PR #5, opened 2024-02-06
Miraclx/p2p exec

xilosada:
Made the use of ipfs nodes conditional, takes a bit to sync anyway, better off setting up a local intermediary node as a boot node

```
$ cargo run -- --home data/node1 init
$ cargo run -- --home data/node2 init --port 2429
$ cargo run -- --home data/node3 init --port 2430
```

This will setup 3 new nodes with differing identities in their respective directories, with zero bootstrap nodes and mDNS turned on for local discovery.

You can directly edit the config files in `data/*/config.toml` to specify the boot nodes for example
Once everything is good, you can start the network
In separate terminals, you can run:

```
$ cargo run -- --home data/node1
$ cargo run -- --home data/node2
$ cargo run -- --home data/node3
```

By default each of these would use mDNS to discover themselves over the local network and everything should just work.
Each node prints a log of active peers every 5 seconds and you can proceed to send messages in real-time from one to the other.

If you want, you can choose to disable mDNS, and opt, instead to use node2 as the bootstrap node (for example).
Add `/ip4/127.0.0.1/tcp/2429/p2p/{PeerId}` to the config files of both node1 and node3
where PeerId is the value in Identity.PeerId of node2
Start node2, then proceed to start nodes 1 and 3, and they should all connect.

Fixed the DHT issue

### DIFF
--- a/chat-p0c/src/network/mod.rs
+++ b/chat-p0c/src/network/mod.rs
@@ -2,6 +2,7 @@ use std::collections::hash_map::{self, HashMap};
 use std::collections::HashSet;
 
 use color_eyre::eyre;
+use color_eyre::owo_colors::OwoColorize;
 use libp2p::futures::prelude::*;
 use libp2p::multiaddr::{self, Multiaddr};
 use libp2p::swarm::behaviour::toggle::Toggle;
@@ -63,22 +64,25 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
                     peer_id: their_peer_id,
                     topic: topic_hash,
                 } => {
-                    info!("Other peer subscribed to {:?}", topic_hash);
-
                     if our_topic_hash == topic_hash {
+                        println!("info: {} joined the chat.", their_peer_id.cyan());
+
                         client
-                            .publish(
-                                our_topic_hash,
-                                format!("Hi {}, I'm {}", their_peer_id, peer_id).into_bytes(),
-                            )
+                            .publish(our_topic_hash, b"Welcome to the chat".to_vec())
                             .await?;
                     }
                 }
                 Event::Message { message, .. } => {
-                    info!(
-                        "Received message from {:?}: {:?}",
-                        message.source,
-                        std::str::from_utf8(&message.data)
+                    println!(
+                        "{}: {}",
+                        match message.source {
+                            Some(peer_id) => peer_id.green().to_string(),
+                            None => "<unknown>".to_owned(),
+                        },
+                        match std::str::from_utf8(&message.data) {
+                            Ok(s) => s,
+                            Err(_) => "<binary>",
+                        }
                     );
                 }
             }


chat-p0c/src/config.rs
```diff
@@ -0,0 +1,230 @@
...
+
+// https://github.com/ipfs/kubo/blob/efdef7fdcfeeb30e2f1ce3dbf65b6460b58afaaf/config/bootstrap_peers.go#L17-L24
+pub const DEFAULT_BOOTSTRAP_NODES: &[&str] = &[
+    "/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN",
+    "/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa",
+    "/dnsaddr/bootstrap.libp2p.io/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb",
+    "/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt",
+    "/ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
+    "/ip4/104.131.131.82/udp/4001/quic-v1/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
+];
```
jknezevic, line 21: it might make sense to create a simple nodes which would advertise themselves as boot nodes, so we can have our own bootstrap nodes, without relying on libp2p public nodes.

chat-p0c/src/cli.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub port: u16,
+
+    /// Enable mDNS discovery
+    #[clap(long, default_value_t = true)]
+    #[clap(overrides_with("no_mdns"))]
+    pub mdns: bool,
+
+    #[clap(long, hide = true)]
+    #[clap(overrides_with("mdns"))]
+    pub no_mdns: bool,
```
jknezevic, line 60: what is the other discovery method, if mdns is set to false? 

i don't particulary like mdns solution as it is not designed to work over internet, and although you can achieve it with having static bootstrap servers, in the future improvements, i'd like to go into direction where we support some other discovery protocols like dht discovery.

chat-p0c/src/network/events/gossipsub.rs
```diff
@@ -0,0 +1,33 @@
...
+                if let Err(_) = self
+                    .event_sender
+                    .send(Event::Subscribed { peer_id, topic })
+                    .await
+                {
+                    error!("Failed to send subscribed event");
+                }
+            }
+            _ => {}
+        }
```
jknezevic, line 31: do we know what are the defaults for full connections and metadata connections in rust libp2p? 

also, does it make sense to limit number of full connections and metadata connections in the gossip sub here?

if we are going with full connections everywhere it comes with certain tradeoffs (in terms of scalability, resource usage, latency, etc). 

it is ok to have full connections everywhere in small network, but the problem grows exponentially when a new peer is added.  

IMO, this is more a comment to think about a possible issue, not something that we should solve immediately for this POC app

xilosada: Yes, good point. Let's address it later

chat-p0c/src/cli.rs
miraclx: I've already implemented DHT discovery, but it requires at least one address to begin discovery from.

If mDNS is activated, that seeds the DHT, and helps discover other nodes.

And when it's turned off, well the only option left is boot nodes.

PR #6, opened 2024-02-06
feat: initial version of a basic rpc to allow sending message through the endpoint instead of terminal stdin

chefsale:
Uploading Screen Recording 2024-02-06 at 23.46.28.mov…

### DIFF
--- a/chat-p0c/src/endpoint.rs
+++ b/chat-p0c/src/endpoint.rs
@@ -1,50 +1,49 @@
 use std::sync::{Arc, Mutex};
 
-use jsonrpc_core::{Error, IoHandler, Params, Value};
-use jsonrpc_http_server::{AccessControlAllowOrigin, DomainsValidation, ServerBuilder};
-use tokio;
+use jsonrpsee_core::async_trait;
+use jsonrpsee_proc_macros::rpc;
+use jsonrpsee_types::ErrorObjectOwned;
 use tracing::info;
 
-pub struct CalimeroRPCHandler {
+#[rpc(server)]
+pub trait CalimeroRPC {
+    #[method(name = "send")]
+    async fn send(&self, message: String);
+
+    #[method(name = "read")]
+    async fn read(&self) -> Result<Option<String>, ErrorObjectOwned>;
+}
+
+pub struct CalimeroRPCImpl {
     mempool: Arc<Mutex<Vec<String>>>,
 }
 
-impl CalimeroRPCHandler {
+impl CalimeroRPCImpl {
     pub fn new() -> Self {
-        CalimeroRPCHandler {
+        CalimeroRPCImpl {
             mempool: Arc::new(Mutex::new(Vec::new())),
         }
     }
+}
 
-    pub async fn send(&self, params: Params) -> Result<Value, Error> {
-        let input: Vec<String> = match params.parse() {
-            Ok(val) => val,
-            Err(_) => return Err(Error::invalid_params("Expected a single string parameter")),
-        };
-
-        if input.len() != 1 {
-            return Err(Error::invalid_params(
-                "Expected exactly one string parameter",
-            ));
-        }
-
+#[async_trait]
+impl CalimeroRPCServer for CalimeroRPCImpl {
+    async fn send(&self, message: String) {
         let mut list = self.mempool.lock().unwrap();
-        list.push(input[0].clone());
+        list.push(message.clone());
 
-        info!("Broadcasting: {}", input[0]);
-
-        Ok(Value::String(format!("Added: {}", input[0])))
+        info!("Broadcasting: {}", message);
     }
 
-    pub async fn read(&self) -> Result<Option<String>, Error> {
+    async fn read(&self) -> Result<Option<String>, ErrorObjectOwned> {
         let mut list = self.mempool.lock().unwrap(); // In real code, handle lock errors
         Ok(list.pop())
     }
 }
 
-impl Clone for CalimeroRPCHandler {
+impl Clone for CalimeroRPCImpl {
     fn clone(&self) -> Self {
-        CalimeroRPCHandler {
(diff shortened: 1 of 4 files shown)


chat-p0c/src/main.rs
```diff
@@ -24,7 +27,7 @@ async fn main() -> eyre::Result<()> {
 pub fn setup() -> eyre::Result<()> {
     tracing_subscriber::registry()
         .with(EnvFilter::builder().parse(format!(
-            "chat_p0c=info,{}",
+            "debug,error,info,{}",
```
miraclx: let's not turn on debug by default, let's leave it opt-in via env `RUST_LOG='chat_p0c=debug'`, and `info > error` so, error is implicitly activated by default.

chat-p0c/src/init.rs
```diff
@@ -63,6 +63,9 @@ pub async fn run(args: cli::RootArgs, init: cli::InitCommand) -> eyre::Result<()
         swarm: SwarmConfig { listen },
         bootstrap: BootstrapConfig { nodes: boot_nodes },
         discovery: DiscoveryConfig { mdns },
+        endpoint: EndpointConfig {
+            port: init.rpc_port,
+        },
```
miraclx: I think I'd lean more towards merging this and `swarm` together to a new `addresses` entry

```toml
[addresses]
api = "127.0.0.1:3030"
swarm = [
    "/ip4/0.0.0.0/tcp/2428",
    "/ip4/0.0.0.0/udp/2428/quic-v1",
    "/ip6/::/tcp/2428",
    "/ip6/::/udp/2428/quic-v1",
]
```

chat-p0c/src/main.rs
```diff
@@ -1,9 +1,12 @@
 use clap::Parser;
 use color_eyre::eyre;
+use jsonrpc_core::*;
+use jsonrpc_http_server::*;
```
miraclx: These are now deprecated in favor of the async-friendly [jsonrpsee](https://github.com/paritytech/jsonrpsee).

chat-p0c/src/config.rs
```diff
@@ -66,6 +69,19 @@ impl Default for DiscoveryConfig {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct EndpointConfig {
+    pub port: u16,
```
miraclx, line 76: I think the `host` addr should be configurable as well.

chat-p0c/src/network/mod.rs
```diff
@@ -37,12 +42,40 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+            .start_http(
+                &format!("127.0.0.1:{}", config.endpoint.port)
+                    .parse()
+                    .unwrap(),
+            )
+            .expect("Unable to start JSON-RPC server");
 
+        info!("RPC Server running on 127.0.0.1:{}", config.endpoint.port);
+
+        server.wait();
```
miraclx: Blocking is a weird pattern here, this library internally uses async-compatible dependencies, but they're exposing a sync-API

thankfully the `jsonrpsee` crate is much better at this

chat-p0c/src/network/mod.rs
```diff
@@ -37,12 +42,40 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+    });
+
+    tokio::task::spawn_blocking(move || {
+        let server = ServerBuilder::new(io)
+            .threads(3)
+            .start_http(
+                &format!("127.0.0.1:{}", config.endpoint.port)
+                    .parse()
+                    .unwrap(),
+            )
```
miraclx: Can we also make this into a service when we port to jsonrpsee

I see it's possible here - https://github.com/paritytech/jsonrpsee/blob/8470f2bb8d2a2b5e3c5cfd07f138aeab9dfbe022/examples/examples/jsonrpsee_as_service.rs#L178-L181

We can use `axum` as the server framework, and delegate route handling to this JSONRPC service

Blobby has an example: https://github.com/calimero-is-near/PortalApps/blob/97adddee03ebf2b21a76d7e22101237f12c8fa86/services/blobby/src/main.rs#L94

chat-p0c/src/network/mod.rs
```diff
@@ -13,8 +13,13 @@ use tokio::sync::{mpsc, oneshot};
 use tokio::time;
 use tracing::{debug, info, trace, warn};
 
+use jsonrpc_core::{IoHandler, Params};
+use jsonrpc_http_server::ServerBuilder;
+use std::sync::{Arc, Mutex};
```
miraclx: PS: let's organize these imports, `std` at the top, crate deps after

chat-p0c/src/endpoint.rs
```diff
@@ -0,0 +1,50 @@
+use jsonrpc_core::{Error, IoHandler, Params, Value};
+use jsonrpc_http_server::{AccessControlAllowOrigin, DomainsValidation, ServerBuilder};
+use std::sync::{Arc, Mutex};
```
miraclx, line 1: here too, `std` on top, separate from crate deps

chat-p0c/src/init.rs
chefsale: i'm leaning more towards 
```
[endpoints]
rpc = "127.0.0.1:3030"
bootstrap = [
   "/ip4/0.0.0.0/tcp/2428",
    "/ip4/0.0.0.0/udp/2428/quic-v1",
    "/ip6/::/tcp/2428",
    "/ip6/::/udp/2428/quic-v1",
]
```

miraclx: That's okay, just a correction - bootstrap goes in a separate section to accommodate other potential configs for that specifically.

So that'll be `swarm`

PR #8, opened 2024-02-07
feat(identity-rs):  add create identity

MatejVukosav:
- Initial setup of identity repo
- method to create an identity from the public key 
- types following document structure

In example I am creating keypair with the same method as in chat-p0c

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/
+res/
 data/
 target/
 .DS_Store
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
- "gimli",
+ "gimli 0.28.1",
 ]
 
 [[package]]
@@ -54,6 +54,17 @@ dependencies = [
 
 [[package]]
 name = "ahash"
+version = "0.7.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
+dependencies = [
+ "getrandom 0.2.12",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
+name = "ahash"
 version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77c3a9648d43b9cd48db467b3f87fdd6e146bcc88ab0180006cef2179fe11d01"
@@ -72,6 +83,12 @@ checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
 
 [[package]]
 name = "alloc-no-stdlib"
@@ -621,6 +638,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed570934406eb16438a4e976b1b4500774099c13b8cb96eec99f620f05090ddf"
 
 [[package]]
+name = "bitvec"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bc2832c24239b0141d5674bb9174f9d68a8b5b3f2753311927c172ca46f7e9c"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "blake2"
 version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -661,6 +690,30 @@ dependencies = [
  "futures-lite 2.2.0",
  "piper",
  "tracing",
+]
+
+[[package]]
+name = "borsh"
+version = "1.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f58b559fd6448c6e2fd0adb5720cd98a2506594cafa4737ff98c396f3e82f667"
+dependencies = [
+ "borsh-derive",
+ "cfg_aliases",
+]
+
+[[package]]
+name = "borsh-derive"
+version = "1.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7aadb5b6ccbd078890f6d7003694e33816e6b784358f18e15e7e6d9f065a57cd"
+dependencies = [
+ "once_cell",
+ "proc-macro-crate",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.48",
+ "syn_derive",
 ]
 
 [[package]]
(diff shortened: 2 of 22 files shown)


identity-rs/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+libp2p = "0.53.2"
+libc = "0.2.110"
+didkit-cli = "0.3.2"
+serde_json = "1.0.113"
+sha2 = "0.10.8"
+ed25519-dalek = "2.1.1"
+multibase = "0.9.1"
+hex = "0.4.3"
+bs58 = "0.5.0"
+serde = "1.0.196"
```
miraclx: A bunch of the deps here aren't used, can you clean them up?

identity-rs/README.md
```diff
@@ -0,0 +1,5 @@
+
+
+# Decentralized identity, also referred to as self-sovereign identity, is an open-standards based identity framework that uses digital identifiers and verifiable credentials that are self-owned, independent, and enable trusted data exchange.
```
miraclx:
```suggestion
# Calimero Identity

Decentralized identity also referred to as self-sovereign identity, is an open-standards-based identity framework that uses digital identifiers and verifiable credentials that are self-owned, independent, and enable trusted data exchange.
```

identity-rs/src/identity_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+///       "publicKeyMultibase": "zCovLVG4fQcqT8sDqj76uQuXtQU2LqABAf6X8vnDW36zAidisFK22Z5Ecm28apKb9Kg6ofRo",
+///       "controller": "did:cali:12D3KooWLU42rBMLrzAyFg8CdpHiPcwJcVsKv9Wx9DtiT4QjPGGV"
+///     }
+///   ]
+/// }
+pub fn create_identity(authentication: Authentication) -> DidDocument {
+    let public_key_id = authentication.public_key.to_peer_id().to_base58();
+    let multibase_encoded = encode(Base::Base58Btc, &public_key_id);
+
+    let did = DID_CALI_IDENTIFIER.to_string() + &public_key_id;
```
miraclx:
```suggestion
    let did = format!("{}{}", DID_CALI_IDENTIFIER, public_key_id);
```

identity-rs/src/identity_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+///       "id": "did:cali:12D3KooWLU42rBMLrzAyFg8CdpHiPcwJcVsKv9Wx9DtiT4QjPGGV",
+///       "type": "Ed25519",
+///       "publicKeyMultibase": "zCovLVG4fQcqT8sDqj76uQuXtQU2LqABAf6X8vnDW36zAidisFK22Z5Ecm28apKb9Kg6ofRo",
+///       "controller": "did:cali:12D3KooWLU42rBMLrzAyFg8CdpHiPcwJcVsKv9Wx9DtiT4QjPGGV"
+///     }
+///   ]
+/// }
+pub fn create_identity(authentication: Authentication) -> DidDocument {
+    let public_key_id = authentication.public_key.to_peer_id().to_base58();
+    let multibase_encoded = encode(Base::Base58Btc, &public_key_id);
```
miraclx: [`PeerId` impls `AsRef<Multihash<u64>>`](https://docs.rs/libp2p-identity/0.2.8/libp2p_identity/struct.PeerId.html#impl-AsRef%3CMultihash%3C64%3E%3E-for-PeerId) and [`Multihash<u64>` has a method `to_bytes`](https://docs.rs/multihash/0.19.1/multihash/struct.Multihash.html#method.to_bytes) that's non-consuming.

Also, [`multibase::encode`](https://docs.rs/multibase/0.9.1/multibase/fn.encode.html) uses the `base` you selected to encode the `input`, which is effectively the same as what you already did with [`.to_base58()`](https://docs.rs/libp2p-identity/0.2.8/src/libp2p_identity/peer_id.rs.html#119-121). So it becomes `base58(base58())`.

Factoring all these, we can instead write this as:

```suggestion
    let public_key_id = authentication.public_key.to_peer_id();
    let multibase_encoded = encode(Base::Base58Btc, &public_key_id.as_ref().to_bytes());
```

We would've been able to simplify to..

```diff
-    let multibase_encoded = encode(Base::Base58Btc, &public_key_id.as_ref().to_bytes());
+    let multibase_encoded = encode(Base::Base58Btc, &public_key_id.to_bytes());
```

.. but unfortunately, the [`PeerId::to_bytes`](https://docs.rs/libp2p-identity/0.2.8/libp2p_identity/struct.PeerId.html#method.to_bytes) method unnecessarily consumes the `PeerId`.

identity-rs/src/identity_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let did = DID_CALI_IDENTIFIER.to_string() + &public_key_id;
+
+    let verification_method: VerificationMethod = VerificationMethod {
+        id: did.clone()+"#key1",
```
miraclx:
```suggestion
        id: format!("{}#key1", did),
```

identity-rs/src/identity_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+        algorithm_type: authentication.algorithm.to_string(),
+        public_key_multibase: multibase_encoded,
+        controller: authentication.controller.unwrap_or(did.clone()),
```
miraclx: Allocate only when absolutely necessary

```suggestion
        controller: authentication.controller.unwrap_or_else(|| did.clone()),
```

identity-rs/src/identity_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+    let did_document: DidDocument = DidDocument {
+        id: did.clone(),
```
miraclx: There are no further consumers of `did` so we can fully consume it here.

```suggestion
        id: did,
```

identity-rs/.gitignore
```diff
@@ -0,0 +1,3 @@
+.idea/
+data/
+target/
```
miraclx: PS: this is duplicated from the root

identity-rs/src/identity_provider.rs
```diff
@@ -0,0 +1,62 @@
...
+        verification_method: vec![verification_method],
+    };
+    //TODO store it
+    return did_document;
```
miraclx: implicit returns :-)

```suggestion
    did_document
```

identity-rs/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+didkit = "0.6.0"
+libp2p = "0.53.2"
+serde_json = "1.0.113"
+sha2 = "0.10.8"
+multibase = "0.9.1"
+serde = "1.0.196"
```
miraclx: Can you move these definitions to the root `Cargo.toml`, you can reference them with `<package>.workspace = true`

identity-rs/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub algorithm_type: String,
+    #[serde(rename = "publicKeyMultibase")]
+    pub public_key_multibase: String,
+    pub controller: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DidDocument {
+    pub id: String,
+    #[serde(rename = "verificationMethod")]
```
miraclx: For these kinds of patterns, we can use `#[serde(rename_all = "camelCase")]` on the struct itself.

identity-rs/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SerializedPublicKey(Vec<u8>);
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct VerificationMethod {
+    pub id: String,
+    #[serde(rename = "type")]
+    pub algorithm_type: String,
+    #[serde(rename = "publicKeyMultibase")]
```
miraclx, line 12: here as well

identity-rs/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+
+fn main() {
+    //generate keypair in any way
+    let keypair = Keypair::generate_ed25519();
+    let public_key = keypair.public();
+
+    let authentication: Authentication = Authentication {
+        algorithm: AlgorithmType::Ed25519,
+        controller: None,
+        public_key: public_key.clone(),
```
miraclx:
```suggestion
        public_key,
```

identity-rs/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "identity-rs"
```
miraclx: Let's remove the `-rs` here in the crate name

identity-rs/src/types.rs
MatejVukosav: I cannot use it here as `algorithm_type` needs to be `type`, and `type` itself is not allowed.

miraclx: That's permissible, retain the explicit rename at that point, the pattern would be:

```rust
pub struct VerificationMethod {
    pub id: String,
    #[serde(rename = "type")]
    pub algorithm_type: String,
    pub public_key_multibase: String,
```

The general idea is "default to camelCase", but use a custom-defined name in certain cases.

But as-is is fine as well.

PR #12, opened 2024-02-09
feat(runtime): impl node runtime with wasmer4 backend

miraclx:
I have been up 30 hours at this point, and typing this doc seems like the hardest thing 😅 

This patch introduces the foundation of the runtime, with 3 new crates

- `vm-runner`, `calimero-sdk` and a `kv-store` app contract

The VM runner hosts all the fancy magic for executing Calimero apps on a node

The `calimero-sdk` hosts the bindings to the FFI layer for which we plug into from the VM (eventually, this library should aid further abstraction making it easier to write Calimero apps)

The `kv-store` crate is an example of a compilable contract that runs in our VM

To test this out, first compile the `kv-store` contract

```console
$ ./apps/kv-store/build.sh
    Finished app-release [optimized] target(s) in 0.16s

$ tree ./apps/kv-store/res
./apps/kv-store/res
└── kv_store.wasm
```

You can then proceed to run the `demo` example as part of the VM runner that sources this example code and runs the functions

CleanShot 2024-02-09 at 23 47 20@2x


simple-wasmer/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
MatejVukosav: Isn't this duplicated from root?

miraclx: @ijerkovic, I can remove these crates, yeah?

PR #17, opened 2024-02-19
initial coordinator implementation

medicz:


Cargo.toml
```diff
@@ -19,7 +19,7 @@ fragile = "2.0.0"
 libp2p = "0.53.2"
 ouroboros = "0.18.3"
 owo-colors = "3.5.0"
-serde = "1.0.196"
+serde = { version = "1.0.196", features = ["derive"] }
```
miraclx: let's localize this definition to the crate

chat-p0c/Cargo.toml
```diff
@@ -21,6 +21,8 @@ jsonrpsee-ws-client = { version = "0.22", default-features = false }
 libp2p = { workspace = true, features = ["full"] }
 once_cell = "1.19.0"
 serde.workspace = true
```
miraclx: this instead

```suggestion
serde = { workspace = true, features = ["derive"] }
```

chat-p0c/src/network/mod.rs
```diff
@@ -16,6 +17,10 @@ use tokio::sync::{mpsc, oneshot};
 use tokio::time;
 use tracing::{debug, info, trace, warn};
 
+use sha2::{Sha256, Digest};
```
miraclx: let's merge these (remove the breaking space at line 19)

```suggestion
use tracing::{debug, info, trace, warn};
use sha2::{Sha256, Digest};
```

chat-p0c/src/network/mod.rs
```diff
@@ -16,6 +17,10 @@ use tokio::sync::{mpsc, oneshot};
...
+use serde::{Deserialize, Serialize};
+use serde_json;
```
miraclx:
```suggestion
```

chat-p0c/src/network/mod.rs
```diff
@@ -35,10 +40,31 @@ struct Behaviour {
...
 
+struct Storage {
+    transactions: Arc<Mutex<HashMap<Hash, Transaction>>>,
+    confirmations: Arc<Mutex<HashMap<Hash, Transaction>>>,
+    senders: Arc<Mutex<HashMap<Hash, PeerId>>>,
+    last_known_transaction_hash: Arc<Mutex<Hash>>,
+    nonce: Arc<Mutex<u64>>,
+    // true if coordinator
+    // TODO proper implementation should use enum
+    node_type: bool,
```
miraclx: Since we can already classify types of these, let's just switch to enums at this point

chat-p0c/src/cli.rs
```diff
@@ -21,6 +21,9 @@ pub struct RootArgs {
     #[clap(long, value_name = "PATH", default_value_t = config::default_chat_dir())]
     #[clap(env = "CALIMERO_CHAT_HOME", hide_env_values = true)]
     pub home: camino::Utf8PathBuf,
+
+    #[clap(long, value_name = "TYPE", default_value = "peer")]
+    pub node_type: String,
```
miraclx: this should work:

```suggestion
    #[clap(long, value_name = "TYPE")]
    #[clap(value_enum, default_value_t)]
    pub node_type: NodeType,
```

and then;

```rust
#[derive(Clone, Debug, Default, ValueEnum)]
pub enum NodeType {
    #[default]
    Peer,
    Coordinator,
}
```

miraclx: Also, at this point, I think it may be wise to introduce a new scope for these types of arguments like `chat-p0c run` and leave `--home` with global args.

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
-
             }
-            /*
-            pop_result = handler.read() => {
-                match pop_result {
-                    Ok(Some(value)) => {
+            line = stdin.next_line() => {
+                match line {
+                    Ok(Some(line)) => {
+                        if storage.node_type {
```
miraclx: once we complete the migration to the `NodeType` enum, we can add a convenience method perhaps for this kinds of checks. `node_type.is_coordinator()`

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                            println!("Coordinator can not create transactions!");
```
miraclx: `error!` or perhaps just `info!`? up to you

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+    let mut transactions_mutex = storage.transactions.lock().unwrap();
+    transactions_mutex.insert(transaction_hash.clone(), transaction);
+    if let Some(peer_id) = sender {
+        storage.senders.lock().unwrap().insert(transaction_hash.clone(), peer_id);
+    }
+
+    transaction_hash
+}
+
+async fn event_recipient(mut client: Client, our_topic_hash: gossipsub::TopicHash, event: Event, storage: &Storage) -> Result<(), eyre::Report> {
```
miraclx:
```suggestion
async fn event_recipient(mut client: Client, our_topic_hash: gossipsub::TopicHash, event: Event, storage: &Storage) -> eyre::Result<()> {
```

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                    println!("info: {} joined the chat.", their_peer_id.cyan());
+
+//                    client
+//                        .publish(our_topic_hash, serde_json::to_vec(&create_transaction("Welcome to the chat", &storage)).unwrap())
+//                        .await?;
+                }
+            }
+            Event::Message { message, .. } => {
+                let source = message.source;
+                let message: NetworkAction = serde_json::from_slice(&message.data).unwrap();
```
miraclx: let's take advantage of the type-erased errors, and avoid potential panics anywhere

```suggestion
                let message: NetworkAction = serde_json::from_slice(&message.data)?;
```

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                    NetworkAction::Transaction(transaction) => {
+                        let transaction_hash = store_transaction(transaction, storage, source);
+
+                        if storage.node_type {
+                            let mut nonce_mutex = storage.nonce.lock().unwrap();
+                            let confirmation = NetworkAction::TransactionConfirmation(TransactionConfirmation{
+                                nonce: *nonce_mutex,
+                                transaction_hash: transaction_hash.clone(),
+                                // TODO proper confirmation hash
+                                confirmation_hash: transaction_hash.clone(),
```
miraclx: last use of `transaction_hash` we don't need to clone it

```suggestion
                                // TODO proper confirmation hash
                                confirmation_hash: transaction_hash.clone(),
                                transaction_hash,
```

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
                 }
             }
-            */
         }
     }
 
     Ok(())
 }
 
+fn store_transaction(transaction: Transaction, storage: &Storage, sender: Option<PeerId>) -> Hash {
```
miraclx:
```suggestion
fn store_transaction(transaction: Transaction, storage: &Storage, sender: Option<PeerId>) -> eyre::Result<Hash> {
```

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +111,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                                Err(_) => "<binary>",
+                            }
+                        )
+                    }
+                    _ => println!("UNKNOWN"),
+                }
+            }
+        }
+
+        Ok::<_, eyre::Report>(())
```
miraclx: Since it's now extracted into a function with an explicitly-defined signature, this can be simplified

```suggestion
        Ok(())
```

chat-p0c/src/network/mod.rs
```diff
@@ -555,3 +636,58 @@ pub(crate) enum Event {
         message: gossipsub::Message,
     },
 }
+
+type Hash = Vec<u8>;
+
+fn hash<T: Serialize>(item: &T) -> Hash {
```
miraclx:
```suggestion
fn hash<T: Serialize>(item: &T) -> eyre::Result<Hash> {
```

chat-p0c/src/network/mod.rs
```diff
@@ -35,11 +38,29 @@ struct Behaviour {
...
     if !Config::exists(&args.home) {
         eyre::bail!("chat node is not initialized in {:?}", args.home);
     }
 
+    let storage = Storage {
+        transactions: Arc::new(Mutex::new(HashMap::new())),
+        confirmations: Arc::new(Mutex::new(HashMap::new())),
+        senders: Arc::new(Mutex::new(HashMap::new())),
+        last_known_transaction_hash: Default::default(),
+        nonce: Arc::new(Mutex::new(1)),
```
miraclx: How about start with 0, then incr when handling transaction not after it.

Makes it so it represents active state instead of holding a nonce for the next transaction.

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +106,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                            error!("Coordinator can not create transactions!");
+                            continue;
+                        }
                         if client.mesh_peer_count(topic.hash()).await == 0 {
                             info!("No connected peers to send message to.");
                             continue;
                         }
                         client
-                            .publish(topic.hash(), value.into_bytes())
+                            .publish(topic.hash(), serde_json::to_vec(&create_transaction(&line, &storage, peer_id)?).unwrap())
```
miraclx: Same concern with the unwraps

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +106,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
-            */
         }
     }
 
     Ok(())
 }
 
+fn store_transaction(transaction: Transaction, storage: &Storage, sender: Option<PeerId>) -> eyre::Result<Hash> {
+    let transaction_hash = hash(&transaction)?;
+    let mut transactions_mutex = storage.transactions.lock().unwrap();
```
miraclx: Please grep and replace all instances

medicz: std::sync::MutexGuard<'_, HashMap<Vec<u8>, Transaction>>` cannot be sent between threads safely when replacing with 

`let mut transactions_mutex = storage.transactions.lock()?;`

miraclx: Try force it to it's error representation ahead of time, perhaps `.map_err(|guard| eyre::eyre!(guard))?`

chat-p0c/src/network/mod.rs
```diff
@@ -14,9 +15,11 @@ use libp2p::{gossipsub, identify, kad, mdns, ping, relay, PeerId};
 use tokio::io::AsyncBufReadExt;
 use tokio::sync::{mpsc, oneshot};
 use tokio::time;
-use tracing::{debug, info, trace, warn};
+use tracing::{debug, info, trace, warn, error};
+use sha2::{Sha256, Digest};
+use serde::{Deserialize, Serialize};
```
miraclx: can you run `cargo fmt`? `s-` should preceed `t-`

chat-p0c/src/network/mod.rs
```diff
@@ -35,11 +38,29 @@ struct Behaviour {
     ping: ping::Behaviour,
 }
 
+struct Storage {
+    transactions: Arc<Mutex<HashMap<Hash, Transaction>>>,
+    confirmations: Arc<Mutex<HashMap<Hash, Transaction>>>,
+    senders: Arc<Mutex<HashMap<Hash, PeerId>>>,
+    last_known_transaction_hash: Arc<Mutex<Hash>>,
+    nonce: Arc<Mutex<u64>>,
```
miraclx: Also, at this point, we can only handle one network event at a time, so we don't need this synchronization primitive.

It's guaranteed there will only be one user (r/w) at a time.

And going forward, we should keep it this way. And avoid the overhead of understanding the entire system to make contributions, it'll get real complex, real fast.

Actors simplify things a lot, or we put this on the wall at the office too haha: https://bholley.net/blog/2015/must-be-this-tall-to-write-multi-threaded-code.html

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +106,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+//                        .await?;
+                }
+            }
+            Event::Message { message, .. } => {
+                let source = message.source;
+                let message: NetworkAction = serde_json::from_slice(&message.data)?;
+
+                match message {
+                    NetworkAction::Transaction(transaction) => {
+                        let transaction_hash = store_transaction(transaction, storage, source)?;
```
miraclx: Eventually we should be able to garbage collect, yeah? We can leave this for a future PR

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +106,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                        info!("Confirmation -> nonce: {}, transaction_hash: {:02X?}", confirmation.nonce, confirmation.transaction_hash);
+                        let src = if let Some(peer_id) = storage.senders.lock().unwrap().get(&confirmation.transaction_hash) {
+                            peer_id.green().to_string()
+                        } else {
+                            "<unknown>".to_owned()
+                        };
+                        let payload = if let Some(transaction) = storage.transactions.lock().unwrap().get(&confirmation.transaction_hash) {
+                            transaction.clone().payload
+                        } else {
+                            Vec::new()
```
miraclx: We can just take a slice here, no need to clone

```suggestion
                            &transaction.payload[..]
                        } else {
                            &[]
```

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +106,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+                        };
+                        println!(
+                            "{}: {}",
+                            src,
+                            match std::str::from_utf8(&payload) {
```
miraclx:
```suggestion
                            match std::str::from_utf8(payload) {
```

chat-p0c/src/network/mod.rs
```diff
@@ -85,75 +106,130 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+        }
+
+        Ok(())
+}
+
+fn create_transaction(message: &str, storage: &Storage, peer_id: PeerId) -> eyre::Result<NetworkAction> {
+    let transaction = Transaction {
+        method: String::from("send_message"),
+        payload: message.as_bytes().to_vec(),
+        last_known_transaction_hash: (*storage.last_known_transaction_hash.lock().unwrap().clone()).to_vec(),
```
miraclx: Double allocation here

```suggestion
        last_known_transaction_hash: storage.last_known_transaction_hash.lock().unwrap().clone(),
```

chat-p0c/src/network/mod.rs
```diff
@@ -143,87 +143,119 @@ pub async fn run(args: cli::RootArgs) -> eyre::Result<()> {
...
+    transaction: Transaction,
+    storage: &Storage,
+    sender: Option<PeerId>,
+) -> eyre::Result<Hash> {
     let transaction_hash = hash(&transaction)?;
-    let mut transactions_mutex = storage.transactions.lock().unwrap();
+    let mut transactions_mutex = storage
+        .transactions
+        .lock()
+        .map_err(|guard| eyre::eyre!(format!("{:?}", guard)))?;
```
miraclx: could be simplified, but this is okay

```suggestion
        .map_err(|guard| eyre::eyre!("{:?}", guard))?;
```

chat-p0c/src/network/mod.rs
miraclx: btw, this comment was in reference to all uses of `Arc<Mutex<..>>` not just `nonce`.

medicz: Please expand explanation why is that the case, not sure I am following.

miraclx: We can alleviate the need to deal with lock contention by designing around avoiding potentially shared mutability.

At this point, it is guaranteed we'll only need to deal with one message at a time from the network. So at the moment, we can avoid needing this synchronization primitive.

And going forward, I'd like to redesign the networking stack to take advantage of the actor model similar to what nearcore does. See https://github.com/actix/actix.

PR #18, opened 2024-02-21
feat: implement websocket server

fbozic:
- implement axum websocket service for the subscriptions
- integrate websocket service with the existing server
- emit `TransactionExecuted` and `PeerJoined` events from node

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -369,7 +369,7 @@ impl Node {
         method: String,
         payload: Vec<u8>,
     ) -> eyre::Result<calimero_runtime::logic::Outcome> {
-        self.execute(None, method, payload, false).await
+        self.execute(None, method, payload).await
     }
 
     pub async fn call_mut(
@@ -424,7 +424,7 @@ impl Node {
         };
 
         let outcome = self
-            .execute(Some(hash), transaction.method, transaction.payload, true)
+            .execute(Some(hash), transaction.method, transaction.payload)
             .await?;
 
         if let Some(sender) = outcome_sender {
@@ -439,12 +439,10 @@ impl Node {
         hash: Option<calimero_primitives::hash::Hash>,
         method: String,
         payload: Vec<u8>,
-        writes: bool,
     ) -> eyre::Result<calimero_runtime::logic::Outcome> {
-        let mut storage = if writes {
-            TemporalRuntimeStore::Write(calimero_store::TemporalStore::new(&self.store))
-        } else {
-            TemporalRuntimeStore::Read(calimero_store::ReadOnlyStore::new(&self.store))
+        let mut storage = match hash {
+            Some(_) => TemporalRuntimeStore::Write(calimero_store::TemporalStore::new(&self.store)),
+            None => TemporalRuntimeStore::Read(calimero_store::ReadOnlyStore::new(&self.store)),
         };
 
         let limits = calimero_runtime::logic::VMLimits {


Cargo.toml
```diff
@@ -16,18 +26,25 @@ clap = "4.4.18"
...
 libp2p = "0.53.2"
 ouroboros = "0.18.3"
 owo-colors = "3.5.0"
 serde = "1.0.196"
 serde_json = "1.0.113"
 thiserror = "1.0.56"
 tokio = "1.35.1"
+tokio-stream = { version = "0.1.14", features = ["signal"] }
+tokio-util = { version = "0.7.10", features = ["rt"] }
+tokio-tungstenite = { version = "0.20.1" }
```
miraclx: Let's localize these feature specifications to the dependent crate

calimero-api/src/ws.rs
```diff
@@ -0,0 +1,187 @@
+use std::collections::HashMap;
+use std::net::SocketAddr;
+use std::sync::{
+    atomic::{AtomicU32, Ordering},
+    Arc,
+};
+
+use color_eyre::eyre;
+use futures_util::{SinkExt, StreamExt, TryFutureExt};
+use serde_json;
```
miraclx:
```suggestion
```

calimero-api/src/ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+pub type ClientsState = Arc<RwLock<HashMap<api::WsClientId, mpsc::Sender<api::WsCommand>>>>;
+
+static NEXT_CLIENT_ID: AtomicU32 = AtomicU32::new(1);
+
+pub async fn start(
+    addr: SocketAddr,
+    cancellation_token: CancellationToken,
+    clients: ClientsState,
+    controller_tx: mpsc::Sender<controller::Command>,
```
miraclx: `controller::Command` reads really well, thank you for this design. 👍🏽

calimero-api/src/ws.rs
```diff
@@ -0,0 +1,187 @@
...
+        cancellation_token.cancelled().await;
+        tracing::info!("agraceful api shutdown initiated");
+        futures_util::stream::iter(shutdown_clients.write().await.drain())
+            .for_each_concurrent(None, |(client_id, client)| async move {
+                let command = api::WsCommand::Close(
+                    protocol::frame::coding::CloseCode::Away,
+                    String::from("Server shuting down"),
+                );
+                if let Err(e) = client.send(command).await {
+                    tracing::error!(
```
miraclx: it's okay to bring these in scope, and use `error!` directly. there shouldn't be any ambiguity about these.

calimero-api/src/ws.rs
```diff
@@ -0,0 +1,187 @@
...
+    tracing::info!("api started");
+    server.await;
+}
+
+async fn client_connected(
+    ws: warp::ws::WebSocket,
+    clients: ClientsState,
+    controller_tx: mpsc::Sender<controller::Command>,
+) {
+    let client_id = NEXT_CLIENT_ID.fetch_add(1, Ordering::Relaxed);
```
miraclx: I'd replace `ClientsState` with a struct like `ServerState` that includes the map of `ClientId` <-> `mpsc::Sender`.

`ClientId` could be determined randomly akin to https://github.com/calimero-is-near/PortalApps/blob/97adddee03ebf2b21a76d7e22101237f12c8fa86/services/blobby/src/main.rs#L147

calimero-api/src/ws.rs
```diff
@@ -0,0 +1,187 @@
...
+    tracing::info!("new client connected(client_id={})", client_id);
+
+    let (mut ws_tx, mut ws_rx) = ws.split();
+
+    let (tx, rx) = mpsc::channel::<api::WsCommand>(32);
+    let mut rx = ReceiverStream::new(rx);
```
miraclx: Converting this to a stream would only be necessary if the consumer was a stream.

But the receiver is a manually constructed loop over each item...

calimero-api/src/ws.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    tokio::task::spawn(async move {
+        while let Some(command) = rx.next().await {
```
miraclx: Meaning you can call `rx.recv().await` here

calimero-controller/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use futures_util::StreamExt;
+use tokio_stream::wrappers::ReceiverStream;
+use tokio_util::sync::CancellationToken;
+
+use calimero_api::ws;
+use calimero_primitives::api;
+use calimero_primitives::app;
+use calimero_primitives::controller;
+use subscriptions::Subscriptions;
```
miraclx:
```suggestion
use futures_util::StreamExt;
use tokio_stream::wrappers::ReceiverStream;
use tokio_util::sync::CancellationToken;

use calimero_api::ws;
use calimero_primitives::api;
use calimero_primitives::app;
use calimero_primitives::controller;

mod subscriptions;

use subscriptions::Subscriptions;
```

calimero-primitives/src/api.rs
```diff
@@ -0,0 +1,71 @@
...
+pub enum ApiResponse {
+    ListRemoteApps(Vec<app::App>),
+    ListInstalledApps(Vec<app::InstalledApp>),
+    GetInstalledApp(app::InstalledApp),
+    InstallBinaryApp(app::InstalledAppId),
+    InstallRemoteApp(app::InstalledAppId),
+    UninstallApp(app::InstalledAppId),
+    Subscribe(app::InstalledAppId),
+    Unsubscribe(app::InstalledAppId),
+    UnsubscribeFromAll(),
```
miraclx:
```suggestion
    UnsubscribeFromAll,
```

calimero-peer/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+        config.websocket_api.get_socket_addr()?,
+        token.clone(),
+        clients.clone(),
+        controller_tx.clone(),
+    ));
+
+    signal::ctrl_c().await?;
+    token.cancel();
+    tracker.close();
+    tracker.wait().await;
```
miraclx:
```suggestion
    tokio::select! {
        _ = signal::ctrl_c() => {
            token.cancel();
            tracker.close();
        _ = tracker.wait() => {}
    };
```

calimero-controller/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    command: controller::Command,
+) {
+    match command {
+        controller::Command::WsApiRequest(client_id, request_id, request) => {
+            let response = match request {
+                api::ApiRequest::ListRemoteApps() => handle_list_remote_apps().await,
+                api::ApiRequest::ListInstalledApps() => todo!(),
+                api::ApiRequest::InstallBinaryApp(_) => todo!(),
+                api::ApiRequest::InstallRemoteApp(_) => todo!(),
+                api::ApiRequest::UninstallApp(_) => todo!(),
```
miraclx: I understand this is still WIP but I'd prefer if we just printed an unimplemented message as opposed to panicking from valid code paths before merging to master.

calimero-api/src/ws.rs
fbozic: Left over from official example. Removed stream wrapper now.

Cargo.toml
fbozic: Fixed

calimero-peer/src/main.rs
fbozic: Are you sure this is correct? Once I close the tracker I need to wait for it before main finishes (at least that is how I understand). Here are the docs: https://tokio.rs/tokio/topics/shutdown

calimero-controller/Cargo.toml
```diff
@@ -7,12 +7,12 @@ repository.workspace = true
...
 
 [dependencies]
-futures-util.workspace = true
+futures-util = { workspace = true, default-features = false, features = ["sink"] }
 serde.workspace = true
 serde_json.workspace = true
 tokio.workspace = true
-tokio-stream.workspace = true
-tokio-util.workspace = true
+tokio-stream = { workspace = true, features = ["signal"] }
```
miraclx: Doesn't seem like this feature flag is used.

calimero-controller/Cargo.toml
```diff
@@ -7,12 +7,12 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
-futures-util.workspace = true
+futures-util = { workspace = true, default-features = false, features = ["sink"] }
```
miraclx: `sink` isn't used for controller either, and it's the same with peer I believe

calimero-primitives/src/api.rs
miraclx: This was just one example, there are other cases. It's perfectly fine to use as-is, but it's just a nice little thing for simplicity at the consumer level.

calimero-peer/src/main.rs
miraclx: Ah, I thought `tracker.wait()` could resolve on it's own without calling `close()`

Say, when both futures complete on their own.

Turns out `close` has to be called **and** the futures must have finished.

Yeah, if the only way out of this is to interrupt, then the previous logic is what we should do.

Sorry for the confusion.

calimero-controller/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+                    subscriptions.subscribe(installed_app_id, client_id);
+                    api::ApiResponse::Subscribe(installed_app_id)
+                }
+                api::ApiRequest::Unsubscribe(installed_app_id) => {
+                    subscriptions.unsubscribe(installed_app_id, client_id);
+                    api::ApiResponse::Unsubscribe(installed_app_id)
+                }
+                api::ApiRequest::UnsubscribeFromAll() => {
+                    subscriptions.unsubscribe_from_all(client_id);
+                    api::ApiResponse::UnsubscribeFromAll()
```
frdomovic:
```suggestion
                    api::ApiResponse::UnsubscribeFromAll
```

calimero-api/src/ws.rs
fbozic: Done

calimero-api/src/ws.rs
fbozic: Done a bit differently. Everything now is part of `ws` module.

calimero-primitives/src/api.rs
fbozic: Done everywhere.

calimero-controller/src/lib.rs
fbozic: Done

Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 [workspace]
 resolver = "2"
 members = [
+  "./apps/kv-store",
```
miraclx: `cargo` doesn't complain about this? weird, it's duplicated though

crates/node/src/cli/init.rs
```diff
@@ -50,6 +50,11 @@ pub struct InitCommand {
     #[clap(default_value_t = calimero_server::config::DEFAULT_PORT)]
     pub server_port: u16,
 
+    /// Port to listen on for Websocket
+    #[clap(long, value_name = "WEBSOCKET_PORT")]
+    #[clap(default_value_t = calimero_server::config::DEFAULT_PORT)]
+    pub server_websocket_port: u16,
```
miraclx: use the same server+port as everything else. we can keep them scoped out with the paths

`/graphql`, `/ws` etc

crates/node/src/cli/run.rs
```diff
@@ -49,10 +49,7 @@ impl RunCommand {
                 bootstrap: config.network.bootstrap,
                 discovery: config.network.discovery,
             },
-            server: calimero_server::config::ServerConfig {
-                listen: config.network.server.listen,
-                graphql: config.network.server.graphql,
-            },
+            server: config.network.server,
```
miraclx, line 52: 👍🏽

crates/server/src/websocket.rs
```diff
@@ -0,0 +1,229 @@
...
+
+use calimero_primitives::api;
+use subscriptions::Subscriptions;
+
+mod subscriptions;
+
+type ConnectionsState = Arc<RwLock<HashMap<api::WsClientId, mpsc::Sender<api::WsCommand>>>>;
+type SubscriptionsState = Arc<RwLock<Subscriptions>>;
+
+pub async fn start(addr: SocketAddr) -> () {
```
miraclx: can we write this in the format of how the graphql service is written, returning a service that can be mounted on our router

so we have one server, with a router that has multiple services, this being one of them

so we don't have to listen on a new port just for websockets

Cargo.toml
```diff
@@ -32,22 +33,29 @@ color-eyre = "0.6.2"
 dirs = "5.0.1"
 eyre = "0.6.12"
 fragile = "2.0.0"
+futures-util = "0.3.30"
+jsonrpsee = "0.22.0"
```
miraclx: quick note here, as I know it can be a bit tricky.. their docs isn't the most helpful with figuring out how to plug it in as a tower services

but Sandi & I looked into this before, and this was the result which shows how the RPC handler can be made into a service that then handles individual requests

https://github.com/calimero-is-near/cali2.0-experimental/blob/caedbecd25ccf992980d799b7dc41e88bd5cc354/chat-p0c/src/network/mod.rs#L48-L65

crates/server/src/websocket.rs
fbozic: done, rewrote to axum and dropped warp

Cargo.toml
fbozic: it doesn't complain. removed.

crates/node/src/cli/init.rs
fbozic: done

crates/server/src/websocket.rs
```diff
@@ -1,105 +1,184 @@
...
+    ws: WebSocketUpgrade,
+    Extension(state): Extension<Arc<ServiceState>>,
+) -> impl IntoResponse {
+    ws.on_upgrade(move |socket| handle_socket(socket, state))
+}
+
+async fn handle_socket(socket: WebSocket, state: Arc<ServiceState>) {
+    let (commands_sender, commands_receiver) = mpsc::channel(32);
     let client_id = loop {
         let peer_id = rand::random();
```
miraclx: `client_id` != `peer_id`, can we unify these?

crates/server/src/websocket.rs
```diff
@@ -1,105 +1,184 @@
...
+    ws_state.connections.remove(&client_id);
+}
 
+fn handle_node_events(
+    client_id: server::WsClientId,
+    state: Arc<ServiceState>,
+    mut node_events_receiver: broadcast::Receiver<calimero_primitives::events::NodeEvent>,
+    command_sender: mpsc::Sender<WsCommand>,
+) {
     tokio::task::spawn(async move {
```
miraclx: can you move these `spawn` invocations to the caller?

crates/server/src/websocket.rs
```diff
@@ -1,105 +1,184 @@
...
+    handle_node_events(
+        client_id,
+        state.clone(),
+        state.node_events.subscribe(),
+        commands_sender.clone(),
+    );
+
+    let (socket_sender, mut socket_receiver) = socket.split();
+
+    handle_commands(client_id, commands_receiver, socket_sender);
```
miraclx: once `handle_commands` is async;

```rust
tokio::spawn(handle_commands(client_id, commands_receiver, socket_sender));
```

crates/server/src/websocket.rs
```diff
@@ -1,105 +1,184 @@
...
+
+    while let Some(message) = socket_receiver.next().await {
+        let message = match message {
+            Ok(message) => message,
+            Err(e) => {
+                error!(%e, "failed to read ws::Message(client_id={})", client_id);
```
miraclx: this perhaps?

```suggestion
                error!(%client_id, %e, "failed to read ws::Message");
```

crates/server/src/websocket.rs
```diff
@@ -1,105 +1,184 @@
...
+                command_sender
+                    .send(server::WsCommand::Send(response))
+                    .unwrap_or_else(|e| {
+                        error!(
+                            %e,
+                            "failed to send WsCommand::Event(client_id={})",
+                            client_id,
+                        );
+                    })
+                    .await;
```
miraclx: use the `unwrap_or*` pattern if you care about the `T` in the `Ok` variant.

```suggestion
                if let Err(err) = command_sender.send(server::WsCommand::Send(response)).await {
                    error!(
                        %e,
                        "failed to send WsCommand::Event(client_id={})",
                        client_id,
                    );
                }
```

crates/node/src/lib.rs
```diff
@@ -299,6 +307,15 @@ impl Node {
             } => {
                 if self.app_topic == topic_hash {
                     info!("{} joined the session.", their_peer_id.cyan());
+                    if self.node_events.receiver_count() > 0 {
+                        self.node_events.send(
```
miraclx: `send` only returns an error when there are no receivers. And so, [it does the same check logic as `receiver_count`](https://github.com/tokio-rs/tokio/blob/b2896feb5bf4c123c0c5b92f97f948ae11c7cdd2/tokio/src/sync/broadcast.rs#L590-L592). 

And it looks like this will never return `Err` since it's guaranteed to only attempt a send when there are receivers, meaning you don't care about the error, so this can be simplified to:

```suggestion
                    let _ = self.node_events.send(
```

crates/node/src/lib.rs
```diff
@@ -401,19 +418,30 @@ impl Node {
...
             None => return Ok(None),
         };
 
         let outcome = self
-            .execute(transaction.method, transaction.payload, true)
+            .execute(
+                transaction.method.clone(),
+                transaction.payload.clone(),
+                true,
+            )
```
```suggestion
            .execute(transaction.method, transaction.payload, true)	
```

crates/node/src/lib.rs
```diff
@@ -401,19 +418,30 @@ impl Node {
...
             .await?;
 
-        if let Some(tx) = outcome_rx {
-            let _ = tx.send(outcome);
+        if outcome.returns.is_ok() && self.node_events.receiver_count() > 0 {
+            let event = calimero_primitives::events::NodeEvent::TransactionExecuted(
+                calimero_primitives::events::ExecutedTransactionInfo { hash },
+            );
+            self.node_events.send(event)?;
+        }
```
miraclx: move into `execute` perhaps? ln:484?

we already have this `returns is ok` check there, and as mentioned, you can send without the receiver count check.

crates/node/src/lib.rs
```diff
@@ -513,7 +541,7 @@ impl TransactionPool {
             TransactionPoolEntry {
                 sender,
                 transaction,
-                outcome_rx,
+                outcome_sender: outcome_sender,
```
```suggestion
                outcome_sender,
```

crates/node/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub struct NodeConfig {
 pub struct TransactionPoolEntry {
     sender: calimero_network::types::PeerId,
     transaction: calimero_primitives::transaction::Transaction,
-    outcome_rx: Option<oneshot::Sender<calimero_runtime::logic::Outcome>>,
```
miraclx, line 31: lol, how'd that slip through, thanks for the fix

crates/node/src/lib.rs
fbozic: The entire node would crash unless I handled it this way. It returns an error if the receiver count is 0. Not sure what I am missing here. 
image

miraclx: Well, not exactly.. since it only returns that error when the receiver count is 0, and with the current logic you have, you're not doing anything if receiver_count is 0, so you can delegate the check to `send` and ignore the error.

the `let _ = ...` syntax ignores those errors. this won't cause the whole node to crash.

miraclx: We already have similar patterns:

https://github.com/calimero-is-near/cali2.0-experimental/blob/28c66ee31608d7c57a7e4816363943d1f31e8d21/crates/node/src/lib.rs#L94

https://github.com/calimero-is-near/cali2.0-experimental/blob/28c66ee31608d7c57a7e4816363943d1f31e8d21/crates/node/src/lib.rs#L416

fbozic: Aha, ignore the error. Sorry, I've missed that part. I always have a feeling ignoring errors is some anti-pattern. But I get that here it makes perfect sense.

Cargo.toml
fbozic: done

crates/server/src/websocket.rs
fbozic: done

crates/server/src/websocket.rs
fbozic: done

crates/node/src/lib.rs
fbozic: done, but I also had to send `Option<Hash>` to the execute method. because `call` method doesn't take hash I've sent `None` from it.

crates/node/src/lib.rs
fbozic: done

crates/server/src/websocket.rs
fbozic: done everywhere

crates/server/src/websocket.rs
fbozic: done everywhere. good point, thanks for the info.

crates/node/src/lib.rs
```diff
@@ -401,26 +416,27 @@ impl Node {
...
 
         Ok(Some(()))
     }
 
     async fn execute(
         &mut self,
+        hash: Option<calimero_primitives::hash::Hash>,
         method: String,
         payload: Vec<u8>,
         writes: bool,
```
miraclx: how about we merge `hash` and `writes` since you only have `Some(hash)` when `writes` is `true`

miraclx: but it's fine, we can do this in a secondary PR

crates/node/src/lib.rs
miraclx: looks good, added a secondary comment which is optional, so this is fine as-is

crates/node/src/lib.rs
fbozic: done

crates/node/src/lib.rs
```diff
@@ -439,12 +439,10 @@ impl Node {
         hash: Option<calimero_primitives::hash::Hash>,
         method: String,
         payload: Vec<u8>,
-        writes: bool,
     ) -> eyre::Result<calimero_runtime::logic::Outcome> {
-        let mut storage = if writes {
-            TemporalRuntimeStore::Write(calimero_store::TemporalStore::new(&self.store))
-        } else {
-            TemporalRuntimeStore::Read(calimero_store::ReadOnlyStore::new(&self.store))
+        let mut storage = match hash {
```
miraclx, line 443: 🚀

PR #25, opened 2024-02-28
feat: add in-memory GraphQL server

xilosada:
## Summary:

This PR integrates an GraphQL server with Actix-web and async-graphql for handling of posts and comments.
For simplicity it stores everything in-memory.
It also includes a schema for GraphQL clients.

## Test Plan:

```
cargo run                                         
    Finished dev [unoptimized + debuginfo] target(s) in 0.46s
     Running `/Users/xabi/dev/cali2.0-experimental/target/debug/peer-space`
GraphiQL IDE: http://localhost:8000
```

Using the graphQL playground the following operations are available:

```
query {
  posts {
    id
    title
    content
    comments {
      user
      text
    }
  }
}
```
```
query {
  post(id: 0) {
    id
    title
    content
    comments {
      user
      text
    }
  }
}
```

```
mutation {
  createPost(input:{ title: "New Post", content: "Content of the new post"}) {
    id
    title
    content
  }
}

```

```
mutation {
  createComment(input: { post: 0, user: "Xabi", text: "FFF"}) {
    id
    title
    content
    comments {
      user
      text
    }
  }
}
```

Screenshot 2024-02-25 at 6 35 18 PM

### DIFF
--- a/peer-space/schema.graphql
+++ b/peer-space/schema.graphql
@@ -4,6 +4,7 @@ schema {
 }
 
 type Query {
+    post(id: ID!): Post!,
     posts: [Post!]!
 }
 
@@ -19,19 +20,19 @@ input CreatePost {
 
 input CreateComment {
     post: ID!,
-    user: String,
+    user: String!,
     text: String!
 }
 
 type Comment {
-    text: String,
-    user: String,
+    text: String!,
+    user: String!,
 }
 
 type Post {
     id: ID!,
-    title: String,
-    content: String,
+    title: String!,
+    content: String!,
     comments: [Comment],
 }
 
--- a/peer-space/src/graphql.rs
+++ b/peer-space/src/graphql.rs
@@ -26,6 +26,14 @@ impl QueryRoot {
         let stored_posts = ctx.data_unchecked::<Storage>().lock().await.clone();
         Ok(stored_posts)
     }
+    async fn post<'a>(&self, ctx: &Context<'a>, id: i32) -> async_graphql::Result<Post> {
+        let idx: usize = id as usize;
+        if let Some(post) = ctx.data_unchecked::<Storage>().lock().await.clone().get(idx) {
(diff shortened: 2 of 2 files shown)


peer-space/schema.graphql
```diff
@@ -0,0 +1,37 @@
+schema {
+    query: Query,
+    mutation: Mutation
+}
+
+type Query {
+    posts: [Post!]!
```
frdomovic, line 8: Can also "post" be added ? to get specific by id?

peer-space/schema.graphql
```diff
@@ -0,0 +1,37 @@
...
+}
+
+input CreateComment {
+    post: ID!,
+    user: String,
+    text: String!
+}
+
+type Comment {
+    text: String,
```
MatejVukosav: comment should also have an id

peer-space/schema.graphql
```diff
@@ -0,0 +1,37 @@
...
+}
+
+input CreatePost {
+    title: String!,
+    content: String!,
+}
+
+input CreateComment {
+    post: ID!,
+    user: String,
```
MatejVukosav: User cannot be null => String!

peer-space/schema.graphql
```diff
@@ -0,0 +1,37 @@
...
+    text: String!
+}
+
+type Comment {
+    text: String,
+    user: String,
```
MatejVukosav: both text and user should be mandatory

peer-space/schema.graphql
xilosada: added

peer-space/schema.graphql
xilosada: we don't have any query or mutation that uses the ID. For posts we use it to add comments. Once we have a proper db and the ids can be consistant (right now is just the index in an array) we can add it if there is any use for it

PR #30, opened 2024-02-29
feat: add app wasm path to config

MatejVukosav:
Enable setup of the wasm path through the config file. 

config:

```
[identity]
PeerID = "12D3KooWRqX2NS3kHKY9HJQevEG1X65yKDubc5HJT5GVnLkJrowG"
PrivKey = "23jhTddKp3F7Pmowank6AAZzgjMKyXzKYghSnWJAqGVnLd8UWUNs9NDr6MoXVcpbfJhaYh6rhNhWSfcndpF2NBPvTK55L"

[swarm]
listen = [
    "/ip4/0.0.0.0/tcp/2331",
    "/ip4/0.0.0.0/udp/2331/quic-v1",
    "/ip6/::/tcp/2331",
    "/ip6/::/udp/2331/quic-v1",
]

[bootstrap]
nodes = []

[discovery]
mdns = true

[endpoint]
host = "127.0.0.1"
port = 3031

[app]
path = "apps/only-peers/res/only_peers.wasm"
```

> [!IMPORTANT]
> You need to delete config files and re-create them
>
> ```console
> $ rm -rf data/node{1,2,3}
> $ rm -rf data/coordinator
> ```

### DIFF
--- a/crates/node/src/cli/init.rs
+++ b/crates/node/src/cli/init.rs
@@ -121,8 +121,7 @@ impl InitCommand {
             boot_nodes.extend(BootstrapNodes::ipfs().list);
         }
 
-        let mut app_path = Utf8PathBuf::new();
-        app_path.push("apps/only-peers/res/only_peers.wasm");
+        let app_path = Utf8PathBuf::from("apps/only-peers/res/only_peers.wasm");
 
         let config = ConfigFile {
             identity,


crates/network/src/config.rs
```diff
@@ -98,6 +98,11 @@ impl Default for EndpointConfig {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AppConfig {
+    pub wasm_path: String,
+}
```
miraclx: Move to `calimero_node::config` instead

crates/network/src/config.rs
```diff
@@ -98,6 +98,11 @@ impl Default for EndpointConfig {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AppConfig {
+    pub wasm_path: String,
```
miraclx: Also, maybe this looks better?

```toml
[app]
path = "only_peers.wasm"
```

crates/node/src/cli/init.rs
```diff
@@ -120,11 +121,15 @@ impl InitCommand {
             boot_nodes.extend(BootstrapNodes::ipfs().list);
         }
 
+        let mut app_path = Utf8PathBuf::new();
+        app_path.push("apps/only-peers/res/only_peers.wasm");
```
miraclx: this works, but it could be compressed to

```suggestion
        let app_path = Utf8PathBuf::from("apps/only-peers/res/only_peers.wasm");
```

miraclx: Also, this assumes we're in the root directory, which is not a good assumption

Suppose we build the node with:

```console
$ cargo build -p calimero-node --release
```

We should be able to distribute the binary in `target/release/calimero-node` and have it work there as well without requiring that the `apps/only-peers/res/` folder exists in whatever curdir it's executed in

If you want to use relative paths, it should be relative to the one absolute path we already have (the node's home dir)

One option is to create a folder inside the node's home dir that the path would be relative to. ex:`nodeN/apps`

```console
$ ls ~/.calimero/node1/apps
only_peers.wasm
kv_store.wasm
```

and then select the binary here

```toml
[app]
path = "only_peers.wasm"
```

The downside is everytime you make a change a and rebuild the app, you have to manually copy it to all the nodes

You could also use a symlink, but it's still not as direct as the current approach that specifies a relative path at runtime

MatejVukosav: This is the default value valid for development. Of course, the user can modify that value at any time and adapt it per file location

MatejVukosav: I recommend not focusing too much on this now as we have a fixed location during development and later when we define app download location, we can redesign this.

PR #33, opened 2024-03-01
fix: wait for listeners to become ready before bootstraping the network

fbozic:


crates/network/src/lib.rs
```diff
@@ -33,14 +34,36 @@ pub async fn run(
...
 
     for addr in &config.swarm.listen {
         client.listen_on(addr.clone()).await?;
     }
 
+    // Reference: https://github.com/libp2p/rust-libp2p/blob/60fd566a955a33c42a6ab6eefc1f0fedef9f8b83/examples/dcutr/src/main.rs#L118
+    loop {
+        tokio::select! {
+            Some(event) = event_receiver.recv() => {
+                match event {
```
miraclx, line 49: I don't particularly like this approach because it will never print `Listening on <address>` if it takes longer than a second to return.

Which it shouldn't, but still.. racy cond

miraclx: answered privately: the node will log a warning of an unexpected listen after bootstrapping

crates/node/src/lib.rs
```diff
@@ -274,6 +274,15 @@ impl Node {
                     message => error!("Unhandled PeerAction: {:?}", message),
                 }
             }
+            calimero_network::types::NetworkEvent::ListeningOn {
+                listener_id,
+                address,
+            } => {
+                warn!(
+                    "listening on not really expected here(listener_id={}): {}",
```
miraclx: do we need listener id here?

crates/network/src/lib.rs
```diff
@@ -33,14 +34,36 @@ pub async fn run(
...
+                    types::NetworkEvent::ListeningOn {
+                        listener_id,
+                        address,
+                    } => {
+                        info!("Listening on(listener_id={}): {}", listener_id, address)
```
miraclx: here as well

crates/node/src/lib.rs
fbozic: Not sure, maybe only in debug mode?

miraclx: looks like it just counts upwards from 0, indexes in a vec, don't think we need to show this

fbozic: done

crates/network/src/lib.rs
fbozic: done

PR #34, opened 2024-03-04
feat: integrate graphql server with node

miraclx:
Changes

- A new `server` section is now required as part of the config, you can add this to existing config to get set up:

    ```toml
	[server]
	listen = ["/ip4/0.0.0.0/tcp/2528", "/ip6/::/tcp/2528"]
	
	[server.graphql]
	enabled = true
    ```

Testing

First, compile the `only-peers` application

```console
$ ./apps/only-peers/build.sh
   Compiling only-peers v0.1.0 (/git/calimero-is-near/cali2.0-experimental/apps/only-peers)
    Finished app-release [optimized] target(s) in 1.40s
```

> [!IMPORTANT]
>
> You need to dump the previous storage state (incase it's been previously initialized for a different app)
>
> ```console
> $ rm -rf node{1,2,3}/data
> ```

Use different ports for each node in their respective config, and then spin up your nodes as before

#### Create Post

```console
$ xh POST http://localhost:2528/graphql query='mutation { createPost(input: {title: "Something Happened", content: "This is a post about something that happened"}) { title, content } }'
{
    "data": {
        "createPost": {
            "title": "Something Happened",
            "content": "This is a post about something that happened"
        }
    }
}
```

```console
$ xh POST http://localhost:2528/graphql query='{ post(id:0) {comments { text } title content} }'
{
    "data": {
        "post": {
            "comments": [],
            "title": "Something Happened",
            "content": "This is a post about something that happened"
        }
    }
}
```

#### todo
- [X] GraphiQL

### DIFF
--- a/crates/node/src/cli/init.rs
+++ b/crates/node/src/cli/init.rs
@@ -40,7 +40,7 @@ pub struct InitCommand {
 
     /// Host to listen on for RPC
     #[clap(long, value_name = "HOST")]
-    #[clap(default_value = "127.0.0.1,::")]
+    #[clap(default_value = "127.0.0.1,::1")]
     pub server_host: Vec<IpAddr>,
 
     /// Port to listen on for RPC


apps/only-peers/src/lib.rs
```diff
@@ -0,0 +1,72 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use calimero_sdk::env;
+use serde::Serialize;
+
+mod code_generated_from_calimero_sdk_macros;
+
+#[derive(Default, BorshSerialize, BorshDeserialize)]
+struct OnlyPeers {
```
MatejVukosav, line 8: What is idea behind this struct? Why so generic name?

apps/only-peers/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    pub fn create_post(&mut self, title: String, content: String) -> &Post {
+        env::log(&format!(
+            "Creating post with title: {:?} and content: {:?}",
+            title, content
+        ));
+
+        self.posts.push(Post {
+            id: self.posts.len(),
```
MatejVukosav, line 46: This looks dangerous. Ok for the demo, but should be converted to more robust generation like uuids at some point

apps/only-peers/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+            content,
+            comments: Vec::new(),
+        });
+
+        self.posts.last().unwrap()
+    }
+
+    pub fn create_comment(
+        &mut self,
+        post: usize,
```
MatejVukosav: Shouldn't this be just postId?

apps/only-peers/src/lib.rs
miraclx: 🤷🏽‍♂️ I literally have no idea, followed this - https://github.com/calimero-is-near/only-peers-client

apps/only-peers/src/lib.rs
miraclx: 😆 it was, and then last minute I changed it to match the previous impl

https://github.com/calimero-is-near/cali2.0-experimental/blob/df84a93aba3d2965fe25bc9e0b2c34d26f2c860d/peer-space/src/graphql.rs#L40

I'll change it back

crates/network/src/config.rs
```diff
@@ -69,35 +65,16 @@ impl BootstrapNodes {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct DiscoveryConfig {
-    #[serde(default = "bool_true")]
+    #[serde(default = "calimero_primitives::common::bool_true")]
```
MatejVukosav, line 68: How is the decision made when to use stuff from import and when to write the full path?

crates/server/src/config.rs
```diff
@@ -0,0 +1,17 @@
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+
+use multiaddr::Multiaddr;
+use serde::{Deserialize, Serialize};
+
+pub const DEFAULT_PORT: u16 = 2528; // (CHAT in T9) + 100
```
MatejVukosav, line 6: what is `CHAT in T9` ?

apps/only-peers/src/lib.rs
MatejVukosav: @frdomovic ^

apps/only-peers/src/lib.rs
miraclx: Agree, should be `sha256(AuthorId, ChannelId, CreationTimestamp)` or something of the sort.

But first, we'll need to expose the caller's ID and provide timestamp information to the runtime context.

crates/server/src/config.rs
miraclx: CHAT on a phone keyboard is 2428, we're using that for the swarm listen address by default.

crates/network/src/config.rs
miraclx: - if there can be any ambiguity, use the full path to qualify it.
- if it's used a lot, import it, else - full path
- avoid wildcards in all cases, but it's okay when you're prototyping. by the time you want to want to commit, be explicit

crates/server/examples/demo.rs
```diff
@@ -0,0 +1,100 @@
...
+use tracing::{error, info};
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::EnvFilter;
+
+#[tokio::main]
+async fn main() -> eyre::Result<()> {
+    setup()?;
+
+    let config = calimero_server::config::ServerConfig {
+        listen: vec!["/ip4/127.0.0.1/tcp/2528".parse()?],
```
MatejVukosav: Can u use `DEFAULT_PORT` const?

crates/node/src/lib.rs
```diff
@@ -55,6 +56,10 @@ pub async fn start(config: NodeConfig) -> eyre::Result<()> {
 
     let mut node = Node::new(&config, network_client).await?;
 
+    let (client_tx, mut client_rx) = mpsc::channel(32);
```
fbozic:
```suggestion
    let (gql_command_sender, mut gql_command_reciever) = mpsc::channel(32);
```

fbozic: We are using `sender` and `receiver` for the channels for most of the code base. Maybe keep the same approach here as well?

crates/node/src/lib.rs
```diff
@@ -55,6 +56,10 @@ pub async fn start(config: NodeConfig) -> eyre::Result<()> {
...
+
+    let mut server = Box::pin(calimero_server::start(config.server, client_tx));
```
fbozic: Random question: Why do we need a pin? I'm having a hard time understanding what exactly is pin doing.

miraclx: `calimero_server::start` returns a future, which may be polled repeatedly by the `tokio::select!` macro

Futures are state machines, which is syntactic sugar for a structure that hosts all it's dependencies (akin to closures).

`calimero_node::start` is a future that hosts the `calimero_server::start` future which is polled by reference.

So this future (`calimero_node::start`) is self-referential, because it references a value defined within itself.

We need to guarantee that the referenced memory location won't be moved at runtime giving us undefined behaviour.

This is a good read - https://rust-lang.github.io/async-book/04_pinning/01_chapter.html

apps/only-peers/src/lib.rs
MatejVukosav: I don't think we need this. Client can combine queries as he wish

miraclx: The `OnlyPeers` struct is the state storage.

PR #37, opened 2024-03-04
feat(identity): add auth to gql server

MatejVukosav:
Add authentication using a signature. The client needs to sign msg with a peer private key and add it to the header.

To obtain example headers run `cargo test `in `crates/identity`.

**Each node has its own keypair so headers are different for each peer.** 

Screenshot 2024-03-08 at 01 11 39

Screenshot 2024-03-08 at 01 11 33

### DIFF
--- a/crates/identity/src/auth.rs
+++ b/crates/identity/src/auth.rs
@@ -9,11 +9,10 @@ mod tests {
     use super::*;
 
     #[test]
-    // Use for obtaining test samples
     fn valid_headers() {
         let keypair = get_peer_keypair().unwrap();
         let msg = "blabla";
-        println!("content header= {:?}", bs58::encode(msg).into_string());
+        println!("challenge header= {:?}", bs58::encode(msg).into_string());
 
         let signature = keypair.sign(msg.as_bytes()).unwrap();
         let signature_header = bs58::encode(&signature).into_string();
--- a/crates/node/src/config.rs
+++ b/crates/node/src/config.rs
@@ -2,7 +2,7 @@ use std::fs;
 
 use camino::Utf8PathBuf;
 use eyre::WrapErr;
-use libp2p::identity;
+use libp2p::{identity, Multiaddr};
 use serde::{Deserialize, Serialize};
 
 const CONFIG_FILE: &str = "config.toml";
@@ -29,13 +29,24 @@ pub struct ConfigFile {
 pub struct NetworkConfig {
     pub swarm: calimero_network::config::SwarmConfig,
 
-    pub server: calimero_server::config::ServerConfig,
+    pub server: ServerConfig,
 
     #[serde(default)]
     pub bootstrap: calimero_network::config::BootstrapConfig,
 
     #[serde(default)]
     pub discovery: calimero_network::config::DiscoveryConfig,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ServerConfig {
+    pub listen: Vec<Multiaddr>,
+
+    #[serde(default)]
+    pub graphql: Option<calimero_server::graphql::GraphQLConfig>,
+
+    #[serde(default)]
+    pub websocket: Option<calimero_server::websocket::WsConfig>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
--- a/crates/server/src/config.rs
+++ b/crates/server/src/config.rs
@@ -1,7 +1,6 @@
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 use multiaddr::Multiaddr;
-use serde::{Deserialize, Serialize};
 
 pub const DEFAULT_PORT: u16 = 2528; // (CHAT in T9) + 100
 pub const DEFAULT_ADDRS: [IpAddr; 2] = [
@@ -9,21 +8,15 @@ pub const DEFAULT_ADDRS: [IpAddr; 2] = [
     IpAddr::V6(Ipv6Addr::LOCALHOST),
 ];
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug)]
 pub struct ServerConfig {
     pub listen: Vec<Multiaddr>,
 
-    #[serde(
-        with = "calimero_identity::config::serde_identity",
-        default = "libp2p::identity::Keypair::generate_ed25519"
-    )]
     pub identity: libp2p::identity::Keypair,
 
-    #[serde(default)]
     #[cfg(feature = "graphql")]
     pub graphql: Option<crate::graphql::GraphQLConfig>,
 
-    #[serde(default)]
     #[cfg(feature = "websocket")]
     pub websocket: Option<crate::websocket::WsConfig>,
 }
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -1,6 +1,7 @@
 use std::net::{IpAddr, SocketAddr};
 
 use axum::routing::Router;
+use config::ServerConfig;
 use tokio::sync::{broadcast, mpsc, oneshot};
 use tracing::warn;
 
@@ -20,7 +21,7 @@ type ServerSender = mpsc::Sender<(
 )>;
 
 pub async fn start(
-    config: config::ServerConfig,
+    config: ServerConfig,
     server_sender: ServerSender,
     node_events: broadcast::Sender<calimero_primitives::events::NodeEvent>,
 ) -> eyre::Result<()> {
--- a/crates/server/src/middleware/auth.rs
+++ b/crates/server/src/middleware/auth.rs
@@ -69,6 +69,7 @@ where
     }
 }
 
+#[derive(Debug)]
 struct AuthHeaders {
     signature: Vec<u8>,
     challenge: Vec<u8>,
--- a/crates/node/src/cli/init.rs
+++ b/crates/node/src/cli/init.rs
@@ -3,7 +3,6 @@ use std::net::IpAddr;
 
 use calimero_network::config::{BootstrapConfig, BootstrapNodes, DiscoveryConfig, SwarmConfig};
 use calimero_node::config::{self, AppConfig, ConfigFile, NetworkConfig, StoreConfig};
-use calimero_server::config::ServerConfig;
 use camino::Utf8PathBuf;
 use clap::{Parser, ValueEnum};
 use eyre::WrapErr;
@@ -135,7 +134,7 @@ impl InitCommand {
                     nodes: BootstrapNodes { list: boot_nodes },
                 },
                 discovery: DiscoveryConfig { mdns },
-                server: ServerConfig {
+                server: calimero_node::config::ServerConfig {
                     listen: self
                         .server_host
                         .into_iter()
@@ -144,7 +143,6 @@ impl InitCommand {
                         })
                         .collect(),
                     graphql: Some(calimero_server::graphql::GraphQLConfig { enabled: true }),
-                    identity,
                     websocket: Some(calimero_server::websocket::WsConfig { enabled: true }),
                 },
             },
--- a/crates/node/src/cli/run.rs
+++ b/crates/node/src/cli/run.rs
@@ -49,7 +49,12 @@ impl RunCommand {
                 bootstrap: config.network.bootstrap,
                 discovery: config.network.discovery,
             },
-            server: config.network.server,
+            server: calimero_server::config::ServerConfig {
+                listen: config.network.server.listen,
+                identity: config.identity,
+                graphql: config.network.server.graphql,
+                websocket: config.network.server.websocket,
+            },
         })
         .await
     }


crates/identity/src/auth.rs
```diff
@@ -0,0 +1,24 @@
+use libp2p::identity::Keypair;
+
+pub fn get_peer_keypair() -> Result<Keypair, String> {
+    let private_key = "...";
```
MatejVukosav: @miraclx I need a private key from the config file here. What is the best way to obtain it?

miraclx: Pass it from the node, use it to initialize the middleware, store it as part of the middleware state

https://github.com/calimero-is-near/cali2.0-experimental/blob/df84a93aba3d2965fe25bc9e0b2c34d26f2c860d/crates/node/src/lib.rs#L18

crates/server/examples/demo.rs
```diff
@@ -50,7 +51,10 @@ async fn main() -> eyre::Result<()> {
 
     let (server_sender, mut server_receiver) = mpsc::channel(32);
 
-    let mut server = Box::pin(calimero_server::start(config, server_sender));
+    let keypair = Keypair::generate_ed25519();
+    let pk = &bs58::encode(&keypair.to_protobuf_encoding().unwrap()).into_string();
+    println!("Private key {:?}", pk);
```
MatejVukosav, line 62: Print so headers can be generated from demo pk

crates/server/src/lib.rs
```diff
@@ -14,11 +18,16 @@ type Sender = mpsc::Sender<(
     oneshot::Sender<calimero_runtime::logic::Outcome>,
 )>;
 
-pub async fn start(config: config::ServerConfig, sender: Sender) -> eyre::Result<()> {
+pub async fn start(
+    config: config::ServerConfig,
+    sender: Sender,
+    keypair: Keypair,
```
miraclx: Add `identity` to `ServerConfig` instead.

crates/server/examples/demo.rs
miraclx:
```suggestion
    let pk = bs58::encode(&keypair.to_protobuf_encoding()?).into_string();
    println!("Private key {:?}", pk);
```

crates/server/Cargo.toml
```diff
@@ -11,15 +11,21 @@ axum.workspace = true
...
+tower = "0.4.13"
 tower-http = { workspace = true, optional = true }
 tracing.workspace = true
 
 calimero-primitives = { path = "../primitives" }
 calimero-runtime = { path = "../runtime" }
+calimero-identity = { path = "../identity" }
+tower-service = "0.3.2"
+tower-layer = "0.3.2"
+bs58.workspace = true
```
miraclx: nit: order

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+    S::Future: Send + 'static,
+{
+    type Response = S::Response;
+    type Error = Infallible;
+    // `BoxFuture` is a type alias for `Pin<Box<dyn Future + Send + 'a>>`
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // self.inner.poll_ready(cx)
+        self.inner.poll_ready(cx).map_err(|e| e.into())
```
miraclx: doesn't it work without the map?

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
+        let result = auth(request.headers(), &self.keypair);
+
+        if result.is_err() {
+            let error_response: Response<Body> = result.err().unwrap().into_response();
```
miraclx:
```suggestion
        if let Err(err) = result {
            let error_response = err.into_response();
```

crates/server/src/auth_signature_middleware.rs
miraclx, line 1: can you make it into a structure like this instead: `calimero::server::middleware::auth`?

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+    if signature.is_err() {
+        return Err(UnauthorizedError::new("Cannot unwrap signature"));
+    }
+    let signature = bs58::decode(signature.unwrap()).into_vec();
+    if signature.is_err() {
+        return Err(UnauthorizedError::new("Invalid base58"));
+    }
+    let signature = signature.unwrap();
+
+    let content = headers.get("content");
```
miraclx: challenge maybe?

crates/identity/src/auth.rs
```diff
@@ -0,0 +1,39 @@
+use libp2p::identity::Keypair;
+
+pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> Result<bool, String> {
```
miraclx: can this potentially fail? shouldn't it all be encapsulated in a `true` / `false` regardless?

```suggestion
pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> Result<bool, String> {
```

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+fn get_auth_headers(headers: &HeaderMap) -> Result<AuthHeaders, UnauthorizedError> {
+    let signature = headers.get("signature");
+    if signature.is_none() {
+        return Err(UnauthorizedError::new("Missing signature header"));
+    }
+    let signature = signature.unwrap().to_str();
+    if signature.is_err() {
+        return Err(UnauthorizedError::new("Cannot unwrap signature"));
+    }
+    let signature = bs58::decode(signature.unwrap()).into_vec();
```
miraclx: the signature of `bs58::decode` is `fn(T)` where `T: AsRef<[u8]>`, so we don't need to first convert the `HeaderValue` to a string.

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+    let content = content.unwrap().to_str();
+    if content.is_err() {
+        return Err(UnauthorizedError::new("Cannot unwrap content"));
+    }
+    let content = bs58::decode(content.unwrap()).into_vec();
+    if content.is_err() {
+        return Err(UnauthorizedError::new("Invalid base58"));
+    }
+    let content = content.unwrap();
```
miraclx: please abstain from panicking

```suggestion
    let signature = headers
        .get("signature")
        .ok_or_else(|| UnauthorizedError::new("Missing signature header"))?;

    let signature = bs58::decode(signature)
        .into_vec()
        .map_err(|_| UnauthorizedError::new("Invalid base58"))?;

    let content = headers
        .get("content")
        .ok_or_else(|| UnauthorizedError::new("Missing content header"))?;

    let content = bs58::decode(content)
        .into_vec()
        .map_err(|_| UnauthorizedError::new("Invalid base58"))?;
```

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl<'a> UnauthorizedError<'a> {
+    pub fn new(reason: &'a str) -> Self {
+        Self { reason }
+    }
+}
+
+impl std::fmt::Display for UnauthorizedError<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.reason)
```
miraclx:
```suggestion
        f.pad(self.reason)
```

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+impl std::error::Error for UnauthorizedError<'_> {}
+
+impl IntoResponse for UnauthorizedError<'_> {
+    fn into_response(self) -> Response<Body> {
+        let response = Response::builder()
+            .status(StatusCode::UNAUTHORIZED)
+            .body(Body::from(self.reason.to_string()))
+            .unwrap();
```
miraclx: ⚠️

miraclx: [`(StatusCode, R)` implements `IntoResponse` where `R: IntoResponse`](https://docs.rs/axum/latest/axum/response/trait.IntoResponse.html#impl-IntoResponse-for-(StatusCode,+R))

Meaning, you can simplify this to:

```suggestion
        (StatusCode::UNAUTHORIZED, self.reason).into_response()
```

crates/identity/Cargo.toml
```diff
@@ -14,3 +14,4 @@ serde_json.workspace = true
 sha2.workspace = true
 multibase.workspace = true
 serde.workspace = true
+bs58.workspace = true
```
miraclx: nit: please order this, keeping with the convention

crates/server/examples/demo.rs
```diff
@@ -1,6 +1,7 @@
 use std::env;
 use std::net::{Ipv4Addr, SocketAddr};
 
+use libp2p::identity::Keypair;
```
miraclx: let's keep with convention

```suggestion
use libp2p::identity;
```

crates/server/src/auth_signature_middleware.rs
```diff
@@ -0,0 +1,166 @@
...
+    task::{Context, Poll},
+};
+
+use axum::{
+    body::Body,
+    extract::Request,
+    http::{HeaderMap, StatusCode},
+    response::{IntoResponse, Response},
+};
+
```
miraclx: merge these sections of third-party imports

crates/server/src/lib.rs
```diff
@@ -57,7 +66,9 @@ pub async fn start(config: config::ServerConfig, sender: Sender) -> eyre::Result
     #[cfg(feature = "graphql")]
     {
         if let Some((path, handler)) = graphql::service(&config, sender.clone())? {
-            app = app.route(path, handler);
+            app = app
+                .route(path, handler)
+                .layer(AuthSignatureLayer::new(keypair));
             serviced = true;
         }
     }
```
miraclx: Let's move this out. It's not just the graphql service that needs this authentication.

```suggestion
            app = app.route(path, handler);
            serviced = true;
        }
    }
    app = app.layer(AuthSignatureLayer::new(keypair));
```

crates/server/src/config.rs
```diff
@@ -13,6 +13,12 @@ pub const DEFAULT_ADDRS: [IpAddr; 2] = [
 pub struct ServerConfig {
     pub listen: Vec<Multiaddr>,
 
+    #[serde(
+        with = "calimero_identity::config::serde_identity",
+        default = "libp2p::identity::Keypair::generate_ed25519"
+    )]
+    pub identity: libp2p::identity::Keypair,
```
miraclx: let's duplicate the `ServerConfig` types, to avoid having duplicate entries of the identity in the config file

One for it's representation in the config file, and another for the runtime

You can already see a similar pattern for `NetworkConfig`

crates/node/src/config.rs
```diff
@@ -12,7 +12,7 @@ pub const DEFAULT_CALIMERO_CHAT_HOME: &str = ".calimero/experiments/chat-p0c";
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ConfigFile {
     #[serde(
-        with = "serde_identity",
+        with = "calimero_identity::config::serde_identity",
```
miraclx, line 15: +1

crates/server/examples/demo.rs
```diff
@@ -1,7 +1,6 @@
 use std::env;
 use std::net::{Ipv4Addr, SocketAddr};
 
-use libp2p::identity::Keypair;
```
miraclx: just pull in `libp2p::identity` and then use `identity::Keypair`

crates/server/src/middleware/auth.rs
```diff
@@ -0,0 +1,152 @@
+pub mod auth {
```
miraclx: this will be `middleware::auth::auth`, just remove this extra one

crates/server/src/middleware/auth.rs
```diff
@@ -0,0 +1,152 @@
...
+    use std::{
+        convert::Infallible,
+        task::{Context, Poll},
+    };
```
miraclx: please run `cargo +nightly fmt`

crates/server/src/middleware/auth.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn token_is_valid(keypair: &Keypair, auth_headers: &AuthHeaders) -> bool {
+        let verify_result = verify_peer_auth(
+            keypair,
+            auth_headers.challenge.as_slice(),
+            auth_headers.signature.as_slice(),
+        );
+        if verify_result.is_err() {
+            return false;
+        }
+        verify_result.unwrap()
```
miraclx: `bool` defaults to `false`, so this can be written as

```suggestion
        verify_result.unwrap_or_default()
```

crates/server/src/middleware/auth.rs
```diff
@@ -0,0 +1,152 @@
...
+        S: Service<Request<Body>, Response = Response<Body>, Error = Infallible> + Send + 'static,
+        S::Future: Send + 'static,
+    {
+        type Response = S::Response;
+        type Error = Infallible;
+        // `BoxFuture` is a type alias for `Pin<Box<dyn Future + Send + 'a>>`
+        type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+        fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+            // self.inner.poll_ready(cx)
```
miraclx: ;-)

crates/server/examples/demo.rs
MatejVukosav: Why do we have in some places `calimero_server::graphql::GraphQLConfig` instead of  `graphql::GraphQLConfig`?  

It is not clear to me which pattern exactly we should follow as I can see both of them in the code.

crates/server/src/middleware/auth.rs
MatejVukosav: This was updated in the followed commit.

crates/server/src/config.rs
MatejVukosav: `You can already see a similar pattern for NetworkConfig` can u point exactly where? Not sure I am looking for the right thing

crates/server/examples/demo.rs
miraclx: Y'know what? leave it as-is, I'll make a wrapper in `calimero_primitives` that derefs to this with transparent serialization.

So we won't need to depend on all of libp2p just to use the `Keypair`

Leave it as is for now.

And to answer the question, we've already established a convention everywhere else of importing libp2p modules because we consistently use multiple parts of each module.

For `Keypair`, I'm not in favor of importing it globally because it's likely we'll revise this at which point we'll have a type fork, which will take some effort to resolve.

And the `graphql::*` pattern may introduce some reading ambiguity, which graphql? scrolls to the top of the file, Oh, it's not a dependency, it's a module instead.

But as mentioned, leave as-is for now, we need to clean that up from before.

crates/identity/src/auth.rs
```diff
@@ -0,0 +1,39 @@
...
+pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> bool {
+    keypair.public().verify(msg, signature)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    // Use for obtaining test samples
```
miraclx: is this still the case?

crates/server/src/config.rs
miraclx: there's one in the node's config, and in network's config. the one in the node config matches the expectations of the config file, and the one in the network's config has identity, which isn't serialized. it's only used at runtime.

crates/server/Cargo.toml
```diff
@@ -10,18 +10,24 @@ license.workspace = true
...
 tokio.workspace = true
+tower = "0.4.13"
 tower-http = { workspace = true, optional = true }
 tracing.workspace = true
 
 calimero-primitives = { path = "../primitives" }
 calimero-runtime = { path = "../runtime" }
+calimero-identity = { path = "../identity" }
+tower-service = "0.3.2"
+tower-layer = "0.3.2"
```
miraclx, line 30: order & versions should be organized at the root

crates/server/Cargo.toml
```diff
@@ -10,18 +10,24 @@ license.workspace = true
...
+bs58.workspace = true
 eyre.workspace = true
+libp2p.workspace = true
 futures-util = { workspace = true, optional = true }
 multiaddr.workspace = true
 rand.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 tokio.workspace = true
+tower = "0.4.13"
```
miraclx, line 22: same here

PR #38, opened 2024-03-05
fix: missing server-host

MatejVukosav:
Fix `error: invalid value '127.0.0.1,::' for '--server-host <HOST>': invalid IP address syntax` after running `./crates/node/gen_localnet_configs.sh 3`

### DIFF
--- a/crates/node/src/cli/init.rs
+++ b/crates/node/src/cli/init.rs
@@ -40,7 +40,7 @@ pub struct InitCommand {
 
     /// Host to listen on for RPC
     #[clap(long, value_name = "HOST")]
-    #[clap(default_value = "127.0.0.1,::1")]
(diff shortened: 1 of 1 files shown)


crates/node/src/cli/init.rs
```diff
@@ -40,7 +40,7 @@ pub struct InitCommand {
 
     /// Host to listen on for RPC
     #[clap(long, value_name = "HOST")]
-    #[clap(default_value = "127.0.0.1,::1")]
+    #[clap(default_value = "127.0.0.1")]
```
miraclx, line 43: ipv6 should be bound to by default

Here's the proper fix: https://github.com/calimero-is-near/cali2.0-experimental/pull/39

PR #43, opened 2024-03-18
fix(server): cors

MatejVukosav:
Fix cors issues while calling server from frontend.

First few calls without fix, and last one with the fix. 

Screenshot 2024-03-18 at 21 02 16


crates/server/src/lib.rs
```diff
@@ -1,9 +1,11 @@
 use std::net::{IpAddr, SocketAddr};
 
-use axum::Router;
+use axum::{http, Router};
 use tokio::sync::{mpsc, oneshot};
 use tracing::warn;
 
+use tower_http::cors;
```
miraclx, line 7: Please merge these sections

PR #44, opened 2024-03-18
feat: verify eth signature

MatejVukosav:
Verify eth signature functions.

### DIFF
--- a/crates/identity/src/auth.rs
+++ b/crates/identity/src/auth.rs
@@ -5,7 +5,7 @@ pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> bool
     keypair.public().verify(msg, signature)
 }
 
-pub fn verify_eth_signature(account: String, message: String, signature: String) -> bool {
+pub fn verify_eth_signature(account: &str, message: &str, signature: &str) -> bool {
     let signature_bytes = match hex::decode(signature.trim_start_matches("0x")) {
         Ok(bytes) => bytes,
         Err(_) => return false,
@@ -31,7 +31,7 @@ pub fn verify_eth_signature(account: String, message: String, signature: String)
     }
 }
 
-pub fn eth_message(message: String) -> [u8; 32] {
+pub fn eth_message(message: &str) -> [u8; 32] {
     keccak256(
         format!(
             "{}{}{}",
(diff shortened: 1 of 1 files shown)


crates/identity/src/auth.rs
```diff
@@ -1,13 +1,68 @@
 use libp2p::identity::Keypair;
+use web3::signing::{keccak256, recover};
 
 pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> bool {
     keypair.public().verify(msg, signature)
 }
 
+pub fn verify_eth_message(account: String, message: String, signature: String) -> bool {
```
xilosada: technically you  are verifying the signature

crates/identity/src/auth.rs
```diff
@@ -1,13 +1,68 @@
...
+            // Format the recovered public key as a hex string
+            let pubkey_hex = format!("{:02X}", pubkey);
+            // Compare the recovered public key with the account address in a case-insensitive manner
+            account.eq_ignore_ascii_case(&pubkey_hex)
+        }
+        Err(_) => false,
+    }
+}
+
+pub fn eth_message(message: String) -> [u8; 32] {
```
xilosada: str instead of String

crates/identity/src/auth.rs
```diff
@@ -1,13 +1,68 @@
 use libp2p::identity::Keypair;
+use web3::signing::{keccak256, recover};
 
 pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> bool {
     keypair.public().verify(msg, signature)
 }
 
+pub fn verify_eth_signature(account: String, message: String, signature: String) -> bool {
```
xilosada: str's instead of Strings

Cargo.toml
```diff
@@ -50,6 +50,7 @@ owo-colors = "3.5.0"
 rand = "0.8.5"
 rocksdb = "0.22.0"
 sha2 = "0.10.8"
+semver = "1.0.22"
 serde = "1.0.196"
 serde_json = "1.0.113"
```
miraclx, line 58: let's move `sha2` past these, danke

```suggestion
semver = "1.0.22"
serde = "1.0.196"
serde_json = "1.0.113"
sha2 = "0.10.8"
```

crates/identity/src/auth.rs
```diff
@@ -1,13 +1,68 @@
 use libp2p::identity::Keypair;
+use web3::signing::{keccak256, recover};
```
miraclx, line 2: I'd be cautious of `web3` due to all the deps it pulls in.

I'd be happy to help here. But I think we can add custom impls of these methods. They're also just delegating to other dependencies anyway.

`keccak256` - https://github.com/tomusdrw/rust-web3/blob/eccb0aa0ba846f6831a6a0149e2e318073f60f2c/src/signing.rs#L186-L193
`recover` - https://github.com/tomusdrw/rust-web3/blob/eccb0aa0ba846f6831a6a0149e2e318073f60f2c/src/signing.rs#L139-L148

Can be done in a separate PR, so as not to block this.

MatejVukosav: I have just merged this 2 minutes before your comment. xd 
Anyhow we can come back to this later again.

miraclx: 😅 yeah, that's okay

PR #45, opened 2024-03-19
fix: config borrow

MatejVukosav:
Fix for 

```
 if let Some((path, handler)) = websocket::service(&config, node_events.clone())? {
   |                                                           ^^^^^^^ value borrowed here after partial move
```

### DIFF
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -85,6 +85,12 @@ pub async fn start(
         }
     }
 
+    if !serviced {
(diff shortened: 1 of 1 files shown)


crates/server/src/lib.rs
```diff
@@ -94,6 +85,15 @@ pub async fn start(
         }
     }
 
+    app = app
```
miraclx, line 94: can you move it after the `serviced` check?

MatejVukosav: sure. done

PR #47, opened 2024-03-22
feat: implement JsonRpc server

fbozic:
- Implement "simple" JsonRpc server using plain axum
- Integrate server with the rest of the node

### DIFF
--- a/crates/server/src/jsonrpc.rs
+++ b/crates/server/src/jsonrpc.rs
@@ -149,3 +149,27 @@ pub(crate) async fn call(
         None => Ok(None),
     }
 }
+
+macro_rules! _mount_method {
+    ($request:ident -> Result<$response:ident, $error:ident>, $handle:path) => {
+        impl crate::jsonrpc::Request for $request {
+            type Response = $response;
+            type Error = $error;
+
+            async fn handle(
+                self,
+                state: std::sync::Arc<crate::jsonrpc::ServiceState>,
+            ) -> std::result::Result<Self::Response, crate::jsonrpc::RpcError<Self::Error>> {
+                match $handle(self, state).await {
+                    Ok(response) => Ok(response),
+                    Err(err) => match err.downcast::<Self::Error>() {
+                        Ok(err) => Err(jsonrpc::RpcError::MethodCallError(err)),
+                        Err(err) => Err(jsonrpc::RpcError::InternalError(err)),
+                    },
+                }
+            }
+        }
+    };
+}
+
+pub(crate) use _mount_method as mount_method;
--- a/crates/server/src/admin.rs
+++ b/crates/server/src/admin.rs
@@ -1,18 +1,14 @@
-use axum::{
-    http::StatusCode,
-    response::IntoResponse,
-    routing::{get, post},
-    Json, Router,
-};
-use base64::{engine::general_purpose::STANDARD, Engine};
+use axum::http::StatusCode;
+use axum::response::IntoResponse;
+use axum::routing::{get, post};
+use axum::{Json, Router};
+use base64::engine::general_purpose::STANDARD;
+use base64::Engine;
 use rand::{thread_rng, RngCore};
 use serde::{Deserialize, Serialize};
-use tower_http::{
-    services::{ServeDir, ServeFile},
-    set_status::SetStatus,
-};
-use tower_sessions::Session;
-use tower_sessions::{MemoryStore, SessionManagerLayer};
+use tower_http::services::{ServeDir, ServeFile};
+use tower_http::set_status::SetStatus;
+use tower_sessions::{MemoryStore, Session, SessionManagerLayer};
 use tracing::{error, info};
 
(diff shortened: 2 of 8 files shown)


crates/primitives/src/server.rs
```diff
@@ -55,3 +55,81 @@ pub enum WsCommand {
...
+    pub app_id: String,
+    pub method: String,
+    pub args_json: serde_json::Value,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(untagged)]
+pub enum JsonRpcRequestParams {
+    Call(JsonRpcRequestParamsCall),
+}
```
miraclx:
```suggestion
#[serde(tag = "method", content = "params")]
pub enum JsonRpcRequestPayload {
    Call(JsonRpcRequestParamsCall),
}
```

crates/primitives/src/server.rs
```diff
@@ -55,3 +55,81 @@ pub enum WsCommand {
...
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct JsonRpcRequest {
+    pub jsonrpc: JsonRpcVersion,
+    pub method: String,
+    pub params: Option<JsonRpcRequestParams>,
+    pub id: Option<WsRequestId>,
+}
```
miraclx:
```suggestion
    pub id: Option<WsRequestId>,
    #[serde(flatten)]
    pub payload: JsonRpcRequestPayload,
}
```

crates/primitives/src/server.rs
```diff
@@ -55,3 +55,81 @@ pub enum WsCommand {
...
+    pub code: u64,
+    pub message: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct JsonRpcResponse {
+    pub jsonrpc: JsonRpcVersion,
+    pub result: Option<JsonRpcResponseResult>,
+    pub error: Option<JsonRpcResponseError>,
```
miraclx: let's merge these by using a custom enum, take advantage of compiler-backed prohibitions as opposed to developer-informed ones

crates/server/src/jsonrpc.rs
```diff
@@ -0,0 +1,90 @@
...
+
+struct ServiceState {
+    server_sender: ServerSender,
+}
+
+pub(crate) fn service(
+    config: &crate::config::ServerConfig,
+    server_sender: ServerSender,
+) -> eyre::Result<Option<(&'static str, MethodRouter)>> {
+    let path = "/jsonrpc"; // todo! source from config
```
miraclx, line 37: you're not checking if `server.jsonrpc.enabled == true`

crates/server/src/jsonrpc.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    Ok(Some((path, post(handle_request).layer(Extension(state)))))
+}
+
+async fn handle_request(
+    Extension(state): Extension<Arc<ServiceState>>,
+    extract::Json(request): extract::Json<calimero_primitives::server::JsonRpcRequest>,
+) -> Json<calimero_primitives::server::JsonRpcResponse> {
+    let result = match request.method.as_str() {
+        "execute" => match request.params {
```
miraclx: I think a better model in the future would be (considering we'll be optimistically executing transactions)

1. There's only one way to call methods. If at runtime, the method mutates the state in any way, then once the call succeeds, broadcast the transaction. And if not, keep that execution private.
2. But right now, it's okay. Let's maybe use names like `call`, `call_mut`?

But even now, I'd design this differently and model it like meroship instead.

So, you have one enum for all the methods, impl some trait over each of those methods and associate their response & error types.

crates/server/Cargo.toml
```diff
@@ -38,5 +38,6 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [features]
+jsonrpc = []
 graphql = ["dep:async-graphql", "dep:async-graphql-axum"]
```
miraclx: it's fine, but:

```suggestion
graphql = ["dep:async-graphql", "dep:async-graphql-axum"]
jsonrpc = []
```

crates/primitives/src/server.rs
fbozic: Done, but probably could be improved more.

crates/server/src/jsonrpc.rs
fbozic: Done (to some extent), but most likely can be improved more.

crates/primitives/src/server.rs
fbozic: Improved more.

crates/server/src/jsonrpc.rs
fbozic: Done with macro all the way (I think) 😄

PR #49, opened 2024-03-25
feat: added simple smart contract for package management

chefsale:
# feat: added simple smart contract for package management)

## Test plan:
```
./test.sh
```

### DIFF
--- a/package-manager/contract/build.sh
+++ b/package-manager/contract/build.sh
@@ -6,7 +6,7 @@ cd "$(dirname $0)"
 
 TARGET="${CARGO_TARGET_DIR:-../../target}"
 
-cargo build --target wasm32-unknown-unknown --app-release
+cargo build --target wasm32-unknown-unknown --profile app-release
 
 mkdir -p res
 


package-manager/contract/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "package_manager"
+version = "0.1.0"
+authors = ["Calimero Limited <info@calimero.network>"]
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+near-sdk = "4.1.1"
```
MatejVukosav: Use versions from the workspace

package-manager/contract/sandbox-rs/src/tests.rs
```diff
@@ -0,0 +1,77 @@
...
+    assert_eq!(package["name"], "application".to_string());
+    assert_eq!(package["owner"], user.id().to_string());
+
+    let _ = user
+        .call(contract.id(), "add_release")
+        .args_json(json!({
+            "name": "application",
+            "version": "0.1.0",
+            "notes": "",
+            "path": "https://gateway/ipfs/CID",
```
MatejVukosav, line 59: Is `path` the same as `repository` above?

package-manager/contract/Cargo.toml
chefsale: Not sure if that is a good idea, as this is a smart contract and probably best to have versions pinned per smart contract. @miraclx wdyt?

package-manager/contract/sandbox-rs/src/tests.rs
chefsale: path is the link to the downloadable binary.

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    #[test]
+    fn test_add_package() {
+        let context = get_context(false);
+        testing_env!(context);
+        let mut contract = PackageManager::default();
+
+        contract.add_package(
+            "application".to_string(),
+            "Demo Application".to_string(),
+            "https://shithub.com/application".to_string(),
```
xilosada: man we will open source it xD

package-manager/contract/Cargo.toml
miraclx: I can think of compelling arguments for both approaches, but it should be fine, let's keep with convention and if needed, a special-needs contract can eventually opt-out.

Worth noting though is (provided you choose to reference root manifest version specifications) the `[profile.release]` section would have to be tuned to something like we currently have in `[profile.app-release]`. You can have a `lean` profile that NEAR smart contracts & Calimero Apps inherit from. Just in case we eventually wanted special tuning for each case independently.

And then to build the app in this profile, you would have to manually activate it via `cargo build --profile app-release`.

But ultimately, I won't worry much about this, eventually we'll prob replace this with a Calimero App instead, so it's fine in the interim.

package-manager/contract/sandbox-rs/src/tests.rs
```diff
@@ -0,0 +1,77 @@
...
+        .transact()
+        .await?
+        .into_result()?;
+
+    // begin tests
+    test_add_package_and_release(&bobo, &contract).await?;
+    Ok(())
+}
+
+async fn test_add_package_and_release(
```
miraclx, line 28: why not merge this into the contract itself and do all this under unit tests?

if the concern is the wasm path, replicate the behavior of the build script for calimero's apps

build, copy the wasm to the `res/..` folder. then the test can depend on it from there.

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        let package = self.packages.get(&name).unwrap();
+        if package.owner != env::signer_account_id() {
+            panic!("Sender is not the owner of the package");
+        }
+
+        if !self.releases.contains_key(&name) {
+            self.releases
+                .insert(name.clone(), LookupMap::new(StorageKeys::Release));
```
miraclx: The `Release` key isn't domain-scoped, so apps all write to the same storage entry. You can use the package name to scope it out. For example: `Release { pkg: sha256(name) }`

https://docs.near.org/sdk/rust/contract-structure/nesting#generating-unique-prefixes-for-persistent-collections

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+        version: String,
+        notes: String,
+        path: String,
+        hash: String,
+    ) {
+        if !self.packages.contains_key(&name) {
+            panic!("Package doesn't exist.")
+        }
+
+        let package = self.packages.get(&name).unwrap();
```
miraclx:
```suggestion
        let Some(package) = self.packages.get(&name) else {
            env::panic("Package doesn't exist.");
        };
```

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        self.releases.get_mut(&name).unwrap().insert(
+            version.clone(),
+            Release {
+                version,
+                notes,
+                path,
+                hash,
+            },
+        );
```
miraclx: here's a better pattern (pseudo-code):

```suggestion
        self.releases
            .entry(name.clone())
            .or_insert_with(|| LookupMap::new(StorageKeys::Release { .. }))
            .insert(..);
```

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // TODO: implement `pub fn get_packages(&self) -> Vec<Package> {}`
```
miraclx, line 96: prob with pagination (offset+limit) to avoid hitting the gas limit from excessive storage reads

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    // TODO: implement `pub fn get_releases(&self) -> Vec<Release> {}`
+
+    pub fn get_package(&self, name: String) -> Package {
+        self.packages
+            .get(&name)
+            .expect("Package doesn't exist")
+            .clone()
```
miraclx: You can return the reference to avoid the clone, the `#[near_bindgen]` macro just serializes it anyway, which only needs a reference.

```suggestion
    pub fn get_package(&self, name: String) -> &Package {
        self.packages
            .get(&name)
            .expect("Package doesn't exist")
```

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    pub fn get_release(&self, name: String, version: String) -> Release {
```
miraclx: same

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+            owner: owner,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use near_sdk::test_utils::VMContextBuilder;
+    use near_sdk::MockedBlockchain;
```
miraclx, line 132: surprised this works, does it?

package-manager/contract/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+#[derive(BorshStorageKey, BorshSerialize)]
+pub enum StorageKeys {
+    Packages,
+    Release,
+    Releases,
+}
+
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct PackageManager {
```
miraclx, line 16: perhaps for a later PR, but we need to enable ABI generation support

at which point we can replace the build script with [`cargo-near`](https://github.com/near/cargo-near)

package-manager/contract/src/lib.rs
chefsale: Works, I mean it compiles and tests are passing?

package-manager/contract/src/lib.rs
chefsale: Sounds good will add as TODO

package-manager/contract/src/lib.rs
chefsale: added to TODO

package-manager/contract/sandbox-rs/src/tests.rs
chefsale: This is a boiler plate code from near-create-app because they split unit test from integration tests which makes sense to me. Not sure that you want to tie them together?

package-manager/contract/build.sh
```diff
@@ -1,3 +1,17 @@
...
 rustup target add wasm32-unknown-unknown
-cargo build --target wasm32-unknown-unknown --release
\ No newline at end of file
+set -e
+
+cd "$(dirname $0)"
+
+TARGET="${CARGO_TARGET_DIR:-../../target}"
+
+cargo build --target wasm32-unknown-unknown --app-release
```
miraclx:
```suggestion
cargo build --target wasm32-unknown-unknown --profile app-release
```

package-manager/contract/src/lib.rs
miraclx: Interesting, I thought it was deprecated. Okay

PR #50, opened 2024-03-26
feat: add basic admin site

xilosada:
## Test plan

First go to node-ui folder and build the site:
`pnpm install && pnpm build`
for development you can just run the site using `pnpm dev`

Once the dist folder is created run the node and navigate to `http://127.0.0.1:2428/admin/`

https://github.com/calimero-is-near/cali2.0-experimental/assets/2929173/a1435e86-212a-4e62-8087-5bf52db4f28e

### DIFF
--- a/crates/server/src/admin.rs
+++ b/crates/server/src/admin.rs
@@ -39,7 +39,7 @@ pub(crate) fn service(
     let session_layer = SessionManagerLayer::new(session_store);
     let admin_router = Router::new()
         .route("/health", get(health_check_handler))
-        .route("/node-key", post(create_root_key_handler))
+        .route("/root-key", post(create_root_key_handler))
         .route("/request-challenge", post(request_challenge_handler))
         .layer(session_layer);
 
--- a/node-ui/src/pages/ConfirmWallet.jsx
+++ b/node-ui/src/pages/ConfirmWallet.jsx
@@ -26,7 +26,7 @@ const getParams = (location) => {
 
 
 const submitRootKeyRequest = async (params) => {
-    const response = await axios.post("/admin-api/node-key", params);
+    const response = await axios.post("/admin-api/root-key", params);
     const data = response.data;
     console.log('Response received:', data);
     return data;


crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let result = hasher.finalize();
+
+    let mut hash_array = [0u8; 32];
+    hash_array.copy_from_slice(&result);
+
+    hash_array
+}
+
+pub(crate) fn verify_signature(
```
xilosada, line 29: I think this should be moved to a crypto crate with different implementations for each network

crates/server/src/admin.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+async fn health_check_handler() -> Json<&'static str> {
+    Json("{\"status\":\"ok\"}")
+}
+
+async fn create_root_key_handler(session: Session, Json(req): Json<PubKeyRequest>) -> impl IntoResponse {
+    let message = "helloworld";
+    let app = "me";
+    let curl = "http://127.0.0.1:2428/admin/confirm-wallet";
```
xilosada, line 133: temporal until we define the challenge generation tomorrow

crates/server/src/admin.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+pub fn admin_router() -> Router {
+    let session_store = MemoryStore::default();
+    let session_layer = SessionManagerLayer::new(session_store);
+    Router::new()
+        .route("/health", get(health_check_handler))
+        .route("/node-key", post(create_root_key_handler))
+        .route("/request-challenge", post(request_challenge_handler))
+        .layer(session_layer)
```
xilosada: Ideally we can make the server stateless and use jwt. TBD tomorrow

node-ui/vite.config.js
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from 'vite'
+import { nodePolyfills } from 'vite-plugin-node-polyfills'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  base: '/admin/',
+  plugins: [
+    nodePolyfills(),
```
xilosada, line 8: required for crypto and buffer, both of them used by near libs

node-ui/src/App.jsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <>
+    <BrowserRouter basename="/admin">
+      <Routes>
+      <Route path="/" element={<Bootstrap />}/>
+      <Route path="/confirm-wallet" element={<ConfirmWallet />}/>
+          {/* <Route index element={<Home />} />
+          <Route path="blogs" element={<Blogs />} />
+          <Route path="contact" element={<Contact />} />
+          <Route path="*" element={<NoPage />} /> */}
```
xilosada, line 16: just a reference, Fran will take over the admin site

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+    let admin_router = Router::new()
+        .route("/health", get(health_check_handler))
+        .route("/node-key", post(create_root_key_handler))
+        .route("/request-challenge", post(request_challenge_handler))
+        .layer(session_layer);
+
+    Ok(Some((admin_path, admin_router)))
+}
+
+pub(crate) fn site(
```
miraclx, line 49: https://github.com/pyrossh/rust-embed/blob/master/examples/axum.rs should solve the bundling issue, it also has compression

And it should help merge everything into a single service

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+    config: &crate::config::ServerConfig,
+) -> eyre::Result<Option<(&'static str, ServeDir<SetStatus<ServeFile>>)>> {
+    let _config = match &config.admin {
+        Some(config) if config.enabled => config,
+        _ => {
+            info!("Admin site is disabled");
+            return Ok(None);
+        }
+    };
+    let path = "/admin";
```
miraclx, line 59: And as for paths, I think it may be permissible to use `/` as the site URL, so `localhost:2529` opens the admin UI if enabled.

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+        ServeFile::new(format!("{}/index.html", react_static_files_path)),
+    );
+
+    Ok(Some((path, react_app_serve_dir)))
+}
+
+pub const CHALLENGE_KEY: &str = "challenge";
+
+pub async fn request_challenge_handler(session: Session) -> impl IntoResponse {
+    if let Some(challenge) = session.get(CHALLENGE_KEY).await.unwrap_or(None) {
```
miraclx: the `Err` is to represent deserialization errors, it still returns an `Option<T>` from the happy path

```suggestion
    if let Some(challenge) = session.get(CHALLENGE_KEY).await.ok().flatten() {
```

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+        (StatusCode::OK, challenge)
+    } else {
+        // No challenge in session, generate a new one
+        let challenge = generate_challenge();
+        session.insert(CHALLENGE_KEY, &challenge).await.unwrap();
```
miraclx:
```suggestion
        if let Err(err) = session.insert(CHALLENGE_KEY, &challenge).await {
            error!("...");
            return (StatusCode::InternalError, ..);
```

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+fn generate_random_bytes() -> [u8; 32] {
+    let mut rng = thread_rng();
+    let mut buf = [0u8; 32];
+    rng.fill_bytes(&mut buf);
+    buf
+}
+
+fn generate_challenge() -> String {
+    let random_bytes = generate_random_bytes();
+    let encoded = STANDARD.encode(&random_bytes);
```
miraclx, line 119: we already have `bs58` in the dep tree, how about we reuse that instead of reintroducing a new one

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+    encoded
+}
+
+async fn health_check_handler() -> Json<&'static str> {
+    Json("{\"status\":\"ok\"}")
```
miraclx: I fear this may not work as you expect, because `Json` will serialize the contained data (which in this case is a string. So it (may) return `"{\\"status\\":\\"ok\\"}"`.

```suggestion
use serde_json::json;
...
async fn health_check_handler() -> Json<serde_json::Value> {
    Json(json!({ "status": "ok" }))
```

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+
+async fn create_root_key_handler(
+    session: Session,
+    Json(req): Json<PubKeyRequest>,
+) -> impl IntoResponse {
+    let message = "helloworld";
+    let app = "me";
+    let curl = "http://127.0.0.1:2428/admin/confirm-wallet";
+
+    match session.get::<String>(CHALLENGE_KEY).await.unwrap_or(None) {
```
miraclx:
```suggestion
    match session.get::<String>(CHALLENGE_KEY).await.ok() {
```

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+        Some(challenge) => {
+            if verifysignature::verify_signature(
+                &challenge,
+                message,
+                app,
+                curl,
+                &req.signature,
+                &req.public_key,
+            ) {
+                (StatusCode::OK, Json("\"status\": \"Root key created\""))
```
miraclx: same serialization concern here: might be worth creating a custom type at this point that impls `Serialize` and is passed here instead

crates/server/src/admin.rs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+        None => (
+            StatusCode::BAD_REQUEST,
+            Json("\"status\": \"No challenge found\""),
+        ),
+    }
+}
+
+#[derive(serde::Deserialize, serde::Serialize)]
```
miraclx: already imported, and doesn't need to be serializable

```suggestion
#[derive(Deserialize)]
```

crates/server/src/verifysignature.rs
miraclx, line 1: can you scope this under admin? and simplify the name?

crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
+use base64::{engine::general_purpose::STANDARD, Engine};
+use borsh::{BorshDeserialize, BorshSerialize};
+use near_crypto::{KeyType, PublicKey, Signature};
+use sha2::{Digest, Sha256};
+use std::convert::TryInto;
+use std::str::FromStr;
```
miraclx:
```suggestion
use std::convert::TryInto;
use std::str::FromStr;

use base64::{engine::general_purpose::STANDARD, Engine};
use borsh::{BorshDeserialize, BorshSerialize};
use near_crypto::{KeyType, PublicKey, Signature};
use sha2::{Digest, Sha256};
```

crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
...
+    let decoded_bytes = match STANDARD.decode(&challenge) {
+        Ok(bytes) => bytes,
+        Err(err) => {
+            eprintln!("Error decoding base64: {:?}", err);
+            return false;
+        }
+    };
+
+    let mut nonce_vec = decoded_bytes.to_vec();
+    nonce_vec.resize(32, 0);
```
miraclx:
```suggestion
    nonce_vec.truncate(32);
```

miraclx: but really, I think you mean to reject non-32 bytes instead? return false?

crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
...
+    let nonce: [u8; 32] = nonce_vec.try_into().unwrap();
+
+    let payload = create_payload(message, nonce, app, curl);
+    let mut borsh_payload: Vec<u8> = Vec::new();
+    payload.serialize(&mut borsh_payload).unwrap();
```
miraclx, line 43: please avoid the `unwrap`s

crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let signature = match Signature::from_parts(signature_type, &real_signature) {
+        Ok(sig) => sig,
+        Err(err) => {
+            eprintln!("Error creating signature: {:?}", err);
+            return false;
+        }
+    };
+
+    signature.verify(&message_signed, &public_key)
```
miraclx: we can rely on [ed25519_dalek](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/) directly for this, seems like the only other reason to pull in `near-crypto` is the `ed25519:<bs58-encoded public key>` parsing logic, which is replaceable, but permissible for now

crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let mut nonce_vec = decoded_bytes.to_vec();
+    nonce_vec.resize(32, 0);
+    let nonce: [u8; 32] = nonce_vec.try_into().unwrap();
+
+    let payload = create_payload(message, nonce, app, curl);
+    let mut borsh_payload: Vec<u8> = Vec::new();
+    payload.serialize(&mut borsh_payload).unwrap();
+
+    let message_signed = hash_bytes(&borsh_payload);
```
miraclx: or just message

```suggestion
    let message_unsigned = hash_bytes(&borsh_payload);
```

crates/server/src/verifysignature.rs
```diff
@@ -0,0 +1,136 @@
...
+    tag: u32,
+    message: String,
+    nonce: [u8; 32],
+    recipient: String,
+    callback_url: Option<String>,
+}
+
+fn create_payload(message: &str, nonce: [u8; 32], recipient: &str, callback_url: &str) -> Payload {
+    Payload {
+        tag: 2147484061,
```
miraclx, line 8: what is tag?

crates/server/src/admin.rs
xilosada: can you elaborate this? What bundling issue? A single service for the frontend and the backend?

crates/server/src/admin.rs
xilosada: we depend on the wallets so... https://github.com/mynearwallet/my-near-wallet/blob/master/packages/frontend/src/redux/slices/signMessage/index.js#L64

crates/server/src/verifysignature.rs
xilosada: Fixed, migrated to ed25519_dalek

crates/server/src/verifysignature.rs
xilosada: 👍

crates/server/src/verifysignature.rs
xilosada: created a helper method to do this conversions

crates/server/src/admin.rs
xilosada: TLDR; we need both

crates/server/src/admin.rs
xilosada: Fixed

crates/server/src/admin.rs
xilosada: uff, let's discuss this in the KT

crates/server/src/verifysignature.rs
xilosada: A number added in the [wallet-selector](https://github.com/near/wallet-selector/blob/fb0ee241ed1d700da9c46c21ae4107172651bd73/packages/core/src/lib/helpers/verify-signature/payload.ts#L14), it's briefly mentioned in the [nep](https://github.com/near/NEPs/pull/461)

crates/server/src/admin.rs
```diff
@@ -0,0 +1,162 @@
...
+            return Ok(None);
+        }
+    };
+    let admin_path = "/admin-api";
+
+    let session_store = MemoryStore::default();
+    let session_layer = SessionManagerLayer::new(session_store);
+    let admin_router = Router::new()
+        .route("/health", get(health_check_handler))
+        .route("/node-key", post(create_root_key_handler))
```
MatejVukosav: `/root-key` to follow nomenclature?

crates/server/src/admin.rs
miraclx: Bundling of the assets into the final binary. In response to:

> I still don't have a final decission about adding the dist folder as an asset and regenerate it in the pipeline.

PR #55, opened 2024-03-28
feat(package-manager): add package getter

frdomovic:
# feat(package-manager): add package getter
 
## Summary:
Included
-> getting all packages list
-> getting all releases with offsets and limits 
-> updated adding releases with fail if version is less than last added

### DIFF
--- a/package-manager/contract/src/lib.rs
+++ b/package-manager/contract/src/lib.rs
@@ -2,6 +2,7 @@ use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::store::UnorderedMap;
 use near_sdk::{env, near_bindgen, AccountId, BorshStorageKey};
+use semver;
 
 #[derive(BorshStorageKey, BorshSerialize)]
 #[borsh(crate = "near_sdk::borsh")]
@@ -74,13 +75,36 @@ impl PackageManager {
         path: String,
         hash: String,
     ) {
-        let Some(package) = self.packages.get(&name) else {
-            env::panic_str("Package doesn't exist.");
-        };
-        if package.owner != env::signer_account_id() {
-            env::panic_str("Sender is not the owner of the package");
+        // Get the last release version for the package
+        let last_release_version = self.releases.get(&name).map(|version_map| {
+            version_map
+                .keys()
+                .max_by(|a, b| {
+                    semver::Version::parse(a)
+                        .unwrap()
+                        .cmp(&semver::Version::parse(b).unwrap())
+                })
+                .expect("No versions found for the package")
+        });
+
+        // Check if the last release version exists and is less than the current version
+        if let Some(last_version) = last_release_version {
+            let last_version = semver::Version::parse(&last_version)
+                .expect("Failed to parse last release version");
+            let current_version =
+                semver::Version::parse(&version).expect("Failed to parse current version");
+            if current_version <= last_version {
+                panic!("New release version must be greater than the last release version.");
+            }
         }
 
+        // Check if the sender is the owner of the package
+        let package = self.packages.get(&name).expect("Package doesn't exist.");
+        if package.owner != env::signer_account_id() {
+            panic!("Sender is not the owner of the package");
+        }
+
+        // Insert the new release
         self.releases
             .entry(name.clone())
             .or_insert_with(|| {
@@ -108,8 +132,14 @@ impl PackageManager {
             .collect()
     }
 
-    // TODO: implement `pub fn get_releases(&self) -> Vec<Release> {}`
-    // with pagination (offset+limit) to avoid hitting the gas limit from excessive storage reads
+    pub fn get_releases(&self, offset: usize, limit: usize) -> Vec<Release> {
+        self.releases
+            .values()
+            .flat_map(|version_map| version_map.values().cloned())
+            .skip(offset)
+            .take(limit)
+            .collect()
+    }
 
     pub fn get_package(&self, name: String) -> &Package {
         self.packages.get(&name).expect("Package doesn't exist")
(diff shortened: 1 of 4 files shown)


package-manager/contract/src/lib.rs
```diff
@@ -1,9 +1,11 @@
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::serde::{Deserialize, Serialize};
-use near_sdk::store::LookupMap;
+use near_sdk::store::UnorderedMap;
 use near_sdk::{env, near_bindgen, AccountId, BorshStorageKey};
+use semver;
```
miraclx, line 5:
```suggestion
```

package-manager/contract/src/lib.rs
```diff
@@ -72,17 +75,40 @@ impl PackageManager {
...
+        // Get the last release version for the package
+        let last_release_version = self.releases.get(&name).map(|version_map| {
+            version_map
+                .keys()
+                .max_by(|a, b| {
+                    semver::Version::parse(a)
+                        .unwrap()
+                        .cmp(&semver::Version::parse(b).unwrap())
+                })
+                .expect("No versions found for the package")
```
miraclx, line 87:
```suggestion
        let last_release_version = self.releases.get(&name).and_then(|version_map| {
            version_map
                .keys()
                .and_then(|ver| semver::Version::parse(a).ok())
                .max_by(|a, b| a.cmp(&b))
```

package-manager/contract/src/lib.rs
```diff
@@ -72,17 +75,40 @@ impl PackageManager {
...
+        });
+
+        // Check if the last release version exists and is less than the current version
+        if let Some(last_version) = last_release_version {
+            let last_version = semver::Version::parse(&last_version)
+                .expect("Failed to parse last release version");
+            let current_version =
+                semver::Version::parse(&version).expect("Failed to parse current version");
```
miraclx, line 95: make use `version: semver::Version` in argument position instead

package-manager/contract/src/lib.rs
```diff
@@ -72,17 +75,40 @@ impl PackageManager {
...
+            if current_version <= last_version {
+                panic!("New release version must be greater than the last release version.");
+            }
+        }
+
+        // Check if the sender is the owner of the package
+        let package = self.packages.get(&name).expect("Package doesn't exist.");
         if package.owner != env::signer_account_id() {
-            env::panic_str("Sender is not the owner of the package");
+            panic!("Sender is not the owner of the package");
```
miraclx, line 104:
```suggestion
            env::panic_str("Sender is not the owner of the package");
```

https://docs.near.org/sdk/rust/best-practices#stdpanic-vs-envpanic

package-manager/contract/src/lib.rs
```diff
@@ -72,17 +75,40 @@ impl PackageManager {
...
+                .expect("Failed to parse last release version");
+            let current_version =
+                semver::Version::parse(&version).expect("Failed to parse current version");
+            if current_version <= last_version {
+                panic!("New release version must be greater than the last release version.");
+            }
+        }
+
+        // Check if the sender is the owner of the package
+        let package = self.packages.get(&name).expect("Package doesn't exist.");
```
miraclx, line 102: move this lookup to the top of the function, should happen first

package-manager/contract/src/lib.rs
```diff
@@ -13,15 +15,16 @@ pub enum StorageKeys {
 // TODO: enable ABI generation support
 #[near_bindgen]
 #[derive(BorshDeserialize, BorshSerialize)]
+#[borsh(crate = "near_sdk::borsh")]
 pub struct PackageManager {
-    packages: LookupMap<String, Package>,
-    releases: LookupMap<String, LookupMap<String, Release>>,
+    packages: UnorderedMap<String, Package>,
+    releases: UnorderedMap<String, UnorderedMap<String, Release>>,
```
miraclx, line 21: @chefsale, why not keep all the releases for each package inside the package itself?

package-manager/contract/src/lib.rs
```diff
@@ -97,11 +123,23 @@ impl PackageManager {
...
+            .keys()
+            .skip(offset)
+            .take(limit)
+            .filter_map(|key| self.packages.get(key).cloned())
+            .collect()
+    }
 
-    // TODO: implement `pub fn get_releases(&self) -> Vec<Release> {}`
-    // with pagination (offset+limit) to avoid hitting the gas limit from excessive storage reads
+    pub fn get_releases(&self, offset: usize, limit: usize) -> Vec<Release> {
```
miraclx, line 135: this should be scoped to a package anyway

package-manager/contract/src/lib.rs
```diff
@@ -97,11 +123,23 @@ impl PackageManager {
             );
     }
 
-    // TODO: implement `pub fn get_packages(&self) -> Vec<Package> {}`
-    // with pagination (offset+limit) to avoid hitting the gas limit from excessive storage reads
+    pub fn get_packages(&self, offset: usize, limit: usize) -> Vec<Package> {
```
miraclx, line 126:
```rust
pub fn get_packages(&self, offset: usize, limit: usize) -> Vec<&JustPackageMetadata>;
pub fn get_package(&self, name: String) -> PackageWithReleases;
pub fn search(
    &self,
    query: String,
    ct: Option<ContinuationToken>
) -> (Vec<&JustPackageMetadata>, Option<ContinuationToken>);
```

package-manager/contract/src/lib.rs
```diff
@@ -97,11 +123,23 @@ impl PackageManager {
...
+        self.packages
+            .keys()
```
miraclx, line 128: `UnorderedMap` has `.iter()` that let's you iterate over `(K, V)` pairs.

package-manager/contract/src/lib.rs
chefsale: Because it wasn't able to deserialize the near collection. And in case if we use an normal std map it will reload the whole thing.

package-manager/contract/src/lib.rs
chefsale: semver is needed.

PR #56, opened 2024-03-29
feat: implement support for routing events from multiple apps

fbozic:
- Refactor WebSocket server (same approach as JSON RPC server)
- Implement support for routing events from multiple apps (and potentially more types of node events)

### DIFF
--- a/crates/node/src/application_manager.rs
+++ b/crates/node/src/application_manager.rs
@@ -40,10 +40,8 @@ impl ApplicationManager {
             application.name, application_topic
         );
 
-        self.applications.insert(
-            application_topic.as_str().to_owned().into(),
-            application.clone(),
-        );
+        self.applications
+            .insert(application_topic.as_str().to_owned().into(), application);
 
         Ok(())
     }
@@ -68,10 +66,7 @@ impl ApplicationManager {
     ) -> eyre::Result<Vec<u8>> {
         match self.applications.get(application_id) {
             Some(application) => Ok(fs::read(&application.path)?),
-            None => eyre::bail!(
-                "failed to get application with id: {}",
-                application_id.as_ref()
-            ),
+            None => eyre::bail!("failed to get application with id: {}", application_id),
         }
     }
 }
--- a/crates/store/src/lib.rs
+++ b/crates/store/src/lib.rs
@@ -111,7 +111,8 @@ pub fn get_application_key(
     application_id: calimero_primitives::application::ApplicationId,
     key: &db::Key,
 ) -> Vec<u8> {
-    let mut application_key = Vec::from((application_id.as_ref().to_owned() + ":").as_bytes());
-    application_key.append(&mut key.clone());
+    let mut application_key = String::from(application_id.as_ref()).into_bytes();
+    application_key.push(b':');
+    application_key.extend_from_slice(&key[..]);
     application_key
 }


crates/node/src/lib.rs
```diff
@@ -323,16 +323,24 @@ impl Node {
             } => {
                 if self
                     .application_manager
-                    .is_application_registered(topic_hash)
+                    .is_application_registered(topic_hash.clone())
```
miraclx: change the signature of `is_application_registered` to take a ref instead. looks like it doesn't need an owned value

crates/server-primitives/src/ws.rs
```diff
@@ -26,29 +27,64 @@ pub enum RequestBody {
...
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub enum ResponseBodyResult {
-    Subscribed,
-    Unsubscribed,
-    Event(calimero_primitives::events::NodeEvent),
+pub enum ResponseBody {
+    Result(ResponseBodyResult),
```
miraclx: you can just put the `Value` in here directly

crates/server-primitives/src/ws.rs
```diff
@@ -26,29 +27,64 @@ pub enum RequestBody {
...
+pub struct SubscribeResponse {
+    pub application_ids: Vec<String>,
+}
+// *************************************************************************
+
+// **************************** unsubscribe method *******************************
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UnsubscribeRequest {
+    pub application_ids: Vec<String>,
```
miraclx:
```suggestion
    pub application_ids: Vec<ApplicationId>,
```

and other places

crates/server/src/ws.rs
```diff
@@ -180,67 +238,152 @@ async fn handle_commands(
...
             }
         }
     }
 }
 
 async fn handle_text_message(
-    client_id: ws_primitives::ClientId,
+    connection_id: ws_primitives::ConnectionId,
     state: Arc<ServiceState>,
+    connection_state: ConnectionState,
```
miraclx: with the other methods, you're extracting the connection state from the `state.connections` map, but here you're passing it directly

let's pick a pattern

crates/server/src/ws.rs
```diff
@@ -180,67 +238,152 @@ async fn handle_commands(
...
-            let error_body = ws_primitives::ResponseBodyError::SerdeError(String::from(format!(
-                "failed to deserialize request: {message}"
-            )));
-            ws_primitives::Response {
-                id: None,
-                body: ws_primitives::ResonseBody::Error(error_body),
-            }
+            let body =
+                ws_primitives::ResponseBody::Error(ws_primitives::ResponseBodyError::ServerError(
+                    ws_primitives::ServerResponseError::ParseError(String::from(format!(
```
miraclx: `format!` already returns a string, no need for `String::from` again

crates/server/src/ws.rs
```diff
@@ -180,67 +238,152 @@ async fn handle_commands(
...
+                        "failed to deserialize request: {message}"
+                    ))),
+                ));
+
+            ws_primitives::Response { id: None, body }
```
miraclx: If `id` is provided, but the body is invalid, like an invalid method call. Then we need to know this. And respond with the right id.

You can experiment with the pattern meroship uses.

Deserialize the json skeleton with `method`+`params` contained in a `serde_json::Value` that can be further deserialized in a secondary step.

If parsing fails at the first level, nothing is returned. You sent gibberish. I can't understand you.

If parsing fails at the second level, we can provide useful feedback about what occurred and for which `id`.

crates/server-primitives/src/ws.rs
fbozic: I don't understand what you mean

crates/node/src/lib.rs
fbozic: fixed, also application_manager uses ApplicationId for the keys in the map

crates/server-primitives/src/ws.rs
fbozic: fixed everywhere (I think)

crates/server/src/ws.rs
fbozic: done, good point

crates/server/src/ws.rs
fbozic: done

crates/server/src/ws.rs
fbozic: done

crates/server-primitives/src/ws.rs
miraclx: The type `ResponseBodyResult` only contains a `Value`, can be flattened here for brevity.

crates/node/src/application_manager.rs
```diff
@@ -14,13 +13,13 @@ pub struct Application {
 
 pub(crate) struct ApplicationManager {
     pub network_client: NetworkClient,
-    pub applications: HashMap<TopicHash, Application>,
+    pub applications: HashMap<calimero_primitives::application::ApplicationId, Application>,
```
miraclx, line 16: thank you

crates/node/src/application_manager.rs
```diff
@@ -37,8 +36,10 @@ impl ApplicationManager {
             .unwrap()
             .hash();
 
-        self.applications
-            .insert(app_topic.clone(), application.clone());
+        self.applications.insert(
+            calimero_primitives::application::ApplicationId(app_topic.clone().into_string()),
+            application.clone(),
+        );
```
miraclx: this operation is infallible, so can you move it after the `info!` step to save the clones?

crates/primitives/src/application.rs
```diff
@@ -2,3 +2,9 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Eq, Hash, Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct ApplicationId(pub String);
+
+impl ApplicationId {
+    pub fn to_string(self) -> String {
+        self.0
+    }
+}
```
miraclx: impl `Display` (which comes with `ToString`) and `Deref<Target = &str>` instead.

Your `Display` impl should forward to `String`'s implementation:

```rust
fmt::Display::fmt(&self.0, f)
```

Optionally consider `Into<String>` and `From<String>`

crates/node/src/lib.rs
```diff
@@ -119,7 +119,9 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
 
                         let tx_hash = match node
                             .call_mut(
-                                application_id.to_string(),
+                                calimero_primitives::application::ApplicationId(
```
miraclx: if you've got `From<String>` implemented for `ApplicationId`, then you can `application_id.to_owned().into()`

crates/store/src/lib.rs
```diff
@@ -101,8 +107,11 @@ impl ReadOnlyStore {
     }
 }
 
-pub fn get_application_key(application_id: String, key: &db::Key) -> Vec<u8> {
-    let mut application_key = Vec::from((application_id.clone() + ":").as_bytes());
+pub fn get_application_key(
+    application_id: calimero_primitives::application::ApplicationId,
+    key: &db::Key,
+) -> Vec<u8> {
+    let mut application_key = Vec::from((application_id.clone().to_string() + ":").as_bytes());
```
miraclx: you can drop the `clone`

crates/node/src/application_manager.rs
```diff
@@ -67,6 +67,12 @@ impl ApplicationManager {
         &self,
         application_id: &calimero_primitives::application::ApplicationId,
     ) -> eyre::Result<Vec<u8>> {
-        Ok(fs::read(&self.applications.get(application_id).unwrap().path).unwrap())
+        match self.applications.get(application_id) {
+            Some(application) => Ok(fs::read(&application.path)?),
+            None => eyre::bail!(
+                "failed to get application with id: {}",
+                application_id.clone().to_string()
```
miraclx: Once you have `Deref` implemented for `ApplicationId`, this can be:

```rust
&*application_id
```

or.. and preferably, impl `AsRef<str>` for `ApplicationId`, so this can be:

```rust
application_id.as_ref()
```

crates/server-primitives/src/ws.rs
fbozic: done

crates/node/src/application_manager.rs
```diff
@@ -35,18 +35,16 @@ impl ApplicationManager {
...
-        );
-
         info!(
             "Registered application {} with hash: {}",
             application.name, application_topic
         );
 
+        self.applications.insert(
+            application_topic.as_str().to_owned().into(),
+            application.clone(),
```
miraclx:
```suggestion
            application,
```

crates/node/src/application_manager.rs
```diff
@@ -72,7 +70,7 @@ impl ApplicationManager {
             Some(application) => Ok(fs::read(&application.path)?),
             None => eyre::bail!(
                 "failed to get application with id: {}",
-                application_id.clone().to_string()
+                application_id.as_ref()
```
miraclx: Since it now impls `Display`, this is much cleaner:

```suggestion
                application_id
```

crates/store/src/lib.rs
```diff
@@ -111,7 +111,7 @@ pub fn get_application_key(
     application_id: calimero_primitives::application::ApplicationId,
     key: &db::Key,
 ) -> Vec<u8> {
-    let mut application_key = Vec::from((application_id.clone().to_string() + ":").as_bytes());
+    let mut application_key = Vec::from((application_id.as_ref().to_owned() + ":").as_bytes());
```
miraclx: reading this now, I'd say since we already pass the application_id by value, maybe this pattern:

```suggestion
    let mut application_key = String::from(application_id).into_bytes();
    application_key.push(b':');
```

crates/store/src/lib.rs
```diff
@@ -111,7 +111,7 @@ pub fn get_application_key(
...
     application_key.append(&mut key.clone());
```
miraclx:
```suggestion
    application_key.extend_from_slice(&key[..]);
```

this should work

crates/store/src/lib.rs
fbozic: This doesn't compile. 

I've went with this
```rust

    let mut application_key = String::from(application_id.as_ref()).into_bytes();
    application_key.push(b':');
```

miraclx: Ah right, `From<T>` for `U` implies `Into<U>` for `T` but not the other way around. correct.

crates/server/src/ws.rs
```diff
@@ -118,43 +135,79 @@ async fn handle_socket(socket: WebSocket, state: Arc<ServiceState>) {
...
+            Err(err) => {
+                ws_primitives::ResponseBody::Error(ws_primitives::ResponseBodyError::ServerError(
+                    ws_primitives::ServerResponseError::InternalError {
+                        err: Some(err.into()),
+                    },
+                ))
+            }
+        };
+
+        let response = ws_primitives::Response { id: None, body };
```
miraclx, line 194: eventually, we might need to merge jsonrpc + websocket and make it generic over any transport, so we're not duplicating a lot of this logic

PR #57, opened 2024-03-29
feat: added the missing pkg name to get releases

chefsale:
```
./test.sh
```

### DIFF
--- a/package-manager/contract/src/lib.rs
+++ b/package-manager/contract/src/lib.rs
@@ -58,7 +58,7 @@ impl Default for PackageManager {
 impl PackageManager {
     pub fn add_package(&mut self, name: String, description: String, repository: String) {
         if self.packages.contains_key(&name) {
-            panic!("Package already exists.")
+            env::panic_str("Package already exists.")
         }
 
         self.packages.insert(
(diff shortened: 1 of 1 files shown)


package-manager/contract/src/lib.rs
```diff
@@ -132,12 +134,14 @@ impl PackageManager {
             .collect()
     }
 
-    pub fn get_releases(&self, offset: usize, limit: usize) -> Vec<Release> {
+    pub fn get_releases(&self, name: String, offset: usize, limit: usize) -> Vec<Release> {
```
miraclx:
```suggestion
    pub fn get_releases(&self, name: String, offset: usize, limit: usize) -> Vec<&Release> {
```

package-manager/contract/src/lib.rs
```diff
@@ -132,12 +134,14 @@ impl PackageManager {
...
         self.releases
-            .values()
-            .flat_map(|version_map| version_map.values().cloned())
+            .get(&name)
+            .expect("Package doesn't exist.")
+            .iter()
             .skip(offset)
             .take(limit)
+            .map(|(_, release)| release.clone())
```
miraclx:
```suggestion
            .map(|(_, release)| release)
```

PR #84, opened 2024-04-03
feat(admin-api): applications binary download

chefsale:
feat(admin-api): applications binary download

# Summary:
-> from Admin-UI select the package and release version to download
-> downloads the app wasm from the url saved in the contract and saves it under  `data/nodeX/apps/<app_name>/<release>/binary.wasm` 
-> verifies the release (if it fails deletes the local binary)
-> returns response (installed / failed + status)

### DIFF
--- a/crates/node/src/cli/init.rs
+++ b/crates/node/src/cli/init.rs
@@ -145,7 +145,6 @@ impl InitCommand {
                         enabled: true,
                         application_dir: root_args.home.join("apps"),
                     }),
-                    graphql: Some(calimero_server::graphql::GraphQLConfig { enabled: true }),
                     jsonrpc: Some(calimero_server::jsonrpc::JsonRpcConfig { enabled: true }),
                     websocket: Some(calimero_server::ws::WsConfig { enabled: true }),
                 },
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,8 +1140,8 @@ dependencies = [
  "calimero-primitives",
  "calimero-runtime",
  "calimero-server-primitives",
- "camino",
  "calimero-store",
+ "camino",
  "chrono",
  "color-eyre",
  "ed25519-dalek",
@@ -2782,20 +2782,6 @@ dependencies = [
  "tokio",
  "tokio-util 0.7.10",
  "tracing",
-]
-
-[[package]]
-name = "handlebars"
-version = "4.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "faa67bab9ff362228eb3d00bd024a4965d8231bbb7921167f0cfa66c6626b225"
-dependencies = [
- "log",
- "pest",
- "pest_derive",
- "serde",
- "serde_json",
- "thiserror",
 ]
 
 [[package]]


crates/node/src/cli/init.rs
```diff
@@ -2,7 +2,10 @@ use std::fs;
 use std::net::IpAddr;
 
 use calimero_network::config::{BootstrapConfig, BootstrapNodes, DiscoveryConfig, SwarmConfig};
-use calimero_node::config::{self, ConfigFile, NetworkConfig, StoreConfig};
+use calimero_node::{
```
miraclx: Please split these up, or run:

```console
cargo +nightly fmt
```

crates/server/src/admin.rs
```diff
@@ -1,2 +1,291 @@
...
+use std::io::Write;
+use std::sync::Arc;
+
+use axum::http::StatusCode;
+use axum::response::IntoResponse;
+use axum::routing::{get, post};
+use axum::{Extension, Json, Router};
+use base64::engine::general_purpose::STANDARD;
+use base64::Engine;
+use eyre::eyre;
```
miraclx: Let's localize this import. use `eyre::eyre!()` at the calling position.

crates/server/src/admin/service.rs
```diff
@@ -42,11 +52,15 @@ pub(crate) fn setup(
...
     let session_layer = SessionManagerLayer::new(session_store);
+    let state = Arc::new(ServiceState {
+        application_dir: config.application_dir.clone(),
+    });
     let admin_router = Router::new()
         .route("/health", get(health_check_handler))
         .route("/root-key", post(create_root_key_handler))
         .route("/request-challenge", post(request_challenge_handler))
         .route("/install-application", post(install_application_handler))
+        .layer(Extension(state))
```
MatejVukosav, line 68: check `with_state` method

crates/server/src/admin/service.rs
```diff
@@ -221,24 +235,71 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
     if let QueryResponseKind::CallResult(result) = response.kind {
         return Ok(from_slice::<Release>(&result.result)?);
     } else {
-        eyre::bail!("Failed to fetch data from the rpc endpoint")
+        Err(eyre!("Failed to fetch data from the rpc endpoint"))
```
MatejVukosav: Isn't bail exactly the same thing?

crates/server/src/admin/service.rs
```diff
@@ -221,24 +235,71 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+async fn install_application_handler(
+    Extension(state): Extension<Arc<ServiceState>>,
+    session: Session,
+    Json(req): Json<InstallApplicationRequest>,
+) -> impl IntoResponse {
+    let result = install_application(&req.application, &req.version, &state.application_dir).await;
+
+    match result {
+        Ok(()) => (StatusCode::OK, "Application Installed"),
+        Err(_) => (StatusCode::BAD_REQUEST, "Failed to install application"),
```
MatejVukosav: It may be interesting to know why install failed

crates/server/src/admin.rs
```diff
@@ -1,2 +1,291 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct Release {
+    pub version: String,
+    pub notes: String,
+    pub path: String,
+    pub hash: String,
+}
+
+pub async fn get_release(application: &String, version: &String) -> eyre::Result<Release> {
```
miraclx: wherever possible, try to reduce the requirement to `&str` unless you're absolutely certain you need a reference to the owned value.

crates/server/src/admin.rs
```diff
@@ -1,2 +1,291 @@
...
+use base64::engine::general_purpose::STANDARD;
+use base64::Engine;
+use eyre::eyre;
+use near_jsonrpc_client::{methods, JsonRpcClient};
+use near_jsonrpc_primitives::types::query::QueryResponseKind;
+use near_primitives::types::{AccountId, BlockReference, Finality, FunctionArgs};
+use near_primitives::views::QueryRequest;
+use rand::{thread_rng, RngCore};
+use serde::{Deserialize, Serialize};
+use serde_json::{from_slice, json};
```
miraclx: localize the use of `from_slice`, it has an ambiguous name.

crates/server/src/admin/service.rs
frdomovic: updated

crates/server/src/admin/service.rs

crates/server/src/admin/service.rs
frdomovic: added the error message parsing

crates/node/src/cli/init.rs
frdomovic: used the command to split

crates/server/src/admin.rs
frdomovic: updated

crates/server/src/admin.rs
frdomovic: updated

crates/server/src/admin.rs
frdomovic: updated

crates/server/src/admin/service.rs
miraclx: I'd be concerned here, because the `Router` is `nest`-ed into the main `Router` for all services, implying all other services now have the same state as this service.

Use `Extension` instead to keep things scoped out.

https://docs.rs/axum/latest/axum/extract/struct.State.html#combining-stateful-routers

Not a huge problem atm because no other service uses the `State<_>` extractor, but it's one of those things to keep in mind.

crates/server/src/admin/service.rs
```diff
@@ -219,26 +232,72 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+            eprintln!("Failed to delete file: {}", e);
+        }
+        return Err(e.into());
+    }
+
     Ok(())
 }
 
-pub async fn verify_release(release: Release, blob: String) {}
+pub async fn verify_release(hash: &String, release_hash: &String) -> eyre::Result<()> {
```
miraclx: same alloc-requirement

```suggestion
pub async fn verify_release(hash: &str, release_hash: &str) -> eyre::Result<()> {
```

crates/server/src/admin/service.rs
```diff
@@ -219,26 +232,72 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+        ));
+    }
+    Ok(())
+}
 
-pub async fn install_application(application: &String, version: &String) -> eyre::Result<()> {
-    download_release(get_release(application, version).await?).await
+pub async fn install_application(
+    application: &String,
+    version: &String,
```
miraclx: seim

crates/server/src/admin/service.rs
```diff
@@ -219,26 +232,72 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+    dir: &camino::Utf8PathBuf,
```
miraclx: here as well

```suggestion
    dir: &camino::Utf8Path,
```

crates/server/src/admin/service.rs
```diff
@@ -219,26 +232,72 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
     }
 }
 
-pub async fn download_release(release: Release) -> eyre::Result<()> {
-    let app_path = "";
-    //verify_release(release, blob);
+pub async fn download_release(
+    application: &String,
+    release: &Release,
+    dir: &camino::Utf8PathBuf,
```
```suggestion
    application: &str,
    release: &Release,
    dir: &camino::Utf8Path,
```

crates/server/src/admin/service.rs

crates/server/src/admin/service.rs

crates/server/src/admin/service.rs
frdomovic: reverted

PR #89, opened 2024-04-04
feat: implement jsonrpc client sdk

fbozic:


packages/calimero-sdk/src/rpc.ts
```diff
@@ -0,0 +1,92 @@
...
+export interface RpcClient {
+    call(params: RpcCallRequestParams, config: RequestConfig): Promise<RpcCallResponse>;
+    callMut(params: RpcCallRequestParams, config: RequestConfig): Promise<RpcCallResponse>;
+}
+
+export interface RequestConfig {
+    timeout?: number
+}
+
+// **************************** request *******************************
```
xilosada: Is this a code convention?

fbozic: not really, we have it like that in rust. probably can be removed in both places.

PR #102, opened 2024-04-08
sdk: codegen

miraclx:
Add macros for SDK code generation

- `#[app::logic]`
- `#[app::state]` & `#[app::state(emits = for<'a> Event<'a>)]`
- `#[app::event]`
- `app::emit!()`

I tried to avoid side-effect patterns, but this way of emitting events was the most intuitive, permitting event emission from anywhere, but ultimately, giving the liberty to determine which events should be serialized over the wire as "app-events" to the app author.

In a follow up PR, I'll define nested relationships between events, permitting an `#[app::event]` item to embed other `#[app::event]` items, delegating its serialization to the other event. (https://github.com/calimero-network/core/issues/243)

I went with this approach because I want to define a standard for what events look like. So they don't switch up on you and you have to test everything.

~In a follow up PR, I'll add the host syscalls for smuggling events, distinguishing them from logs.~ Done

~I also need to decouple the event item parsing to prevent serde injection. Yum~ Done, now I need to impl custom serialization to match the encode pattern defined in `AppEvent`. (https://github.com/calimero-network/core/issues/244)

I also need to clean up the tests, add all the cases living rent free in my head, as well as the documentation. (https://github.com/calimero-network/core/issues/245)

Follow up PR, follow up PR, follow up PR

But I'm pretty happy with what we have so far.

Notes for myself in a future PR:

~the proc-macro `emit!` doesn't play well with rust-analyzer, so I can't cmd select items. Needs investigating. Worst case, introduce an intermediate crate `calimero-sdk-macros-impl`, lol. That contains `macro_rules!` definitions that are annotated with `#[macro_export]` and can be carefully exposed in `calimero_sdk::app` like the rest of the proc macros.~

Restarted my computer and this has since been resolved, not sure if it's due to RA getting refreshed or if it got updated.

### DIFF
--- a/crates/sdk/macros/src/event.rs
+++ b/crates/sdk/macros/src/event.rs
@@ -34,13 +34,19 @@ impl<'a> ToTokens for EventImpl<'a> {
             #orig
 
             impl #impl_generics ::calimero_sdk::event::AppEvent for #ident #ty_generics #where_clause {
-                fn encode(&self) -> ::calimero_sdk::event::EncodedAppEvent {
+                fn kind(&self) -> ::std::borrow::Cow<str> {
                     // todo! revisit quick
                     match ::calimero_sdk::serde_json::to_value(self) {
-                        Ok(data) => ::calimero_sdk::event::EncodedAppEvent {
-                            kind: ::std::borrow::Cow::Owned(data["kind"].as_str().expect("Failed to get event kind").to_string()),
-                            data: ::std::borrow::Cow::Owned(::calimero_sdk::serde_json::to_vec(&data["data"]).expect("Failed to serialize event data")),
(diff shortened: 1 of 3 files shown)


apps/kv-store/src/lib.rs
```diff
@@ -1,35 +1,59 @@
...
 
+#[app::event]
+pub enum Event<'a> {
+    Inserted { key: &'a str, value: &'a str },
+    Updated { key: &'a str, value: &'a str },
+    Removed { key: &'a str },
+    Cleared,
+}
+
+#[app::logic]
```
fbozic, line 21: looks amazing, great stuff 🚀

PR #104, opened 2024-04-08
fix(admin-ui): fetch pk from storage

frdomovic:
# fix(admin-ui): fetch pk from storage

### DIFF
--- a/node-ui/src/utils/displayFunctions.js
+++ b/node-ui/src/utils/displayFunctions.js
@@ -9,9 +9,9 @@ export const truncatePublicKey = (publicKey) => {
 };
 
 export const getStatus = (active, revoked) => {
-  if (active && !revoked) {
+  if (active) {
     return "active";
-  } else if (revoked && !active) {
+  } else if (revoked) {
     return "revoked";
   } else {
     return "";


node-ui/src/utils/displayFunctions.js
```diff
@@ -0,0 +1,19 @@
+export const getDisplayPublicKey = (publicKey) => {
```
MatejVukosav: I would call it `truncatePublicKey`. 
In the simplest version you can just pass publicKey but it is open to more granular modification by keeping the same name.

node-ui/src/utils/displayFunctions.js
```diff
@@ -0,0 +1,19 @@
...
+  return `
+      ${publicKey.split(":")[1].substring(0, 4)}...${publicKey
+    .split(":")[1]
+    .substring(
+      publicKey.split(":")[1].length - 4,
+      publicKey.split(":")[1].length
+    )}`;
+};
+
+export const getStatus = (active, revoked) => {
```
MatejVukosav, line 11: I don't understand this function. Can `active` and `revoke` be true at the same time?

frdomovic: No they can't, the other option i'm referring to in else is both false meaning -> exist in node and not revoked

node-ui/src/utils/displayFunctions.js
frdomovic: renamed

node-ui/src/utils/displayFunctions.js
frdomovic: I mean the component part could look like this
```
{revoked && <p>revoked</p>}
{active && <p>active</p>} 
```
but just minified code with `<div class=status()>status()</div>`

MatejVukosav: Would this work?

```
if (active) {
    return "active";
  } else if (revoked) {
    return "revoked";
  } else {
    return "";
  }
```

frdomovic: right, updated

PR #106, opened 2024-04-08
feat: store root keys

MatejVukosav:
- store root keys under fixed node key `did:cali`. 
Root keys are on the node level with shared access for app auth.


crates/server/src/storage.rs
miraclx, line 1: Please move this under admin, I don't think anything else will use it. At least not these modules.

crates/server/src/storage/did.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    let did_document = Did {
+        id: DID_KEY.to_string(),
+        root_keys: Vec::<RootKey>::new(),
+    };
+
+    let did_document_vec = serde_json::to_vec(&did_document)
+        .map_err(|e| eyre::Report::new(e).wrap_err("Serialization error"))?;
+
+    storage.put(DID_KEY.to_string().into_bytes(), did_document_vec);
```
miraclx:
```suggestion
    storage.put(DID_KEY.as_bytes().to_owned(), did_document_vec);
```

crates/server/src/storage/did.rs
```diff
@@ -0,0 +1,52 @@
...
+    storage.commit()?;
+
+    Ok(did_document)
+}
+
+pub fn get_or_create_did(application_id: ApplicationId, store: &Store) -> eyre::Result<Did> {
+    let mut storage = calimero_store::ReadOnlyStore::new(application_id.clone(), &store);
+
+    let did_vec = storage.get(&DID_KEY.as_bytes().to_vec())?;
```
miraclx: 👍🏽

crates/server/src/storage/did.rs
```diff
@@ -0,0 +1,52 @@
...
+
+//TODO extract this to identity where suitable
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Did {
+    pub(crate) id: String,
+    pub(crate) root_keys: Vec<RootKey>,
+}
+
+pub fn create_did(application_id: ApplicationId, store: &Store) -> eyre::Result<Did> {
+    let mut storage = calimero_store::TemporalStore::new(application_id.clone(), &store);
```
miraclx: seems we can skip the clone here

```suggestion
    let mut storage = calimero_store::TemporalStore::new(application_id, &store);
```

crates/server/src/storage/root_key.rs
```diff
@@ -0,0 +1,46 @@
...
+        .root_keys
+        .iter()
+        .any(|k| k.signing_key == root_key.signing_key)
+    {
+        Ok(true)
+    } else {
+        did_document.root_keys.push(root_key);
+        update_did(application_id, store, did_document)?;
+        Ok(true)
+    }
```
miraclx: this is fine, but I would've written this instead

```suggestion
    if !did_document
        .root_keys
        .iter()
        .any(|k| k.signing_key == root_key.signing_key)
    {
        did_document.root_keys.push(root_key);
        update_did(application_id, store, did_document)?;
    }
    Ok(true)
```

crates/server/src/storage/root_key.rs
```diff
@@ -0,0 +1,46 @@
...
+pub(crate) struct RootKey {
+    pub(crate) signing_key: String,
+}
+
+pub fn add_root_key(
+    application_id: ApplicationId,
+    store: &Store,
+    root_key: RootKey,
+) -> eyre::Result<bool> {
+    let mut storage = calimero_store::TemporalStore::new(application_id.clone(), &store);
```
miraclx: this isn't used right?

PR #107, opened 2024-04-09
feat(typescript-sdk): implement subscrtions manager interface and websocket implementation

fbozic:


examples/only-peers-simple/README.md
```diff
@@ -1,56 +1,47 @@
...
 To build the production-ready version of the application, run:
 
 ```
-yarn build
+pnmp build
 ```
 This will create a dist directory containing the compiled and optimized version of the application. To preview the production build, you can run:
 
-```
+pnmp
```
MatejVukosav: wrong edit?

crates/primitives/src/events.rs
```diff
@@ -17,7 +17,7 @@ pub struct ApplicationEventPayload {
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
-#[serde(rename_all = "camelCase")]
+#[serde(tag = "type", content = "payload", rename_all = "snake_case")]
```
miraclx: use `#[serde(untagged)]` on `NodeEvent`

so notifs that include `application_id` are automatically inferred as `ApplicationEvent`

Also just pointing out that the pattern across the codebase is `type` & `data`, to be sure you've considered that.

Also, use `PascalCase` when serializing a type in value position.

examples/only-peers-simple/README.md
fbozic: yes

crates/primitives/src/events.rs
fbozic: I'll update it here, and I've added comment to this issue to follow up: https://github.com/calimero-network/core/issues/115#issuecomment-2049478257

PR #108, opened 2024-04-09
feat(admin-api): release verification

frdomovic:
# feat(admin-api): release verification
 
## Summary:
Fixed creation of folders and files for apps binaries.
Added verification for releases.

*ipfs urls down for relases
Screenshot 2024-04-09 at 16 40 55

Fixes # C20-47

## Test plan:

https://github.com/calimero-network/core/assets/93442516/f9fc1599-ed1a-42f9-9c70-9f27ad554a04

### DIFF
--- a/crates/server/src/admin.rs
+++ b/crates/server/src/admin.rs
@@ -14,7 +14,6 @@ use near_jsonrpc_primitives::types::query::QueryResponseKind;
 use near_primitives::types::{AccountId, BlockReference, Finality, FunctionArgs};
 use near_primitives::views::QueryRequest;
 use rand::{thread_rng, RngCore};
-use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use serde_json::{from_slice, json};
 use sha2::{Digest, Sha256};
@@ -221,8 +220,7 @@ pub async fn download_release(
     let file_path = format!("{}/binary.wasm", base_path);
     let mut file = File::create(&file_path)?;
 
-    let client = Client::new();
-    let mut response = client.get(&release.path).send().await?;
+    let mut response = reqwest::Client::new().get(&release.path).send().await?;
     let mut hasher = Sha256::new();
     while let Some(chunk) = response.chunk().await? {
         hasher.update(&chunk);
@@ -231,7 +229,12 @@ pub async fn download_release(
     let result = hasher.finalize();
     let hash = format!("{:x}", result);
 
-    verify_release(&hash, &release.hash).await?;
+    if let Err(e) = verify_release(&hash, &release.hash).await {
+        if let Err(e) = std::fs::remove_file(&file_path) {
+            eprintln!("Failed to delete file: {}", e);
+        }
+        return Err(e.into());
+    }
 
     Ok(())
 }


crates/server/src/admin.rs
```diff
@@ -207,21 +209,45 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+}
 
 pub async fn install_application(
     application: &String,
     version: &String,
     dir: &camino::Utf8PathBuf,
 ) -> eyre::Result<()> {
     let release = get_release(application, version).await?;
+    verify_release(&release.path, &release.hash).await?;
     download_release(application, &release, dir).await
```
miraclx: Looks like this implementation is downloading the blob twice. Why not hash the downloaded blob to verify instead?

crates/server/src/admin.rs
```diff
@@ -207,21 +209,45 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
 }
 
-pub async fn download_release(application: &String, release: &Release, dir: &camino::Utf8PathBuf) -> eyre::Result<()> {
-    let mut file = std::fs::File::create(dir.join("/".to_owned() + application + "/" + &release.version + "/binary.wasm"))?;
+pub async fn download_release(
+    application: &String,
+    release: &Release,
+    dir: &camino::Utf8PathBuf,
+) -> eyre::Result<()> {
     let mut content = Cursor::new(reqwest::get(&release.path).await?.bytes().await?);
```
miraclx: I'm quite concerned about this download implementation as it buffers the whole blob into memory, which is not the best.

`sha256` supports streaming, so we're not blocked on that

crates/server/src/admin.rs
frdomovic: Updated the download_release function to hash the blob and added verify to download_relase to remove redundant code

crates/server/src/admin.rs
```diff
@@ -207,14 +210,40 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+    dir: &camino::Utf8PathBuf,
+) -> eyre::Result<()> {
+    let base_path = format!("./{}/{}/{}", dir, application, &release.version);
+    fs::create_dir_all(&base_path)?;
+
+    let file_path = format!("{}/binary.wasm", base_path);
+    let mut file = File::create(&file_path)?;
+
+    let client = Client::new();
+    let mut response = client.get(&release.path).send().await?;
```
miraclx: I think the verbose `Client::` pattern is mostly useful when you intend to re-use the client across requests. Since you aren't, you can retain the same shorter pattern as before.

Shouldn't change much.

https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.chunk

crates/server/src/admin.rs
```diff
@@ -207,14 +210,40 @@ pub async fn get_release(application: &String, version: &String) -> eyre::Result
...
+    let mut hasher = Sha256::new();
+    while let Some(chunk) = response.chunk().await? {
+        hasher.update(&chunk);
+        file.write_all(&chunk)?;
```
miraclx, line 227: I like this much better now, thank you!

crates/server/src/admin.rs
frdomovic: Updated as before

PR #123, opened 2024-04-11
feat(package-manager): add id to package

frdomovic:
# feat(package-manager): add id to package
 
## Summary:
Added UID for packages where they are now fetched by its id (hash) and not by name. 
Hash is generated from application-name + caller 
Also releases are fetched by version + uid

Fixes # C20-50

## Test plan:
Screenshot 2024-04-11 at 15 53 14

### DIFF
--- a/contracts/package-manager/Cargo.toml
+++ b/contracts/package-manager/Cargo.toml
@@ -10,10 +10,10 @@ license.workspace = true
 crate-type = ["rlib", "cdylib"]
 
 [dependencies]
+hex.workspace = true
 near-sdk.workspace = true
 semver.workspace = true
 sha2.workspace = true
-hex.workspace = true
 
 [dev-dependencies]
 near-sdk = { workspace = true, features = ["unit-testing"] }
--- a/contracts/package-manager/src/lib.rs
+++ b/contracts/package-manager/src/lib.rs
@@ -1,6 +1,4 @@
-use hex;
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
-use near_sdk::env::sha256;
 use near_sdk::serde::{Deserialize, Serialize};
 use near_sdk::store::UnorderedMap;
 use near_sdk::{env, near_bindgen, AccountId, BorshStorageKey};
@@ -79,7 +77,7 @@ impl PackageManager {
     fn calculate_id_hash(name: &str) -> String {
         let author = env::signer_account_id();
         let id = format!("{}:{}", name, author);
-        let hash_bytes = sha256(id.as_bytes());
+        let hash_bytes = env::sha256(id.as_bytes());
         let hash_string = hex::encode(hash_bytes);
         return hash_string;
     }
@@ -185,11 +183,11 @@ impl Package {
         owner: AccountId,
     ) -> Self {
         Self {
-            id: id,
-            name: name,
-            description: description,
-            repository: repository,
-            owner: owner,
+            id,
+            name,
+            description,
+            repository,
+            owner,
         }
     }
 }


contracts/package-manager/src/lib.rs
```diff
@@ -56,13 +59,16 @@ impl Default for PackageManager {
 #[near_bindgen]
 impl PackageManager {
     pub fn add_package(&mut self, name: String, description: String, repository: String) {
-        if self.packages.contains_key(&name) {
+        let author = env::signer_account_id();
+        let id = format!("{}{}", name, author);
+        let id_hash = format!("{:x}", Sha256::digest(id.as_bytes()));
```
miraclx: You'd want to delegate that computation to the host

https://docs.rs/near-sdk/latest/near_sdk/env/fn.sha256.html

contracts/package-manager/src/lib.rs
```diff
@@ -56,13 +59,16 @@ impl Default for PackageManager {
 #[near_bindgen]
 impl PackageManager {
     pub fn add_package(&mut self, name: String, description: String, repository: String) {
-        if self.packages.contains_key(&name) {
+        let author = env::signer_account_id();
+        let id = format!("{}{}", name, author);
```
miraclx: Use a domain separator, for example these could collide:

name: `pkg`
author: `erlang.near`

and

name: `pkger`
author: `lang.near`

will both serialize to the `sha256("pkgerlang.near")` hash entry

It's guaranteed that near account IDs don't have `:` in them, so you can use that as a separator.

contracts/package-manager/src/lib.rs
```diff
@@ -56,13 +59,16 @@ impl Default for PackageManager {
...
+        let id_hash = format!("{:x}", Sha256::digest(id.as_bytes()));
+        if self.packages.contains_key(&id_hash) {
             env::panic_str("Package already exists.")
         }
 
         self.packages.insert(
-            name.clone(),
-            Package::new(name, description, repository, env::signer_account_id()),
+            id_hash.clone(),
```
miraclx, line 66: I think you can drop clone now

contracts/package-manager/src/lib.rs
```diff
@@ -74,8 +80,11 @@ impl PackageManager {
         path: String,
         hash: String,
     ) {
+        let author = env::signer_account_id();
+        let id = format!("{}{}", name, author);
```
miraclx: Just add a private convenience method for deriving the hash from these components so you don't have to duplicate the logic

contracts/package-manager/src/lib.rs
frdomovic: updated

contracts/package-manager/src/lib.rs
frdomovic: updated

contracts/package-manager/src/lib.rs
frdomovic: extracted

contracts/package-manager/src/lib.rs
frdomovic: does not work without it, if i use id_hash says the value is moved for the next line, if i use &id_hash for key it says to remove &

contracts/package-manager/Cargo.toml
```diff
@@ -13,6 +13,7 @@ crate-type = ["rlib", "cdylib"]
 near-sdk.workspace = true
 semver.workspace = true
 sha2.workspace = true
+hex.workspace = true
```
miraclx: nit: order :-)

contracts/package-manager/src/lib.rs
```diff
@@ -1,9 +1,9 @@
+use hex;
```
miraclx: no need for this, weird that the SDK doc has it there

contracts/package-manager/src/lib.rs
```diff
@@ -1,9 +1,9 @@
...
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+use near_sdk::env::sha256;
```
miraclx: just use `env::sha256` at the calling location.

contracts/package-manager/src/lib.rs
miraclx: Ah, right.. `Package::new` requires it, missed that.

contracts/package-manager/src/lib.rs
```diff
@@ -144,22 +163,29 @@ impl PackageManager {
...
+        id: String,
+        name: String,
+        description: String,
+        repository: String,
+        owner: AccountId,
+    ) -> Self {
         Self {
+            id: id,
             name: name,
             description: description,
```
miraclx: for brevity

```suggestion
            id,
            name,
            description,
```

... and others

PR #133, opened 2024-04-11
feat(app-manager): update app manager

frdomovic:
# feat(app-manager): update app manager
 
## Summary:

Removed application registration and added check for package locally -> installs the latest version of wasm

### DIFF
--- a/crates/node/src/application_manager.rs
+++ b/crates/node/src/application_manager.rs
@@ -63,8 +63,8 @@ impl ApplicationManager {
                     let entry_path = entry.path();
 
                     let version = {
-                        let folder_name = entry_path.file_name()?.to_string_lossy().into_owned();
-                        semver::Version::parse(&folder_name).ok()?
+                        semver::Version::parse(entry_path.file_name()?.to_string_lossy().as_ref())
+                            .ok()?
                     };
 
                     let binary_path = entry_path.join("binary.wasm");


crates/node/Cargo.toml
```diff
@@ -21,6 +21,7 @@ serde_json.workspace = true
 tokio = { workspace = true, features = ["io-std", "macros"] }
 toml.workspace = true
 tracing.workspace = true
+semver.workspace = true
```
miraclx: nit: order

crates/node/src/application_manager.rs
```diff
@@ -57,16 +35,70 @@ impl ApplicationManager {
         &self,
         application_id: &calimero_primitives::application::ApplicationId,
     ) -> bool {
-        self.applications.contains_key(application_id)
+        if let Some(latest_version_path) = self.get_latest_application_path(application_id) {
+            true
+        } else {
+            false
+        }
```
miraclx:
```suggestion
	self.get_latest_application_path(application_id).is_some()
```

crates/node/src/application_manager.rs
```diff
@@ -57,16 +35,70 @@ impl ApplicationManager {
...
+        &self,
+        application_id: &calimero_primitives::application::ApplicationId,
+    ) -> Option<String> {
+        let application_base_path = self.application_dir.join(application_id.to_string());
+
+        if let Ok(entries) = fs::read_dir(&application_base_path) {
+            // Collect version folders that contain binary.wasm into a vector
+            let mut versions_with_binary = entries
+                .filter_map(|entry| {
+                    if let Ok(entry) = entry {
```
miraclx: compress these depths

```suggestion
                    let entry = entry.ok()?;
```

crates/node/src/application_manager.rs
```diff
@@ -57,16 +35,70 @@ impl ApplicationManager {
...
+                        let entry_path = entry.path();
+                        if entry_path.is_dir() {
```
miraclx:
```suggestion
                        entry_path.is_dir().then_some(())?;
```

crates/node/src/application_manager.rs
```diff
@@ -57,16 +35,70 @@ impl ApplicationManager {
...
+                            if let Some(folder_name) = entry_path.file_name() {
```
miraclx: just follow the pattern, you should have fewer indents.

crates/node/Cargo.toml
frdomovic: updated

crates/node/src/application_manager.rs
frdomovic: updated

crates/node/src/application_manager.rs
```diff
@@ -63,30 +59,16 @@ impl ApplicationManager {
...
-                        }
-                    } else {
-                        None
-                    }
+                    let entry = entry.ok()?;
+                    let entry_path = entry.path();
+
+                    let version = {
+                        let folder_name = entry_path.file_name()?.to_string_lossy().into_owned();
+                        semver::Version::parse(&folder_name).ok()?
```
miraclx: You should be able to avoid an allocation here since `Version::parse` only requires a `&str`.

Did you try

```rust
semver::Version::parse(entry_path.file_name()?.to_string_lossy().as_ref()).ok()?
```

frdomovic: Yes this works

crates/node/src/application_manager.rs
```diff
@@ -13,39 +13,17 @@ pub struct Application {
...
-    pub async fn register_application(&mut self, application: Application) -> eyre::Result<()> {
-        let application_blob = fs::read(&application.path)?;
-        let application_topic = self
-            .network_client
-            .subscribe(calimero_network::types::IdentTopic::new(format!(
-                "/calimero/experimental/app/{}",
-                calimero_primitives::hash::Hash::hash(&application_blob),
-            )))
-            .await?
-            .hash();
```
fbozic, line 36: I think this is the root cause for this issue: https://github.com/calimero-network/core/issues/209
We are not subscribing to a topic anymore at all from application_manager.

frdomovic: @miraclx

PR #138, opened 2024-04-15
feat(docs): admin api docs

frdomovic:
# feat(docs): admin api docs
 

https://github.com/calimero-network/core/assets/93442516/d91aa271-da90-40fa-a374-019b3d295587


docs/docs/admin-api.mdx
```diff
@@ -0,0 +1,187 @@
...
+   -H "Content-Type: application/json" \
+   -d '{"wallet_signature": "<WALLET_SIGNATURE>", "payload": { "message": <MESSAGE>, "metadata": <METADATA_VALUE>}. "wallet_metadata": {"wallet_type": <WALLET_TYPE>, "signing_key": <SIGNING_KEY>}}' \
+   https://<NODE-URL>/admin-api/add-client-key
+   ```
+
+   **Example response**
+
+   ```
+   > Success response
+       Client key stored
```
MatejVukosav, line 152: This is not the correct response. 

client-key returns data:{} or error: {}.

Also, we will refactor other endpoints to return the same.

docs/docs/admin-api.mdx
```diff
@@ -0,0 +1,187 @@
...
+- Request validation challenge
+- Install and list node applications
+- Fetch decentralized identifiers (DID)
+- Manage client keys
+
+In essence, the Admin API simplifies node management, while the Admin UI offers an intuitive interface for seamless control.
+
+### API Endpoints
+
+1. **Root Key Request** - This endpoint allows administrators to request root keys for secure authentication and access control within the decentralized network environment.
```
MatejVukosav, line 26: POST /root-key enables adding root key, not requesting.

docs/docs/admin-api.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+   Note: The parameters for this POST request, including accountId, signature, and publicKey, are generated by signing the challenge and message with the wallet.
+
+   **Example call with curl**
+
+   ```
+   curl -X POST \
+   -H "Content-Type: application/json" \
+   -d '{"accountId": "<ACCOUNT_ID>", "signature": "<SIGNATURE>", "publicKey": "<PUBLIC_KEY>", "callbackUrl": "<CALLBACK_URL>"}' \
+   https://<NODE-URL>/admin-api/root-key
```
MatejVukosav, line 46: Currently, we don't have `https` so it should be `http`

docs/docs/admin-api.mdx
```diff
@@ -0,0 +1,187 @@
...
+- Manage client keys
+
+In essence, the Admin API simplifies node management, while the Admin UI offers an intuitive interface for seamless control.
+
+### API Endpoints
+
+1. **Root Key Request** - This endpoint allows administrators to request root keys for secure authentication and access control within the decentralized network environment.
+   Validation Challenge Generation
+
+   **POST** `/root-key`
```
MatejVukosav, line 29: This endpoint requires calling `/request-challenge` as currently challenge is added to the session directly. Otherwise it won't work.

docs/docs/admin-api.mdx
```diff
@@ -0,0 +1,187 @@
...
+### About Admin Client API
+
+Our Admin Client API offers a comprehensive set of endpoints designed to facilitate the management and administration of the node states.
+With intuitive routes and robust handlers, administrators can seamlessly navigate various node operations with ease and efficiency.
+The Admin Client API, paired with the Admin UI—a web application, is your go-to tool for managing node states,
+simplifying interactions with the server.
+
+The Admin API empowers you to:
+
+- Generate root keys
```
xilosada, line 16: Revoke root keys

docs/docs/admin-api.mdx
xilosada: 👍

docs/docs/admin-api.mdx
xilosada: I wouldn't go so detailed about the methods we expose in the API but how to do it through the UI. With some screenshots even better

docs/docs/admin-api.mdx
```diff
@@ -0,0 +1,187 @@
...
+            [{
+                "signing_key": <signing_key>
+            }]
+        }
+       }
+   > Error response
+       Failed to fetch DID: <error_message>
+   ```
+
+### How to Use
```
xilosada, line 183: Ok these should go first. Or even in another page, we can create subsections in the docs

docs/docs/admin-api.mdx
MatejVukosav: In that case we should only keep `/request-challenge` and `/add-client-key`

PR #153, opened 2024-04-15
feat(admin-ui + admin-api): connect apps api with ui

frdomovic:
# feat(admin-ui + admin-api): connect apps api with ui
 
Screenshot 2024-04-15 at 18 17 32


node-ui/src/components/applications/ApplicationsTable.jsx
```diff
@@ -129,16 +129,18 @@ export function ApplicationsTable({ applications, install, uninstall }) {
...
-                  <MenuIconDropdown
-                    options={[
-                      {
-                        buttonText: t.uninstallButtonText,
-                        onClick: () => uninstall(id),
-                      },
-                    ]}
-                  />
-                </div>
+                {false && (
```
frdomovic, line 132: not implemented

crates/server/src/admin/service.rs
```diff
@@ -60,14 +61,17 @@ pub(crate) fn setup(
...
         application_dir: config.application_dir.clone(),
     });
+
     let admin_router = Router::new()
         .route("/health", get(health_check_handler))
         .route("/root-key", post(create_root_key_handler))
         .route("/request-challenge", post(request_challenge_handler))
         .route("/install-application", post(install_application_handler))
-        .layer(Extension(state))
+        .layer(Extension(state.clone()))
```
MatejVukosav: I have updated this part in the new PR so we have shared state https://github.com/calimero-network/core/pull/154

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
...
+    let application_base_path = dir.join(application_id.to_string());
+
+    if let Ok(entries) = fs::read_dir(&application_base_path) {
+        let mut versions_with_binary = entries
+            .filter_map(|entry| {
+                let entry = entry.ok()?;
+                let entry_path = entry.path();
+
+                let version = {
+                    semver::Version::parse(entry_path.file_name()?.to_string_lossy().as_ref())
```
MatejVukosav, line 430: why `to_string_lossy()`?

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
...
+                };
+
+                let binary_path = entry_path.join("binary.wasm");
+                binary_path.exists().then_some((version, entry_path))
+            })
+            .collect::<Vec<_>>();
+
+        versions_with_binary.sort_by(|a, b| b.0.cmp(&a.0));
+
+        if let Some((_, path)) = versions_with_binary.first() {
```
MatejVukosav: First already returns none if there are no element
Check this
```
versions_with_binary.first()
            .map(|(_, path)| path.to_string_lossy().into_owned())
```

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
...
+                    get_latest_application_path(&state.application_dir, &file_name);
+                if let Some(latest_version) = latest_version {
+                    let app_name = file_name.to_string();
+                    if let Some(version) = latest_version.rsplit('/').next() {
+                        applications.insert(app_name, version.to_string());
+                    }
+                }
+            }
+        });
+        let response_body = json!({
```
MatejVukosav: Create struct and return it

node-ui/src/pages/Applications.jsx
```diff
@@ -27,6 +29,21 @@ export default function Applications() {
     }
   }, [packages]);
 
+  useEffect(() => {
+    const setApps = async () => {
+      const applicationIds = Object.keys(
+        JSON.parse((await axios.get("/admin-api/applications")).data).apps
```
MatejVukosav: Encapsulate api call outside page. You can check only-peers frontend for reference

crates/server/src/admin/service.rs
frdomovic: Will rebease after merged

crates/server/src/admin/service.rs
frdomovic: to_string()  / to_str() / just as_ref() -> the trait bound `OsStr: AsRef<str>` is not satisfied

crates/server/src/admin/service.rs
frdomovic: Yep this works, updated

node-ui/src/pages/Applications.jsx
frdomovic: Updated

crates/server/src/admin/service.rs
miraclx: why duplicate these entries?

Just put the extension at the end, all the handlers will be able to access it

crates/server/src/admin/service.rs
miraclx: OS file names are not guaranteed to be UTF-8 strings, so you have to attempt at either a lossy or fallible conversion

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
     .into_response())
 }
 
+fn get_latest_application_path(dir: &camino::Utf8Path, application_id: &str) -> Option<String> {
+    let application_base_path = dir.join(application_id.to_string());
+
+    if let Ok(entries) = fs::read_dir(&application_base_path) {
```
miraclx, line 423: sucks that we're doing blocking IO in an async environment, we can refactor after token

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
     .into_response())
 }
 
+fn get_latest_application_path(dir: &camino::Utf8Path, application_id: &str) -> Option<String> {
```
miraclx: Return an `Option<PathBuf>` here

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
...
+                let binary_path = entry_path.join("binary.wasm");
+                binary_path.exists().then_some((version, entry_path))
+            })
+            .collect::<Vec<_>>();
+
+        versions_with_binary.sort_by(|a, b| b.0.cmp(&a.0));
+
+        versions_with_binary
+            .first()
+            .map(|(_, path)| path.to_string_lossy().into_owned())
```
miraclx: not a healthy conversion, if the path is useless to read from the file

and ideally, we should be using [camino](https://docs.rs/camino) which will do a falible conversion to UTF-8 to get that guarantee, and we already have it in the codebase

And this already has serde impls (behind a feature flag), so you won't need to convert to `String` at all

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
...
+
+async fn fetch_application_handler(
+    Extension(state): Extension<Arc<ServiceState>>,
+    session: Session,
+) -> impl IntoResponse {
+    if let Ok(entries) = fs::read_dir(&state.application_dir) {
+        let mut applications: HashMap<String, String> = HashMap::new();
+
+        entries.filter_map(|entry| entry.ok()).for_each(|entry| {
+            if let Some(file_name) = entry.file_name().to_str() {
```
miraclx, line 468: maybe we add a log to the CLI that we're unable to read the application ID, so it doesn't silently ignore it

a lossy conversion to aid display will be necessary since we can only print utf-8 data

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,72 @@ async fn install_application_handler(
...
+                let latest_version =
+                    get_latest_application_path(&state.application_dir, &file_name);
+                if let Some(latest_version) = latest_version {
+                    let app_name = file_name.to_string();
+                    if let Some(version) = latest_version.rsplit('/').next() {
```
miraclx: looks like this is guatanteed to be true, unless you intend to use `get_latest_application_path` elsewhere in this admin page, you can just return the version itself.

crates/server/Cargo.toml
```diff
@@ -25,6 +25,7 @@ rand.workspace = true
 reqwest.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
+semver.workspace = true
```
miraclx: nit: order

crates/server/src/admin/service.rs
frdomovic: Added at the end

crates/server/src/admin/service.rs
frdomovic: upated

crates/server/src/admin/service.rs

crates/server/src/admin/service.rs
```diff
@@ -336,7 +336,7 @@ async fn install_application_handler(
     .into_response())
 }
 
-fn get_latest_application_path(dir: &camino::Utf8Path, application_id: &str) -> Option<String> {
+fn get_latest_application_path(dir: &camino::Utf8Path, application_id: &str) -> Option<PathBuf> {
```
miraclx: just saw the input here is `camino::Utf8Path`, please return `camino::Utf8PathBuf`

crates/server/src/admin/service.rs
```diff
@@ -383,10 +386,15 @@ async fn fetch_application_handler(
                     get_latest_application_path(&state.application_dir, &file_name);
                 if let Some(latest_version) = latest_version {
                     let app_name = file_name.to_string();
-                    if let Some(version) = latest_version.rsplit('/').next() {
-                        applications.insert(app_name, version.to_string());
-                    }
+                    let version = latest_version
+                        .file_name()
+                        .unwrap()
+                        .to_str()
```
miraclx: why this change?

miraclx: to clarify, I meant return the version itself from the `get_latest_application_path` instead of a path since it's not used elsewhere

make it `find_latest_version`, and return the `(version, path)` pair you already have

crates/server/src/admin/service.rs
```diff
@@ -332,6 +336,80 @@ async fn install_application_handler(
...
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+
+        versions_with_binary.sort_by(|a, b| b.0.cmp(&a.0));
+
+        versions_with_binary
+            .first()
+            .map(|(_, path)| PathBuf::from(path))
```
miraclx: that is, remove this

```suggestion
```

PR #154, opened 2024-04-15
feat:  implement challenge endpoint

MatejVukosav:
Request challenge

applicationId is optional for admin ui and node related requests.

```
Response
{
    "data": {
        "nonce": "RpnSWfVAm1Cz+Tg2vxc2/DV9lRXLJAirm8Xkv5eZbHo=",
        "applicationId": "some_app_id",
        "timestamp": 1713211405,
        "nodeSignature": "nYV2gjlWRAfICUPR3M3e4knDMq4c5YvaByt59DXjF6Z6MYUKHVSA8KNU/wIBfYLLV68Hl9t91wQmyicfJ1IgBQ=="
    }
}
```

### DIFF
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -52,7 +52,7 @@ pub struct AdminState {
 pub(crate) fn setup(
     config: &crate::config::ServerConfig,
     store: Store,
-) -> eyre::Result<Option<(&'static str, Router<Arc<AdminState>>)>> {
+) -> eyre::Result<Option<(&'static str, Router)>> {
     let admin_config = match &config.admin {
         Some(config) if config.enabled => config,
         _ => {
@@ -167,7 +167,7 @@ struct RequestChallengeResponse {
 
 pub async fn request_challenge_handler(
     session: Session,
-    State(state): State<Arc<AdminState>>,
+    Extension(state): Extension<Arc<AdminState>>,
     Json(req): Json<RequestChallenge>,
 ) -> impl IntoResponse {
     if let Some(challenge) = session.get::<String>(CHALLENGE_KEY).await.ok().flatten() {
@@ -272,7 +272,7 @@ async fn health_check_handler() -> impl IntoResponse {
 
 async fn create_root_key_handler(
     session: Session,
-    State(state): State<Arc<AdminState>>,
+    Extension(state): Extension<Arc<AdminState>>,
     Json(req): Json<PubKeyRequest>,
 ) -> impl IntoResponse {
     let message = "helloworld";
--- a/crates/server/src/admin/handlers/add_client_key.rs
+++ b/crates/server/src/admin/handlers/add_client_key.rs
@@ -1,9 +1,8 @@
 use std::sync::Arc;
 
-use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
-use axum::Json;
+use axum::{Extension, Json};
 use base64::engine::general_purpose::STANDARD;
 use base64::Engine;
 use calimero_identity::auth::verify_eth_signature;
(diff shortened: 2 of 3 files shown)


crates/server/src/admin/service.rs
```diff
@@ -37,38 +38,52 @@ pub struct AdminConfig {
     pub application_dir: camino::Utf8PathBuf,
 }
 
+#[derive(Clone)]
 pub(crate) struct ServiceState {
     application_dir: camino::Utf8PathBuf,
 }
 
+#[derive(Clone)]
```
miraclx: Can you externalize this instead?

So callers of `state.clone()` are aware of it's behaviour, it's unclear from the caller perspective if this will be a ref-counted clone or re-alloc. You cannot guarantee the `Clone` behavior of your dependencies. `Store` and `Keypair`.

Wrap all your uses of `AdminState` in an `Arc`, so you have `Arc<AdminState>`

crates/server/src/admin/service.rs
```diff
@@ -315,7 +402,7 @@ pub async fn install_application(
 
 async fn install_application_handler(
     Extension(state): Extension<Arc<ServiceState>>,
-    session: Session,
+    _session: Session,
```
miraclx: you should be able to just remove this extractor if it's unused

crates/server/src/admin/service.rs
```diff
@@ -37,38 +38,52 @@ pub struct AdminConfig {
...
         .route("/health", get(health_check_handler))
         .route("/root-key", post(create_root_key_handler))
         .route("/request-challenge", post(request_challenge_handler))
         .route("/install-application", post(install_application_handler))
-        .layer(Extension(state))
         .route("/add-client-key", post(add_client_key_handler))
         .route("/did", get(fetch_did_handler))
-        .layer(session_layer)
-        .with_state(store);
+        .with_state(shared_state)
```
miraclx: please use `Extension`, otherwise this state will be inherited by non-admin UI method routers as well

crates/server/src/admin/service.rs
```diff
@@ -175,10 +218,54 @@ fn generate_random_bytes() -> [u8; 32] {
...
+    application_id: Option<String>,
+    keypair: &Keypair,
+) -> Result<NodeChallenge, ApiError> {
     let random_bytes = generate_random_bytes();
     let encoded = STANDARD.encode(&random_bytes);
-    encoded
+
+    let node_challenge_message = NodeChallengeMessage {
+        nonce: encoded,
+        application_id: application_id.unwrap_or_else(|| "".to_string()),
```
miraclx: hint: `unwrap_or_default()` is a little convenience pattern for the same thing

crates/server/src/admin/service.rs
```diff
@@ -175,10 +218,54 @@ fn generate_random_bytes() -> [u8; 32] {
...
+        timestamp: chrono::Utc::now().timestamp(),
+    };
+
+    let serialized_message =
+        serde_json::to_string(&node_challenge_message).map_err(|_| ApiError {
```
miraclx: Looks like you don't need a `String`: use `to_vec` then, although this is negligible

crates/server/src/admin/service.rs
```diff
@@ -43,17 +43,16 @@ pub(crate) struct ServiceState {
...
+    pub service: ServiceState,
     pub store: Store,
     pub keypair: Keypair,
 }
 
 pub(crate) fn setup(
     config: &crate::config::ServerConfig,
     store: Store,
-) -> eyre::Result<Option<(&'static str, Router)>> {
+) -> eyre::Result<Option<(&'static str, Router<Arc<AdminState>>)>> {
```
miraclx: does rust complain without it?

crates/server/src/admin/service.rs
```diff
@@ -52,7 +52,7 @@ pub struct AdminState {
 pub(crate) fn setup(
     config: &crate::config::ServerConfig,
     store: Store,
-) -> eyre::Result<Option<(&'static str, Router<Arc<AdminState>>)>> {
+) -> eyre::Result<Option<(&'static str, Router)>> {
```
miraclx, line 55: ah it's been removed, 👍🏽

PR #157, opened 2024-04-15
feat: add client key signature and enable auth in jsonrpc

MatejVukosav:
- store client key on login
- use client key to verify signature - currently only in json rpc
- update calimeroSdk example to test newer version
- fix admin usage of challenge to add root key for testing client keys

### DIFF
--- a/crates/server/src/middleware/auth.rs
+++ b/crates/server/src/middleware/auth.rs
@@ -88,7 +88,7 @@ pub fn auth(headers: &HeaderMap, store: &Store) -> Result<(), UnauthorizedError<
     })?;
 
     let client_key = ClientKey {
-        wallet_type: auth_headers.wallet_type.clone(),
+        wallet_type: auth_headers.wallet_type,
         signing_key: auth_headers.signing_key.clone(),
     };
     let key_exists = exists_client_key(store, &client_key)


crates/identity/src/auth.rs
```diff
@@ -1,10 +1,17 @@
...
 use web3::signing::{keccak256, recover};
 
 pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> bool {
     keypair.public().verify(msg, signature)
 }
 
+pub fn verify_client_key(client_key: &str, msg: &[u8], signature: &[u8]) -> bool {
+    let client_key = bs58::decode(client_key).into_vec().unwrap();
+    let public_key =
+        PublicKey::try_decode_protobuf(&client_key).expect("Public key creation failed");
```
miraclx: return a result, or convert to bool, but don't panic

crates/server/src/admin/handlers/add_client_key.rs
```diff
@@ -51,12 +52,22 @@ struct WalletMetadata {
     signing_key: String,
 }
 
-#[derive(Debug, Deserialize, PartialEq)]
-enum WalletType {
+#[derive(Debug, Deserialize, PartialEq, Serialize, Clone)]
```
miraclx: Can impl `Copy` so you don't have to be explicit about `.clone()`

crates/server/src/middleware/auth.rs
```diff
@@ -71,31 +75,58 @@ where
...
-        Ok(_) => Err(UnauthorizedError::new("Keypair not matching signature.")),
-        Err(error) => Err(error),
+pub fn auth<'a>(headers: &'a HeaderMap, store: &'a Store) -> Result<(), UnauthorizedError<'a>> {
+    let auth_headers = get_auth_headers(headers).map_err(|e| {
+        debug!("Failed to extract authentication headers: {}", e);
+        UnauthorizedError::new("Failed to extract authentication headers")
+    })?;
+
+    let client_key = ClientKey {
+        wallet_type: auth_headers.wallet_type.clone(),
```
miraclx: like here

```suggestion
        wallet_type: auth_headers.wallet_type,
```

crates/server/src/admin/storage/client_keys.rs
```diff
@@ -0,0 +1,56 @@
...
+use super::did::{get_or_create_did, update_did};
+use crate::admin::handlers::add_client_key::WalletType;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub(crate) struct ClientKey {
+    pub(crate) wallet_type: WalletType,
+    pub(crate) signing_key: String,
+}
+
+pub fn add_client_key(store: &Store, client_key: ClientKey) -> eyre::Result<bool> {
```
miraclx, line 13: These APIs are a bit weird, imo.. I mean, it works, but it's an antipattern.

Ideally, all DID features should be implemented on a `Did` type that wraps the store for it's backing functionality

This way you have to know these stand-alone methods exist and remember to do the right thing

We can refactor post-token

crates/server/src/admin/storage/client_keys.rs
```diff
@@ -0,0 +1,56 @@
...
+        .iter()
+        .any(|k| k.signing_key == client_key.signing_key)
+    {
+        did_document.client_keys.push(client_key);
+        update_did(store, did_document)?;
+    }
+    Ok(true)
+}
+
+pub fn get_client_key(store: &Store, client_key: &ClientKey) -> eyre::Result<Option<ClientKey>> {
```
miraclx: especially this.. you need a client key to get back a client key

crates/server/src/admin/storage/client_keys.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub fn remove_client_key(store: &Store, client_key: &ClientKey) -> eyre::Result<()> {
+    let mut did_document = get_or_create_did(store)?;
+
+    if let Some(pos) = did_document
+        .client_keys
+        .iter()
+        .position(|x| x.signing_key == client_key.signing_key)
+    {
+        did_document.client_keys.remove(pos);
```
miraclx, line 52: forgot to write

crates/server/src/admin/storage/did.rs
```diff
@@ -21,6 +23,7 @@ pub fn create_did(store: &Store) -> eyre::Result<Did> {
     let did_document = Did {
         id: DID_KEY.to_string(),
         root_keys: Vec::<RootKey>::new(),
+        client_keys: Vec::<ClientKey>::new(),
```
miraclx: no need for this verbose syntax: `vec![]` or `Vec::new()` should suffice

crates/server/src/middleware/auth.rs
```diff
@@ -71,31 +75,58 @@ where
...
-                keypair,
-                auth_headers.challenge.as_slice(),
-                auth_headers.signature.as_slice(),
-            ) =>
-        {
-            Ok(())
-        }
-        Ok(_) => Err(UnauthorizedError::new("Keypair not matching signature.")),
-        Err(error) => Err(error),
+pub fn auth<'a>(headers: &'a HeaderMap, store: &'a Store) -> Result<(), UnauthorizedError<'a>> {
```
miraclx: looks like this can be static

```suggestion
pub fn auth(headers: &HeaderMap, store: &Store) -> Result<(), UnauthorizedError<'static>> {
```

or just let rustc infer it, like you're already doing with `get_auth_headers`:

```suggestion
pub fn auth(headers: &HeaderMap, store: &Store) -> Result<(), UnauthorizedError> {
```

crates/server/src/middleware/auth.rs
```diff
@@ -71,31 +75,58 @@ where
...
+        Ok(())
+    } else {
+        Err(UnauthorizedError::new(
+            "Invalid signature for provided key.",
+        ))
     }
 }
 
 fn get_auth_headers(headers: &HeaderMap) -> Result<AuthHeaders, UnauthorizedError> {
+    print!("Headers: {:?}", headers);
```
miraclx: don't forget this

crates/identity/src/auth.rs
```diff
@@ -1,10 +1,17 @@
-use libp2p::identity::Keypair;
+use libp2p::identity::{Keypair, PublicKey};
 use web3::signing::{keccak256, recover};
 
 pub fn verify_peer_auth(keypair: &Keypair, msg: &[u8], signature: &[u8]) -> bool {
     keypair.public().verify(msg, signature)
 }
 
+pub fn verify_client_key(client_key: &str, msg: &[u8], signature: &[u8]) -> bool {
+    let client_key = bs58::decode(client_key).into_vec().unwrap();
```
fbozic: Can we not use `unwrap` here?

crates/identity/src/auth.rs
fbozic: Also this `except` might panic, would be nice to avoid it unless you are 100% sure it won't panic.

crates/server/src/admin/handlers/add_client_key.rs
```diff
@@ -51,12 +52,22 @@ struct WalletMetadata {
...
+pub enum WalletType {
     NEAR,
     ETH,
 }
 
+impl WalletType {
+    pub fn from_str(input: &str) -> Result<Self, eyre::Report> {
```
fbozic: Why not use `eyre::Result<Self>` instead of `Result<Self, eyre::Report>`?

crates/server/src/middleware/auth.rs
fbozic: Leftovers from debug?

examples/only-peers-simple/package.json
```diff
@@ -10,10 +10,11 @@
     "redev": "rm -rf node_modules && pnpm install && vite"
   },
   "dependencies": {
+    "@calimero-is-near/calimero-p2p-sdk": "0.0.11",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^13.5.0",
-    "@calimero-is-near/calimero-p2p-sdk": "0.0.6",
+    "axios": "^1.6.7",
```
fbozic: Why add axios? It would make more sense to update client-sdk not to expose anything related to axios towards end user (because at the end of the day we might switch from axios to something else and it shouldn't affect users)

MatejVukosav: Leftover. I have update it at the end but forgot to remove it here

crates/server/src/admin/storage/client_keys.rs
MatejVukosav: It should be just key id

crates/server/src/admin/storage/client_keys.rs
MatejVukosav: Sure, we can encapsulate the store inside the did and provide traits for it to simplify. I am still not sure if client keys should be under did as they are more related to session rather than identity itself.  This is just first iteration which we can improve later.

crates/server/src/middleware/auth.rs
MatejVukosav: I can implement Copy but auth_headers is used after and without clone, move occurs.

miraclx: if you impl `Copy`, a move shouldn't occur

crates/server/src/admin/storage/client_keys.rs
miraclx: in that case, you can use a trait for specialization "on" the Did. But we'll revise it later.

crates/server/src/middleware/auth.rs
MatejVukosav:
```
#[derive(Debug, Deserialize, PartialEq, Serialize, Clone, Copy)]
pub enum WalletType {
    NEAR,
    ETH,
}
```

 auth_headers.signing_key.as_str(),
    |         ^^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after move

AuthHeaders cannot impl Copy.

@miraclx

MatejVukosav: Ok I thing I catch up with your thoughts. I have removed the clone from the wallet type only.

PR #167, opened 2024-04-18
fix(admin-ui): upload app ui

frdomovic:
# fix(admin-ui): upload app ui

### DIFF
--- a/node-ui/src/components/applications/ApplicationsTable.jsx
+++ b/node-ui/src/components/applications/ApplicationsTable.jsx
@@ -98,7 +98,7 @@ export function ApplicationsTable({ applications, install, uninstall }) {
 
   return (
     <Table>
-      {applications && applications.length === 0 ? (
+      {!applications || applications.length === 0 ? (
         <div className="header">
           <p className="item-no-apps">{t.noApplicationsText}</p>
         </div>


node-ui/src/components/applications/ApplicationsTable.jsx
```diff
@@ -92,7 +98,11 @@ export function ApplicationsTable({ applications, install, uninstall }) {
 
   return (
     <Table>
-      {applications && (
+      {applications.length === 0 ? (
```
MatejVukosav: Are you sure that apps always exists?

node-ui/src/constants/en.global.json
```diff
@@ -19,7 +19,8 @@
             "headerDescText": "Description",
             "headerRepoText": "Repository",
             "uninstallButtonText": "Uninstall",
-            "installButtonText": "Install new application"
+            "installButtonText": "Install new application",
+            "noApplicationsText": "No applicatons found in the node."
```
MatejVukosav: Applications

MatejVukosav: No applications installed in the node.

node-ui/src/constants/en.global.json
```diff
@@ -65,10 +67,14 @@
     },
     "uploadPage": {
         "title": "Upload Application",
-        "walletButtonText": "Add wallet account"
+        "walletButtonText": "Add wallet account",
```
MatejVukosav, line 70: This should be redefined imo. It is not clear why user needs to add new wallet.

node-ui/src/constants/en.global.json
```diff
@@ -88,6 +94,7 @@
     },
     "addPackageForm": {
         "title": "Package Details",
+        "subtitle": "To successfully install the application, first you need to add the package details and procceed to add relase.",
```
MatejVukosav, line 97: We need better copy here @matejpribanic

node-ui/src/components/applications/ApplicationsTable.jsx
frdomovic: updated

node-ui/src/constants/en.global.json
frdomovic: can i merge this and next comment and we fix it next week?

MatejVukosav: Sure, but should be properly redesigned with @matejpribanic inputs. 

Can you open a task for this and lead it with @matejpribanic ?

node-ui/src/components/applications/ApplicationsTable.jsx
```diff
@@ -92,7 +98,11 @@ export function ApplicationsTable({ applications, install, uninstall }) {
 
   return (
     <Table>
-      {applications && (
+      {applications && applications.length === 0 ? (
```
MatejVukosav: Shouldn't this rather be `if(!applications || applications.length===0)`  so you don't need to check later if applications exist

node-ui/src/constants/en.global.json
frdomovic: https://calimero.atlassian.net/jira/software/projects/C20/boards/28?selectedIssue=C20-53

PR #171, opened 2024-04-22
fix(admin-ui + admin-api) : listing applications

frdomovic:
# fix(admin-ui + admin-api) : listing applications
 
## Summary:
First issue was with api returning "Failed to read application directory" when no applications were installed as the "apps" folder did not exist. Now there is a check for "apps" folder existence and it returns empty hash map if no applications are installed yet 

The check "response?.apps?.length" was not working as there is no length for KV object in javascript without using Object.keys. Changing the response to have different key like "apps2" (img2) throws no errors and just returns empty array.

### DIFF
--- a/node-ui/src/pages/Applications.jsx
+++ b/node-ui/src/pages/Applications.jsx
@@ -45,7 +45,7 @@ export default function Applications() {
       }
     };
     setApps();
-  }, []);
+  }, [swithInstall]);
 
   return (
     <FlexLayout>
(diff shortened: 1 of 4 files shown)


node-ui/src/pages/Applications.jsx
```diff
@@ -45,7 +45,7 @@ export default function Applications() {
       }
     };
     setApps();
-  }, []);
+  }, [swithInstall]);
```
MatejVukosav, line 48: what is `swithInstall`?

frdomovic: Its the application page content switch, so it fetches apps when you go to install new app and back to app list. When having [] deps it wont give you newest application installed as the page doesent reload after installing. (In this approach, will be updated when i add popup modals in next PR's)

PR #173, opened 2024-04-22
feat(admin-ui): install applications status handler

frdomovic:
# feat(admin-ui): install applications status handler
 
## Summary:
Added popup status modal for installing applications. It can show success status and error status. Success status on continue navigates you to applications list and resets the state, while error status closes the modal and leaves you on the select package and release page to try again.

Fixes # [C20-54](https://calimero.atlassian.net/browse/C20-54)

### DIFF
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -356,7 +356,7 @@ async fn fetch_application_handler(
     Extension(state): Extension<Arc<AdminState>>,
 ) -> impl IntoResponse {
 
-    if !&state.service.application_dir.exists() {
+    if !state.service.application_dir.exists() {
         return ApiResponse {
             payload: ApplicationListResult { apps: Vec::new() },
         }


node-ui/src/hooks/useAdminClient.js
```diff
@@ -2,11 +2,28 @@ import axios from "axios";
...
+    try {
+      const response = await axios.post("/admin-api/install-application", {
+        application: selectedPackage,
+        version: selectedVersion,
+      });
+      if (response.status === 200) {
+        return { data: response.data };
+      } else {
+        return {
+          error: { code: 500, message: "Failed to install application" },
```
MatejVukosav: Use error returned from server

node-ui/src/hooks/useAdminClient.js
```diff
@@ -2,11 +2,28 @@ import axios from "axios";
...
+        };
+      }
+    } catch (error) {
+      if (error.response) {
```
MatejVukosav, line 12: Can silently ignore error if `error.response `does not exist.

node-ui/src/pages/Applications.jsx
```diff
@@ -45,29 +53,64 @@ export default function Applications() {
...
+            setShowInstallApplications={setShowInstallApplications}
+            showStatusModal={showStatusModal}
+            closeModal={closeStatusModal}
+            installationStatus={installationStatus}
           />
         ) : (
           <ApplicationsTable
             applications={applications}
-            install={() => setSwitchInstall(true)}
+            install={() => setShowInstallApplications(true)}
```
MatejVukosav: I think a different approach is needed here. As this is in the form of tabs, the tabs work on a base where a single item is selected.  To enable tabs to grow it is pretty messy to maintain semaphore states with on and off. A better way is to use a selected item where you define what should be visible in the tab and not which item is on and off. Saying that remove `setShowInstallApplications` and define an enum with possibilities and iterate over options.

frdomovic: Updated with JS version of enum

node-ui/src/hooks/useAdminClient.js
frdomovic: updated

node-ui/src/hooks/useAdminClient.js
frdomovic: ?

MatejVukosav: If an error does not have a response field, an error will not be shown to the user right? We should have a generic error in that case indicating that something is wrong and we don't have more details about the problem so the user knows that his action has not finished successfully.

frdomovic: Updated , will be displayed as:
title: Failed to install application
sub: Try again later.

PR #174, opened 2024-04-22
fix(admin-api): additional brackets

frdomovic:
# fix(admin-api): additional brackets


crates/server/src/admin/service.rs
```diff
@@ -351,7 +351,7 @@ async fn fetch_application_handler(
     Extension(state): Extension<Arc<AdminState>>,
 ) -> impl IntoResponse {
 
-    if(!&state.service.application_dir.exists()){
+    if !&state.service.application_dir.exists() {
```
miraclx: shouldn't be necessary, I believe

```suggestion
    if !state.service.application_dir.exists() {
```

PR #175, opened 2024-04-22
feat(admin-ui): add version and id to applications list

frdomovic:
# feat(admin-ui): add version and id to applications list
 
## Summary:
Add application id and version to apps table list with option to copy id.

Fixes # [C20-55](https://calimero.atlassian.net/browse/C20-55)

### DIFF
--- a/node-ui/src/api/dataSource/AppsDataSource.js
+++ b/node-ui/src/api/dataSource/AppsDataSource.js
@@ -11,11 +11,11 @@ export class AppsDataSource {
       if (response?.apps) {
         return response.apps;
       } else {
-        return {};
+        return [];
       }
     } catch (error) {
       console.error("Error fetching installed applications:", error);
-      return {};
+      return [];
     }
   }
 }
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::error::Error;
 use std::fmt::{self, Display, Formatter};
 use std::fs::{self, File};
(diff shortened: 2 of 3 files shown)


node-ui/src/api/dataSource/AppsDataSource.js
```diff
@@ -9,12 +9,13 @@ export class AppsDataSource {
     try {
       const response = await this.client.get("/admin-api/applications");
       if (response?.apps) {
-        return Object.keys(response.apps);
+        return response.apps;
       } else {
-        return [];
+        return {};
```
MatejVukosav: shouldn't this be an empty array?

node-ui/src/constants/en.global.json
```diff
@@ -17,6 +17,8 @@
             "title": "Applications",
             "headerNameText": "Name",
             "headerDescText": "Description",
+            "headerVersionText": "Version",
```
MatejVukosav, line 20: Maybe latest version as you can have more versions

frdomovic: admin api only returns latest version installed so there will never be more than 1 version of the same application in the list

node-ui/src/api/dataSource/AppsDataSource.js
frdomovic: previous implementation where i only used applications id array ( Object.keys(x)) . now since i need version as well which is value of the json object i return empty json {}. Later it checks  Object.keys(response).length so its covered the right way and using array here would throw error.
The object returned from api looks like this 
`
{
"hash1" : "1.0.0",
"hash2" : "1.0.1",
"hash3": "1.4.3"
}
`
And its not array but parsed JSON version of hashmap

node-ui/src/constants/en.global.json
MatejVukosav: Ok, forgot that this is the list of installed applications only.

node-ui/src/api/dataSource/AppsDataSource.js
MatejVukosav: Huh, why are we using object for a list? :/ API should return list of applications where each app in this example have only hash value.

frdomovic: Updated api to return list to looks like 
`[{ id: "hash1", version: "1.0.0"}, {...}, {...}]`

PR #210, opened 2024-04-23
feat(admin-ui): error handling for package - release upload

frdomovic:
# feat(admin-ui): error handling for package - release upload
 
## Summary:
Error handling for adding new package and release to package. The error at the end "package does not exist" is because contract error handling needs to be updated.  It takes the id as sha256(name+signer) but it should check if the signer is owner

Fixes # C20-56

### DIFF
--- a/node-ui/src/pages/UploadApp.jsx
+++ b/node-ui/src/pages/UploadApp.jsx
@@ -157,7 +157,7 @@ export default function UploadApp() {
           ?.ExecutionError ?? "An error occurred while adding the package";
 
       setDeployStatus({
-        title: "Failed to add Package",
+        title: "Failed to add package",
         message: errorMessage,
         error: true,
       });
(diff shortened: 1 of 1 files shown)


node-ui/src/pages/UploadApp.jsx
```diff
@@ -15,28 +15,69 @@ import * as nearAPI from "near-api-js";
...
+    fetchWalletAccounts();
+  }, []);
 
   const addWalletAccount = async () => {
     const selector = await setupWalletSelector({
       network: import.meta.env.VITE_NEAR_ENVIRONMENT ?? "testnet",
       modules: [setupMyNearWallet()],
     });
     const wallet = await selector.wallet("my-near-wallet");
+    await wallet.signOut();
```
frdomovic, line 80: This wont work if the user is signed in already thats why this signOut is added. Only 1 account can be "active" "logged in" with this approach

MatejVukosav: There is an option to switch wallet if you are already logged in

Issue #209, opened 2024-04-23
Failed to send transaction: No connected peers to send message to.


MatejVukosav:
Application is installed through admin ui to node 1. Then copied to node2 and the coordinator as networks are currently live before app installation.
Coordinator, node 1, and node 2 are up and running with the same app.
Open only-peers frontend and create a new post with node 1.  Error appears.

```
2024-04-23T11:34:13.947392Z ERROR calimero_node: Failed to send transaction: No connected peers to send message to.
2024-04-23T11:34:13.947500Z ERROR calimero_server::jsonrpc: Failed to execute JSON RPC method err=HandlerError
```

The post should be submitted as node 1 has two active peers.

PR #229, opened 2024-04-25
fix: fix topic subscription on app install and reboot of node

fbozic:
Closes: #209

### DIFF
--- a/crates/application/src/lib.rs
+++ b/crates/application/src/lib.rs
@@ -40,7 +40,7 @@ impl ApplicationManager {
     pub async fn install_application(
         &self,
         application_id: calimero_primitives::application::ApplicationId,
-        version: &str,
+        version: &semver::Version,
     ) -> eyre::Result<()> {
         let release = self.get_release(&application_id, version).await?;
         self.download_release(&application_id, &release, &self.application_dir)
@@ -51,7 +51,7 @@ impl ApplicationManager {
             .subscribe(calimero_network::types::IdentTopic::new(application_id))
             .await?;
 
-        info!(%topic_hash, "Subscribed to application topic");
+        info!(%topic_hash, "Subscribed to network topic");
         return Ok(());
     }
 
@@ -71,7 +71,7 @@ impl ApplicationManager {
                     if let Some((version, _)) = self.get_latest_application_info(&application_id) {
                         applications.push(calimero_primitives::application::Application {
                             id: application_id,
-                            version: version.to_string(),
+                            version,
                         });
                     }
                 }
@@ -127,7 +127,7 @@ impl ApplicationManager {
     async fn get_release(
         &self,
         application_id: &calimero_primitives::application::ApplicationId,
-        version: &str,
+        version: &semver::Version,
     ) -> eyre::Result<calimero_primitives::application::Release> {
         let client = JsonRpcClient::connect("https://rpc.testnet.near.org");
         let request = methods::query::RpcQueryRequest {
@@ -138,7 +138,7 @@ impl ApplicationManager {
                 args: FunctionArgs::from(
                     serde_json::json!({
                         "id": application_id,
-                        "version": version
+                        "version": version.to_string()
                     })
                     .to_string()
                     .into_bytes(),
--- a/crates/primitives/src/application.rs
+++ b/crates/primitives/src/application.rs
@@ -32,17 +32,12 @@ impl AsRef<str> for ApplicationId {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Application {
     pub id: ApplicationId,
-    pub version: String,
-}
-
-#[derive(Clone, Serialize, Deserialize, Debug)]
-pub struct ApplicationListResult {
-    pub apps: Vec<Application>,
+    pub version: semver::Version,
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Release {
-    pub version: String,
+    pub version: semver::Version,
     pub notes: String,
     pub path: String,
     pub hash: String,
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -15,7 +15,7 @@ use tower_http::set_status::SetStatus;
 use tower_sessions::{MemoryStore, Session, SessionManagerLayer};
 use tracing::{error, info};
 
-use super::handlers::add_client_key::{add_client_key_handler, parse_api_error};
+use super::handlers::add_client_key::add_client_key_handler;
 use super::handlers::challenge::{request_challenge_handler, NodeChallenge, CHALLENGE_KEY};
 use super::handlers::fetch_did::fetch_did_handler;
 use super::storage::root_key::{add_root_key, RootKey};
@@ -191,18 +191,16 @@ async fn create_root_key_handler(
 
 async fn install_application_handler(
     Extension(state): Extension<Arc<AdminState>>,
-    Json(req): Json<InstallApplicationRequest>,
+    Json(req): Json<calimero_server_primitives::admin::InstallApplicationRequest>,
 ) -> impl IntoResponse {
-    let result = state
+    match state
         .application_manager
-        .install_application(req.application.into(), &req.version)
-        .await;
-
-    Ok(match result {
-        Ok(()) => (StatusCode::OK, "Application Installed"),
-        Err(err) => return Err(parse_api_error(err)),
+        .install_application(req.application, &req.version)
+        .await
+    {
+        Ok(()) => ApiResponse { payload: () }.into_response(),
+        Err(err) => (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response(),
     }
-    .into_response())
 }
 
 async fn list_applications_handler(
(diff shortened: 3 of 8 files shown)


crates/application/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+pub struct ApplicationManager {
+    pub network_client: NetworkClient,
+    pub application_dir: Utf8PathBuf,
+}
+
+pub async fn start_manager(
+    config: &config::ApplicationConfig,
+    network_client: NetworkClient,
+) -> eyre::Result<ApplicationManager> {
+    let application_manager = ApplicationManager::new(network_client.clone(), config.dir.clone());
```
miraclx:
```suggestion
    let application_manager = ApplicationManager::new(network_client, config.dir.clone());
```

crates/application/src/lib.rs
```diff
@@ -0,0 +1,234 @@
+use std::fs::{self, File};
```
miraclx, line 1: This was a lingering problem before this patch, so the fix doesn't have to happen here. Unless you have the time to look into it. But we need to avoid these kinds of blocking fs ops and migrate to the non-blocking alternatives provided by tokio:

https://docs.rs/tokio/latest/tokio/fs/index.html

But like I said, doesn't have to happen right away. It's just one more thing in the backlog.

crates/primitives/src/application.rs
```diff
@@ -28,3 +28,22 @@ impl AsRef<str> for ApplicationId {
         &self.0
     }
 }
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct Application {
+    pub id: ApplicationId,
+    pub version: String,
```
miraclx: make this a `semver::Version` forcing its validity across all `Application` instances

crates/primitives/src/application.rs
```diff
@@ -28,3 +28,22 @@ impl AsRef<str> for ApplicationId {
...
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct ApplicationListResult {
+    pub apps: Vec<Application>,
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct Release {
+    pub version: String,
```
miraclx: Seim 🙃

crates/application/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    pub async fn list_installed_applications(
+        &self,
+    ) -> eyre::Result<Vec<calimero_primitives::application::Application>> {
+        if !self.application_dir.exists() {
+            return Ok(Vec::new());
+        }
+
+        if let Ok(entries) = fs::read_dir(&self.application_dir) {
+            let mut applications: Vec<calimero_primitives::application::Application> = Vec::new();
```
miraclx: type inference go brrr

```suggestion
            let mut applications = Vec::new();
```

crates/application/src/lib.rs
```diff
@@ -39,18 +39,16 @@ impl ApplicationManager {
...
+        let release = self.get_release(&application_id, version).await?;
+        self.download_release(&application_id, &release, &self.application_dir)
             .await?;
 
         let topic_hash = self
             .network_client
-            .subscribe(calimero_network::types::IdentTopic::new(
-                application_id.as_ref(),
-            ))
+            .subscribe(calimero_network::types::IdentTopic::new(application_id))
```
miraclx, line 51: looks like this can maybe also take a reference?

crates/application/src/lib.rs
```diff
@@ -39,18 +39,16 @@ impl ApplicationManager {
 
     pub async fn install_application(
         &self,
-        application_id: &calimero_primitives::application::ApplicationId,
+        application_id: calimero_primitives::application::ApplicationId,
```
miraclx, line 42: so we can relax this requirement to a ref

crates/server/src/admin/service.rs
```diff
@@ -195,7 +195,7 @@ async fn install_application_handler(
 ) -> impl IntoResponse {
     let result = state
         .application_maneger
-        .install_application(&(req.application.into()), &req.version)
+        .install_application(req.application.into(), &req.version)
```
miraclx: let's derive `Deserialize` on `ApplicationId` at this point

crates/node/src/lib.rs
```diff
@@ -498,10 +511,14 @@ impl Node {
             ),
         };
 
+        info!(%application_id, %method, "Executing method");
+        let short_application_id = application_id.as_ref().split('/').last().unwrap();
+        info!(%short_application_id, "Executing method");
```
miraclx: tolerable for now, but unideal. we'll revisit

crates/server/src/admin/service.rs
```diff
@@ -64,21 +52,19 @@ pub(crate) fn setup(
     let session_layer = SessionManagerLayer::new(session_store);
 
     let shared_state = Arc::new(AdminState {
-        service: ServiceState {
-            application_dir: admin_config.application_dir.clone(),
-        },
         store,
         keypair: config.identity.clone(),
+        application_maneger: application_manager,
```
miraclx: is it actually `maneger`? 😅

crates/server/src/lib.rs
```diff
@@ -80,7 +81,7 @@ pub async fn start(
     {
         if let Some((path, handler)) = jsonrpc::service(&config, server_sender.clone())? {
             app = app.route(path, handler);
-            app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
+            // app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
```
miraclx: don't forget this 🙃

crates/application/src/lib.rs
chefsale: Issue a ticket.

crates/node/src/lib.rs
fbozic: leftover, sorry, will clean it up

crates/application/src/lib.rs
fbozic: good point :)

crates/server/src/admin/service.rs
fbozic: 🤦🏼

crates/application/src/lib.rs
fbozic: I'm not 100% sure, but I can see that all client methods take ownership of the topic (and they pass it to the sender). Can we pass a reference in the message?
image

crates/server/src/admin/service.rs
fbozic: we already derive `Deserialize`
image

crates/primitives/src/application.rs
fbozic: I don't think this is going to be trivial, just changing `String` to `semver:Version` cause this
```
   |
45 |     pub version: semver::Version,
   |     ^^^ the trait `Deserialize<'_>` is not implemented for `semver::Version`
```
`Release` struct is used to deserialize contract output as well, we need to make sure that we don't break that part.  

Let's create a ticket to clean up the entire admin server.

miraclx: `semver::Version` impls `Serialize` & `Deserialize` with the `serde` feature flag activated.

https://github.com/dtolnay/semver/blob/c520af7e3cdd4c949b40dc20c945b525f34a2dc5/Cargo.toml#L18-L19

So you can modify primitives's Cargo.toml to include the feature flag `"serde"`

miraclx: > Release struct is used to deserialize contract output as well, we need to make sure that we don't break that part.

It should be consistent there too, but non-blocking for this patch at least

crates/application/src/lib.rs
miraclx: Yeah, I was working with the previous logic, taking `AsRef<str>` and upcasting it to a `String` via `AsRef<str>: Into<String>` which is the only requirement for [`IdentTopic::new`](https://docs.rs/libp2p/latest/libp2p/gossipsub/struct.Topic.html#tymethod.new)

But this would still do an allocation anyway, so it's perfectly fine as is.

crates/application/src/lib.rs
miraclx: resolved above

crates/server/src/admin/service.rs
miraclx: In that case, you should be able to use it directly in `InstallApplicationRequest`

crates/application/src/lib.rs
miraclx: Extracted out into https://github.com/calimero-network/core/issues/234

crates/primitives/src/application.rs
fbozic: thank you, that helps. will keep in mind to check feature flags of libs we use in future.

crates/application/src/lib.rs
fbozic: oki, keeping it as it is

crates/server/src/admin/service.rs
fbozic: good point, done

PR #253, opened 2024-05-06
feat: add context crd

MatejVukosav:
Endpoint for creating, reading and deleting context.

**### Important
Delete all nodes storage and initialize it again**

Endpoints
```
POST http://localhost:2428/admin-api/contexts

{
    "applicationId": "5d77b7852d5c2fa9483ece70ada7e3cc7d2a9df5aa056d3916103dbe783df6e2"
}

{
    "data": {
        "id": "bjXWsETACza9PsejKCgRr5ZZvUndaJDSKMDCUR66zRw",
        "signingKey": {
            "signingKey": "BFBMA2pZH6aKMJKPnXYSJwHVpSYZpw5jiLywGyppCYzn"
        },
        "applicationId": "5d77b7852d5c2fa9483ece70ada7e3cc7d2a9df5aa056d3916103dbe783df6e2"
    }
}
```

```
GET http://localhost:2428/admin-api/contexts/bjXWsETACza9PsejKCgRr5ZZvUndaJDSKMDCUR66zRw
{
    "data": {
        "id": "bjXWsETACza9PsejKCgRr5ZZvUndaJDSKMDCUR66zRw",
        "signingKey": {
            "signingKey": "BFBMA2pZH6aKMJKPnXYSJwHVpSYZpw5jiLywGyppCYzn"
        },
        "applicationId": "5d77b7852d5c2fa9483ece70ada7e3cc7d2a9df5aa056d3916103dbe783df6e2"
    }
}
```

```
GET http://localhost:2428/admin-api/contexts
{
    "data": [
        {
        "id": "bjXWsETACza9PsejKCgRr5ZZvUndaJDSKMDCUR66zRw",
        "signingKey": {
            "signingKey": "BFBMA2pZH6aKMJKPnXYSJwHVpSYZpw5jiLywGyppCYzn"
        },
        "applicationId": "5d77b7852d5c2fa9483ece70ada7e3cc7d2a9df5aa056d3916103dbe783df6e2"
    }
    ]
}
```

```
DELETE http://localhost:2428/admin-api/contexts/bla
{
    "data": true
}
```

Refactor
- move identity types to primitives
- rename PeerID to peer_id and PrivKey to keypair

### DIFF
--- a/crates/primitives/src/identity.rs
+++ b/crates/primitives/src/identity.rs
@@ -134,7 +134,7 @@ pub mod serde_identity {
         let mut keypair = serializer.serialize_map(Some(2))?;
         keypair.serialize_entry("peer_id", &key.public().to_peer_id().to_base58())?;
         keypair.serialize_entry(
-            "private_key",
+            "keypair",
             &bs58::encode(&key.to_protobuf_encoding().map_err(ser::Error::custom)?).into_string(),
         )?;
         keypair.end()
@@ -163,7 +163,7 @@ pub mod serde_identity {
                 while let Some(key) = map.next_key::<String>()? {
                     match key.as_str() {
                         "peer_id" => peer_id = Some(map.next_value()?),
-                        "private_key" => priv_key = Some(map.next_value()?),
+                        "keypair" => priv_key = Some(map.next_value()?),
                         _ => {
                             let _ = map.next_value::<de::IgnoredAny>();
                         }
@@ -171,7 +171,7 @@ pub mod serde_identity {
                 }
 
                 let peer_id = peer_id.ok_or_else(|| de::Error::missing_field("peer_id"))?;
-                let priv_key = priv_key.ok_or_else(|| de::Error::missing_field("private_key"))?;
+                let priv_key = priv_key.ok_or_else(|| de::Error::missing_field("keypair"))?;
 
                 let priv_key = bs58::decode(priv_key)
                     .into_vec()
@@ -188,6 +188,6 @@ pub mod serde_identity {
             }
         }
 
-        deserializer.deserialize_struct("Keypair", &["peer_id", "private_key"], IdentityVisitor)
+        deserializer.deserialize_struct("Keypair", &["peer_id", "keypair"], IdentityVisitor)
     }
 }


crates/server/Cargo.toml
```diff
@@ -28,6 +28,11 @@ tower.workspace = true
 tower-http = { workspace = true, features = ["cors", "fs"] }
 tower-sessions = { workspace = true, optional = true }
 tracing.workspace = true
+uuid =  { workspace = true, features = [
+    "v4",                # Lets you generate random UUIDs
+    "fast-rng",          # Use a faster (but still sufficiently random) RNG
+    "macro-diagnostics", # Enable better diagnostics for compile-time UUIDs
+]}
```
miraclx: I'd like us to cryptographically enforce context exclusivity by using a public key derived from a cryptographic secret

The idea is - the cx author would generate a random seed as a signing key for the context discriminator, the public component of this key would be the context ID.

For now, this should suffice:

```rust
let mut context_id = [0; 32];
rand.thread_rng().fill_bytes(&mut context_id);
```

string representations of this can be encoded in bs58

crates/server/src/admin/handlers/context.rs
```diff
@@ -0,0 +1,113 @@
...
+    data: Context,
+}
+
+pub async fn create_context_handler(
+    Extension(state): Extension<Arc<AdminState>>,
+    Json(req): Json<CreateContextRequest>,
+) -> impl IntoResponse {
+    let context = Context {
+        id: Uuid::new_v4().to_string(),
+        identity: libp2p::identity::Keypair::generate_ed25519(),
```
miraclx: not libp2p KeyPair, use `ed25519_dalek::SigningKey` instead

crates/primitives/src/identity.rs
```diff
@@ -67,6 +116,6 @@ pub mod serde_identity {
             }
         }
 
-        deserializer.deserialize_struct("Keypair", &["PeerID", "PrivKey"], IdentityVisitor)
+        deserializer.deserialize_struct("Keypair", &["peerId", "privateKey"], IdentityVisitor)
```
miraclx: `toml` conventions typically use `kebab-case`, `PascalCase` or `snake_case`

crates/primitives/src/identity.rs
```diff
@@ -1,3 +1,52 @@
...
+}
+
+#[derive(Debug, Deserialize, PartialEq, Serialize, Clone, Copy)]
+pub enum WalletType {
+    NEAR,
+    ETH,
+}
+
+impl WalletType {
+    pub fn from_str(input: &str) -> eyre::Result<Self> {
```
miraclx: Use https://doc.rust-lang.org/std/str/trait.FromStr.html with a custom error variant, so we don't have to introduce `eyre` into `calimero-primitives` just for this.

crates/server/src/admin/handlers/context.rs
MatejVukosav: updated. Suggest better naming if you have.

crates/primitives/src/identity.rs
```diff
@@ -60,9 +124,9 @@ pub mod serde_identity {
         S: Serializer,
     {
         let mut keypair = serializer.serialize_map(Some(2))?;
-        keypair.serialize_entry("peerId", &key.public().to_peer_id().to_base58())?;
+        keypair.serialize_entry("peer_id", &key.public().to_peer_id().to_base58())?;
         keypair.serialize_entry(
-            "privateKey",
+            "private_key",
```
miraclx: let's use `keypair` here

crates/server/src/middleware/auth.rs
```diff
@@ -1,4 +1,5 @@
 use std::convert::Infallible;
+use std::str::FromStr;
```
miraclx: you can use `"ETH".parse()?` or `"ETH".parse::<WalletType>()` if you want to be explicit.

In this file, the first should suffice, Rust should type-infer it just fine

crates/primitives/src/identity.rs
MatejVukosav: `keypair` for private key value?

crates/server/src/middleware/auth.rs
MatejVukosav: I am using WalletType::from_str(&wallet_type). I don't know the type yet

MatejVukosav: I can do this `wallet_type.parse::<WalletType>()` instead

crates/primitives/src/identity.rs
miraclx: it is a keypair

crates/server/src/middleware/auth.rs
miraclx: > I don't know the type yet

yeah it was an example

in this case `wallet_type.parse()` should suffice too since rust will infer the correct type

crates/primitives/src/identity.rs
MatejVukosav: `&bs58::encode(&key.to_protobuf_encoding().map_err(ser::Error::custom)?)` is encoded private key?

crates/server/src/middleware/auth.rs
MatejVukosav: It can't. I need to add explicit type

crates/primitives/src/identity.rs
miraclx: `key` is a keypair

https://github.com/calimero-network/core/blob/64f415fcf5b42eb16cee472b0fbed8b6fce069ac/crates/primitives/src/identity.rs#L130

Issue #254, opened 2024-05-06
Investigate channel closed reports


miraclx:
```
2024-04-23T11:34:13.947392Z ERROR calimero_node: Failed to send transaction: No connected peers to send message to.
2024-04-23T11:34:13.947500Z ERROR calimero_server::jsonrpc: Failed to execute JSON RPC method err=HandlerError(Object {"ExecutionError": Object {"message": String("channel closed")}})
```

_Originally posted by @MatejVukosav in https://github.com/calimero-network/core/issues/176#issuecomment-2072319367_

PR #258, opened 2024-05-08
fix: handle outcome sender errors in the node and JSONRPC server

fbozic:
Fixes: #254 

Implement error handling for the outcome sender in node and JSON RPC. 
Implement `node-primitives` crate with the `ServerSender` and `CallError` types. `CallError` should cover almost all error cases which can occur during `call`/`call_mut`.
Refactored error handling in the node a bit.

Example of output from JSONRPC server.
```
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "error": {
    "cause": {
      "cause": {
        "application_id": "7039141d6a4bfbee3daa019746b3a44d13267f97fbdd6c8872f68da533c65d93"
      },
      "name": "ApplicationNotInstalled"
    },
    "name": "CallError"
  }
}
```

Example of logs from node:
```
2024-05-08T09:35:19.767562Z ERROR calimero_node: Failed to send transaction: Application is not installed.
2024-05-08T09:35:19.767565Z ERROR calimero_server::jsonrpc: Failed to execute JSON RPC method err=HandlerError(Object {"cause": Object {"cause": Object {"application_id": String("7039141d6a4bfbee3daa019746b3a44d13267f97fbdd6c8872f68da533c65d93")}, "name": String("ApplicationNotInstalled")}, "name": String("CallError")})
```

We can maybe reduce node logs a bit. Let me know what you think.

## Summary by CodeRabbit

- **New Features**
  - Introduced new types and error handling mechanisms in node communication.

- **Refactor**
  - Updated various dependencies and type references to use `calimero-node-primitives` for improved modularity and maintainability.
  - Revised error handling and logging for better clarity and robustness.

- **Bug Fixes**
  - Enhanced error mapping in JSON-RPC calls to provide more specific error messages.

### DIFF
--- a/crates/node-primitives/src/lib.rs
+++ b/crates/node-primitives/src/lib.rs
@@ -40,9 +40,7 @@ pub enum QueryCallError {
     ApplicationNotInstalled {
         application_id: calimero_primitives::application::ApplicationId,
     },
-    ExecutionError {
-        message: String,
-    },
+    InternalError,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Error)]
@@ -56,11 +54,5 @@ pub enum MutateCallError {
         application_id: calimero_primitives::application::ApplicationId,
     },
     NoConnectedPeers,
-    ExecutionError {
-        message: String,
-    },
-    FailedToPushTransaction {
-        message: String,
-    },
     InternalError,
 }
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -459,11 +459,10 @@ impl Node {
 
         self.execute(application_id, None, method, payload)
             .await
-            .map_err(
-                |e| calimero_node_primitives::QueryCallError::ExecutionError {
-                    message: e.to_string(),
-                },
-            )
+            .map_err(|e| {
+                error!(%e,"Failed to execute query call.");
+                calimero_node_primitives::QueryCallError::InternalError
+            })
     }
 
     async fn call_mutate(
@@ -518,20 +517,16 @@ impl Node {
         };
 
         if let Err(err) = self
-            .push_action(
-                application_id.clone(),
-                types::PeerAction::Transaction(transaction),
-            )
+            .push_action(application_id, types::PeerAction::Transaction(transaction))
             .await
         {
-            let _ = self.tx_pool.remove(&tx_hash).expect(
-            "Failed to remove just inserted transaction from the pool. This is a bug and should be reported.",
-        );
-            return Err(
-                calimero_node_primitives::MutateCallError::FailedToPushTransaction {
-                    message: err.to_string(),
-                },
-            );
+            if self.tx_pool.remove(&tx_hash).is_none() {
+                error!("Failed to remove just inserted transaction from the pool. This is a bug and should be reported.");
+                return Err(calimero_node_primitives::MutateCallError::InternalError);
+            }
+
+            error!(%err, "Failed to push transaction over the network.");
+            return Err(calimero_node_primitives::MutateCallError::InternalError);
         }
 
         self.last_tx = tx_hash;
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -71,7 +71,7 @@ pub async fn start(
     {
         if let Some((path, handler)) = jsonrpc::service(&config, server_sender.clone())? {
             app = app.route(path, handler);
-            // app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
+            app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
 
             serviced = true;
         }


crates/node/src/lib.rs
```diff
@@ -149,50 +146,56 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
-                                    println!("{IND}   Logs:");
-
-                                    for log in outcome.logs {
-                                        println!("{IND}     > {}", log.cyan());
+                                            for log in outcome.logs {
+                                                println!("{IND}     > {}", log.cyan());
+                                            }
+                                        }
                                     }
+                                    Err(_) => todo!(),
```
fbozic: Need to update this, some nice format would be nice. Or we can drop this completely, there is even todo to drop and replace with JSONRPC server.

crates/node/src/lib.rs
```diff
@@ -404,32 +407,59 @@ impl Node {
...
+        outcome_sender: oneshot::Sender<
+            Result<calimero_runtime::logic::Outcome, calimero_node_primitives::CallError>,
+        >,
     ) -> eyre::Result<calimero_primitives::hash::Hash> {
         if self.typ.is_coordinator() {
+            let _ =
+                outcome_sender.send(Err(calimero_node_primitives::CallError::InvalidNodeType {
+                    node_type: self.typ,
+                }));
             eyre::bail!("Coordinator can not create transactions!");
```
fbozic: Not sure if this is the practice, but atm we are returning error to 2 different consumers.

crates/sdk/macros/src/event.rs
```diff
@@ -78,40 +78,50 @@ impl<'a> TryFrom<EventImplInput<'a>> for EventImpl<'a> {
         match vis {
             syn::Visibility::Public(_) => {}
             syn::Visibility::Inherited => {
-                return Err(errors.finish(syn::Error::new_spanned(ident, errors::ParseError::NoPrivateEvent)));
+                return Err(errors.finish(syn::Error::new_spanned(
+                    ident,
+                    errors::ParseError::NoPrivateEvent,
+                )));
```
fbozic: Ran `cargo +nightly fmt`

crates/server/Cargo.toml
```diff
@@ -30,15 +30,13 @@ tower-sessions = { workspace = true, optional = true }
...
+calimero-node-primitives = { path = "../node-primitives" }
 calimero-identity = { path = "../identity" }
 calimero-primitives = { path = "../primitives" }
 calimero-runtime = { path = "../runtime" }
 calimero-server-primitives = { path = "../server-primitives" }
 calimero-store = { path = "../store" }
 
-[[example]]
-name = "demo"
-
```
fbozic, line 41: It was broken so I've removed it

crates/node/src/lib.rs
miraclx: At this point, I'd distinguish between the server's channel and the internal one.

This function should return a structured error as a result, the caller (from the node) would then serialize that over the server's sender.

crates/sdk/macros/src/event.rs
miraclx: Thanks - https://github.com/calimero-network/core/pull/259

crates/server/src/lib.rs
```diff
@@ -20,19 +20,9 @@ pub mod ws;
...
     "/calimero/experimental/app/9SFTEoc6RBHtCn9b6cm4PPmhYzrogaMCd5CRiYAQichP";
 
-// TODO: add comments or even better make it explicit types
-type ServerSender = mpsc::Sender<(
-    // todo! move to calimero-node-primitives
-    calimero_primitives::application::ApplicationId,
-    String,
-    Vec<u8>,
-    bool,
-    oneshot::Sender<calimero_runtime::logic::Outcome>,
```
miraclx, line 30: `calimero_runtime` dep can be removed now

or at least moved to dev-deps since `demo` still depends on it

crates/server/src/jsonrpc.rs
```diff
@@ -138,29 +136,32 @@ impl<T: Serialize, E: Serialize> ToResponseBody for Result<T, RpcError<E>> {
...
 pub(crate) async fn call(
-    sender: crate::ServerSender,
+    sender: calimero_node_primitives::ServerSender,
     application_id: calimero_primitives::application::ApplicationId,
     method: String,
     args: Vec<u8>,
     writes: bool,
 ) -> eyre::Result<Option<String>> {
-    let (result_sender, result_receiver) = oneshot::channel();
+    let (outcome_sender, result_receiver) = oneshot::channel();
```
miraclx: these should at least share a common denominator for idents to show that they are related

crates/server/Cargo.toml
miraclx: what was the error?

crates/server-primitives/src/jsonrpc.rs
```diff
@@ -134,8 +135,9 @@ pub struct MutateResponse {
 
 #[derive(Debug, Serialize, Deserialize, Error)]
 #[error("MutateError")]
+#[serde(tag = "name", content = "cause")]
```
miraclx: Let's keep the current standard to be consistent with the rest of the codebase, so when we migrate to this pattern in a follow-up PR, it would be for all the wire types.

crates/node-primitives/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+license.workspace = true
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+thiserror.workspace = true
+tokio.workspace = true
+serde_json.workspace = true
+
+calimero-primitives = { path = "../primitives" }
+calimero-runtime = { path = "../runtime" }
```
miraclx, line 17: put a todo here to extract primitives from runtime (future effort), as right now we're unintentionally bundling `wasmer` into `calimero-node-primitives`

as a rule of thumb, primitives should depend on primitives, and nothing else.

crates/node-primitives/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+thiserror.workspace = true
+tokio.workspace = true
```
miraclx: This, I'd expect to fail when building just `calimero-node-primitives`, because `mpsc` and `oneshot` rely on the tokio `sync` feature flag being activated (it's implicitly activated by other dependencies of `calimero-node` but this crate has no such luck.

Ideally, as part of our CI efforts, we should build each crate independently to ensure all their minimum requirements for independent builds are met and not require "peer dependencies" for side effect activations.

crates/server/Cargo.toml
fbozic: it wasn't compiling, some invalid method signature(s). should I add it back and fix it?

miraclx: It's fine, leave it out.

crates/server/src/jsonrpc.rs
fbozic: good point, missed this one

crates/node/src/lib.rs
fbozic: I have to pass outcome_sender to call_mut method because it needs to be stored in tx_pool. I am not sure how to push the Result to outcome_sender from the caller of call_mut because I've moved ownership to call_mut. I would have to return outcome_sender back to the caller. Is that the idea?
TBH I would prefer to drop duplicate error reporting and not move outcome_sender soo much.

crates/node-primitives/Cargo.toml
fbozic: Added explicit `sync` feature flag
> Ideally, as part of our CI efforts, we should build each crate independently to ensure all their minimum requirements for independent builds are met and not require "peer dependencies" for side effect activations.

I agree

crates/server/src/lib.rs
fbozic: removed, because `demo` is removed as well

crates/node-primitives/Cargo.toml
miraclx: Linked in https://github.com/calimero-network/core/issues/265

crates/server-primitives/src/jsonrpc.rs
```diff
@@ -111,7 +111,6 @@ pub struct QueryResponse {
 
 #[derive(Debug, Serialize, Deserialize, Error)]
 #[error("QueryError")]
-#[serde(tag = "name", content = "cause")]
```
miraclx: should be `#[serde(tag = "type", content = "data")]` right?

https://github.com/calimero-network/core/blob/5bba02a22bc9ffc2f0677036fdd0e10dd942300e/crates/server-primitives/src/jsonrpc.rs#L87-L88

crates/node/src/lib.rs
fbozic: updated

crates/node/src/lib.rs
```diff
@@ -414,60 +473,44 @@ impl Node {
...
+    async fn prepare_transaction(
         &mut self,
         application_id: calimero_primitives::application::ApplicationId,
         method: String,
         payload: Vec<u8>,
-        outcome_sender: oneshot::Sender<
-            Result<calimero_runtime::logic::Outcome, calimero_node_primitives::CallError>,
-        >,
-    ) -> eyre::Result<calimero_primitives::hash::Hash> {
+    ) -> eyre::Result<calimero_primitives::transaction::Transaction> {
```
miraclx: there already is a concrete error type definition of the only way this function can fail, type erasure and eventual downcasting is not an ideal implementation

use the concrete `Result<Tx, CallError>` instead

elaborated in second comment

crates/node/src/lib.rs
```diff
@@ -414,60 +473,44 @@ impl Node {
...
+            .await
+            .map_err(|e| {
+                eyre!(calimero_node_primitives::CallError::ExecutionError {
+                    message: e.to_string(),
+                })
+            })
     }
 
-    pub async fn call_mut(
+    async fn prepare_transaction(
```
miraclx: It's not guaranteed that the transaction returned here will be committable

The caller would have to know to immediately make the commit call which increases the scope of things that aren't intuitive and rely on prior knowledge on how the system works

These things are halves of the same intent, keeping them together is the better approach

So while in the scope of one function, it can check to guarantee the minimum requirements, and proceed to send

If I understand correctly, this design was to mitigate the fact that we pass the `Sender` by value to the `call_mut` method

Here's what I'll do instead (at this point, all the server handling logic that's embedded inside the `select!` block can be extracted into a method)

```rust
enum CallError {
  Query(QueryCallError),
  Mutable(MutableCallError),
}

enum QueryCallError {}
enum MutableCallError {}
```

the node<->server `Sender` should pass `Result<Outcome, CallError>`
but while handling the server's request, create a local sender that will be used for the rest of the transaction

The `call` and `call_mut` signatures don't need to change, the error handling does though.

Just return the relevant error out of them.

Then in the server handler, we receive the `Sender<Result<Outcome, CallError>>`, create a shiny new `oneshot::Sender<Outcome>` that we can pass to `call_mut` which should return immediately with a result whose `Err(MutCallErr)` we can wrap in a `CallError` to be written to the received `Sender`.

If things worked fine `Ok(TxHash)`, we should `tokio::spawn` a task that awaits the locally created `Sender` and on result, finishes the request `Ok(Outcome)`

crates/server/src/jsonrpc/mutate.rs
```diff
@@ -35,8 +35,9 @@ async fn handle(
             }),
         },
         Ok(None) => Ok(MutateResponse { output: None }),
-        Err(err) => eyre::bail!(MutateError::ExecutionError {
-            message: err.to_string()
-        }),
+        Err(err) => match err.downcast::<calimero_node_primitives::CallError>() {
+            Ok(err) => eyre::bail!(MutateError::CallError(err)),
+            Err(err) => eyre::bail!(err),
+        },
```
fbozic: Is downcasting here okay? Or should I update upstream `call` return `super::CallError` which will wrap inner `CallError`, `FunctionCallError` and `InternalError`.

fbozic: I've removed downcasting.

crates/server/src/jsonrpc.rs
```diff
@@ -135,32 +136,51 @@ impl<T: Serialize, E: Serialize> ToResponseBody for Result<T, RpcError<E>> {
...
+}
+
+pub(crate) async fn call2(
     sender: calimero_node_primitives::ServerSender,
     application_id: calimero_primitives::application::ApplicationId,
     method: String,
     args: Vec<u8>,
     writes: bool,
-) -> eyre::Result<Option<String>> {
+) -> Result<Option<String>, CallError> {
```
miraclx, line 153: don't think you should be forced to reuse the same type here

`CallError` is a node error indication, the server can have it's own that wraps the node's

Especially since the node can't construct the `UpstreamFunctionCallError` variant

So it's not well descriptive of the error cases when calling a method on the node

crates/node/src/lib.rs
```diff
@@ -467,48 +387,103 @@ impl Node {
...
+                |e| calimero_node_primitives::QueryCallError::ExecutionError {
                     message: e.to_string(),
-                })
-            })
+                },
+            )
     }
 
-    async fn prepare_transaction(
+    async fn call_mututate(
```
miraclx: typo

crates/node/src/lib.rs
```diff
@@ -518,50 +493,46 @@ impl Node {
...
-        let tx_hash = self
+        let tx_hash = match self
             .tx_pool
-            .insert(self.id, transaction.clone(), Some(outcome_sender))?;
+            .insert(self.id, transaction.clone(), Some(outcome_sender))
+        {
+            Ok(tx_hash) => tx_hash,
+            Err(err) => {
+                return Err(
+                    calimero_node_primitives::MutateCallError::FailedToInsertTransaction {
```
miraclx: opaque `InternalError` with log is fine

crates/node/src/lib.rs
```diff
@@ -467,48 +387,103 @@ impl Node {
...
+                    payload,
+                    inner_outcome_sender,
+                )
+                .await
+            {
+                let _ = outcome_sender.send(Err(calimero_node_primitives::CallError::Mutate(err)));
+                return;
+            }
+
+            match inner_outcome_receiver.await {
```
miraclx: this model will never resolve because it blocks receiving events from the network that would advance the state

let's `tokio::spawn` at this point, since we no longer need `Node`

crates/node/src/lib.rs
```diff
@@ -92,7 +92,7 @@ pub async fn start(config: NodeConfig) -> eyre::Result<()> {
                 continue;
             }
             Some((application_id, method, payload, write, outcome_sender)) = server_receiver.recv() => {
-                node.handle_transaction(application_id, method, payload, write, outcome_sender).await;
+                node.handle_call(application_id, method, payload, write, outcome_sender).await;
```
miraclx, line 95: danke

crates/node/src/lib.rs
fbozic: my mistake, thanks for pointing it out

crates/node/src/lib.rs
fbozic: awesome stuff, thank you

crates/server/src/jsonrpc.rs
fbozic: I am not, I've implemented `CallError` in the jsonrpc module which wraps upstream errors.

crates/node/src/lib.rs
```diff
@@ -449,16 +506,33 @@ impl Node {
...
+        };
+
+        if let Err(err) = self
+            .push_action(
+                application_id.clone(),
+                types::PeerAction::Transaction(transaction),
+            )
+            .await
+        {
+            let _ = self.tx_pool.remove(&tx_hash).expect(
```
```suggestion
            self.tx_pool.remove(&tx_hash).expect(
```

can also return `Internal` error

crates/node/src/lib.rs
```diff
@@ -449,16 +506,33 @@ impl Node {
...
+            Ok(tx_hash) => tx_hash,
+            Err(err) => {
+                error!(%err, "Failed to insert transaction into the pool.");
+                return Err(calimero_node_primitives::MutateCallError::InternalError);
+            }
+        };
+
+        if let Err(err) = self
+            .push_action(
+                application_id.clone(),
```
```suggestion
                application_id,
```

crates/node/src/lib.rs
```diff
@@ -449,16 +506,33 @@ impl Node {
...
+                types::PeerAction::Transaction(transaction),
+            )
+            .await
+        {
+            let _ = self.tx_pool.remove(&tx_hash).expect(
+            "Failed to remove just inserted transaction from the pool. This is a bug and should be reported.",
+        );
+            return Err(
+                calimero_node_primitives::MutateCallError::FailedToPushTransaction {
```
miraclx: let's treat this as an opaque error maybe? since there's nothing the caller can do about it

crates/node-primitives/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+#[serde(tag = "type", content = "data")]
+pub enum MutateCallError {
+    InvalidNodeType {
+        node_type: NodeType,
+    },
+    ApplicationNotInstalled {
+        application_id: calimero_primitives::application::ApplicationId,
+    },
+    NoConnectedPeers,
+    ExecutionError {
```
miraclx: unused, I think?

crates/node-primitives/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Error)]
+#[error("QueryCallError")]
+#[serde(tag = "type", content = "data")]
+pub enum QueryCallError {
+    ApplicationNotInstalled {
+        application_id: calimero_primitives::application::ApplicationId,
+    },
+    ExecutionError {
```
miraclx: can be opaque, just use `InternalError` here. the runtime-related errors the client cares about are already encoded in `Outcome`

should be fine

crates/server/src/lib.rs
```diff
@@ -81,7 +71,7 @@ pub async fn start(
     {
         if let Some((path, handler)) = jsonrpc::service(&config, server_sender.clone())? {
             app = app.route(path, handler);
-            app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
+            // app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
```
miraclx: don't forget this

PR #261, opened 2024-05-08
feat(sdk): drop custom type discriminant

miraclx:
We already have `unsafe` present so we no longer need this hack. For non-'static events, we lower the event handler to treat non-'static events as 'static. And by reducing it's effect to the `emit` scope, it's guaranteed we should have no unintended side effects.

### DIFF
--- a/crates/sdk/src/event.rs
+++ b/crates/sdk/src/event.rs
@@ -19,7 +19,7 @@ thread_local! {
 
 #[track_caller]
 #[inline(never)]
-fn handler<E: AppEvent + AppEventExt>(event: Box<dyn AppEventExt>) {
+fn handler<E: AppEventExt + 'static>(event: Box<dyn AppEventExt>) {
     if let Ok(event) = E::downcast(event) {
         env::emit(event);
     }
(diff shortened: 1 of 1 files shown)


crates/sdk/src/event.rs
```diff
@@ -40,21 +40,14 @@ pub fn emit<'a, E: AppEventExt + 'a>(event: E) {
 }
 
 mod reflect {
-    #[derive(PartialEq)]
-    pub struct TypeId {
-        id: usize,
-    }
-
-    #[inline(never)]
```
miraclx, line 48: Also, this should've been a plain `#[inline]` so we're not actually calling the function, but it's enough for `rustc` to generate the function pointer.


But that's taking advantage of unintended features in `rustc` that may be patched in the future so this is not guaranteed to work.

To summarize, uncalled functions shouldn't exist. And so, in future versions of `rustc` the function may be optimized away. Which makes this unsound.

https://github.com/sagebind/castaway/pull/6#issuecomment-1150952050

manually casting non-'static -> 'static is much more stable so far as you manage the side-effects of non-static data being treated as one

Issue #235, opened 2024-04-30
Add mock host iface for non-wasm32 targets


miraclx:
When compiling for non-wasm32 targets, you get this error:

![CleanShot 2024-04-30 at 15 10 03@2x](https://github.com/calimero-network/core/assets/16881812/220c88a3-ecc3-401f-bba4-98eebe54c1aa)

Which also means, in IDEs, you get this:

![CleanShot 2024-04-30 at 15 10 44@2x](https://github.com/calimero-network/core/assets/16881812/a6d5af00-c147-4c09-9c3c-a242c6039657)

We can provide mock host interfaces that panic at runtime to prevent its direct use in unit tests, instead favoring [simulation testing](https://github.com/calimero-network/core/issues/236).

PR #281, opened 2024-05-14
feat(app): rock-paper-scissors game

petarjuki7:
##### Run the demo

```console
$ cargo run -p calimero-runtime --example rps
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a Rock-Paper-Scissors game with commit-reveal mechanism.
  - Added game logic for player creation, commitment, revealing choices, and determining the winner.

- **Build Improvements**
  - Automated build process for both `commit-reveal-rps` and `rock-paper-scissors` applications, targeting WebAssembly.

- **Error Handling**
  - Enhanced error handling with new error types for various scenarios in the game logic.

- **Event Handling**
  - Implemented event handling to emit game-related events based on actions.

### DIFF
--- a/apps/rock-paper-scissors/src/repr.rs
+++ b/apps/rock-paper-scissors/src/repr.rs
@@ -1,5 +1,4 @@
 use std::marker::PhantomData;
-use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::{fmt, io};
 
@@ -149,26 +148,5 @@ impl<T: ReprBytes + BorshDeserialize> BorshDeserialize for Repr<T, Bs58> {
         let data = T::deserialize_reader(reader)?;
 
         Ok(Repr::from(data))
-    }
-}
-
-impl<const N: usize> ReprBytes for [u8; N] {
-    type Bytes = [u8; N];
-
-    fn to_bytes(&self) -> Self::Bytes {
-        *self
-    }
-
-    fn from_bytes<F, E>(f: F) -> Option<Result<Self, E>>
-    where
-        F: FnOnce(&mut Self::Bytes) -> Option<E>,
-    {
-        let mut bytes = MaybeUninit::zeroed();
-
-        if let Some(err) = f(unsafe { &mut *bytes.as_mut_ptr() }) {
-            return Some(Err(err));
-        }
-
-        Some(Ok(unsafe { bytes.assume_init() }))
     }
 }


apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+                reveal_correct: false,
+                id: id_insert,
+            },
+        );
+
+        self.commit_state[id_insert] = true;
+
+        env::log(&format!("{:?}", hash));
+        Ok(())
+    }
```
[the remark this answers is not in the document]
miraclx: bad rabbit

apps/commit-reveal-rps/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname $0)"
```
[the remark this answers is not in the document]
miraclx: good suggestions, but we can do them in a separate PR along with updating kv-store & only-peers

apps/commit-reveal-rps/build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+TARGET="${CARGO_TARGET_DIR:-../../target}"
+
+rustup target add wasm32-unknown-unknown
+
+cargo build --target wasm32-unknown-unknown --profile app-release
+
+mkdir -p res
+
+cp $TARGET/wasm32-unknown-unknown/app-release/commit_reveal_rps.wasm ./res/
```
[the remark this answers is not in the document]
miraclx: good suggestions, but we can do them in a separate PR along with updating kv-store & only-peers

apps/commit-reveal-rps/src/lib.rs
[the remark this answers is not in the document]
miraclx: the `env::log` method doesn't return a result, so this suggestion is incorrect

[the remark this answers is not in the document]
miraclx: so there's no applicable suggestion here anymore, am I right? can you resolve this conversation now?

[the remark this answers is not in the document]
miraclx: it's still unresolved, do you not know how? should I resolve it myself?

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+                    env::log(&format!("The game ended in a draw"));
+                    Ok(())
+                }
+            }
+        } else {
+            return Err(Error::NotFound(
+                "One of the players salt wasn't correct or didn't reveal yet".to_string(),
+            ));
+        }
+    }
```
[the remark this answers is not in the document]
miraclx: same as above

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    fn calculate_hash(choice: &str, salt: &str) -> Result<[u8; 32], Error> {
+        Sha3_256::new()
+            .chain_update(Choices::from_str(&choice)?)
+            .chain_update(salt)
+            .finalize()
+            .deref()
+            .try_into()
+            .map_err(|_| Error::ConversionError("Failed to convert ".to_string()))
+    }
```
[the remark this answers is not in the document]
miraclx: bad rabbit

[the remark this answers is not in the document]
miraclx: `Sha3_256::new` doesn't return a result

apps/commit-reveal-rps/src/lib.rs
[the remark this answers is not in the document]
miraclx: you're the problem ✨

apps/commit-reveal-rps/src/lib.rs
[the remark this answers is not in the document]
miraclx: so.. no change here either then, thanks

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+        for choice in choices {
+            if Game::calculate_hash(choice, &salt).unwrap() == hash {
+                return Ok(choice.to_string());
+            }
+        }
+
+        Err(Error::ConversionError(
+            "No choices match the hash and salt".to_string(),
+        ))
+    }
```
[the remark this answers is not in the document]
miraclx: and finally folks, we have a winner.. yeah, let's avoid unwrapping here

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+calimero-sdk = { path = "../core/crates/sdk" }
+ring = "0.17.8"
```
miraclx: unused

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+sha3 = "0.10"
+base64ct = "1.6.0"
```
miraclx: unused

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+base16ct = "0.2.0"
```
miraclx: unused

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+eyre = "0.6.12"
```
miraclx: unused

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+serde = "1.0.201"
```
miraclx: you're using `calimero_sdk::serde` already, so this too is unused

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+repository.workspace = true
+license.workspace = true
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+calimero-sdk = { path = "../core/crates/sdk" }
+ring = "0.17.8"
+sha3 = "0.10"
```
miraclx: as for this, since we're inside the core repo, we are keeping all the versions consolidated at the root manifest 

https://github.com/calimero-network/core/blob/68308bcad459c582ad715b4b7a369e6c206c19e7/Cargo.toml#L28

while workspace member crates reference the root versions, for example:

https://github.com/calimero-network/core/blob/68308bcad459c582ad715b4b7a369e6c206c19e7/crates/runtime/Cargo.toml#L10

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+serde = "1.0.201"
+
+[profile.app-release]
+inherits = "release"
+codegen-units = 1
+opt-level = "z"
+lto = true
+debug = false
+panic = "abort"
+overflow-checks = true
```
miraclx: Add this crate to the workspace members at the root manifest:

https://github.com/calimero-network/core/blob/68308bcad459c582ad715b4b7a369e6c206c19e7/Cargo.toml#L22-L23

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+#[borsh(crate = "calimero_sdk::borsh")]
+struct Game {
+    players: HashMap<usize, Player>,
+    ids: bool,
+    commit_state: [bool; 2], //state for the two id's (did they commit)
+    choices: [Option<String>; 2],
+    results: [u8; 3],
+}
+
+#[app::state]
```
miraclx: just the app state needs to be annotated

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+impl FromStr for Choices {
+    type Err = Error;
+
+    fn from_str(input: &str) -> Result<Choices, Self::Err> {
+        match input {
+            "Rock" => Ok(Choices::Rock),
+            "Paper" => Ok(Choices::Paper),
+            "Scissors" => Ok(Choices::Scissors),
+            _ => Err(Error::ConversionError(
+                "Choose one of Rock, Papar or Scissors".to_string(),
```
miraclx:
```suggestion
                "Choose one of Rock, Paper or Scissors".to_string(),
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+#[derive(Debug)]
+enum Choices {
+    Rock,
+    Paper,
+    Scissors,
+}
+
+impl FromStr for Choices {
```
miraclx: instead of `FromStr`, why not implement `serde::Deserialize` so it's auto-deserialized from the user inputs

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+            Choices::Scissors => b"Scissors",
+        }
+    }
+}
+
+#[app::logic]
+impl Game {
+    pub fn create_player_and_commit(
+        &mut self,
+        choice: &str,
```
miraclx: and this can be

```suggestion
        choice: Choice,
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Default, Debug, BorshSerialize, BorshDeserialize)]
+#[borsh(crate = "calimero_sdk::borsh")]
+struct Player {
+    hash: [u8; 32],
+    reveal_correct: bool,
+    id: usize,
+}
+
+#[derive(Debug)]
+enum Choices {
```
miraclx: should be `Choice` too

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+
+#[app::state(emits = for<'a> Event<'a>)]
+#[derive(Default, Debug, BorshSerialize, BorshDeserialize)]
+#[borsh(crate = "calimero_sdk::borsh")]
+struct Game {
+    players: HashMap<usize, Player>,
+    ids: bool,
+    commit_state: [bool; 2], //state for the two id's (did they commit)
+    choices: [Option<String>; 2],
+    results: [u8; 3],
```
miraclx: as for the state, I think we can reduce all these to this:

```rust
struct Game {
    players: [Option<Player>, 2]
}
```

with `Player` represented as:

```rust
struct Player {
    state: Option<State>,
    name: String
    key: PublicKey,
}
```

and `State` as:

```rust
enum State {
    Committed(Commitment),
    Revealed(Choice),
}
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+            "Paper" => Ok(Choices::Paper),
+            "Scissors" => Ok(Choices::Scissors),
+            _ => Err(Error::ConversionError(
+                "Choose one of Rock, Papar or Scissors".to_string(),
+            )),
+        }
+    }
+}
+
+impl AsRef<[u8]> for Choices {
```
miraclx: derive `BorshSerialize` for `Choice` instead

here's a snip:

```rust
for choice in &[Choice::Rock, Choice::Paper, Choice::Scissors] {
    let mut bytes = [0u8; 1];
    choice.serialize(&mut &mut bytes[..]).unwrap();

    let deserialized = Choice::deserialize(&mut &bytes[..]).unwrap();

    println!("{:?} => {:?} => {:?}", choice, bytes, deserialized);
}
```

```console
Rock => [0] => Rock
Paper => [1] => Paper
Scissors => [2] => Scissors
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+                return Ok(choice.to_string());
+            }
+        }
+
+        Err(Error::ConversionError(
+            "No choices match the hash and salt".to_string(),
+        ))
+    }
+
+    fn determine_winner(choice0: Choices, choice1: Choices) -> Option<usize> {
```
miraclx: this can be a method bound to `Choice`, and in fact, this can a custom `PartialOrd` implementation where

- `Rock > Scissors`
- `Scissors > Paper`
- `Paper > Rock`

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+use std::ops::Deref;
+use std::str::FromStr;
+
+use calimero_sdk::borsh::{BorshDeserialize, BorshSerialize};
+use calimero_sdk::serde::Serialize;
+use calimero_sdk::{app, env};
+use sha3::{Digest, Sha3_256};
+
+#[derive(Debug, Serialize)]
+pub enum Error {
```
miraclx: let's break up these errors so we're being more specific about the types of errors each function actually returns

the model I have above includes this

Cargo.toml
```diff
@@ -22,7 +22,7 @@ members = [
   "./apps/kv-store",
   "./apps/only-peers",
 
-  "./contracts/package-manager",
+  "./contracts/package-manager", "apps/commit-reveal-rps",
```
miraclx: lets follow the order and format, and keep it with the apps section above

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -11,18 +11,4 @@ crate-type = ["cdylib"]
...
-
-[profile.app-release]
-inherits = "release"
-codegen-units = 1
-opt-level = "z"
-lto = true
-debug = false
-panic = "abort"
-overflow-checks = true
+sha3.workspace = true
```
miraclx: this isn't defined in the root manifest, does it build?

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-                reveal_correct: false,
-                id: id_insert,
-            },
-        );
+    fn create_keypair(random_bytes: &[u8; 32]) -> SigningKey {
+        SigningKey::from_bytes(random_bytes)
+    }
 
-        self.commit_state[id_insert] = true;
+    fn sign(mut secret_key: SigningKey, message: &[u8]) -> Result<Signature, SignError> {
```
miraclx: you just defined a private method, this will not be exposed as part of the public interface for the app

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
 
-        self.players.insert(
-            id_insert,
-            Player {
-                hash,
-                reveal_correct: false,
-                id: id_insert,
-            },
-        );
+    fn create_keypair(random_bytes: &[u8; 32]) -> SigningKey {
```
miraclx: same here

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-                    self.results[id] += 1;
-
-                    self.commit_state = Default::default();
-                    self.choices = Default::default();
-                    env::log(&format!("Player {} won", id));
-                    Ok(())
-                }
-                None => {
-                    self.results[2] += 1;
+    fn state(&self) -> [Option<(String, &State)>; 2] {
```
miraclx: and here

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
 
-    fn calculate_hash(choice: &str, salt: &str) -> Result<[u8; 32], Error> {
-        Sha3_256::new()
-            .chain_update(Choices::from_str(&choice)?)
-            .chain_update(salt)
-            .finalize()
-            .deref()
-            .try_into()
-            .map_err(|_| Error::ConversionError("Failed to convert ".to_string()))
+    fn prepare(
```
miraclx: and here

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+                return Ok(());
+            }
+        } else {
+            return Err(RevealError::NotRevealed(
+                "One of the players hasn't revealed yet".to_string(),
+            ));
         }
     }
+
+    fn reset(&mut self) {
```
miraclx: I think reset should only permit the active members of the game state to call it

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+            .chain_update(salt)
+            .finalize()
+            .deref()
+            .try_into()
+            .map_err(|_| Error::ConversionError("Failed to convert ".to_string()))
+    }
 
-        env::log(&format!("New hash is: {:?}", hash));
+    fn compare_hashes(hash: Commitment, salt: &str) -> Result<Choice, Error> {
+        let choices = vec![Choice::Rock, Choice::Paper, Choice::Scissors];
```
miraclx: no need for the heap allocation

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-            .chain_update(Choices::from_str(&choice)?)
-            .chain_update(salt)
-            .finalize()
-            .deref()
-            .try_into()
-            .map_err(|_| Error::ConversionError("Failed to convert ".to_string()))
+    fn prepare(
+        signing_key: &mut SigningKey,
+        choice: Choice,
+        salt: &str,
```
miraclx: I prefer the term `nonce` since by definition, it's only to be used once

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+//#[app::logic]
+impl Game {
+    fn calculate_hash(choice: &Choice, salt: &str) -> Result<Commitment, Error> {
+        Sha3_256::new()
+            .chain_update(choice)
+            .chain_update(salt)
+            .finalize()
+            .deref()
+            .try_into()
+            .map_err(|_| Error::ConversionError("Failed to convert ".to_string()))
```
miraclx: `SHA3_256` output is guaranteed to be 32 bytes, so this can be reduced to:

```rust
Sha2_256::new().chain_update().chain_update().into()
```

making this function infallible

apps/commit-reveal-rps/src/lib.rs
miraclx: return a `Keypair` here, so the caller doesn't have to compute the public key

should be serializable into both components

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+    }
+}
+
+impl Serialize for PublicKey {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let bytes = self.0.to_bytes();
+        serializer.serialize_bytes(&bytes)
```
miraclx: let's `bs58` encode the `SigningKey` and `VerifyingKey` for their serializable representations

apps/commit-reveal-rps/src/lib.rs
miraclx: does `SigningKey` impl `Deserialize`?

I mean, it does, but only when the `serde` flag is activated on `ed25519-dalek` but it doesn't seem to.

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
 use std::ops::Deref;
-use std::str::FromStr;
 
-use calimero_sdk::borsh::{BorshDeserialize, BorshSerialize};
-use calimero_sdk::serde::Serialize;
-use calimero_sdk::{app, env};
+use calimero_sdk::app;
+use calimero_sdk::borsh::{io, BorshDeserialize, BorshSerialize};
+use calimero_sdk::serde::{Deserialize, Serialize};
+use ed25519_dalek::{Signature, Signer, SigningKey, Verifier, VerifyingKey, PUBLIC_KEY_LENGTH};
```
miraclx: this crate is not part of the app's deps

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-            },
-        );
+    fn create_keypair(random_bytes: &[u8; 32]) -> SigningKey {
+        SigningKey::from_bytes(random_bytes)
+    }
 
-        self.commit_state[id_insert] = true;
+    fn sign(mut secret_key: SigningKey, message: &[u8]) -> Result<Signature, SignError> {
+        Ok(secret_key.sign(message))
+    }
```
miraclx: this is infallible, you can remove `Result`

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
 
-        env::log(&format!("{:?}", hash));
-        Ok(())
+    fn verify(
+        &self,
+        player_idx: usize,
+        message: &[u8],
+        signature: Signature,
+    ) -> Result<bool, Error> {
```
miraclx: Seems this can return `Option<bool>`? especially since the error isn't actually serialized

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+        if let Some((index, slot)) = self
+            .players
+            .iter_mut()
+            .enumerate()
+            .find(|(_, player)| player.is_none())
+        //find an empty spot in the game
+        {
+            app::emit!(Event::PlayerInserted {
+                id: index,
+                public_key: &public_key,
```
miraclx: we don't need the public key here

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+            let new_player = Player {
+                state: None,
+                key: public_key,
+                name: player_name,
+            };
+
+            *slot = Some(new_player);
+
+            return Ok(index);
+        } else {
```
miraclx: keep alternative actions close to the conditional check

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-    pub fn reveal(&mut self, salt: &str, id: usize) -> Result<(), Error> {
-        if self.commit_state.into_iter().any(|x| x == false) {
-            return Err(Error::InvalidCall(
-                "Can not reveal yet, one of the players did not commit".to_string(),
+    pub fn join(&mut self, player_name: String, public_key: PublicKey) -> Result<usize, JoinError> {
+        if let Some((index, slot)) = self
+            .players
+            .iter_mut()
+            .enumerate()
+            .find(|(_, player)| player.is_none())
```
miraclx: essentially, rewrite this as:

```rust
let Some((index, slot)) = self.players.iter_mut().enumerate().find(|(_, player)| player.is_none()) else {
    return Err(JoinError::GameFull(
        "Two players are already in the game".to_string(),
    ));
};
```

https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+                name: player_name,
+            };
+
+            *slot = Some(new_player);
+
+            return Ok(index);
+        } else {
+            return Err(JoinError::GameFull(
+                "Two players are already in the game".to_string(),
             ));
```
miraclx: also, your errors are too verbose. is there any other reason why `GameFull` may fail other than "Two players are already in the game"?

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+    AlreadyCommitted(String),
+}
+
+#[derive(Debug, Serialize)]
+pub enum RevealError {
+    PlayerNotFound(String),
+    InvalidNonce(String),
+    NotCommitted(String),
+    NotRevealed(String),
 }
```
miraclx: Variants that contain static strings should have them removed (and just be unit variants), and for those that don't, just encode the variable fragments

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+        &mut self,
+        player_idx: usize,
+        commitment: Commitment,
+        signature: Signature,
+    ) -> Result<(), CommitError> {
+        let player: &mut Player =
+            self.players[player_idx]
+                .as_mut()
+                .ok_or(CommitError::PlayerNotFound(format!(
+                    "Can't find player with idx {}",
```
miraclx: btw, this doesn't qualify as "variable fragment" because there's only one `player_idx` in this call, so it's constant within this context

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-                Choices::Rock => Some(0),
-                Choices::Paper => None,
-                Choices::Scissors => Some(1),
-            },
-            Choices::Scissors => match choice1 {
-                Choices::Rock => Some(1),
-                Choices::Paper => Some(0),
-                Choices::Scissors => None,
-            },
+    fn reveal(&mut self, player_idx: usize, nonce: &str) -> Result<(), RevealError> {
```
miraclx: these non-public methods won't be part of your public API

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+        let player: &mut Player = self
+            .players
+            .get_mut(player_idx)
```
miraclx: Maybe add a new type `PlayerIdx(usize)` that validates it's either `0` or `1`?

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-    fn compare_hashes(hash: [u8; 32], salt: &str) -> Result<String, Error> {
-        let choices = vec!["Rock", "Paper", "Scissors"];
+    fn commit(
+        &mut self,
+        player_idx: usize,
+        commitment: Commitment,
+        signature: Signature,
+    ) -> Result<(), CommitError> {
+        let player: &mut Player =
+            self.players[player_idx]
```
miraclx: especially useful here, this will cause a runtime panic

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
 
-                    self.commit_state = Default::default();
-                    self.choices = Default::default();
-                    env::log(&format!("The game ended in a draw"));
-                    Ok(())
+        for (i, player) in self.players.as_ref().into_iter().enumerate() {
+            match player {
+                Some(p) => {
+                    if let Some(player_state) = &p.state {
+                        states[i] = Some((p.name.clone(), player_state));
```
miraclx: can return a `&str` to avoid the clone

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-                None => {
-                    self.results[2] += 1;
+    fn state(&self) -> [Option<(String, &State)>; 2] {
+        let mut states = [None, None];
 
-                    self.commit_state = Default::default();
-                    self.choices = Default::default();
-                    env::log(&format!("The game ended in a draw"));
-                    Ok(())
+        for (i, player) in self.players.as_ref().into_iter().enumerate() {
```
miraclx:
```suggestion
        for (i, player) in self.players.iter().enumerate() {
```

this works right?

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+            match player {
+                Some(p) => {
+                    if let Some(player_state) = &p.state {
```
miraclx: can be compressed

```suggestion
            if let Some(Player { state: Some(state), .. }) = player {
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+            .filter_map(|p| match p {
+                Some(Player {
+                    state: Some(State::Revealed(choice)),
+                    ..
+                }) => Some(choice),
+                _ => None,
+            })
+            .collect::<Vec<&Choice>>();
+
+        if revealed_choices.len() == 2 {
```
miraclx: just check if the other index is also revealed

```rust
self.players[(player_idx + 1) % 2]
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-        for choice in choices {
-            if Game::calculate_hash(choice, &salt).unwrap() == hash {
-                return Ok(choice.to_string());
+        match player.key.0.verify(&commitment, &signature) {
+            Ok(_) => {
+                app::emit!(Event::PlayerCommited {
+                    id: player_idx,
+                    hash: &commitment
+                });
+                player.state = Some(State::Commited(commitment));
```
miraclx: don't permit single-play, wait until the second player is ready before permitting commit

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+        name: &'a str,
+    },
+    PlayerRevealed {
+        id: usize,
+        reveal: &'a Choice,
+    },
+    PlayerWon {
+        id: usize,
+    },
+    StateDumped {},
```
miraclx:
```suggestion
    StateDumped,
```

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
-        choice: &str,
-        salt: &str,
-        id: Option<usize>,
-    ) -> Result<(), Error> {
-        if self.players.len() == 2 && id == None {
-            return Err(Error::WrongInput(
-                "You have to provide your correct id".to_string(),
-            ));
-        }
+impl Ord for Choice {
```
miraclx: I'd say the expected behavior feels more like `ParialOrd` than `Ord` since

```console
Rock > Scissors > Paper > Rock
```

We don't need `Ord`

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -1,236 +1,368 @@
...
+        } else {
+            return Err(RevealError::NotRevealed(
+                "One of the players hasn't revealed yet".to_string(),
+            ));
         }
     }
+
+    fn reset(&mut self) {
+        self.players[0] = None;
+        self.players[1] = None;
```
miraclx: works, but we don't need to manually index and define them. either of these should suffice

```rust
self.players = [None; 2];
```

```rust
self.players = Default::default();
```

Cargo.toml
```diff
@@ -19,6 +19,7 @@ members = [
   "./crates/server-primitives",
   "./crates/store",
 
+  "./apps/commit-reveal",
```
miraclx: btw, why not use the name of the game?

also consider `./apps/games/{game}`

apps/commit-reveal-rps/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "commit-reveal-rps"
```
miraclx: same here

apps/commit-reveal-rps/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+#[derive(Default, Debug)]
+pub struct PublicKey(VerifyingKey);
+
+impl BorshSerialize for PublicKey {
+    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<(), std::io::Error> {
+        let bytes = self.0.to_bytes();
+        writer.write_all(&bytes)?;
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
miraclx: wrong suggestion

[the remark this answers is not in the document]
miraclx: the errors are the same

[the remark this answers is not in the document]
miraclx: lol

apps/commit-reveal-rps/src/lib.rs
petarjuki7: SigningKey is already a keypair of (secret_key, verifying key).

Should I return my own PublicKey type also?

miraclx: The `serialize`-d representation of `SigningKey` isn't

https://github.com/dalek-cryptography/curve25519-dalek/blob/9252fa5c0d09054fed4ac4d649e63c40fad7abaf/ed25519-dalek/src/signing.rs#L730

> Should I return my own PublicKey type also?

I'd say yeah, you can have a `KeyComponent<T: AsKeyBytes>: Serialize`

then implement `AsKeyBytes` for `SigningKey` and `VerifyingKey` independently

and then follow that up with (pseudocode)

```rust
#[derive(Serialize)]
struct KeyComponents {
    pk: KeyComponent<VerifyingKey>,
    sk: KeyComponent<SigningKey>
}
```

so you can return `KeyComponents`

something of the sort

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    fn into(self) -> usize {
+        self.0
+    }
+}
+
+#[derive(Debug, Serialize)]
+#[serde(crate = "calimero_sdk::serde")]
+pub enum Error {
+    ConversionError(),
+    ResetError(),
```
miraclx:
```suggestion
    ConversionError,
    ResetError,
```

et al.

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+pub enum RevealError {
+    PlayerNotFound(),
+    InvalidNonce(),
+    NotCommitted(),
+    NotRevealed(),
+}
+
+#[app::event]
+pub enum Event<'a> {
+    PlayerCommited { id: usize, hash: &'a [u8; 32] },
```
miraclx: do you think the hash will be useful on the frontend? I can't think of any use for it. So maybe skip

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    PlayerInserted { id: usize, name: &'a str },
```
miraclx: `NewPlayer`?

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            if Game::calculate_hash(&choice, &salt) == hash {
+                return Ok(choice);
+            }
+        }
+
+        Err(Error::ConversionError())
+    }
+
+    pub fn create_keypair(random_bytes: [u8; 32]) -> SigningKey {
+        SigningKey::from_bytes(&random_bytes)
```
miraclx: use this to seed a CSPRNG, like https://rust-random.github.io/rand/rand_chacha/struct.ChaCha20Rng.html#method.from_seed

after which you should be able to:

```rust
SigningKey::generate(ChaCha20Rng::from_seed(seed))
```

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -68,6 +69,70 @@ impl<'de> Deserialize<'de> for PublicKey {
...
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(crate = "calimero_sdk::serde")]
+pub struct KeyComponent<T: AsKeyBytes> {
+    #[serde(
+        serialize_with = "serialize_base58",
+        deserialize_with = "deserialize_base58"
+    )]
+    key_bytes: Vec<u8>,
+    #[serde(skip)]
+    _marker: PhantomData<T>,
```
miraclx:
```suggestion
    key: T,
```

and then you can call `AsKeyBytes::as_key_bytes` on serialization

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -68,6 +69,70 @@ impl<'de> Deserialize<'de> for PublicKey {
...
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(crate = "calimero_sdk::serde")]
+pub struct KeyComponent<T: AsKeyBytes> {
+    #[serde(
+        serialize_with = "serialize_base58",
+        deserialize_with = "deserialize_base58"
+    )]
+    key_bytes: Vec<u8>,
```
miraclx: or at least avoid the allocation, since the size is known

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -68,6 +69,70 @@ impl<'de> Deserialize<'de> for PublicKey {
     }
 }
 
+pub trait AsKeyBytes {
+    fn as_key_bytes(&self) -> &[u8];
```
miraclx:
```suggestion
    fn as_key_bytes(&self) -> &[u8; 32];
```

we can genericize this with an associated type, but it's not necessary since we are dealing with known types

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+use calimero_sdk::app;
+use calimero_sdk::borsh::{io, BorshDeserialize, BorshSerialize};
+use calimero_sdk::serde::{Deserialize, Serialize};
+use ed25519_dalek::{Signature, Signer, SigningKey, Verifier, VerifyingKey, PUBLIC_KEY_LENGTH};
+use sha3::{Digest, Sha3_256};
+
+pub(crate) type Commitment = [u8; 32];
+
+#[derive(Default, Debug)]
+pub struct PublicKey(VerifyingKey);
```
miraclx: this can be a type alias now that `KeyComponent` is present, and move the `Borsh{Se,Des}erialize` to `KeyComponent`

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -68,6 +69,70 @@ impl<'de> Deserialize<'de> for PublicKey {
     }
 }
 
+pub trait AsKeyBytes {
```
miraclx: let's also move these to a new file so it doesn't clutter the core logic of the game

apps/rock-paper-scissors/src/lib.rs
miraclx: actually, since we have `Signature` that also needs to be serialized, use this:

```rust
pub trait AsKeyBytes {
    type Bytes;

    fn as_key_bytes(&self) -> &Self::Bytes;
}
```

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            }
+        }
+        states
+    }
+
+    pub fn prepare(
+        signing_key: SigningKey,
+        choice: Choice,
+        nonce: &str,
+    ) -> Result<(Commitment, Signature), Error> {
```
miraclx: `bs58` encodings for all byte formats

```suggestion
    ) -> Result<(KeyComponent<Commitment>, KeyComponent<Signature>), Error> {
```

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            }) = player
+            {
+                states[i] = Some((name.as_str(), state));
+            }
+        }
+        states
+    }
+
+    pub fn prepare(
+        signing_key: SigningKey,
```
miraclx: that would also apply to inputs, btw

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        Ok((hash, signature))
+    }
+
+    pub fn commit(
+        &mut self,
+        player_idx: PlayerIdx,
+        commitment: Commitment,
+        signature: Signature,
+    ) -> Result<(), CommitError> {
+        if self.players[(player_idx.value() + 1) % 2].is_none() {
```
miraclx: I'd introduce a `self.other_player(idx) -> Option<&Player>` here

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+                id: player_idx.into(),
+                reveal: &choice
+            });
+            player.state = Some(State::Revealed(choice));
+        } else {
+            return Err(RevealError::NotCommitted());
+        }
+
+        let other_idx = (player_idx.value() + 1) % 2;
+        if let Some(other_player) = &self.players[other_idx] {
```
miraclx: should help simplify this too

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+                return Err(RevealError::NotRevealed());
+            }
+        } else {
+            return Err(RevealError::PlayerNotFound());
+        }
+    }
+
+    fn determine_winner(choice0: &Choice, choice1: &Choice) {
+        if choice0 == choice1 {
+            app::emit!(Event::PlayerWon { id: 3 });
```
miraclx: what does 3 mean?

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        let player: &mut Player = self
+            .players
+            .get_mut(player_idx.value())
+            .ok_or_else(|| RevealError::PlayerNotFound())?
+            .as_mut()
+            .unwrap();
+
+        if let Some(State::Commited(commitment)) = player.state {
+            choice =
+                Game::compare_hashes(commitment, nonce).map_err(|_| RevealError::InvalidNonce())?;
```
miraclx: when you remove the empty data, this can be simplified to

```suggestion
                Game::compare_hashes(commitment, nonce).map_err(RevealError::InvalidNonce)?;
```

_but (see next)_

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    fn compare_hashes(hash: Commitment, salt: &str) -> Result<Choice, Error> {
+        let choices: [Choice; 3] = [Choice::Rock, Choice::Paper, Choice::Scissors];
+
+        for choice in choices {
+            if Game::calculate_hash(&choice, &salt) == hash {
+                return Ok(choice);
+            }
+        }
+
+        Err(Error::ConversionError())
```
miraclx: this error is not used, return `Option<Choice>` instead

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    pub fn verify(
+        &self,
+        player_idx: PlayerIdx,
+        message: &[u8],
+        signature: Signature,
+    ) -> Option<bool> {
+        let signing_key = &self.players[player_idx.value()]
+            .as_ref()
+            .or_else(|| return None)
```
miraclx: this seems like a no-op?

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            .unwrap()
```
miraclx: let's avoid unwrapping, `?` should suffice here

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    pub fn reveal(&mut self, player_idx: PlayerIdx, nonce: &str) -> Result<(), RevealError> {
+        let choice: Choice;
+
+        let player: &mut Player = self
+            .players
+            .get_mut(player_idx.value())
+            .ok_or_else(|| RevealError::PlayerNotFound())?
+            .as_mut()
+            .unwrap();
```
miraclx: maybe instead of just `other_player`, make it a `players` method that returns `(&self, &other)`

seems like the present panic points are concerning getting players, would help simplify that part

apps/rock-paper-scissors/src/lib.rs
petarjuki7: It was just a way to represent a draw for now, changed it to `Event::Draw`

apps/rock-paper-scissors/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+license.workspace = true
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+bs58.workspace = true
+calimero-sdk = { path = "../../crates/sdk" }
+ed25519-dalek = { workspace = true, features = ["serde", "rand_core"] }
+getrandom = { version = "0.2", features = ["js"] }
```
miraclx: ?

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -260,7 +260,7 @@ impl PartialOrd for Choice {
     }
 }
 
-#[app::logic]
+//#[app::logic]
```
miraclx: I guess this is why you commented it out?

https://github.com/calimero-network/core/issues/235

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -195,6 +195,7 @@ pub enum Event<'a> {
     NewPlayer { id: usize, name: &'a str },
     PlayerRevealed { id: usize, reveal: &'a Choice },
     PlayerWon { id: usize },
+    Draw,
```
miraclx: `GameOver(Option<idx>)`

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+            _ => Err("Player index must be either 0 or 1."),
+        }
+    }
+
+    fn value(self) -> usize {
+        self.0
+    }
+}
+
+impl Into<usize> for PlayerIdx {
```
miraclx: to avoid clutter, let's pick between `PlayerIdx::value` and this

since this is not an off-the-shelf library, it's the app in and of itself

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(crate = "calimero_sdk::serde")]
+
+pub struct KeyComponents {
+    pub pk: KeyComponent<PublicKey>,
+    pub sk: KeyComponent<SigningKey>,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Deserialize)]
```
miraclx: The point of this new type is to apply the validation checks on deserialization, this isn't doing that

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+#[serde(crate = "calimero_sdk::serde")]
+struct PlayerIdx(usize);
+
+impl PlayerIdx {
+    fn new(value: usize) -> Result<Self, &'static str> {
```
miraclx: just realized this isn't used

apps/rock-paper-scissors/src/lib.rs
petarjuki7: Yeah, it's just distracting sometimes, so I comment it out when writing code

miraclx: yeah, I get that 👍🏽

apps/rock-paper-scissors/Cargo.toml
petarjuki7: To use the rand crate for wasm32-unknown-unknow I had to import it like that. I didn't put it in the top level Cargo.toml because my logic was not to clutter it if not needed.

https://github.com/rust-random/getrandom/issues/208#issuecomment-791896409

miraclx: But you aren't using the `rand` crate? You're using `rand-core`.

And you're not directly using that either.

https://github.com/calimero-network/core/blob/990ab6d9028859f070293520254a49e382abd029/apps/rock-paper-scissors/src/lib.rs#L8-L9

Everything is packaged behind `rand_chacha`, which doesn't require any host ifaces.

petarjuki7: Ah, I had a leftover import, fixed now

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -1,204 +1,34 @@
-use std::cmp::Ordering;
-use std::marker::PhantomData;
+mod choice;
```
miraclx: `cargo +nightly fmt`

we're going with this order:

- std imports
- third-party lib imports
- crate imports
- mod declarations
- self imports

apps/rock-paper-scissors/src/events.rs
miraclx, line 1: it's small enough and easier if this stays in lib

apps/rock-paper-scissors/src/errors.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[derive(Debug, Serialize)]
+#[serde(crate = "calimero_sdk::serde")]
+pub enum Error {
+    ConversionError,
+    ResetError,
+    ByteSizeError,
+}
+
+impl fmt::Display for Error {
```
miraclx: is this used?

apps/rock-paper-scissors/src/events.rs
```diff
@@ -0,0 +1,12 @@
+use calimero_sdk::app;
+
+use crate::Choice;
+
+#[app::event]
+pub enum Event<'a> {
+    PlayerCommited { id: usize },
+    NewPlayer { id: usize, name: &'a str },
+    PlayerRevealed { id: usize, reveal: &'a Choice },
+    GameOver(Option<usize>),
```
miraclx: ah.. let's make it use named fields instead

apps/rock-paper-scissors/src/keys.rs
```diff
@@ -0,0 +1,68 @@
+use ed25519_dalek::{Signature, SigningKey, VerifyingKey};
+
+use crate::errors::Error;
+use crate::{Commitment, PublicKey};
+
+pub trait AsKeyBytes {
```
miraclx: since this is into+from, can be `KeyBytes`

apps/rock-paper-scissors/src/keys.rs
```diff
@@ -0,0 +1,68 @@
...
+    type Bytes: AsRef<[u8]>;
+
+    fn as_key_bytes(&self) -> Self::Bytes;
+    fn from_key_bytes(bytes: &[u8]) -> Result<Self, Error>
```
miraclx: we already know the expected data and its size, so we can eliminate the length check

and reduce the error cases to a validation check, so the only error would be `InvalidData`

though trivial, we also don't need to impose a `Copy` requirement on the bytes, just reference it

```suggestion
    fn as_key_bytes(&self) -> &Self::Bytes;
    fn from_key_bytes(bytes: Self::Bytes) -> Result<Self, Self::Bytes>;
```

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -1,204 +1,34 @@
...
-where
-    D: calimero_sdk::serde::Deserializer<'de>,
-{
-    let s = <std::string::String as Deserialize>::deserialize(deserializer)?;
-    let decoded = bs58::decode(&s)
-        .into_vec()
-        .map_err(calimero_sdk::serde::de::Error::custom)?;
-    Ok(decoded)
-}
+pub(crate) type PublicKey = VerifyingKey;
```
miraclx: since we're using `SigningKey` everywhere else, we can remove this now and use `VerifyingKey` in the same capacity.

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -283,31 +77,37 @@ impl Game {
         None
     }
 
-    pub fn create_keypair(random_bytes: [u8; 32]) -> KeyComponents {
-        let mut csprng = ChaCha20Rng::from_seed(random_bytes);
+    pub fn create_keypair(random_bytes: KeyComponent<[u8; 32]>) -> KeyComponents {
```
miraclx: let's not require clients to bs58 encode their seed

the general idea is we validate things we give clients, and we encode it

the clients should not have data on their end that they "need" to encode/decode

they can leave it to in-memory storage

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -223,55 +53,19 @@ enum State {
...
+            key: Sha3_256::new()
+                .chain_update(choice)
+                .chain_update(salt)
+                .finalize()
+                .into(),
+        }
     }
 
-    fn compare_hashes(hash: Commitment, salt: &str) -> Option<Choice> {
+    fn compare_hashes(hash: KeyComponent<Commitment>, salt: &str) -> Option<Choice> {
```
miraclx: can remain `Commitment`

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -350,22 +152,22 @@ impl Game {
...
-        signing_key: SigningKey,
+        signing_key: KeyComponent<SigningKey>,
         choice: Choice,
         nonce: &str,
-    ) -> Result<(Commitment, Signature), Error> {
-        let hash: Commitment = Game::calculate_hash(&choice, nonce);
-        let signature = SigningKey::sign(&signing_key, &hash);
-        Ok((hash, signature))
+    ) -> Result<(KeyComponent<Commitment>, KeyComponent<Signature>), Error> {
+        let hash: Commitment = Game::calculate_hash(&choice, nonce).key;
```
miraclx: if you don't care about the `KeyComponent` at this point, you shouldn't be returning it

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -223,55 +53,19 @@ enum State {
...
-
 #[app::logic]
 impl Game {
-    fn calculate_hash(choice: &Choice, salt: &str) -> Commitment {
-        Sha3_256::new()
-            .chain_update(choice)
-            .chain_update(salt)
-            .finalize()
-            .into()
+    fn calculate_hash(choice: &Choice, salt: &str) -> KeyComponent<Commitment> {
```
miraclx: revert back to returning `Commitment`

apps/rock-paper-scissors/src/key_component.rs
```diff
@@ -0,0 +1,61 @@
...
+        let bytes = binding.as_ref();
+        writer.write_all(&bytes)?;
+        Ok(())
+    }
+}
+
+impl<T: AsKeyBytes> BorshDeserialize for KeyComponent<T> {
+    fn deserialize_reader<R: std::io::prelude::Read>(reader: &mut R) -> std::io::Result<Self> {
+        let mut key_bytes = vec![];
+        reader.read(&mut key_bytes)?;
```
miraclx: this is not what we want

first, with this condition:

```rust
impl<T: KeyBytes, T::Bytes: BorshDeserialize> BorshDeserialize for KeyComponent<T> {
```

then you can:

```rust
let key_bytes = <T::Bytes as BorshDeserialize>::deserialize_reader(reader);

let key = T::from_key_bytes(key_bytes).map_err(|_| std::io::ErrorKind::InvalidData)?;
```

might be possible to shorten `<T::Bytes as BorshDeserialize>::` to `T::Bytes::` but idk what `rustc` would say without testing it

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -425,26 +231,31 @@ impl Game {
...
             return Err(Error::ResetError);
         }
 
         self.players = Default::default();
-        app::emit!(Event::StateDumped {});
+        app::emit!(Event::StateDumped);
         Ok(())
     }
+
+    fn other_player(&self, my_idx: PlayerIdx) -> Option<&Player> {
```
miraclx: `players` instead, so there's no further indexing anywhere else

https://github.com/calimero-network/core/pull/281#discussion_r1607344267

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -395,23 +201,23 @@ impl Game {
         let player: &mut Player = self
             .players
             .get_mut(player_idx.value())
-            .ok_or_else(|| RevealError::PlayerNotFound)?
+            .ok_or(RevealError::PlayerNotFound)?
             .as_mut()
-            .unwrap();
+            .ok_or(RevealError::PlayerNotFound)?;
```
miraclx: like here, since `player_idx` has already been validated through `Deserialize`, these `PlayerNotFound` errors will never be constructed

apps/rock-paper-scissors/src/lib.rs
```diff
@@ -425,26 +231,31 @@ impl Game {
 
     fn determine_winner(choice0: &Choice, choice1: &Choice) {
         if choice0 == choice1 {
-            app::emit!(Event::Draw);
+            app::emit!(Event::GameOver(None));
         }
         if choice0 > choice1 {
```
miraclx: here we're trusting the implementation of `Choice: ParialOrd` to be consistent with `a == b && !(a > b)`

and it will always call `PartialOrd` twice

you can reduce this to a single `match` statement on `choice0.partial_cmp(choice1)`

apps/rock-paper-scissors/src/keys.rs
petarjuki7: I'm having a problem when I try to implement Deserialize.
```rust
impl<'de, T: KeyBytes> Deserialize<'de> for KeyComponent<T> {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let encoded = <std::string::String as Deserialize>::deserialize(deserializer)?;
        let bytes_decoded = bs58::decode(&encoded)
            .into_vec()
            .map_err(calimero_sdk::serde::de::Error::custom)?;

        let key = match bytes_decoded.len() {
            32 => {
                let mut array = [0u8; 32];
                array.copy_from_slice(&bytes_decoded);
                T::from_key_bytes(array).map_err(calimero_sdk::serde::de::Error::custom(
                    "Invalid byte length",
                ))?
            }
            64 => {
                let mut array = [0u8; 64];
                array.copy_from_slice(&bytes_decoded);
                T::from_key_bytes(array).map_err(calimero_sdk::serde::de::Error::custom(
                    "Invalid byte length",
                ))?
            }
            _ => {
                return Err(calimero_sdk::serde::de::Error::custom(
                    "Invalid byte length",
                ))
            }
        };

        Ok(KeyComponent { key })
    }
}
``` 
Something like this was my idea, but looks way too messy. Since `T::from_key_bytes` takes `Self::Bytes`, I dont know how to coerce either `[u8;32]` or `[u8; 64]` to `<T as KeyBytes>::Bytes`

petarjuki7: Going to push with this error for now

PR #338, opened 2024-05-28
feat(admin-ui): context invitation component

frdomovic:
# feat(admin-ui): context invitation component

## Summary
Component for displaying context invitations, will be connected when API is available

## Test plan
Screenshot 2024-05-28 at 12 51 16

## Summary by CodeRabbit

- **New Features**
	- Added styled button properties for different states like disabled or hover.
	- Updated `ListTableProps` interface to accept an additional parameter in `onRowItemClick` function.
	- Introduced a new `Invitation` entity in the `ContextTable` component.
	- Added a new `handleInvitation` function to handle invitations.
	- Added a React component `invitationRowItem` for rendering invitation details.
	- Included new text related to context invitations in `en.global.json`.
	- Introduced new interfaces `Invitation` and `ContextObject` in `Contexts.tsx`.
	- Added a function `formatDateWithTime` to format timestamps into human-readable dates with time.

### DIFF
--- a/node-ui/src/utils/date.ts
+++ b/node-ui/src/utils/date.ts
@@ -1,26 +1,14 @@
 export function formatDateWithTime(timestamp: string): string {
   const date = new Date(Number(timestamp));
 
-  const day = date.getDate().toString().padStart(2, "0");
-  const monthNames = [
-    "Jan",
-    "Feb",
-    "Mar",
-    "Apr",
-    "May",
-    "Jun",
-    "Jul",
-    "Aug",
-    "Sep",
-    "Oct",
-    "Nov",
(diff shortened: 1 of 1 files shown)


node-ui/src/utils/date.ts
```diff
@@ -0,0 +1,26 @@
+export function formatDateWithTime(timestamp: string): string {
```
MatejVukosav, line 1: There should be localized function to construct format based on users device. We should not hardcode format.

frdomovic: Updated 

test:
Screenshot 2024-05-28 at 15 27 41

PR #341, opened 2024-05-31
feat(network): implement hole punching and rendezvous discovery

fbozic:
## Summary by CodeRabbit

- **New Features**
	- Introduced new behaviors (`dcutr`, `rendezvous`, `relay`) in the `Behaviour` struct.
	- Added a `discovery` module with a `DiscoveryState` struct.

- **Documentation**
	- Updated the documentation for the changes made in the `lib.rs` file.

### DIFF
--- a/crates/network/src/lib.rs
+++ b/crates/network/src/lib.rs
@@ -21,6 +21,8 @@ use client::NetworkClient;
 use config::NetworkConfig;
 
 const PROTOCOL_VERSION: &str = concat!("/", env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
+const CALIMERO_KAD_PROTO_NAME: libp2p::StreamProtocol =
+    libp2p::StreamProtocol::new("/calimero/kad/1.0.0");
 
 #[derive(NetworkBehaviour)]
 struct Behaviour {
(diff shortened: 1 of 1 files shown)


crates/network/src/config.rs
```diff
@@ -61,20 +66,75 @@ impl BootstrapNodes {
...
+    )]
+    pub namespace: rendezvous::Namespace,
+
+    pub discovery_rpm: f32,
+}
+
+impl Default for RendezvousConfig {
+    fn default() -> Self {
+        Self {
+            namespace: rendezvous::Namespace::from_static("/calimero/devnet/global"),
```
MatejVukosav, line 113: Should `devnet` be configurable?

crates/network/src/discovery.rs
```diff
@@ -0,0 +1,414 @@
...
+    pub(crate) fn pending_addr_changes(&self) -> bool {
+        self.pending_addr_changes
+    }
+
+    pub(crate) fn set_pending_addr_changes(&mut self) {
+        self.pending_addr_changes = true;
+    }
+
+    pub(crate) fn clear_pending_addr_changes(&mut self) {
+        self.pending_addr_changes = true;
```
MatejVukosav: Should this be `false`?

crates/network/src/config.rs
fbozic: Once contexts are introduced we are probably going to have mapping `context` ~= `namespace`. This is configurable, and this value is the default. If you have better default value let me know.

crates/network/src/discovery.rs
fbozic: Yeah, good catch.

crates/network/src/lib.rs
```diff
@@ -96,14 +98,24 @@ async fn init(
...
+
                 kad.set_mode(Some(kad::Mode::Client));
+
                 for (peer_id, addr) in bootstrap_peers {
                     kad.add_address(&peer_id, addr);
                 }
                 if let Err(err) = kad.bootstrap() {
-                    warn!("Failed to bootstrap with Kademlia: {}", err);
-                }
+                    warn!(%err, "Failed to bootstrap Kademlia");
```
MatejVukosav, line 116: Why did you change log format? Is that some convention?

fbozic: I thought it was convention (or at least a standard in our repo). I scrolled a bit and realized we have no convention (although a lot of logging is in this format). I prefer to log it like this because it gives more context (the log line will include the object's name before embedding the object).

We should probably standardize all logging in the repo.

PR #354, opened 2024-06-06
feat(admin-api): enable auth headers for admin api routes

frdomovic:
# feat(admin-api): enable auth headers for admin api routes

## Summary
In this PR admin API routes were split into two different routers.
One router is protected routes which should be always protected, and unprotected routes which should not be protected when there are no root keys saved in the DID.

Unprotected routes:
- "/health" - to ping node when setting up the node URL
- "/fetch-challenge" - to be able to sign challenge in process of adding the root key
- "/root-key"- to be able to add the first root key

Once the DID has saved root_key, auth layer is also applied to unprotected routes. 

Requires adding auth headers to calimero sdk and admin dashboard

## Summary by CodeRabbit

- **New Features**
    - Added a constant `APP_URL` and functions for managing application endpoints in `storage.ts`.
    - Introduced functionality for creating authentication headers in `auth/headers.ts`.
    - Updated version from "0.0.22" to "0.0.23" in `package.json`.

- **Refactor**
    - Reorganized routing logic in `admin/service.rs` with new routers and function imports.
    - Updated interfaces in `nodeApi.ts` to include a `contextId` field.
    - Renamed and added variables in routing logic within `admin/service.rs`.

- **Documentation**
    - No alterations to exported or public entities in `package.json` and other files.

### DIFF
--- a/packages/calimero-sdk/package.json
+++ b/packages/calimero-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@calimero-is-near/calimero-p2p-sdk",
-  "version": "0.0.22",
+  "version": "0.0.23",
   "description": "Javascript library to interact with Calimero P2P node",
   "type": "module",
   "main": "lib/index.js",


crates/server/src/admin/storage/did.rs
```diff
@@ -25,6 +25,18 @@ pub fn create_did(store: &Store) -> eyre::Result<Did> {
     Ok(did_document)
 }
 
+pub fn get_did(store: &Store) -> eyre::Result<Option<Did>> {
```
MatejVukosav: This will crash if there is no did. Use `get_or_create_did`

frdomovic:
```
 let did_vec = storage.get(&DID_KEY.as_bytes().to_vec())?;
    match did_vec {
        Some(bytes) => serde_json::from_slice(&bytes)
            .map_err(|e| eyre::Report::new(e).wrap_err("Deserialization error")),
        None => Ok(None)
    }
```
Don't understand, wont it just return none? It says the return type for `storage.get` is `Option<Vec<u8>>` so returns `some` or `none`. And if you are talking about `ReadOnlyStore` implementation, is there something blocking  this code from having crash recovery or?
```
 pub fn get(&self, key: &db::Key) -> eyre::Result<Option<db::Value>> {
        self.inner
            .get(&get_application_key(self.application_id.clone(), key))
    }
```

frdomovic: cc @miraclx

MatejVukosav: Few things:

1. auth should be added to the whole admin service and not just to separate routes
2. we need DID and I don't see any value in not creating it if it does not exist. It eliminates useless checks. What you care about are data inside.

miraclx: seconded

crates/server/src/admin/service.rs
```diff
@@ -186,11 +186,19 @@ fn handle_root_key_result(result: eyre::Result<bool>) -> axum::http::Response<ax
...
     Extension(state): Extension<Arc<AdminState>>,
     Json(req): Json<PubKeyRequest>,
 ) -> impl IntoResponse {
     let recipient = "me";
+    
+    let did_result = crate::admin::storage::did::get_did(&state.store);
+
+    if let Ok(Some(mut did)) = did_result {
+        let result = crate::middleware::auth::auth(&headers, &state.store);
+    }
```
miraclx: and I'm not exactly sure I understand what this is supposed to do

- are we mutating `did` or not?
- do we need `result`?

crates/server/src/admin/service.rs
```diff
@@ -76,7 +74,29 @@ pub(crate) fn setup(
...
+
+    let mut exempted_router = Router::new()
+        .route("/health", get(health_check_handler))
+        .route("/root-key", post(create_root_key_handler))
+        .route("/request-challenge", post(request_challenge_handler))
+        .layer(Extension(shared_state.clone()));
+
+    if has_root_keys {
+        exempted_router = exempted_router.layer(middleware::auth::AuthSignatureLayer::new(store));
+    }
```
miraclx: if I read this right, then the first time this is executed without any root keys, there will be no authentication

and it will stay unauthenticated until we restart the node right?

frdomovic: Yeah, I see the issue now, will add this logic as a layer

crates/server/src/admin/service.rs
```diff
@@ -76,7 +74,29 @@ pub(crate) fn setup(
...
+
+    let did_result = get_or_create_did(&store).map_err(|err| parse_api_error(err));
+
+    let has_root_keys = match did_result {
+        Ok(did) => did.root_keys.len() > 0,
+        Err(_) => false,
+    };
+
+    let mut exempted_router = Router::new()
+        .route("/health", get(health_check_handler))
```
MatejVukosav, line 91: health should remain unprotected

crates/server/src/admin/service.rs
```diff
@@ -76,7 +74,29 @@ pub(crate) fn setup(
...
+        .route("/root-key", post(create_root_key_handler))
+        .route("/request-challenge", post(request_challenge_handler))
```
MatejVukosav: How will you login if you cannot retrieve challenge?

crates/server/src/admin/service.rs
frdomovic: updated

crates/server/src/admin/service.rs
frdomovic: removed this as we dont need auth on exempted routes

crates/server/src/admin/service.rs
```diff
@@ -76,7 +73,19 @@ pub(crate) fn setup(
         .route("/contexts/:context_id/client-keys", get(get_context_client_keys_handler))
         .route("/contexts/:context_id/storage", get(get_context_storage_handler))
         .route("/contexts", get(get_contexts_handler))
-        .layer(Extension(shared_state))
+        .layer(middleware::auth::AuthSignatureLayer::new(store.clone()))
+        .layer(Extension(shared_state.clone()));
+
+    let exempted_router = Router::new()
```
MatejVukosav: non_protected_router

miraclx: unprotected

MatejVukosav: Even better

crates/server/src/admin/service.rs
```diff
@@ -76,7 +73,19 @@ pub(crate) fn setup(
...
+        .route("/health", get(health_check_handler))
+        .route("/root-key", post(create_root_key_handler))
```
MatejVukosav: Should be protected

packages/calimero-sdk/src/api/nodeApi.ts
```diff
@@ -50,12 +50,14 @@ export interface LoginRequest {
...
   payload: Payload;
   walletMetadata: WalletMetadata;
+  contextId: string;
 }
 
 export interface RootKeyRequest {
   walletSignature: String;
   payload: Payload;
   walletMetadata: WalletMetadata;
+  contextId: string;
```
MatejVukosav: why do you need contextId for root key?

packages/calimero-sdk/src/crypto/crypto.ts
```diff
@@ -0,0 +1,54 @@
+import { unmarshalPrivateKey } from '@libp2p/crypto/keys';
```
MatejVukosav: Why crypto/crypto?

`/src/auth/headers`

packages/calimero-sdk/src/storage/storage.ts
```diff
@@ -1,16 +1,17 @@
...
+export const APP_URL = "app-url";
+export const AUTHORIZED = "node-authorized";
 
 export const setStorageClientKey = (clientKey: ClientKey) => {
   localStorage.setItem(CLIENT_KEY, JSON.stringify(clientKey));
 };
 
 export const getStorageClientKey = (): ClientKey | null => {
-  if (typeof window !== 'undefined' && window.localStorage) {
+  if (typeof window !== "undefined" && window.localStorage) {
```
MatejVukosav: Hm formatter should change this to ' . Run it manually

Do you have hooks installed?

packages/calimero-sdk/src/wallets/MetamaskLogin/LoginWithMetamask.tsx
```diff
@@ -118,6 +118,7 @@ export function LoginWithMetamask({
         walletSignature: signData,
         payload: walletSignatureData?.payload,
         walletMetadata: walletMetadata,
+        contextId: applicationId,
```
MatejVukosav: This is not part of this PR.

frdomovic: Screenshot 2024-06-19 at 17 52 33
Auth wrapper already has this context_id header check thats why adding this is needed.

packages/calimero-sdk/src/api/nodeApi.ts
frdomovic: leftover, removed

packages/calimero-sdk/src/crypto/crypto.ts
frdomovic: updated

packages/calimero-sdk/src/storage/storage.ts
frdomovic: updated

packages/calimero-sdk/src/auth/headers.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ClientKey } from '../types/storage';
+import { getStorageClientKey } from '../storage/storage';
+
+export interface AxiosHeader {
+  [key: string]: string;
+}
+
+export async function createAuthHeader(
+  payload: string,
+  contextId: string = "admin-ui"
```
MatejVukosav: this is in sdk. Why admin-ui?

packages/calimero-sdk/src/auth/headers.ts
```diff
@@ -0,0 +1,56 @@
...
+  const privateKey: PrivateKey = await getPrivateKey();
+
+  if (!privateKey) {
+    return null;
+  }
+
+  const encoder = new TextEncoder();
+  const contentBuff = encoder.encode(payload);
+
+  const signing_key = bs58.encode(privateKey.public.bytes);
```
MatejVukosav: signingKey

packages/calimero-sdk/src/wallets/MetamaskLogin/LoginWithMetamask.tsx
MatejVukosav: Now, when I think more about it, I think it will not work. We don't know the context id during login. We only know the application id... and multiple contexts can use the same application

packages/calimero-sdk/src/auth/headers.ts
frdomovic: Linked to previous comment; as we don't have the context id for admin ui login I added this as default value. 
Client key must have contextId as we are linking them later in context details

packages/calimero-sdk/src/wallets/MetamaskLogin/LoginWithMetamask.tsx
frdomovic: It will work when context id replaces application id. When this happens we can just put context id in envs of application e.g. only peers client. 
Right now this works when you pass application id as context id and apart from checking if the header exists there are no other tests with context_id. This new field is only used in context details > client keys so it does not break anything else.

packages/calimero-sdk/src/auth/headers.ts
frdomovic: updated

packages/calimero-sdk/src/auth/headers.ts
MatejVukosav: But add it to admin-ui when it uses sdk, not in sdk itself

packages/calimero-sdk/src/wallets/MetamaskLogin/LoginWithMetamask.tsx
MatejVukosav: But you don't know context upfront

MatejVukosav: and you will be able to switch contexts inside the app

frdomovic: So what do you propose should be done?

packages/calimero-sdk/src/auth/headers.ts
frdomovic: updated

crates/server/src/admin/service.rs
```diff
@@ -91,7 +84,24 @@ pub(crate) fn setup(
             get(handlers::context::get_context_storage_handler),
         )
         .route("/contexts", get(handlers::context::get_contexts_handler))
-        .layer(Extension(shared_state))
+        .layer(middleware::auth::AuthSignatureLayer::new(store.clone()))
```
MatejVukosav: do you need clone here

frdomovic: removed

packages/calimero-sdk/src/wallets/MetamaskLogin/LoginWithMetamask.tsx
MatejVukosav: Lets wait for Mira to deploy context and then we can check it. I think we will need to remove client keys from context and bring them to level of root keys, independent.

crates/server/src/admin/service.rs
```diff
@@ -47,27 +49,18 @@ pub(crate) fn setup(
     let session_layer = SessionManagerLayer::new(session_store).with_secure(false);
 
     let shared_state = Arc::new(AdminState {
-        store,
+        store: store.clone(),
```
MatejVukosav, line 52: Check if this will work across multiple calls. We had some issues with cloning store before but can't recall exactly what was it

packages/calimero-sdk/src/wallets/MetamaskLogin/LoginWithMetamask.tsx
frdomovic: Okay

frdomovic: cc @miraclx

frdomovic: Tested the admin ui and it works

PR #356, opened 2024-06-07
feat: add endpoint for deleting identity keys

MatejVukosav:
During development, we have a need to clean up identity keys. Add endpoint but disable it by default.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to delete authentication keys.
  
- **Bug Fixes**
  - Renamed the function `clean_keys` to `clean_auth_keys` for improved clarity.


crates/server/src/admin/handlers/root_keys.rs
```diff
@@ -68,3 +67,21 @@ pub fn store_root_key(
...
+    data: bool,
+}
+pub async fn delete_auth_keys_handler(
+    Extension(state): Extension<Arc<AdminState>>,
+) -> impl IntoResponse {
+    clean_auth_keys(&state.store).map_or_else(
+        |err| parse_api_error(err).into_response(),
+        |_| {
+            ApiResponse {
+                payload: DeleteKeysResponse { data: true },
```
miraclx: if it can never return `false`, is this information even useful?

crates/server/src/admin/service.rs
```diff
@@ -62,6 +62,7 @@ pub(crate) fn setup(
 
     let admin_router = Router::new()
         .route("/health", get(health_check_handler))
+        // .route("/identity/keys", delete(delete_auth_keys_handler)) // Only for development
```
miraclx: any plans to get this merged in?

MatejVukosav: It is useful for development; ideally, I would explore "dangerous" flag so the user needs to confirm his intention before enabling this by default. CLI is a good candidate for that. Until that, I have disabled it so someone does not just "explore" the API and delete his keys without confirmation.

crates/server/src/admin/handlers/root_keys.rs
MatejVukosav: It is not. I don't yet see the option that will return false, but having it in advance allows us not to change API later.

miraclx: I think `data: false` is not very descriptive even if we did have a use for it.

all cases where `false` could've been returned could just be errors, right?

xilosada: if we return anything in the future it is no going to be a boolean. Errors should be returned instead of false, if it worked the status code will be enough

crates/server/src/admin/handlers/root_keys.rs
```diff
@@ -67,3 +67,21 @@ pub fn store_root_key(
     info!("Root key stored successfully.");
     Ok(true)
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DeleteKeysResponse {
+    data: Empty,
+}
```
miraclx, line 74: Hm, do we need a structured response to represent the absence of data, or am I missing something?

crates/server/src/admin/handlers/root_keys.rs
```diff
@@ -67,3 +67,21 @@ pub fn store_root_key(
...
+pub async fn delete_auth_keys_handler(
+    Extension(state): Extension<Arc<AdminState>>,
+) -> impl IntoResponse {
+    clean_auth_keys(&state.store).map_or_else(
+        |err| parse_api_error(err).into_response(),
+        |_| {
+            ApiResponse {
+                payload: DeleteKeysResponse { data: Empty {} },
```
miraclx, line 82:
```suggestion
                payload: (),
```

but that will return `null` as the body with the `json` content-type

crates/server/src/admin/handlers/root_keys.rs
```diff
@@ -67,3 +67,21 @@ pub fn store_root_key(
...
+            }
+            .into_response()
```
miraclx, line 84: ..or better yet, just `().into_response()` for an empty body, status code 200, no content-type

MatejVukosav: I don't want empty body. Return structure is either data or error

miraclx (approved):

crates/server/src/admin/handlers/root_keys.rs
frdomovic: Why not boolean true?

MatejVukosav: There is no current case that will return false but not error so it does not bring any value

PR #357, opened 2024-06-10
build: add macos workflows

xilosada:
This workflow builds, compresses, caches, and uploads binaries for `calimero-node` for `x86_64-apple-darwin` and `aarch64-apple-darwin` targets. The shared job `build` runs on every push and closed pull request to handle the building and caching of the binaries.

When a push occurs on any branch that is not `main`, the `upload_branch_artifact` job runs. This job restores the cached build artifact and uploads it with a retention period of two days.

When changes are pushed to `main` or a pull request is merged into `main`, the `release` job is triggered. This job also restores the cached build artifact, checks if a release for the current version exists, and if not, creates a new release and uploads the artifact.

Screenshot 2024-06-13 at 3 33 04 PM

## Summary by CodeRabbit

- **New Features**
  - Added a GitHub Actions workflow for building and uploading a binary for the `calimero-node` project on macOS. 
  - The workflow triggers on pushes and pull requests to the `main` branch, includes steps for setting up Rust, installing the target architecture, caching dependencies, building the crate, compressing the artifact using gzip, and uploading the binary as an artifact. 
  - Additional steps involve checking the existence of a release, creating a new release if it doesn't exist, and uploading artifacts to the release.

### DIFF
--- a/.github/workflows/calimero_node_macos.yml
+++ b/.github/workflows/calimero_node_macos.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       matrix:
         target: [x86_64-apple-darwin, aarch64-apple-darwin]
-    if: ${{ github.ref != 'refs/heads/main' }}
+    if: ${{ github.ref != 'refs/heads/master' }}
 
     steps:
       - name: Restore build artifact
@@ -81,7 +81,7 @@ jobs:
     strategy:
       matrix:
         target: [x86_64-apple-darwin, aarch64-apple-darwin]
-    if: ${{ github.ref == 'refs/heads/main' || (github.event.pull_request.merged == true && github.event.pull_request.base.ref == 'main') }}
+    if: ${{ github.ref == 'refs/heads/master' || (github.event.pull_request.merged == true && github.event.pull_request.base.ref == 'master') }}
 
     steps:
       - name: Setup gh CLI


.github/workflows/calimero_node_macos.yml
```diff
@@ -0,0 +1,55 @@
...
+      - main
+
+jobs:
+  build:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        target: [x86_64-apple-darwin, aarch64-apple-darwin]
+
+    steps:
```
MatejVukosav, line 21: Use latest `actions` versions

.github/workflows/calimero_node_macos.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
```
fbozic: That repo is deprecated: https://github.com/actions-rs/toolchain

This should be enough: https://github.com/calimero-network/boot-node/blob/b460f3f2260037a1d2ba1f96c9f9da11e084a5dd/.github/workflows/release.yml#L45

.github/workflows/calimero_node_macos.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: rustup target add ${{ matrix.target }}
+
+      - name: Cache cargo registry
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-registry-
```
fbozic: Maybe this one is more suitable: https://github.com/Swatinem/rust-cache?

.github/workflows/calimero_node_macos.yml
```diff
@@ -0,0 +1,55 @@
...
+          restore-keys: ${{ runner.os }}-cargo-build-
+
+      - name: Build the crate
+        run: cargo build -p calimero-node --release --target ${{ matrix.target }}
+
+      - name: Upload binary
+        uses: actions/upload-artifact@v3
+        with:
+          name: calimero-node_${{ matrix.target }}
+          path: target/${{ matrix.target }}/release/calimero-node
```
fbozic: If we run this on pull_requests, should we skip upload part?

.github/workflows/calimero_node_macos.yml
miraclx: or https://github.com/dtolnay/rust-toolchain which abstracts over the command line

.github/workflows/calimero_node_macos.yml
miraclx: If we determine that we need them, we can define retention periods

https://github.com/actions/upload-artifact#retention-period

xilosada: I guess we can keep the artifacts for testing purposes or builds depending on these binaries. I'll add a retention policy. @fbozic let's catch up in the 1:1

.github/workflows/calimero_node_macos.yml
```diff
@@ -0,0 +1,125 @@
...
+        with:
+          path: calimero-node_${{ matrix.target }}.tar.gz
+          key: build-artifact-${{ matrix.target }}-${{ github.sha }}
+          restore-keys: |
+            build-artifact-${{ matrix.target }}
+
+      - name: Upload binary as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: calimero-node_${{ github.ref_name }}_${{ matrix.target }}.tar.gz
```
miraclx, line 74:
```suggestion
          name: calimero-node_${{ matrix.target }}.tar.gz
```

is this necessary?

fbozic: I think the idea is to have temporary artifacts for the PRs and for the master branch it should be just a version.

xilosada: Yes, the idea is we build the binaries for each branch on every push so we can test them (like for the macOS app). New pushed should overwrite the artifact for the branch.

miraclx: but artifacts are scoped out by job runtime, aren't they?

so two jobs of the same workflow shouldn't provide artifacts that overwrite each other

PR #360, opened 2024-06-10
feat(admin-api): mock fetch context client keys and users

frdomovic:
# feat(admin-api): mock fetch context client keys and users

## Summary
As this should be handled on the application layer, right now only mock logic added and also updated UI to handle when this is available

Fixes # C20-85 + C20-86 (client keys + users)

## Test plan
Screenshot 2024-06-10 at 14 24 06

## Summary by CodeRabbit

- **New Features**
  - Added `ClientKey` and `ContextData` structs to improve context handling.
  - Enhanced `GetContextResponse` structure to include `ContextData`.
  - Updated `ClientKey` struct with an `application_id` field.
  - Implemented `get_context_client_key` function for retrieving client keys.
  - Modified `AuthHeaders` struct to include `application_id` and `context_id` fields.
  - Updated `AddPublicKeyRequest` and `IntermediateAddPublicKeyRequest` with a `context_id` field.


crates/server/src/admin/storage/client_keys.rs
```diff
@@ -25,6 +25,15 @@ pub fn get_client_key(store: &Store, signing_key: &str) -> eyre::Result<Option<C
         .find(|k| k.signing_key == signing_key))
 }
 
+pub fn get_context_client_key(store: &Store, application_id: &str) -> eyre::Result<Vec<ClientKey>> {
+    let did = get_or_create_did(store)?;
+    Ok(did
+        .client_keys
+        .into_iter()
+        .filter(|k| k.application_id == application_id)
```
frdomovic: update to context id later

crates/server/src/middleware/auth.rs
```diff
@@ -152,11 +154,18 @@ fn get_auth_headers(headers: &HeaderMap) -> Result<AuthHeaders, UnauthorizedErro
         .into_vec()
         .map_err(|_| UnauthorizedError::new("Invalid base58 challenge"))?;
 
+    let applicaton_id = headers
+        .get(" applicaton_id")
```
frdomovic:
```suggestion
        .get("applicaton_id")
```

crates/server/src/admin/handlers/add_client_key.rs
```diff
@@ -91,6 +91,7 @@ pub fn store_client_key(
         wallet_type: WalletType::NEAR,
         signing_key: req.payload.message.public_key.clone(),
         created_at: Utc::now().timestamp_millis() as u64,
+        application_id: req.payload.message.application_id.clone()
```
frdomovic: context_id later

MatejVukosav: You don't need to follow the real state. If you use application_id now and our plan is to switch to context_id as otherwise nothing works, you will need to update the server and it will break all apps simultaneously. 

By using contextId server is using proper data and you will only need to pass temp value on client side thus, updating clients to use context id returned from service.

MatejVukosav: I prefer building the system properly than having temp states

frdomovic: alright, will update right away then

frdomovic: updated

PR #369, opened 2024-06-12
feat: add a fetch function inside the runtime host environment

its-saeed:
# [product] short description

## Summary

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Test plan

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a new public function `fetch` in `crates/sdk/src/env.rs` for fetching data from a URL using low-level sys functions.
  - Introduced a new module `ext` in `crates/sdk/src/env.rs` containing the `fetch` function for external data retrieval.
  - Added a function `expected_boolean` in `crates/sdk/src/env.rs` to handle expected boolean values with error handling enhancements.

- **Bug Fixes**
  - Enhanced error handling in `crates/sdk/src/env.rs` for `read_register` and `storage_write` functions.

- **Documentation**
  - Updated the summary in `crates/sdk/src/env.rs` to reflect recent changes and additions.

### DIFF
--- a/crates/sdk/src/env.rs
+++ b/crates/sdk/src/env.rs
@@ -76,12 +76,12 @@ pub fn read_register(register_id: sys::RegisterId) -> Option<Vec<u8>> {
 
     let mut buffer = Vec::with_capacity(len);
 
-    let succeed = unsafe {
+    let succeed: bool = unsafe {
         buffer.set_len(len);
 
         sys::read_register(register_id, sys::BufferMut::new(&mut buffer))
             .try_into()
-            .unwrap_or_else(expected_boolean::<bool>)
+            .unwrap_or_else(expected_boolean)
     };
 
     if !succeed {


crates/runtime/src/logic.rs
```diff
@@ -305,4 +306,50 @@ impl<'a> VMHostFunctions<'a> {
...
+        headers_ptr: u64,
+        headers_len: u64,
+        body_ptr: u64,
+        body_len: u64,
+        out_register_id: u64,
+    ) -> Result<()> {
+        let method = self.get_string(method_ptr, method_len)?;
+        let url = self.get_string(url_ptr, url_len)?;
+        let headers = self.read_guest_memory(headers_ptr, headers_len)?;
+        let headers: HashMap<String, String> = borsh::from_slice(&headers).unwrap();
```
miraclx: handle this error

crates/runtime/src/logic.rs
```diff
@@ -305,4 +306,50 @@ impl<'a> VMHostFunctions<'a> {
...
+
+        let response = if !body.is_empty() {
+            request.send_bytes(&body)
+        } else {
+            request.call()
+        }
+        .map_err(|e| HostError::FetchError {
+            url: url.clone(),
+            error: e.to_string(),
+        })?
```
miraclx: let's pass this error back to the guest, write it to the data register

and return a usize where `0` implies no error (and the register can be read as the response body), and `1` when there's an error and the error is to be read from the out register.

crates/runtime/src/logic/imports.rs
```diff
@@ -44,6 +44,8 @@ impl<'a> VMLogic<'a> {
                 register_id: u64,
             ) -> u32;
             fn storage_read(key_ptr: u64, key_len: u64, register_id: u64) -> u32;
+            fn fetch(method_ptr: u64, method_len: u64, url_ptr: u64, url_len: u64, headers_ptr: u64,
+                 headers_len: u64, body_ptr: u64, body_len: u64, out_register_id: u64);
```
miraclx: break this up for readability, the same way it's broken up in it's other half in `logic`

crates/sdk/src/env.rs
```diff
@@ -145,3 +147,16 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
     };
     storage_write(STATE_KEY, &data);
 }
+
+pub fn fetch(method: &str, url: &str, headers: HashMap<String, String>, body: Vec<u8>) -> String {
```
miraclx:
```suggestion
pub fn fetch(method: &str, url: &str, headers: &[&str, &str], body: &[u8]) -> Vec<u8> {
```

miraclx: this also shouldn't be part of the `env` module, and it should be marked `unsafe`

miraclx: also.. let's put `url` first

crates/sdk/src/env.rs
```diff
@@ -148,30 +148,36 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
...
-    match unsafe { sys::fetch(method, url, headers, body, DATA_REGISTER).try_into() } {
-        Ok(true) => {
-            let data = read_register(DATA_REGISTER).unwrap_or_else(expected_register);
-            Ok(data)
-        }
-        Ok(false) => {
-            let data = read_register(DATA_REGISTER).unwrap_or_else(expected_register);
-            Err(String::from_utf8(data)
-                .unwrap_or_else(|_| panic_str("Cannot convert fetch response to UTF-8 string.")))
+pub mod internal {
```
miraclx: let's use `ext` instead

and mark it `#[doc(hidden)]` for the time being

and pull it out into a separate file under this module

crates/sdk/src/env.rs
```diff
@@ -148,30 +148,36 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
...
+    use super::*;
+
+    #[inline]
```
miraclx: test how this specification affects the wasm binary size, my gut feeling is not inlining would be the better approach

essentially, we shouldn't apply the inline specification for methods that seem particularly chunky

crates/sdk/src/env.rs
```diff
@@ -148,30 +148,36 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
...
+                let data = read_register(DATA_REGISTER).unwrap_or_else(expected_register);
+                Ok(data)
+            }
+            Ok(false) => {
+                let data = read_register(DATA_REGISTER).unwrap_or_else(expected_register);
+                Err(String::from_utf8(data).unwrap_or_else(|_| {
+                    panic_str("Cannot convert fetch response to UTF-8 string.")
+                }))
+            }
+            Err(val) => panic_str(&format!("Expected bool as 0|1, got: {}.", val)),
```
miraclx: I think at this point, we can define a helper method much like `expected_register` for printing bool issues.

so you can call `unwrap_or_else` with it which would help to collapse the match to an `if` statement

crates/runtime/src/logic.rs
```diff
@@ -322,7 +322,11 @@ impl<'a> VMHostFunctions<'a> {
         let url = self.get_string(url_ptr, url_len)?;
         let method = self.get_string(method_ptr, method_len)?;
         let headers = self.read_guest_memory(headers_ptr, headers_len)?;
-        let headers: HashMap<String, String> = borsh::from_slice(&headers).unwrap();
+        let headers: HashMap<String, String> =
+            borsh::from_slice(&headers).map_err(|e| HostError::FetchError {
```
miraclx: I'm not convinced we need a special error type here, and if we did, that'll be credit to a deserialization issue, rather than the actual fetch, so `FetchError` isn't the right wording, I think

miraclx: I'm nearly-temped to restore the unwrap since it's super unlikely to happen, and if it did, it'll be caught by https://github.com/calimero-network/core/blob/203e1469d02695bf1949d352ab7863cd978bb89d/crates/runtime/src/logic/imports.rs#L115-L130

and represented as a `HostError` with contextual information about the source of the error.

crates/sdk/src/env.rs
its-saeed: 291K with inline
292K without inline!

miraclx: how many invocations did you test it with?

its-saeed: 1. Does that matter for the size of the wasm file? 🤔

miraclx: yep, inlining will flatten out the instructions representing the function's logic in the calling position which I'd expect to be more expansive depending on the complexity of the method, whereas not inlining will always jump to the function's definition

which is why it's worth comparing them if you're going to be suggesting inline to the compiler

personally, I lean removing the inline specification in this case

but try some things, report back

crates/runtime/Cargo.toml
```diff
@@ -13,8 +13,10 @@ owo-colors = { workspace = true, optional = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 thiserror.workspace = true
+ureq = "2.9.7"
```
miraclx: specify the version in the root manifest, and follow the convention of referencing the root definitions

crates/runtime/Cargo.toml
```diff
@@ -13,8 +13,10 @@ owo-colors = { workspace = true, optional = true }
...
 wasmer.workspace = true
 wasmer-types.workspace = true
+borsh = { workspace = true, features = ["derive"] }
```
miraclx: order

crates/runtime/src/logic.rs
```diff
@@ -305,4 +305,46 @@ impl<'a> VMHostFunctions<'a> {
...
+        headers_ptr: u64,
+        headers_len: u64,
+        body_ptr: u64,
+        body_len: u64,
+        out_register_id: u64,
+    ) -> Result<u32> {
+        let url = self.get_string(url_ptr, url_len)?;
+        let method = self.get_string(method_ptr, method_len)?;
+        let headers = self.read_guest_memory(headers_ptr, headers_len)?;
+        let headers: Vec<(String, String)> = borsh::from_slice(&headers).unwrap();
```
miraclx: document why this is safe

https://github.com/calimero-network/core/blob/203e1469d02695bf1949d352ab7863cd978bb89d/crates/server/src/lib.rs#L53

crates/runtime/src/logic.rs
```diff
@@ -305,4 +305,46 @@ impl<'a> VMHostFunctions<'a> {
...
+        }
+
+        let response = if !body.is_empty() {
+            request.send_bytes(&body)
+        } else {
+            request.call()
+        };
+
+        let (status, data) = match response {
+            Ok(response) => (0, response.into_string().map_err(|_| HostError::BadUTF8)?),
```
miraclx: it's not guaranteed that the response body will be a string

crates/sdk/src/env.rs
```diff
@@ -19,7 +19,14 @@ pub fn panic_str(message: &str) -> ! {
 #[track_caller]
 #[inline]
 fn expected_register<T>() -> T {
-    panic_str("Expected a register to be set, but it was not.");
+    panic_str("Expected 0|1");
```
miraclx: ?

crates/sdk/src/env.rs
```diff
@@ -19,7 +19,14 @@ pub fn panic_str(message: &str) -> ! {
...
+}
+
+#[track_caller]
+#[inline]
+// TODO: Do I need to make `e` generic?
```
miraclx: not for this case, u32 is fine

crates/sdk/src/env.rs
```diff
@@ -145,3 +152,37 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
     };
     storage_write(STATE_KEY, &data);
 }
+
+pub mod ext {
```
miraclx: please pull this out to a separate file

crates/sdk/src/env.rs
```diff
@@ -145,3 +152,37 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
...
+    use super::*;
```
miraclx: this is discouraged (if you're not using everything in a module), specify your imports

crates/sdk/src/env.rs
```diff
@@ -145,3 +152,37 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
...
+        let headers = match borsh::to_vec(&headers) {
+            Ok(data) => data,
+            Err(err) => panic_str(&format!("Cannot serialize headers: {:?}", err)),
+        };
+        let method = sys::Buffer::from(method);
+        let url = sys::Buffer::from(url);
+        let headers = sys::Buffer::from(headers.as_slice());
+        let body = sys::Buffer::from(body);
+
+        let failed = unsafe {
```
miraclx: only wrap unsafe things in `unsafe` blocks

in this case, `.try_into().unwrap_or_else(..)` don't need to be inside the `unsafe` block

crates/sdk/src/env.rs
```diff
@@ -145,3 +152,37 @@ pub fn state_write<T: crate::state::AppState>(state: &T) {
...
+            sys::fetch(url, method, headers, body, DATA_REGISTER)
+                .try_into()
+                .unwrap_or_else(expected_boolean)
+        };
+        let data = read_register(DATA_REGISTER).unwrap_or_else(expected_register);
+        if failed {
+            Err(String::from_utf8(data)
+                .unwrap_or_else(|_| panic_str("Cannot convert fetch response to UTF-8 string.")))
```
miraclx: change the error message

crates/sdk/src/sys.rs
```diff
@@ -18,4 +19,12 @@ extern "C" {
     // --
     pub fn storage_read(key: Buffer, register_id: RegisterId) -> Bool;
     pub fn storage_write(key: Buffer, value: Buffer, register_id: RegisterId) -> Bool;
+
+    pub fn fetch(
+        url: Buffer,
+        method: Buffer,
+        headers: Buffer,
+        body: Buffer,
+        out_register_id: RegisterId,
```
miraclx: also, can we leave it as `register_id`? out is implied

apps/only-peers/src/lib.rs
```diff
@@ -43,6 +43,18 @@ pub enum Event<'a> {
 
 #[app::logic]
 impl OnlyPeers {
+    pub fn fetch(&mut self) -> String {
+        // let near_sdk = NearSdk::testnet();
+        let response = env::ext::fetch(
```
miraclx: it shouldn't work without being wrapped in an `unsafe` block right?

apps/only-peers/src/lib.rs
```diff
@@ -43,6 +43,18 @@ pub enum Event<'a> {
 
 #[app::logic]
 impl OnlyPeers {
+    pub fn fetch(&mut self) -> String {
```
miraclx: also, if you don't mind and you want to retain the fetch runtime example, create a new app for fetch, call it whatever

apps/only-peers/src/lib.rs
```diff
@@ -43,6 +43,18 @@ pub enum Event<'a> {
...
+        // let near_sdk = NearSdk::testnet();
+        let response = env::ext::fetch(
+            "https://httpbin.org/get",
+            "GET",
+            &[("khar", "gav")],
+            &vec![],
```
miraclx: this can just be `&[]`

crates/sdk/src/env.rs
```diff
@@ -19,7 +19,14 @@ pub fn panic_str(message: &str) -> ! {
...
 fn expected_register<T>() -> T {
-    panic_str("Expected a register to be set, but it was not.");
+    panic_str("Expected 0|1");
+}
+
+#[track_caller]
+#[inline]
+// TODO: Do I need to make `e` generic?
+fn expected_boolean<T>(e: u32) -> T {
+    panic_str(&format!("Expected 0|1. Got {e}"));
```
miraclx, line 31: also, there's other places from the old code that could use this new method

apps/only-peers/src/lib.rs
its-saeed: It was accidentally merged to the branch :D

apps/only-peers/src/lib.rs
its-saeed: Accidental merge!

crates/sdk/src/env.rs
its-saeed: LMTAL

crates/runtime/src/logic.rs
```diff
@@ -305,4 +305,55 @@ impl<'a> VMHostFunctions<'a> {
...
+        };
+
+        let (status, data) = match response {
+            Ok(response) => {
+                let mut buffer = vec![];
+                match response.into_reader().read_to_end(&mut buffer) {
+                    Ok(_) => (0, buffer),
+                    Err(_) => (
+                        1,
+                        "Failed to read the response body.".to_string().into_bytes(),
```
miraclx:
```suggestion
                        "Failed to read the response body.".into(),
```

crates/runtime/src/logic.rs
```diff
@@ -305,4 +305,55 @@ impl<'a> VMHostFunctions<'a> {
...
+                    ),
+                }
+            }
+            Err(e) => (1, e.to_string().into_bytes()),
+        };
+
+        self.with_logic_mut(|logic| {
+            logic
+                .registers
+                .set(&logic.limits, register_id, data.as_slice())
```
miraclx: this will force data to be copied into a new heap allocation

```suggestion
                .set(&logic.limits, register_id, data)
```

crates/sdk/src/env.rs
```diff
@@ -1,5 +1,7 @@
 use crate::sys;
 
+pub mod ext;
```
miraclx, line 4: `#[doc(hidden)]`

crates/sdk/src/env.rs
```diff
@@ -70,10 +78,12 @@ pub fn read_register(register_id: sys::RegisterId) -> Option<Vec<u8>> {
     unsafe {
         buffer.set_len(len);
 
-        match sys::read_register(register_id, sys::BufferMut::new(&mut buffer)).try_into() {
-            Ok(true) => (),
-            Ok(false) => panic_str("Buffer is too small."),
-            Err(val) => panic_str(&format!("Expected bool as 0|1, got: {}.", val)),
+        match sys::read_register(register_id, sys::BufferMut::new(&mut buffer))
```
miraclx: I think you can make this if-else now

crates/sdk/src/env.rs
```diff
@@ -110,10 +120,12 @@ pub fn emit<T: crate::event::AppEvent>(event: T) {
 
 #[inline]
 pub fn storage_read(key: &[u8]) -> Option<Vec<u8>> {
-    match unsafe { sys::storage_read(sys::Buffer::from(key), DATA_REGISTER) }.try_into() {
-        Ok(false) => None,
-        Ok(true) => Some(read_register(DATA_REGISTER).unwrap_or_else(expected_register)),
-        Err(val) => panic_str(&format!("Expected bool as 0|1, got: {}.", val)),
+    match unsafe { sys::storage_read(sys::Buffer::from(key), DATA_REGISTER) }
```
miraclx: [`.then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then)

crates/sdk/src/env/ext.rs
```diff
@@ -0,0 +1,32 @@
...
+    let url = sys::Buffer::from(url);
+    let headers = sys::Buffer::from(headers.as_slice());
+    let body = sys::Buffer::from(body);
+
+    let failed = unsafe { sys::fetch(url, method, headers, body, DATA_REGISTER) }
+        .try_into()
+        .unwrap_or_else(expected_boolean);
+    let data = read_register(DATA_REGISTER).unwrap_or_else(expected_register);
+    if failed {
+        Err(String::from_utf8(data).unwrap_or_else(|_| {
```
miraclx: typically for these tail patterns, this would be more compact:

```rust
if failed {
    return Err(..);
}

Ok(data)
```

crates/runtime/src/logic.rs
miraclx: for context, (`Registers::set` needs any `T: Into<Box<[u8]>>`)

https://github.com/calimero-network/core/blob/cf9d816b98708c39b154de24b81cc36f57b1bea5/crates/runtime/src/logic/registers.rs#L27-L29

where, in the case of a `Vec<u8>`, it just reuses the allocation

https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3471-3495

but in the case of a `&[u8]`, makes a new allocation

https://doc.rust-lang.org/src/alloc/boxed.rs.html#1521-1539

crates/sdk/src/env.rs
```diff
@@ -75,16 +76,16 @@ pub fn read_register(register_id: sys::RegisterId) -> Option<Vec<u8>> {
...
 
-        match sys::read_register(register_id, sys::BufferMut::new(&mut buffer))
+        sys::read_register(register_id, sys::BufferMut::new(&mut buffer))
             .try_into()
-            .unwrap_or_else(expected_boolean)
-        {
-            true => (),
-            false => panic_str("Buffer is too small."),
-        }
+            .unwrap_or_else(expected_boolean::<bool>)
```
miraclx: did it complain without the qualifying turbofish?

its-saeed: yes! But let me try again to be sure!

PR #374, opened 2024-06-13
feat(ext): add app-side library for NEAR

its-saeed:
## Summary

We need to develop a high-level library that wraps around the low-level `fetch` function to call query methods of the NEAR blockchain in a deterministic way. This library should abstract the complexities of the low-level function and provide a simple and intuitive interface for developers.

## Test plan

For internal tests, I added this function to the only-peers app to call NEAR query methods:
```rust
#[app::logic]
impl OnlyPeers {
    pub fn fetch(&mut self) -> String {
        let client = Client::testnet();
        let request = RpcQueryRequest {
            block_id: calimero_sdk_near::BlockId::Height(166710735),
            request: QueryRequest::ViewAccount {
                account_id: "nearkat.testnet".parse().unwrap(),
            },
        };
        match client.call(request) {
            Ok(r) => format!("{:?}", r),
            Err(e) => format!("{:?}", e),
        }
    }
```

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `Client` struct for interacting with NEAR Protocol RPC endpoints.
  - Added new RPC query requests and response structs to handle NEAR Protocol queries.
  - Implemented JSON-RPC client functionality for NEAR Protocol communication.

- **Enhancements**
  - Updated dependency in `Cargo.toml` to `calimero-sdk-near`.
  - Added a new public function `foo` to the `OnlyPeers` struct for retrieving account balance.
  - Extended serialization and deserialization traits for custom enums.

- **Bug Fixes**
  - Updated `peer_id` field type in `PeerJoinedPayload` struct to `libp2p_identity::PeerId`.

- **Refactor**
  - Changed dependency from `libp2p` to `libp2p-identity` with additional features.
  - Updated import statements and type declarations to reflect new dependencies.

- **Chores**
  - Modified example runtime script to change the function argument, affecting the program's behavior.


Cargo.toml
```diff
@@ -15,6 +15,7 @@ members = [
   "./crates/primitives",
   "./crates/runtime",
   "./crates/sdk",
+  "./crates/near-sdk",
```
miraclx: revise this name, it's not near's SDK, but calimero's libraries for other SDKs.

something like `./crates/sdk/(shim|libs)/near`

crates/near-sdk/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "near-sdk"
```
miraclx: `calimero-sdk-near` or something of that order

crates/near-sdk/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
+use calimero_sdk::env::internal::fetch;
```
miraclx: import `calimero_sdk::env` and use `env::ext::fetch` at the caller

crates/near-sdk/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
...
+use serde::de::DeserializeOwned;
+use serde::Deserialize;
+use std::collections::HashMap;
```
miraclx: follow the import order in other crates

`std` comes first, then third-party libs

crates/near-sdk/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub struct Client {
```
miraclx: none of our exports should permit arbitrary request construction

```suggestion
pub(crate) struct Client {
```

crates/near-sdk/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
...
+    ) -> Response<T, E> {
+        let headers = HashMap::from([("Content-Type".to_string(), "application/json".to_string())]);
+
+        let body = serde_json::to_vec(&serde_json::json!({
+            "jsonrpc": "2.0",
+            "id": "1",
+            "method": method,
+            "params": params,
+        }))
+        .unwrap();
```
miraclx: no panics please

crates/near-sdk/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
...
+
+        let response = unsafe { fetch("POST", &self.url, headers, body) };
+        serde_json::from_str(&response).unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Response<T, E> {
+    pub result: Option<T>,
+    pub error: Option<RpcError<E>>,
```
miraclx: this should be `result` or `error`, never both

with

```suggestion
    #[serde(with = `calimero_primitives::common::ResultAlt`)]
    pub data: Result<T, RpcError<E>>
```

you can update `calimero_primitives::common` to include this:

```rust
#[derive(Serialize, Deserialize)]
#[serde(remote = "Result")]
pub enum ResultAlt<T, E> {
    #[serde(rename = "result")]
    Ok(T),
    #[serde(rename = "error")]
    Err(E),
}

impl<T, E> From<ResultAlt<T, E>> for Result<T, E> {
    fn from(result: ResultAlt<T, E>) -> Self {
        match result {
            ResultAlt::Ok(value) => Ok(value),
            ResultAlt::Err(err) => Err(err),
        }
    }
}
```

crates/near-sdk/src/jsonrpc.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub id: String,
+
+    pub jsonrpc: Option<String>,
```
miraclx: the order `jsonrpc`, `id`, `data` is much more intuitive for progression

crates/near-sdk/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+use jsonrpc::Response;
+use serde::Deserialize;
+
+mod jsonrpc;
+
+pub struct NearSdk {
+    jsonrpc_client: jsonrpc::Client,
```
miraclx:
```suggestion
    client: jsonrpc::Client,
```

[this is fine](https://media.npr.org/assets/img/2023/01/14/this-is-fine_custom-b7c50c845a78f5d7716475a92016d52655ba3115.jpg?s=1300&c=85&f=webp)

crates/near-sdk/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+                "request_type": "view_account",
+                "finality": "final",
+                "account_id": account_id,
+            }),
+        );
+
+        if let Some(response) = response.result {
+            response
+        } else if let Some(error) = response.error {
+            panic!("Error: {:?}", error.message);
```
miraclx: no panics please

crates/near-sdk/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    pub fn mainnet() -> Self {
+        Self {
+            jsonrpc_client: jsonrpc::Client::new("https://rpc.mainnet.near.org/".to_string()),
+        }
+    }
+
+    pub fn get_balance(&self, account_id: &str) -> BalanceResponse {
```
miraclx: see [near_jsonrpc_primitives](https://docs.rs/near-jsonrpc-client/latest/near_jsonrpc_client/methods/index.html)

crates/near-sdk/Cargo.toml
its-saeed: what about calimero-near-bridge? 🤔

crates/runtime/src/logic.rs
```diff
@@ -334,22 +333,20 @@ impl<'a> VMHostFunctions<'a> {
...
-        })?;
+        };
 
-        self.with_logic_mut(|logic| {
-            logic
-                .registers
-                .set(&logic.limits, out_register_id, response.into_bytes())
-        })?;
-        Ok(())
+        let (status, data) = match response {
```
miraclx: @its-saeed, can you rebase on master to remove the noise?

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl<T, E> From<ResultAlt<T, E>> for Result<T, E> {
+    fn from(result: ResultAlt<T, E>) -> Self {
+        match result {
+            ResultAlt::Ok(value) => Ok(value),
+            ResultAlt::Err(err) => Err(err),
+        }
+    }
+}
```
miraclx: move to `calimero-primitives`

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn call<T: DeserializeOwned, E: DeserializeOwned>(
+        &self,
+        method: &str,
+        params: serde_json::Value,
+    ) -> Result<Response<T, E>, String> {
+        let headers = [("Content-Type", "application/json")];
+
+        let body = serde_json::to_vec(&serde_json::json!({
+            "jsonrpc": "2.0",
+            "id": "1",
```
miraclx: don't assume an id

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,67 @@
...
+            "method": method,
+            "params": params,
+        }))
+        .map_err(|err| format!("Cannot serialize request: {:?}", err))?;
+
+        let response = unsafe { env::ext::fetch(&self.url, "POST", &headers, &body) }?;
+        let response = String::from_utf8(response).map_err(|e| e.to_string())?;
+        serde_json::from_str::<Response<T, E>>(&response).map_err(|e| e.to_string())
```
miraclx: `serde_json::from_slice`

crates/runtime/src/logic.rs
```diff
@@ -309,20 +308,20 @@ impl<'a> VMHostFunctions<'a> {
...
-        out_register_id: u64,
-    ) -> Result<()> {
-        let method = self.get_string(method_ptr, method_len)?;
+        register_id: u64,
+    ) -> Result<u32> {
         let url = self.get_string(url_ptr, url_len)?;
+        let method = self.get_string(method_ptr, method_len)?;
         let headers = self.read_guest_memory(headers_ptr, headers_len)?;
-        let headers: HashMap<String, String> = borsh::from_slice(&headers).unwrap();
+        let headers: Vec<(String, String)> = borsh::from_slice(&headers).unwrap(); // safety: headers are coming from an inner source. Safe to deserialize.
```
miraclx: what "inner source"?

crates/runtime/src/logic.rs
```diff
@@ -334,22 +333,20 @@ impl<'a> VMHostFunctions<'a> {
...
-        Ok(())
+        let (status, data) = match response {
+            Ok(response) => {
+                let mut buffer = vec![];
+                match response.into_reader().read_to_end(&mut buffer) {
+                    Ok(_) => (0, buffer),
+                    Err(_) => (1, "Failed to read the response body.".into()),
+                }
+            }
+            Err(e) => (1, e.to_string().into_bytes()),
```
miraclx: just realized: how about the HTTP status code, response headers, etc.. it's not just the body we should care about

crates/sdk/libs/near/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+            Err(e) => Err(e.message),
+        }
+    }
+
+    pub fn view_code(&self, account_id: &str) -> Result<views::ContractCodeView, String> {
+        let response: Response<views::ContractCodeView, String> = self.client.call(
+            "query",
+            serde_json::json!({
+                "request_type": "view_code",
+                "finality": "final",
```
miraclx: this is by definition, non-deterministic

reduce it to a block hash / block height instead

crates/sdk/libs/near/src/jsonrpc.rs
its-saeed: I did that before! why did I change it again?!

crates/primitives/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license.workspace = true
 [dependencies]
 bs58.workspace = true
 ed25519-dalek.workspace = true
-libp2p = { workspace = true, features = ["serde"] }
+libp2p-identity = {workspace=true, features = ["peerid", "serde"]}
```
miraclx:
```suggestion
libp2p-identity = {workspace = true, features = ["peerid", "serde"]}
```

crates/sdk/libs/near/Cargo.toml
```diff
@@ -10,4 +10,5 @@ license.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 calimero-sdk = { path = "../../" }
```
miraclx:
```suggestion

calimero-sdk = { path = "../../" }
```

crates/runtime/src/logic.rs
```diff
@@ -321,7 +321,9 @@ impl<'a> VMHostFunctions<'a> {
         let url = self.get_string(url_ptr, url_len)?;
         let method = self.get_string(method_ptr, method_len)?;
         let headers = self.read_guest_memory(headers_ptr, headers_len)?;
-        let headers: Vec<(String, String)> = borsh::from_slice(&headers).unwrap(); // safety: headers are coming from an inner source. Safe to deserialize.
+
+        // safety: It's not possible to call `fetch` directly by apps. So headers is generated by our code. Safe to deserialize.
```
miraclx: it's possible, all they have to do is add this to their code:

```rust
#[link(wasm_import_module = "env")]
extern "C" {
    pub fn fetch(
        url_ptr: u64, url_len: u64,
        method_ptr: u64, method_len: u64,
        headers_ptr: u64, headers_len: u64,
        body_ptr: u64, body_len: u64,
        register_id: u64,
    ) -> u32;
}
```

crates/sdk/libs/near/src/query.rs
```diff
@@ -0,0 +1,33 @@
+use crate::{
+    types::{BlockHash, BlockHeight, BlockId},
+    views::{
+        AccessKeyList, AccessKeyView, AccountView, CallResult, ContractCodeView, QueryRequest,
+        ViewStateResult,
+    },
+};
```
fbozic: By some internal convention we tend to split this into one liners.

Something like this

```
use crate::types::{BlockHash, BlockHeight, BlockId};
use crate::views::{
        AccessKeyList, AccessKeyView, AccountView, CallResult, ContractCodeView, QueryRequest,
        ViewStateResult,
};
```

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
...
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+
+pub(crate) struct Client {
+    url: String,
+    id: std::cell::RefCell<u64>,
+}
+
+impl Client {
+    pub fn new(url: String) -> Self {
```
fbozic: ~question: Does this mean we can create a 'plain' JsonRpc client in the app and actually `call` any method that we want?

crates/sdk/libs/near/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+license.workspace = true
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
+serde_with = {workspace = true, features = ["base64"]}
+
+calimero-sdk = { path = "../../" }
+calimero-primitives = { path = "../../../primitives" }
+
```
fbozic: ~nit: Extra new line at the end

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
+use calimero_sdk::env;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+
+pub(crate) struct Client {
+    url: String,
+    id: std::cell::RefCell<u64>,
```
fbozic, line 10: ~question: Why do we need `RefCell` here? Isn't our runtime synchronous?

crates/sdk/libs/near/src/jsonrpc.rs
its-saeed: Good question. No, it doesn't. Take a look at `pub(crate) struct Client`

fbozic: A cool, missed that, thanks

crates/sdk/libs/near/src/jsonrpc.rs
its-saeed: It's for [interior mutability pattern](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html). Instead of making the whole client mutable to increase the `id`, the id itself is now mutable.

crates/sdk/libs/near/src/query.rs
its-saeed: Can't we put in `rustfmt.toml` :) Every time I change it and save the document it changes back to what it was!

crates/runtime/src/logic.rs
miraclx: worth looking into, when we redesign this to accommodate returning more complex data like status codes, and response headers

let's break up how linked data is sent over the bridge, building a linked list backed by the runtime registers, akin to how curl stores headers backed by heap memory

the idea is:

runtime:

1. define the 0th register as reserved
2. expose a method taking 3 parameters to the guest that appends to the link list
  - the first parameter should be the last register id we stored at
  - the second parameter should be a string (`sys::Buffer`)
  - the third parameter should be the register id that stores the new entry
3. to store a new entry in the linked list, we insert this at the new register location:
  - LE encoded register id for the last entry
  - the data we intend to store
4. reading this linked list would always start from the last register id and so would freeing it

guest:

how are register id's derived?

one option is to create a `Box::new(0u8)`, and use it's address for every entry.. since the guest is in a 32-bit environment, all guest memory addresses would be valid register IDs

see https://curl.se/libcurl/c/curl_slist_append.html

---

Non-blocking for this PR though, since we'll need to revise the `fetch` host-function

crates/sdk/libs/near/src/query.rs
fbozic: @miraclx made some efforts in that area. Can you run `cargo +nightly fmt`? 

We have this config. 
```
group_imports = "StdExternalCrate"
imports_granularity = "Module"
```

crates/sdk/libs/near/src/lib.rs
```diff
@@ -133,4 +133,34 @@ impl Client {
...
+            self.client.call("query", request)?;
+
+        match response.data {
+            Ok(r) => {
+                if let QueryResponseKind::CallResult(cr) = r.kind {
+                    return Ok(cr);
+                }
+                return Err("Unexpected response returned.".to_string());
+            }
+            Err(e) => Err(format!("Error: {}, Code: {}", e.message, e.code,)),
```
miraclx: Why not return the `RpcError`?

its-saeed: What if `client.call` fails for some other errors other than the error part in the JSON-RPC response body? We need some more general error type here

miraclx: imo, `RpcError` should be this

crates/runtime/src/logic.rs
its-saeed: > it's possible, all they have to do is add this to their code:

So it's not safe to call `unwrap` :D

miraclx: not exactly, but leave it for #401

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
...
+        serde_json::from_slice::<Response<T, E>>(&response)
+            .map_err(|e| format!("Failed to parse response: {}", e.to_string(),))
+    }
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct Request<P: Serialize> {
+    pub jsonrpc: &'static str,
+    pub id: String,
+    pub method: String,
```
miraclx: these two can be `&str` too

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        let response = unsafe { env::ext::fetch(&self.url, "POST", &headers, &body) }?;
+        serde_json::from_slice::<Response<T, E>>(&response)
+            .map_err(|e| format!("Failed to parse response: {}", e.to_string(),))
+    }
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct Request<P: Serialize> {
+    pub jsonrpc: &'static str,
```
miraclx: doesn't have to be static

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
...
+        .map_err(|err| format!("Cannot serialize request: {:?}", err))?;
+
+        let response = unsafe { env::ext::fetch(&self.url, "POST", &headers, &body) }?;
+        serde_json::from_slice::<Response<T, E>>(&response)
+            .map_err(|e| format!("Failed to parse response: {}", e.to_string(),))
+    }
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct Request<P: Serialize> {
```
miraclx: why `pub`?

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
...
+        &self,
+        method: &str,
+        params: P,
+    ) -> Result<Response<T, E>, String> {
+        let headers = [("Content-Type", "application/json")];
+
+        *self.id.borrow_mut() += 1;
+        let body = serde_json::to_vec(&Request {
+            jsonrpc: "2.0",
+            id: self.id.borrow().to_string(),
```
miraclx: it's fine to be a int

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,63 @@
...
+            method: method.to_string(),
+            params,
+        })
+        .map_err(|err| format!("Cannot serialize request: {:?}", err))?;
```
miraclx: `String` errors are unideal, please fix error handling

miraclx: If you don't care about the error, then type erase, it's at least a midway point to structured errors

crates/sdk/libs/near/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn view_account(
+        &self,
+        account_id: &str,
+        block_id: BlockId,
+    ) -> Result<views::AccountView, String> {
+        let request = RpcQueryRequest {
+            block_id,
+            request: QueryRequest::ViewAccount {
+                account_id: account_id.to_string(),
+            },
```
miraclx: if you're not exposing these types, you might as well use references, avoid the clones

miraclx: and if you're exposing the types, then this might be a better design (since these methods are just wrappers around generic functionality):

https://docs.rs/near-jsonrpc-client/latest/near_jsonrpc_client/struct.JsonRpcClient.html#method.call

crates/sdk/libs/near/src/types.rs
```diff
@@ -0,0 +1,30 @@
+use serde_with::base64::Base64;
+use serde_with::serde_as;
+
+pub type BlockHeight = u64;
+pub type BlockHash = String;
+pub type AccountId = String;
```
miraclx: https://docs.rs/near-account-id/latest/near_account_id/index.html

crates/sdk/libs/near/src/types.rs
```diff
@@ -0,0 +1,30 @@
+use serde_with::base64::Base64;
+use serde_with::serde_as;
+
+pub type BlockHeight = u64;
+pub type BlockHash = String;
```
miraclx: `calimero_primitives::hash::Hash` should work here

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,105 @@
...
+    ViewAccessKey {
+        account_id: AccountId,
+        public_key: String,
+    },
+    ViewAccessKeyList {
+        account_id: AccountId,
+    },
+    CallFunction {
+        account_id: AccountId,
+        method_name: String,
```
miraclx: since these are views into the response, let's perhaps use immutable representations

```suggestion
        method_name: Box<str>,
```

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub storage_usage: StorageUsage,
+    pub storage_paid_at: BlockHeight,
+}
+
+#[serde_as]
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct ContractCodeView {
+    #[serde(rename = "code_base64")]
+    #[serde_as(as = "Base64")]
+    pub code: Vec<u8>,
```
miraclx:
```suggestion
    pub code: Box<[u8]>,
```

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub value: StoreValue,
+}
+
+#[serde_as]
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct ViewStateResult {
+    pub values: Vec<StateItem>,
+    #[serde_as(as = "Vec<Base64>")]
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub proof: Vec<Arc<[u8]>>,
```
miraclx:
```suggestion
    pub proof: Box<[Box<[u8]>]>,
```

and others

crates/runtime/src/logic.rs
its-saeed: Will be addressed in #401

crates/near-sdk/src/jsonrpc.rs
MatejVukosav: Just help yourself with "cargo +nightly fmt"

Cargo.toml
```diff
@@ -53,6 +53,7 @@ multiaddr = "0.18.1"
 near-jsonrpc-client = "0.8.0"
 near-jsonrpc-primitives = "0.20.0"
 near-primitives = "0.20.0"
+near-account-id = "1.0.0"
```
miraclx: nit: order

crates/sdk/libs/near/Cargo.toml
```diff
@@ -7,6 +7,7 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
+near-account-id = { workspace = true }
```
miraclx:
```suggestion
near-account-id.workspace = true
```

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -25,8 +25,8 @@ impl Client {
         *self.id.borrow_mut() += 1;
         let body = serde_json::to_vec(&Request {
             jsonrpc: "2.0",
-            id: self.id.borrow().to_string(),
-            method: method.to_string(),
+            id: &*self.id.borrow().to_string(),
```
miraclx: can be a number

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -38,10 +38,10 @@ impl Client {
 }
 
 #[derive(Debug, Clone, Serialize)]
-pub struct Request<P: Serialize> {
-    pub jsonrpc: &'static str,
-    pub id: String,
-    pub method: String,
+struct Request<'a, P: Serialize> {
+    pub jsonrpc: &'a str,
+    pub id: &'a str,
```
miraclx:
```suggestion
    pub id: u64,
```

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -38,10 +38,10 @@ impl Client {
 }
 
 #[derive(Debug, Clone, Serialize)]
-pub struct Request<P: Serialize> {
-    pub jsonrpc: &'static str,
-    pub id: String,
-    pub method: String,
+struct Request<'a, P: Serialize> {
+    pub jsonrpc: &'a str,
```
miraclx: I don't think these fields need to be `pub` either

crates/sdk/libs/near/src/jsonrpc.rs
miraclx: not resolved

crates/sdk/libs/near/src/lib.rs
```diff
@@ -25,142 +31,10 @@ impl Client {
...
-                }
-                return Err("Unexpected response returned.".to_string());
-            }
-            Err(e) => Err(format!("Error: {}, Code: {}", e.message, e.code,)),
-        }
+    pub fn call<M>(&self, method: M) -> Result<Response<M::Response, String>, NearLibError>
+    where
+        M: RpcMethod,
+    {
+        self.client.call(method.method_name(), &method.params()?)
```
miraclx: inline this

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -2,6 +2,8 @@ use calimero_sdk::env;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
+use crate::error::NearLibError;
+
 pub(crate) struct Client {
```
miraclx: so you no longer need this

crates/sdk/libs/near/src/lib.rs
```diff
@@ -25,142 +31,10 @@ impl Client {
...
-        match response.data {
-            Ok(r) => {
-                if let QueryResponseKind::CallResult(cr) = r.kind {
-                    return Ok(cr);
-                }
-                return Err("Unexpected response returned.".to_string());
-            }
-            Err(e) => Err(format!("Error: {}, Code: {}", e.message, e.code,)),
-        }
+    pub fn call<M>(&self, method: M) -> Result<Response<M::Response, String>, NearLibError>
```
miraclx:
```suggestion
    pub fn call<M>(&self, method: M) -> Result<M::Response, Error<M::Error>>
```

you should be able to encode the whole state into this flattened representation

crates/sdk/libs/near/src/error.rs
```diff
@@ -0,0 +1,13 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum NearLibError {
```
miraclx: please revise the name, `Error` is fine, and it can be exposed as `calimero_sdk_near::Error`

crates/sdk/libs/near/src/lib.rs
```diff
@@ -1,8 +1,7 @@
+use error::NearLibError;
 use jsonrpc::Response;
-use query::{QueryResponseKind, RpcQueryRequest};
-use types::{BlockId, FunctionArgs, StoreKey};
-use views::QueryRequest;
 
+pub mod error;
```
miraclx:
```suggestion
mod error
```

and then:

```rust
pub use error::Error;
```

crates/sdk/libs/near/Cargo.toml
its-saeed: I needed `serde` feature, I'm not sure why it's not like `near-account-id = { workspace = true, features = ["serde"]` in the PR, but it is in my machine (It works on my machine! :D)

crates/sdk/libs/near/src/error.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use crate::jsonrpc::RpcError;
+
+#[derive(Debug, Error)]
+pub enum Error<R> {
+    #[error(transparent)]
+    JsonError(#[from] serde_json::Error),
+
+    #[error(transparent)]
+    IoError(#[from] std::io::Error),
```
miraclx: Looks like this will never happen?

crates/sdk/libs/near/src/query.rs
```diff
@@ -0,0 +1,47 @@
...
+impl RpcMethod for RpcQueryRequest {
+    type Response = RpcQueryResponse;
+    type Error = String;
+
+    fn method_name(&self) -> &str {
+        "query"
+    }
+
+    fn params(&self) -> Result<serde_json::Value, std::io::Error> {
+        Ok(json!(self))
```
miraclx: yep, just take out the error path

crates/sdk/libs/near/src/jsonrpc.rs
```diff
@@ -0,0 +1,74 @@
...
+            jsonrpc: "2.0",
+            id: *self.id.borrow(),
+            method: method.method_name(),
+            params: method.params()?,
+        })?;
+
+        let response = unsafe { env::ext::fetch(&self.url, "POST", &headers, &body) }
+            .map_err(Error::FetchError)?;
+
+        serde_json::from_slice::<Response<M::Response, M::Error>>(&response)?
```
miraclx: rust should be able to infer the types here, 

```suggestion
        serde_json::from_slice::<Response<_, _>>(&response)?
```

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,111 @@
...
+        allowance: Option<Balance>,
+        receiver_id: Box<str>,
+        method_names: Box<[Box<str>]>,
+    },
+    FullAccess,
+}
+
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct AccessKeyList {
+    pub keys: Vec<AccessKeyInfoView>,
```
miraclx:
```suggestion
    pub keys: Box<[AccessKeyInfoView]>,
```

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,111 @@
...
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct StateItem {
+    pub key: StoreKey,
+    pub value: StoreValue,
+}
+
+#[serde_as]
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct ViewStateResult {
+    pub values: Vec<StateItem>,
```
miraclx:
```suggestion
    pub values: Box<[StateItem]>,
```

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,111 @@
...
+    ViewState {
+        account_id: AccountId,
+        #[serde(rename = "prefix_base64")]
+        prefix: StoreKey,
+        #[serde(default)]
+        include_proof: bool,
+    },
+    ViewAccessKey {
+        account_id: AccountId,
+        public_key: String,
```
miraclx:
```suggestion
        public_key: Box<str>,
```

crates/sdk/libs/near/src/types.rs
```diff
@@ -0,0 +1,31 @@
...
+#[serde(untagged)]
+pub enum BlockId {
+    Height(BlockHeight),
+    Hash(BlockHash),
+}
+
+#[serde_as]
+#[derive(serde::Deserialize, Clone, Debug)]
+#[serde(transparent)]
+pub struct StoreValue(#[serde_as(as = "Base64")] pub Vec<u8>);
```
miraclx:
```suggestion
pub struct StoreValue(#[serde_as(as = "Base64")] pub Box<[u8]>);
```

crates/sdk/libs/near/src/types.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[serde_as]
+#[derive(serde::Serialize, serde::Deserialize, Clone, Debug)]
+#[serde(transparent)]
+pub struct StoreKey(#[serde_as(as = "Base64")] pub Vec<u8>);
```
miraclx:
```suggestion
pub struct StoreKey(#[serde_as(as = "Base64")] pub Box<[u8]>);
```

crates/sdk/libs/near/src/types.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[serde_as]
+#[derive(serde::Serialize, Clone, Debug)]
+#[serde(transparent)]
+pub struct FunctionArgs(#[serde_as(as = "Base64")] Vec<u8>);
```
miraclx:
```suggestion
pub struct FunctionArgs(#[serde_as(as = "Base64")] Box<[u8]>);
```

miraclx: why aren't these with view?

crates/sdk/libs/near/src/query.rs
```diff
@@ -0,0 +1,47 @@
...
+    ViewCode(ContractCodeView),
+    ViewState(ViewStateResult),
+    AccessKey(AccessKeyView),
+    AccessKeyList(AccessKeyList),
+    CallResult(CallResult),
+}
+
+impl RpcMethod for RpcQueryRequest {
+    type Response = RpcQueryResponse;
+    type Error = String;
```
miraclx: how about `RpcQueryError` - https://docs.rs/near-jsonrpc-client/latest/near_jsonrpc_client/methods/query/enum.RpcQueryError.html?

crates/sdk/libs/near/src/query.rs
```diff
@@ -93,7 +35,7 @@ pub enum QueryResponseKind {
 
 impl RpcMethod for RpcQueryRequest {
     type Response = RpcQueryResponse;
-    type Error = RpcQueryError;
```
miraclx: seems you were on the right path, what happened?

miraclx:
@coderabbitai pause

crates/sdk/libs/near/src/query.rs
its-saeed: Yep. But I thought I was overengeenering.

miraclx: Ah, it's just the right amount.

crates/sdk/libs/near/src/types.rs
its-saeed: The folder structure is based on near-premitives crate.

miraclx: Uh huh.. feel free to organize things in your own way, they would've done what they did for reasons beyond the scope of this PR

Like for example, these structures could be reused when submitting transactions, which we don't do right now.

its-saeed: ok ok :)

xilosada:
@its-saeed Please add a test plan and a summary. Can you prepare the docs? This must have public docs @MatejVukosav

crates/sdk/libs/near/src/types.rs
its-saeed: Currently, everything is `query` related so that we can move everything to `query` module.

crates/sdk/libs/near/src/error.rs
```diff
@@ -1,18 +1,39 @@
...
-    #[error(transparent)]
-    IoError(#[from] std::io::Error),
-
     #[error("Failed to fetch: {0}")]
     FetchError(String),
 
-    #[error(transparent)]
-    ServerError(#[from] RpcError<R>),
+    #[error("Server error {0}")]
+    ServerError(RpcError<R>),
```
miraclx: looks like this requires `RpcError<R>: fmt::Display`, is that implemented?

miraclx (commented):
lgtm, now let's add an example that tests the library for basic things

like your test case in the description, but it's own application, you should receive specific variables from the inputs, and not hard coded

apps/only-peers/Cargo.toml
```diff
@@ -11,3 +11,4 @@ crate-type = ["cdylib"]
 
 [dependencies]
 calimero-sdk = { path = "../../crates/sdk" }
+calimero-sdk-near = { path = "../../crates/sdk/libs/near" }
```
miraclx: this shouldn't be here btw

crates/sdk/libs/near/src/views.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub value: StoreValue,
+}
+
+#[serde_as]
+#[derive(serde::Deserialize, Debug, Clone)]
+pub struct ViewStateResult {
+    pub values: Box<[StateItem]>,
+    #[serde_as(as = "Vec<Base64>")]
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub proof: Vec<Arc<[u8]>>,
```
miraclx:
```suggestion
    pub proof: Box<[Arc<[u8]>]>,
```

its-saeed:
> lgtm, now let's add an example that tests the library for basic things
> 
> like your test case in the description, but it's own application, you should receive specific variables from the inputs, and not hard coded

Should it be in the `apps` folder?

crates/runtime/src/logic.rs
xilosada: check grammar
```suggestion
// Safety: The `fetch` function cannot be directly called by applications.
// Therefore, the headers are generated exclusively by our code, ensuring
// that it is safe to deserialize them.
```

xilosada (commented):
@miraclx let's keep current scope. If the code is ok please approve the PR. @its-saeed review the comments you are adding in the codebase.

crates/runtime/src/logic.rs
miraclx: I think it's fine for now, we'll be revising some details after this - https://github.com/calimero-network/core/issues/401

miraclx (approved):
approving to unblock, let's get this merged 🎉

PR #402, opened 2024-06-21
feat: add nodes script

frdomovic:
# feat: add nodes script

## Summary
Shell script for starting nodes and has 2 options:
1. delete-and-start - deletes the data folder and recreates all the files : coordinator, node1, node2
2. start - starts the nodes + coordinator with existing files in data folder.

Video: 

https://github.com/calimero-network/core/assets/93442516/6bae697f-2715-40b6-9efe-a2b61166b34f

## Summary by CodeRabbit

- **New Features**
    - Added functionality to manage the initialization and running of multiple nodes using Cargo.
    - Options to delete and start nodes or simply start them were included.

### DIFF
--- a/start-nodes.sh
+++ b/start-nodes.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+show_usage() {
+    echo "Usage: $0 delete-and-start || start"
+    exit 1
+}
+
+current_dir=$(pwd)
+
+if [ $# -ne 1 ]; then
+    show_usage
+fi
+
+case "$1" in
+    delete-and-start)
+        rm -rf "$current_dir/data"
+        mkdir -p "$current_dir/data"
+        osascript <<EOF
+tell application "Terminal"
+    activate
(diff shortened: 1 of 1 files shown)


start-nodes.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+show_usage() {
+    echo "Usage: $0 delete-and-start || start"
```
its-saeed, line 4: Shouldn't we send this error to STDERR? 🤔

start-nodes.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
its-saeed, line 1: what about having `set -e` at the beginning?

xilosada:
@frdomovic I can see this as functionality provided by dev-tool as. It can be added to our app template as developer dependencies and configured during the installation. IMO this should not be in this repo. Cc: @MatejVukosav

xilosada:
Also I think you don't need osascript for any of this

PR #405, opened 2024-06-24
feat(cli): add base commands

petarjuki7:
# CLI init, config, run

## Summary

Our meroctl CLI tool, with run, config and init features for now.

## Test plan

### Set env variable CALIMERO_HOME
example: export CALIMERO_HOME=/Users/<user_name>/Documents/core/data 

### Initialize a new node
$ cargo run -p meroctl -- --node-name node1 init

### Configure an existing node
$ cargo run -p meroctl -- --node-name node1 config --server-host 143.34.182.202 --server-port 3000

### Run a node as a peer
$ cargo run -p meroctl -- --node-name node1 run

### Run a node as a coordinator
$ cargo run -p meroctl -- --node-name node1 run --node-type coordinator

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a command-line interface (CLI) for the `meroctl` tool with commands for initializing, configuring, and running nodes.
  - Added functionality to manage and save node configurations, including network settings and mDNS discovery.
  - Implemented default path settings for configuration files.
  
- **Enhancements**
  - Improved logging and application initialization with `tracing_subscriber` and `eyre`.
  - Added extensive dependency support to ensure robust and extendable functionality.

- **Configuration**
  - Added new configuration structures and methods to handle various settings related to nodes, networks, servers, and storage.


crates/meroctl/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+libp2p.workspace = true
+dirs.workspace = true
+eyre.workspace = true
+tokio = { workspace = true, features = ["io-std", "macros"] }
+multiaddr.workspace = true
+serde = { workspace = true, features = ["derive"] }
+sha2.workspace = true
+toml.workspace = true
+tracing.workspace = true
+hex.workspace = true
```
its-saeed: They should be in alphabetical order.

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,214 @@
...
+    Ipfs,
+}
+
+impl ConfigCommand {
+    pub fn run(mut self, root_args: cli::RootArgs) -> eyre::Result<()> {
+        let path = root_args.home.join(&root_args.node_name);
+
+        let matches = cli::RootCommand::command().get_matches();
+
+        let setup_matches = matches.subcommand_matches("config").unwrap();
```
its-saeed: Try not to `unwrap`

crates/meroctl/src/cli/init.rs
```diff
@@ -0,0 +1,164 @@
...
+    Ipfs,
+}
+
+impl InitCommand {
+    pub fn run(self, root_args: cli::RootArgs) -> eyre::Result<()> {
+        let mdns = self.mdns && !self.no_mdns;
+
+        let path = root_args.home.join(root_args.node_name);
+
+        println!("{}", path);
```
its-saeed: I think having a raw print is discouraged

crates/meroctl/src/cli/init.rs
```diff
@@ -0,0 +1,164 @@
...
+                "/{}/{}",
+                match host {
+                    std::net::IpAddr::V4(_) => "ip4",
+                    std::net::IpAddr::V6(_) => "ip6",
+                },
+                host,
+            );
+            listen.push(format!("{}/tcp/{}", host, self.swarm_port).parse()?);
+            listen.push(format!("{}/udp/{}/quic-v1", host, self.swarm_port).parse()?);
+        }
```
its-saeed: Can't we do it with `map` to keep the listen immutable?

crates/meroctl/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+    let command = cli::RootCommand::parse();
+
+    command.run().await
+}
+
+fn setup() -> eyre::Result<()> {
+    tracing_subscriber::registry()
+        .with(EnvFilter::builder().parse(format!(
+            "info,{}",
+            // "debug,libp2p_core=warn,libp2p_gossipsub=warn,{}",
```
its-saeed: Uncomment it or remove it :D

crates/meroctl/src/cli/init.rs
petarjuki7: I can do something like:
```
let listen: Vec<Multiaddr> = self.swarm_host
    .iter()
    .flat_map(|host| {
        let host_str = format!(
            "/{}/{}",
            match host {
                std::net::IpAddr::V4(_) => "ip4",
                std::net::IpAddr::V6(_) => "ip6",
            },
            host
        );
        vec![
            format!("{}/tcp/{}", host_str, self.swarm_port).parse().ok(),
            format!("{}/udp/{}/quic-v1", host_str, self.swarm_port).parse().ok(),
        ]
    })
    .flatten()
    .collect();
```
but I think the original version is more readable.
What do you think?

crates/meroctl/src/cli.rs
```diff
@@ -0,0 +1,45 @@
+use clap::{Parser, Subcommand};
+
+use crate::config as struct_config;
+
+mod config;
+mod init;
+mod run;
+#[derive(Debug, Parser)]
```
```suggestion
mod run;

#[derive(Debug, Parser)]
```

crates/meroctl/src/config.rs
miraclx, line 1: `config_file.rs`

crates/meroctl/src/cli.rs
```diff
@@ -0,0 +1,45 @@
...
+pub enum SubCommands {
+    Init(init::InitCommand),
+    Config(config::ConfigCommand),
+    Run(run::RunCommand),
+}
+
+#[derive(Debug, Parser)]
+pub struct RootArgs {
+    /// Directory for config and data
+    #[arg(long, value_name = "PATH", default_value_t = struct_config::default_chat_dir())]
```
```suggestion
    #[arg(long, value_name = "PATH", default_value_t = defaults::chat_dir())]
```

if you're going to have more than one default value, might make sense to collect them into a module

crates/meroctl/src/cli.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    #[command(subcommand)]
+    pub action: SubCommands,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum SubCommands {
+    Init(init::InitCommand),
+    Config(config::ConfigCommand),
+    Run(run::RunCommand),
```
miraclx, line 47: what do you think about `meroctl up`?

cc: @fbozic @xilosada @MatejVukosav @chefsale

crates/meroctl/src/cli.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[arg(short, long, value_name = "NAME")]
+    pub node_name: camino::Utf8PathBuf,
+}
+
+impl RootCommand {
+    pub async fn run(self) -> eyre::Result<()> {
+        match self.action {
+            SubCommands::Init(init) => return init.run(self.args),
+            SubCommands::Config(config) => return config.run(self.args),
+            SubCommands::Run(run) => return run.run(self.args).await,
```
```suggestion
            SubCommands::Init(init) => init.run(self.args),
            SubCommands::Config(config) => config.run(self.args),
            SubCommands::Run(run) => run.run(self.args).await,
```

crates/meroctl/src/cli.rs
chefsale: Dont have stron opinions 😁

xilosada:
Missing test plan 🙏

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,216 @@
...
+        let setup_matches = matches
+            .subcommand_matches("config")
+            .ok_or_else(|| eyre::eyre!("Command not recognized. Expected 'config' subcommand."))?;
+
+        let boot_network_provided = check_if_provided(&setup_matches, "boot_network");
+        let swarm_host_provided = check_if_provided(&setup_matches, "swarm_host");
+        let swarm_port_provided = check_if_provided(&setup_matches, "swarm_port");
+        let server_host_provided = check_if_provided(&setup_matches, "server_host");
+        let server_port_provided = check_if_provided(&setup_matches, "server_port");
+        let mdns_provided = check_if_provided(&setup_matches, "mdns");
```
miraclx: why this pattern? if these values are options, can't you just test if they're `Some(_)`?

reads to me like you can just make the non-option args options, and remove default values

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        let config_new = ConfigFile {
+            identity,
+            store: StoreConfig {
+                path: "data".into(),
+            },
+            application: ApplicationConfig {
+                path: "apps".into(),
+            },
```
miraclx: this will override the file contents, we shouldn't be creating a new config.. we should be mutating the existing one

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,216 @@
...
+            network: NetworkConfig {
+                swarm: SwarmConfig { listen },
+                bootstrap: BootstrapConfig {
+                    nodes: BootstrapNodes { list: boot_nodes },
+                },
+                discovery: DiscoveryConfig {
+                    mdns,
+                    rendezvous: Default::default(),
```
miraclx: same here

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,216 @@
...
+                            .collect(),
+                    },
+                    admin: Some(calimero_server::admin::service::AdminConfig { enabled: true }),
+                    jsonrpc: Some(calimero_server::jsonrpc::JsonRpcConfig { enabled: true }),
+                    websocket: Some(calimero_server::ws::WsConfig { enabled: true }),
+                },
+            },
+        };
+
+        config_new.save(&path)?;
```
miraclx: we should also be using `toml_edit` to preserve user-modifications like comments, whitespace in their config file

https://docs.rs/toml-edit

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        calimero_store::Store::open(&calimero_store::config::StoreConfig {
+            path: path.join(config_new.store.path),
+        })?;
```
miraclx: we no longer need to open.. this is init only

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        info!("Initialized confing for a node in {:?}", path);
```
miraclx: ?

crates/meroctl/src/cli/config.rs
xilosada: is there any project we can use to get some inspiration?

miraclx: I don't think we need one, we just need to wrap each field in an `Option<_>`, except `Vec`s of course, which can default to being empty. The options when not provided will default to `None` which tells us all we need to know here.

petarjuki7: It was a pattern left over from when I was doing some initialisation in init, and rest in config so I needed to know if it default or provided from user.
I agree it's not needed anymore, will change to `Option<_>`

crates/meroctl/src/cli/config.rs
petarjuki7: I agree, but right now it isn't even an option from the cli to change those two settings.
Should I add an option to change those paths?
Or just leave it but in case the user manually changes his config.toml file?

crates/meroctl/src/cli.rs
petarjuki7: Do I leave it as `run` or should I go with `up`?

crates/meroctl/src/cli/config.rs
miraclx: no, we don't need to do that.. just mutate the same object, so you won't have to bother about these

crates/meroctl/src/cli.rs
miraclx: It's in your hands, whatever you pick. We can also use aliases so it's not one or the other.

crates/meroctl/src/cli/config.rs
```diff
@@ -58,44 +56,46 @@ pub enum BootstrapNetwork {
...
-                _ => None,
-            });
-
-            let ipv6_host = self.swarm_host.iter().find_map(|ip| match ip {
-                IpAddr::V6(v6) => Some(*v6),
-                _ => None,
-            });
-            for addr in config.network.swarm.listen.iter() {
+            let swarm_table = doc["swarm"].as_table_mut().unwrap();
+            let listen_array = swarm_table["listen"].as_array_mut().unwrap();
```
petarjuki7: I `unwrap()` here because if the config.toml is created through the cli, the swarm table and listen array are guaranteed.
But, the user could manually delete those tables/arrays which would cause a panic.
Should I cover those cases? Open to suggestions @miraclx

crates/meroctl/src/cli/config.rs
```diff
@@ -115,28 +115,20 @@ impl ConfigCommand {
...
+                *item = Value::from(new_addr.to_string());
+            }
         }
 
-        // Update server host and/or port if provided
+        // Update server listen addresses
         if !self.server_host.is_empty() || self.server_port.is_some() {
-            let mut new_listen = Vec::new();
+            let server_table = doc["server"].as_table_mut().unwrap();
+            let listen_array = server_table["listen"].as_array_mut().unwrap();
```
petarjuki7: Same here as above.

crates/meroctl/src/cli/config.rs
miraclx: how about `unwrap_or_default`?

crates/meroctl/src/cli.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use crate::defaults;
+
+mod config;
+mod init;
+mod run;
+
+pub const EXAMPLES: &str = r#"
+  # Initialize a new node
+  $ meroctl --home data/ --node-name node1 init
```
miraclx, line 13:
```suggestion
  $ meroctl --home data/node1 init
```

are these different?

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,185 @@
...
+
+            for item in listen_array.iter_mut() {
+                let addr: Multiaddr = item.as_str().unwrap().parse().unwrap();
+                let mut new_addr = Multiaddr::empty();
+
+                for protocol in addr.iter() {
+                    match protocol {
+                        Protocol::Ip4(_) if ipv4_host.is_some() => {
+                            new_addr.push(Protocol::Ip4(ipv4_host.unwrap()));
+                        }
```
miraclx: match on the case of interest and the option

```suggestion
                    match (protocol, ipv4_host, ipv6_host, swarm_port) {
                        (Protocol::Ip4(_), Some(ipv4_host), _, _) => {
                            new_addr.push(Protocol::Ip4(ipv4_host));
                        }
```

miraclx: what if I want to add a new address, and not replace the existing one?

crates/meroctl/src/cli/config.rs
miraclx, line 1: tbh, I'm not a fan of this structured approach to config mutation

there's too many logical questions that we cannot safely assume answers for

I still think "simplifying" it would be incredibly valuable

https://github.com/calimero-network/core/blob/422025a53a9f72c83d965833e696cb3459c7ffb4/crates/node/src/cli/init.rs#L16-L19

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        // Update boot nodes if provided
+        if !self.boot_nodes.is_empty() {
+            let bootstrap_table = doc["bootstrap"].as_table_mut().unwrap();
+            let list_array = bootstrap_table["nodes"].as_array_mut().unwrap();
+            list_array.clear();
+            for node in self.boot_nodes.iter() {
+                list_array.push(node.to_string());
+            }
```
miraclx, line 166: like here.. what if I want to keep the existing boot nodes, but add a new one?

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,185 @@
...
+                BootstrapNetwork::CalimeroDev => BootstrapNodes::calimero_dev().list,
+                BootstrapNetwork::Ipfs => BootstrapNodes::ipfs().list,
+            };
+            for node in new_nodes.iter() {
+                list_array.push(node.to_string());
+            }
+        }
+
+        // Update mDNS setting if provided
+        if self.mdns != self.no_mdns {
```
miraclx: won't this always evaluate to `true`

miraclx (commented):
rest LGTM, we should really think about `config`'s behaviour especially since it's a potentially destructive action, you won't want unintentional loss of data from it doing something you didn't expect

and finally, when do you plan to remove the existing CLI code from `calimero-node`

crates/meroctl/src/cli/config.rs
miraclx: (doesn't have to happen in this PR)

crates/meroctl/src/cli.rs
petarjuki7: My logic was that it's easier to remember just the node name, and not the whole path.
That's why the user should export `CALIMERO_HOME` as an env variable, so he doesn't have to input `--home` eveytime, which is just path to a folder where the node is stored.
The node name is then the name of the node in that directory (the one provided in` --home` or env variable)
I can change the pattern if needed

crates/meroctl/src/cli/config.rs
petarjuki7: I agree, my logic was:
- lets say the user initialized with calimero-dev boot nodes
- now he wants to switch to ipfs boot nodes

Should I just append ipfs boot nodes to already existing ones from calimero-dev?
To me it made sense that it overwrites the existing config.
Should I add an append flag or something similar?
Not sure about the cleanest way to handle the situation.

petarjuki7:
> rest LGTM, we should really think about `config`'s behaviour especially since it's a potentially destructive action, you won't want unintentional loss of data from it doing something you didn't expect
> 
> and finally, when do you plan to remove the existing CLI code from `calimero-node`

My plan was to remove the cli in a separate PR right after this one, so its more independant

crates/meroctl/src/cli/config.rs
miraclx: > Should I add an append flag or something similar?

that just further complicates it, let's just give the user a flexible interface to mutate whatever data they want inside the config (without being opinionated about it)

miraclx:
> My plan was to remove the cli in a separate PR right after this one

reviewing will be much easier when we can see a diff rather than blocks of new code that is a duplication of old code

so we don't have to re-review thing that are already in master

crates/meroctl/src/cli.rs
miraclx: I think `node-name` is useful for aliasing nodes that `meroctl` is a client of.. not nodes that `meroctl` "is".

By that, I mean:

```console
$ meroctl --home data/node1 init
$ meroctl --home data/node1 config ..
$ meroctl --home data/node1 run
```

```console
$ meroctl --home data/node2 init
$ meroctl --home data/node2 config ..
$ meroctl --home data/node2 run
```

```console
$ meroctl --home data/node3 init
$ meroctl --home data/node3 config ..
$ meroctl --home data/node3 run
```

and for remote-control commands:

1. Connect to the node running at `data/node1`, and list it's contexts

```console
$ meroctl --home data/node1 context ls
```

2. Connect to the node running at `http://localhost:2528`, and list it's contexts

```console
$ meroctl --api http://localhost:2528 context ls
```

3. Connect to the node aliased `mein-node`, and list it's contexts

```console
$ meroctl connect local-node --path data/node1
$ meroctl --node local-node context ls
```

```console
$ meroctl connect remote-node --api https://my-remote-node:2528
$ meroctl --node remote-node context ls
```

crates/meroctl/src/cli/config.rs
```diff
@@ -170,10 +169,9 @@ impl ConfigCommand {
         }
 
         // Update mDNS setting if provided
-        if self.mdns != self.no_mdns {
-            let discovery_table = doc["discovery"].as_table_mut().unwrap();
-            discovery_table["mdns"] = toml_edit::value(self.mdns && !self.no_mdns);
-        }
+        let mdns_enabled = if self.no_mdns { false } else { self.mdns };
+        let discovery_table = doc["discovery"].as_table_mut().unwrap();
+        discovery_table["mdns"] = toml_edit::value(mdns_enabled);
```
miraclx: aren't these the same thing?

```suggestion
        doc["discovery"]["mdns"] = toml_edit::value(self.mdns && !self.no_mdns);
```

miraclx: yet, it doesn't do what you want it to. And the problem is the `mdns` options aren't `Option`-s, so they'll always be defined, implying they'll always be overwritten.

.github/workflows/calimero_node_linux.yml
miraclx, line 1: Not sure I understand this, why are we building and bundling the lib instead of the binary?

What I expect to see is.. build the `meroctl` binary since the `calimero-node` binary has now been deleted.

crates/meroctl/src/cli/config.rs
```diff
@@ -169,9 +169,14 @@ impl ConfigCommand {
...
-        let mdns_enabled = if self.no_mdns { false } else { self.mdns };
         let discovery_table = doc["discovery"].as_table_mut().unwrap();
-        discovery_table["mdns"] = toml_edit::value(mdns_enabled);
+        let mdns_current = &discovery_table["mdns"]
+            .as_bool()
+            .ok_or(eyre::eyre!("mdns value wasn't set as a bool"))?;
+        let mdns = self.mdns && !self.no_mdns;
+        if *mdns_current != mdns {
+            discovery_table["mdns"] = toml_edit::value(mdns);
+        }
```
miraclx: if I have my `mdns` setting as `false`, this will always override my `mdns` settings since `self.mdns` defaults to `true`

.github/workflows/calimero_node_linux.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and Upload Library for Linux
+name: Build and Upload MeroCtl Binary for Linux
```
miraclx: `MeroCtl` here

.github/workflows/calimero_node_macos.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and Upload Library for MacOS
+name: Build and Upload Meroctl for MacOS
```
miraclx, line 1: `Meroctl` here, can we pick a style?

crates/meroctl/src/cli/context/common.rs
```diff
@@ -31,8 +31,7 @@ pub(crate) fn multiaddr_to_url(
 
     let mut url = Url::parse(&format!("{}://{}:{}", scheme, ip, port))?;
 
-    if let Some(api_path) = api_path {
-        url.set_path(&api_path);
-    }
+    url.set_path(&api_path);
```
miraclx, line 24: looks like `api_path` doesn't need an owned `String` either, should be fine as a `&str`

crates/meroctl/src/cli.rs
chefsale: I preffer we add both options then with aliases.

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,188 @@
...
+            }
+        });
+
+        let ipv6_host = self.swarm_host.iter().find_map(|ip| {
+            if let IpAddr::V6(v6) = ip {
+                Some(*v6)
+            } else {
+                None
+            }
+        });
```
miraclx: iterate once, select for both, extract each on each iteration

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,188 @@
...
+        if !self.boot_nodes.is_empty() {
+            let bootstrap_table = doc["bootstrap"].as_table_mut().unwrap();
+            let list_array = bootstrap_table["nodes"].as_array_mut().unwrap();
+            list_array.clear();
+            for node in self.boot_nodes.iter() {
+                list_array.push(node.to_string());
+            }
+        } else if let Some(network) = self.boot_network {
+            let bootstrap_table = doc["bootstrap"].as_table_mut().unwrap();
+            let list_array = bootstrap_table["nodes"].as_array_mut().unwrap();
```
miraclx: also, I don't think you need to break these into separate statements with individual `unwrap`s

as highlighted here - https://github.com/calimero-network/core/pull/405#discussion_r1658368702

`doc["bootstrap"]["nodes"]` accesses the same item

crates/meroctl/src/cli/context/common.rs
```diff
@@ -0,0 +1,37 @@
...
+        .ok_or_else(|| eyre::eyre!("No TCP port found in Multiaddr"))?;
+
+    let scheme = multiaddr
+        .iter()
+        .find_map(|p| match p {
+            libp2p::multiaddr::Protocol::Http => Some("http"),
+            libp2p::multiaddr::Protocol::Https => Some("https"),
+            _ => None,
+        })
+        .unwrap_or("http");
```
miraclx: iterate once, extract all three items on each iteration that matches

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use calimero_network::config::BootstrapNodes;
+use clap::{Parser, ValueEnum};
+use multiaddr::{Multiaddr, Protocol};
+use toml_edit::{DocumentMut, Value};
+use tracing::info;
+
+use crate::cli;
+
+/// Initialize node configuration
```
miraclx: incorrect description

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,193 @@
...
+struct ListApplicationsResponse {
+    data: ApplicationListResult,
+}
+
+#[derive(Debug, Parser)]
+#[clap(group(
+    ArgGroup::new("dev_args")
+        .multiple(true)
+        .requires_all(&["dev", "path", "version"])
+))]
```
miraclx: also, why not use the short syntax? with which `multiple = true` [is implied](https://docs.rs/clap/latest/clap/_derive/index.html#arggroup-attributes)

```suggestion
#[group(requires_all(&["dev", "path", "version"]))]
```

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,193 @@
...
+pub struct CreateCommand {
+    /// The application ID to attach to the context
+    #[clap(
+        long,
+        short = 'a',
+        default_value = "",
+        exclusive = true,
+        value_name = "APP_ID"
+    )]
+    application_id: String,
```
miraclx: `default` seems wrong here

CleanShot 2024-07-01 at 14 04 59@2x

crates/server/src/admin/service.rs
```diff
@@ -99,6 +99,19 @@ pub(crate) fn setup(
...
+            post(install_dev_application_handler),
+        )
+        .route(
+            "/contexts-dev",
+            post(handlers::context::create_context_handler),
+        )
+        .route(
+            "/contexts-dev",
+            get(handlers::context::get_contexts_handler),
+        )
```
```suggestion
        .route(
            "/contexts-dev",
            get(handlers::context::get_contexts_handler)
                .post(handlers::context::create_context_handler),
        )
```

crates/meroctl/src/cli.rs
petarjuki7: > I preffer we add both options then with aliases.

Added!

crates/meroctl/src/cli/context/create.rs
miraclx: not resolved

petarjuki7: Ah, didn't remove it from application-id, fixed now

crates/application/src/lib.rs
```diff
@@ -55,6 +59,28 @@ impl ApplicationManager {
...
+        &self,
+        application_id: calimero_primitives::application::ApplicationId,
+        version: &semver::Version,
+        path: Option<Utf8PathBuf>,
+    ) -> eyre::Result<()> {
+        self.link_release(
+            &application_id,
+            &self.application_dir,
+            version,
+            &path.unwrap(),
```
miraclx: can we not?

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,172 @@
...
+                    }
+                }
+
+                *item = Value::from(new_addr.to_string());
+            }
+        }
+
+        // Update server listen addresses
+        if !self.server_host.is_empty() || self.server_port.is_some() {
+            let listen_array = doc["server"]["listen"].as_array_mut().unwrap();
```
miraclx: all these `unwrap` points should have proper error handling

petarjuki7: Sure, I think `unwrap_or_default()` doesn't really make sense because there is no "default" to assign it to, so I'm going to put `ok_or()`

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,187 @@
...
+            for node in new_nodes.iter() {
+                list_array.push(node.to_string());
+            }
+        }
+
+        // Update mDNS setting if provided
+        self.mdns
+            .map(|mdns| doc["discovery"]["mdns"] = toml_edit::value(mdns));
+        self.no_mdns
+            .map(|no_mdns| doc["discovery"]["mdns"] = toml_edit::value(no_mdns));
```
miraclx: https://github.com/calimero-network/core/pull/438

crates/application/src/lib.rs
```diff
@@ -187,6 +208,29 @@ impl ApplicationManager {
...
+        application_id: &calimero_primitives::application::ApplicationId,
+        dir: &camino::Utf8Path,
+        version: &semver::Version,
+        link_path: &camino::Utf8Path,
+    ) -> eyre::Result<()> {
+        let base_path = format!("{}/{}/{}", dir, application_id, version);
+        fs::create_dir_all(&base_path)?;
+
+        let file_path = format!("{}/binary.wasm", base_path);
+        info!("{}", file_path);
```
miraclx: why are we logging this with no further information?

crates/application/src/lib.rs
```diff
@@ -55,6 +59,23 @@ impl ApplicationManager {
...
+    ) -> eyre::Result<()> {
+        self.link_release(&application_id, &self.application_dir, version, &path)?;
+
+        let topic_hash = self
+            .network_client
+            .subscribe(calimero_network::types::IdentTopic::new(application_id))
+            .await?;
+
+        info!(%topic_hash, "Subscribed to network topic");
+        return Ok(());
```
miraclx:
```suggestion
        Ok(())
```

crates/meroctl/src/cli/config.rs
```diff
@@ -0,0 +1,193 @@
...
+            .join(&root_args.node_name)
+            .join("config.toml");
+
+        // Load the existing TOML file
+        let toml_str =
+            fs::read_to_string(&path).map_err(|_| eyre!("Node must be initialized first."))?;
+        let mut doc = toml_str.parse::<DocumentMut>()?;
+
+        if self.print {
+            println!("{}", doc.to_string());
```
miraclx: this will call `Display::fmt` twice, you should instead delegate that call to `println`

```suggestion
            println!("{}", doc);
```

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    /// Version of the application
+    #[clap(
+        short,
+        long,
+        group = "dev_args",
+        default_value = "0.0.0",
+        value_name = "VERSION"
+    )]
+    version: Version,
```
miraclx, line 69: `version` is only useful when `--dev` is provided, the definition should reflect that relationship

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[clap(long, short = 'a', exclusive = true, value_name = "APP_ID")]
+    application_id: String,
+
+    /// Enable dev mode
+    #[clap(long, group = "dev_args")]
+    dev: bool,
+
+    /// Path to use in dev mode
+    #[clap(short, long, group = "dev_args", value_name = "PATH")]
+    path: Utf8PathBuf,
```
miraclx, line 60: same with `path`. you should play with group semantics

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,177 @@
...
+    if response.status().is_success() {
+        let api_response: ListApplicationsResponse = response.json().await?;
+        let app_list = api_response.data.apps;
+        let x = app_list
+            .iter()
+            .any(|app| app.id.as_ref() == *application_id);
+        if x {
+            return Ok(());
+        } else {
+            eyre::bail!("The app with the id {} is not installed\nTo create a context, install an app first", application_id)
```
miraclx: this is a weird pattern.. why not return `Result<bool>`?

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        if self.dev {
+            return link_local_app(multiaddr, self.path, self.version).await;
+        }
+        create_context(multiaddr, self.application_id).await
+    }
+}
+
+async fn create_context(base_multiaddr: &Multiaddr, application_id: String) -> eyre::Result<()> {
+    app_installed(&base_multiaddr, &application_id).await?;
```
miraclx:
```rust
if !application_installed(..).await? {
    eyre::bail!
}
```

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,177 @@
...
+    let client = Client::new();
+    let response = client.get(url).send().await?;
+    if response.status().is_success() {
+        let api_response: ListApplicationsResponse = response.json().await?;
+        let app_list = api_response.data.apps;
+        let x = app_list
+            .iter()
+            .any(|app| app.id.as_ref() == *application_id);
+        if x {
+            return Ok(());
```
miraclx: looks like all paths in this function short-circuit so it never gets to the end, refactor this so `Ok(())` ends up at the bottom of the function

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,177 @@
...
+            status,
+            error_text
+        );
+    }
+    Ok(())
+}
+
+async fn app_installed(base_multiaddr: &Multiaddr, application_id: &String) -> eyre::Result<()> {
+    let url = multiaddr_to_url(base_multiaddr, "admin-api/applications-dev")?;
+    let client = Client::new();
```
miraclx: you should try to reuse `Client` here

https://docs.rs/reqwest/latest/reqwest/struct.Client.html

crates/meroctl/src/cli/init.rs
```diff
@@ -72,17 +68,13 @@ impl InitCommand {
...
-                fs::create_dir_all(&root_args.home)
-            } else {
-                fs::create_dir(&root_args.home)
-            }
-            .wrap_err_with(|| format!("failed to create directory {:?}", root_args.home))?;
-        }
+        let path = root_args.home.join(root_args.node_name);
+
+        fs::create_dir_all(&path)
+            .wrap_err_with(|| format!("failed to create directory {:?}", path))?;
```
miraclx: the idea behind this was to be careful about recursive directory creation, which should only be permitted when explicitly performed by the user

so, unless it's the calimero default directory which we know is defined as `~/.calimero`, we should not be recursively creating nested diretcories.

crates/meroctl/src/cli/run.rs
```diff
@@ -29,21 +29,23 @@ impl From<NodeType> for calimero_node_primitives::NodeType {
 
 impl RunCommand {
     pub async fn run(self, root_args: cli::RootArgs) -> eyre::Result<()> {
-        if !ConfigFile::exists(&root_args.home) {
-            eyre::bail!("chat node is not initialized in {:?}", root_args.home);
+        let path = root_args.home.join(root_args.node_name);
+
+        if !ConfigFile::exists(&path) {
+            eyre::bail!("chat node is not initialized in {:?}", path);
```
miraclx:
```suggestion
            eyre::bail!("Node is not initialized in {:?}", path);
```

miraclx: We also need to unify this error message across commands for consistency

crates/meroctl/src/config_file.rs
```diff
@@ -62,11 +60,15 @@ pub struct ApplicationConfig {
...
-    pub fn exists(dir: &camino::Utf8Path) -> bool {
+    pub(crate) fn exists(dir: &camino::Utf8Path) -> bool {
         dir.join(CONFIG_FILE).is_file()
     }
 
-    pub fn load(dir: &camino::Utf8Path) -> eyre::Result<Self> {
+    pub(crate) fn load(dir: &camino::Utf8Path) -> eyre::Result<Self>
+    where
+        Self: Sized,
+        for<'de> Self: Deserialize<'de>,
```
miraclx: isn't this guaranteed? or am I missing something?

crates/meroctl/src/config_file.rs
```diff
@@ -78,7 +80,10 @@ impl ConfigFile {
         toml::from_str(&content).map_err(Into::into)
     }
 
-    pub fn save(&self, dir: &camino::Utf8Path) -> eyre::Result<()> {
+    pub(crate) fn save(&self, dir: &camino::Utf8Path) -> eyre::Result<()>
+    where
+        Self: Serialize,
```
miraclx: same here

crates/meroctl/src/defaults.rs
```diff
@@ -0,0 +1,10 @@
+pub(crate) const DEFAULT_CALIMERO_HOME: &str = "data";
+
+pub fn default_chat_dir() -> camino::Utf8PathBuf {
```
miraclx: rename this

crates/server-primitives/src/admin.rs
```diff
@@ -8,6 +9,13 @@ pub struct InstallApplicationRequest {
     pub version: semver::Version,
 }
 
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct InstallDevApplicationRequest {
+    pub application: calimero_primitives::application::ApplicationId, // TODO: rename to application_id
```
miraclx: I think you have the liberty to rename it now, since it's just now being introduced

crates/server/src/admin/service.rs
```diff
@@ -209,6 +219,20 @@ async fn install_application_handler(
     }
 }
 
+async fn install_dev_application_handler(
```
MatejVukosav: Extract handler into handlers. Keep service clean

crates/server/src/admin/service.rs
```diff
@@ -99,6 +99,16 @@ pub(crate) fn setup(
             "/add-client-key",
             post(handlers::add_client_key::add_client_key_handler),
         )
+        .route(
+            "/install-dev-application",
```
MatejVukosav: Why do you need all these endpoints?

Also, keep them organized under
/dev/ something restful

crates/server/src/admin/service.rs
```diff
@@ -209,6 +219,20 @@ async fn install_application_handler(
     }
 }
 
+async fn install_dev_application_handler(
+    Extension(state): Extension<Arc<AdminState>>,
+    Json(req): Json<calimero_server_primitives::admin::InstallDevApplicationRequest>,
+) -> impl IntoResponse {
+    match state
+        .application_manager
+        .install_dev_application(req.application_id, &req.version, req.path)
```
MatejVukosav: Does the user need to increase the version every time or he can use the same version but with changed wasm? Like hot reload.

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,196 @@
...
+struct ListApplicationsResponse {
+    data: ApplicationListResult,
+}
+
+#[derive(Debug, Parser)]
+#[clap(group(
+    ArgGroup::new("mode")
+        .required(true)
+        .args(&["application_id", "dev"]),
+))]
```
miraclx: do you need this?

this should suffice, right?

```suggestion
#[group(required = true)]
```

crates/server/src/admin/service.rs
miraclx: Wasm is linked so the next execution will use the new version.

crates/server/src/admin/handlers/install_dev_application.rs
```diff
@@ -0,0 +1,21 @@
+use std::sync::Arc;
+
+use axum::http::StatusCode;
+use axum::response::IntoResponse;
+use axum::{Extension, Json};
+
+use crate::admin::service::{AdminState, ApiResponse};
+
+pub async fn install_dev_application_handler(
```
MatejVukosav: call file applications

crates/meroctl/src/cli/config.rs
fbozic: I agree with Mira. Maybe another alternative (similar to kubectl edit) would be to open config file with the default shell editor and before saving the file do the validations. Docs for the edit: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_edit/

fbozic: I think with both approaches user should be able to achieve the same

miraclx: Totally agree, `kubo` has both too

```console
ipfs help config
USAGE
  ipfs config <key> [<value>] - Get and set IPFS config values.

SYNOPSIS
  ipfs config [--bool] [--json] [--] <key> [<value>]

ARGUMENTS

  <key>     - The key of the config entry (e.g. "Addresses.API").
  [<value>] - The value to set the config entry to.

OPTIONS

  --bool  bool - Set a boolean value.
  --json  bool - Parse stringified JSON.

DESCRIPTION

  'ipfs config' controls configuration variables. It works
  much like 'git config'. The configuration values are stored in a config
  file inside your IPFS repository (IPFS_PATH).

  Examples:

  Get the value of the 'Datastore.Path' key:

    $ ipfs config Datastore.Path

  Set the value of the 'Datastore.Path' key:

    $ ipfs config Datastore.Path ~/.ipfs/datastore

SUBCOMMANDS
  ipfs config edit           - Open the config file for editing in $EDITOR.
  ipfs config profile        - Apply profiles to config.
  ipfs config replace <file> - Replace the config with <file>.
  ipfs config show           - Output config file contents.

  For more information about each command, use:
  'ipfs config <subcmd> --help'
```

crates/server/src/admin/service.rs
MatejVukosav: New version or new wasm?

crates/meroctl/src/cli/config.rs
petarjuki7: Is it fine if I expand on it in another PR?

crates/meroctl/src/cli/context/create.rs
```diff
@@ -31,18 +36,16 @@ struct ListApplicationsResponse {
 }
 
 #[derive(Debug, Parser)]
-#[group(required = true)]
```
miraclx: does this mean `meroctl context create` now works?

by "works" I mean, `clap` won't report an issue

miraclx (approved):
provided other comments are iterated upon in follow-up PRs, this so far, LGTM

thanks, @petarjuki7 ✨

crates/server/src/admin/service.rs
miraclx: yeah, new wasm.. new version of the wasm, new update..

crates/meroctl/src/cli/context/create.rs
petarjuki7: Yes, I had some weird issue that I couldn't figure out, so I just put application_id as exclusive so it keeps the same functionality and it fixes the issue.

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Enable dev mode
+    #[clap(long)]
+    dev: bool,
+
+    /// Path to use in dev mode
+    #[clap(
+        short,
+        long,
+        help = "Path to use in dev mode (requires --dev and --version)",
+        required = true
```
miraclx: I think you can drop this now

miraclx:
PS: we need to update and move this file - https://github.com/calimero-network/core/blob/0f821050a556573197bdba655c4e6c58d5d99f6d/crates/node/gen_localnet_configs.sh

crates/meroctl/src/defaults.rs
```diff
@@ -0,0 +1,10 @@
+pub(crate) const DEFAULT_CALIMERO_HOME: &str = "data";
```
miraclx, line 1: why `data` btw?

`meroctl init` now creates a home dir as `~/data` instead of `~/.calimero`

PR #410, opened 2024-06-25
chore(store-revamp): fix `calimero-server` compilation

miraclx:
Opening this for visibility

Careful review is required here, as parts of the API have now changed

Notably, `application_id` use has been replaced with `context_id`, everywhere except the application installation flow.

This will also require patching the client side.

### DIFF
--- a/crates/server-primitives/src/admin.rs
+++ b/crates/server-primitives/src/admin.rs
@@ -19,7 +19,7 @@ pub struct AddPublicKeyRequest {
     pub wallet_signature: String,
     pub payload: Payload,
     pub wallet_metadata: WalletMetadata,
-    pub context_id: String,
+    pub context_id: calimero_primitives::context::ContextId,
 }
 
 #[derive(Debug, Deserialize)]
@@ -32,8 +32,8 @@ pub struct Payload {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SignatureMessage {
+    pub context_id: calimero_primitives::context::ContextId,
     pub nonce: String,
-    pub application_id: String,
     pub timestamp: i64,
     pub node_signature: String,
     pub message: String,
@@ -74,7 +74,7 @@ pub struct IntermediateAddPublicKeyRequest {
     pub wallet_signature: String,
     pub payload: IntermediatePayload,
     pub wallet_metadata: WalletMetadata, // Reuse WalletMetadata as it fits the intermediate step
-    pub context_id: String,
+    pub context_id: calimero_primitives::context::ContextId,
 }
 
 #[derive(Debug, Deserialize)]
@@ -96,7 +96,7 @@ pub struct NodeChallenge {
 #[serde(rename_all = "camelCase")]
 pub struct NodeChallengeMessage {
     pub nonce: String,
-    pub application_id: String, //optional if challenge is used on admin level
+    pub context_id: Option<calimero_primitives::context::ContextId>,
     pub timestamp: i64,
 }
 
--- a/crates/server-primitives/src/jsonrpc.rs
+++ b/crates/server-primitives/src/jsonrpc.rs
@@ -98,7 +98,7 @@ pub enum ServerResponseError {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct QueryRequest {
-    pub application_id: calimero_primitives::application::ApplicationId,
+    pub context_id: calimero_primitives::context::ContextId,
     pub method: String,
     pub args_json: serde_json::Value,
 }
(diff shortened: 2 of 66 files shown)


crates/server/src/admin/storage/client_keys.rs
```diff
@@ -1,4 +1,4 @@
-use calimero_primitives::identity::ClientKey;
+use calimero_primitives::{context::ContextId, identity::ClientKey};
```
miraclx, line 1: I'll run `cargo +nightly fmt` on the main PR after everything is merged

crates/server/src/admin/storage/context.rs
```diff
@@ -29,12 +25,16 @@ pub fn delete_context(store: &mut Store, context_id: &str) -> eyre::Result<bool>
...
+    // Ok(did.contexts.into_iter().find(|k| k.id == context_id))
 }
 
 pub fn get_contexts(store: &mut Store) -> eyre::Result<Vec<Context>> {
     let did = get_or_create_did(store)?;
-    Ok(did.contexts)
+
+    todo!("Implement get_contexts")
+    // Ok(did.contexts)
 }
```
miraclx: Moved `Context` out of `Did`, this pending logic depends on the unimplemented logic for reading the context metadata from the db that's beyond the scope of this patch.

crates/server-primitives/src/admin.rs
```diff
@@ -96,7 +96,7 @@ pub struct NodeChallenge {
 #[serde(rename_all = "camelCase")]
 pub struct NodeChallengeMessage {
     pub nonce: String,
-    pub application_id: String, //optional if challenge is used on admin level
+    pub context_id: Option<calimero_primitives::context::ContextId>,
```
miraclx, line 99: is this safe to do now?

frdomovic: Why wouldn't it be safe? Will update [this](https://github.com/calimero-network/core/blob/master/packages/calimero-sdk/src/wallets/NearLogin/NearLogin.tsx#L303) and other parts of sdk to replace application id

crates/server-primitives/src/jsonrpc.rs
```diff
@@ -123,7 +123,7 @@ pub enum QueryError {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct MutateRequest {
-    pub application_id: calimero_primitives::application::ApplicationId,
+    pub context_id: calimero_primitives::context::ContextId,
```
frdomovic, line 126: Just for visibility; only-peers-client will be updated [here](https://github.com/calimero-network/only-peers-client/blob/master/app/src/api/dataSource/ClientApiDataSource.ts#L39)

crates/server/src/admin/handlers/challenge.rs
```diff
@@ -60,15 +61,15 @@ pub async fn request_challenge_handler(
...
+    context_id: Option<ContextId>,
     keypair: &Keypair,
 ) -> Result<NodeChallenge, ApiError> {
     let random_bytes = generate_random_bytes();
     let encoded = STANDARD.encode(&random_bytes);
 
     let node_challenge_message = NodeChallengeMessage {
         nonce: encoded,
-        application_id: application_id.unwrap_or_default(),
+        context_id,
```
frdomovic, line 72: Why was this `application_id.unwrap_or_default()` before? Should it also be `context_id.unwrap_or_default()` because of Option or not?

crates/server-primitives/src/admin.rs
miraclx: Not sure why it was only a comment before

frdomovic (approved):
Looks good to me

crates/server/src/admin/handlers/challenge.rs
miraclx: Connected to https://github.com/calimero-network/core/pull/410#discussion_r1652280855

Figured it's best to retain the option than to assume a default

Just need to ensure the associated behaviours on admin UI & apps are consistent with this change

crates/server-primitives/src/admin.rs
frdomovic: As we don't need application_id if we login from admin dashboard as it doesn't have appid or contextid. Currently we just pass "admin-ui" for this value.

crates/server/src/admin/storage/context.rs
xilosada: Do we have a gh issue to track this?

crates/server/src/ws/unsubscribe.rs
```diff
@@ -13,11 +13,11 @@ async fn handle(
...
 ) -> eyre::Result<UnsubscribeResponse> {
     let mut inner = connection_state.inner.write().await;
-    request.application_ids.iter().for_each(|id| {
+    request.context_id.iter().for_each(|id| {
         inner.subscriptions.remove(id);
     });
 
     Ok(UnsubscribeResponse {
-        application_ids: request.application_ids,
+        context_id: request.context_id,
```
xilosada: What if I want to unsubscribe to multiple contexts?

miraclx: Good catch, the data type didn't change, just the field name. I'll fix that.

crates/server/src/admin/storage/context.rs
miraclx: Not needed, implementing it rn

PR #413, opened 2024-06-25
feat(cli): add context commands

petarjuki7:
# CLI context ls, create, create --dev

## Summary

This PR build un top of #405

In introduces functionality to manage apps and contexts from the cli.

Workflow:

```
$ calimero context ls 
$ calimero context create --application-id APwGT9wuLwjkBoGxRR9thpS89VsbazYd6CpMpyqx4b5J 
  Created BTgDJ5FezfFJCMWzRWScXiwkgfMHX3zX4s3Qum7Gf4pb with application APwGT9wuLwjkBoGxRR9thpS89VsbazYd6CpMpyqx4b5J
$ calimero context create --dev --path ./path/to/binary.wasm --version 1.0.0
  Created BTgDJ5FezfFJCMWzRWScXiwkgfMHX3zX4s3Qum7Gf4pb linked to application ./path/to/binary.wasm
```

## Test plan

### Set env variable CALIMERO_HOME
example: export CALIMERO_HOME=/Users/<user_name>/Documents/core/data 

### List all contexts
$ cargo run -p meroctl --  --node-name node1 context ls

### Create a new context
$ cargo run -p meroctl -- --node-name node1 context create --application-id my-app-id

### Create a new context in dev mode
$ cargo run -p meroctl -- --node-name node1 context create --dev --path /path/to/app --version 1.0.0

## Checklist


crates/meroctl/src/cli/context/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[derive(Debug, Parser)]
+pub struct ContextCommand {
+    #[command(subcommand)]
+    pub subcommand: ContextSubCommands,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum ContextSubCommands {
+    Ls(ls::LsCommand),
```
miraclx: for the sanity of the type system, let's use `List` here, with an alias for `ls`

crates/meroctl/src/cli/context/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+            ContextSubCommands::Ls(ls) => ls.run(args).await,
+            ContextSubCommands::Create(create) => create.run(args).await,
+            ContextSubCommands::Query(query) => query.run(args).await,
+            ContextSubCommands::Mutate(mutate) => mutate.run(args).await,
+            ContextSubCommands::Members(members) => members.run(args).await,
+        }
+    }
+}
+
+pub(crate) fn get_ip(multiaddr: &libp2p::Multiaddr) -> eyre::Result<reqwest::Url> {
```
miraclx: create a `common` module instead?

crates/meroctl/src/cli/context/ls.rs
```diff
@@ -0,0 +1,57 @@
...
+                    }
+                } else {
+                    return Err(eyre!("Request failed with status: {}", response.status()));
+                }
+            } else {
+                return Err(eyre!("Failed to load config file"));
+            }
+        } else {
+            return Err(eyre!("Config file does not exist"));
+        }
```
miraclx: move short short-circuit cases closer to the condition to flatten out the indentation

```rust
if !condition {
    return ..
}

// case is true
```

crates/meroctl/src/cli/context/ls.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+#[derive(Debug, Parser)]
+pub struct LsCommand {}
+
+impl LsCommand {
+    pub async fn run(self, root_args: RootArgs) -> eyre::Result<()> {
+        let path = root_args.home.join(&root_args.node_name);
+        if ConfigFile::exists(&path) {
+            if let Ok(config) = ConfigFile::load(&path) {
```
miraclx: for `if let <PATTERN> = ` cases, rewrite them with `let <PATTERN> = .. else {}`

```rust
let Ok(config) = ConfigFile::load(&path) else {
    return ..
}

config.whatever..
```

crates/meroctl/src/cli/context/ls.rs
```diff
@@ -0,0 +1,57 @@
...
+                    .network
+                    .server
+                    .listen
+                    .first()
+                    .ok_or_else(|| eyre!("No address."))?;
+                let base_url = get_ip(multiaddr)?;
+                let url = format!("{}admin-api/contexts-dev", base_url);
+
+                let client = Client::new();
+                let response = client.get(&url).send().await?;
```
miraclx: Ideally, we abstract over a generic `Transport` that could be HTTP{admin-api} or whatever else, like raw sockets.. but this is fine for now

crates/meroctl/src/cli/context/ls.rs
```diff
@@ -0,0 +1,57 @@
...
+
+                if response.status().is_success() {
+                    let api_response: GetContextsResponse = response.json().await?;
+                    let contexts = api_response.data;
+
+                    println!("Contexts:");
+                    for context in contexts {
+                        println!("App ID: {}", context.application_id);
+                        println!("Context ID: {}", context.id);
+                        println!();
```
miraclx: Ideally, we factor in outputs being consumable by other programs like JSON output, which is worth noting for now, even worth opening a tracking issue.

Non blocking for this PR though

miraclx: why's there an extra new line at the end? personal preference?

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
...
+    #[clap(long, short = 'a')]
+    application_id: Option<String>,
+
+    /// Enable dev mode
+    #[clap(long)]
+    dev: bool,
+
+    /// Path to use in dev mode (required in dev mode)
+    #[clap(short, long, requires = "dev")]
+    path: Option<Utf8PathBuf>,
```
miraclx: what if I specify `--path` and `--application-id` together?

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
...
+            }
+        } else {
+            Err(eyre!("Config file does not exist"))
+        }
+    }
+}
+
+async fn create_context(base_url: Url, application_id: Option<String>) -> eyre::Result<()> {
+    let application_id =
+        application_id.ok_or_else(|| eyre!("Application ID is required for starting context"))?;
```
miraclx: this should be encoded in clap, so the requirements validation is unified by clap

explore `Arg` options on how to specify inter-argument relationships & dependence

https://docs.rs/clap/latest/clap/struct.Arg.html

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+async fn link_local_app(
+    base_url: Url,
+    path: Option<Utf8PathBuf>,
+    version: Option<Version>,
+) -> eyre::Result<()> {
+    let path = path.ok_or_else(|| eyre!("Path is required in dev mode"))?;
+    let version = version.ok_or_else(|| eyre!("Version is required in dev mode"))?;
```
miraclx: same with these

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
...
+        let api_response: ListApplicationsResponse = response.json().await?;
+        let app_list = api_response.data.apps;
+        let x = app_list
+            .iter()
+            .map(|app| app.id.0.clone())
+            .any(|app| app == *application_id);
+        if x {
+            return Ok(());
+        } else {
+            return Err(eyre!("The app with the id {} is not installed\nTo create a context, install an app first", application_id));
```
miraclx: `eyre::bail`?

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
+use calimero_primitives::application::ApplicationId;
+use calimero_primitives::identity::Context;
+use calimero_server_primitives::admin::ApplicationListResult;
+use camino::Utf8PathBuf;
+use clap::Parser;
+use eyre::eyre;
```
miraclx: most cases don't actually require bringing this in

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub async fn run(self, root_args: RootArgs) -> eyre::Result<()> {
+        let path = root_args.home.join(&root_args.node_name);
+        if ConfigFile::exists(&path) {
+            if let Ok(config) = ConfigFile::load(&path) {
+                let multiaddr = config
+                    .network
+                    .server
+                    .listen
+                    .first()
+                    .ok_or_else(|| eyre!("No address."))?;
```
miraclx: this for example, can be rewritten to:

```rust
let Some(multiaddr) = config...... else {
    eyre::bail!("No address")
}
```

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,179 @@
...
+    let url = format!("{}admin-api/applications-dev", base_url);
+    let client = Client::new();
+    let response = client.get(&url).send().await?;
+    if response.status().is_success() {
+        let api_response: ListApplicationsResponse = response.json().await?;
+        let app_list = api_response.data.apps;
+        let x = app_list
+            .iter()
+            .map(|app| app.id.0.clone())
+            .any(|app| app == *application_id);
```
miraclx: this seems wasteful.. the allocation is unnecessary

```suggestion
            .any(|app| &app.id.0 == application_id);
```

or better yet, lean into the available methods on `ApplicationId`, accessing the contained type in a "Newtype" is not recommended

```suggestion
            .any(|app| app.id.as_ref() == application_id);
```

crates/meroctl/src/cli/context/ls.rs
```diff
@@ -0,0 +1,57 @@
+use calimero_primitives::identity::Context;
+use clap::Parser;
+use eyre::eyre;
+use reqwest::Client;
+use serde::{Deserialize, Serialize};
+
+use super::get_ip;
+use crate::cli::RootArgs;
+use crate::config::ConfigFile;
+#[derive(Debug, Serialize, Deserialize)]
```
miraclx:
```suggestion

#[derive(Debug, Serialize, Deserialize)]
```

xilosada:
@petarjuki7 we need to add a `man` or `help` option to each command. Ideally in the same PR and add it to the test plan. This also applies to the base branch.

miraclx:
> or help option to each command

thankfully, `clap` generates one by default from the rustdoc so we just need to have `///` over things of interest, so that needs to be properly done

but we can expand on the help produced by `clap` with an `after_help` directive with examples on using it for example

(see here: https://github.com/calimero-archive/nearx/blob/2f8bf5bab032e4c049a79d02ea9068c8ec1d4248/src/cli/call.rs#L26)

crates/meroctl/src/cli/context/ls.rs
petarjuki7: Just a personal preference, I think it is easier to read when there are a lot of contexts.
Probably should also return a `json` so its usable if piped into other cli programs.

miraclx: ah, okay..

how about listing just the context IDs by default

```console
$ meroctl context ls
94G5Y6mNvHAZTS61JNCF1rvj4kwyn6m7Ns3TtexztFjW
3vthqNdhcaFHjRCdGcirjvX5qFcUjGNYaUCXs3oyBPQp
3NTfcKz5j4iDLi4mUiqsgpLdgdWtomD7Jd1jZTCV8MjH
```

and something like this with `-l, --long`:

```console
$ meroctl context ls -l
├─[94G5Y6mNvHAZTS61JNCF1rvj4kwyn6m7Ns3TtexztFjW]
│   ├─[Application ID]: Hcf7pWQqFNzCRvkyHSdjDQjagKh4QKhoZYKBfuANeqgt
│   │   ├─[Name]: only-peers
│   │   └─[Version]: 0.2.0
│   └─[Identities]: 12
│       ├─ 7e7d5Rag4HRx2Tp2BTMLqi4m8nn2bswHFRzjEEDLrjs8
│       ├─ ...
│       ├─ ...
│       └─ 6w931mWdGyfX3XC5QjP3qhNLE6SC2ybYGHbV5L8vLGwn
├─[3vthqNdhcaFHjRCdGcirjvX5qFcUjGNYaUCXs3oyBPQp]
│   ├─[Application ID]: D36d8HH1gwEtc7furQdiXr8GvGysCKYR6JevcQgN3HeH
│   │   ├─[Name]: rock-paper-scissors
│   │   └─[Version]: 0.1.0
│   └─[Identities]: 3
│       ├─ 3imfDqhqvFNrsn3dmgK9315vyuvtLWu22vY7Pd8vwZfp
│       ├─ GNu4AVzgDU9QRKScg1d2qMM3XL7FmBRatkjPhQWDPK9B
│       └─ 7vkuKGGYapXnkT4dL5Sapt5MLKpjzCaJ3S5HehQFrA1M
└─[3NTfcKz5j4iDLi4mUiqsgpLdgdWtomD7Jd1jZTCV8MjH]
    ├─[Application ID]: 3D9R6ZsPL6j8JawaAZo9Jwi8uGJGHt6NXSqA16oP8cqR
    │   ├─[Name]: kv-store
    │   └─[Version]: 0.1.0
    └─[Identities]: 2
        ├─ EfHYJ3oNhg5QBQgGe8ZxSLtS5Tpu2tBVnH9HfAX2JwgU
        └─ CJDuNj6RaQQFydu8ueWTXR1VsZZtdoFmtJp7YaScS6vd
```

crates/application/src/lib.rs
```diff
@@ -187,6 +213,31 @@ impl ApplicationManager {
...
 
+    fn link_release(
+        &self,
+        application_id: &calimero_primitives::application::ApplicationId,
+        dir: &camino::Utf8Path,
+        version: &semver::Version,
+        link_path: &camino::Utf8Path,
+    ) -> eyre::Result<()> {
+        let base_path = format!("{}/{}/{}", dir, application_id, version);
+        info! {"{}", base_path};
```
its-saeed: Are these types of logs without any prefixes helpful?

crates/application/src/lib.rs
```diff
@@ -187,6 +213,31 @@ impl ApplicationManager {
...
+        fs::create_dir_all(&base_path)?;
+
+        let file_path = format!("{}/binary.wasm", base_path);
+        info!("{}", file_path);
+        match symlink(link_path, &file_path) {
+            Ok(_) => {}
+            Err(err) => eyre::bail!("Symlinking failed: {}", err),
+        }
```
miraclx: this wants to be an `if let`

```suggestion
        if let Err(err) = symlink(link_path, &file_path) {
            eyre::bail!("Symlinking failed: {}", err);
        }
```

crates/meroctl/src/cli/context/common.rs
```diff
@@ -0,0 +1,21 @@
...
+
+    let port = multiaddr
+        .iter()
+        .find_map(|p| match p {
+            libp2p::multiaddr::Protocol::Tcp(port) => Some(port),
+            _ => None,
+        })
+        .ok_or_else(|| eyre::eyre!("No TCP port found in Multiaddr"))?;
+    let url = Url::parse(&format!("http://{}:{}", ip, port))?;
+    return Ok(url);
```
miraclx:
```suggestion
    Ok(url)
```

crates/meroctl/src/cli/context/create.rs
```diff
@@ -0,0 +1,190 @@
...
+        let Some(multiaddr) = config.network.server.listen.first() else {
+            eyre::bail!("No address.")
+        };
+
+        let base_url = get_ip(multiaddr)?;
+
+        match self.dev {
+            true => Ok(link_local_app(base_url, self.path, self.version).await?),
+            false => Ok(create_context(base_url, self.application_id).await?),
+        }
```
miraclx: an `if` statement here reads better, no?

```suggestion
        if self.dev {
            return link_local_app(base_url, self.path, self.version).await
        }

        create_context(base_url, self.application_id).await
```

crates/meroctl/src/cli/context/list.rs
```diff
@@ -0,0 +1,53 @@
...
+                eyre::bail!("Failed to load config file");
+            };
+            let multiaddr = config
+                .network
+                .server
+                .listen
+                .first()
+                .ok_or_else(|| eyre!("No address."))?;
+            let base_url = get_ip(multiaddr)?;
+            let url = format!("{}admin-api/contexts-dev", base_url);
```
miraclx: reads to me like instead of `get_ip`, we can have a method that takes the path plus an address, and returns the `Url` for the full path

instead of adding a secondary path concatenation line

crates/meroctl/src/cli/context/list.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Debug, Parser)]
+pub struct ListCommand {}
+
+impl ListCommand {
+    pub async fn run(self, root_args: RootArgs) -> eyre::Result<()> {
+        let path = root_args.home.join(&root_args.node_name);
+        if !ConfigFile::exists(&path) {
+            eyre::bail!("Config file does not exist")
+        } else {
```
miraclx: `else` is unnecessary here, bail has already `return`-ed

crates/meroctl/src/cli/context/list.rs
```diff
@@ -0,0 +1,53 @@
...
+            let Ok(config) = ConfigFile::load(&path) else {
+                eyre::bail!("Failed to load config file");
+            };
+            let multiaddr = config
+                .network
+                .server
+                .listen
+                .first()
+                .ok_or_else(|| eyre!("No address."))?;
```
miraclx: https://github.com/calimero-network/core/pull/413#discussion_r1653988545

crates/meroctl/src/cli/context/common.rs
```diff
@@ -1,6 +1,9 @@
 use reqwest::Url;
 
-pub(crate) fn get_ip(multiaddr: &libp2p::Multiaddr) -> eyre::Result<reqwest::Url> {
+pub(crate) fn get_ip(
```
miraclx: should be renamed

crates/meroctl/src/cli/context/common.rs
```diff
@@ -16,6 +19,11 @@ pub(crate) fn get_ip(multiaddr: &libp2p::Multiaddr) -> eyre::Result<reqwest::Url
             _ => None,
         })
         .ok_or_else(|| eyre::eyre!("No TCP port found in Multiaddr"))?;
-    let url = Url::parse(&format!("http://{}:{}", ip, port))?;
-    return Ok(url);
+
+    let mut url = Url::parse(&format!("http://{}:{}", ip, port))?;
```
miraclx: we're assuming the scheme here, but we can derive that from the multiaddr

crates/meroctl/src/cli/context/create.rs
```diff
@@ -94,12 +94,12 @@ impl CreateCommand {
             eyre::bail!("No address.")
         };
 
-        let base_url = get_ip(multiaddr)?;
+        let base_url = get_ip(multiaddr, None)?;
```
miraclx: what do you need the base url for?

crates/meroctl/src/cli/context/ls.rs
xilosada: Please create a new issue

crates/meroctl/src/cli/context/common.rs
```diff
@@ -1,6 +1,6 @@
 use reqwest::Url;
 
-pub(crate) fn get_ip(
+pub(crate) fn multiaddr_to_url(
     multiaddr: &libp2p::Multiaddr,
     api_path: Option<String>,
```
miraclx, line 5: this doesn't need to be an option

you never use the `None` path

https://github.com/calimero-network/core/pull/413#discussion_r1658374106

PR #417, opened 2024-06-27
chore: add mock host methods for non-wasm32 targets

### LINKED ISSUE #235: Add mock host iface for non-wasm32 targets

its-saeed:
## Summary

Let's define a module that mocks out all the expected host functions in `sys`, to prevent the IDE from showing us this error:
image

The mocked out methods should **effectively be noops**, but should **panic** to discourage using it

Fixes #235 

## Test plan
Only-peers app is compiled successfully on a non-wasm target.

## Summary by CodeRabbit

- **Chores**
  - Added conditional module `mock` for better architecture-specific handling.
  - Included `cfg-if` dependency for conditional compilation.

- **Documentation**
  - Removed unnecessary compile error message related to target architecture for cleaner code.


crates/sdk/src/sys/mock.rs
```diff
@@ -0,0 +1,62 @@
+use super::types::*;
+
+#[no_mangle]
+extern "C" fn panic(_loc: Location) -> ! {
+    panic!("Although your build was successful, it doesn't mean you're allowed to call this function in a non WASM target!")
```
miraclx: can you define a declarative macro to reduce duplication?

miraclx: let's also simplify the message

its-saeed: Something like `The function_name is supposed to call only on WASM target. Panicking!`?

its-saeed: The funciton_name is supposed to call only on WASM target. Panicking!

its-saeed: Does this work?

miraclx: what do you think of something like:

```text
host function `panic` is only available when compiled for wasm32
```

crates/sdk/src/sys/mock.rs
```diff
@@ -0,0 +1,50 @@
+use super::types::*;
+
+macro_rules! non_wasm_panic {
+    (fn $func:ident($($arg:ident: $arg_ty:ty),*$(,)?) $(-> $returns:ty)?;) => {
+        #[no_mangle]
+        extern "C" fn $func($($arg: $arg_ty),*) $(-> $returns)? {
+            panic!("host function `{}` is only available when compiled for wasm32", stringify!($func_name));
```
miraclx:
```suggestion
            panic!("host function `{}` is only available when compiled for wasm32", stringify!($func));
```

miraclx (commented):
bit more complex than I expected, but this pattern is starting to give me ideas

what do you think about making them all one statement?

and if you're doing that, then we can merge everything into sys and make it something like this:

```rust
wasm_imports! {
    "env" => {
        pub fn panic(loc: Location) -> !;
        pub fn panic_utf8(msg: Buffer, loc: Location) -> !;
    },
}
```

what do you think?

its-saeed:
> bit more complex than I expected, but this pattern is starting to give me ideas
> 
> what do you think about making them all one statement?
> 
> and if you're doing that, then we can merge everything into sys and make it something like this:
> 
> ```rust
> wasm_imports! {
>     "env" => {
>         pub fn panic(loc: Location) -> !;
>         pub fn panic_utf8(msg: Buffer, loc: Location) -> !;
>     },
> }
> ```
> 
> what do you think?

Good idea, I did it, but now the compiler complains about unnecessary `unsafe` block. Should I do something unsafe but dummy before panicking?

crates/sdk/src/sys.rs
```diff
@@ -1,32 +1,51 @@
...
-    // --
-    pub fn input(register_id: RegisterId);
-    pub fn value_return(value: ValueReturn);
-    pub fn log_utf8(msg: Buffer);
-    pub fn emit(event: Event);
-    // --
-    pub fn storage_read(key: Buffer, register_id: RegisterId) -> Bool;
-    pub fn storage_write(key: Buffer, value: Buffer, register_id: RegisterId) -> Bool;
+macro_rules! wasm_imports {
+    ($module:expr => { $(fn $func_name:ident($($arg:ident: $arg_ty:ty),*) $(-> $returns:ty)?;)* }) => {
```
miraclx:
```suggestion
    ($module:literal => { $(fn $func_name:ident($($arg:ident: $arg_ty:ty),*) $(-> $returns:ty)?;)* }) => {
```

crates/sdk/src/sys.rs
```diff
@@ -1,32 +1,51 @@
...
+                #[link(wasm_import_module = $module)]
+                extern "C" {
+                    $(
+                        pub fn $func_name($($arg: $arg_ty),*) $(-> $returns)?;
+                    )*
+                }
+            } else {
+                $(
+                    #[no_mangle]
+                    pub extern "C" fn $func_name($($arg: $arg_ty),*) $(-> $returns)? {
```
miraclx: since we're using this model, we no longer need the `extern "C"`, and to match the behaviour of the `extern`, we can mark this `unsafe`

```suggestion
                    pub unsafe fn $func_name($($arg: $arg_ty),*) $(-> $returns)? {
```

miraclx: I suspect this \*might\* complain about unused args?

its-saeed: Yes, it did, but if you look at the code, I underscored the arg names.

its-saeed: Now the only thing that it complains about only-peers build is:

```
warning: methods `post`, `posts`, `create_post`, and `create_comment` are never used
  --> apps/only-peers/src/lib.rs:46:12
   |
45 | impl OnlyPeers {
   | -------------- methods in this implementation
46 |     pub fn post(&self, id: usize) -> Option<&Post> {
   |            ^^^^
...
52 |     pub fn posts(&self) -> &[Post] {
   |            ^^^^^
...
58 |     pub fn create_post(&mut self, title: String, content: String) -> &Post {
   |            ^^^^^^^^^^^
...
84 |     pub fn create_comment(
   |            ^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `only-peers` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.69s
```

miraclx: what was the command?

its-saeed: `cargo build -p only-peers`

miraclx: Ah, I think that's because the `OnlyPeers` struct is private

its-saeed: ah, Wait! Whenever I run `./apps/only-peers/build.sh` I get a lot of `cannot find type Bool in this scope` errors for all of the imported types!

its-saeed: How can I use `cargo expand` to expand sys.rs to see the generated code?

apps/only-peers/src/lib.rs
```diff
@@ -5,14 +5,14 @@ use calimero_sdk::{app, env};
 #[app::state(emits = for<'a> Event<'a>)]
 #[derive(Default, BorshSerialize, BorshDeserialize)]
 #[borsh(crate = "calimero_sdk::borsh")]
-struct OnlyPeers {
+pub struct OnlyPeers {
```
miraclx, line 8: let's do the same for `kv-store` and `gen-ext`

crates/sdk/src/sys.rs
miraclx: > How can I use cargo expand to expand sys.rs to see the generated code?

you can use the rust-analyzer option `expand macro recursively at caret`

miraclx: Alternatively, `cargo expand -p calimero-sdk sys`

crates/sdk/src/sys.rs
```diff
@@ -1,30 +1,54 @@
 #![allow(dead_code)]
 
+#[cfg(not(target_arch = "wasm32"))]
```
miraclx: you forgot this

crates/sdk/src/sys.rs
miraclx: https://github.com/calimero-network/core/pull/417#discussion_r1657416877

crates/sdk/src/sys.rs
```diff
@@ -1,30 +1,54 @@
...
-    pub fn read_register(register_id: RegisterId, buf: BufferMut) -> Bool;
-    // --
-    pub fn input(register_id: RegisterId);
-    pub fn value_return(value: ValueReturn);
-    pub fn log_utf8(msg: Buffer);
-    pub fn emit(event: Event);
-    // --
-    pub fn storage_read(key: Buffer, register_id: RegisterId) -> Bool;
-    pub fn storage_write(key: Buffer, value: Buffer, register_id: RegisterId) -> Bool;
+macro_rules! wasm_imports {
```
miraclx: let's move the macro definition to the bottom of the file since the methods are the important bit here

it'll complain, since macros must be in scope before their use, that's okay, we can abuse `use`

just add this after the macro definition:

```rust
use wasm_imports;
```

miraclx (approved):
lgtm, thanks ✨

PR #422, opened 2024-06-28
fix(calimero-sdk): update auth logic for root key

frdomovic:
# fix(calimero-sdk): update auth logic for root key

## Summary
Overlooked in process of updating auth headers -> auth headers added to the root key sdk.


danwilliams (approved):
I can confirm that these changes resolve the error 🙂

PR #438, opened 2024-07-01
chore: fix `mdns` discrimination logic

miraclx:
- `None` no user specification
- `Some({ mdns: true, no_mdns: false })` when `--mdns`
- `Some({ mdns: true, no_mdns: true })` when `--no-mdns`

Specifying both `--mdns` and `--no-mdns` is forbidden


petarjuki7 (approved):
Makes sense, lgtm!

miraclx:
Realized I could further compress the check, so we only rely on a single source of truth.. `Some(mdns)?.mdns` since `clap` defaults `bool`s to `false`

PR #443, opened 2024-07-02
feat(node): persist executed transaction in the store

fbozic:
Node store prerequisites for the catchup logic
- Persist executed transactions in the store

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -608,8 +608,24 @@ impl Node {
         };
 
         let outcome = self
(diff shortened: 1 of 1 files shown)


miraclx (approved):
lgtm.

chefsale:
Mira approved without a comment again? Bozic are you paying him to close eyes? 😂

miraclx:
😅 the only issue here is the clones, but that's a consequence of type convenience on the runtime side which could eventually be optimized. Aside that, 🍾 .

Issue #184, opened 2024-04-23
Node State Consistency - Coordinator version


chefsale:
State needs to become consistent across all peers with a coordinator, store all transaction forever, modes should be able to have a catchup.

Estimate experimental: ~1 week
Estimate production: ~2 weeks

xilosada:
@fbozic is this fixed?

PR #445, opened 2024-07-03
feat(node): persist last executed transaction hash in the store

fbozic:
Node store prerequisites for the catchup logic
- Persist last executed transaction hash in the store

### DIFF
--- a/crates/store/src/types/context.rs
+++ b/crates/store/src/types/context.rs
@@ -7,14 +7,14 @@ use crate::key;
 use crate::slice::Slice;
 use crate::types::PredefinedEntry;
 
-pub type Hash = [u8; 32];
+pub type TransactionHash = [u8; 32];
 
 #[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
 pub struct ContextMeta {
     // todo! make [u8; 32] when application_id<->meta is a separate record
     pub application_id: Box<str>,
 
-    pub last_transaction_hash: Hash,
+    pub last_transaction_hash: TransactionHash,
 }
 
 impl DataType<'_> for ContextMeta {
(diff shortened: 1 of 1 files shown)


crates/node/src/types.rs
```diff
@@ -43,3 +46,24 @@ pub struct SignedPeerAction {
     pub action: PeerAction,
     pub signature: Signature,
 }
+
+pub(crate) struct LastTxEntry {
+    key: Generic,
+}
```
miraclx: why not expand `calimero_store::types::ContextMeta` to include last tx entry?

fbozic: because I'm not that well acquainted with the store yet. that sounds great, I will implement it.

crates/store/src/types/context.rs
```diff
@@ -11,6 +11,8 @@ use crate::types::PredefinedEntry;
 pub struct ContextMeta {
     // todo! make [u8; 32] when application_id<->meta is a separate record
     pub application_id: Box<str>,
+
+    pub last_transaction_hash: [u8; 32],
```
fbozic: Is `[u8; 32]` okay as type here?

miraclx: yeah, it's fine.. but you can choose to use a type alias so we can unify them since it's also used in `ContextTransaction{prior_hash}`

fbozic: done

crates/store/src/types/context.rs
```diff
@@ -7,12 +7,14 @@ use crate::key;
 use crate::slice::Slice;
 use crate::types::PredefinedEntry;
 
+pub type Hash = [u8; 32];
```
miraclx: can we be more descriptive on the type of hash?

fbozic: `CalimeroPrimitiveHash`?

miraclx: `TransactionHash` should suffice, I think

miraclx (approved):

PR #447, opened 2024-07-03
feat(node): implement catchup stream handler

fbozic:
- Implement a catchup stream handler. 
- Cleanup unused types from the initial implementation.

In the following PR I will implement initiation of stream connection and request sending. 

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -3,7 +3,7 @@ use std::collections::VecDeque;
 use calimero_primitives::events::OutcomeEvent;
 use calimero_runtime::logic::VMLimits;
 use calimero_runtime::Constraint;
-use calimero_store::{key, Store};
+use calimero_store::Store;
 use futures_util::{SinkExt, StreamExt};
 use libp2p::gossipsub::TopicHash;
 use libp2p::identity;


crates/node/src/lib.rs
```diff
@@ -704,12 +707,111 @@ impl Node {
...
         Ok(outcome)
     }
-
     async fn handle_stream(
         &mut self,
-        _stream: calimero_network::stream::Stream,
+        mut stream: calimero_network::stream::Stream,
     ) -> eyre::Result<()> {
-        // TODO: implement catchup mechanism once storage is ready
+        let request = match stream.next().await {
```
miraclx: use `let-else` to short-circuit early, which helps compress indentation

```rust
let Some(message) = stream.next().await else {
    eyre::bail!("Stream closed unexpectedly")
}

let request = match serde_json::from_slice(&message?.data)? {
    types::CatchupStreamMessage::Request(req) => req,
    message => {
        eyre::bail!("Unexpected message: {:?}", message)
    }
}
```

crates/node/src/lib.rs
```diff
@@ -704,12 +707,111 @@ impl Node {
...
+            Some(message) => match message {
+                Ok(message) => match serde_json::from_slice(&message.data)? {
+                    types::CatchupStreamMessage::Request(req) => req,
+                    message => {
+                        eyre::bail!("Unexpected message: {:?}", message)
+                    }
+                },
+                Err(err) => eyre::bail!(err),
```
miraclx: `?` would've helped here

crates/node/src/catchup.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub(crate) async fn flush_with_error(
+        &mut self,
+        error: &types::CatchupError,
+    ) -> eyre::Result<()> {
+        self.flush().await?;
+
+        let message = serde_json::to_vec(error)?;
+        self.stream
+            .send(calimero_network::stream::Message { data: message })
+            .await?;
```
miraclx, line 68: if we're not clearing the pending messages from the queue, why send the error separately?

crates/node/src/lib.rs
```diff
@@ -704,12 +707,111 @@ impl Node {
...
+            }
+
+            key = calimero_store::key::ContextTransaction::new(
+                request.context_id,
+                transaction.prior_hash.into(),
+            );
+        }
+
+        if !last_executed_transaction_hash_discovered {
+            let message = serde_json::to_vec(&types::CatchupError::TransactionNotFound {
```
miraclx: just noting that - depending on how the transactions are applied on receipt, we may need a different error for this case

`TransactionNotFound` can be quickly returned without traversing tx links

Just do a `get` on the last executed we're given so failure is instant and we don't spend a lot of time looking for it

This case, however, represents the potential for us to have transactions that aren't part of the canonical history

Which, should be managed such that it doesn't happen (but \*might\* warrant a separate error case)

crates/node/src/lib.rs
```diff
@@ -704,12 +707,111 @@ impl Node {
...
+            batch_writer
+                .send(types::TransactionWithStatus {
+                    transaction_hash: hash.into(),
+                    transaction: calimero_primitives::transaction::Transaction {
+                        context_id: request.context_id,
+                        method: transaction.method.into(),
+                        payload: transaction.payload.into(),
+                        prior_hash: calimero_primitives::hash::Hash::from(transaction.prior_hash),
+                    },
+                    status: types::TransactionStatus::Executed,
```
miraclx, line 793: Hm, is this useful?

fbozic: I think yes. When received, executed transactions should be executed immediately and transactions from the pool should be inserted in the pool (and executed later once TransactionConfirmation is received). Let me know if I am missing something.

crates/node/src/catchup.rs
fbozic: I don't understand the question. First I flush the batch and after that, I send an error message.

miraclx: yeah, why not add the error to the batch to be sent together?

miraclx: ah - different types, the batch isn't a "message" it's transactions, oke oke

crates/node/src/lib.rs
miraclx: Ah, yeah.. 👌🏽

crates/node/src/catchup.rs
fbozic: yeah that is why. let me know if we want to open a issue to implement BatchMessage on the stream protocol itself.

miraclx: ye, it'll be useful for Blob API for example, but for now this is okay

miraclx (commented):
lgtm so far, will you be implementing the apply side in this PR or a follow-up?

miraclx (approved):
lgtm

crates/node/src/catchup.rs
fbozic: https://github.com/calimero-network/core/issues/449

PR #421, opened 2024-06-28
feat(store): rollup of 3 patches

miraclx:
#255 + #409 + #410

PR #448, opened 2024-07-03
feat: subscribe to context IDs

miraclx:
following #421, we now need to subscribe to context IDs, as opposed to application IDs

Added `context` subcommand to interactive CLI

```console
help
 │ Unknown command: `help`
 │ Usage: [call|peers|pool|gc|store|context] [args]
context
 │ Usage: context [ls|join|leave|create|delete|state] [args]
```


crates/node/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub struct Node {
     typ: calimero_node_primitives::NodeType,
     store: calimero_store::Store,
     tx_pool: transaction_pool::TransactionPool,
-    application_manager: calimero_application::ApplicationManager,
+    ctx_mgr: calimero_context::ContextManager,
```
fbozic: ~nit: Not a fan of `mgr` part. `ctx` is already short for context and it is kind more standard (especially if you have worked Golang). I would keep `manager` long.

crates/context/src/lib.rs
```diff
@@ -17,46 +17,158 @@ use tracing::{error, info};
...
+    ) -> eyre::Result<()> {
+        // todo! ensure application is installed
+
+        let mut handle = self.store.handle();
+
+        handle.put(
+            &calimero_store::key::ContextMeta::new(context.id),
+            &calimero_store::types::ContextMeta {
+                application_id: context.application_id.0.into(),
+                last_transaction_hash: [0; 32],
```
fbozic: ~nit: Why not `*calimero_primitives::hash::Hash::default().as_bytes()`?

miraclx: aside it being compact, `Hash` right now is more costly memory-wise than `[0; 32]`

but fair enough, since you introduced the type alias, I can rewrite this

frdomovic (approved):

PR #452, opened 2024-07-04
feat(node): implement client catchup logic

fbozic:
- Implement the client part of the catchup logic
- Extend `context::ApplicationConfig` with catchup config (for now only `batch_size`)
- Remove unnecessary `pub` from some `ContextManager` methods

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -40,6 +40,46 @@ impl ContextManager {
         Ok(this)
     }
 
+    async fn boot(&self) -> eyre::Result<()> {
+        let handle = self.store.handle();
+
+        let mut iter = handle.iter(&calimero_store::key::ContextMeta::new([0; 32].into()))?;
+
+        for key in iter.keys() {
+            self.subscribe(&key.context_id()).await?;
+        }
+
+        Ok(())
+    }
+
+    async fn subscribe(
+        &self,
+        context_id: &calimero_primitives::context::ContextId,
+    ) -> eyre::Result<()> {
+        self.network_client
+            .subscribe(calimero_network::types::IdentTopic::new(context_id))
+            .await?;
+
+        info!(%context_id, "Subscribed to context");
+
+        Ok(())
+    }
+
+    async fn unsubscribe(
+        &self,
+        context_id: &calimero_primitives::context::ContextId,
+    ) -> eyre::Result<()> {
+        self.network_client
+            .unsubscribe(calimero_network::types::IdentTopic::new(context_id))
+            .await?;
+
+        info!(%context_id, "Unsubscribed from context");
+
+        Ok(())
+    }
+}
+
+impl ContextManager {
     pub async fn join_context(
         &self,
         context_id: &calimero_primitives::context::ContextId,
@@ -378,43 +418,5 @@ impl ContextManager {
         } else {
             None
         }
-    }
-
-    async fn boot(&self) -> eyre::Result<()> {
-        let handle = self.store.handle();
-
-        let mut iter = handle.iter(&calimero_store::key::ContextMeta::new([0; 32].into()))?;
-
-        for key in iter.keys() {
-            self.subscribe(&key.context_id()).await?;
-        }
-
-        Ok(())
-    }
-
-    async fn subscribe(
-        &self,
-        context_id: &calimero_primitives::context::ContextId,
-    ) -> eyre::Result<()> {
-        self.network_client
-            .subscribe(calimero_network::types::IdentTopic::new(context_id))
-            .await?;
-
-        info!(%context_id, "Subscribed to context");
-
-        Ok(())
-    }
-
-    async fn unsubscribe(
-        &self,
-        context_id: &calimero_primitives::context::ContextId,
-    ) -> eyre::Result<()> {
-        self.network_client
-            .unsubscribe(calimero_network::types::IdentTopic::new(context_id))
-            .await?;
-
-        info!(%context_id, "Unsubscribed from context");
-
-        Ok(())
     }
 }


crates/context/src/config.rs
```diff
@@ -3,4 +3,11 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ApplicationConfig {
     pub dir: camino::Utf8PathBuf,
+
+    pub cathup: CatchupConfig,
```
miraclx:
```suggestion
    pub catchup: CatchupConfig,
```

crates/meroctl/src/config_file.rs
```diff
@@ -57,6 +57,12 @@ pub struct StoreConfig {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ApplicationConfig {
     pub path: camino::Utf8PathBuf,
+    pub cathup: CatchupConfig,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CatchupConfig {
+    pub batch_size: u8,
```
miraclx: this should be a networking config imo

crates/context/Cargo.toml
```diff
@@ -14,10 +14,11 @@ near-jsonrpc-client.workspace = true
 near-jsonrpc-primitives.workspace = true
 near-primitives.workspace = true
 reqwest.workspace = true
-semver.workspace = true
+semver = { workspace = true, features = ["serde"] }
```
miraclx: why's this needed?

crates/context/Cargo.toml
```diff
@@ -14,10 +14,11 @@ near-jsonrpc-client.workspace = true
...
 serde.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
+tokio = { workspace = true, features = ["sync"] }
```
miraclx: or this

crates/node/src/lib.rs
```diff
@@ -986,6 +1028,112 @@ impl Node {
 
         Ok(())
     }
+
+    async fn perform_catchup(
+        &mut self,
+        context_id: calimero_primitives::context::ContextId,
+        choosen_peer: libp2p::PeerId,
```
```suggestion
        chosen_peer: libp2p::PeerId,
```

crates/node/src/lib.rs
```diff
@@ -504,41 +504,56 @@ impl Node {
...
-                                            peer_id: their_peer_id,
-                                        },
-                                    ),
-                            },
-                        ));
+                    match self.typ {
+                        calimero_node_primitives::NodeType::Peer => {
+                            match self.perform_catchup(context_id, their_peer_id).await {
+                                Ok(_) => {
+                                    info!(%their_peer_id, "Attempting to perform catchup");
```
miraclx: this should come before `perform_catchup`, yes?

crates/node/src/lib.rs
```diff
@@ -504,41 +504,56 @@ impl Node {
...
+                        "Failed to parse topic hash into context ID, ignoring.."
                     );
 
                     return Ok(());
                 };
 
                 if self
                     .ctx_manager
-                    .is_application_installed(&context.application_id)
+                    .is_context_pending_catchup(&context_id)
```
miraclx: why not leave it to the new subscriber to make an explicit request to catchup?

and make the catchup process bidirectional?

so:

node A joins the context when nobody is online
node B comes online, subscribes - notifying node A
nothing happens just yet
node B sends a catchup request to all who may be available
node A responds, but since he lags behind node B
A stream is opened for node B to aid node A catchup

miraclx: And the reason for this, is.. what if node B helps node A get caught up, but node B is itself behind

when node C which has the latest state comes online, neither node A or B will know to make a catchup request

crates/context/src/lib.rs
```diff
@@ -21,6 +24,12 @@ pub struct ContextManager {
     pub config: config::ApplicationConfig,
     pub store: calimero_store::Store,
     pub network_client: NetworkClient,
+    state: Arc<RwLock<State>>,
```
miraclx: that revision will also help remove this managed mutation pattern

crates/node/src/lib.rs
```diff
@@ -986,6 +1028,112 @@ impl Node {
...
+                types::CatchupStreamMessage::Error(err) => {
+                    eyre::bail!(err);
+                }
+                event => {
+                    warn!(?event, "Unexpected event");
+                }
+            }
+        }
+
+        return Ok(Some(()));
```
```suggestion
        Ok(Some(()))
```

crates/context/Cargo.toml
fbozic: for this `use tokio::sync::RwLock;`

miraclx: yeah, saw that later.. forgot to delete this

crates/context/Cargo.toml
fbozic: not needed, not sure why I added it though

fbozic: removed

crates/node/src/lib.rs
fbozic: hmm, not sure if I understand the flow fully. 

> why not leave it to the new subscriber to make an explicit request to catchup?   

immediately after calling `network_client.subscribe(topic).await?;` there are no mesh peers, so I can not perform the request directly. or am I missing something? 
- node A comes online, and subscribes (no notifications)
- node B comes online, and subscribes (A gets notified that B subscribed, and B gets notified that A subscribed)
- node C comes online, and subscribes (A gets notified that C subscribed, and B gets notified that C subscribed, C gets 2 notifications (that A and B subscribed))

do you mean something like this? make a direct request (`WhatIsYourLatestTransactionHash`) to the new subscriber when we get notified that someone subscribed. when we receive the response (`MyLatestTransactionHas`), we check if we have this transaction locally, and if we don't we open a stream to the peer which responded.

crates/meroctl/src/config_file.rs
fbozic: moved it to the networking config, but I am storing it in the node. do you think `NetworkClient` might be a better place to store this config?

miraclx: yeah

fbozic: moved to `NetworkClient`

crates/node/src/lib.rs
miraclx: > there are no mesh peers,

good point..

> make a direct request (WhatIsYourLatestTransactionHash)

that can be an accommodation for this, unless there's some other mechanism (event) that indicates mesh peer connection

but if there isn't, I'm okay with that approach

fbozic: >>make a direct request (WhatIsYourLatestTransactionHash)

> that can be an accommodation for this, unless there's some other mechanism (event) that indicates mesh peer connection

`Subscribed` event indicates new mesh peer connection (and we are already reacting to it)

how about we try to perform a catchup whenever we get a `Subscribed` event? this guarantees that we try to catchup with all possible peers and it works bidirectionally (one stream from A to B where A requests a catchup, and one stream from B to A where B requests a catchup).

miraclx: > > that can be an accommodation for this, unless there's some other mechanism (event) that indicates mesh peer connection
>
> Subscribed event indicates new mesh peer connection (and we are already reacting to it)

when node C subscribes, it doesn't notify node C about node A and node B's prior subscription

so node C won't know to initiate the catchup request

I was hoping for another mechanism entirely, although that might require manual implementation (since it's eventually possible to know how many other peers you're connected to are subscribed to a topic)

> how about we try to perform a catchup whenever we get a Subscribed event?

my concern is the potential for all other peers to immediately spam you with catchup requests which I'd expect to be overwhelming

as opposed to the requesting node selectively requesting a catchup payload in an iterative process

crates/node/src/lib.rs
```diff
@@ -546,6 +546,25 @@ impl Node {
...
                 match serde_json::from_slice(&message.data)? {
                     types::PeerAction::Transaction(transaction) => {
+                        let handle = self.store.handle();
+
+                        if handle
+                            .get(&calimero_store::key::ContextTransaction::new(
+                                transaction.context_id,
+                                transaction.prior_hash.into(),
+                            ))?
+                            .is_none()
```
```suggestion
                        if !handle
                            .has(&calimero_store::key::ContextTransaction::new(
                                transaction.context_id,
                                transaction.prior_hash.into(),
                            ))?
```

crates/node/src/lib.rs
```diff
@@ -923,6 +955,20 @@ impl Node {
             return Ok(());
         };
 
+        let application_version = self.ctx_manager.get_application_latest_version(
+            &meta_value.application_id.clone().into_string().into(),
+        )?;
+
+        let message = serde_json::to_vec(&types::CatchupStreamMessage::ResponseMeta(
+            types::CatchupResponseMeta {
+                application_id: meta_value.application_id.into_string().into(),
```
miraclx: reuse `meta_value.application_id.clone().into_string().into()`

crates/node/src/lib.rs
```diff
@@ -986,6 +1032,112 @@ impl Node {
...
+                                )
+                                .await?;
+                            }
+                        }
+                    }
+                }
+                types::CatchupStreamMessage::ResponseMeta(response) => match context {
+                    Some(ref context) => {
+                        if context.application_id != response.application_id {
+                            eyre::bail!(
```
miraclx: this should be expected, and accounted for.. we should adopt the new application ID and download the blob..._

crates/node/src/lib.rs
```diff
@@ -923,6 +955,20 @@ impl Node {
...
+
+        let message = serde_json::to_vec(&types::CatchupStreamMessage::ResponseMeta(
+            types::CatchupResponseMeta {
+                application_id: meta_value.application_id.into_string().into(),
+                version: application_version,
+            },
+        ))?;
+        stream
+            .send(calimero_network::stream::Message { data: message })
+            .await?;
```
miraclx: _...so long as it's guaranteed that we only send `ResponseMeta` when there's a hash variance

so, we can add a check above this that responds with a variant for "yah, you're good" or "you're up to date"

crates/node/src/lib.rs
```diff
@@ -986,6 +1032,112 @@ impl Node {
...
+                types::CatchupStreamMessage::Error(err) => {
+                    eyre::bail!(err);
+                }
+                event => {
+                    warn!(?event, "Unexpected event");
+                }
+            }
+        }
+
+        Ok(Some(()))
```
miraclx: can't find any `return Ok(None)` so this can be simplified to return `eyre::Result<()>`

crates/node/src/lib.rs
miraclx: We have a similar thing at https://github.com/calimero-network/core/blob/6357c80691b57078c80260118f21d6396c2e1849/crates/node/src/lib.rs#L906-L911 that needs some love

crates/node/src/lib.rs
fbozic: huh, I was almost certain application id in the `ContextMeta` is immutable, will accommodate

crates/node/src/lib.rs
fbozic: I need to check if the application_id has changed even if there are no transactions to catchup as well

crates/node/src/lib.rs
miraclx: you can't directly modify the value view.. but you can re-define it in with `handle.put`

crates/node/src/lib.rs
miraclx: Ideally all synchronizable activity including things like application upgrade should be a transaction

(doesn't have to happen in this PR, just noting so we can factor that into the design)

crates/node/src/lib.rs
fbozic: yeah I get that you can mutate the value in the store, I just thought that changing application_id for the existing context is not desired behaviour

crates/context/src/lib.rs
```diff
@@ -72,6 +72,26 @@ impl ContextManager {
         Ok(())
     }
 
+    pub async fn update_context_application_id(
```
miraclx: move this past `context` methods so it's collected with `application` methods

crates/node/src/lib.rs
fbozic: I've implemented new variant `ApplicationChanged` and renamed `Response` to `TransactionsBatch`.

crates/context/Cargo.toml
```diff
@@ -18,6 +18,7 @@ semver.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sha2.workspace = true
+tokio.workspace = true
```
miraclx:
```suggestion
```

crates/context/src/lib.rs
```diff
@@ -263,6 +242,55 @@ impl ContextManager {
...
+        context_id: &calimero_primitives::context::ContextId,
+    ) -> eyre::Result<()> {
+        self.network_client
+            .unsubscribe(calimero_network::types::IdentTopic::new(context_id))
+            .await?;
+
+        info!(%context_id, "Unsubscribed from context");
+
+        Ok(())
+    }
```
miraclx: these network methods are in-between application methods

fbozic: I'll move them to the bottom since they are private, that okay with you?

crates/node/src/lib.rs
fbozic: after internal discussion we are going to implement the following:
- perform catchup when we receive transaction for which we do not have prior_transaction in our local storage
- perform periodic catchups against randomly chosen peer 

I'll implement periodic catchups in the following PR.

crates/context/src/lib.rs
miraclx: you can use a different `impl` block for maintenance methods at the top (including `start`)

but yeah that works too

fbozic: cool, never cross my mind 😄

crates/node/src/lib.rs
```diff
@@ -986,6 +1037,118 @@ impl Node {
...
+
+        let request = serde_json::to_vec(&types::CatchupStreamMessage::Request(request))?;
+
+        stream
+            .send(calimero_network::stream::Message { data: request })
+            .await?;
+
+        while let Some(message) = stream.next().await {
+            match serde_json::from_slice(&message?.data)? {
+                types::CatchupStreamMessage::TransactionsBatch(response) => {
```
miraclx, line 1083: btw, this could be `CatchupError`, it's not handled

fbozic: I am propagating to the handler
image

miraclx: in that case, these:

https://github.com/calimero-network/core/blob/1631e492df629569b86086756efbcef4953a04ba/crates/node/src/lib.rs#L926-L931

https://github.com/calimero-network/core/blob/1631e492df629569b86086756efbcef4953a04ba/crates/node/src/lib.rs#L939-L944

https://github.com/calimero-network/core/blob/1631e492df629569b86086756efbcef4953a04ba/crates/node/src/catchup.rs#L63-L70

have wrong implementations, right?

fbozic: yes that is a very good point, missed it completely 😅

miraclx: that's why it's good to rely on the type system to guide these things, I'll patch stream to be generic over a `codec` like `MessageJsonCodec` which would then be generic over an serializable type

fbozic: here is the fix: https://github.com/calimero-network/core/pull/458

Issue #401, opened 2024-06-21
Re-implement `fetch` host method


its-saeed:
worth looking into, when we redesign this to accommodate returning more complex data like status codes, and response headers

let's break up how linked data is sent over the bridge, building a linked list backed by the runtime registers, akin to how curl stores headers backed by heap memory

the idea is:

runtime:

1. define the 0th register as reserved
2. expose a method taking 3 parameters to the guest that appends to the link list
  - the first parameter should be the last register id we stored at
  - the second parameter should be a string (`sys::Buffer`)
  - the third parameter should be the register id that stores the new entry
3. to store a new entry in the linked list, we insert this at the new register location:
  - LE encoded register id for the last entry
  - the data we intend to store
4. reading this linked list would always start from the last register id and so would freeing it

guest:

how are register id's derived?

one option is to create a `Box::new(0u8)`, and use it's address for every entry.. since the guest is in a 32-bit environment, all guest memory addresses would be valid register IDs

see https://curl.se/libcurl/c/curl_slist_append.html

---

Non-blocking for this PR though, since we'll need to revise the `fetch` host-function

_Originally posted by @miraclx in https://github.com/calimero-network/core/pull/374#discussion_r1648754735_

PR #453, opened 2024-07-04
docs: Add runtime crate README.md

danwilliams:
Added a README for the runtime, which covers the main areas and operations.

### DIFF
--- a/crates/runtime/README.md
+++ b/crates/runtime/README.md
@@ -1,0 +1,592 @@
+# Calimero runtime
+
+[Serde]: https://serde.rs/
+[Wasmer]: https://wasmer.io/
+
+  - [Introduction](#introduction)
+  - [Core components](#core-components)
+    - [Overview](#overview)
+    - [VMLogic](#vmlogic)
+    - [Host functions](#host-functions)
+    - [Resource limits](#resource-limits)
+    - [Storage](#storage)
+    - [Error handling](#error-handling)
+    - [Serialisation and deserialisation](#serialisation-and-deserialisation)
+  - [Operational flow](#operational-flow)
+  - [Detailed component descriptions](#detailed-component-descriptions)
+    - [Wasm module execution (`run()` function)](#wasm-module-execution-run-function)
+    - [VMLogic](#vmlogic-1)
+    - [Host functions](#host-functions-1)
+    - [Memory management](#memory-management)
+    - [Storage](#storage-1)
+    - [Error handling](#error-handling-1)
+  - [Usage examples](#usage-examples)
+
+## Introduction
+
+The Calimero runtime is a lightweight, sandboxed execution environment for
+WebAssembly (Wasm) modules. It provides a secure and controlled platform for
+running untrusted code with defined resource limits and a set of host-provided
+functions. The runtime is designed to facilitate the execution of smart
+contracts or similar applications in a blockchain-like environment. Its modular
+design allows for easy extension and customisation, making it suitable for
+various applications requiring sandboxed execution of untrusted code.
+
+Key features of the Calimero runtime include:
+
+  - Secure execution of Wasm modules
+  - Resource limitation and metering
+  - Host-provided functions for storage, logging, and external interactions
+  - Support for deterministic execution
+  - Error handling and panic recovery
+
+This document provides a [high-level summary](#core-components) of the runtime's
+architecture and core components, followed by a [detailed description of each
+component](#detailed-component-descriptions) . It also includes a brief overview
+of the [operational flow](#operational-flow), and [usage examples](#usage-examples).
+
+## Core components
+
+### Overview
+
+```mermaid
+graph TD
+    A[Wasm Module] --> B[Calimero Runtime]
+    B --> C[VMLogic]
+    C --> D[Host Functions]
+    C --> E[Storage]
+    C --> F[Resource Limits]
+    D --> H[Memory Operations]
+    D --> I[Register Operations]
+    D --> J[Storage Operations]
+    D --> K[Logging/Events]
+    D --> L[External Requests]
+    F --> N[VMLimits]
+```
+
+### VMLogic
+
+The `VMLogic` struct is the central component of the runtime. It encapsulates:
+
+  - Storage access
+  - Memory management
+  - Context information
+  - Resource limits
+  - Logging and event emission
+
+*[Read more...](#vmlogic-1)*
+
+### Host functions
+
+The runtime provides a set of host functions that can be called from the guest
+Wasm module:
+
+  - Memory and register operations
+  - Storage read/write
+  - Logging and event emission
+  - Cryptographic operations (signing and verification)
+  - External HTTP requests (fetch)
+
+*[Read more...](#host-functions-1)*
+
+### Resource limits
+
+The `VMLimits` struct defines constraints on various resources:
+
+  - Memory usage
+  - Stack size
+  - Number and size of registers
+  - Logging and event limits
+  - Storage key and value size limits
+
+### Storage
+
+The runtime includes an abstract `Storage` trait and an in-memory implementation
+(`InMemoryStorage`) for persisting state.
+
+*[Read more...](#storage-1)*
+
+### Error handling
+
+A comprehensive error handling system is implemented, covering various
+scenarios:
+
+  - Host errors
+  - Wasm execution errors
+  - Storage errors
+  - Function call errors
+
+*[Read more...](#error-handling-1)*
+
+## Operational flow
+
+  1. Load Wasm module
+  2. Set up execution environment (memory, limits, storage)
+  3. Resolve and link host functions
+  4. Execute specified function in the Wasm module
+  5. Handle return values, logs, and events
+  6. Clean up and return execution outcome
+
+```mermaid
+graph TD
+    A[Load Wasm Module] --> B[Set up Execution Environment]
+    B --> C[Create Wasmer Engine and Store]
+    C --> D[Set Memory Limits and Tuning]
+    D --> E[Instantiate Wasm Module]
+    E --> F[Resolve and Link Host Functions]
+    F --> G[Execute Specified Function]
+    G --> H{Execution Successful?}
+    H -->|Yes| I[Collect Results]
+    H -->|No| J[Handle Errors]
+    I --> K[Process Logs and Events]
+    J --> K
+    K --> L[Clean Up Resources]
+    L --> M[Return Execution Outcome]
+
+    subgraph "Host Function Calls"
+        G --> N[Memory Operations]
+        G --> O[Register Operations]
+        G --> P[Storage Interactions]
+        G --> Q[Logging and Events]
+        G --> R[External Requests]
+    end
+```
+
+## Detailed component descriptions
+
+### Topology
+
+```mermaid
+graph TD
+    B[VMLogic]
+    C[Wasm Module - Guest Code]
+    
+    B --> E[Storage]
+    B --> F[Memory Management]
+    B --> G[Registers]
+    B --> H[Context]
+    B --> I[Host Functions]
+    
+    I --> K[Memory Operations]
+    I --> L[Register Operations]
+    I --> M[Storage Operations]
+    I --> N[Logging/Events]
+    I --> O[External Requests]
+    
+    C -.-> I
+    
+    subgraph "External Environment"
+        AB[Host System]
+    end
+    
+    E -.-> AB
+    O -.-> AB
+```
+
+### Wasm module execution (`run()` function)
+
+The `run()` function in `lib.rs` is the entry point for executing a Wasm module:
+
+  1. Creates a Wasmer engine and store
+  2. Sets up memory limits and tuning
+  3. Instantiates the Wasm module
+  4. Links host functions
+  5. Executes the specified method
+  6. Handles errors and returns the execution outcome
+
+The `run()` function is the primary entry point for executing WebAssembly (Wasm)
+modules within the Calimero runtime. Its purpose is to orchestrate the entire
+process of loading, preparing, and executing a Wasm module, while ensuring that
+all operations occur within the defined constraints and security boundaries of
+the runtime environment.
+
+At the core of the `run()` function's operation is [Wasmer][], a
+high-performance WebAssembly runtime. Wasmer is an open-source project that
+allows for the execution of Wasm modules outside of web browsers. It provides a
+sandboxed environment for running Wasm code, handling tasks such as module
+compilation, memory management, and function invocation. Wasmer's architecture
+is designed to be extensible and performant, making it an ideal choice for the
+Calimero runtime.
+
+The execution process begins with the creation of a Wasmer engine and store. The
+engine is responsible for compiling Wasm modules into executable code, while the
+store serves as a container for all runtime objects, including the compiled
+module, its memory, and any imported functions. In the Calimero runtime, a
+custom engine is created with specific tuning parameters defined by the
+`WasmerTunables` struct, allowing for fine-grained control over memory
+allocation and other runtime behaviours.
+
+Next, the `run()` function sets up the execution environment by creating an
+instance of `VMLogic`. This struct encapsulates the runtime's core logic,
+including storage access, resource limits, and the context for the current
+execution. The `VMLogic` instance serves as a mediator between the Wasm module
+and the host environment, ensuring that all interactions occur within the
+defined boundaries.
+
+The Wasm module is then loaded from the provided byte array and compiled using
+Wasmer's module compilation facilities. This step involves parsing the Wasm
+binary, validating its structure, and transforming it into a format that can be
+efficiently executed by the host system. If any errors occur during this
+process, such as invalid Wasm code, they are caught and returned as part of the
+execution outcome.
+
+Once the module is compiled, the `run()` function proceeds to link the required
+import functions. These imports, defined by the `VMLogic::imports()` method,
+provide the interface between the Wasm module and the host environment. They
+include functions for memory operations, storage access, logging, and other
+runtime services. By controlling these imports, the Calimero runtime ensures
+that the Wasm module can only interact with the host system in predefined,
+secure ways.
+
+With the module compiled and imports linked, the `run()` function creates an
+instance of the Wasm module. This instance represents a unique execution context
+for the module, with its own memory and global state. The function then
+retrieves the module's memory and associates it with the `VMLogic` instance,
+allowing for safe, controlled access to the Wasm module's memory space.
+
+The next step is to locate and prepare the specified entry point function within
+the Wasm module. The `run()` function checks that the requested function exists
+and has the correct signature (no parameters and no return value). This check
+ensures that the function can be called safely within the constraints of the
+runtime.
+
+Finally, the `run()` function invokes the specified Wasm function. During
+execution, any calls to imported functions are handled by the `VMLogic`
+instance, ensuring that all operations adhere to the defined resource limits and
+security policies. If any errors or exceptions occur during execution, they are
+caught, wrapped in appropriate error types, and included in the execution
+outcome.
+
+Throughout the execution process, the `run()` function leverages Wasmer's
+capabilities for sandboxing and resource control. Wasmer provides mechanisms for
+limiting memory usage, controlling CPU usage through "fuel metering", and
+isolating the Wasm execution from the host system. These features, combined with
+the additional controls implemented in the Calimero runtime, ensure a high level
+of security and predictability in the execution of untrusted Wasm code.
+
+After the function execution completes (or if an error occurs), the `run()`
+function collects the results, logs, and any emitted events. It then calls
+`VMLogic::finish()` to package all this information into an `Outcome` struct,
+which provides a comprehensive summary of the execution results.
+
+### VMLogic
+
+The `VMLogic` struct manages the execution context:
+
+  - Holds references to storage and resource limits
+  - Manages registers for temporary data storage
+  - Collects logs and events during execution
+  - Provides methods for interacting with the host environment
+
+The `VMLogic` struct is the centre of the Calimero runtime, orchestrating the
+execution of WebAssembly modules within a controlled and sandboxed environment.
+Its primary purpose is to manage the state, resources, and interactions between
+the guest Wasm code and the host system, ensuring safe and deterministic
+execution.
+
+At its core, `VMLogic` encapsulates several crucial elements: a mutable
+reference to the storage system, a [Wasmer][] memory instance, the execution
+context, resource limits, a register system for temporary data storage, and
+collections for logs and events. This comprehensive set of components allows
+`VMLogic` to maintain full control over the execution environment, mediating all
+interactions between the guest code and the host system.
+
+One of the key responsibilities of `VMLogic` is to enforce resource limits
+defined in the `VMLimits` struct. These limits cover various aspects of
+execution, including memory usage, stack size, register count and size, and
+constraints on storage operations. By strictly adhering to these limits,
+`VMLogic` prevents potential abuse of resources and ensures that all executions
+remain within predefined boundaries, which is crucial for maintaining the
+integrity and security of the runtime.
+
+`VMLogic` also plays a vital role in managing the execution context. It holds
+the `VMContext` struct, which contains input data for the current execution.
+This allows the runtime to provide necessary contextual information to the guest
+code, such as function parameters or environmental data, in a controlled and
+standardised manner.
+
+Another critical aspect of `VMLogic` is its role in facilitating host function
+calls. Through the `host_functions()` method, it provides a set of functions
+that the guest code can invoke to interact with the host environment. These
+functions cover a [wide range of operations](#host-functions-1). By channeling
+all these interactions through `VMLogic`, the runtime can ensure that they are
+performed safely and within the defined limits.
+
+Throughout the execution of a Wasm module, `VMLogic` collects and manages logs
+and events emitted by the guest code. This feature is crucial for debugging,
+monitoring, and auditing purposes, allowing developers and system administrators
+to gain insights into the execution flow and any significant occurrences during
+runtime.
+
+When the execution of a Wasm function concludes, `VMLogic` is responsible for
+gathering all the results, logs, and events, and packaging them into an
+`Outcome` struct. This struct provides a comprehensive summary of the execution,
+including any return values or errors encountered during the process. By
+centralising this information, `VMLogic` ensures that the caller of the runtime
+receives a complete and consistent report of the execution results.
+
+### Host functions
+
+Host functions are implemented in the `imports.rs` file:
+
+  - Panic handling
+  - Register operations (read, write, get length)
+  - Storage interactions
+  - Logging and event emission
+  - External HTTP requests (fetch)
+
+Host functions serve as the bridge between the sandboxed WebAssembly (Wasm)
+environment and the host system. These functions provide a carefully curated set
+of capabilities to the guest Wasm code, allowing it to interact with the outside
+world in a controlled and secure manner. The primary purpose of host functions
+is to extend the functionality of the Wasm module beyond its inherent
+limitations while maintaining strict control over its access to system resources
+and external services.
+
+At the core of the host function implementation is the `VMHostFunctions` struct,
+which encapsulates the runtime's logic and provides a safe interface for the
+guest code to interact with.
+
+One of the fundamental responsibilities of host functions is memory management.
+They provide methods for reading from and writing to the guest's memory space,
+ensuring that all memory accesses are bounds-checked and secure. This includes
+functions for reading raw bytes from memory as well as higher-level operations
+like retrieving UTF-8 strings.
+
+Register operations form another critical aspect of the host functions. The host
+functions ensure that all register operations adhere to the size and count
+limits specified in `VMLimits`, preventing resource abuse and maintaining the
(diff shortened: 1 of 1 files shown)


xilosada (approved):

xilosada:
Regarding the content and the format nothing to add. @miraclx ?

chefsale:
@miraclx is this ready to be landed?

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+
+### Host functions
+
+The runtime provides a set of host functions that can be called from the guest
+Wasm module:
+
+  - Memory and register operations
+  - Storage read/write
+  - Logging and event emission
+  - Cryptographic operations (keypair creation)
```
miraclx: we don't do creation, just sign & verify

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+    C --> D[Host Functions]
+    C --> E[Storage]
+    C --> F[Resource Limits]
+    C --> G[Error Handling]
+    D --> H[Memory Operations]
+    D --> I[Register Operations]
+    D --> J[Storage Operations]
+    D --> K[Logging/Events]
+    D --> L[External Requests]
+    E --> M[InMemoryStorage]
```
miraclx: this is just "an" implementation of the storage iface, I don't think it's relevant to include

unless the graph uses a different model for indicating that relationship

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+### Overview
+
+```mermaid
+graph TD
+    A[Wasm Module] --> B[Calimero Runtime]
+    B --> C[VMLogic]
+    C --> D[Host Functions]
+    C --> E[Storage]
+    C --> F[Resource Limits]
+    C --> G[Error Handling]
```
miraclx: not sure I understand this one

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+  - Wasm execution errors
+  - Storage errors
+  - Function call errors
+
+*[Read more...](#error-handling-1)*
+
+### Serialisation and deserialisation
+
+The runtime uses [Serde][] for JSON serialisation and deserialisation of
+input/output data.
```
miraclx: which parts?

input is an arbitrary slice of bytes, output too

it's up to each individual app to determine how it wants to parse the input & write the output

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+        G --> Q[Logging and Events]
+        G --> R[External Requests]
+    end
+
+    subgraph "Error Handling"
+        J --> S[VM Runtime Errors]
+        J --> T[Function Call Errors]
+        J --> U[Host Errors]
+        J --> V[Wasm Traps]
+    end
```
miraclx: I don't think there's much value in having a section on error handling at all? it's non-integral to the core functionalities

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+    
+    T[Error Handling] --> U[VMRuntimeError]
+    T --> V[FunctionCallError]
+    T --> W[HostError]
+    T --> X[WasmTrap]
+    
+    A --> T
+    
+    Y[Serialisation] --> Z[Input Deserialisation]
+    Y --> AA[Output Serialisation]
```
miraclx: not needed

I'm also a bit confused on the meaning of the arrows

seems to conflate between "dependence" and "has"

like

"Calimero Runtime" has "Error Handling" to indicate it's own behaviour
"Calimero Runtime" depends on "VMLogic"

maybe we need different arrow types?

But personally, I think "has" type relationships deal better textually than graphically

it would significantly simplify this graph

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+
+### VMLogic
+
+The `VMLogic` struct is the central component of the runtime. It encapsulates:
+
+  - Storage access
+  - Memory management
+  - Context information
+  - Resource limits
+  - Registers for inter-function communication
```
miraclx: not quite.. think about it as a facility for host allocation with privileged guest access without giving the guest any further detail about host memory

see https://github.com/calimero-network/core/issues/401

and so, it's used to pass data from the host to the guest

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+  - Provides methods for interacting with the host environment
+
+The `VMLogic` struct is the centre of the Calimero runtime, orchestrating the
+execution of WebAssembly modules within a controlled and sandboxed environment.
+Its primary purpose is to manage the state, resources, and interactions between
+the guest Wasm code and the host system, ensuring safe and deterministic
+execution.
+
+At its core, `VMLogic` encapsulates several crucial elements: a mutable
+reference to the storage system, an optional [Wasmer][] memory instance, the
```
miraclx: it's not exactly "supposed" to be optional - it needs to be there, but we can't exactly provide it when `VMLogic` is constructed since we can't have a reference to memory until the module has been instantiated, yet we use `VMLogic` to manage all error handling, so we defer for later to eventually be provided with `VMLogic::with_memory`

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+One of the key responsibilities of `VMLogic` is to enforce resource limits
+defined in the `VMLimits` struct. These limits cover various aspects of
+execution, including memory usage, stack size, register count and size, and
+constraints on storage operations. By strictly adhering to these limits,
+`VMLogic` prevents potential abuse of resources and ensures that all executions
+remain within predefined boundaries, which is crucial for maintaining the
+integrity and security of the runtime.
+
+The register system implemented within `VMLogic` provides a flexible mechanism
+for temporary data storage and inter-function communication. It allows the guest
```
miraclx: again.. inter-function communication "can" happen, but it's not the core reason for this

most cases would be one function.. that perhaps needs to pass more complex data than just a slice of bytes so it takes advantage of multiple registers, but to be consumed by one function

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+code to store and retrieve data in a controlled manner, with `VMLogic` ensuring
+that all register operations adhere to the specified size and count limits. This
+system is particularly useful for passing data between different parts of the
+guest code or for storing intermediate results during computation.
```
miraclx: this would/should happen fully within the guest code, registers use for this would be questionable

the guest can heap allocate within it's own memory

registers are only useful when you want the host to

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+`Outcome` struct. This struct provides a comprehensive summary of the execution,
+including any return values or errors encountered during the process. By
+centralising this information, `VMLogic` ensures that the caller of the runtime
+receives a complete and consistent report of the execution results.
+
+### Host functions
+
+Host functions are implemented in the `imports.rs` file:
+
+  - Panic handling and UTF-8 string management
```
miraclx: can you expand on "UTF-8 string management"? Otherwise I think this is just:

```suggestion
  - Panic handling
```

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+code, such as function parameters or environmental data, in a controlled and
+standardised manner.
+
+Another critical aspect of `VMLogic` is its role in facilitating host function
+calls. Through the `host_functions()` method, it provides a set of functions
+that the guest code can invoke to interact with the host environment. These
+functions cover a wide range of operations, including memory and register
+manipulation, storage access, logging, event emission, and even external HTTP
+requests. By channeling all these interactions through `VMLogic`, the runtime
+can ensure that they are performed safely and within the defined limits.
```
miraclx: since this is defined below, seems unnecessary to go into here

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+world in a controlled and secure manner. The primary purpose of host functions
+is to extend the functionality of the Wasm module beyond its inherent
+limitations while maintaining strict control over its access to system resources
+and external services.
+
+At the core of the host function implementation is the `VMHostFunctions` struct,
+which encapsulates the runtime's logic and provides a safe interface for the
+guest code to interact with. This struct is created using self-references,
+allowing it to maintain a consistent view of the execution environment
+throughout the lifetime of a function call.
```
miraclx: the self-reference is for memory, but I don't think it's relevant to mention it's purely an implementation detail (not integral to core function, it's just an accommodation to avoid getting multiple memory views)

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+
+One of the fundamental responsibilities of host functions is memory management.
+They provide methods for reading from and writing to the guest's memory space,
+ensuring that all memory accesses are bounds-checked and secure. This includes
+functions for reading raw bytes from memory as well as higher-level operations
+like retrieving UTF-8 strings. By mediating all memory interactions, the host
+functions prevent potential security vulnerabilities that could arise from
+unchecked memory access.
```
miraclx:
```suggestion
like retrieving UTF-8 strings.
```

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+
+Register operations form another critical aspect of the host functions. They
+allow the guest code to store and retrieve data in a temporary storage system,
+which is particularly useful for passing information between different parts of
+the Wasm module or for storing intermediate results. The host functions ensure
+that all register operations adhere to the size and count limits specified in
+`VMLimits`, preventing resource abuse and maintaining the integrity of the
+execution environment.
```
miraclx: another register block that needs to be updated

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+
+Storage interactions are also facilitated through host functions. These
+functions provide a key-value storage interface, allowing the guest code to read
+from and write to a persistent storage system. By channeling all storage
+operations through these functions, the runtime can enforce access controls,
+manage resource usage, and potentially implement advanced features like rollback
+or versioning.
```
miraclx:
```suggestion
from and write to a persistent storage system.
```

- access-control (I guess in the context of contexts not reading other context data) isn't managed by the runtime, but by the node
- resource usage (I guess storage usage) for now isn't metered nor is it constrained
- rollback and versioning would also fall outside the purview of the runtime

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+and emit events, which are collected by the runtime for later analysis. These
+functions ensure that logging and event emission adhere to the specified limits,
+preventing potential denial-of-service attacks through excessive logging.
+
+One of the more powerful capabilities provided by the host functions is the
+ability to make external HTTP requests. The `fetch()` function allows the guest
+code to interact with external services, opening up possibilities for complex,
+interconnected applications. However, this function is implemented with careful
+consideration for security, potentially including restrictions on allowed URLs
+or rate limiting to prevent abuse.
```
miraclx: we're unable to do any of these generically without being opinionated, but higher-level libraries like `calimero-sdk-near` unsafely use the `calimero_sdk::env::ext::fetch` method and can apply their service-specific behaviour

so this is out of scope for the runtime

https://github.com/calimero-network/core/blob/a1376c7f823397015fa4d45b7542d6a05c95a1a8/crates/sdk/src/env/ext.rs#L5

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+Error handling and panic management are also key responsibilities of the host
+functions. They provide mechanisms for the guest code to signal errors or panic
+conditions, which are then safely propagated to the runtime environment. This
+includes functions for reporting panics with detailed location information,
+allowing for better debugging and error reporting.
+
+Cryptographic operations, such as keypair creation, are also made available
+through host functions. These provide secure, audited implementations of
+cryptographic primitives to the guest code, ensuring that sensitive operations
+are performed correctly and securely.
```
miraclx: We don't have this

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+defined in the `memory.rs` file. This struct extends Wasmer's base tuning
+capabilities, allowing the Calimero runtime to exert fine-grained control over
+memory allocation and usage. The `WasmerTunables` struct encapsulates both the
+base tuning parameters and a custom `VMConfig`, which together define the memory
+constraints and behaviour for Wasm execution.
+
+One of the key aspects of the memory management system is the enforcement of
+memory limits. The `WasmerTunables` struct uses the `VMLimits` provided by the
+runtime to set strict boundaries on memory usage. This includes setting a
+maximum number of memory pages that can be allocated (`max_memory_pages`), which
```
miraclx: `max_stack_size` too

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+directly controls the total amount of memory available to a Wasm module. By
+enforcing these limits, the runtime prevents potential out-of-memory scenarios
+and ensures that each Wasm module operates within its allocated resources.
+
+The memory management system also implements guard pages, both for static and
+dynamic memory. These guard pages serve as a security measure, creating a buffer
+zone at the edges of allocated memory. This helps prevent potential buffer
+overflow attacks and makes it easier to detect and handle attempts to access
+memory outside of the allocated regions.
```
miraclx: Huh.. did you find this logic inside wasmer?

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+memory access to the guest Wasm code. The `VMLogic` struct includes methods for
+reading from and writing to the Wasm module's memory, ensuring that all memory
+accesses are bounds-checked and secure.
+
+The memory management system also plays a crucial role in the implementation of
+host functions. Many host functions, such as those for reading and writing data,
+interacting with registers, or performing I/O operations, require careful
+management of memory. The system ensures that these operations are performed
+safely, preventing buffer overflows, use-after-free vulnerabilities, and other
+memory-related security issues.
```
miraclx:
```suggestion
management of memory.
```

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+achieves a high degree of flexibility, allowing for different storage
+implementations to be used interchangeably depending on the specific needs of
+the application or deployment environment.
+
+The most basic implementation provided in the runtime is the `InMemoryStorage`
+struct, which uses a `BTreeMap` to store key-value pairs in memory. This
+implementation is particularly useful for testing, development, and scenarios
+where persistent storage is not required. However, the trait-based approach
+allows for easy integration of more sophisticated storage solutions, such as
+disk-based storage, distributed databases, or blockchain-specific storage
```
miraclx: what is meant by "blockchain-specific storage"?

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+systems, without requiring changes to the rest of the runtime code.
+
+One of the key aspects of the storage system is its integration with the
+`VMLogic` struct. `VMLogic` holds a mutable reference to the storage
+implementation, allowing it to mediate all storage interactions between the Wasm
+module and the underlying storage system. This mediation is crucial for
+maintaining control over storage operations, enforcing access restrictions, and
+implementing additional features like metering or transaction management.
```
miraclx:
```suggestion
module and the underlying storage system.
```

crates/runtime/README.md
```diff
@@ -0,0 +1,662 @@
...
+    information about panics, including the panic message and the location
+    (file, line, and column) where the panic occurred. This information is then
+    wrapped in a `HostError::Panic` variant, allowing panics to be handled
+    consistently with other types of errors. This is notable because, of course,
+    it is usually an ambition to avoid panics in Rust, but in this case, they
+    cannot currently be avoided.
+
+Security considerations are paramount in the design of the error handling
+system. Care is taken to ensure that error messages and stack traces do not leak
+sensitive information that could be exploited by malicious actors.
```
miraclx: I don't think there's any such risk or consideration

miraclx (commented):
Nice but quite verbose imo, since this documents the code as implemented, it doesn't leave much room to modify the code without requiring a subsequent doc update

Essentially like freezing the code in-place

Ideally we document high-level interaction of components and leave the rest as implementation detail where the code is the source of truth

There's also multiple references in the doc to certain concepts, where when changed, might be tricky to find all it's references

xilosada:
@miraclx @danwilliams can we prioritize this?

crates/runtime/README.md
danwilliams: Ah, my bad! Is this something we will be adding in...? I would have thought it would be under our remit - do we expect people to always have and provide a keypair?

crates/runtime/README.md
danwilliams: That's a fair point 🙂 Do you think that I should modify it to make this point clear, or remove it? Probably the former?

crates/runtime/README.md
danwilliams: I will remove it; on reflection that didn't end up so great. It seems a bit pointless.

crates/runtime/README.md
danwilliams: That's fair. I will remove that as it is indeed the responsibility of the wider ecosystem to decide what to actually do with the binary data. From a runtime perspective I should constrain the statement. Good catch! 👍

crates/runtime/README.md
danwilliams: I am happy to remove it 🙂 I personally always like to see what kind of errors can be emitted, as it's good to know, but perhaps this is not really the place. It was part of the notes I created in reviewing, and maybe too much leaked in - I can see the argument for excluding this.

crates/runtime/README.md
danwilliams: Fair, I think that's more significance than I was intending - it's just a visual representation for topology rather than implying anything deeper. As a relationship diagram was desired, that's why I put one in 🙂 

I'll remove those entries based on the general points about (de)serialisation responsibility.

crates/runtime/README.md
danwilliams: Ah, okay - I may have misunderstood this part. Is there no intention to be able to do this? I think the rest of my description was accurate? Maybe I should adjust the wording here...? I mean, passing data from host to guest fits what I have written, but I've implied wider reach, which it seems is the error.

crates/runtime/README.md
danwilliams: Ah right, interesting. I hadn't realised that. I will amend 👍

crates/runtime/README.md
miraclx: Keypair creation and it's management happen outside the runtime, the runtime will eventually expose signing operations, but it's missing right now. You can retain it if you mention signing instead.

crates/runtime/README.md
miraclx: If there's some representation that can indicate that generic relationship, then refining is good.. if not, it's not that valuable to have it in.

crates/runtime/README.md
danwilliams: So wait... I don't follow now. I thought in the comment a couple above, you were saying this was not a thing - but - it _is_ a thing? I'm a little confused; I thought I'd got it wrong 😅 

Could you elaborate a little so that I can adjust what I've written?

crates/runtime/README.md
danwilliams: Okay, noted. I have used registers for similar things and thought that this was how it was designed here. I will amend it 👍

crates/runtime/README.md
miraclx: > I personally always like to see what kind of errors can be emitted, as it's good to know

I agree, but Rust is typed so it's pretty easy to find. That, plus.. we have docs.rs

crates/runtime/README.md
danwilliams: With you having pointed out that the runtime is only responsible for the binary payload, this doesn't have any scope anymore, so I will remove it 👍

crates/runtime/README.md
danwilliams: Okay, that's fair 👍

crates/runtime/README.md
danwilliams: Fair, I'm getting the feel some of these points are maybe too technical - I had written this to be technical documentation, for developers to read, but perhaps I've got the level wrong...? From a tech perspective I think knowing there's a self-reference is arguably useful, but otherwise not 🙂

crates/runtime/README.md
miraclx: "inter-function" is confusing, are these functions in the guest space or functions in the host space?

it's a facility for aiding bridging, "inter-function" doesn't give that hint

crates/runtime/README.md
miraclx: I mean, it's like.. I can either talk to you directly, or I can put it in a note, send it to France, tell you it's in France and you ship it back just to read it.

crates/runtime/README.md
miraclx: Well, I think from a tech perspective, you'll be reading the code, and in there the structure is annotated with `#[self_referencing]`

How I think about these docs is in the sense that some random person wants to get an idea why this component exists, and high-level - what it means to have it and how it functions.

crates/runtime/README.md
danwilliams: Mmmm maybe that was a bit superfluous 😄

crates/runtime/README.md
danwilliams: 👍 Will do - I think I need a little more info as per one of the comments threads above, just so I'm clear, but then I can update this.

crates/runtime/README.md
danwilliams: - Access control - okay, that's my bad
- Resource usage - the angle I was going for was "can", but maybe that's not quite relevant here
- Versioning - fair, I was trying to list some benefits and I think I went to far outside the bounds of what is strictly runtime 🙂

crates/runtime/README.md
danwilliams: Okay, that's fair - I did say "potentially", but part of the problem is I don't quite know yet where the line is drawn between some parts 😄 So I will happily remove this 👍

crates/runtime/README.md
danwilliams: I must say I am surprised - it was touched on in another thread, but is it a design choice that we will _not_ provide this? I expected us to be able to. Will we always require a user to supply a keypair? Surely it would ease the process if we allow the option to create one? I'm just keen to understand more about this 🙂

crates/runtime/README.md
danwilliams: 👍 I will mention that too

crates/runtime/README.md
danwilliams: Yes! 😄 Well... sort of. I looked into Wasmer and specifically some of the security aspects, and found that there were some issues with overflows, fixed in v2.3.0 about a year ago. Any buffer overflows are dealt with better now, and I believe Wasm in general is restricted to the data section so cannot overflow buffers beyond that. From what I read, guard pages and buffer zones are in use, so I described them a bit.

Do you think that's okay, or is that going too far given that it's a Wasmer thing and not Calimero specifically? I kinda felt like Wasmer's abilities are Calimero's, in this regard?

crates/runtime/README.md
danwilliams: I'll wait for you to confirm in response to the above comment before I remove this, just in case 😉 Given that Wasmer does provide us with those abilities, and indeed Rust too to some degree.

crates/runtime/README.md
danwilliams: Xabi has been talking to me about the possibilities of linking into blockchain systems from our storage, so I thought I would mention it here. Too much...? Too tenuous?

crates/runtime/README.md
danwilliams: 👍

crates/runtime/README.md
danwilliams: There is definitely that risk and we need to take that consideration 🤔 Does that mean we are currently NOT considering this? I kinda hoped we were... if not, we should! 😅

danwilliams:
@miraclx thanks for the thorough and detailed review! 🙌 I really appreciate it.

It seems that I have misunderstood a couple of points, so thank you for picking up on those. I do have a couple of areas where I am unsure about making a change, or what the change would be. Otherwise:

> Nice but quite verbose imo, since this documents the code as implemented, it doesn't leave much room to modify the code without requiring a subsequent doc update

This was on purpose, as I have see this as Documentation as Code (DoC) to live in the repository. It's entirely possible that my focus here has been wrong 🙂 Is this not aimed at developers wanting to have a technical overview...? I can of course tweak anything necessary, but it was a conscious action to closely mirror the working setup of the code. So yes, there would indeed be the effect of needing to keep it up-to-date when those aspects change.

> Ideally we document high-level interaction of components and leave the rest as implementation detail where the code is the source of truth

Could you elaborate on how you would want this to be achieved? I'm concerned that it would be so high-level as to not serve as an introduction to the codebase (which was a stated goal) - I agree the code is always the source of truth, but in terms of an introduction to the code, how can that be done _without_ some close coupling to what's being introduced? 😅 

> There's also multiple references in the doc to certain concepts, where when changed, might be tricky to find all it's references

That's a fair point, as some concepts do crop up in more than one place, but, for the most part I think they are fairly well segregated?

crates/runtime/README.md
miraclx: the executor's identity is already defined once they join the context, and that's the only identity that matters wrt the runtime

any other keypairs you want to create can be done within the guest environment like we currently do with `rock-paper-scissors`

https://github.com/calimero-network/rock-paper-scissors/blob/de166ecf11d41e963df5facb8a2457edc1bf6320/logic/rock-paper-scissors/src/lib.rs#L63

Although it's use in RPS was an accommodation for lacking runtime identities, which is changing

crates/runtime/README.md
miraclx: The eventual shape is not to be too coupled to any one engine, we always had plans to impl `wasmtime` and consider others.

And following from an earlier conversation this week, that's also going to expand to non-WASM engines.

So it's okay to speak generally about the details of upstream deps, especially since we're not maintaining those pieces and can't follow the changes to keep the doc updated.

crates/runtime/README.md
miraclx: Not quite sure how Wasmer itself prevents buffer overflows or use-after-frees, I suppose if you mean in the sense that pointers within the guest environment can't be dereferenced to read things in the host, then sure.. but that also falls out the scope of the runtime and our memory definitions.

Especially since from the pov of Wasmer, WASM is just a bunch of numbers interacting together, it lacks an understanding of what's a buffer and its length.

Within the bounds of what the runtime has defined as the memory region for the guest, the guest can do whatever it feels like 😅

crates/runtime/README.md
miraclx: Hm, I.. think we can remove it here

crates/runtime/README.md
miraclx: So it comes down to this - `malicious actors`

Execution is local, so the only potential actor is by definition, you.

miraclx:
> Is this not aimed at developers wanting to have a technical overview...?

yes, it is.. but as you mentioned "overview" 😅 

It's already documented, and good work, really.. so you shouldn't have to change much, I would've just preferred to focus on key pieces that are relevant, summarize them high-level, and relationships that may be hard to understand by reading the code

For example, here's NEAR's technical documentation for their runtime

- https://nomicon.io/RuntimeSpec/Runtime
- https://nomicon.io/RuntimeSpec/Components/BindingsSpec

> but in terms of an introduction to the code, how can that be done without some close coupling to what's being introduced? 

it's okay to document what exists, to clarify, it's the verbosity that might make it require significant effort to maintain

for every tiny change, you have to find all references to it in the doc to update them all, that'll be costly

crates/runtime/README.md
danwilliams: As well as removing those entries I have also simplified the graph.

MatejVukosav:
@danwilliams please do not use force push. It is much easier to review only new changes than everything again.

We stil squash all commits before merge so you don't need to worry about it

crates/runtime/README.md
danwilliams: That's not really the angle I was going for - I have removed it, but consider this: we are adding functionality for diagnostic information to be reported. That is what things like stack traces are for... any system that allows stack traces or equivalents to be collected generally has an approach towards sanitisation for privacy. So no, the malicious actor is not the user; it's whoever may receive the diagnostic data.

crates/runtime/README.md
```diff
@@ -0,0 +1,592 @@
...
+    I --> O[External Requests]
+    
+    C -.-> I
+    
+    subgraph "External Environment"
+        AB[Host System]
+    end
+    
+    AB -.-> E
+    AB -.-> O
```
miraclx: wdyt about flipping these? since `Storage` depends on stuff provided by the host, and so does the request, being fulfilled by the host

crates/runtime/README.md
miraclx: Ah, I see what you mean. There's a couple of things to note here:

One is we don't collect stack traces on panic nor include any debug info

https://github.com/calimero-network/core/blob/ed5f17f750e3bbbfb1e3af7648db0d471d4aa62f/Cargo.toml#L103-L104

But we do include location information for where the panic originated, which could potentially leak information about the machine the binary was compiled on.

Not sure how much of an issue this is, but we (could) potentially investigate sanitizing these pieces at compilation.

Because at runtime, it doesn't matter. The information leak is in the wasm blob. So it doesn't even need to be executed to have that issue.

miraclx (approved):
lgtm

Issue #451, opened 2024-07-03
Change default initialization dir


miraclx:
`meroctl init` should default to initializing nodes in `~/.calimero` instead of `~/data`

Issue #450, opened 2024-07-03
Update localnet build script


miraclx:
> PS: we need to update and move this file - https://github.com/calimero-network/core/blob/0f821050a556573197bdba655c4e6c58d5d99f6d/crates/node/gen_localnet_configs.sh

_Originally posted by @miraclx in https://github.com/calimero-network/core/issues/405#issuecomment-2206754173_

PR #454, opened 2024-07-05
chore: update localnet build script

petarjuki7:
# Update default dir and move build script

This solves #450 and #451 

## Summary

The build script is updated so it builds the new `meroctl` binary, and uses new initialization methods. 
Also default directory for `meroctl` changed from `/data` to `/.calimero`

## Test plan

running `./crates/meroctl/gen_localnet_configs.sh 3` from the root of this repository creates three nodes (`node1`, `node2`, `node3`) in the `~./calimero` directory

### DIFF
--- a/crates/meroctl/gen_localnet_configs.sh
+++ b/crates/meroctl/gen_localnet_configs.sh
@@ -2,6 +2,9 @@
 
 set -o pipefail
 
+# Set default NODE_HOME 
+: ${NODE_HOME:="$HOME/.calimero"}
+
 # Check if an argument was provided
 if [ $# -eq 0 ]; then
   echo "Please provide the number of local nodes argument."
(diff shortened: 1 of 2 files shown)


MatejVukosav:
Pls update docs accordingly before merge

petarjuki7:
> Pls update docs accordingly before merge

These updates don't reflect anything in the docs, no updated needed.
Except for the `gen_localnet_configs.sh`, but it wasn't documented before...
Should we document it publicly or is it more of an internal tool? (cc: @miraclx)

crates/meroctl/gen_localnet_configs.sh
```diff
@@ -11,15 +11,15 @@ fi
...
+cargo build --bin meroctl
 
 # Iterate in a loop N times
 for ((i = 1; i <= N; i++)); do
   node_home="$HOME/.calimero/node$i"
   echo "\x1b[1;36m(i)\x1b[39m Initializing Node $i at \x1b[33m$node_home\x1b[0m"
   rm -rf "$node_home"
   mkdir -p "$node_home"
-  ./target/debug/calimero-node --home "$node_home" \
+  ./target/debug/meroctl --home "$HOME/.calimero" --node-name "node$i" \
```
miraclx: this is implied

```suggestion
  ./target/debug/meroctl --node-name "node$i" \
```

MatejVukosav:
@petarjuki7 I think it affects docs https://calimero-network.github.io/getting-started/setup

crates/meroctl/gen_localnet_configs.sh
miraclx: Unless you make it further customizable by defining an intermediary var that defaults to `$HOME`

```bash
: ${NODE_HOME:="$HOME/.calimero"}
```

with

```suggestion
  ./target/debug/meroctl --home "$NODE_HOME" --node-name "node$i" \
```

petarjuki7:
@MatejVukosav in the examples we are creating our own data folder and specifying it with `--home`, this `~/.calimero` folder is used only when there is no `--home` flag provided.
Maybe I can add in the docs that if the user doesn't specify `--home`, the node config will be saved in the default dir `~/.calimero`?

miraclx (approved):

PR #462, opened 2024-07-10
chore: set docs line length limit

MatejVukosav:
Add prettier to format md/mdx files and husky for pre-commit hooks in root. 

Try it manually from the root:
`pnpm format:md`

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+pnpm format:md
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,13 +4,14 @@ about: Create a report to help us improve
 title: ''
 labels: ''
 assignees: ''
-
 ---
 
-**Describe the bug**
+## Describe the bug
+
 A clear and concise description of what the bug is.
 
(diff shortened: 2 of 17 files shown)


.husky/pre-commit
miraclx, line 1: Can we somehow limit this to run on just md/mdx changes? so Rust changes won't trigger this hook.

fbozic: > Add prettier to format md/mdx files

I understood we are only doing md/mdx files with this 😄 
If I am wrong, yes please do as mira asked

MatejVukosav: It is pre-commit hook. It will be triggered every time there is a commit. That's the purpose 😄 
Not sure how or even why, alter that.

miraclx: Something like this should suffice - https://stackoverflow.com/a/76811469/9806233

MatejVukosav: Ok, now I understand what you wanted. That is just check inside the hook script. It will still be executed. I have used that approach already for folders https://github.com/calimero-network/core-app-template/blob/master/.husky%2Fpre-commit 

I will add extensions check next week when I return :)

miraclx: Amazing, thank you!

PR #463, opened 2024-07-10
feat(node): persist transactions on coordinator and verify transaction chain

fbozic:
This PR will allow the coordinator to participate in the catchup reply process.
- Persist transactions on the coordinator node as well (without actually executing transaction)
- Validate transaction chain on coordinator node before sending Confirmation/Rejection
- Validate transaction chain on all nodes before executing/persisting transaction
- Move handling of a specific network event to dedicated methods and improve error handling

### DIFF
--- a/crates/node-primitives/src/lib.rs
+++ b/crates/node-primitives/src/lib.rs
@@ -58,6 +58,5 @@ pub enum MutateCallError {
     },
     NoConnectedPeers,
     TransactionRejected,
-    ExecutionError,
     InternalError,
 }
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -902,9 +902,10 @@ impl Node {
 
         if context.last_transaction_hash != transaction.prior_hash {
             error!(
+                context_id=%transaction.context_id,
                 %transaction_hash,
                 prior_hash=%transaction.prior_hash,
-                "Transaction from the pool doesn't build on last executed transaction",
+                "Transaction from the pool doesn't build on last transaction",
             );
             return Err(calimero_node_primitives::MutateCallError::TransactionRejected);
         }
@@ -914,7 +915,7 @@ impl Node {
             .await
             .map_err(|e| {
                 error!(%e, "Failed to execute transaction");
-                calimero_node_primitives::MutateCallError::ExecutionError
+                calimero_node_primitives::MutateCallError::InternalError
             })?;
 
         Ok(outcome)


crates/node/src/lib.rs
```diff
@@ -496,100 +496,141 @@ impl Node {
...
+                %their_peer_id,
+                "Observed subscription to unknown context, ignoring.."
+            );
+            return Ok(());
+        };
+
+        if self
+            .ctx_manager
+            .is_application_installed(&context.application_id)
+        {
```
miraclx: I think we no longer need this, yeah?

crates/node/src/lib.rs
```diff
@@ -496,100 +496,141 @@ impl Node {
...
+                    },
+                ));
+        }
+
+        Ok(())
+    }
+
+    async fn handle_message(&mut self, message: libp2p::gossipsub::Message) -> eyre::Result<()> {
+        let Some(source) = message.source else {
+            return Ok(());
```
miraclx, line 597: let's log here to see when this comes up, if ever

crates/node/src/lib.rs
```diff
@@ -601,38 +642,37 @@ impl Node {
...
-                    types::PeerAction::TransactionConfirmation(confirmation) => {
-                        // todo! ensure this was only sent by a coordinator
-                        self.execute_in_pool(
-                            confirmation.context_id,
-                            confirmation.transaction_hash,
+                            .await
+                        {
+                            self.delete_transaction(transaction.context_id, transaction_hash)?;
+                            eyre::bail!("Failed to push confirmation action: {}", err);
+                        };
```
miraclx: why not persist after push?

crates/node/src/lib.rs
```diff
@@ -1110,36 +1178,103 @@ impl Node {
...
+                                            transaction.context_id,
+                                            types::PeerAction::TransactionRejection(
+                                                types::TransactionRejection {
+                                                    context_id: transaction.context_id,
+                                                    transaction_hash,
+                                                },
+                                            ),
+                                        )
+                                        .await?;
+                                    }
```
miraclx: can we deduplicate this with it's counterpart above?

crates/node/src/lib.rs
```diff
@@ -1110,36 +1178,103 @@ impl Node {
...
+                                            transaction_hash,
+                                        )?;
+
+                                        if let Err(err) = self
+                                            .push_action(
+                                                transaction.context_id,
+                                                types::PeerAction::TransactionConfirmation(
+                                                    types::TransactionConfirmation {
+                                                        context_id: transaction.context_id,
+                                                        nonce: self.nonce,
```
miraclx: PS: we still need to increment the nonce

crates/node/src/lib.rs
```diff
@@ -886,6 +901,25 @@ impl Node {
         Ok(outcome)
     }
 
+    async fn reject_from_pool(
+        &mut self,
+        hash: calimero_primitives::hash::Hash,
+    ) -> eyre::Result<()> {
+        while let Some(transaction_pool::TransactionPoolEntry { transaction, .. }) =
+            self.tx_pool.remove(&self.last_tx)
```
miraclx: ah, we're gonna need to remove last_tx from `Node` and read it from `ContextMeta` now

crates/node/src/lib.rs
```diff
@@ -852,10 +855,22 @@ impl Node {
...
+        let Some(ctx_meta) = handle.get(&calimero_store::key::ContextMeta::new(
+            transaction.context_id,
+        ))?
+        else {
+            eyre::bail!("Context '{}' not found", transaction.context_id);
         };
+        let context = calimero_primitives::context::Context {
+            id: transaction.context_id,
+            application_id: ctx_meta.application_id.to_string().into(),
+        };
```
miraclx: why inline the function? .._

crates/node/src/lib.rs
```diff
@@ -852,10 +855,22 @@ impl Node {
...
+
+        if ctx_meta.last_transaction_hash != *transaction.prior_hash {
```
miraclx: _.. ah.. because of `last_transaction_hash`, why not add `last_transaction_hash` to `Context`?

crates/node/src/lib.rs
```diff
@@ -852,10 +855,22 @@ impl Node {
...
+            eyre::bail!("Context '{}' not found", transaction.context_id);
```
miraclx: wrong message

crates/node/src/lib.rs
```diff
@@ -886,6 +901,25 @@ impl Node {
...
     }
 
+    async fn reject_from_pool(
+        &mut self,
+        hash: calimero_primitives::hash::Hash,
+    ) -> eyre::Result<()> {
+        while let Some(transaction_pool::TransactionPoolEntry { transaction, .. }) =
+            self.tx_pool.remove(&self.last_tx)
+        {
+            self.last_tx = transaction.prior_hash;
```
miraclx: although, this seems weird.. `last_tx` isn't set until the transaction is executed, so it won't be in the pool

crates/node/src/lib.rs
fbozic: not sure why, overthinking on my side

crates/node/src/lib.rs
fbozic: done, looks wonky because I have to perform 2 validations in the catchup flow. or I have to pass a bool flag to the method.

crates/node/src/lib.rs
fbozic: not really, I've renamed `last_tx` from `Node` to `last_pending_transaction_hash` so it is more descriptive. 

e.g. in case we have fired 5 transactions and we get a rejection for the 3rd transaction. we need to move `last_pending_transaction_hash` to point to 2nd transaction and we need to drop 3 transactions from the pool. because we still might get a confirmation for 2nd transaction so we can build upon it.

crates/node/src/lib.rs
fbozic: question actually, we have a similar code here: https://github.com/calimero-network/core/blob/cb1bd854ef0d4fe822dacfb01b14f37a21fc3931/crates/node/src/lib.rs#L772-L794

I think we already discussed this, do you remember why we ended up with insert/delete instead of insert after the push?

crates/node/src/lib.rs
fbozic: okay, my bad, I've removed `last_tx` from the `Node`.
the comment above would make sense if we would have `tx_pool` per context (or we would track `last_transaction_hash` per context)

crates/node/src/lib.rs
miraclx: it's quite different, the current one is in-memory and so, ephemeral.. if the node is killed before the confirmation is received, we don't retain those transactions, so there's no risk of holding on to data that hasn't formed the canonical state

and secondly, we were able to consolidate tx_hash generation by passing it to the pool, which also uses it

crates/node/src/lib.rs
```diff
@@ -1218,100 +1257,54 @@ impl Node {
                         status,
                     } in response.transactions
                     {
-                        let is_transaction_valid = last_transaction_hash == transaction.prior_hash;
+                        if last_transaction_hash != transaction.prior_hash {
+                            continue;
```
miraclx: how about `bail!`-ing here?

miraclx (approved):
lgtm. thanks for being thorough!! 🎉

crates/node-primitives/src/lib.rs
```diff
@@ -57,5 +57,7 @@ pub enum MutateCallError {
         application_id: calimero_primitives::application::ApplicationId,
     },
     NoConnectedPeers,
+    TransactionRejected,
```
miraclx, line 60: Y'know, thinking about it.. rejected transactions aren't the client's fault, the node should be able to resubmit with new knowledge of the current transaction hash

crates/node/src/lib.rs
```diff
@@ -793,37 +880,44 @@ impl Node {
...
+            return Err(calimero_node_primitives::MutateCallError::TransactionRejected);
         }
 
-        Ok(Some(()))
+        let outcome = self
+            .execute_transaction(context, transaction, transaction_hash)
+            .await
+            .map_err(|e| {
+                error!(%e, "Failed to execute transaction");
+                calimero_node_primitives::MutateCallError::ExecutionError
```
miraclx: non-contextual errors can just be `InternalError`

crates/node-primitives/src/lib.rs
fbozic: yeah, that would be nice I agree. but since I am going on PTO I will open an issue to track that feature and leave rejection propagation back to the client (so the client is aware at least of what happened).

crates/node/src/lib.rs
fbozic: yeah, forgot that I've left it like this. it makes no sense to continue forward.

crates/node-primitives/src/lib.rs
fbozic: https://github.com/calimero-network/core/issues/466

crates/node/src/lib.rs
fbozic: talked on Slack, linked code is safe to refactor as well

PR #465, opened 2024-07-12
Moved structs to server-primitives

petarjuki7:
# Meroctl fix - short description

## Summary

Moved shared structs from `meroctl` and admin handler methods to `server-primitives`.
Node no longer subscribes to `application_id` when installing the app in `dev-mode`.

### DIFF
--- a/crates/meroctl/src/cli/context/create.rs
+++ b/crates/meroctl/src/cli/context/create.rs
@@ -1,8 +1,3 @@
-use calimero_primitives::application::ApplicationId;
-use calimero_server_primitives::admin::{
-    CreateContextRequest, CreateContextResponse, InstallDevApplicationRequest,
-    ListApplicationsResponse,
-};
 use camino::Utf8PathBuf;
 use clap::{Args, Parser};
 use libp2p::Multiaddr;
@@ -91,14 +86,15 @@ async fn create_context(
     }
 
     let url = multiaddr_to_url(base_multiaddr, "admin-api/dev/contexts")?;
-    let request = CreateContextRequest {
+    let request = calimero_server_primitives::admin::CreateContextRequest {
         application_id: calimero_primitives::application::ApplicationId(application_id),
     };
 
     let response = client.post(url).json(&request).send().await?;
 
     if response.status().is_success() {
-        let context_response: CreateContextResponse = response.json().await?;
+        let context_response: calimero_server_primitives::admin::CreateContextResponse =
+            response.json().await?;
         let context = context_response.data;
 
         println!("Context created successfully:");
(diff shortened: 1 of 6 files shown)


crates/meroctl/src/cli/context/create.rs
```diff
@@ -1,40 +1,20 @@
 use calimero_primitives::application::ApplicationId;
-use calimero_primitives::context::Context;
-use calimero_server_primitives::admin::{ApplicationListResult, InstallDevApplicationRequest};
+use calimero_server_primitives::admin::{
+    CreateContextRequest, CreateContextResponse, InstallDevApplicationRequest,
+    ListApplicationsResponse,
+};
```
fbozic: Some time ago we had an agreement to use fully qualified names as the preferred way. Exceptions are okay when they make sense (https://github.com/calimero-network/core/blob/56ee23d5a17b81b0e558f059a1bda74c3ea9e158/crates/server/src/jsonrpc/mutate.rs#L3). Here I think it makes more sense to fully qualify requests objects as well.

crates/meroctl/src/cli/context/list.rs
```diff
@@ -1,22 +1,11 @@
-use calimero_primitives::context::Context;
+use calimero_server_primitives::admin::GetContextsResponse;
```
fbozic: And here

crates/server-primitives/src/admin.rs
```diff
@@ -1,3 +1,5 @@
+use calimero_primitives::application::ApplicationId;
+use calimero_primitives::context::Context;
 use calimero_primitives::identity::WalletType;
```
fbozic: And here (example: https://github.com/calimero-network/core/blob/56ee23d5a17b81b0e558f059a1bda74c3ea9e158/crates/server-primitives/src/jsonrpc.rs#L1-L2)

crates/server/src/admin/handlers/context.rs
```diff
@@ -6,7 +6,10 @@ use axum::{Extension, Json};
 use calimero_primitives::application::ApplicationId;
 use calimero_primitives::context::{Context, ContextId};
 use calimero_primitives::identity::{ClientKey, ContextUser};
```
fbozic: And here (I know this is not your code but it would be nice to make it consistent finally)

crates/server/src/admin/handlers/context.rs
```diff
@@ -6,7 +6,10 @@ use axum::{Extension, Json};
...
-use calimero_server_primitives::admin::ContextStorage;
+use calimero_server_primitives::admin::{
+    ContextList, ContextResponse, ContextStorage, CreateContextRequest, CreateContextResponse,
+    GetContextsResponse,
+};
```
fbozic: IMO this is okay since we are in the admin server

xilosada:
This PR was opened 5 days ago. What is missing to get it merged? @petarjuki7  @fbozic is gonna be out next week

PR #470, opened 2024-07-12
fix: catchup and state reconciliation bug fixes

fbozic:
* node - fix how `last_transaction_hash` is moved backwards during reply chain construction
* node - remove subscription triggered catchup due to many timeout errors
* node - perform catchup on transaction rejection event
* node - remove executed transactions from the `tx_pool` during catchup
* node - fix coordinator catchup criteria
* node - sanitize error struct
* node - implement list context transactions in the interactive CLI
---
* store - use `store.get` and check if value is present as `store.has` implementation
---
* context - check if the application is already installed during installation
* context - check if the application is installed before creating new context
---
* network - fix stream length codec initialization bug
* network - implement timeout on message receive

### DIFF
--- a/crates/node/src/catchup.rs
+++ b/crates/node/src/catchup.rs
@@ -65,7 +65,10 @@ impl Node {
 
         let application_id = context.application_id.clone();
 
-        if request.application_id.is_none() || application_id != request.application_id.unwrap() {
+        if request
+            .application_id
+            .map_or(false, |id| id != application_id)
+        {
             let application_version = self
                 .ctx_manager
                 .get_application_latest_version(&application_id)?;


crates/context/src/lib.rs
```diff
@@ -415,29 +413,34 @@ impl ContextManager {
             eyre::bail!("Release hash does not match the hash of the downloaded file");
         }
 
-        Ok(())
+        Ok(Some(()))
```
miraclx: `Ok(true)`?

crates/node/src/catchup.rs
```diff
@@ -58,7 +65,7 @@ impl Node {
 
         let application_id = context.application_id.clone();
 
-        if request.application_id.is_none() || application_id == request.application_id.unwrap() {
+        if request.application_id.is_none() || application_id != request.application_id.unwrap() {
```
miraclx: either of these get the same result without invoking `unwrap`

```suggestion
        if matches!(request.application_id, Some(id) if application_id == id) {}
```

```suggestion
        if Some(application_id) == request.application_id {}
```

crates/node/src/lib.rs
```diff
@@ -428,6 +428,37 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
+                        let key = calimero_store::key::ContextTransaction::new(
+                            context_id,
+                            [0; 32].into(),
+                        );
+
+                        println!("{IND} {c1:44} | {c2:44}", c1 = "Hash", c2 = "Prior Hash");
+
+                        for (k, v) in &mut handle.iter(&key)?.entries() {
+                            let entry = format!(
+                                "{c1:44} | {c2:?}",
```
miraclx: is it more useful to `Debug` than `Display` the prior hash here?

```suggestion
                                "{c1:44} | {c2}",
```

crates/server/src/lib.rs
```diff
@@ -69,7 +69,7 @@ pub async fn start(
     {
         if let Some((path, handler)) = jsonrpc::service(&config, server_sender.clone())? {
             app = app.route(path, handler);
-            app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
+            // app = app.layer(middleware::auth::AuthSignatureLayer::new(store.clone()));
```
miraclx: don't forget this

crates/store/src/db/rocksdb.rs
```diff
@@ -37,11 +37,7 @@ impl Database for RocksDB {
     }
 
     fn has(&self, col: Column, key: Slice) -> eyre::Result<bool> {
-        let cf_handle = self.try_cf_handle(&col)?;
-
-        let exists = self.db.key_may_exist_cf(cf_handle, key.as_ref());
-
-        Ok(exists)
+        self.get(col, key).map(|value| value.is_some())
```
miraclx: make this `self.db.key_may_exist_cf(..) && self.get(..).map(..)`

since we know for certain when it doesn't exist, but are not quite sure if it does

so we only check to be sure when it may exist

miraclx (approved):
lgtm

crates/node/src/catchup.rs
fbozic: With second approach I have to clone `application_id` that is being wrapped in `Some`

fbozic: And with the first approach I also have to clone `application_id`

fbozic: I've managed to remove `unwrap` with the third approach (`map_or`)😄

miraclx: > With second approach I have to clone application_id that is being wrapped in Some

nothing a little `as_ref` won't fix

```suggestion
        if Some(&application_id) == request.application_id.as_ref() {}
```

> And with the first approach I also have to clone application_id

weird, works on my end

> I've managed to remove unwrap with the third approach (map_or)😄

noice..

PR #471, opened 2024-07-15
docs(context): add context crate readme

miraclx:
Documents the `context` crate

For now, we're leaving out encryption, in-state representation, multiple-identities on a single node, etc.

We'll revisit when we have implementations of these pieces.

### DIFF
--- a/crates/context/README.md
+++ b/crates/context/README.md
@@ -34,7 +34,7 @@ The unfortunate downside of this approach is that we're unnecessarily executing 
 
 ## Context Membership
 
-The membership of a context is a list of member identifiers. Each member is identified by a 32-byte key, which represents the public key of the member.
+This is not implemented at this time.
 
 ### Invitations
 
--- a/crates/context/README.md
+++ b/crates/context/README.md
@@ -12,7 +12,7 @@
 
 A context in Calimero is an instance of a deployed application where members share a synchronized state.
 
-A context is identified by a 32-byte key, which is randomly generated when the context is created. As of now, the key is not necessarily cryptographically secure, nor is it a dependency of cryptographic operations, but it is unique enough to mitigate collisions. Worth noting that this is likely to change in the future.
+Contexts are identified by a 32-byte key, which is randomly determined when each context is created. As of now, this key is not necessarily cryptographically secure, nor is it a dependency of cryptographic operations, but it is unique enough to mitigate collisions. Worth noting that this is likely to change in the future.
 
 All state mutations within a context performed by members will be broadcasted to all other members.
 
--- a/crates/context/README.md
+++ b/crates/context/README.md
@@ -23,7 +23,7 @@ The state of a context is a key-value store, where the key is a 32-byte identifi
 
 ## State mutations
 
-A state mutation is an operation that changes the state of the context. As of the time of writing, state mutations are represented as transactions that define the application method to be called, and it's input (arguments). The method is called on the application, which is a WebAssembly module, and the output is the new state of the context. We make no assumptions about the application, nor it's inputs and outputs, as long as clients of each context agree on the schema & serialization method.
+A state mutation is an operation that changes the state of the context. As of the time of writing, state mutations are represented as transactions that define the application method to be called, and it's input (arguments). We make no assumptions about the application, nor it's inputs and outputs, as long as it's clients agree on the schema & serialization method.
 
 All members of the context, following the reception of a transaction, will execute the transaction on their local copy of the state to advance it to the new state.
 
--- a/crates/context/README.md
+++ b/crates/context/README.md
@@ -1,7 +1,6 @@
 # Calimero Contexts
 
 - [Context Lifecycle](#context-lifecycle)
-- [Administrative Operations](#administrative-operations)
 - [Context State](#context-state)
 - [State Mutations](#state-mutations)
 - [Context Membership](#context-membership)
@@ -15,22 +14,15 @@ A context in Calimero is an instance of a deployed application where members sha
 
 A context is identified by a 32-byte key, which is randomly generated when the context is created. As of now, the key is not necessarily cryptographically secure, nor is it a dependency of cryptographic operations, but it is unique enough to mitigate collisions. Worth noting that this is likely to change in the future.
 
-All operations within a context performed by members will be broadcasted to all other members. This includes:
-
-- State mutations
-- Membership changes (invitation, leaving, kickouts)
+All state mutations within a context performed by members will be broadcasted to all other members.
 
 ## Context Lifecycle
 
 Following the creation of a context, the only member is the creator. The creator can invite other members to join the context, and members can leave the context at any time.
 
-## Administrative Operations
-
-Application authors can define gated behaviour for administrative operations, but by default all members have the same rights. Which, depending on the application, can be an undesirable behaviour.
-
 ## Context State
 
-The state of a context is a key-value store, where the key is a 32-byte identifier and the value is a byte array. The state is synchronized across all members of the context, and all state mutations are broadcasted to all members.
+The state of a context is a key-value store, where the key is a 32-byte identifier and the value is a byte array. The state is synchronized across all members of the context, since all state mutations are broadcasted to all members.
 
 ## State Mutations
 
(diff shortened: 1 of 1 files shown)


crates/context/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+All members of the context, following the reception of a transaction, will execute the transaction on their local copy of the state to advance it to the new state.
+
+The unfortunate downside of this approach is that we're unnecessarily executing the same transaction on all nodes without using that information to achieve trustless consensus. We're currently working on a solution to this problem, which reimagines consensus.
+
+## Consensus
+
+As of the time of writing, we're using a simple consensus mechanism, where all members of the context must agree on the order of transactions. This is achieved by broadcasting transactions to all members, who queue them up in a pool, waiting for a confirmation from an elected coordinator. This coordinator is responsible for ordering transactions, which ensures that all members of the context eventually reach the same state.
+
+The coordinator also stores the transaction history, and so, can reject transactions that have already been processed as well as ones that don't reference the latest state.
```
fbozic: I think this part makes more sense as a part of node crate doc, since the logic is there.

crates/context/README.md
```diff
@@ -0,0 +1,60 @@
+# Calimero Contexts
+
```
fbozic, line 2: Can we add TOC?

fbozic: It would be nice to have a consistent layout of all crate docs.

crates/context/README.md
```diff
@@ -0,0 +1,65 @@
...
+- [Context State](#context-state)
+- [State Mutations](#state-mutations)
+- [Context Membership](#context-membership)
+  - [Invitations](#invitations)
+  - [Leaving](#leaving)
+- [Encryption](#encryption)
+- [Application Upgrade](#application-upgrade)
+- [Context Deletion](#context-deletion)
+
+A context in Calimero is an instance of a deployed application members of which share a synchronized state.
```
MatejVukosav: Can you simplify this intro? 

In terminology we have this `Context is the core of the Calimero ecosystem. It is an application specific network designed to enable direct communication between users, eliminating the need for intermediaries`

More questions than answers arise from this which we should prevent:
- where are members deployed
- are they application members or context members
- is there then a global state that they all share

crates/context/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+A context is identified by a 32-byte key, which is randomly generated when the context is created. As of now, the key is not necessarily cryptographically secure, nor is it a dependency of cryptographic operations, but it is unique enough to mitigate collisions. Worth noting that this is likely to change in the future.
+
+All operations within a context performed by members will be broadcasted to all other members. This includes:
+
+- State mutations
+- Membership changes (invitation, leaving, kickouts)
```
MatejVukosav: Do we have membership changes right now? If not, lets remove it

crates/context/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Context Lifecycle
+
+Following the creation of a context, the only member is the creator. The creator can invite other members to join the context, and members can leave the context at any time.
+
+## Administrative Operations
+
+Application authors can define gated behaviour for administrative operations, but by default all members have the same rights. Which, depending on the application, can be an undesirable behaviour.
```
MatejVukosav: "Application authors can define gated behaviour for administrative operations" <- do we really have this atm?

crates/context/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+The unfortunate downside of this approach is that we're unnecessarily executing the same transaction on all nodes without using that information to achieve trustless consensus. We're currently working on a solution to this problem, which reimagines consensus.
+
+## Context Membership
+
+The membership of a context is a list of member identifiers. Each member is identified by a 32-byte key, which represents the public key of the member.
+
+### Invitations
+
+Bob wants to join a context. He shares his public key with Alice, who is already a member of the context. Alice then broadcasts an invitation to all members of the context, which is made up of Bob's public key. Following this invitation, Alice shares the context ID with Bob. At this point, Bob can make a catchup request, which will be responded to by any member of the context who has some of the state and now knows about Bob. Eventually, Bob receives all transactions required to make up the complete state, at which point he is considered a full member of the context. And can proceed to make mutations.
```
MatejVukosav: Where can I see that implementation? Join context requires only `contextId`

crates/context/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Leaving
+
+Bob wants to leave a context. He broadcasts a leave request to all members of the context, which is made up of his public key. Following this request, Bob is removed from the membership list, and all members will stop broadcasting state mutations to Bob.
```
MatejVukosav: Where is context leaving implemented? `context/src/lib.rs` does not have `leave` or similar methods

crates/context/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+As of the time of writing, all messages are sent in plaintext. But we're currently working on encrypting all messages using distinct, yet, deterministic keys by employing the double-ratchet algorithm, which provides forward secrecy, and post-compromise security. We'll revise this document once this, among other features, is implemented.
+
+## Application Upgrade
+
+As noted [above](#administrative-operations), application authors can define gated behaviour for administrative operations. This includes the ability to upgrade the application. Subsequent transactions made by members from the old version of the application will be rejected by the new version.
+
+## Context Deletion
+
+There's no way to "delete a context" for all members outside of the scope of all members leaving it.
```
MatejVukosav:
```
    pub async fn delete_context(
        &self,
        context_id: &calimero_primitives::context::ContextId,
    ) -> eyre::Result<bool> {
        let mut handle = self.store.handle();

        let key = calimero_store::key::ContextMeta::new(*context_id);

        if !handle.has(&key)? {
            return Ok(false);
        }

        handle.delete(&key)?;

        self.unsubscribe(context_id).await?;

        Ok(true)
    }
```

So this does not delete context?

miraclx: That removes the context from your local node, it does not "delete" it from existence

crates/context/README.md
miraclx: Outside of the scope of what's here for the eventual shape, there's not much material to document about the current implementation of a context that is worthy of a doc, it'll all stay dynamic for a while

Unlike the other pieces

MatejVukosav: So, do we have it right now or not? Lets not write things that will eventually come.

crates/context/README.md
MatejVukosav: It is then more of `leave_context`. Delete is misleading as I as an owner could assume that it does not exists anymore

miraclx: The issue is thinking anything wrt the node happens globally, when we're starting from the premise of local-first.

There's no such thing as global actions / global state, everything is local and node-oriented.

You can certainly create a context that has already been created before.

MatejVukosav: But shouldn't delete context allow me as an owner to kick everyone from the context and then close it?  Others can then create new ones if they want.

We can continue this discussion outside this PR as it is change of concept and not related to doc itself.

miraclx: > kick everyone from the context

you can only perform actions as long as other members validate it, other members can choose to ignore your transaction and continue on with their fork, forks of the same context can exist independently so long as their histories are divergent, this stays the same even with the new state diff model

crates/context/README.md
miraclx: In that case, there's no value to documenting this piece (yet). What we have now (context only, non-networking, non-storage) is barely half a page.

MatejVukosav: Then let's have half a page. We must be transparent with what is available. The roadmap will answer what will come.

crates/context/README.md
miraclx: > where are members deployed

I mention not long after:

> All operations within a context performed by members will be broadcasted to all other members. This includes:
>
> - State mutations
> - Membership changes (invitation, leaving, kickouts)

Indicating it's local

> are they application members or context members

> is an "instance" of a deployed application

is that vague?

> is there then a global state that they all share

hm, that's implied, is it not? at least by the section quoted above

> - State mutations

in the sense that any operation executed on any node will be broadcasted to the rest

MatejVukosav: Later it is explained well and pieces come together. But as this is the first sentence all those info is not known. That is only my concern

crates/context/README.md
miraclx: > do we really have this atm?

depends on how you define this

do we have applications that do this right now? no, well kinda, with RPS

can app developers do this if they wanted to? yep

do we provide easy-to-use mechanisms to facilitate this? also no

I'll take it out anyway.

crates/context/README.md
miraclx: reworded

crates/context/README.md
miraclx: reworded

crates/context/README.md
miraclx: removed

crates/context/README.md
miraclx: removed

crates/context/README.md
miraclx: simplified

MatejVukosav (approved):
Lgtm

Issue #307, opened 2024-05-20
Optimize app blob reads


miraclx:
We can `mmap` apps from disk to avoid buffering the whole thing in memory, it's immutable anyway so it will never be copied into memory.

https://github.com/RazrFalcon/memmap2-rs

PR #472, opened 2024-07-15
docs: Add store crate README.md

danwilliams:
Added a README for the store, which covers the main areas and operations.

### DIFF
--- a/crates/store/README.md
+++ b/crates/store/README.md
@@ -1,0 +1,802 @@
+# Calimero store
+
+[RocksDB]: https://rocksdb.org/
+
+  - [Introduction](#introduction)
+  - [Core components](#core-components)
+    - [Database interface](#database-interface)
+    - [Key structure](#key-structure)
+    - [Data types and entries](#data-types-and-entries)
+    - [Layers](#layers)
+    - [Transactions](#transactions)
+    - [Iterators](#iterators)
+  - [Operation flow](#operation-flow)
+  - [Key concepts](#key-concepts)
+  - [Advanced features](#advanced-features)
+  - [Detailed component descriptions](#detailed-component-descriptions)
+    - [Database interface](#database-interface-1)
+    - [Key structure](#key-structure-1)
+    - [Data types and entries](#data-types-and-entries-1)
+    - [Layers](#layers-1)
+    - [Transactions](#transactions-1)
+    - [Iterators](#iterators-1)
+
+## Introduction
+
+The store library is a versatile and efficient key-value storage system designed
+for managing application state and transactions. It provides a flexible and
+extensible architecture for storing, retrieving, and manipulating data across
+different storage backends.
+
+Key features of the store library include:
+
+  - Support for potentially multiple storage backends (currently just
+    [RocksDB][])
+  - Transactional operations with atomic commits
+  - Flexible key-value structure with custom data types
+  - Iteration and range queries
+  - Temporal layers for augmenting/adapting lower-level database functionality,
+    allowing, for example, atomic transactions, caching, etc.
+
+## Core components
+
+### Database interface
+
+The `Database` trait defines the core interface for interacting with the
+underlying storage system. It provides methods for basic CRUD operations,
+iteration, and transaction application.
+
+Key operations:
+
+  - `open()`: Initialise the database with a given configuration
+  - `has()`, `get()`, `put()`, `delete()`: Basic key-value operations
+  - `iter()`: Create an iterator for range queries
+  - `apply()`: Apply a transaction to the database
+
+The `RocksDB` struct implements this trait, providing a concrete implementation
+using the [RocksDB][] storage engine.
+
+### Key structure
+
+The library uses a flexible key structure defined by the `Key` type and
+associated traits:
+
+  - `KeyComponent`: Defines a part of a key with a specific length
+  - `KeyComponents`: Combines multiple `KeyComponent`s into a complete key
+  - `AsKeyParts`: Converts a key into its column and components
+  - `FromKeyParts`: Constructs a key from its components
+
+This structure allows for efficient encoding and decoding of complex keys,
+supporting various data models.
+
+### Data types and entries
+
+The `DataType` trait defines how values are serialised and deserialised. The
+`Entry` trait combines a key with its associated data type, providing a complete
+representation of a stored item.
+
+Predefined entry types include:
+
+  - `ContextMeta`: Metadata for a context
+  - `ContextState`: State data for a context
+  - `ContextIdentity`: Identity information for a context
+  - `ContextTransaction`: Transaction data for a context
+  - `Generic`: A generic key-value pair
+
+### Layers
+
+The library implements a layered architecture for managing data access and
+modifications:
+
+  - `Layer`: Base trait for all layers
+  - `ReadLayer`: Provides read-only access to the underlying data
+  - `WriteLayer`: Extends `ReadLayer` with write operations
+
+### Transactions
+
+The `Transaction` struct represents a set of operations to be applied
+atomically. It supports putting and deleting entries, as well as merging with
+other transactions.
+
+### Iterators
+
+The library provides a flexible iteration system:
+
+  - `Iter`: Base iterator type
+  - `DBIter`: Trait for database-specific iterators
+
+## Operation flow
+
+  1. Initialise the store with a specific backend (e.g., [RocksDB][])
+  2. Create a `StoreHandle` to interact with the store
+  3. Perform read/write operations directly on the store
+  4. For atomic write operations:
+       - Create a `Temporal` layer
+       - Accumulate changes in the temporal layer
+       - Commit changes to the underlying store
+  5. Use transactions for atomic operations across multiple keys
+  6. Utilise iterators for range queries and scans
+
+```mermaid
+graph TD
+    A[Application] --> B[StoreHandle]
+    B --> C[Store]
+    C --> D[Database Backend]
+    D --> E[RocksDB and others]
+    B --> F
+    F --> C
+    B --> G[Read-Only Layer]
+    G --> C
+
+subgraph F[Temporal Layer]
+    H[Transaction]
+end
+```
+
+## Key concepts
+
+  1. **Columns**: The library uses a column-based structure to organise
+     different types of data (`Identity`, `State`, `Transaction`, `Generic`).
+
+  2. **Context**: Many operations are context-aware, allowing for isolation
+     between different application contexts.
+
+  3. **Serialisation**: The library supports custom serialisation through the
+     `DataType` trait.
+
+  4. **Error handling**: The library uses the `eyre` crate for error handling,
+     providing detailed error context.
+
+  5. **Generic programming**: Extensive use of generic programming allows for
+     flexibility and type safety.
+
+## Advanced features
+
+  1. **Temporal layer**: Allows for accumulating changes in memory before
+     committing to the underlying storage, useful for managing complex
+     transactions or temporary state changes as an atomic operation.
+
+  2. **Read-only layer**: Provides a way to ensure that certain operations are
+     read-only, enhancing safety and preventing accidental modifications.
+
+  3. **Custom key types**: The key system allows for creating complex, typed
+     keys that can encode multiple pieces of information efficiently.
+
+  4. **Flexible iteration**: The iterator system supports both unstructured and
+     structured data, allowing for type-safe iteration over complex data
+     structures.
+
+  5. **Transaction support**: Built-in support for creating and applying
+     transactions, ensuring atomic updates across multiple operations.
+
+## Detailed component descriptions
+
+### Database interface
+
+The database interface is a fundamental component of the store library, serving
+as the foundation for all data storage and retrieval operations. Defined by the
+`Database` trait, this interface provides a uniform way to interact with the
+underlying storage system, abstracting away the specifics of any particular
+database implementation.
+
+At its core, the database interface is designed to support a key-value storage
+model, where each piece of data is associated with a unique key. This model is
+further enhanced by the concept of columns, which allow for logical separation
+of different types of data within the same database. The interface defines four
+primary columns: `Identity`, `State`, `Transaction`, and `Generic`, each serving
+a specific purpose in the overall data architecture.
+
+The `Database` trait specifies a set of primary operations that any implementing
+storage system must provide. These operations include checking for the existence
+of a key (`has()`), retrieving data associated with a key (`get()`), storing new
+data or updating existing data (`put()`), and removing data (`delete()`). These
+operations therefore align to the well-known, fundamental "CRUD" actions. Each
+of these operations is column-aware, allowing for efficient organisation and
+retrieval of data based on its type or purpose.
+
+In addition to these basic operations, the interface also defines methods for
+more complex interactions. The `iter()` method allows for iteration over a range
+of keys within a specific column, enabling efficient scanning and querying of
+data. This is particularly useful for operations that need to process multiple
+related entries, such as retrieving all state data for a particular context.
+
+One of the most important features of the database interface is its support for
+transactions through the `apply()` method. This method takes a `Transaction`
+object, which represents a set of changes to be applied atomically to the
+database. By supporting transactions, the interface ensures that complex
+operations involving multiple keys can be performed consistently, maintaining
+data integrity even in the face of concurrent access or system failures.
+
+The interface is designed with flexibility and extensibility in mind. It uses
+Rust's powerful type system to ensure type safety and to allow for custom key
+and value types. The use of the `eyre` crate for error handling provides rich
+context for any errors that might occur during database operations, aiding in
+debugging and error recovery.
+
+Importantly, the database interface is designed to be asynchronous and
+thread-safe. This is enforced by applying `Send + Sync` bounds to the `Database`
+trait, ensuring that implementations can be safely shared and used across
+multiple threads. This design choice allows for high-performance, concurrent
+access to the database, which is crucial for applications dealing with large
+amounts of data or high request volumes.
+
+While the interface defines the contract that any database implementation must
+fulfill, it doesn't prescribe how these operations should be implemented. This
+abstraction allows the library to support different storage backends without
+changing the core logic of the store system. Currently, the library provides an
+implementation for [RocksDB][], a high-performance key-value store, but the
+design allows for easy addition of other backends in the future.
+
+### Key structure
+
+The key structure in the store library is a sophisticated and flexible system
+designed to efficiently encode and manage complex keys for data storage and
+retrieval. This structure is fundamental to the library's ability to organise
+and access data in a type-safe and performant manner.
+
+#### Core components
+
+At the heart of the key structure are several interrelated traits and types:
+
+  - **`KeyComponent`**: This trait defines a single part of a key. It associates
+    a type with a specific length, allowing for fixed-size key components. This
+    is crucial for efficient storage and retrieval in many database systems.
+
+  - **`KeyComponents`**: Building on `KeyComponent`, this trait allows for the
+    composition of multiple components into a single key. It uses Rust's type
+    system to calculate the total length of the combined components at
+    compile-time, ensuring type safety and optimal performance.
+
+  - **`Key<T>`**: This is a newtype wrapper around a `GenericArray` of bytes.
+    The type parameter `T` must implement `KeyComponents`, which determines the
+    size and structure of the key.
+
+  - **`AsKeyParts`**: This trait defines how a key type can be decomposed into
+    its column and components. It's essential for the library's ability to work
+    with keys in a generic manner while still preserving type information.
+
+  - **`FromKeyParts`**: The counterpart to `AsKeyParts`, this trait allows for
+    the construction of a key from its components. It includes an associated
+    `Error` type, allowing for type-specific error handling during key
+    construction.
+
+#### Flexibility and type safety
+
+The key structure's design prioritises both flexibility and type safety. By
+using Rust's generic programming features, the library allows for the creation
+of complex, multi-component keys while still maintaining strong type checking at
+compile-time. This prevents many potential runtime errors related to key
+formatting or interpretation.
+
+For example, a key might be composed of a context ID and a state key:
+
+```rust
+pub struct ContextState(Key<(ContextId, StateKey)>);
+```
+
+Here, `ContextId` and `StateKey` are separate `KeyComponent` types, combined
+into a single key structure. The library can then provide methods to safely
+access these components:
+
+```rust
+use calimero_primitives::context::ContextId;
+
+impl ContextState {
+    pub fn context_id(&self) -> ContextId { ... }
+    pub fn state_key(&self) -> [u8; 32] { ... }
+}
+```
+
+#### Performance considerations
+
+The key structure is designed with performance in mind. By using fixed-size
+components and compile-time length calculation, the library can avoid runtime
+allocations and provide predictable performance characteristics. This is
+particularly important for database operations where key comparisons and sorting
+are frequent operations.
+
+#### Integration with database operations
+
+The key structure integrates seamlessly with the database interface. The
+`AsKeyParts` trait allows the database to efficiently determine the column for a
+given key, which is crucial for operations like `put()`, `get()`, and
+`delete()`. Similarly, the `FromKeyParts` trait enables the creation of
+structured keys from raw byte data, which is essential for operations like
+iteration and range queries.
+
+#### Predefined key types
+
+The library provides several predefined key types that leverage this structure:
+
+  - `ContextMeta`
+  - `ContextIdentity`
+  - `ContextState`
+  - `ContextTransaction`
+  - `Generic`
+
+These predefined types demonstrate how the key structure can be used to create
+semantically meaningful keys that encode multiple pieces of information.
+
+### Data types and entries
+
+The store library implements a sophisticated system for handling diverse data
+types and entries, providing a flexible and type-safe approach to storing and
+retrieving various kinds of information. This system is built around two key
+traits: `DataType` and `Entry`, which work together to define how data is
+serialised, deserialised, and associated with keys in the store.
+
+#### The `DataType` trait
+
+The `DataType` trait is at the core of the library's value handling system. It
+defines how different types of data are converted to and from raw bytes for
+storage. This trait has two primary methods:
+
+  1. **`from_slice()`**: This method converts a slice of bytes into the specific
+     data type.
+
+  2. **`as_slice()`**: This method converts the data type back into a slice of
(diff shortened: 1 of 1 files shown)


crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+    - [Key structure](#key-structure)
+    - [Data types and entries](#data-types-and-entries)
+    - [Layers](#layers)
+    - [Transactions](#transactions)
+    - [Iterators](#iterators)
+  - [Operation flow](#operation-flow)
+  - [Key concepts](#key-concepts)
+  - [Advanced features](#advanced-features)
+  - [Detailed component descriptions](#detailed-component-descriptions)
+    - [Database interface](#database-interface-1)
```
miraclx, line 17: missing link?

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+## Introduction
+
+The state/store library is a versatile and efficient key-value storage system
+designed for managing application state and transactions. It provides a flexible
+and extensible architecture for storing, retrieving, and manipulating data
+across different storage backends.
+
+In this context, "state" refers to the data being stored and managed, while
+"store" refers to the actual storage mechanism and its associated operations.
```
miraclx: I think since "state" is only a piece of this, and is focused on in detail later in this doc, we can leave it out here, and just talk about the store.


```suggestion
The store library is a versatile and efficient key-value storage system
designed for managing application state and transactions. It provides a flexible
and extensible architecture for storing, retrieving, and manipulating data
across different storage backends.
```

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+developers to work with application state seamlessly. From here on, the term
+"store library" or just "store" will be used, but this encompasses state too.
+
+Key features of the store library include:
+
+  - Support for potentially multiple storage backends (currently just
+    [RocksDB][])
+  - Transactional operations with atomic commits
+  - Flexible key-value structure with custom data types
+  - Iteration and range queries
```
miraclx, line 37: range queries are kinda like an implementation detail, and only explicitly named such wrt `store::tx::IterRange`

the idea is we don't need special range specifications since rust's iterators are lazy anyway, and by specifying a start key, you can choose to stop advancing `next()` once you're satisfied with consumption

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+  - Temporal layers for managing changes before committing
```
miraclx: Layers for augmenting/adapting lower-level database functionality, allowing, for example, atomic transactions, caching, etc..

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+`Entry` trait combines a key with its associated data type, providing a complete
+representation of a stored item.
+
+Predefined entry types include:
+
+  - `ContextMeta`: Metadata for a context
+  - `ContextState`: State data for a context
+  - `ContextIdentity`: Identity information for a context
+  - `ContextTransaction`: Transaction data for a context
+  - `Generic`: A generic key-value pair
```
miraclx, line 84: if you're mentioning these here, maybe mention the keys too? unless we can skip them?

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+### Layers
+
+The library implements a layered architecture for managing data access and
+modifications:
+
+  - `Layer`: Base trait for all layers
+  - `ReadLayer`: Provides read-only access to the underlying data
+  - `WriteLayer`: Extends `ReadLayer` with write operations
+  - `Temporal`: A specialised layer that accumulates changes before committing
```
miraclx: `Temporal` is a type, not a trait, so maybe it doesn't fit here. Mentioning it here would call into question how about `ReadOnly`?

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+atomically. It supports putting and deleting entries, as well as merging with
+other transactions.
+
+### Iterators
+
+The library provides a flexible iteration system:
+
+  - `Iter`: Base iterator type
+  - `DBIter`: Trait for database-specific iterators
+  - `IterRange`: Iterator for a specific key range
```
miraclx: we can skip mentioning `IterRange` here

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+## Operation flow
+
+  1. Initialise the store with a specific backend (e.g., [RocksDB][])
+  2. Create a `StoreHandle` to interact with the store
+  3. Perform read operations directly on the store
+  4. For write operations:
+       - Create a `Temporal` layer
+       - Accumulate changes in the temporal layer
+       - Commit changes to the underlying store
```
miraclx:
```suggestion
  3. Perform read/write operations directly on the store
  4. For atomic write operations:
       - Create a `Temporal` layer
       - Accumulate changes in the temporal layer
       - Commit changes to the underlying store
```

The idea for layering was methods specify their contract in the type system. For example:

```rust
fn store_stuff<S: WriteLayer>(store: S) {}
fn read_stuff<S: ReadLayer>(store: S) {}
```

so you know from the signature what each functions does to your state

I can see us potentially encoding more things into the type system at a higher level, allowing functions to specify that they need, for example, to be able to write `ContextIdentity`, so the caller evaluates that expectation

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+graph TD
+    A[Application] --> B[StoreHandle]
+    B --> C[Store]
+    C --> D[Database Backend]
+    D --> E[RocksDB and others]
+    B --> F[Temporal Layer]
+    F --> C
+    B --> G[Read-Only Layer]
+    G --> C
+    H[Transaction] --> F
```
miraclx: can we use a subgraph for `Temporal Layer`?

so it "contains" `Transaction`?

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+    I[Iterator] --> C
```
miraclx: I don't understand this one, a `Store: ReadLayer` "can" make iterators, but it doesn't fit with the rest at this level

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+## Key concepts
+
+  1. **Columns**: The library uses a column-based structure to organise
+     different types of data (`Identity`, `State`, `Transaction`, `Generic`).
+
+  2. **Context**: Many operations are context-aware, allowing for isolation
+     between different application contexts.
+
+  3. **Serialisation**: The library supports custom serialisation through the
```
miraclx, line 144: we use `borsh` in certain cases, `serde_json` for others, but it shouldn't matter it's generic so it's up to each data type to define what they want to use

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+     providing detailed error context.
+
+  5. **Generic programming**: Extensive use of generic programming allows for
+     flexibility and type safety.
+
+## Advanced features
+
+  1. **Temporal layer**: Allows for accumulating changes in memory before
+     committing to the underlying storage, useful for managing complex
+     transactions or temporary state changes.
```
miraclx: The key point here is to allow for atomic application where the db supports it

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+`delete()`. Similarly, the `FromKeyParts` trait enables the creation of
+structured keys from raw byte data, which is essential for operations like
+iteration and range queries.
+
+#### Extensibility
+
+The key structure's design allows for easy extensibility. New key types can be
+created by composing existing `KeyComponent` types or by defining new ones. This
+extensibility is crucial for adapting the library to various use cases and data
+models without modifying the core library code.
```
miraclx: actually, by design, it shouldn't be possible to construct your own keys

`KeyComponent` isn't exposed from `store::key` and neither are it's implementations

you can only use the predefined keys, and this is done to protect columns by reducing the scope of keys that can reference them

All things the "Node" needs to store should exist as pre-defined keys, and for anything beyond that, we have the the `Generic` key within which you can specify a scope (16 bytes) and fragment of it (32-bytes) for multiple entries

For example, Client identities aren't a "feature" of the "Node", but of the "Auth-gated client" (see https://github.com/calimero-network/core#architecture) so it depends on this interface [here](https://github.com/calimero-network/core/blob/74853212aefb652197eede710784ebad7201a695/crates/server/src/admin/storage/did.rs#L22)

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+### Data types and entries
+
+The store library implements a sophisticated system for handling diverse data
+types and entries, providing a flexible and type-safe approach to storing and
+retrieving various kinds of information. This system is built around two key
+traits: `DataType` and `Entry`, which work together to define how data is
+serialised, deserialised, and associated with keys in the store.
+
+#### The `DataType` trait
```
miraclx, line 328: unlike keys that aren't designed to be extended, `DataTypes` are, so we can mention it here

Example use outside the store:

https://github.com/calimero-network/core/blob/74853212aefb652197eede710784ebad7201a695/crates/server/src/admin/storage/did.rs#L10-L17

looking at this again, we (could) potentially make `Generic` generic as `Generic<T>` for any `T` that is some other public trait, unsure if that association is useful or how that key will be consumed, but it might make sense to type-separate `Generic`s of `DID`s from others.

And as a neat trick, we can use const-generics to avoid scope collisions.

But for now, the model we have is simple enough and works

miraclx: Extracted latter comment into https://github.com/calimero-network/core/issues/476.

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+  4. **`ContextTransaction`**: Stores information about a transaction, including
+     the method name, payload, and the hash of the prior transaction.
+
+  5. **`Generic`**: A flexible entry type for storing arbitrary data associated
+     with a scope and fragment.
+
+Each of these predefined types implements both the `DataType` and
+`PredefinedEntry` traits, with the latter being a convenience trait that
+combines `Entry` with `AsKeyParts` and `FromKeyParts`.
```
miraclx: that implements `Entry` for all defined keys, with the entry key as the key it's defined for

also, `PredefinedEntry` can't be implemented outside the store crate because it requires it's implementor to impl `AsKeyParts` which, as mentioned earlier is not implementable.

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+#### Serialisation and deserialisation
+
+For the predefined types, the library uses the Borsh serialisation format, which
+provides efficient, schema-driven serialisation and deserialisation. This is
+implemented through the `BorshSerialize` and `BorshDeserialize` traits.
+
+The use of Borsh allows for compact data representation and fast processing,
+which is crucial for a high-performance storage system. However, the `DataType`
+trait's design allows for easy integration of other serialisation formats if
+needed for specific use cases.
```
miraclx: As mentioned earlier, `borsh` isn't used for all data types

Some, like `ContextState` don't need serialization at all, while others like for `Generic<Did>`, the DID is serialized using `Json`

https://github.com/calimero-network/core/blob/74853212aefb652197eede710784ebad7201a695/crates/server/src/admin/storage/did.rs#L12

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+#### Flexibility and type safety
+
+The combination of the `DataType` and `Entry` traits with Rust's type system
+provides a high degree of flexibility while maintaining strong type safety. This
+design allows developers to define custom data types and entries that fit their
+specific needs, while the compiler ensures that keys and values are used
+consistently throughout the application.
```
miraclx: Hm, consistently? correctly perhaps?

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+     effectively preventing any write operations from being performed through
+     it.
+
+#### Layer composition
+
+One of the most powerful aspects of the layer system is the ability to compose
+layers. Layers can be stacked on top of each other, each adding or modifying
+functionality. For example, you might create a `Temporal` layer on top of a
+`ReadOnly` layer, allowing for transactional-style operations without the
+ability to actually modify the underlying store.
```
miraclx: Seems interesting, but I couldn't think of a case where you would want that 😅 

So right now, `Temporal` can only be constructed from a `WriteLayer` and it's only even just a `Layer` at all when it wraps a `WriteLayer`

https://github.com/calimero-network/core/blob/74853212aefb652197eede710784ebad7201a695/crates/store/src/layer/temporal.rs#L12-L29

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+  - It interacts with the transaction system, particularly in the case of the
+    `Temporal` layer.
+
+#### Extensibility
+
+The layer system is designed to be extensible. New layers can be easily created
+by implementing the appropriate traits. This allows for the creation of
+specialised layers for specific use cases, such as logging layers, caching
+layers, or layers that implement specific business logic.
```
miraclx, line 539: ah, caching is mentioned here ✨ 

another interesting one is `Tee`, which wraps two `Layer`s, reading from one, falling back on to the other when it doesn't exist, or writing to both

this is part of the design considerations that influenced the decision of why `WriteLayer::apply` takes a `&Transaction` for example

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+#### Performance considerations
+
+The layered architecture does introduce some overhead, particularly in terms of
+method call indirection. However, this is generally outweighed by the benefits
+in terms of flexibility and functionality. Moreover, the use of Rust's zero-cost
+abstractions means that much of this overhead can be optimised away at compile
+time.
```
miraclx, line 554: yeah, everything is staticly dispatching, so Rust should in-theory optimize all of it away. the only piece that it cannot is `Store`'s dependence on `dyn Database` which will dynamically dispatch and so for convenience, we're out of luck

although, that part was designed prior to redesigning the store interface and it was influenced by runtime's requirements, so we might revisit now that we have read/write contracts defined in the type system

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+### Transactions
+
+The store library implements a robust transaction system that allows for atomic,
+consistent updates to the data store. This system is crucial for maintaining
+data integrity, especially in scenarios involving complex operations or
+concurrent access.
+
+#### Core transaction  oncept
```
miraclx: 404

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+  1. **Atomic operations**: All operations in a transaction are applied
+     together, or not at all. This ensures consistency in the face of failures
+     or interruptions.
+  
+  2. **Isolation**: Changes made within a transaction are not visible to other
+     parts of the system until the transaction is committed.
+  
+  3. **Durability**: Once a transaction is committed, its changes are
+     permanently stored and will survive system failures.
```
miraclx: that's dependent on the `Database` implementation, but `Transaction` allows capable databases to take advantage of this

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+The execution of a transaction typically follows these steps:
+
+  1. A new `Transaction` is created.
+
+  2. Operations (`put()` and `delete()`) are added to the transaction.
+
+  3. The transaction is passed to a layer's `apply()` method.
+
+  4. The layer (often a `Temporal` layer) processes the transaction's
```
miraclx, line 641: > (often a `Temporal` layer)

`Store` is a `WriteLayer` too

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+  - The use of a `BTreeMap` for storing operations allows for efficient lookups
+    and range queries.
+
+  - The ability to batch multiple operations into a single transaction can
+    significantly reduce the number of interactions with the underlying storage
+    system.
+
+  - The `Temporal` layer's ability to accumulate changes before committing
+    allows for efficient batching of writes.
```
miraclx, line 672: this section seems extraneous

> allows for efficient lookups
>    and range queries.

mentioned above under "Transaction structure"

> The ability to batch multiple operations into a single transaction can
>    significantly reduce the number of interactions with the underlying storage
>    system.

Can be mentioned in "Core transaction concept"

> The `Temporal` layer's ability to accumulate changes before committing
>    allows for efficient batching of writes.

we can do without this one

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+  - Custom layers could implement specialised transaction handling logic.
+
+  - The system could be extended to support more complex transaction patterns,
+    such as nested transactions or savepoints.
+
+#### Error handling
+
+The transaction system is designed to work with Rust's `Result` type and the
+`eyre` error handling library. This allows for detailed error reporting and
+proper error propagation throughout the system.
```
miraclx: Let's take these out too

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+The iterator system is built around several key types and traits:
+
+  1. **`Iter`**: This is the primary iterator type in the library. It's generic
+     over two type parameters, `K` and `V`, which represent the key and value
+     types respectively.
+
+  2. **`DBIter`**: This trait defines the core functionality that any
+     database-specific iterator must implement. It includes methods for moving
+     to the next item (`next()`) and reading the current value (`read()`). In
+     this way it is similar to Rust's `Iterator` trait.
```
miraclx: One thing to mention here in terms of expectations, is.. When an iterator (over a database, not layer) is created, it should essentially become an iterator over a snapshot of the db at that point in time.

And while mutations to the store can still happen during this iterator's lifetime (notice none of the `Database` methods are `&mut`), it means the items that (could) be returned that might've been deleted only get freed at that point.

This also means, we're able to iterate over all the keys, and delete ones of interest (which deviates a bit from Rust's iterator behaviour)

RocksDB upholds these expectations: https://github.com/facebook/rocksdb/wiki/Iterator

> A consistent-point-in-time view of the database is created when the Iterator is created. Thus, all keys returned via the Iterator are from a consistent view of the database.

and the in-memory store I've got in my WIP patch does as well

miraclx: Another thing to note here about why there's a layer of indirection to reading values is to avoid unnecessary IO.

Unsure how well RocksDB upholds those expectations of permitting key iteration without values considering (ref: #475)

> Worth noting that RocksDB has a boat-load of optimizations, one of which is being able to mmap SST files to avoid memory copying (much akin to our efforts with https://github.com/calimero-network/core/issues/307).
>
> In which case, the PinnableSlice<'_> encapsulated into Slice<'_> from RocksDB::get might just reference an index in these mmap-ed pages, and as such, a further optimization might not perform better.

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+begin at a specific point in the key space, enabling efficient partial scans of
+the data.
+
+#### Performance considerations
+
+The iterator system is designed with performance in mind:
+
+  1. **Lazy evaluation**: Iterators are lazy, meaning they only do work when
+     asked for the next item. This allows for efficient processing of large
+     datasets without unnecessary memory usage.
```
miraclx: Also allows you to stop at will, even though the iterator is only created with `start`

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+#### Limitations and considerations
+
+While powerful, users of the iterator system should be aware of certain
+considerations:
+
+  1. **Consistency**: In a multi-threaded environment, iterators provide a
+     point-in-time view and may not reflect concurrent modifications. If an
+     iterator is to be shared across threads, care should be taken to ensure
+     consistency — and it may be preferable to use [Rayon][]'s parallel
+     iterators instead for concurrent processing.
```
miraclx: ah, it's mentioned here.. I think it needs better placement than down here

Also, iterators should behave no differently from other Rust types across threads, we don't need to mention Rayon.

Rust won't let incorrect use compile.

crates/store/README.md
```diff
@@ -0,0 +1,829 @@
...
+
+  2. **Resource usage**: Long-running iterations may hold resources for extended
+     periods. Care should be taken in concurrent scenarios.
```
miraclx, line 802: by resources you mean? potentially deleted values on the live view of the storage? or what?

crates/store/README.md
danwilliams: I did think it fitted here, and had mentioned that it was a specialised layer, but I have removed it as requested.

crates/store/README.md
danwilliams: I feel a little disappointed, as I felt that was a cool scenario 😄 

There are situations where it can be useful to do this - for instance, dry runs of a sequence of events.

However, I think I got carried away, and didn't notice this isn't actually possible - and my example is perhaps not all that relevant to the target audience, I dunno.

crates/store/README.md
danwilliams: Looks like I hit space instead of "c"...

crates/store/README.md
danwilliams: I've left this in, as I wrote the document with each area having a performance section, which summarises the key points from that perspective. So, although it is covered elsewhere, it's consistent to include it.

Regarding the batching of writes, I left that in too, as it is a benefit?

crates/store/README.md
danwilliams: This is covered under the limitations section, but I added a link to it 🙂

danwilliams: This is covered in the performance section 🙂

crates/store/README.md
miraclx: > dry runs of a sequence of events.

I can get behind that, we can patch it quite simply.

Especially since the embeddable model will expose these pieces to consumers, who can then choose to compose them how they want.

> and my example is perhaps not all that relevant to the target audience

it's spot on

> didn't notice this isn't actually possible

it's fine, let's continue with documenting "as-is", we'll revisit this piece

crates/store/README.md
danwilliams: I've added a link 👍

crates/store/README.md
miraclx: I noticed you removed `Generic` here, to clarify.. my comment wasn't about this, but mentioning predefined entries at all at this level

And that's only in question because when you discuss keys a section above, we didn't talk about the types of keys.

Figured it's worth clarifying just in case it wasn't clear, it's not that I think it's wrong to mention or there's something wrong with `Generic` specifically.

Not sure what model is best, since (due to `Predefined`'s behaviour `Keys` are treated as entries, so it might seem like duplication.

Unless there's some way of talking about both cases.

crates/store/README.md
```diff
@@ -0,0 +1,805 @@
...
+    B --> F[Temporal Layer]
+    F --> C
+    B --> G[Read-Only Layer]
+    G --> C
+
+subgraph Temporal Layer
+F
+H[Transaction]
+F --> H
+end
```
miraclx: what do you think of this

```mermaid
graph TD
    A[Application] --> B[StoreHandle]
    B --> C[Store]
    C --> D[Database Backend]
    D --> E[RocksDB and others]
    B --> F
    F --> C
    B --> G[Read-Only Layer]
    G --> C
subgraph F[Temporal Layer]
    H[Transaction]
end
```

<details>
<summary>patch</summary>

```suggestion
    B --> F
    F --> C
    B --> G[Read-Only Layer]
    G --> C

subgraph F[Temporal Layer]
    H[Transaction]
end
```

</details>

crates/store/README.md
danwilliams: > Also, iterators should behave no differently from other Rust types across threads, we don't need to mention Rayon.

That's... not quite right. It's true that in Rust, iterators themselves don't have special thread-safety properties. The behaviour of iterators across threads is determined by the types they iterate over and the ownership model of Rust. If a type is `Send` and `Sync`, it can be safely transferred and accessed across threads. This applies to iterators as well.

So although you are correct in implying that iterators don't inherently need special mention of Rayon regarding thread safety, as long as the underlying types conform to Rust's concurrency guarantees, it does not therefore automatically follow that everything iterable will work exactly as expected. Hence the mention of Rayon as a _possible_ ("it _may_ be preferable") option for concurrent processing.

crates/store/README.md
danwilliams: No... anything that's in use by the iterator. So, borrowed data, or anything in scope of the iterator logic. In a situation where that might be used across multiple threads, that can lead to thread starvation.

crates/store/README.md
danwilliams: Sure! If you prefer 🙂

crates/store/README.md
miraclx: To clarify, I meant "our" iterators. In Rust, iterators take a `&mut self` which according to Rust's rules, forbids multiple references, that's where Rayon helps.

I mention this because you said "consistency", which I'm taking to mean "in the face of concurrent writes", your "point-in-time" view should stay "consistent" at that point in time.

In Rust, this is forbidden because the iterator you hold references some held data, while the mutation you intend to make, requires an exclusive reference.

In our case, `Database::{put, delete}` take shared references, implying that each database must implement snapshotting and manage RW contention internally (which RocksDB does as mentioned above - https://github.com/calimero-network/core/pull/472#discussion_r1678942579)

This means we're able to provide an uncomplicated way to iterate without requiring caveat communication.

And as such, however you choose to consume the output iterator should be safe.

This is why I said, our iterators should behave no differently from other Rust types across threads.

The same considerations you would make for a `Vec`'s or `BTreeMap`'s iterator, are the same you would make here.

Which is why I said, it's not necessary to mention.

crates/store/README.md
danwilliams: Hmmmm, I see them as different... I am not sure what outcome is best.

For now I've re-added `Generic` - thanks for clarifying that.

crates/store/README.md
miraclx: Ah, so this is directed at people implementing `Database`?

crates/store/README.md
danwilliams: Thanks for the clarification. You're right that Rust's type system prevents many incorrect uses at compile time, but when dealing with persistent storage and database-like structures there are additional considerations beyond what Rust's type system can enforce.

Even if the Rust code is safe, the data being iterated over could change if there are concurrent modifications to the database. This "consistency" is what I was referring to, rather than memory safety.

Separately, the mention of Rayon was indeed not necessary, but I felt it was a useful note for anyone needing to perform concurrent iteration. It is not critical, and I agree it can be removed.

You're absolutely right that our iterators should behave similarly to other Rust types in terms of thread safety, and I haven't disagreed with that, but equally the discussion there is going off at a tangent unrelated to what I wrote.

As the way I've structured the sentence has caused conflation, and the mention of Rayon is not necessary, I've simply removed it.

crates/store/README.md
danwilliams: Well, this whole README is about the store. Regardless of who implements `Database`, iterating through data may cause concurrency issues if the things being held by the iteration process are required elsewhere. It is simply something to be aware of.

miraclx (approved):

PR #513, opened 2024-07-18
fix/ssl-server-wrapper

alenmestrov:
# TLS Wrapper for Server

## Summary:
When trying to access the local server from a browser (such as GitHub Pages), users receive CORS issues for serving insecure content. This issue is mentioned here: [calimero-network/admin-dashboard#26](https://github.com/calimero-network/admin-dashboard/issues/26).

To resolve this, we generate a self-signed SSL certificate for the server based on the current local IP address. The implementation includes:
- If a certificate doesn't exist, create a new certificate.
- If a certificate exists for the current IP address, use the existing one.
- If a certificate exists but is not configured for the current IP address, create a new certificate.

Additionally, instructions have been added with steps to add the certificate to the device to the CLI. The server is wrapped in TLS using the `axum-server` crate.

## Changes:
- **/crates/server/Cargo.toml**: Modified to include necessary crates.
- **Cargo.lock**: Updated to reflect changes in dependencies.
- **/crates/server/src/lib.rs**: Updated to call methods from `certificates.rs` and wrap the server with TLS.
- **/crates/server/src/certificates.rs**: Added for SSL certificate generation and validation.
- **/crates/server/certificates/**: Created a folder for generated certificates.
- **/crates/meroctl/src/cli/init.rs**: Updated for initial setup to start the server on the local IP address.

## How to test:
1. Initialize the node.
```sh 
cargo run -p meroctl -- --node-name node1 --home data init --server-port 2428 --swarm-port 2528
```
3. Start the node.
```sh
cargo run -p meroctl -- --node-name node1 --home data run
```
4. Follow the instructions in CLI to add the SSL certificate to your device.
5. Try accessing the local server using the HTTPS URL: [https://calimero-network.github.io/getting-started/admin-dashboard/](https://calimero-network.github.io/getting-started/admin-dashboard/)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,11 @@ ureq = "2.9.7"
 wasmer = "4.2.5"
 wasmer-types = "4.2.5"
 web3 = "0.19.0"
+local-ip-address = "0.6.1"
+rcgen = "0.13.1"
+x509-parser = "0.16.0"
+axum-server = { version = "0.6", features = ["tls-rustls"] }
+axum-server-dual-protocol = "0.6.0"
 
 [profile.app-release]
 inherits = "release"
--- a/crates/meroctl/src/cli/init.rs
+++ b/crates/meroctl/src/cli/init.rs
@@ -35,7 +35,7 @@ pub struct InitCommand {
 
     /// Host to listen on for RPC
     #[clap(long, value_name = "HOST")]
-    #[clap(default_value = "127.0.0.1,::1")]
+    #[clap(default_value = "0.0.0.0,::1")]
     #[clap(use_value_delimiter = true)]
     pub server_host: Vec<IpAddr>,
 
--- a/crates/meroctl/src/cli/init.rs
+++ b/crates/meroctl/src/cli/init.rs
@@ -35,7 +35,7 @@ pub struct InitCommand {
 
     /// Host to listen on for RPC
     #[clap(long, value_name = "HOST")]
-    #[clap(default_value = "0.0.0.0,::1")]
+    #[clap(default_value = "127.0.0.1,::1")]
     #[clap(use_value_delimiter = true)]
     pub server_host: Vec<IpAddr>,
 
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -29,10 +29,11 @@ tower.workspace = true
 tower-http = { workspace = true, features = ["cors", "fs"] }
 tower-sessions = { workspace = true, optional = true }
 tracing.workspace = true
-axum-server = { version = "0.6", features = ["tls-rustls"] }
-local-ip-address = "0.6.1"
-rcgen = "0.13.1"
-x509-parser = "0.16.0"
+axum-server.workspace = true
+axum-server-dual-protocol.workspace = true
+rcgen.workspace = true
+x509-parser.workspace = true
+local-ip-address.workspace = true
 
 calimero-context = { path = "../context" }
 calimero-node-primitives = { path = "../node-primitives" }
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -29,6 +29,10 @@ tower.workspace = true
 tower-http = { workspace = true, features = ["cors", "fs"] }
 tower-sessions = { workspace = true, optional = true }
 tracing.workspace = true
+axum-server = { version = "0.6", features = ["tls-rustls"] }
+local-ip-address = "0.6.1"
+rcgen = "0.13.1"
+x509-parser = "0.16.0"
 
 calimero-context = { path = "../context" }
 calimero-node-primitives = { path = "../node-primitives" }
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -8,6 +8,8 @@ license.workspace = true
 
 [dependencies]
 axum.workspace = true
+axum-server.workspace = true
+axum-server-dual-protocol.workspace = true
 base64.workspace = true
 borsh = { workspace = true, features = ["derive"] }
 bs58.workspace = true
@@ -16,8 +18,10 @@ ed25519-dalek.workspace = true
 eyre.workspace = true
 futures-util = { workspace = true, optional = true }
 libp2p.workspace = true
+local-ip-address.workspace = true
 multiaddr.workspace = true
 rand.workspace = true
+rcgen.workspace = true
 reqwest.workspace = true
 semver.workspace = true
 serde = { workspace = true, features = ["derive"] }
@@ -29,11 +33,7 @@ tower.workspace = true
 tower-http = { workspace = true, features = ["cors", "fs"] }
 tower-sessions = { workspace = true, optional = true }
 tracing.workspace = true
-axum-server.workspace = true
-axum-server-dual-protocol.workspace = true
-rcgen.workspace = true
 x509-parser.workspace = true
-local-ip-address.workspace = true
 
 calimero-context = { path = "../context" }
 calimero-node-primitives = { path = "../node-primitives" }
--- a/crates/server/src/admin/storage/ssl.rs
+++ b/crates/server/src/admin/storage/ssl.rs
@@ -59,7 +59,7 @@ pub fn get_ssl(store: Store) -> eyre::Result<Option<SSLCert>> {
     let handle = store.handle();
 
     match handle.get(&entry) {
-        Ok(Some(ssl_document)) => Ok(Some(ssl_document.value().clone())),
+        Ok(Some(ssl_document)) => Ok(Some(ssl_document.value())),
         Ok(None) => Ok(None),
         Err(e) => Err(e.into()),
     }
--- a/crates/server/src/admin/storage/ssl.rs
+++ b/crates/server/src/admin/storage/ssl.rs
@@ -1,0 +1,66 @@
+use calimero_store::entry::{Entry, Json};
+use calimero_store::key::Generic;
+use calimero_store::Store;
+use serde::{Deserialize, Serialize};
+
+struct SSLEntry {
+    key: Generic,
+}
+
+impl Entry for SSLEntry {
+    type Key = Generic;
+    type DataType<'a> = Json<SSLCert>;
+
+    fn key(&self) -> &Self::Key {
+        &self.key
+    }
+}
+
+impl SSLEntry {
+    fn new() -> Self {
+        Self {
+            key: Generic::new(*b"ssl_certs:server", [0; 32]),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct SSLCert {
+    cert: Vec<u8>,
+    key: Vec<u8>,
+}
+
+impl SSLCert {
+    pub fn cert(&self) -> &Vec<u8> {
+        &self.cert
+    }
+
+    pub fn key(&self) -> &Vec<u8> {
+        &self.key
+    }
+}
+
+pub fn insert_or_update_ssl(store: Store, cert: &[u8], key: &[u8]) -> eyre::Result<SSLCert> {
+    let ssl_cert = SSLCert {
+        cert: cert.to_vec(),
+        key: key.to_vec(),
+    };
+
+    let ssl_document = Json::new(ssl_cert.clone());
+    let entry = SSLEntry::new();
+    let mut handle = store.handle();
+    handle.put(&entry, &ssl_document)?;
+
+    Ok(ssl_cert)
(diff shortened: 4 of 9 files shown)


alenmestrov:
This PR is connected to documentation update PR [PR#10](https://github.com/calimero-network/calimero-network.github.io/pull/10)

crates/meroctl/src/cli/init.rs
```diff
@@ -35,7 +35,7 @@ pub struct InitCommand {
 
     /// Host to listen on for RPC
     #[clap(long, value_name = "HOST")]
-    #[clap(default_value = "127.0.0.1,::1")]
+    #[clap(default_value = "0.0.0.0,::1")]
```
miraclx: let's not listen on all interfaces by default, keep that explicit for the node operator

it's different with the swarm host since that needs to be as accessible as possible through all means for network communication

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,164 @@
+use rcgen::{CertificateParams, DnType};
+use local_ip_address::local_ip;
+use std::fs;
+use std::path::Path;
+use x509_parser::prelude::*;
+use x509_parser::extensions::GeneralName;
+use std::path::PathBuf;
```
miraclx: `cargo +nightly fmt`

crates/server/Cargo.toml
```diff
@@ -29,6 +29,10 @@ tower.workspace = true
 tower-http = { workspace = true, features = ["cors", "fs"] }
 tower-sessions = { workspace = true, optional = true }
 tracing.workspace = true
+axum-server = { version = "0.6", features = ["tls-rustls"] }
+local-ip-address = "0.6.1"
+rcgen = "0.13.1"
+x509-parser = "0.16.0"
```
miraclx: move these to the root manifest `Cargo.toml` and link to the versions with `workspace = true` like the others

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,164 @@
...
+    let key_pem = key_pair.serialize_pem();
+
+    // Get the path to the certificate and key files
+    let certificate_dir = get_certificate_dir();
+    let cert_file_path = certificate_dir.join("cert.pem");
+    let key_file_path = certificate_dir.join("key.pem");
+    
+    // Save the certificate and key to files
+    fs::write(cert_file_path, cert_pem)?;
+    fs::write(key_file_path, key_pem)?;
```
miraclx: Consider putting these in storage instead of on-disk (non-blocker though)

You can use this for inspiration - https://github.com/calimero-network/core/blob/9896f712f887f65a7b2b62f6defd962ad28b4463/crates/server/src/admin/storage/did.rs

crates/server/src/lib.rs
```diff
@@ -108,14 +111,42 @@ pub async fn start(
...
+    // Check if the certificate exists and if they contain the current local IP address
+    certificates::check_for_certificate()?;
+
+    // Get the certificate directory and the paths to the certificate and private key
+    let certificate_dir = certificates::get_certificate_dir();
+    let cert_path = certificate_dir.join("cert.pem");
+    let key_path = certificate_dir.join("key.pem");
+
+    // Configure certificate and private key used by https
+    let rustls_config = match RustlsConfig::from_pem_file(cert_path, key_path).await {
```
miraclx: why not use `from_pem`? .._

crates/server/src/lib.rs
```diff
@@ -108,14 +111,42 @@ pub async fn start(
                 http::Method::DELETE,
                 http::Method::PUT,
                 http::Method::OPTIONS,
-            ]),
+            ])
+            .allow_private_network(true)
     );
+    // Check if the certificate exists and if they contain the current local IP address
+    certificates::check_for_certificate()?;
```
miraclx: _.. and combine the check & read instead

crates/server/src/lib.rs
```diff
@@ -108,14 +111,42 @@ pub async fn start(
...
 
     let mut set = tokio::task::JoinSet::new();
 
     for listener in listeners {
+        let rustls_config = rustls_config.clone();
         let app = app.clone();
-        set.spawn(async { axum::serve(listener, app).await });
+        let addr = listener.local_addr().unwrap();
+        set.spawn(async move {
+            if let Err(e) = axum_server::bind_rustls(addr, rustls_config)
```
miraclx: looks like this will only permit TLS-able requests, we still need to conditionally allow HTTP right?

or are we forbidding that, favoring HTTPS always?

There are some options for this:

- https://crates.io/crates/axum-server-dual-protocol
- https://users.rust-lang.org/t/how-to-serve-http-and-https-on-the-same-port-with-hyper/103759/2

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,164 @@
...
+    println!("     a. Open Keychain Access and select the 'System' keychain.");
+    println!("     b. Drag the cert.pem file into Keychain Access.");
+    println!("     c. Double-click the certificate, expand the 'Trust' section, and set 'When using this certificate' to 'Always Trust'.");
+    println!("   - For Windows:");
+    println!("     a. Double-click the cert.pem file.");
+    println!("     b. Select 'Install Certificate...', choose the 'Local Machine' store location, and place it in the 'Trusted Root Certification Authorities' store.");
+    println!("   - For Linux:");
+    println!("     a. The process varies by distribution and browser. For most browsers, importing the certificate through the browser's settings by navigating to 'Privacy and Security' -> 'Manage Certificates' should work.");
+    println!("3. After installation, you may need to restart your browser or any application that needs to trust the certificate.");
+    println!("*******************************************************************************");
```
miraclx: I think these should be part of the doc instead, and linked to by the CLI

https://github.com/calimero-network/calimero-network.github.io/pull/10/files#r1682828762

Otherwise, since this is running locally, we can know what platform it's executing on and only show instructions for that platform.

There's also the potential for this to run in a background process, `docker compose up` for example

Logs shouldn't include useful instructions (unless contextual from runtime knowledge), docs should.

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,164 @@
...
+    let (_, pem) = parse_x509_pem(cert_pem.as_bytes())?;
+    let x509_cert = pem.parse_x509()?;
+
+    // Get the local IP address
+    let local_ip = local_ip()?;
+
+    // Parse the local IP address to Vec<u8>
+    let local_ip_bytes = match local_ip {
+        std::net::IpAddr::V4(ip) => ip.octets().to_vec(),
+        std::net::IpAddr::V6(ip) => ip.octets().to_vec(),
```
miraclx: You can avoid the allocation by moving this to the point of consumption .._

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,164 @@
...
+      // Check if it's a SubjectAlternativeName extension
+      if let ParsedExtension::SubjectAlternativeName(san) = parsed_ext {
+          // Iterate over the general names
+          for gn in &san.general_names {
+              // Check if the general name is an IP address
+              if let GeneralName::IPAddress(dns_name) = gn {
+                  if *dns_name == local_ip_bytes {
+                      ip_found = true;
+                      break;
+                  }
```
miraclx: _.. like this

```suggestion
                  ip_found = match local_ip {
                      std::net::IpAddr::V4(ip) => ip.octets() == dns_name,
                      std::net::IpAddr::V6(ip) => ip.octets() == dns_name,
                  };

                  if ip_found {
                      break;
                  }
```

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,164 @@
+use rcgen::{CertificateParams, DnType};
+use local_ip_address::local_ip;
+use std::fs;
+use std::path::Path;
+use x509_parser::prelude::*;
```
miraclx: how many things would you have to pull in to avoid this blanket import? let's favor explicitness, unless with good reason

alenmestrov:
I've made the following updates based on feedback and further improvements:
- Changed the default value for the HOST field from '0.0.0.0,::1' back to '127.0.0.1,::1'.
- Moved `axum-server`, `local-ip-address`, `rcgen`, and `x509-parser` dependencies to the root `Cargo.toml` and linked them in the workspace.
- Refactored `certificates.rs` to use in-memory certificate and key handling.
- Updated `start` function to call `certificates::get_certificate` for checking and reading the certificate.
- Switched to using `axum_server_dual_protocol::bind_dual_protocol` for dual protocol support.
- Linked SSL certification steps to the github documentation page

Cargo.toml
```diff
@@ -94,6 +94,11 @@ ureq = "2.9.7"
 wasmer = "4.2.5"
 wasmer-types = "4.2.5"
 web3 = "0.19.0"
+local-ip-address = "0.6.1"
+rcgen = "0.13.1"
+x509-parser = "0.16.0"
+axum-server = { version = "0.6", features = ["tls-rustls"] }
+axum-server-dual-protocol = "0.6.0"
```
miraclx, line 101: nit: order

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,155 @@
...
+                        break;
+                    }
+                }
+            }
+        }
+    }
+
+    if !ip_found {
+        delete_certificate()?;
+        return Ok(generate_certificate()?);
```
miraclx: `generate_certificate`, uses blocking operations .._

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,155 @@
+use std::fs;
```
miraclx: \_.. specifically through this .._

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+    }
+
+    if !ip_found {
+        delete_certificate()?;
+        return Ok(generate_certificate()?);
+    }
+
+    let cert_pem = tokio::fs::read(&cert_file).await?;
```
miraclx: _.. prefer `tokio::fs` like you do here

alenmestrov:
I switched from using std::fs to tokio::fs.
Reordered creates alphabetically.

crates/server/src/certificates.rs
alenmestrov: For now, we agreed to leave the certificate as a file for easier installation. For future we can expand the functionality so that user can download the certificate by accessing endpoint of the server.

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,155 @@
+use tokio::fs;
```
miraclx: `cargo +nightly fmt`

crates/server/src/certificates.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    let cert_pem = fs::read(&cert_file).await?;
+    let key_pem = fs::read(&key_file).await?;
+    Ok((cert_pem, key_pem))
+}
+
+fn write_out_instructions() {
+    println!("*******************************************************************************");
+    println!("To install the generated self-signed SSL certificate, follow the steps in our documentation:");
+    println!("https://https://calimero-network.github.io/getting-started/setup");
```
miraclx:
```suggestion
    println!("https://calimero-network.github.io/getting-started/setup");
```

crates/server/src/lib.rs
```diff
@@ -1,11 +1,13 @@
 use std::net::{IpAddr, SocketAddr};
 
 use axum::{http, Router};
+use axum_server::tls_rustls::RustlsConfig;
 use calimero_store::Store;
 use config::ServerConfig;
 use tokio::sync::broadcast;
 use tower_http::cors;
 use tracing::warn;
+pub mod certificates;
```
miraclx, line 11: `cargo +nightly fmt`

alenmestrov:
Fixed typo in instructions
Formatted code with `cargo +nightly fmt`

crates/server/src/admin/storage/ssl.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    Ok(ssl_cert)
+}
+
+pub fn get_ssl(store: Store) -> eyre::Result<Option<SSLCert>> {
+    let entry = SSLEntry::new();
+    let handle = store.handle();
+
+    match handle.get(&entry) {
+        Ok(Some(ssl_document)) => Ok(Some(ssl_document.value().clone())),
```
miraclx: is the `clone` necessary here?

miraclx (approved):

alenmestrov:
Removed unnecessary clone() method

PR #518, opened 2024-07-19
feat(cli): Added auth to dev mode in the cli

petarjuki7:
# Dev Mode Authentification

## Summary

When using the cli, the client signs the timestamp with the private key of the keypair of the node, and appends it to a header (the signature and the timestamp).
On the server side, it verifies the signature using the node's public key.

## Test plan

Just using curl with the dev-mode endpoints doesn't work anymore, they can be accessed only through the cli.


crates/server/src/admin/service.rs
```diff
@@ -111,11 +112,15 @@ pub(crate) fn setup(
...
+        .route_layer(axum::middleware::from_fn(
+            middleware::dev_auth::dev_mode_auth,
+        ));
 
     let admin_router = Router::new()
-        .nest("/", unprotected_router)
-        .nest("/", protected_router)
+        .merge(unprotected_router)
+        .merge(protected_router)
+        .merge(dev_router)
```
MatejVukosav, line 122: Do you still need dev routes? Can you use regular routes?

crates/server/src/middleware/dev_auth.rs
```diff
@@ -0,0 +1,38 @@
...
+    request: Request<axum::body::Body>,
+    next: Next,
+) -> Result<Response, StatusCode> {
+    let public_key = &state.keypair.public();
+
+    let signature = request
+        .headers()
+        .get("X-Signature")
+        .and_then(|v| hex::decode(v).ok())
+        .ok_or(StatusCode::UNAUTHORIZED)?;
```
MatejVukosav, line 21: This is more BAD_REQUEST than unauthorized

crates/meroctl/src/cli/context/list.rs
```diff
@@ -23,13 +24,22 @@ impl ListCommand {
 
         let url = multiaddr_to_url(multiaddr, "admin-api/dev/contexts")?;
         let client = Client::new();
-        let response = client.get(url).send().await?;
+        let keypair = config.identity;
+
+        let timestamp = Utc::now().timestamp().to_string();
```
MatejVukosav, line 29: We are creating auth headers and then later checking if they are correct. What is the purpose of that?

petarjuki7: Not sure if I understand the question.

The auth headers (timestamp and signature) are created on the client side (through the cli) using the node's private key. The server then verifies these headers using the node's public key. That means someone (a bad actor) can't just `curl` our endpoints without access to our nodes private key. The server is not creating the headers, but verifying them.

I might have misunderstood something so open to suggestions.

crates/server/src/admin/service.rs
petarjuki7: I need them because they don't have the same auth as the regular routes. Also, for example, `/dev/install-application` and `/install-application` don't install the app in the same way

Issue #433, opened 2024-07-01
chore: add `rust-toolchain` file


miraclx:
Add a `rust-toolchain` file to mitigate potential compilation issues from compiler version variance across contributors.

All rust source within the workspace must be guaranteed to be compilable on a rust compiler at least as recent as the one specified here.

And [CI testing](https://github.com/calimero-network/core/issues/265) should reflect this, essentially we should treat this version as us being "pinned".

As a consequence, any patch from a contributor that uses advanced features warranting a bump of the toolchain will allow for a conversation.

And in the least, provide visibility.

- [x] define required rust version in rust-toolchain and make sure required rust version is downloaded - 1h

PR #519, opened 2024-07-19
feat: blob store

miraclx:
Setup a blob store for applications to be installed in and introduces a watcher process to the CLI for development mode hot-reload

_lifetimes are the bane of my existence_


packages/calimero-sdk/src/api/nodeApi.ts
miraclx, line 1: @frdomovic, I tweaked these pieces a bit thinking the path of having a generic interface over blob sources would be nice, but the current impl of client sends blob URL is much better.

So for now, these pieces aren't used by the blobstore and are still only used in the login+did flow which means I can undo these changes, but you can lmk if this is a better interface and we should retain it.

crates/store/src/types/blobs.rs
```diff
@@ -0,0 +1,18 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use crate::entry::Borsh;
+use crate::key;
+use crate::types::PredefinedEntry;
+
+#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct BlobMeta {
+    // todo! impl proper entry reference count
+    // pub refs: usize,
```
miraclx, line 10: highlighting for visibility

multiple applications (could) link to the same blob, we need to track those references to know when we can garbage collect blobs (after all the applications referencing it have been deleted)

crates/store/src/types/application.rs
```diff
@@ -0,0 +1,19 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+
+use crate::entry::Borsh;
+use crate::key;
+use crate::types::PredefinedEntry;
+
+#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ApplicationMeta {
+    // todo! impl proper entry reference count
+    // pub refs: usize,
```
miraclx, line 10: highlighting for visibility

multiple contexts (could) link to the same application, we need to track those references to know when we can garbage collect application (after all the contexts referencing it have been deleted)

crates/store/src/iter.rs
```diff
@@ -2,20 +2,29 @@ use std::fmt;
...
+impl<'a, K, V> fmt::Debug for Iter<'a, K, V> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Iter").field(&self.inner).finish()
+    }
+}
+
 pub trait DBIter {
+    // todo! indicate somehow that Key<'a> doesn't contain mutable references to &'a mut self
+    fn seek(&mut self, key: Key) -> eyre::Result<Option<Key>>;
     fn next(&mut self) -> eyre::Result<Option<Key>>;
```
miraclx, line 26: this will require interior mutability, but all iterators will have to guarantee that all slices live at least as long as the iterator

this is guaranteed for rocksdb and our in-memory iterator

crates/store/src/db/memory/raw.rs
```diff
@@ -0,0 +1,217 @@
...
+    type Value: CastsTo<Slice<'a>>;
+
+    fn db(&self) -> &RwLock<InMemoryDBInner<Self::Key, Self::Value>>;
+
+    fn key_from_slice(slice: Slice<'a>) -> Self::Key;
+    fn value_from_slice(slice: Slice<'a>) -> Self::Value;
+}
+
+pub struct DBArena<V> {
+    // todo! Slice::clone points to the same object, can save one allocation here
```
miraclx, line 25: highlighting for visibility, do this with the same efforts of `miraclx/slice/multi-thread-capable`

crates/store/src/db/memory.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+pub struct InMemoryDB<T> {
+    inner: T,
+}
+
+// todo! vvvvv remove this once miraclx/slice/multi-thread-capable is merged in
+unsafe impl<T> Sync for InMemoryDB<T> {}
+unsafe impl<T> Send for InMemoryDB<T> {}
+// todo! ^^^^^ remove this once miraclx/slice/multi-thread-capable is merged in
```
miraclx, line 63: highlighting for visibility

this is in fact, NOT SAFE. Since `Slice` uses `Rc`

[`miraclx/slice/multi-thread-capable`](https://github.com/calimero-network/core/blob/f74f4f48e39a8ad2556881c4a0f599fba346f766/crates/store/src/slice.rs) defines a `Slice` generic over a `ThreadMode` that would fix this, it would've required more effort than I had to integrate that with the codebase as we have it

So this hack works for now, since in prod, we're not using `InMemoryDB`

crates/store/src/db.rs
```diff
@@ -5,30 +5,35 @@ use crate::iter::Iter;
...
     fn get(&self, col: Column, key: Slice) -> eyre::Result<Option<Slice>>;
-    fn put(&self, col: Column, key: Slice, value: Slice) -> eyre::Result<()>;
+    fn put(&self, col: Column, key: Slice<'a>, value: Slice<'a>) -> eyre::Result<()>;
     fn delete(&self, col: Column, key: Slice) -> eyre::Result<()>;
-    fn iter(&self, col: Column, key: Slice) -> eyre::Result<Iter>;
+    fn iter(&self, col: Column) -> eyre::Result<Iter>;
 
-    fn apply(&self, tx: &Transaction) -> eyre::Result<()>;
+    // todo! redesign this, each DB should return a transaction
+    // todo! modelled similar to Iter - {put, delete, clear}
```
miraclx, line 37: highlighting for visibility

crates/store/blobs/src/lib.rs
miraclx, line 1: highlighting for visibility

we have MVP, but there's a lot to come back to here

- `AsyncBufRead` + `AsyncRead` as opposed to raw `Stream`s
- Impl link depth `>1`
- Generic `BlobStore`
- `Blob` should include `size` and `hash`, and still be a stream of bytes.

crates/primitives/src/application.rs
```diff
@@ -1,39 +1,103 @@
...
-pub struct ApplicationId(pub String);
+use crate::blobs::BlobId;
+use crate::hash::{Error as HashError, Hash};
 
-impl Display for ApplicationId {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        Display::fmt(&self.0, f)
+#[derive(Eq, Copy, Hash, Clone, Debug, PartialEq, Serialize, Deserialize)]
+// todo! define macros that construct newtypes
+// todo! wrapping Hash<N> with this interface
```
miraclx, line 13: highlighting for visibility

was working on this in a now stashed commit, but hit some snags as `sha2::Digest` uses a version of `generic-array` that lacks the accommodation for `const-generic`s.

Started to look wonky, so I removed it

Worth revisiting

crates/node/src/lib.rs
```diff
@@ -488,12 +565,29 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
+                        //                   ^^^^~ ContextState<'a> lends the `iter`, while `.entries()` attempts to mutate it
+                        // };
+
+                        for (k, v) in iter.entries() {
+                            let (k, v) = (k?, v?);
+                            if k.context_id() != context_id {
+                                // todo! revisit this when DBIter::seek no longer returns
+                                // todo! the sought item, you have to call next(), read()
+                                continue;
+                            }
```
miraclx, line 673: highlighting for visibility

either `DBIter::seek` mustn't eagerly return a key or the adapters `IterKeys`, `IterEntries` must accommodate that eagerness, and themselves, return without a value.

crates/node/src/lib.rs
```diff
@@ -312,6 +317,73 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
+                                    .map(|ver| ver.to_string())
+                                    .unwrap_or_default(),
+                                c4 = application.source
+                            );
+                            for line in entry.lines() {
+                                println!("{IND} {}", line.cyan());
+                            }
+                        }
+                    }
+                    // todo! a "show" subcommand should help keep "ls" compact
```
miraclx, line 419: highlighting for visibility

consider this at some point

crates/node/src/catchup.rs
```diff
@@ -339,26 +338,23 @@ impl Node {
 
                 if !self
                     .ctx_manager
-                    .is_application_installed(&change.application_id)
+                    .is_application_installed(&change.application_id)?
                 {
+                    // note! for now, we assume all paths are urls
+                    // todo! for path sources, share the blob peer to peer
```
miraclx, line 388: highlighting for visibility

catchups in development mode will attempt to download the "file", which "just might work" locally

worth testing

and if not, well.. odds are, your other node over the local network is in dev mode too

eventually we'll change this to peer-to-peer blob sharing

crates/node/src/catchup.rs
```diff
@@ -65,28 +65,29 @@ impl Node {
...
             let message = serde_json::to_vec(&types::CatchupStreamMessage::ApplicationChanged(
                 types::CatchupApplicationChanged {
                     application_id,
-                    version: application_version,
-                    url,
-                    hash,
+                    blob_id: application.blob,
+                    version: application.version,
+                    source: url,
+                    hash: None, // todo! blob_mgr(application.blob)?.hash
```
miraclx, line 91: highlighting for visibility

we should be able to share the expected hash here once blob mgr returns a hash

crates/context/src/lib.rs
```diff
@@ -283,172 +271,171 @@ impl ContextManager {
...
+        value.application = calimero_store::key::ApplicationMeta::new(application_id);
 
         handle.put(&key, &value)?;
 
         Ok(())
     }
+}
 
-    pub async fn list_installed_applications(
+// vv~ these would be more appropriate in an ApplicationManager
```
miraclx, line 303: highlighting for visibility

crates/context/src/lib.rs
```diff
@@ -283,172 +271,171 @@ impl ContextManager {
...
         &self,
-        application_id: &calimero_primitives::application::ApplicationId,
-    ) -> eyre::Result<semver::Version> {
-        let Some((version, _)) = self.get_latest_application_info(application_id) else {
-            eyre::bail!("failed to get application with id: {}", application_id)
-        };
+        url: Url,
+        version: Option<semver::Version>,
+        // hash: calimero_primitives::hash::Hash,
+        // todo! BlobMgr should return hash of content
```
miraclx, line 355: highlighting for visibility

crates/context/src/lib.rs
```diff
@@ -283,172 +271,171 @@ impl ContextManager {
...
+            return Ok(None);
+        };
+
+        let Some(mut stream) = self.blob_manager.get(application.blob.blob_id()).await? else {
+            eyre::bail!("fatal: application points to dangling blob");
+        };
+
+        // todo! we can preallocate the right capacity here
+        // todo! once `blob_manager::get` -> Blob{size}:Stream
+        let mut buf = vec![];
```
miraclx, line 452: highlighting for visibility

crates/context/src/lib.rs
```diff
@@ -283,172 +271,171 @@ impl ContextManager {
...
+
+        // todo! guard against loading excessively large blobs into memory
```
miraclx, line 454: highlighting for visibility

crates/store/src/types/blobs.rs
MatejVukosav: Can you add those tasks as well as issues in gh so we can track everything there?

crates/context/src/lib.rs
MatejVukosav: What is "vv"? 

 - this and other similar comments should be converted to the issues as otherwise they will get lost

crates/meroctl/src/cli/context/create.rs
```diff
@@ -73,56 +83,137 @@ async fn create_context(
...
+    client: &Client,
+) -> eyre::Result<()> {
+    let (tx, mut rx) = mpsc::channel(1);
+
+    let handle = tokio::runtime::Handle::current();
+    let mut watcher = notify::recommended_watcher(move |evt| {
+        handle.block_on(async {
+            let _ = tx.send(evt).await;
+        })
+    })?;
```
danwilliams, line 119: This is a blocking operation... I am concerned about calling a sync blocking operation in an async context to block on an async operation. Could we consider using `spawn_blocking()` here instead? That would use Tokio's dedicated blocking pool.

danwilliams (changes requested):
This is a big PR, and undocumented. It's quite hard to review. Other than the bits already drawn attention to (such as the todos and what is "vv") I have managed to find a potential blocking issue, but I can't really review the whole shape of the approach as I am not entirely sure what the intent is in some areas.

I would advise breaking things like this into smaller pieces where possible (in future... can't really do that here right now) but also to consider:

  - Can we add Rustdocs for all new things?
  - Can the approach be explained in overarching module documentation?
  - Can there be examples in the documentation? (There is a code example but there are no accompanying docs.)
  - Can the README files be updated to explain what is going on?

I feel quite uneasy about us continuing to add new features, some quite large and significant, with zero documentation.

crates/meroctl/src/cli/context/create.rs
miraclx: `notify` [maintains it's own OS thread](https://github.com/notify-rs/notify/blob/128bf6230c03d39dbb7f301ff7b20e594e34c3a2/notify/src/fsevent.rs#L435-L437), which calls the closure once an event occurs, and that's why we're using `block_on`, and smuggling the tokio runtime handler across thread boundaries.

We shouldn't care about blocking in the closure because it executes in a non-async world, sending events to the async world with our `mpsc` channel which is non-blocking for tokio.

Keep in mind that tokio requires all it's structures to share a runtime environment, so while we (could) use an unbounded sender that can send without awaiting, it won't be able to send across thread boundaries safely without doing that in the tokio environment (rust will pass it, but `tokio` will panic at runtime).

But again, blocking in the closure doesn't affect anything, `notify` is designed with that in mind.

crates/context/src/lib.rs
miraclx: `^`, `<`, `>`, `v`

arrow pointing down

crates/store/src/types/blobs.rs
miraclx: And while I agree about making them issues, I didn't have the time before EOD Friday. And most of them are highlighted for me to come back to because IMO, this is still incomplete. It's an MVP.

I'll follow it up with more non-blocking iterations of much smaller patches.

MatejVukosav: Can u add them now? 

Lets not have private comments in the code as development is dynamic and time allocations are changing.

MatejVukosav (changes requested):
error[E0277]: `[calimero_store::key::BlobMeta]` is not an iterator

miraclx:
> error[E0277]: [calimero_store::key::BlobMeta] is not an iterator

can't reproduce, can you include the command that resulted in this?

MatejVukosav:
> > error[E0277]: [calimero_store::key::BlobMeta] is not an iterator
> 
> can't reproduce, can you include the command that resulted in this?

just  `cargo build` from the root

crates/store/blobs/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        let stream = Box::pin(try_stream!({
+            if blob_meta.links.is_empty() {
+                let maybe_blob = blob_mgr.blob_store.get(id).await;
+                let maybe_blob = maybe_blob.map_err(BlobError::RepoError)?;
+                let blob = maybe_blob.ok_or_else(|| BlobError::DanglingBlob { id })?;
+                return yield blob;
+            }
+
+            for link in blob_meta.links {
```
MatejVukosav, line 136:
```suggestion
           for link in &*blob_meta.links {
```

as [calimero_store::key::BlobMeta]` is not an iterator

miraclx: https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-IntoIterator-for-Box%3C%5BI%5D,+A%3E

`Box<[T]>: IntoIterator<Item = T>` seems to have been stabilized in rust `1.80`, you need to update rust

CleanShot 2024-08-06 at 11 24 20@2x

cc: https://github.com/calimero-network/core/issues/433

MatejVukosav: Ok, it wasn't noted. We have minimum requirements in the docs for this kind of cases: https://calimero-network.github.io/getting-started/setup

Currently, it is 1.79. I'll try with 1.80

MatejVukosav: https://github.com/calimero-network/calimero-network.github.io/pull/11

MatejVukosav (approved):
Open issue to improve code next week https://github.com/calimero-network/core/issues/546

crates/server/src/admin/handlers/context.rs
```diff
@@ -217,23 +216,12 @@ struct JoinContextResponse {
 }
 
 pub async fn join_context_handler(
-    Path(context_id): Path<String>,
+    Path(context_id): Path<calimero_primitives::context::ContextId>,
```
MatejVukosav: Having object as Path type is returning rust error, instead of formatted api error

crates/primitives/src/identity.rs
```diff
@@ -78,17 +78,28 @@ pub struct ContextUser {
...
 }
 
-#[derive(Debug, Deserialize, PartialEq, Serialize, Clone, Copy)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "UPPERCASE")]
 #[serde(tag = "type")]
 pub enum WalletType {
-    NEAR,
+    NEAR {
+        network_id: NearNetworkId,
```
MatejVukosav: This killed login in sdk

MatejVukosav: removed as not related to this PR

PR #521, opened 2024-07-19
feat: subscribe to meta topic

MatejVukosav:
.


crates/node/src/lib.rs
```diff
@@ -77,6 +77,19 @@ pub async fn start(config: NodeConfig) -> eyre::Result<()> {
 
     let mut stdin = tokio::io::BufReader::new(tokio::io::stdin()).lines();
 
+    match network_client
+        .subscribe(IdentTopic::new(
```
frdomovic, line 81: Is this just a workaround or do we plan to leave it like this?

crates/node/src/lib.rs
```diff
@@ -77,6 +77,19 @@ pub async fn start(config: NodeConfig) -> eyre::Result<()> {
...
+            "9df3h13HTCTgEoEnX77VaSoWrVkNxFiQmtuVGLorjycX".to_string(),
```
miraclx: where's this derived from?

MatejVukosav: Check description

miraclx: Understood, but then we can use non-valid context IDs here, what do you think?

MatejVukosav: Updated

crates/node/src/lib.rs
MatejVukosav: For now, we don't have a different solution in the plan.

Issue #477, opened 2024-07-16
App installation on client side


MatejVukosav:
Instead of the node making assumptions on the application registry and contract (currently fetching from NEAR testnet on a pre-defined contract `calimero-package-manager.testnet`), the new design of application install from #461 allows the node to consume and store streams of arbitrary data.

The new model for app installation should be:

- From the client side, armed with knowledge of the user's registry source (NEAR mainnet / testnet or wherever else) and the registry contract address, the client retrieves the URL for the application.
- The client makes a request to the node, opening up a stream for a blob upload.
- The client makes a request to the blob URL, piping it's response stream to the node.
- The node responds with a blob ID.
- The client makes a request to the node once more, to create an application record, passing it the `BlobID` and other metadata like a source URL.
- The node responds with an `ApplicationID` == `BlobID`
- The client can now create contexts using the `ApplicationID`

Notable changes in client:
   - frontend : get stream from ipfs and send it to node 

Blocked on #461

frdomovic:
```mermaid
sequenceDiagram
    title App Installation Workflow

    participant Client
    participant Package Manager Contract
    participant Node
    participant IPFS

    Client->>Package Manager Contract: Retrieve IPFS URL for application
    Package Manager Contract->>Client: Application IPFS URL
    Client->>Node: Open stream for blob upload
    Node-->>Client: Blob URL
    Client->>IPFS: Request blob
    IPFS-->>Client: Blob response stream
    Client->>Node: Pipe blob response stream to node
    Node-->>Client: Blob ID
    Client->>Node: Create application record (BlobID & metadata)
    Node-->>Client: ApplicationID == BlobID
    Client->>Node: Create contexts using ApplicationID
```

frdomovic:
cc @miraclx 
```mermaid
sequenceDiagram
    title App Installation Workflow

    participant Client
    participant Package Manager Contract
    participant Node
    participant IPFS

    Client->>Package Manager Contract: Retrieve IPFS URL for application
    Package Manager Contract->>Client: Application IPFS URL
    Client->>Node: Open stream for blob upload
    Node->>Node: Opening stream
    Node-->>Client: Blob URL
    Client->>IPFS: Request blob
    IPFS-->>Client: Blob response stream
    Client->>Node: Pipe blob response stream to node
    Node-->>Client: Blob ID
    Client->>Node: Create application record (BlobID & metadata)
    Node-->>Client: ApplicationID == BlobID
[...]

PR #530, opened 2024-07-22
feat(admin-api): update install application logic

frdomovic:
# feat(admin-api): update install application logic

## Summary
Goal is that the node does not have to know where the application binary is hosted (testnet / mainnet NEAR view calls). Node should only get the path and download the release from the path + compare downloaded hash with hash passed in the request. 

With this update we don't need function to fetch application release to get that metadata because its passed in the API request. 

Connected with [this](https://github.com/calimero-network/admin-dashboard/pull/43) PR. 

Fixes # [Issue](https://github.com/calimero-network/core/issues/477)

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -252,7 +252,8 @@ impl ContextManager {
         path: &str,
         hash: Option<&str>,
     ) -> eyre::Result<()> {
-        self.download_and_install_release(&application_id, &version, &path, hash).await?;
+        self.download_and_install_release(&application_id, &version, &path, hash)
+            .await?;
 
         Ok(())
     }
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -249,10 +249,10 @@ impl ContextManager {
         application_id: &calimero_primitives::application::ApplicationId,
         // todo! permit None version for latest
         version: &semver::Version,
+        path: &str,
+        hash: &str,
     ) -> eyre::Result<()> {
-        let release = self.get_release(&application_id, version).await?;
-
-        self.download_release(&application_id, &release).await?;
+        self.download_and_install_release(&application_id, &version, &path, &hash).await?;
 
         Ok(())
     }
@@ -345,48 +345,16 @@ impl ContextManager {
         Ok(version)
     }
 
-    async fn get_release(
+    async fn download_and_install_release(
         &self,
         application_id: &calimero_primitives::application::ApplicationId,
         version: &semver::Version,
-    ) -> eyre::Result<calimero_primitives::application::Release> {
-        // todo! the node shouldn't know anything about where
-        // todo! apps are to be sourced from, keep it generic
-
-        let client = JsonRpcClient::connect("https://rpc.testnet.near.org");
-        let request = methods::query::RpcQueryRequest {
-            block_reference: BlockReference::Finality(Finality::Final),
-            request: QueryRequest::CallFunction {
-                account_id: "calimero-package-manager.testnet".parse()?,
-                method_name: "get_release".to_string(),
-                args: FunctionArgs::from(
-                    serde_json::json!({
-                        "id": application_id,
-                        "version": version.to_string()
-                    })
-                    .to_string()
-                    .into_bytes(),
-                ),
-            },
-        };
-
-        let response = client.call(request).await?;
-
-        let QueryResponseKind::CallResult(result) = response.kind else {
-            eyre::bail!("Failed to fetch data from the rpc endpoint")
-        };
-
-        Ok(serde_json::from_slice(&result.result)?)
-    }
-
-    async fn download_release(
-        &self,
-        application_id: &calimero_primitives::application::ApplicationId,
-        release: &calimero_primitives::application::Release,
+        path: &str,
+        hash: &str,
     ) -> eyre::Result<bool> {
         // todo! download to a tempdir
         // todo! Blob API
-        let base_path = format!("{}/{}/{}", self.config.dir, application_id, release.version);
+        let base_path = format!("{}/{}/{}", self.config.dir, application_id, version);
         fs::create_dir_all(&base_path)?;
 
         let file_path = format!("{}/binary.wasm", base_path);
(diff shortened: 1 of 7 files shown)


crates/node/src/catchup.rs
```diff
@@ -187,6 +202,8 @@ impl Node {
                     application_id: Some(context.application_id),
                     last_executed_transaction_hash: context.last_transaction_hash,
                     batch_size: self.network_client.catchup_config.batch_size,
+                    path: None,
```
frdomovic: Unclear what to do with this part

crates/node/src/catchup.rs
```diff
@@ -196,6 +213,8 @@ impl Node {
                     application_id: None,
                     last_executed_transaction_hash: calimero_primitives::hash::Hash::default(),
                     batch_size: self.network_client.catchup_config.batch_size,
+                    path: None,
```
frdomovic: And this

crates/context/src/lib.rs
```diff
@@ -396,16 +364,16 @@ impl ContextManager {
 
         let mut file = File::create(&file_path)?;
 
-        let mut response = reqwest::Client::new().get(&release.path).send().await?;
+        let mut response = reqwest::Client::new().get(path.to_string()).send().await?;
```
```suggestion
        let mut response = reqwest::Client::new().get(path).send().await?;
```

crates/context/src/lib.rs
```diff
@@ -345,48 +345,16 @@ impl ContextManager {
...
-
-        Ok(serde_json::from_slice(&result.result)?)
-    }
-
-    async fn download_release(
-        &self,
-        application_id: &calimero_primitives::application::ApplicationId,
-        release: &calimero_primitives::application::Release,
+        path: &str,
+        hash: &str,
```
miraclx: let's make hash optional here, since it's unnecessary for local file paths

crates/node/src/catchup.rs
miraclx: I think we can do without this piece

crates/node/src/catchup.rs
miraclx: same here, we already have `application_id`

crates/node/src/lib.rs
```diff
@@ -372,15 +373,17 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                     }
                     "create" => {
-                        let Some((context_id, application_id, version)) = args.and_then(|args| {
+                        let Some((context_id, application_id, version, path, hash)) = args.and_then(|args| {
                             let mut iter = args.split(' ');
                             let context = iter.next()?;
                             let application = iter.next()?;
                             let version = iter.next()?;
+                            let path = iter.next()?;
+                            let hash= iter.next()?;
```
miraclx: `cargo +nightly fmt`

frdomovic: updated

crates/context/src/lib.rs
frdomovic: Added as optional and updated other parts of code as well

crates/context/src/lib.rs
frdomovic: updated

crates/node/src/lib.rs
```diff
@@ -372,15 +373,19 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
+                                let context = iter.next()?;
+                                let application = iter.next()?;
+                                let version = iter.next()?;
+                                let path = iter.next()?;
+                                let hash = iter.next()?;
+
+                                Some((context, application, version, path, hash))
+                            })
+                        else {
+                            println!("{IND} Usage: context create <context_id> <application_id> <version> <path> <hash>");
```
miraclx: we don't need hash here

crates/node/src/lib.rs
```diff
@@ -385,7 +385,7 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
                                 Some((context, application, version, path, hash))
                             })
                         else {
-                            println!("{IND} Usage: context create <context_id> <application_id> <version> <path> <hash>");
+                            println!("{IND} Usage: context create <context_id> <application_id> <version> <path>");
```
miraclx: I meant we don't need hash in the CLI at all, so it doesn't need to be parsed in 385 either

frdomovic: Updated

crates/node/src/lib.rs
frdomovic: Updated

frdomovic:
Added additional comments next to path objects until further updates

crates/server-primitives/src/admin.rs
```diff
@@ -6,6 +6,8 @@ use serde_json::Value;
 pub struct InstallApplicationRequest {
     pub application: calimero_primitives::application::ApplicationId, // TODO: rename to application_id
     pub version: semver::Version,
+    pub path: String,  // represents the url path to the release binary (e.g. ipfs path)
```
miraclx: why not just use url here? it's fine as-is, just checking

crates/node/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 use calimero_primitives::events::OutcomeEvent;
+use calimero_primitives::hash;
```
miraclx: unused import?

crates/node/src/catchup.rs
```diff
@@ -65,6 +65,18 @@ impl Node {
 
         let application_id = context.application_id.clone();
 
+        let path = if let Some(path) = request.path {
+            path
+        } else {
+            eyre::bail!("Path is missing in the request")
+        };
```
miraclx:
```suggestion
        let Some(path) = request.path else {
            eyre::bail!("Path is missing in the request")
        };
```

crates/node/src/catchup.rs
```diff
@@ -65,6 +65,18 @@ impl Node {
...
+
+        let hash = if let Some(hash) = request.hash {
```
miraclx: same here

crates/server-primitives/src/admin.rs
frdomovic: i actually updated the name to use url so its more clear. path can also represent file dirs just to avoid confusion what this param is...

crates/node/src/catchup.rs
frdomovic: updated

crates/node/src/catchup.rs
frdomovic: updated

crates/node/src/lib.rs
frdomovic: removed

miraclx (approved):

crates/server-primitives/src/admin.rs
miraclx: Aha, in my PR I'm using source URIs. Will coalesce.

Issue #247, opened 2024-05-02
Define the context initialization process at the app level


miraclx:
Right now, `read-only` calls on uninitialized state panics at runtime, while `mutable` calls on "previously"-uninitialized state will delegate to calling `<{AppState} as Default>::default` which makes an assumption on the capacity for the state to be initialized with "default" params. Which may not always be the case.

So.

We need to specify how applications get their initial state.

One option is to use a reserved method called `new` which is idiomatic for Rust, but that would be a weird pattern in other SDKs when we decide to have them

Another approach is to have a custom annotation like `#[app::init]` that informs the codegen logic to treat the return type as the state to be written.

Worth noting that annotated methods *MUST* return `Self`.

And calling them from the node/clients will always return a context ID.

Linked:
- #248

MatejVukosav:
- app init - schema is currently too much work - rather for dev preview call init method with parameters
  - couple context creation with initialization
  - add to docs example of init parameters for our apps  ( op, rps)
  - send params to init function 
  - add init to sdk macro for new init attribute (add init method (mutate method) to initialize contract )

miraclx:
Alright, so I did some thinking.. and being able to manually implement `AppStateInit` without any init method actually being defined is unideal.

This is a non-blocker for this PR, though it's worth thinking through what a better implementation would be.

If we move the initializer method to `AppStateInit`:

```rust
pub trait AppStateInit: Sized {
    type InitParams<'a>: Deserialize<'a>;

    fn init<'a>(params: Self::InitParams<'a>) -> Self;
}
```

and since this returns an explicit `Self`, we can relinquish `Identity` of it's duty _<sub>thanks for your service</sub>_

Now, instead of code-generating an `init` method from the method, we'll move that to `#[app::state]` instead.


```rust
#[cfg(target_arch = "wasm32")]
#[no_mangle]
pub extern "C" fn init() {
    ::calimero_sdk::env::setup_panic_hook();

    ::calimero_sdk::event::register::<#ident>();

    let Some(input) = ::calimero_sdk::env::input() else {
        ::calimero_sdk::env::panic_str("Expected input since method has arguments.")
    };

    let params = match ::calimero_sdk::serde_json::from_slice(&input) {
        Ok(value) => value,
        Err(err) => ::calimero_sdk::env::panic_str(
            &format!("Failed to deserialize input from JSON: {:?}", err)
        ),
    };

    let state = <#ident as ::calimero_sdk::state::AppStateInit>::init(params);

[...]

PR #532, opened 2024-07-23
feat(sdk): app::init codegen

petarjuki7:
# Initialization function in sdk

## Summary

Added #[app::init] attribute, needed for initialization of application.

Fixes https://github.com/calimero-network/core/issues/247

## Improvements

It generates code that expects you to return Self, could make it enforced during compilation (building of ast).


crates/sdk/macros/src/lib.rs
```diff
@@ -25,10 +25,38 @@ pub fn logic(args: TokenStream, input: TokenStream) -> TokenStream {
     reserved::init();
     let _args = parse_macro_input!({ input } => args as items::Empty);
     let block = parse_macro_input!(input as syn::ItemImpl);
-    let tokens = match logic::LogicImpl::try_from(logic::LogicImplInput { item: &block }) {
+
+    // Find the #[app::init] method
+    let init_method = match block.items.iter().find_map(|item| {
```
miraclx: A pattern you'll notice that defines the design of our macros is delegation.

This function delegates to `LogicImpl` which delegates to another parser for each individual function in the impl block.

This should've been in `LogicImpl` instead (after parsing all the `fn`s)

crates/sdk/macros/src/lib.rs
```diff
@@ -25,10 +25,38 @@ pub fn logic(args: TokenStream, input: TokenStream) -> TokenStream {
...
+        }
+    }) {
+        Some(method) => method,
+        None => {
+            return syn::Error::new(
+                proc_macro2::Span::call_site(),
+                "An #[app::init] method is required",
+            )
+            .to_compile_error()
+            .into()
```
miraclx: also, it should be possible to break up `#[app::logic]` into multiple blocks, so we can't make the assumption that the absence of the `init` method in this one, means there's not one in another

would be nice if we're able to add a compile-time assertion from `#[app::state]` that `<Ty>::init` is defined

As for now, I'm not sure exactly what that would look like since `Fn<A: Tuple>` and `Tuple` is nightly-only, so we might not be able to do it without making a custom trait for all `Fn` with assumptions on parameter count.

crates/sdk/macros/src/logic.rs
```diff
@@ -107,25 +110,54 @@ impl<'a> TryFrom<LogicImplInput<'a>> for LogicImpl<'a> {
...
+            type_: &type_,
+            item: input.init_method,
+        }) {
+            Ok(method::LogicMethod::Init(method)) => method,
+            _ => {
+                errors.subsume(syn::Error::new_spanned(
+                    input.init_method,
+                    "The #[app::init] method is not properly defined",
+                ));
+                return Err(errors);
```
miraclx: this is not a fatal parse error, collect it and proceed (one of the design considerations is to be as exhaustive as possible in a single pass)

crates/sdk/macros/src/logic.rs
```diff
@@ -107,25 +110,54 @@ impl<'a> TryFrom<LogicImplInput<'a>> for LogicImpl<'a> {
...
+        // Process other methods
         for item in &input.item.items {
             if let syn::ImplItem::Fn(method) = item {
-                match method::LogicMethod::try_from(method::LogicMethodImplInput {
-                    type_: &type_,
-                    item: method,
+                if !method.attrs.iter().any(|attr| {
+                    attr.path().segments.len() == 2
+                        && attr.path().segments[0].ident == "app"
+                        && attr.path().segments[1].ident == "init"
```
miraclx: why not leave this to `LogicMethod`?

crates/sdk/macros/src/logic/method.rs
```diff
@@ -6,6 +6,7 @@ use crate::{errors, reserved};
 pub enum LogicMethod<'a> {
     Public(PublicLogicMethod<'a>),
     Private,
+    Init(InitMethod<'a>),
```
miraclx: why not merge this with `PublicLogicMethod`?

crates/sdk/macros/src/logic/method.rs
```diff
@@ -19,6 +20,51 @@ pub struct PublicLogicMethod<'a> {
...
+        quote! {
+            #[cfg(target_arch = "wasm32")]
+            #[no_mangle]
+            pub extern "C" fn init() {
+                ::calimero_sdk::env::setup_panic_hook();
+
+                ::calimero_sdk::event::register::<#self_>();
+
+                let input = ::calimero_sdk::env::input().expect("Init parameters are required");
+                let (#(#arg_idents),*): (#(#arg_types),*) = ::calimero_sdk::serde_json::from_slice(&input)
```
miraclx: ah, neat trick.. helps avoid the struct definition

crates/sdk/macros/src/lib.rs
miraclx: On second thought, you can define a marker trait like `AppStateInit` that only gets implemented when a method is annotated with `#[app::init]` to guard against duplication.

And make `AppState: AppStateInit` so it requires that to be implemented.

Then we can use the `on_unimplemented` debug assertion introduced in Rust 1.78 to customize the error message.

https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#diagnostic-attributes

miraclx (commented):
PS: we also want to do away with the `.unwrap_or_default()` accommodation for `&mut self` methods.

As well as update the apps in this repo.

petarjuki7:
> PS: we also want to do away with the `.unwrap_or_default()` accommodation for `&mut self` methods.
> 
> As well as update the apps in this repo.

Haven't added init methods yet, but changed `.unwrap_or_default()` in  `&mut self` methods.

crates/sdk/macros/src/lib.rs
petarjuki7: I'm having a problem with circular dependencies if I do `AppState: AppStateInit`, and then try to implement `AppStateInit` when I find the `#[app::init]` func  in the `macros` sdk.

I may not be doing what was intended, so feedback appreciated.

miraclx: You want the reverse, not this one, so - `AppState: AppStateInit` instead

crates/sdk/macros/src/logic/method.rs
```diff
@@ -113,7 +113,12 @@ impl<'a> ToTokens for PublicLogicMethod<'a> {
             Some(type_) => {
                 let def = match type_ {
```
miraclx: seems to me, this match statement is now irrelevant?

crates/sdk/src/state.rs
```diff
@@ -2,6 +2,12 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
 
-pub trait AppState: Default + BorshSerialize + BorshDeserialize {
+pub trait AppState: Default + BorshSerialize + BorshDeserialize + AppStateInit {
     type Event<'a>: AppEvent + 'a;
 }
+
+#[diagnostic::on_unimplemented(
+    message = "The type `{Self}` doesn't have an `#[app::init]` method",
+    label = "add an `#[app::init]` method to this type"
+)]
```
miraclx: can you include screenshots of this? so we know it behaves correctly?

crates/sdk/macros/src/logic/method.rs
petarjuki7: Should we just `.unwrap()` then? If we can guarantee that the apps state was initialized.

crates/sdk/macros/src/logic.rs
```diff
@@ -9,25 +9,19 @@ mod ty;
...
-            methods,
-            ..
-        } = self;
+        let LogicImpl { orig, methods, .. } = self;
 
         quote! {
             #orig
 
             #(#methods)*
+
```
miraclx:
```suggestion
```

crates/sdk/src/state.rs
petarjuki7: Screenshot 2024-07-25 at 17 30 27

I have the init method commented out, and we can see that the error message shows

miraclx: Amazing! But, please make this match the format for other types of error reports.

apps/gen-ext/src/lib.rs
```diff
@@ -11,6 +11,10 @@ pub struct GenExt;
 
 #[app::logic]
 impl GenExt {
+    #[app::init]
+    pub fn init() -> GenExt {
+        GenExt
+    }
```
miraclx, line 17:
```suggestion
    }

```

and the others

crates/sdk/macros/src/logic/method.rs
```diff
@@ -8,6 +8,10 @@ pub enum LogicMethod<'a> {
     Private,
 }
 
+pub enum Modifer {
+    Init(),
```
miraclx:
```suggestion
pub enum Modifier<'a> {
    Init(&'a syn::Path),
```

any reason this isn't this?

crates/sdk/macros/src/logic/method.rs
```diff
@@ -18,51 +22,35 @@ pub struct PublicLogicMethod<'a> {
...
-                        }
-                }.to_tokens(tokens)
-        }
+        let init_method = if modifiers.is_empty() {
+            false
+        } else {
+            modifiers
+                .iter()
+                .any(|modifier| matches!(modifier, Modifer::Init()))
+        };
```
miraclx: I think this is the same?

```suggestion
        let init_method = modifiers
            .iter()
            .any(|modifier| matches!(modifier, Modifer::Init(_)));
```

crates/sdk/macros/src/logic/method.rs
```diff
@@ -101,55 +89,66 @@ impl<'a> ToTokens for PublicLogicMethod<'a> {
...
-                        .into_result()
-                        .to_json()
-                    {
-                        Ok(output) => output,
-                        Err(err) => ::calimero_sdk::env::panic_str(
-                            &format!("Failed to serialize output to JSON: {:?}", err)
-                        ),
-                    }
+            //only when it's not init
+            if !init_method {
```
miraclx:
```rust
        if let (Some(_), false) = (&self.ret, init_method) {
```

crates/sdk/macros/src/logic/method.rs
```diff
@@ -101,55 +89,66 @@ impl<'a> ToTokens for PublicLogicMethod<'a> {
...
+            }
         }
 
-        let state_finalizer = match &self.self_type {
-            Some(arg::SelfType::Mutable) => quote! {
-                ::calimero_sdk::env::state_write(&app);
-            },
-            _ => quote! {},
+        let state_finalizer = {
+            if init_method == true || matches!(&self.self_type, Some(arg::SelfType::Mutable)) {
```
miraclx:
```suggestion
            if init_method || matches!(&self.self_type, Some(arg::SelfType::Mutable)) {
```

or

```rust
        let state_finalizer = match (&self.self_type, init_method) {
            (Some(arg::SelfType::Mutable), _) | (_, true) => quote! {
                ::calimero_sdk::env::state_write(&app);
            },
            _ => quote! {},
        };
```

crates/sdk/macros/src/logic/method.rs
```diff
@@ -101,55 +89,66 @@ impl<'a> ToTokens for PublicLogicMethod<'a> {
...
+                quote! {
+                    ::calimero_sdk::env::state_write(&app);
+                }
+            } else {
+                quote! {}
+            }
+        };
+
+        let init_impl = {
+            if init_method {
```
miraclx:
```suggestion
        let init_impl = if init_method {
```

crates/sdk/macros/src/logic/method.rs
```diff
@@ -168,6 +167,7 @@ impl<'a> ToTokens for PublicLogicMethod<'a> {
 
                 #state_finalizer
             }
+            #init_impl
```
miraclx, line 160: let's space this out

crates/sdk/macros/src/logic/method.rs
```diff
@@ -257,9 +257,24 @@ impl<'a, 'b> TryFrom<LogicMethodImplInput<'a, 'b>> for LogicMethod<'a> {
...
-        let attrs = &input.item.attrs;
+        for attr in &input.item.attrs {
+            if attr.path().segments.len() == 2
+                && attr.path().segments[0].ident == "app"
+                && attr.path().segments[1].ident == "init"
+            {
+                if let Some(_self_type) = &self_type {
+                    errors.subsume(syn::Error::new_spanned(
+                        input.item,
+                        "The argument to #[app::init] method can't be self or mut self",
```
miraclx: Use `errors::ParseError` and \`escape\` code references

crates/sdk/src/state.rs
```diff
@@ -2,13 +2,18 @@ use borsh::{BorshDeserialize, BorshSerialize};
...
 
-pub trait AppState: Default + BorshSerialize + BorshDeserialize //+ AppStateInit
-{
+pub trait AppState: Default + BorshSerialize + BorshDeserialize + AppStateInit {
     type Event<'a>: AppEvent + 'a;
 }
 
+pub trait Identity<This> {}
+
+impl<T> Identity<T> for T {}
```
miraclx:
```suggestion
pub trait Identity<This = Self> {}

impl<T: AppState> Identity<T> for T {}
```

which now satisfies the reverse constraint - `AppStateInit: AppState`

miraclx: Can you screenshot what it's like to implement `AppStateInit` manually?

In case they try to bypass using `#[app::init]`

crates/sdk/macros/src/logic/method.rs
miraclx: I suspect it's because it's unused, in which case you can remove the `()`

petarjuki7: Yes, it's unused, I'm going to remove the brackets

crates/sdk/macros/src/logic/method.rs
```diff
@@ -215,6 +257,23 @@ impl<'a, 'b> TryFrom<LogicMethodImplInput<'a, 'b>> for LogicMethod<'a> {
...
         }
 
+        let mut modifiers = vec![];
+
+        for attr in &input.item.attrs {
+            if attr.path().segments.len() == 2
+                && attr.path().segments[0].ident == "app"
+                && attr.path().segments[1].ident == "init"
+            {
+                if let Some(_self_type) = &self_type {
```
miraclx: another thing to check for is that the `init` method is `pub`lic

crates/sdk/src/state.rs
petarjuki7: It would throw no errors

For example if you just write:
```
impl AppStateInit for KvStore {
    type Return = KvStore;
}
```

It throws no errors

crates/sdk/macros/src/logic/method.rs
petarjuki7: https://github.com/calimero-network/core/blob/d107917304df78a1ee75c5d8ee7fd4702dc98cbd/crates/sdk/macros/src/logic/method.rs#L144

Is it fine because here I already make sure it is public here?

miraclx: Aha, right.. we're not bothering to parse private methods

Although, there's a concern here..

Annotating private methods with `#[app::init]` won't indicate any errors.

And you can annotate any number of them. I'll leave it to you to determine how you want to address this.

crates/sdk/macros/src/errors.rs
```diff
@@ -70,6 +70,8 @@ pub enum ParseError<'a> {
     NoComplexVisibility,
     #[error("explicit ABIs are not supported")]
     NoExplicitAbi,
+    #[error("argument to `#[app::init]` method can't be `self` or `mut self`")]
+    SelfNotPermitted,
```
miraclx: perhaps we don't need to mention the macro? and since the error message is specific to init, maybe expand on the variant name? for example:

```suggestion
    #[error("an initializer, by definition, has no `self` to reference")]
    NoSelfReceiverAtInit,
```

but the final form is up to you

crates/sdk/src/state.rs
```diff
@@ -6,12 +6,12 @@ pub trait AppState: Default + BorshSerialize + BorshDeserialize + AppStateInit {
...
 
-pub trait Identity<This> {}
+pub trait Identity<This = Self> {}
 
-impl<T> Identity<T> for T {}
+impl<T: AppState> Identity<T> for T {}
 
 #[diagnostic::on_unimplemented(
-    message = "The type `{Self}` doesn't have an `#[app::init]` method",
+    message = "no method named `#[app::init]` found for type `{Self}`",
```
miraclx: `no method annotated`

but.. good highlight, we also need to ensure the annotated method is actually called "init"

I can't think of any good arguments for making this configurable atm

You'll also need to forbid the use of function identifiers named `init` when it's not annotated with the macro

You might ask.. why do we need the macro then? why not just check the function name to determine it's behaviour.

The idea is we don't want to leave anything implicit so while we "could" assume that your `init` method initializes the app state, you may not have had that in mind when writing the app.

So, we error on `pub fn init` with no annotation, suggesting they specify the annotation to fix it.

crates/sdk/src/state.rs
```diff
@@ -6,12 +6,12 @@ pub trait AppState: Default + BorshSerialize + BorshDeserialize + AppStateInit {
...
     label = "add an `#[app::init]` method to this type"
```
miraclx, line 15: this piece might be worth revisiting depending on how `rustc` displays this, since it might just be rephrasing the `message`

crates/sdk/src/state.rs
```diff
@@ -2,6 +2,18 @@ use borsh::{BorshDeserialize, BorshSerialize};
 
 use crate::event::AppEvent;
 
-pub trait AppState: Default + BorshSerialize + BorshDeserialize {
+pub trait AppState: Default + BorshSerialize + BorshDeserialize + AppStateInit {
```
miraclx: we no longer need `Default` here

```suggestion
pub trait AppState: BorshSerialize + BorshDeserialize + AppStateInit {
```

miraclx:
Alright, so I did some thinking.. and being able to manually implement `AppStateInit` without any init method actually being defined is unideal.

This is a non-blocker for this PR, though it's worth thinking through what a better implementation would be.

If we move the initializer method to `AppStateInit`:

```rust
pub trait AppStateInit: Sized {
    type InitParams<'a>: Deserialize<'a>;

    fn init<'a>(params: Self::InitParams<'a>) -> Self;
}
```

and since this returns an explicit `Self`, we can relinquish `Identity` of it's duty _<sub>thanks for your service</sub>_

Now, instead of code-generating an `init` method from the method, we'll move that to `#[app::state]` instead.


```rust
#[cfg(target_arch = "wasm32")]
#[no_mangle]
pub extern "C" fn init() {
    ::calimero_sdk::env::setup_panic_hook();

    ::calimero_sdk::event::register::<#ident>();

    let Some(input) = ::calimero_sdk::env::input() else {
        ::calimero_sdk::env::panic_str("Expected input since method has arguments.")
    };

    let params = match ::calimero_sdk::serde_json::from_slice(&input) {
        Ok(value) => value,
        Err(err) => ::calimero_sdk::env::panic_str(
            &format!("Failed to deserialize input from JSON: {:?}", err)
        ),
    };

    let state = <#ident as ::calimero_sdk::state::AppStateInit>::init(params);

    ::calimero_sdk::env::state_write(&state);
}
```

With that handled, `#[app::init]` becomes the `AppStateInit` definition.

```rust
#[app::logic]
impl Adder {
    #[app::init]
    pub fn init(a: u8, b: u8) -> Adder {
        Adder { a, b }
    }

    pub fn add(&self) -> u8 {
        self.0 + self.1
    }
}
```

expanding to

```rust
impl Adder {
    pub fn init(a: u8, b: u8) -> Adder {
        Adder { a, b }
    }

    pub fn add(&self) -> u8 {
        self.0 + self.1
    }
}

const _: () = {
    struct CalimeroInput {
        a: u8,
        b: u8,
    }

    impl ::calimero_sdk::state::AppStateInit for Adder {
        type InitParams<'a> = CalimeroInput;

        fn init<'a>(
            CalimeroInput {
                a,
                b,
            }: Self::InitParams<'a>,
        ) -> Self {
            Adder::init(a, b)
        }
    }
};

#[cfg(target_arch = "wasm32")]
#[no_mangle]
pub extern "C" fn add() {
    // Snip..
}
```

This forces implementors of `AppStateInit` to also provide the actual initialization logic.

The unfortunate downside of this is, we'll lose the ability to define generics for `#[app::state]`

Although, I suppose there's no good arguments for having them in the first place? Just that we "could".

Even though `#[app::logic]` already forbids the use of generics. So it might make sense to outlaw them for `#[app::state]` to open up this possibility.

crates/sdk/macros/src/logic/method.rs
miraclx: For posterity, the reason we use struct definitions is so you can provide a JSON payload with the field names in an object instead of just an array of parameters.

Worth noting: `serde` also attempts to deserialize an array of parameters, so we get the best of both worlds.

crates/sdk/src/state.rs
miraclx: By this, I mean prefix the tag, so users can tell which errors are `rustc` errors and which are `calimero`-flavored.

https://github.com/calimero-network/core/blob/d107917304df78a1ee75c5d8ee7fd4702dc98cbd/crates/sdk/macros/src/errors.rs#L36

petarjuki7:
> ```rust
> pub trait AppStateInit: Sized {
>     type InitParams<'a>: Deserialize<'a>;
> 
>     fn init<'a>(params: Self::InitParams<'a>) -> Self;
> }
> ```

I was thinking of something similar, but maybe the problem is that now the user couldn't use the init method without any params.

For example how we are initializing our apps right now:
```rust
#[app::init]
pub fn init() -> KvStore {
    KvStore::default()
}
 ```
 
I agree that generics for `#[app::state]` are not needed.

**Comment by miraclx** [comment]
> but maybe the problem is that now the user couldn't use the init method without any params.

it's still possible, the macro still passes the parameters to the `AppStateInit` definition

```rust
        fn init<'a>(
            CalimeroInput {
                a,
                b,
            }: Self::InitParams<'a>,
        ) -> Self {
            Adder::init(a, b)
        }
```

crates/sdk/macros/src/logic/method.rs
```diff
@@ -237,6 +259,20 @@ impl<'a, 'b> TryFrom<LogicMethodImplInput<'a, 'b>> for LogicMethod<'a> {
...
+        match (is_init, &self_type, name.to_string().as_str()) {
+            (true, Some(_), _) => errors.subsume(syn::Error::new_spanned(
+                input.item,
+                errors::ParseError::NoSelfReceiverAtInit,
+            )),
+            (false, _, "init") => errors.subsume(syn::Error::new_spanned(
+                input.item,
+                errors::ParseError::InitMethodWithoutNoInitAttribute,
+            )),
+            _ => {}
```
miraclx: `Ident: PartialEq<T>` where `T: AsRef<str>` [(src)](https://docs.rs/syn/latest/syn/struct.Ident.html#impl-PartialEq%3CT%3E-for-Ident)

```rust
match (is_init, &self_type) {
    (true, Some(_)) => errors.subsume(syn::Error::new_spanned(
        input.item,
        errors::ParseError::NoSelfReceiverAtInit,
    )),
    (true, None) if name != "init" => { /* forbid this */ },
    (false, _) if name == "init" => errors.subsume(syn::Error::new_spanned(
        input.item,
        errors::ParseError::InitMethodWithoutNoInitAttribute,
    )),
    _ => {}
```

petarjuki7: Ah, okay, we are also making sure the actual function is named init
Fixed!

miraclx (approved):

miraclx:
Reposted https://github.com/calimero-network/core/issues/532#issuecomment-2251864232 in #247. We'll continue the conversation there, and follow it up with it's implementation.

Issue #177, opened 2024-04-23
Runtime Identity


chefsale:
Expose a host method to the guest for fetching the executor identity from inside the runtime

And another method for signing messages with the executor's identity
- Note that this won't be possible under the current implementation of the transactional model without optimistic execution, so this is blocked by #440

Currently 
- we have node identity stored in config file
- identity should be exposed in runtime
- runtime can sign transactions and verify them

PR #536, opened 2024-07-27
feat: runtime identity

danwilliams:
Adds runtime identity to satisfy ticket #177.


crates/store/src/types/context.rs
```diff
@@ -54,11 +54,6 @@ impl PredefinedEntry for key::ContextState {
     type DataType<'a> = ContextState<'a>;
 }
 
-#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
-pub struct ContextIdentity {
-    pub private_key: Option<[u8; 32]>,
-}
```
miraclx: please restore this, the separation is intentional to prevent mindless modification of the storage format which could be breaking changes (especially since we're using `borsh`), we have the same consideration for wire types.

For clarity, this is the "value" of the datastore entry where the key is the [public key](https://github.com/calimero-network/core/blob/8f7b16459132ed113438d51268bc65b6e0b8decb/crates/store/src/key/context.rs#L61-L62).

crates/runtime/src/logic.rs
```diff
@@ -1,5 +1,6 @@
 use std::num::NonZeroU64;
 
+use calimero_sdk::sys::RegisterId;
```
miraclx: `runtime` shouldn't depend on `sdk`

crates/runtime/src/logic.rs
```diff
@@ -80,6 +82,16 @@ impl<'a> VMLogic<'a> {
         }
         .build()
     }
+
+    pub fn get_executor_public_key(&mut self, register_id: u64) -> Result<()> {
+        self.registers
+            .set(self.limits, register_id, *&self.context.executor_public_key)
+    }
```
miraclx: this is unused, if you want to make it available to the guest, you have to move it to `VMHostFunctions` (which will then be exposed via `logic/imports.rs`)

crates/sdk/src/sys.rs
```diff
@@ -27,6 +27,8 @@ wasm_imports! {
             body: Buffer,
             register_id: RegisterId
         ) -> Bool;
+        fn get_executor_identity(register_id: RegisterId);
```
miraclx: but then this expects a different host method name entirely, btw I would also move it to the category that hosts `input` for example

crates/sdk/src/sys.rs
```diff
@@ -27,6 +27,8 @@ wasm_imports! {
...
+        fn sign_message(message: Buffer, register_id: RegisterId);
```
miraclx: as previously mentioned, we're not doing this in this iteration

crates/sdk/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub mod env;
 pub mod event;
 mod returns;
 pub mod state;
-mod sys;
+pub mod sys;
```
miraclx: please revert this change, providing a direct, unsafe interface to `sys` is not part of our objectives.

we export safe interfaces via `env`

crates/sdk/src/env.rs
```diff
@@ -31,6 +31,18 @@ fn expected_boolean<T>(e: u32) -> T {
     panic_str(&format!("Expected 0|1. Got {e}"));
 }
 
+pub fn get_executor_identity() -> String {
+    unsafe { sys::get_executor_identity(DATA_REGISTER) };
+    read_register(DATA_REGISTER)
+        .map(|data| String::from_utf8(data).expect("Invalid UTF-8"))
```
miraclx: if it can fail, return `Result` or `Option`, defaulting doesn't make sense here

or terminate the runtime if it's absence doesn't make sense (in this case, it doesn't) the execution must always have an associated identity.

returning a string also doesn't work here, it should be a `[u8; 32]` instead

crates/runtime/src/logic.rs
```diff
@@ -80,6 +82,16 @@ impl<'a> VMLogic<'a> {
...
+
+    pub fn write_to_register(&mut self, register_id: RegisterId, data: &[u8]) -> Result<()> {
+        self.registers
+            .set(self.limits, register_id.as_usize() as u64, data)
+    }
```
miraclx: this doesn't seem to have any utility, as we can read from guest memory directly, and for the purposes of this patch, we shouldn't need this anyway

crates/runtime/src/logic.rs
```diff
@@ -18,6 +19,7 @@ pub type Result<T, E = errors::VMLogicError> = std::result::Result<T, E>;
 
 pub struct VMContext {
     pub input: Vec<u8>,
+    pub executor_public_key: [u8; 32],
```
miraclx, line 21: starting to think perhaps `author` might be a better name.

crates/runtime/src/lib.rs
```diff
@@ -8,6 +8,8 @@ pub mod store;
 
 pub use constraint::Constraint;
 
+use crate::logic::{VMLogic, VMLogicError};
+
```
miraclx: unused imports

crates/runtime/Cargo.toml
```diff
@@ -18,6 +18,9 @@ ureq.workspace = true
 wasmer.workspace = true
 wasmer-types.workspace = true
 
+calimero-identity = { path = "../identity" }
+calimero-sdk = { path = "../sdk" }
+
```
miraclx: unused

crates/primitives/src/identity.rs
```diff
@@ -1,7 +1,19 @@
...
+#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ContextIdentity {
+    pub public_key: [u8; 32],
+    pub private_key: Option<[u8; 32]>,
+}
+
+#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ContextIdentities {
+    pub identities: Vec<ContextIdentity>,
+}
```
miraclx: unused

crates/node/tests/identity.rs
```diff
@@ -0,0 +1,97 @@
+use calimero_node::{Node, NodeConfig};
+use calimero_primitives::context::ContextId;
+use camino::Utf8PathBuf;
+use libp2p::identity::Keypair;
+use tokio::sync::{broadcast, oneshot};
+use {calimero_context, calimero_network, calimero_server, calimero_store};
```
miraclx: this is unnecessary, deps are made available by default

and if they weren't, `use {<crate>};` is not a replacement for `extern crate <crate>;`

crates/primitives/src/identity.rs
```diff
@@ -1,7 +1,19 @@
+use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 
 use crate::context::ContextId;
 
+#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ContextIdentity {
+    pub public_key: [u8; 32],
```
miraclx: you can define a new type here to represent the public key, I suggest following the pattern used for `ContextId`

crates/node/tests/identity.rs
```diff
@@ -0,0 +1,97 @@
...
+
+fn create_test_config() -> NodeConfig {
+    let identity = Keypair::generate_ed25519();
+    NodeConfig {
+        home: Utf8PathBuf::from("/tmp"),
```
miraclx: this would fail on macOS and Windows

crates/node/src/lib.rs
```diff
@@ -1042,15 +1082,24 @@ impl Node {
             },
         )?;
 
+        let identity_key =
+            calimero_store::key::ContextIdentity::new(context.id, transaction.executor_public_key);
+        let identity = calimero_primitives::identity::ContextIdentity {
+            public_key: transaction.executor_public_key,
+            private_key: None, // We don't have the private key here
+        };
+        handle.put(&identity_key, &identity)?;
```
miraclx: this isn't meant to happen here, it should happen at `add_context` and `join_context`.

crates/node/src/lib.rs
```diff
@@ -143,11 +143,25 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
+                            bs58::decode(executor_key)
+                                .into_vec()
+                                .map_err(|_| eyre::eyre!("Invalid executor public key"))?
+                                .try_into()
+                                .map_err(|_| eyre::eyre!("Executor public key must be 32 bytes"))?
+                        } else {
+                            // Use a default key or generate a temporary one
+                            // TODO: This is just a placeholder, it needs appropriate logic
+                            [0u8; 32]
+                        };
```
miraclx: it's bad to be empty, I'd also update the usage message below to indicate the appropriate parameters

crates/node/src/lib.rs
```diff
@@ -1,9 +1,10 @@
+use bs58;
```
miraclx: unnecessary

crates/node/src/lib.rs
```diff
@@ -143,11 +143,25 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
                             return Ok(());
                         };
 
+                        // Parse the executor's public key if provided, or use a default
+                        let executor_public_key = if !executor_key.is_empty() {
+                            bs58::decode(executor_key)
+                                .into_vec()
+                                .map_err(|_| eyre::eyre!("Invalid executor public key"))?
+                                .try_into()
+                                .map_err(|_| eyre::eyre!("Executor public key must be 32 bytes"))?
```
miraclx, line 152: you can `decode(..).onto(&mut key)` where `key: [u8; 32]`

crates/node/src/catchup.rs
```diff
@@ -9,6 +9,28 @@ use crate::{types, Node};
...
+            handle.get(&identity_key)?.ok_or_else(|| {
+                eyre::eyre!(
+                    "ContextIdentity not found for context {} and public key {:?}",
+                    context_id,
+                    executor_public_key
+                )
+            })?;
+
+        Ok(identity.public_key)
+    }
```
miraclx: I'm not sure I understand this method

Doesn't it just return the `executor_public_key` that we already have in the best case (passed in args)? and fail otherwise? (which will happen during initial catchup?)

crates/node/Cargo.toml
```diff
@@ -7,6 +7,7 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
+borsh = { workspace = true, features = ["derive"] }
```
miraclx: unused?

crates/network/Cargo.toml
```diff
@@ -37,6 +37,7 @@ tokio.workspace = true
 tokio-util = { workspace = true, features = ["codec", "compat"] }
 tracing.workspace = true
 
+calimero-identity = { path = "../identity" }
```
miraclx: unused?

crates/meroctl/Cargo.toml
```diff
@@ -30,6 +30,7 @@ tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 calimero-context = { path = "../context" }
+calimero-identity = { path = "../identity" }
```
miraclx, line 33: unused?

crates/store/src/types/context.rs
danwilliams: I've restored this - thanks for the explanation, that makes sense 🙂

crates/runtime/src/logic.rs
danwilliams: Good point - this actually has disappeared now with the removal of `write_to_register()`.

crates/runtime/src/logic.rs
danwilliams: I've removed this as requested 👍

crates/sdk/src/sys.rs
danwilliams: I might need some clarification here, as this is already in the same place as `input`?

crates/sdk/src/sys.rs
danwilliams: Good spot - I had overlooked this. Removed 👍

crates/sdk/src/lib.rs
danwilliams: Thanks for the info - this has actually gone now due to the removal of the `write_to_register()` function.

crates/sdk/src/env.rs
danwilliams: I've adjusted the return type - originally I had implemented the identity using `String`, as it wasn't clear until later that it would be a public key, and when I updated that I missed this location. Good spot! 👍

crates/runtime/src/logic.rs
danwilliams: That's fair - removed.

crates/runtime/src/logic.rs
danwilliams: We discussed this and agreed that, for now, `executor_public_key` matches the usage elsewhere, so I have left this as-is, as `author` is not used anywhere.

crates/node/tests/identity.rs
danwilliams: I'll update tests once I know the source code is all correct 🙂

crates/primitives/src/identity.rs
danwilliams: I've introduced `KeyPair` and `PublicKey` to be more appropriate here, with `ContextIdentity` now representing semantic usage elsewhere.

crates/node/tests/identity.rs
danwilliams: I'll update tests once I know the source code is all correct 🙂

crates/node/src/lib.rs
danwilliams: I've updated these, but note, although I initially added the context identity to `join_context()`, I removed it again as it felt wrong.

crates/node/src/lib.rs
danwilliams: Updated to return an error, and thanks for pointing out the usage instructions! I hadn't seen those. Updated 👍

crates/node/src/catchup.rs
danwilliams: This evolved over time and became essentially useless, so I've removed it 👍

crates/sdk/src/sys.rs
miraclx: See the demarcated sections `// --`

crates/node/src/catchup.rs
```diff
@@ -417,9 +373,9 @@ impl Node {
...
                         };
 
-                        // We don't have the private key during catchup
-                        let initial_identity = calimero_primitives::identity::ContextIdentity {
-                            public_key: *context_id,
+                        // We don't have the identity during catchup
+                        let initial_identity = KeyPair {
+                            public_key: PublicKey(*context_id),
                             private_key: None,
                         };
```
miraclx: Reusing the context ID as the user's public key is not legal. At this point, the initial identity should already have been defined via `ContextManager::join_context`.

We should also be sharing our public key with the node we're performing a catchup with.

crates/primitives/src/identity.rs
```diff
@@ -3,16 +3,16 @@ use serde::{Deserialize, Serialize};
...
     pub private_key: Option<[u8; 32]>,
 }
 
 #[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
-pub struct ContextIdentities {
-    pub identities: Vec<ContextIdentity>,
-}
+// This could use a Hash, but we need to be able to serialize the PublicKey and
+// create::hash::Hash does not currently implement Borsh.
+pub struct PublicKey(pub [u8; 32]);
```
miraclx, line 16: please make this:

```rust
pub struct PublicKey(Hash);
```

on which you can then impl `Deref<Target = [u8; 32]>`, `From<[u8; 32]>`, `Display`, `Into<String>`, `FromStr`

the same way [`ContextId` does](https://github.com/calimero-network/core/blob/3f20778271969ba0799e666eb5c523616ab7b776/crates/primitives/src/context.rs#L11-L64).

> but we need to be able to serialize the PublicKey

why's that?

crates/runtime/src/logic.rs
miraclx, line 1: we're still missing the `get_executor_identity` host function here

crates/sdk/src/env.rs
```diff
@@ -31,16 +31,12 @@ fn expected_boolean<T>(e: u32) -> T {
...
     read_register(DATA_REGISTER)
-        .map(|data| String::from_utf8(data).expect("Invalid UTF-8"))
-        .unwrap_or_default()
-}
-
-pub fn sign_message(message: &[u8]) -> Vec<u8> {
-    unsafe { sys::sign_message(sys::Buffer::from(message), DATA_REGISTER) };
-    read_register(DATA_REGISTER).unwrap_or_default()
+        .expect("Must have executor identity.")
+        .try_into()
```
miraclx: unnecessary heap allocation here, I'd impl a special `unsafe` `read_register` method generic over some const length.

```rust
unsafe fn read_register_sized<const N: usize>(register_id: sys::RegisterId) -> Option<[u8; N]> {
    let mut buf = [0; N];
    let len = register_len(register_id)?;
    if N != len {
        panic_str("...");
    }
    let succeed = sys::read_register(register_id, sys::BufferMut::new(&mut buf))
        .try_into()
        .unwrap_or_else(expected_boolean);

    if !succeed {
        panic_str("fatal error, length guard didn't hold");
    }

    Some(buf)
}
```

crates/sdk/src/env.rs
```diff
@@ -31,16 +31,12 @@ fn expected_boolean<T>(e: u32) -> T {
     panic_str(&format!("Expected 0|1. Got {e}"));
 }
 
-pub fn get_executor_identity() -> String {
+pub fn get_executor_identity() -> [u8; 32] {
```
miraclx, line 34: also, nit: ordering

crates/server/src/admin/handlers/context.rs
```diff
@@ -174,8 +175,8 @@ pub async fn create_context_handler(
         last_transaction_hash: Default::default(),
     };
 
-    let initial_identity = calimero_primitives::identity::ContextIdentity {
-        public_key: *context_id.as_bytes(),
+    let initial_identity = KeyPair {
+        public_key: PublicKey(*context_id.as_bytes()),
         private_key: Some(*signing_key.as_bytes()),
```
miraclx: the context ID isn't the public key for the member who created it

you need another key pair here

crates/node/src/lib.rs
```diff
@@ -143,17 +143,15 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                             bs58::decode(executor_key)
                                 .into_vec()
                                 .map_err(|_| eyre::eyre!("Invalid executor public key"))?
                                 .try_into()
                                 .map_err(|_| eyre::eyre!("Executor public key must be 32 bytes"))?
                         } else {
-                            // Use a default key or generate a temporary one
-                            // TODO: This is just a placeholder, it needs appropriate logic
-                            [0u8; 32]
+                            return Err(eyre::eyre!("Executor public key is required"));
```
miraclx, line 154: this will kill the node with this error, look at similar patterns:

> print the error, `return Ok(());`

crates/node/src/lib.rs
```diff
@@ -242,7 +240,9 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
                     }
                 }
             } else {
-                println!("{IND} Usage: call <Method> <JSON Payload>");
+                println!(
+                    "{IND} Usage: call <Context ID> <Method> <JSON Payload> [Executor Public Key]"
```
miraclx: not exactly optional

```suggestion
                    "{IND} Usage: call <Context ID> <Method> <JSON Payload> <Executor Public Key>"
```

crates/node/src/lib.rs
```diff
@@ -439,11 +439,12 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                         };
 
+                        // We don't have the identity at this point
                         node.ctx_manager
                             .add_context(
                                 context,
-                                calimero_primitives::identity::ContextIdentity {
-                                    public_key: *context_id,
+                                KeyPair {
+                                    public_key: PublicKey(*context_id),
```
miraclx: the same way the user of the interactive CLI provides a context ID, they can provide a private key for their executor (from which we can extrapolate a `KeyPair`)

crates/store/src/types/context.rs
```diff
@@ -54,6 +54,12 @@ impl PredefinedEntry for key::ContextState {
     type DataType<'a> = ContextState<'a>;
 }
 
+#[derive(Eq, Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub struct ContextIdentity {
+    pub public_key: PublicKey,
```
miraclx, line 59: this is already in the key used to access this entry, `public_key` is not needed here

crates/store/src/key/context.rs
```diff
@@ -62,9 +62,12 @@ impl KeyComponent for PublicKey {
...
 
 impl ContextIdentity {
-    pub fn new(context_id: calimero_primitives::context::ContextId, context_pk: [u8; 32]) -> Self {
+    pub fn new(
+        context_id: calimero_primitives::context::ContextId,
+        context_pk: calimero_primitives::identity::PublicKey,
+    ) -> Self {
         Self(Key(
-            GenericArray::from(*context_id).concat(context_pk.into())
+            GenericArray::from(*context_id).concat(GenericArray::from(context_pk.0))
```
miraclx, line 70: the contained entry should not be `pub`, as it should be a read-only view to a public key, using the interface equally defined for `ContextId`, this should become:

```suggestion
            GenericArray::from(*context_id).concat(GenericArray::from(*context_pk))
```

crates/sdk/src/env.rs
miraclx: On second thought, you can take out the `N != len` guard check, since the host will do it anyway 🤷🏽‍♂️ 

Though it's worth noting that `read_register` will panic if the register doesn't exist, so while the length check is unnecessary, you can use it to check if the register exists.

```rust
fn read_register_sized<const N: usize>(register_id: sys::RegisterId) -> Option<[u8; N]> {
    let len = register_len(register_id)?;

    let mut buf = [0; N];

    let succeed: bool = unsafe {
        sys::read_register(register_id, sys::BufferMut::new(&mut buf))
            .try_into()
            .unwrap_or_else(expected_boolean)
    };

    if !succeed {
        panic_str(&format!("register content length ({}) does not match buffer length ({})", len, N));
    }

    Some(buf)
}
```

crates/sdk/src/sys.rs
danwilliams: Ahhh - it is totally non-obvious what those mean. You might want to consider turning them into actual comment headings, as right now they are a bit weird 🙂

crates/node/src/catchup.rs
danwilliams: I've attempted to apply these changes but there were quite a few of them that resulted, and I am not 100% confident about them. Particularly, they resulted in me having to add handling for `types::CatchupStreamMessage::Request` to `handle_catchup_message()`, based partly on adding the public key to `CatchupRequest`. Honestly, I'm guessing if that is correct, as it hasn't been mentioned, but it seems like the required outcome.

As mentioned previously, it feels quite wrong to me to be adding the key into the join context, when we already have it via create context - I feel like it should be one or other - I don't know? Hopefully this is correct now.

crates/primitives/src/identity.rs
danwilliams: As mentioned, I cannot change this to be based around a `Hash`, as I described - i.e. it needs to be serialised.

The reason it needs to be serialised is that `ContextIdentity` uses it, and implements `BorshSerialize` and `BorshDeserialize`. Remember, at present the public key is stored in the database as part of this struct. In another piece of feedback below, you have suggested removing it from here - but that is not currently feasible (see other reply for that).

If there is a valid way to remove it from `ContextIdentity` without breaking everything for `KeyPair` conversion, or if you have another idea, let me know... I could implement the (de)serialisation functions manually, but that seems wrong here, or also I could add the derives to `Hash`, but that also seems wrong.

crates/runtime/src/logic.rs
danwilliams: We did actually have it before, but you asked me to remove it. It was called `get_executor_public_key()`... some of the naming of things changed along the way, so maybe you meant for me to rename it instead? 🙂 I have brought it back as `get_executor_identity()`.

crates/sdk/src/env.rs
danwilliams: That's a good shout, and I have added that in 🙂 The only thing I am not keen on is the panic, but we have one in `read_register()` anyway, so meh.

crates/sdk/src/env.rs
danwilliams: I'm going to need more of a steer here. I default to using alphabetical order where possible, with a couple of exceptions such as putting things like constructors and id fields first. But in most places of the codebase we don't see to have any particular order that I can determine, and here I have added it in the most alphabetically-ordered place possible. What is the secret order? 😅

crates/server/src/admin/handlers/context.rs
danwilliams: That's fair. Based on what you've said so far, I've added generation of a key here... but I am not 100% sure that this is correct. If `create_context_handler()` is meant to create the actual keypair then great, done - but if it's actually meant to accept one in, then I will need to add it to `CreateContextRequest` instead. As you've not mentioned that at all, or said anything about changing the interface here, I've erred on the side of caution, but I do feel uncomfortable about it. Let me know!

crates/node/src/lib.rs
danwilliams: This one has confused me, as you specifically said to kill the runtime if it's not supplied. So... how does returning `Ok` kill the runtime?

As killing makes more sense to me here than just making it okay, seeing as we cannot really continue, I have left this in place based on the original request. If you do want it to be okay and continue, can you explain how we can do that without a valid key?

Note, I did poke around here a bit but I could not work out any approach that I liked, and that seemed to fit, that would enable us to continue under these circumstances.

crates/node/src/lib.rs
danwilliams: That makes a lot of sense. Thanks - that clears up where to get this from 🙂 

I have added the parameter along with extraction of the key and some `From` conversions for it.

crates/store/src/types/context.rs
danwilliams: That's fair, but I was under the impression that you had said for it to be there, and to convert to and from the other type(s). We ended up changing the `ContextIdentity` primitive to `KeyPair`, and as we still have that type used then there's no way to convert the `KeyPair` to and from the `ContextIdentity` without losing data if we remove this field. I did actually remove it and spend some time carefully exploring what the repercussions would be like, but I did not feel comfortable as they deviate too far from what we've previously discussed. Therefore I've kept this field in for now - if you 100% want it removed then please advise what you want done with the `KeyPair`, the interaction between the two types, and the wider usage.

crates/store/src/key/context.rs
danwilliams: No worries! I was trying to keep with the existing interface, as every single other struct in that file exposes its fields as public. But I have no problem with the change, in principle. The trouble is, this cannot be dereferenced here in the manner suggested, so is there an alternative I'm missing...?

crates/runtime/src/logic.rs
miraclx: > but you asked me to remove it

Hmm, I don't think I did.

I said it wasn't exported correctly:

https://github.com/calimero-network/core/pull/536#discussion_r1694854727

And that the host defines a different symbol than what the guest expects

https://github.com/calimero-network/core/pull/536#discussion_r1694855133

> so maybe you meant for me to rename it instead? 🙂 I have brought it back as `get_executor_identity()`.

👍🏽

crates/primitives/src/identity.rs
miraclx: > `ContextIdentity` uses it

if you mean `calimero_store::types::ContextIdentity`, again, the `PublicKey` makes up the `key` used to access the `ContextIdentity` so it shouldn't need to contain the public key anymore.

crates/store/src/types/context.rs
miraclx: > but I was under the impression that you had said for it to be there, and to convert to and from the other type(s).

Not at all, I don't foresee any use for `KeyPair` aside as an input to `create_context` and `join_context` (which could just be parameters at that point).

But if you had to keep a `KeyPair` in, then it's "conversion to the store entry"  would be:

```rust
let mut handle = store.handle();

handle.put(
    &calimero_store::key::ContextIdentity::new(context_id, key_pair.public_key),
    &calimero_store::types::ContextIdentity {
        private_key: keypair.private_key
    }
);
```

it doesn't have to warrant a conversion to an identical structure for storage

crates/store/src/key/context.rs
miraclx: > this cannot be dereferenced here in the manner suggested, so is there an alternative I'm missing

following the clarifications made above, matching the interface for `ContextId` will make this possible

crates/node/src/lib.rs
miraclx: if you follow the logic, you'll see that the caller of this method `handle_line`:

https://github.com/calimero-network/core/blob/9abbec2d7c57bc19b537260f6c48b10b8aee0b1b/crates/node/src/lib.rs#L103

short-circuits and exits the `loop-select` of the `start` method which keeps the node running

so this should only be done for fatal, unrecoverable errors

in this case, failure to parse a public key isn't unrecoverable

returning `Ok(())` from `handle_line` will proceed to the next iteration of the loop allowing the user to fix their mistake and retry their input which is what you want

you can take advantage of that opportunity to print a guiding message to the user before their next input

For example:

https://github.com/calimero-network/core/blob/9abbec2d7c57bc19b537260f6c48b10b8aee0b1b/crates/node/src/lib.rs#L141-L144

crates/server/src/admin/handlers/context.rs
miraclx: I agree with this decision, if we decide to change it later we'll have that discussion.

crates/sdk/src/env.rs
miraclx: Just try to match how things are defined in the bridge so it's consistently mapped.

Methods are grouped by function

You should get a hint of

- panic
- register-ops
- context & runtime
- state
- external

crates/sdk/src/env.rs
miraclx: We'll eventually communicate the potential for panicking in the docs as the standard library does. I'm tempted to make this `unsafe`, but we've guarded against undefined behavior with the length check on the host, and `std` doesn't mark logically sound and potentially misused methods unsafe, so it's fine.

crates/runtime/src/logic.rs
```diff
@@ -81,6 +81,11 @@ impl<'a> VMLogic<'a> {
         }
         .build()
     }
+
+    pub fn get_executor_identity(&mut self, register_id: u64) -> Result<()> {
+        self.registers
+            .set(self.limits, register_id, self.context.executor_public_key)
+    }
```
miraclx, line 88: Again, as [mentioned](https://github.com/calimero-network/core/pull/536#discussion_r1694854727)

CleanShot 2024-08-01 at 18 25 51@2x

Like the other host functions, it should be part of `VMHostFunctions` and exported via `logic/imports.rs`

To match the guest's host function expectations.

Otherwise, in this form, it's simply unused.

crates/node/src/catchup.rs
```diff
@@ -213,6 +227,9 @@ impl Node {
                     batch_size: self.network_client.catchup_config.batch_size,
                     url: None,
                     hash: None,
+                    // In this situation, where we have no context, we have no way to know the
+                    // public key.
+                    public_key: PublicKey([0; 32]),
```
miraclx, line 232: `ContextManager::join_context` should take an identity for the member who wants to join, so at this point we should have one.

crates/node/src/catchup.rs
```diff
@@ -263,6 +280,29 @@ impl Node {
...
+                        id: request.context_id,
+                        application_id: request.application_id.ok_or_else(|| {
+                            eyre::eyre!("Application ID missing in catchup request")
+                        })?,
+                        last_transaction_hash: request.last_executed_transaction_hash,
+                    });
+                }
+
+                // Note: Might need to prepare and send an initial response?
+            }
```
miraclx, line 305: we can skip doing this if the initial catchup request includes all the identities on our node

crates/node/src/catchup.rs
```diff
@@ -193,17 +193,31 @@ impl Node {
...
+                let identity_key =
+                    calimero_store::key::ContextIdentity::new(context_id, PublicKey([0; 32]));
+                let initial_identity: KeyPair = self
+                    .store
+                    .handle()
+                    .get(&identity_key)?
+                    .ok_or_else(|| {
+                        eyre::eyre!("Initial identity not found for context {}", context_id)
+                    })?
+                    .into();
```
miraclx, line 206:
```rust
let start = calimero_store::key::ContextIdentity::new(context_id, PublicKey::from([0; 32]));
let identities = handle.iter(&start).keys().map(|key| PublicKey::from(key.public_key())).collect();
```

to share all the identities

crates/node/src/lib.rs
```diff
@@ -376,7 +382,26 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                         };
 
-                        node.ctx_manager.join_context(&context_id).await?;
+                        // Parse the private key
+                        let private_key = bs58::decode(private_key)
+                            .into_vec()
+                            .map_err(|_| eyre::eyre!("Invalid private key"))?;
+                        let private_key: [u8; 32] = private_key
+                            .try_into()
+                            .map_err(|_| eyre::eyre!("Private key must be 32 bytes"))?;
```
miraclx, line 389: https://github.com/calimero-network/core/pull/536#discussion_r1694891197

> you can `decode(..).onto(&mut key)` where `key: [u8; 32]`

crates/node/src/lib.rs
```diff
@@ -426,6 +452,23 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
 
                         let application_id = application_id.to_owned().into();
 
+                        // Parse the private key
+                        let private_key = bs58::decode(private_key)
+                            .into_vec()
+                            .map_err(|_| eyre::eyre!("Invalid private key"))?;
+                        let private_key: [u8; 32] = private_key
+                            .try_into()
+                            .map_err(|_| eyre::eyre!("Private key must be 32 bytes"))?;
```
miraclx, line 459: https://github.com/calimero-network/core/pull/536#discussion_r1694891197

> you can `decode(..).onto(&mut key)` where `key: [u8; 32]`

crates/node/src/catchup.rs
```diff
@@ -373,11 +413,29 @@ impl Node {
...
+
+                        // Share our public key with the catchup node
+                        self.network_client
+                            .publish(
+                                libp2p::gossipsub::TopicHash::from_raw(context_id),
+                                serde_json::to_vec(&types::PeerAction::SharePublicKey(
+                                    initial_identity.public_key.clone(),
+                                ))?,
+                            )
+                            .await?;
```
miraclx, line 438: nit here: this will announce this across the network and not just the catch up node

crates/sdk/src/env.rs
```diff
@@ -33,10 +33,7 @@ fn expected_boolean<T>(e: u32) -> T {
 
 pub fn get_executor_identity() -> [u8; 32] {
     unsafe { sys::get_executor_identity(DATA_REGISTER) };
-    read_register(DATA_REGISTER)
-        .expect("Must have executor identity.")
-        .try_into()
-        .expect("Wrong executor identity length.")
+    read_register_sized(DATA_REGISTER).expect("Must have executor identity.")
```
miraclx, line 36: one thing you'll notice is we don't use the traditional panic because panics in WASM will abort the runtime with no information about what happened

so you have to call `panic_str` which will first send a contextual message to the host, after which the host terminates the execution

consider `unwrap_or_else`

xilosada (approved):

MatejVukosav (approved):
Lets merge this and address open comments next week: https://github.com/calimero-network/core/issues/545

crates/context/src/lib.rs
```diff
@@ -122,6 +129,7 @@ impl ContextManager {
     pub async fn join_context(
         &self,
         context_id: &calimero_primitives::context::ContextId,
+        initial_identity: KeyPair,
```
fbozic, line 132: Can we have this as `calimero_primitives::identity::KeyPair`? We tend to use fully qualified names for primitives

danwilliams: I truly do not understand why fully-qualified names are used everywhere, but I was told it was to avoid ambiguity 😅 

Regardless of my feelings about the practice, there is no ambiguity here as there is only one type with that name...?

fbozic: That is the convention @miraclx proposed a long time ago and we are trying to stick with it. To be honest, I have no strong feelings about either approach, but breaking consistency like this is the worst possible choice IMO. If you have strong feelings about this, please bring that topic up for discussion. If we decide to change convention, that is also fine but we are gonna do it everywhere.

PR #538, opened 2024-08-01
fix: json rpc subscriptions

MatejVukosav:
Test:

I have an [app](https://github.com/calimero-network/core-app-template/pull/6) that receives new value after mutation with button. 

https://github.com/user-attachments/assets/64f9f9df-7b3e-4a57-b876-81b19fcf2574

### DIFF
--- a/packages/calimero-sdk/src/subscriptions/subscriptions.ts
+++ b/packages/calimero-sdk/src/subscriptions/subscriptions.ts
@@ -1,10 +1,10 @@
 import { ContextId } from '../types/context';
 
 export interface SubscriptionsClient {
-  connect(connectionId: string): void;
-  disconnect(connectionId: string): void;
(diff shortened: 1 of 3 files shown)


packages/calimero-sdk/src/subscriptions/subscriptions.ts
```diff
@@ -1,10 +1,10 @@
 import { ContextId } from '../types/context';
 
 export interface SubscriptionsClient {
-  connect(connectionId: string): void;
-  disconnect(connectionId: string): void;
-  subscribe(applicationIds: string[], connectionId?: string): void;
-  unsubscribe(applicationIds: string[], connectionId?: string): void;
+  connect(connectionId?: string): Promise<void>;
```
MatejVukosav, line 4: Main issue was that subscription didn't return promise so app didn't wait for connection

PR #543, opened 2024-08-05
fix: Clippy errors

danwilliams:
We had over 130 Clippy warnings, plus some non-Clippy build warnings. I've fixed up all of these, so hopefully we can keep Clippy happy now! 🙂 

**NOTE: I have found tests failing on master, due to test files not found, but I have not attempted to fix these.** Just noting here that these failures were not introduced by this branch.

### Fixed build warnings

  - Updated Near crates to latest version.

  - Changed `UnorderedMap` to `IterableMap`, to deal with the deprecation warning.

  - Removed unused imports.

  - Updated `Cargo.lock`.

### Fixed Clippy warnings

Notable changes:

  - Corrected `Into<String> for ApplicationId` to be `From<ApplicationId> for String`.

  - Renamed `Hash::hash()` to `Hash::new()` to better-describe its purpose.

  - Boxed the `Stream` type.

Plus numerous miscellaneous fixes not worth mentioning individually.


chefsale:
Can we add clippy check to CI?

MatejVukosav:
https://github.com/calimero-network/core/issues/544

crates/network/src/events/identify.rs
```diff
@@ -8,29 +8,26 @@ impl EventHandler<identify::Event> for EventLoop {
     async fn handle(&mut self, event: identify::Event) {
         debug!("{}: {:?}", "identify".yellow(), event);
 
-        match event {
```
fbozic, line 11: What do you think about keeping a `match` arm here? So we have a consistent code base for event handling...

crates/network/src/events/mdns.rs
```diff
@@ -9,28 +9,25 @@ impl EventHandler<mdns::Event> for EventLoop {
     async fn handle(&mut self, event: mdns::Event) {
         debug!("{}: {:?}", "mdns".yellow(), event);
 
-        match event {
```
fbozic, line 12: ... and here

crates/network/src/stream.rs
```diff
@@ -66,7 +66,7 @@ impl EventLoop {
         self.event_sender
             .send(types::NetworkEvent::StreamOpened {
                 peer_id: peer,
-                stream: Stream::new(stream),
+                stream: Box::new(Stream::new(stream)),
```
fbozic, line 69: ~noob question: Why is this required? I thought `rustc` would complain about those kinds of issues.

crates/network/src/events/identify.rs
danwilliams: The issue here is simply that we were only checking one match arm 🙂 If we check against more later, we can revert it to a `match`.

crates/network/src/events/mdns.rs
danwilliams: Same 🙂

crates/network/src/stream.rs
danwilliams: It is not required, but the main reason is that the `Stream` is potentially large, and best boxed, especially as it creates a large size discrepancy in variants.

PR #561, opened 2024-08-08
fix(admin-api): link installed application with package manager

frdomovic:
# fix(admin-api): link installed application with package manager

## Summary
The issue with previous approach is that it is missing a link between application metadata that is saved in the package-manager smart contract and this is needed for properly displaying and identifying applications from the admin dashboard.

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -431,8 +431,7 @@ impl ContextManager {
             blob: application.blob.blob_id(),
             version: application.version.as_deref().map(str::parse).transpose()?,
             source: application.source.parse()?,
-            metadata: application
-                .metadata.to_vec()
+            metadata: application.metadata.to_vec(),
         }))
     }
 
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -334,6 +334,7 @@ impl ContextManager {
         &self,
         path: Utf8PathBuf,
         version: Option<semver::Version>,
+        metadata: Vec<u8>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
         let file = fs::File::open(&path).await?;
 
@@ -346,7 +347,7 @@ impl ContextManager {
             eyre::bail!("non-absolute path")
         };
 
-        self.install_application(blob_id, uri.as_str().parse()?, version, Vec::new())
+        self.install_application(blob_id, uri.as_str().parse()?, version, metadata)
     }
 
     pub async fn install_application_from_url(
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -313,7 +313,7 @@ impl ContextManager {
             blob: calimero_store::key::BlobMeta::new(blob_id),
             version: version.map(|v| v.to_string().into_boxed_str()),
             source: source.to_string().into_boxed_str(),
-            metadata: metadata.map(|s| s.into_boxed_str()),
+            metadata: metadata.map(|s| s.into_bytes().into_boxed_slice()),
         };
 
         let application_id = calimero_primitives::application::ApplicationId::from(
@@ -379,13 +379,12 @@ impl ContextManager {
 
         for (id, app) in iter.entries() {
             let (id, app) = (id?, app?);
-
             applications.push(calimero_primitives::application::Application {
                 id: id.application_id(),
                 blob: app.blob.blob_id(),
                 version: app.version.as_deref().map(str::parse).transpose()?,
                 source: app.source.parse()?,
-                metadata: app.metadata.as_deref().map(str::parse).transpose()?,
+                metadata: app.metadata.map(|m| m.into_vec()),
             })
         }
 
@@ -433,10 +432,7 @@ impl ContextManager {
             version: application.version.as_deref().map(str::parse).transpose()?,
             source: application.source.parse()?,
             metadata: application
-                .metadata
-                .as_deref()
-                .map(str::parse)
-                .transpose()?,
+                .metadata.map(|m| m.into_vec())
         }))
     }
 
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -307,11 +307,18 @@ impl ContextManager {
         blob_id: calimero_primitives::blobs::BlobId,
         source: calimero_primitives::application::ApplicationSource,
         version: Option<semver::Version>,
+        contract_app_id: Option<String>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
+        let boxed_id: Box<str> = match contract_app_id {
+            Some(s) => s.into_boxed_str(),
+            None => Box::from(""),
+        };
+
         let application = calimero_store::types::ApplicationMeta {
             blob: calimero_store::key::BlobMeta::new(blob_id),
             version: version.map(|v| v.to_string().into_boxed_str()),
             source: source.to_string().into_boxed_str(),
+            contract_app_id: Some(boxed_id),
         };
 
         let application_id = calimero_primitives::application::ApplicationId::from(
@@ -344,13 +351,14 @@ impl ContextManager {
             eyre::bail!("non-absolute path")
         };
 
-        self.install_application(blob_id, uri.as_str().parse()?, version)
+        self.install_application(blob_id, uri.as_str().parse()?, version, None)
     }
 
     pub async fn install_application_from_url(
         &self,
         url: Url,
         version: Option<semver::Version>,
+        contract_app_id: Option<String>,
         // hash: calimero_primitives::hash::Hash,
         // todo! BlobMgr should return hash of content
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
@@ -362,7 +370,7 @@ impl ContextManager {
 
         // todo! if blob hash doesn't match, remove it
 
-        self.install_application(blob_id, uri, version)
+        self.install_application(blob_id, uri, version, Some(contract_app_id).unwrap())
     }
 
     pub fn list_installed_applications(
@@ -382,6 +390,7 @@ impl ContextManager {
                 blob: app.blob.blob_id(),
                 version: app.version.as_deref().map(str::parse).transpose()?,
                 source: app.source.parse()?,
+                contract_app_id: app.contract_app_id.as_deref().map(str::parse).transpose()?,
             })
         }
 
@@ -428,6 +437,7 @@ impl ContextManager {
             blob: application.blob.blob_id(),
             version: application.version.as_deref().map(str::parse).transpose()?,
             source: application.source.parse()?,
+            contract_app_id: application.contract_app_id.as_deref().map(str::parse).transpose()?,
         }))
     }
 
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -307,18 +307,13 @@ impl ContextManager {
         blob_id: calimero_primitives::blobs::BlobId,
         source: calimero_primitives::application::ApplicationSource,
         version: Option<semver::Version>,
-        contract_app_id: Option<String>,
+        metadata: Option<String>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
-        let contract_app_id: Box<str> = match contract_app_id {
-            Some(s) => s.into_boxed_str(),
-            None => Box::from(""),
-        };
-
         let application = calimero_store::types::ApplicationMeta {
             blob: calimero_store::key::BlobMeta::new(blob_id),
             version: version.map(|v| v.to_string().into_boxed_str()),
             source: source.to_string().into_boxed_str(),
-            contract_app_id: Some(contract_app_id),
+            metadata: metadata.map(|s| s.into_boxed_str()),
         };
 
         let application_id = calimero_primitives::application::ApplicationId::from(
@@ -358,7 +353,7 @@ impl ContextManager {
         &self,
         url: Url,
         version: Option<semver::Version>,
-        contract_app_id: Option<String>,
+        metadata: Option<String>,
         // hash: calimero_primitives::hash::Hash,
         // todo! BlobMgr should return hash of content
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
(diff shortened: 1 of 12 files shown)


crates/context/src/lib.rs
```diff
@@ -307,11 +307,18 @@ impl ContextManager {
         blob_id: calimero_primitives::blobs::BlobId,
         source: calimero_primitives::application::ApplicationSource,
         version: Option<semver::Version>,
+        contract_app_id: Option<String>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
+        let boxed_id: Box<str> = match contract_app_id {
```
MatejVukosav: what is boxed_id? please use name by meaning, not by type

crates/context/src/lib.rs
```diff
@@ -307,11 +307,18 @@ impl ContextManager {
...
+            Some(s) => s.into_boxed_str(),
+            None => Box::from(""),
+        };
+
         let application = calimero_store::types::ApplicationMeta {
             blob: calimero_store::key::BlobMeta::new(blob_id),
             version: version.map(|v| v.to_string().into_boxed_str()),
             source: source.to_string().into_boxed_str(),
+            contract_app_id: Some(boxed_id),
```
MatejVukosav: what is diff between app_id and contract_app_id?

frdomovic: contract-app-id is application id from the contract package manager so you can fetch the metadata by its id (hash - owner + package name). 
and app_id is hash generated in code in image  (hash - application metadata)
Screenshot 2024-08-08 at 12 09 56

crates/context/src/lib.rs
frdomovic: Updated

crates/context/src/lib.rs
```diff
@@ -307,11 +307,18 @@ impl ContextManager {
         blob_id: calimero_primitives::blobs::BlobId,
         source: calimero_primitives::application::ApplicationSource,
         version: Option<semver::Version>,
+        contract_app_id: Option<String>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
+        let contract_app_id: Box<str> = match contract_app_id {
+            Some(s) => s.into_boxed_str(),
+            None => Box::from(""),
```
MatejVukosav: When can contract_app_id be empty?

crates/context/src/lib.rs
```diff
@@ -307,11 +307,18 @@ impl ContextManager {
...
+        };
+
         let application = calimero_store::types::ApplicationMeta {
             blob: calimero_store::key::BlobMeta::new(blob_id),
             version: version.map(|v| v.to_string().into_boxed_str()),
             source: source.to_string().into_boxed_str(),
+            contract_app_id: Some(contract_app_id),
```
MatejVukosav: hm if the value is `None` and the type is `Option`, why not leave it as `None`?

crates/server-primitives/src/admin.rs
```diff
@@ -5,9 +5,10 @@ use serde_json::Value;
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct InstallApplicationRequest {
+    pub contract_app_id: String,
     pub url: url::Url,
     pub version: Option<semver::Version>,
-    pub hash: calimero_primitives::hash::Hash,
+    pub hash: Option<calimero_primitives::hash::Hash>,
```
MatejVukosav, line 10: What is a scenario of hash missing?

crates/context/src/lib.rs
MatejVukosav: Can you please add the field description to the code as well so we have it documented?

crates/context/src/lib.rs
frdomovic: When installing local application which has no metadata saved in the smart contract package manager

crates/context/src/lib.rs
frdomovic: ApplicationMeta is used somewhere else where contract_app_id can be None, in this case its always Some as contract_app_id is transformed into Box<str> either app_id value or "".

crates/server-primitives/src/admin.rs
frdomovic: Checking if the downloaded application hash is the same as the one saved in the contract. As the source can be corrupt

crates/context/src/lib.rs
frdomovic: Added description

crates/context/src/lib.rs
MatejVukosav: Do you need to transform it to Box or can you leave it as `None`? That is my concern

crates/server-primitives/src/admin.rs
MatejVukosav: Shouldn't it then make an application invalid? And not just the hash part

frdomovic: So previous implementation that I created downloaded the application wasm, check the hash and if they don't match delete the application and return message why it failed.
Can you confirm the updated logic cc @miraclx

crates/context/src/lib.rs
frdomovic: I followed what was already implemented there and using Box probably has its reason. cc @miraclx Could you please answer this ?

crates/primitives/src/application.rs
```diff
@@ -94,7 +94,9 @@ impl fmt::Display for ApplicationSource {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Application {
+    // contract_app_id - Application ID is hash(owner,name) saved in package manager contract
     pub contract_app_id: Option<String>,
+    // id - Applicaiton ID created in the node for identification - see line 324 crates/context/src/lib.rs
```
frdomovic:
```suggestion
    // id - Applicaton ID created in the node for identification - see line 324 crates/context/src/lib.rs
```

crates/store/src/types/application.rs
```diff
@@ -11,6 +11,8 @@ pub struct ApplicationMeta {
     pub blob: key::BlobMeta,
     pub version: Option<Box<str>>,
     pub source: Box<str>,
+    // contract_app_id - Application ID is hash(owner,name) saved in package manager contract
+    pub contract_app_id: Option<Box<str>>,
```
miraclx: We don't want any lock-in to the registry. No allusion to dependence either.

You can use a generic entry called `metadata` as an arbitrary byte slice - `Box<[u8]>`.

And leave it up to clients to visualize that however they deem fit.

crates/server/src/admin/handlers/context.rs
```diff
@@ -244,8 +244,8 @@ pub async fn get_context_storage_handler(
 
 #[derive(Deserialize)]
 pub struct JoinContextRequest {
-    pub public_key: PublicKey,
-    pub private_key: [u8; 32],
+    pub public_key: Option<PublicKey>,
+    pub private_key: Option<[u8; 32]>,
```
miraclx: I don't understand this change. I think you're right to change something here, but this isn't the change to make.

We only need `Option<PrivateKey>`, and nothing else.

crates/server-primitives/src/admin.rs
miraclx: Take the CLI for example, `meroctl app install --url <URL>`

I know where the blob is, but I don't have the hash

For me to know the hash, I would have already downloaded it. In which case, I would use `meroctl app install --path <PATH>` which also, doesn't use the hash.

The idea of the hash is (if) you have it, provide it and we'll test the blob before using it.

And if not, you're implicitly saying you trust it regardless of what we end up downloading.

crates/context/src/lib.rs
```diff
@@ -362,7 +370,7 @@ impl ContextManager {
 
         // todo! if blob hash doesn't match, remove it
 
-        self.install_application(blob_id, uri, version)
+        self.install_application(blob_id, uri, version, Some(contract_app_id).unwrap())
```
```suggestion
        self.install_application(blob_id, uri, version, contract_app_id)
```

crates/context/src/lib.rs
miraclx: As mentioned above, this will change. But aside from that, will it ever be `None`?

frdomovic: Updated

crates/store/src/types/application.rs
frdomovic: Updated to metadata name

crates/server/src/admin/handlers/context.rs
frdomovic: public key removed as we can generate it from private key

crates/context/src/lib.rs
frdomovic: updated

crates/store/src/types/application.rs
frdomovic: updated to u8

crates/context/src/lib.rs
```diff
@@ -307,18 +307,13 @@ impl ContextManager {
         blob_id: calimero_primitives::blobs::BlobId,
         source: calimero_primitives::application::ApplicationSource,
         version: Option<semver::Version>,
-        contract_app_id: Option<String>,
+        metadata: Option<String>,
```
miraclx: how about we take a `Vec<u8>` here instead of assuming a `String` (which is required to be utf8-encoded)?

also, an empty array can indicate absence of metadata.

`Vec::new` does not allocate, so there's no associated heap cost

crates/meroctl/src/cli/app/install.rs
```diff
@@ -30,7 +30,6 @@ impl InstallCommand {
         let Ok(config) = ConfigFile::load(&path) else {
             eyre::bail!("Failed to load config file")
         };
-
```
miraclx: restore whitespace

crates/primitives/src/application.rs
```diff
@@ -94,8 +94,8 @@ impl fmt::Display for ApplicationSource {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Application {
-    // contract_app_id - Application ID is hash(owner,name) saved in package manager contract
-    pub contract_app_id: Option<String>,
+    // Metadata - Contains application ID is hash(owner,name) saved in package manager contract
+    pub metadata: Option<Vec<u8>>,
     // id - Application ID created in the node for identification - see line 324 crates/context/src/lib.rs
```
miraclx: lets move this to the bottom, and we no longer need the comments. especially since we're registry agnostic

crates/server-primitives/src/admin.rs
```diff
@@ -5,8 +5,8 @@ use serde_json::Value;
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct InstallApplicationRequest {
-    // contract_app_id - Application ID is hash(owner,name) saved in package manager contract
-    pub contract_app_id: String,
+    // Metadata - Contains application ID is hash(owner,name) saved in package manager contract
+    pub metadata: String,
```
miraclx: same

crates/store/src/types/application.rs
```diff
@@ -11,8 +11,8 @@ pub struct ApplicationMeta {
     pub blob: key::BlobMeta,
     pub version: Option<Box<str>>,
     pub source: Box<str>,
-    // contract_app_id - Application ID is hash(owner,name) saved in package manager contract
-    pub contract_app_id: Option<Box<str>>,
+    // metadata - Contains application ID is hash(owner,name) saved in package manager contract
+    pub metadata: Option<Box<[u8]>>,
```
miraclx: can just be `Box<[u8]>`, an empty array can indicate a lack of metadata

also, no need for the comment

crates/context/src/lib.rs
frdomovic: Updated

crates/meroctl/src/cli/app/install.rs
frdomovic: reverted

crates/primitives/src/application.rs
frdomovic: updated

crates/server-primitives/src/admin.rs
frdomovic: updated

crates/store/src/types/application.rs
frdomovic: Yes, updated

crates/context/src/lib.rs
```diff
@@ -344,13 +346,14 @@ impl ContextManager {
             eyre::bail!("non-absolute path")
         };
 
-        self.install_application(blob_id, uri.as_str().parse()?, version)
+        self.install_application(blob_id, uri.as_str().parse()?, version, Vec::new())
```
miraclx: let's take in `metadata` here as well

crates/context/src/lib.rs
```diff
@@ -428,6 +431,8 @@ impl ContextManager {
             blob: application.blob.blob_id(),
             version: application.version.as_deref().map(str::parse).transpose()?,
             source: application.source.parse()?,
+            metadata: application
+                .metadata.to_vec()
```
miraclx: `cargo +nightly fmt`

crates/node/src/catchup.rs
```diff
@@ -391,6 +391,7 @@ impl Node {
                         .install_application_from_url(
                             change.source.to_string().parse()?,
                             change.version,
+                            Vec::new()
```
miraclx: we should receive application metadata from the peer, you can introduce a comment here to expand the type of `change` to include metadata information

crates/primitives/src/application.rs
```diff
@@ -94,10 +94,12 @@ impl fmt::Display for ApplicationSource {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct Application {
+    // id - Application ID created in the node for identification - see line 324 crates/context/src/lib.rs
     pub id: ApplicationId,
```
miraclx: nit~ is this comment useful? seems to me it just says `id: ApplicationId` which we already have in-code

miraclx (approved):

crates/primitives/src/application.rs
frdomovic: removed

crates/context/src/lib.rs
frdomovic: updated

crates/context/src/lib.rs
frdomovic: Added, this also required change in CLI commands

miraclx (approved):

PR #589, opened 2024-08-16
fix: Request failed with status: 415 Unsupported Media Type

MatejVukosav:
Private and public keys were required but not provided.

- private and public keys are now optional parameter
- generate new keypair if not provided

### DIFF
--- a/crates/server/src/admin/handlers/context.rs
+++ b/crates/server/src/admin/handlers/context.rs
@@ -14,6 +14,7 @@ use tower_sessions::Session;
 
 use crate::admin::service::{parse_api_error, AdminState, ApiError, ApiResponse, Empty};
 use crate::admin::storage::client_keys::get_context_client_key;
+use crate::admin::utils::identity::generate_identity_keypair;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ContextObject {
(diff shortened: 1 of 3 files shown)


crates/server/src/admin/handlers/context.rs
```diff
@@ -269,10 +261,16 @@ pub async fn join_context_handler(
         }
     };
 
-    // Create a KeyPair from the provided public and private keys
-    let initial_identity = KeyPair {
-        public_key: request.public_key,
-        private_key: Some(request.private_key),
+    //TODO Should we check if the context exists before trying to join it?
```
MatejVukosav: meaning: if user is already part of that context so to not join him twice?

MatejVukosav: maybe only check for the same identity but allow using multiple identities?

miraclx: I think it comes down to the same observable behaviour - user has joined context

whether or not it existed before, what do you think? is there value to feedback about just joined or not?

but also, the identity might already exist on some other machine, just waiting for catchup

in which case, you can't definitively say when the user actually "joins" the context, just when they are registered to have joined on the local machine

I think it's fine if we obscure that detail and focus on what the outcome is

miraclx (approved):

crates/server/src/admin/handlers/context.rs
MatejVukosav: Agree. I don't see value in knowing as output is the same. Just wasn't sure if second join will make some inconsistencies or not.

PR #590, opened 2024-08-16
refactor: unify context creation

MatejVukosav:
Unify context creation code in `node` and `server`.

- remove contextId from parameters during context creation with CLI
- indicate that private_key is optional in CLI

Test
`cargo run -p  meroctl -- --node-name node1 --home data context create --application-id 5Fi5jFcwoNVkV9Ngzc2Gs18GKKFtaAbsg72bdxi4FAhV`

`context create 5Fi5jFcwoNVkV9Ngzc2Gs18GKKFtaAbsg72bdxi4FAhV
`

### DIFF
--- a/crates/server/src/admin/utils/identity.rs
+++ b/crates/server/src/admin/utils/identity.rs
@@ -1,5 +1,6 @@
 use calimero_primitives::identity::{KeyPair, PublicKey};
 use ed25519_dalek::{SigningKey, VerifyingKey};
+use rand::RngCore;
 
 pub fn generate_identity_keypair() -> KeyPair {
     let member_seed = [0u8; 32];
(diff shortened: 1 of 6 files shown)


crates/server/src/admin/utils/identity.rs
```diff
@@ -10,3 +12,14 @@ pub fn generate_identity_keypair() -> KeyPair {
         private_key: Some(*member_signing_key.as_bytes()),
     }
 }
+
+pub fn generate_context_id() -> calimero_primitives::context::ContextId {
+    // Create a Send-able RNG
+    let mut rng = StdRng::from_entropy();
```
miraclx: can we use `rand::thread_rng` here?

miraclx (approved):
lgtm, thank you!

Eventually, all these pieces will be consolidated in the "client" (ref: [Architecture](https://github.com/calimero-network/core#architecture))

crates/server/src/admin/utils/identity.rs
MatejVukosav: Sure! Updated.

PR #591, opened 2024-08-16
refactor: unify context join

MatejVukosav:
Share the code for context joining between `node` and `server`

- remove `public_key` from server parameters as we can construct it from `private_key`

### DIFF
--- a/crates/server/src/admin/handlers/context.rs
+++ b/crates/server/src/admin/handlers/context.rs
@@ -4,15 +4,13 @@ use std::sync::Arc;
 use axum::extract::Path;
 use axum::response::IntoResponse;
 use axum::{Extension, Json};
-use calimero_primitives::identity::{KeyPair, PublicKey};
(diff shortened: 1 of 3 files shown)


crates/server/src/admin/handlers/context.rs
```diff
@@ -238,21 +237,19 @@ pub async fn join_context_handler(
...
     };
 
-    let initial_identity = if let Some(Json(json_body)) = request {
-        // Create a KeyPair from the provided public and private keys
-        KeyPair {
-            public_key: json_body.public_key,
-            private_key: Some(json_body.private_key),
-        }
+    let private_key = if let Some(Json(json_body)) = request {
+        Some(bs58::encode(json_body.private_key).into_string())
```
MatejVukosav, line 239: For some reason we are sending private key as `[u8; 32]`

miraclx (approved):

PR #595, opened 2024-08-19
feat(sdk): update json rpc request type

frdomovic:
# feat(sdk): update json rpc request type

## Summary
Updated types as json rpc now requires executor public key.
Type of number as public key bytes need to be transformed from uint8Array because its throwing 
error from server:` "invalid type: map, expected an array of length 32"`

### DIFF
--- a/packages/calimero-sdk/src/types/rpc.ts
+++ b/packages/calimero-sdk/src/types/rpc.ts
@@ -64,6 +64,8 @@ export interface RpcQueryParams<Args> {
   contextId: ContextId;
   method: string;
   argsJson: Args;
+  executorPublicKey: number[];
+
 }
 
 export interface RpcQueryResponse<Output> {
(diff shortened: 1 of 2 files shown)


packages/calimero-sdk/src/types/rpc.ts
```diff
@@ -74,6 +76,7 @@ export interface RpcMutateParams<Args> {
   contextId: ContextId;
   method: string;
   argsJson: Args;
+  executorPublicKey: number[];
```
miraclx, line 79: Sucks this made it here now (as it should be `bs58`), we could've been able to make this a `string` alias like https://github.com/calimero-network/core/blob/a6ac6526529e60d651ce4c70b51945804f01c364/packages/calimero-sdk/src/types/context.ts

we can't rn, (not trivially anyway), because https://github.com/calimero-network/core/pull/536#discussion_r1698298492 wasn't implemented. we'll have to revisit

miraclx (approved):

Issue #488, opened 2024-07-17
Implement initial context catchup


fbozic:
Currently, the node performs catch-up in case of missing `prior_hash` upon receiving `Transaction`, or in case of `TransactionRejection` from the coordinator. The catch-up process is a blocking process by design on both sides (client and server).

Our first attempt to perform an initial catch-up upon receiving the `Subscribed` event led to a bunch of timeouts because when multiple nodes are started at the same time they try to perform catch-up at the same time, reference: https://github.com/calimero-network/core/blob/dd01008cd8a7352b1c417d5b56289f1e3f079250/crates/node/src/lib.rs#L578-L596

Initial ideas to tackle this issue:
- Setup interval tick event in the [main control loop](https://github.com/calimero-network/core/blob/dd01008cd8a7352b1c417d5b56289f1e3f079250/crates/node/src/lib.rs#L80-L102), pick one random pending context and perform catchup. This interval tick should ideally be random as well because if we start N nodes at the same time with the same interval we will end up with timeouts again.
- Make the server part of the catch-up process non-blocking. E.g. define the maximum number of streams node/network accepts and processes in parallel. For a new incoming stream that exceeds a maximum number of streams, we could send a new message variant indicating that the peer is already serving catch-up data and is not willing to accept a new stream. Keep in mind that we need to access `ContextManager`, `Store`, `NetworkClient` (due to config), and `TxPool` (this one might require extra care) from the catch-up reply task.

PR #602, opened 2024-08-19
feat: implement pending catchup tick

fbozic:
Resolves: https://github.com/calimero-network/core/issues/488

There are other bugs in the catchup itself which I will fix in another PR.

### DIFF
--- a/crates/node/src/catchup.rs
+++ b/crates/node/src/catchup.rs
@@ -206,9 +206,16 @@ impl Node {
             None => return,
         };
 
+        info!(%context_id, %peer_id, "Performing interval catchup");
+
         if let Err(err) = self.perform_catchup(context_id, peer_id).await {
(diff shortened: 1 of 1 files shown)


crates/node/src/catchup.rs
```diff
@@ -188,6 +190,27 @@ impl Node {
         Ok(())
     }
 
+    pub(crate) async fn handle_interval_catchup(&mut self) {
+        let context_id = match self.ctx_manager.get_any_pending_catchup_context().await {
```
MatejVukosav, line 194: Do you want to add log so we can see when it started?

MatejVukosav (approved):

crates/node/src/catchup.rs
fbozic: Sure, but I will put it right before the opening of the stream since.

PR #604, opened 2024-08-19
fix: remove context identity exchange from catchup and other bugs

fbozic:
- In `ctx_manager` hardcode default `KeyPair` and store it (ignoring passed `KeyPair`)
  - All peers in the context will have 1 ContextIdentity with the default `KeyPair`
- Cleanup unused code from catchup

Resolves: #605

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -1,5 +1,4 @@
 use std::collections::HashSet;
-use std::str::FromStr;
 use std::sync::Arc;
 
 use calimero_network::client::NetworkClient;
@@ -339,19 +338,14 @@ impl ContextManager {
         version: Option<semver::Version>,
         metadata: Vec<u8>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
-        let path_without_prefix = path
-            .as_str()
-            .strip_prefix("file://")
-            .unwrap_or(path.as_str());
-        let utf_path_buf = Utf8PathBuf::from_str(path_without_prefix)?;
-        let file = fs::File::open(utf_path_buf).await?;
+        let file = fs::File::open(&path).await?;
 
         let blob_id = self
             .blob_manager
             .put(tokio_util::io::ReaderStream::new(file))
             .await?;
 
-        let Ok(uri) = reqwest::Url::from_file_path(path_without_prefix) else {
+        let Ok(uri) = reqwest::Url::from_file_path(path) else {
             eyre::bail!("non-absolute path")
         };
 
--- a/crates/node/src/catchup.rs
+++ b/crates/node/src/catchup.rs
@@ -370,30 +370,16 @@ impl Node {
                     .ctx_manager
                     .is_application_installed(&change.application_id)?
                 {
-                    if change.source.to_string().starts_with("http://")
-                        || change.source.to_string().starts_with("https://")
-                    {
-                        info!("Installing application from the url");
-                        self.ctx_manager
-                            .install_application_from_url(
-                                change.source.to_string().parse()?,
-                                change.version,
-                                Vec::new(),
-                            )
-                            .await?;
-                    } else {
-                        // TODO: for path sources, share the blob peer to peer
-                        // NOTE: this will fail if the path is not accessible by the node
-                        info!("Installing application from the path");
+                    // note! for now, we assume all paths are urls
+                    // todo! for path sources, share the blob peer to peer
 
-                        self.ctx_manager
-                            .install_application_from_path(
-                                change.source.to_string().parse()?,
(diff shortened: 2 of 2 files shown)


frdomovic (approved):

frdomovic:
Everything looks alright to me.

crates/node/src/catchup.rs
```diff
@@ -418,13 +372,28 @@ impl Node {
                     // note! for now, we assume all paths are urls
                     // todo! for path sources, share the blob peer to peer
```
fbozic: @MatejVukosav can you just update these comments? if I understand correctly this will work for a file as long as the peer is running on the same machine

MatejVukosav: Yes because we are sharing local file path. Should be noted in docs with dev mode example. 

Add whatever you think should be in code as well.

fbozic: I've updated rust comments

crates/context/src/lib.rs
```diff
@@ -111,15 +123,13 @@ impl ContextManager {
...
-        handle.put(&identity_key, &context_identity)?;
-
-        self.subscribe(&context.id).await?;
+        if let Some(initial_identity) = initial_identity {
+            let identity_key = calimero_store::key::ContextIdentity::new(
+                context.id,
+                initial_identity.public_key.clone(),
+            );
+            handle.put(&identity_key, &initial_identity.into())?;
+        }
```
miraclx: can we move this to `create_`?

crates/context/src/lib.rs
```diff
@@ -330,14 +339,19 @@ impl ContextManager {
         version: Option<semver::Version>,
         metadata: Vec<u8>,
     ) -> eyre::Result<calimero_primitives::application::ApplicationId> {
-        let file = fs::File::open(&path).await?;
+        let path_without_prefix = path
+            .as_str()
+            .strip_prefix("file://")
+            .unwrap_or(path.as_str());
+        let utf_path_buf = Utf8PathBuf::from_str(path_without_prefix)?;
+        let file = fs::File::open(utf_path_buf).await?;
```
miraclx: did something break here that's fixed with this patch?

crates/node/src/catchup.rs
```diff
@@ -415,16 +370,30 @@ impl Node {
...
+                        self.ctx_manager
+                            .install_application_from_url(
+                                change.source.to_string().parse()?,
+                                change.version,
+                                Vec::new(),
+                            )
+                            .await?;
+                    } else {
+                        // TODO: for path sources, share the blob peer to peer
+                        // NOTE: this will fail if the path is not accessible by the node
```
miraclx: this is the concern that warranted the use of the URL, we can't assume both nodes are being run on the same machine.

this can "potentially" source any arbitrary file that just happens to be in the same location, even though it's not the same file.

miraclx (approved):
Thanks for being thorough, this is the right direction.

crates/node/src/types.rs
```diff
@@ -37,10 +35,8 @@ pub enum CatchupStreamMessage {
 pub struct CatchupRequest {
     pub context_id: calimero_primitives::context::ContextId,
     pub application_id: Option<calimero_primitives::application::ApplicationId>,
-    pub source: Option<calimero_primitives::application::ApplicationSource>,
     pub last_executed_transaction_hash: calimero_primitives::hash::Hash,
     pub batch_size: u8,
-    pub public_key: PublicKey,
```
miraclx, line 43: we probably want to keep this one? although, right now, there's no demonstration of private key control, so it's not as useful or verifiable, we can maybe reintroduce it when proper catchup of member list are implemented.

fbozic: > we can maybe reintroduce it when proper catchup of member list are implemented.

yes, that is my idea. if it is not used atm, remove it since catchup is complicated as it is plus it is not documented.

crates/node/src/catchup.rs
fbozic: cc @MatejVukosav for this

crates/context/src/lib.rs
fbozic: and cc @MatejVukosav for this as well

crates/node/src/types.rs
miraclx: Sounds good to me

miraclx (approved):

crates/context/src/lib.rs
fbozic: reverted

crates/node/src/catchup.rs
fbozic: reverted

PR #606, opened 2024-08-20
feat: expose context identities

MatejVukosav:
Output example

GET {{url}}/admin-api/contexts/2GAJpURPC2cGYgSkZNjk3oLWwmca6CYyMf7j4CUcWLZA/identities
```
{
    "data": {
        "identities": [
            "4zvwRjXUKGfvwnParsHAS3HuSVzV5cA4McphgmoCtajS"
        ]
    }
}
```

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -273,10 +273,12 @@ impl ContextManager {
         for (k, v) in first.into_iter().chain(iter.entries()) {
             let (k, v) = (k?, v?);
 
-            if k.context_id() == context_id {
-                if !only_owned_identities || v.private_key.is_some() {
-                    ids.push(PublicKey(k.public_key()));
-                }
+            if k.context_id() != context_id {
+                break;
+            }
+
+            if !only_owned_identities || v.private_key.is_some() {
+                ids.push(PublicKey(k.public_key()));
             }
         }
         Ok(ids)
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 use std::sync::Arc;
 
 use calimero_network::client::NetworkClient;
-use calimero_primitives::identity::KeyPair;
+use calimero_primitives::identity::{KeyPair, PublicKey};
 use camino::Utf8PathBuf;
 use futures_util::TryStreamExt;
 use reqwest::Url;
@@ -215,7 +215,7 @@ impl ContextManager {
         Ok(true)
     }
 
-    pub fn get_context_ids(
+    pub fn get_contexts_ids(
         &self,
         start: Option<calimero_primitives::context::ContextId>,
     ) -> eyre::Result<Vec<calimero_primitives::context::ContextId>> {
@@ -235,6 +235,30 @@ impl ContextManager {
             ids.push(key?.context_id());
         }
 
+        Ok(ids)
+    }
+
+    // Iterate over all identities in a context (from members and mine)
+    // and return only public key of identities which contains private key (in value)
+    // If there is private key then it means that identity is mine.
+    pub fn get_context_identities(
+        &self,
+        context_id: calimero_primitives::context::ContextId,
+    ) -> eyre::Result<Vec<PublicKey>> {
+        let handle = self.store.handle();
+
+        let mut iter = handle.iter::<calimero_store::key::ContextIdentity>()?;
+        let mut ids = Vec::<PublicKey>::new();
+
+        for (k, v) in iter.entries() {
+            let (k, v) = (k?, v?);
+
+            if k.context_id() == context_id {
+                if v.private_key.is_some() {
+                    ids.push(PublicKey::from(k.public_key()));
+                }
+            }
+        }
         Ok(ids)
     }
 
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -255,7 +255,7 @@ impl ContextManager {
 
             if k.context_id() == context_id {
                 if v.private_key.is_some() {
-                    ids.push(PublicKey::from(k.public_key()));
+                    ids.push(PublicKey(k.public_key()));
                 }
             }
         }
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -238,12 +238,10 @@ impl ContextManager {
         Ok(ids)
     }
 
-    // Iterate over all identities in a context (from members and mine)
-    // and return only public key of identities which contains private key (in value)
-    // If there is private key then it means that identity is mine.
-    pub fn get_context_identities(
+    fn get_context_identities(
         &self,
         context_id: calimero_primitives::context::ContextId,
+        only_owned_identities: bool,
     ) -> eyre::Result<Vec<PublicKey>> {
         let handle = self.store.handle();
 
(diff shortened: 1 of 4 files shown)


crates/context/src/lib.rs
```diff
@@ -238,6 +238,30 @@ impl ContextManager {
...
+
+        let mut iter = handle.iter::<calimero_store::key::ContextIdentity>()?;
+        let mut ids = Vec::<PublicKey>::new();
+
+        for (k, v) in iter.entries() {
+            let (k, v) = (k?, v?);
+
+            if k.context_id() == context_id {
+                if v.private_key.is_some() {
+                    ids.push(PublicKey::from(k.public_key()));
```
miraclx: can we use this here:

```suggestion
                    ids.push(PublicKey(k.public_key()));
```

just so it breaks when I change the signature of `PublicKey` to be consistent with the pattern established with `ContextId`, `ApplicationId` and `BlobId`

the reason for this is `k.public_key()` "should" return `PublicKey` already, but currently doesn't

that'll change soon

crates/server/src/admin/handlers/context.rs
```diff
@@ -13,8 +14,10 @@ use crate::admin::storage::client_keys::get_context_client_key;
 use crate::admin::utils::context::{create_context, join_context};
 
 #[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct ContextObject {
     context: calimero_primitives::context::Context,
+    context_identities: Vec<String>,
```
miraclx: will we always want all identities for the context when we're getting context metadata?

I think maybe this should be a different endpoint, wdyt?

crates/context/src/lib.rs
```diff
@@ -238,6 +238,30 @@ impl ContextManager {
...
+        let handle = self.store.handle();
+
+        let mut iter = handle.iter::<calimero_store::key::ContextIdentity>()?;
+        let mut ids = Vec::<PublicKey>::new();
+
+        for (k, v) in iter.entries() {
+            let (k, v) = (k?, v?);
+
+            if k.context_id() == context_id {
+                if v.private_key.is_some() {
```
miraclx: maybe introduce a `local_only` boolean parameter?

crates/server/src/admin/handlers/context.rs
MatejVukosav: I think it is relevant context metadata the same as app_id or last_tx_hash. 

Use case for this is to select with which identity you want to join selected context or in context details show context config data. 

Also, the payload should be small, as user won't have too much identities for a single context. I don't see such use case.

crates/context/src/lib.rs
```diff
@@ -238,6 +238,45 @@ impl ContextManager {
...
+
+        for (k, v) in iter.entries() {
+            let (k, v) = (k?, v?);
+
+            if k.context_id() == context_id {
+                if !only_owned_identities || v.private_key.is_some() {
+                    ids.push(PublicKey(k.public_key()));
+                }
+            }
+        }
```
fbozic: This iterates over all entries, but you already know part of the key. Why not use it, so we reduce in memory checks?
Reference: https://github.com/calimero-network/core/blob/1296f7d45974d63c3784d4d64fb29a38b5328652/crates/node/src/lib.rs#L566-L568

crates/server/src/admin/handlers/context.rs
fbozic: I agree this should be split. IMO we should avoid eagerly joining data in the server and allow clients to fetch it when needed.

crates/server/src/admin/handlers/context.rs
```diff
@@ -1,5 +1,6 @@
 use std::str::FromStr;
 use std::sync::Arc;
+use std::vec;
```
miraclx: is this unused?

crates/server/src/admin/handlers/context.rs
```diff
@@ -34,12 +37,26 @@ pub async fn get_context_handler(
...
-            Some(context) => ApiResponse {
-                payload: GetContextResponse {
-                    data: ContextObject { context },
-                },
+            Some(context) => {
+                let context_identities = state
+                    .ctx_manager
+                    .get_context_owned_identities(context.id)
+                    .map_err(|err| parse_api_error(err).into_response())
+                    .unwrap_or_default()
```
miraclx: looks to me like we're throwing this error away, yeah? we should have a `err.into_response` somewhere, unless you think it's irrelevant to share store-errors

crates/context/src/lib.rs
MatejVukosav: Thanks! Updated

crates/server/src/admin/handlers/context.rs
MatejVukosav: Doesn't mark as unused but also does not complain if I remove it.

there is below:
```
data: ContextUsers {
      context_users: vec![],
  },
```

crates/server/src/admin/handlers/context.rs
MatejVukosav: Idea is to not expose internal errors outside. So we map all errors to message we want to expose

MatejVukosav: pub fn parse_api_error(err: eyre::Report) -> ApiError {
    match err.downcast::<ApiError>() {
        Ok(api_error) => api_error,
        Err(original_error) => ApiError {
            status_code: StatusCode::INTERNAL_SERVER_ERROR,
            message: original_error.to_string(),
        },
    }
}

MatejVukosav: Do we really want to expose internal error outside? User can react only on invalid contextId. Not sure what else he can fix?

miraclx: You're probably right, if this fails, for a database issue, then maybe `get_context` failed as well and we would've returned.

Just concerned about saying we don't have context IDs when we do, but had an internal error.

There's always a chance retrying might work, maybe..

crates/server/src/admin/handlers/context.rs
miraclx: Ah, yeah that's already in the standard library prelude. No need to import explicitly.

crates/context/src/lib.rs
```diff
@@ -256,7 +256,21 @@ impl ContextManager {
...
+                break 'first None;
+            };
+
+            Some((k, iter.read()))
+        };
+
+        for (k, v) in first.into_iter().chain(iter.entries()) {
             let (k, v) = (k?, v?);
 
             if k.context_id() == context_id {
```
miraclx: let's flip this. if it doesn't match, we can break

crates/server/src/admin/handlers/context.rs
```diff
@@ -35,6 +33,44 @@ pub async fn get_context_handler(
         .get_context(&context_id)
         .map_err(|err| parse_api_error(err).into_response());
```
miraclx:
```suggestion
        .map_err(|err| parse_api_error(err));
```

or

```suggestion
        ;
```
and ..._

crates/server/src/admin/handlers/context.rs
```diff
@@ -35,6 +33,44 @@ pub async fn get_context_handler(
...
+                },
+            }
+            .into_response(),
+            None => ApiError {
+                status_code: StatusCode::NOT_FOUND,
+                message: "Context not found".into(),
+            }
+            .into_response(),
+        },
+        Err(err) => err.into_response(),
```
miraclx: _...

```suggestion
        Err(err) => parse_api_error(err).into_response(),
```

miraclx (approved):
lgtm

frdomovic (approved):
Tested, all good!

crates/server/src/admin/handlers/context.rs
MatejVukosav: Make sense. Updated to return generic error if there is some internal issue

crates/context/src/lib.rs
MatejVukosav: Updated

miraclx (approved):

PR #607, opened 2024-08-20
feat: Starknet login with ArgentX and Metamask Snap Wallets

alenmestrov:
# Implemented StarkNet Wallet Login with MetaMask and ArgentX

## Overview
This PR implements the functionality to allow users to log in using StarkNet-compatible wallets, specifically MetaMask and ArgentX. The flow includes connecting the wallet, receiving a challenge from the server, signing the challenge, and sending the signed message back to the server for verification. Also, add root key component was added for enabling users to add Starknet root keys

### 1. Wallet Integration
- **MetaMask & ArgentX Support**: Added support for connecting to StarkNet wallets via MetaMask and ArgentX.
- **Wallet Detection**: Implemented logic to detect if the user has a compatible wallet installed and prompt connection.

### 2. Challenge Signing
- **Request Challenge**: Upon successful wallet connection, the frontend requests a challenge message from the server.
- **Challenge Signing**: The user is prompted to sign the challenge using their connected wallet.
- **Send Signed Challenge**: The signed challenge with appropriate metadata (signature, message hash, and signature key) is sent back to the server for verification.

### 3. Server-Side Verification
- **Signature Verification**: The server-side logic verifies the signature using the public key associated with the user's wallet.
- **Message Integrity**: Verification of the message content by hashing the message on server and then comparing the hash to received hash from frontend, to ensure nothing hasn't been tampered with

### Frontend/Calimero SDK
- **Wallet Connection**: Implemented using StarkNet.js to interact with MetaMask and ArgentX wallets.
- **Challenge Handling**: Axios is used to communicate with the server for fetching and sending the challenge.

### Backend/Server
- **Challenge Generation**: A random challenge string is generated and sent to the client.
- **Signature Verification**: Utilized StarkNet’s cryptographic libraries to verify the signature on the server.
- **Error Handling**: Implemented error handling for cases such as wallet disconnection, invalid signatures, and timeouts.

## Notes
- **Verification of ArgentX signature is done on-chain since there isn't currently an option to do it otherwise (waiting for Starknet to implement a new way of signatures that ArgentX is using)**
- **For Metamask Snap verification we are using wallet address and derived public key received from the account, they are also in the process of updating the logic of verification**


packages/calimero-sdk/src/wallets/StarknetLogin/StarknetRootKey.tsx
```diff
@@ -0,0 +1,394 @@
...
+import { getStarknet, StarknetWindowObject } from 'get-starknet-core';
+
+interface StarknetRootKeyProps {
+  contextId?: string;
+  rpcBaseUrl: string;
+  successRedirect: () => void;
+  navigateBack?: () => void | undefined;
+}
+
+export function StarknetRootKey({
```
MatejVukosav, line 30: `props : StarknetRootKeyProps` so all props are automatically there and it is easier to recognize what is input

packages/calimero-sdk/src/wallets/StarknetLogin/StarknetRootKey.tsx
```diff
@@ -0,0 +1,394 @@
...
+          setSignData({
+            signature: signature,
+            messageHash: messageHash,
+          });
+        }
+      }
+    }catch(error) {
+      console.error('Error signing message:', error);
+      setErrorMessage('Error signing message');
+      //TODO handle error
```
MatejVukosav: Is this handled already?

packages/calimero-sdk/src/wallets/StarknetLogin/StarknetRootKey.tsx
```diff
@@ -0,0 +1,394 @@
...
+              { name: "publicKey", type: "string"},
+            ],
+          },
+          primaryType: "Challenge",
+          message: {
+            nodeSignature: walletSignatureData.payload.message.nodeSignature,
+            publicKey: walletSignatureData.payload.message.publicKey
+          }
+        };
+        const signature = await starknetAccount.account.signMessage(message);
```
MatejVukosav: can u add types?

packages/calimero-sdk/src/wallets/StarknetLogin/LoginWithStarknet.tsx
```diff
@@ -0,0 +1,390 @@
...
+          textAlign: 'center',
+        }}
+      >
+        <span
+          style={{
+            marginBottom: '0.5rem',
+            color: '#fff',
+          }}
+        >
+          Login with Starknet
```
MatejVukosav, line 310: translations?

crates/primitives/src/identity.rs
```diff
@@ -90,6 +90,10 @@ pub enum WalletType {
         #[serde(rename = "chainId")]
         chain_id: u64,
     },
+    SN {
```
MatejVukosav: Let's use a full name to make our life easier :)

crates/server/src/admin/utils/auth.rs
```diff
@@ -61,14 +73,79 @@ pub fn verify_node_signature(
...
+                        status_code: StatusCode::BAD_REQUEST,
+                        message: "Invalid wallet signature type for Starknet.".into(),
+                    })
+                }
+            };
+
+            let result = match wallet_name.as_str() {
+                "argentX" => tokio::task::block_in_place(|| {
+                    tokio::runtime::Runtime::new()
+                        .unwrap()
```
MatejVukosav: Will this stop the node if it fails?

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,310 @@
...
+    let parsed_signature: Vec<_> = signature
+        .iter()
+        .map(|s| Felt::from_str(s).unwrap())
+        .collect();
+
+    // Prepare a function call to verify the signature on-chain
+    let function_call = FunctionCall {
+        contract_address: wallet_address,
+        entry_point_selector: Felt::from_str(
+            "0x213dfe25e2ca309c4d615a09cfc95fdb2fc7dc73fbcad12c450fe93b1f2ff9e",
```
MatejVukosav: What is this string? Can u extract it into meaningful variable

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,310 @@
...
+    wallet_address: String,
+    message: &str,
+) -> eyre::Result<bool> {
+    // Convert inputs from strings to StarkNet-compatible types
+    let wallet_address = Felt::from_str(&wallet_address).unwrap();
+    let message_hash = Felt::from_str(&message_hash).unwrap();
+
+    // Set up a JSON-RPC client to interact with the StarkNet blockchain
+    let provider = JsonRpcClient::new(HttpTransport::new(
+        Url::parse("https://starknet-mainnet.public.blastapi.io").unwrap(),
```
MatejVukosav: Extract to const

packages/calimero-sdk/src/api/nodeApi.ts
```diff
@@ -14,14 +14,20 @@ interface ETHWalletType extends WalletTypeBase<'ETH'> {
...
+export type WalletType = ETHWalletType | NEARWalletType | SNWalletType;
 
 export namespace WalletType {
   export let NEAR: WalletType = { type: 'NEAR' } as NEARWalletType;
 
   export function ETH({ chainId = 1 }: { chainId?: number }): WalletType {
     return { type: 'ETH', chainId } as ETHWalletType;
   }
+
+  export function SN({ walletName = 'MS'}: { walletName?: string }): WalletType {
```
MatejVukosav: MS or SN?

packages/calimero-sdk/src/wallets/StarknetLogin/LoginWithStarknet.tsx
```diff
@@ -0,0 +1,390 @@
...
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [signData, setSignData] = useState<any>(null);
+  const [loading, setLoading] = useState<boolean>(false);
+
+  const walletLogin = useCallback(async (walletType: string) => {
+    try {
+      setLoading(true);
+      const starknetInstance = getStarknet();
+      if (starknetInstance) {
+        if(walletType === 'argentX') {
```
MatejVukosav: Extract to const

packages/calimero-sdk/src/wallets/StarknetLogin/LoginWithStarknet.tsx
alenmestrov: Translations are in admin-dashboard.

crates/server/src/admin/utils/auth.rs
alenmestrov: No, this only blocks the current thread where it's executed. It does not interfere with other asynchronous tasks.

crates/primitives/src/identity.rs
alenmestrov: Updated to full name

packages/calimero-sdk/src/wallets/StarknetLogin/StarknetRootKey.tsx
alenmestrov: Added types for signature and message hash

packages/calimero-sdk/src/wallets/StarknetLogin/StarknetRootKey.tsx
alenmestrov: yep

packages/calimero-sdk/src/wallets/StarknetLogin/StarknetRootKey.tsx
alenmestrov: We decided to leave this like this since we are using Callbacks because of which we would need to destructure it anyway before using them as dependencies.

crates/server-primitives/src/admin.rs
```diff
@@ -51,7 +51,7 @@ pub struct GetApplicationResult {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AddPublicKeyRequest {
-    pub wallet_signature: String,
+    pub wallet_signature: WalletSignature,
```
frdomovic, line 119: Are other login cases update to work with this update?

packages/calimero-sdk/src/wallets/StarknetLogin/LoginWithStarknet.tsx
```diff
@@ -0,0 +1,390 @@
...
+              setStorageNodeAuthorized();
+              successRedirect();
+            }
+          })
+          .catch(() => {
+            console.error('error while login!');
+            setErrorMessage('Error while login!');
+          });
+      }
+    }catch(error) {
```
frdomovic:
```suggestion
    } catch(error) {
```
Use lint to format code

crates/server-primitives/src/admin.rs
alenmestrov: yes, they are, now it is separated into two options, one is String (for old ones NEAR and ETH) and the other one is StarknetPayload which serves for Starknet. In verify_node_function there is the extraction of string in the case of ETH and NEAR while in the case of Starknet, it extracts the payload.

crates/primitives/src/identity.rs
```diff
@@ -90,6 +90,10 @@ pub enum WalletType {
         #[serde(rename = "chainId")]
         chain_id: u64,
     },
+    STARKNET {
+        #[serde(rename = "walletName")]
```
miraclx, line 139: can we make this

```suggestion
    #[serde(rename_all = "camelCase")]
    STARKNET {
```

we can apply the same to `ETH`

crates/server/src/admin/utils/auth.rs
miraclx: I think it makes more sense if the function this is in is `async` instead ..._

crates/server/src/admin/utils/auth.rs
```diff
@@ -15,10 +16,11 @@ use tracing::info;
 use crate::admin::service::{parse_api_error, ApiError};
 use crate::admin::storage::root_key::get_root_key;
 use crate::verifysignature::verify_near_signature;
+use crate::verifysnsignature::{verify_argent_signature, verify_metamask_signature};
 
 pub fn verify_node_signature(
```
miraclx: _...

```suggestion
pub async fn verify_node_signature(
```

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,313 @@
...
+    let message_hash = Felt::from_str(&message_hash).unwrap();
+
+    // Set up a JSON-RPC client to interact with the StarkNet blockchain
+    let provider = JsonRpcClient::new(HttpTransport::new(Url::parse(rpc_node_url).unwrap()));
+
+    // Parse the signature strings into Felt types
+    let parsed_signature: Vec<_> = signature
+        .iter()
+        .map(|s| Felt::from_str(s).unwrap())
+        .collect();
```
miraclx: this should work, I think

```suggestion
        .map(|s| Felt::from_str(s))
        .collect::<Result<_, _>>()?;
```

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,313 @@
...
+    message_hash: String,
+    signature: Vec<String>,
+    wallet_address: String,
+    message: &str,
+    rpc_node_url: &str,
+    chain_id: &str,
+) -> eyre::Result<bool> {
+    // Convert inputs from strings to StarkNet-compatible types
+    let wallet_address = Felt::from_str(&wallet_address).unwrap();
+    let message_hash = Felt::from_str(&message_hash).unwrap();
```
miraclx: let's avoid unwrapping

```suggestion
    let wallet_address = Felt::from_str(&wallet_address)?;
    let message_hash = Felt::from_str(&message_hash)?;
```

and others

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,313 @@
...
+        Err(err) => {
+            eyre::bail!("Error verifying signature: {:?}", err);
+        }
+    }
+}
+
+// Function to verify a MetaMask Snap wallet signature off chain
+pub fn verify_metamask_signature(
+    message_hash: String,
+    signature: Vec<String>,
```
miraclx: this seems like it only needs a `&[&str]` maybe?

maybe even a `&[impl AsRef<str>]` so it works for `&[&str]`, `&[String]`, `Vec<&str>`, `Vec<String>`

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+        Err(err) => {
+            eyre::bail!("Error verifying signature: {:?}", err);
+        }
+    }
+}
+
+// Function to verify a MetaMask Snap wallet signature off chain
+pub fn verify_metamask_signature(
+    message_hash: String,
```
miraclx: `&str` is a minimum-requirements `String`

if the function needs a `String` (which is an owned memory location) then it should reflect that in the function signature

and if it only needs a reference to a string, then `&str` is sufficient

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,313 @@
...
+    signature: Vec<String>,
+    signing_key: String,
+    message: &str,
+    wallet_address: String,
```
miraclx:
```suggestion
    wallet_address: &str,
```

crates/server/src/verifysnsignature.rs
```diff
@@ -0,0 +1,313 @@
...
+// Structure representing a field in a StarkNet type with a name and type
+#[derive(Serialize, Deserialize, Debug)]
+struct FieldType {
+    name: String,
+    #[serde(rename = "type")]
+    field_type: String,
+}
+
+// Structure holding definitions of StarkNet types: StarknetDomain and Challenge
+#[derive(Serialize, Deserialize, Debug)]
```
miraclx: is this structure serialized/deserialized? I can find a piece where it's constructed and eventually destructured, but not {de,}serialized

crates/server-primitives/src/admin.rs
```diff
@@ -102,16 +112,34 @@ pub struct NearSignatureMessageMetadata {
...
+#[serde(untagged)]
+pub enum WalletSignature {
+    String(String),
+    StarknetPayload(StarknetPayload),
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StarknetPayload {
+    pub signature: Vec<String>,
```
miraclx, line 241: this is just 2 parameters, yeah?

maybe `[String, 2]` or split it into it's constituents `r` and `s`

crates/server/src/lib.rs
```diff
@@ -18,6 +18,7 @@ pub mod jsonrpc;
 #[cfg(feature = "admin")]
 mod middleware;
 mod verifysignature;
+mod verifysnsignature;
```
miraclx: these look similar, maybe we reorganize to the module structure `verify::starknet`?

MatejVukosav: sgtm. 

even `verifysnsignature` should have full name `verifystarknetnsignature` -> we will forget "s", "p", "n", "k" parts pretty soon if we start using it in that way

frdomovic:
Rebase before merging, apply [this](https://github.com/calimero-network/core/pull/616) changes to your login method

crates/server-primitives/src/admin.rs
alenmestrov: depending on the wallet, the Argent wallet returns 5 elements in the Vec while Metamask Snap returns 3, it isn't constant and is subjugated to a possible change in the future, so I implemented it like this.

crates/server/src/admin/utils/auth.rs
alenmestrov: For this, the thing is that currently, all parent functions are not async, this is why I implemented it in this way. The reason is that I didn't want to change the current implementation of all other wallets.

crates/primitives/src/identity.rs
alenmestrov: here we cannot use rename_all since it is designed to be used on containers, not fields. Since we already have rename_all = "UPPERCASE" on WalletType I would leave it like it is now.

crates/server/src/verifysnsignature.rs
alenmestrov: Removed Serialize and Deserialize from here

crates/server/src/admin/utils/auth.rs
```diff
@@ -106,7 +106,7 @@ pub fn verify_node_signature(
 
             let (message_hash, signature) = match wallet_signature {
                 WalletSignature::StarknetPayload(payload) => {
-                    (payload.message_hash.clone(), payload.signature.clone())
+                    (&payload.message_hash, payload.signature.clone())
```
miraclx:
```suggestion
                    (&payload.message_hash, &payload.signature)
```

crates/server/src/admin/utils/auth.rs
```diff
@@ -131,26 +131,36 @@ pub fn verify_node_signature(
...
+                "metamask" => {
+                    let wallet_address: &str = match &wallet_metadata.wallet_address {
+                        Some(s) => s,
+                        None => {
+                            return Err(ApiError {
+                                status_code: StatusCode::BAD_REQUEST,
+                                message: "Wallet address not present.".into(),
+                            })
+                        }
+                    };
```
miraclx: style: with the let-else syntax, you can rewrite as:

```suggestion
                    let Some(wallet_address) = &wallet_metadata.wallet_address else {
                        return Err(ApiError {
                            status_code: StatusCode::BAD_REQUEST,
                            message: "Wallet address not present.".into(),
                        })
                    };
```

up to you tho

crates/server/src/admin/utils/auth.rs
```diff
@@ -131,26 +131,36 @@ pub fn verify_node_signature(
...
-                    wallet_metadata.signing_key.clone(),
-                    &payload.message.message,
-                    wallet_metadata.wallet_address.clone().unwrap_or_default(),
-                    &chain_id,
-                ),
+                "argentX" => {
+                    verify_argent_signature(
+                        message_hash,
+                        signature,
+                        &wallet_metadata.signing_key.clone(),
```
miraclx: can we use `signing_key` for private key references, and `verifying_key`/`public_key` for public key references?

also:

```suggestion
                        &wallet_metadata.signing_key,

```

crates/server/src/verifywalletsignatures/starknet.rs
```diff
@@ -0,0 +1,320 @@
...
+        challenge: vec![
+            FieldType {
+                name: "nodeSignature".to_string(),
+                field_type: "string".to_string(),
+            },
+            FieldType {
+                name: "publicKey".to_string(),
+                field_type: "string".to_string(),
+            },
+        ],
```
miraclx: will `start_net_domain` always contain 4 items? and `challenge` 2? if so, you can represent that in the type..._

crates/server/src/verifywalletsignatures/starknet.rs
```diff
@@ -0,0 +1,320 @@
...
+    name: String,
+    #[serde(rename = "type")]
+    field_type: String,
+}
+
+// Structure holding definitions of StarkNet types: StarknetDomain and Challenge
+#[derive(Debug)]
+struct Types {
+    stark_net_domain: Vec<FieldType>,
+    challenge: Vec<FieldType>,
```
miraclx: _...

```suggestion
    stark_net_domain: [FieldType; 4],
    challenge: [FieldType; 2],
```

miraclx (approved):
lgtm

PR #611, opened 2024-08-21
feat: enable defining context id in context creation cli command

MatejVukosav:
Test

`cargo run -p meroctl -- --node-name coordinator --home data context create -c <contextId> --watch <path>`

### DIFF
--- a/crates/meroctl/src/cli/context/create.rs
+++ b/crates/meroctl/src/cli/context/create.rs
@@ -44,7 +44,7 @@ impl CreateCommand {
 
         let client = Client::new();
 
-        match dbg!(self) {
+        match self {
             CreateCommand {
                 application_id: Some(app_id),
                 watch: None,
@@ -84,26 +84,18 @@ impl CreateCommand {
             CreateCommand {
                 application_id: None,
                 watch: Some(path),
-                context_id: Some(context_id),
-                metadata: None,
+                context_id,
+                metadata,
             } => {
                 let path = path.canonicalize_utf8()?;
-                let application_id = install_app(multiaddr, path.clone(), &client, None).await?;
-                let context_id =
-                    create_context(multiaddr, application_id, &client, Some(context_id)).await?;
-                watch_app_and_update_context(multiaddr, context_id, path, &client, None).await?;
-            }
-            CreateCommand {
-                application_id: None,
-                watch: Some(path),
-                metadata: None,
-                context_id: None,
-            } => {
-                let path = path.canonicalize_utf8()?;
-                let application_id = install_app(multiaddr, path.clone(), &client, None).await?;
-                let context_id = create_context(multiaddr, application_id, &client, None).await?;
-
-                watch_app_and_update_context(multiaddr, context_id, path, &client, None).await?;
+                let application_id =
+                    install_app(multiaddr, path.clone(), &client, metadata.clone()).await?;
+                let context_id = match context_id {
+                    Some(context_id) => context_id,
+                    None => create_context(multiaddr, application_id, &client, context_id).await?,
+                };
+                watch_app_and_update_context(multiaddr, context_id, path, &client, metadata)
+                    .await?;
             }
             _ => eyre::bail!("Invalid command configuration"),
         }
--- a/crates/meroctl/src/cli/context/create.rs
+++ b/crates/meroctl/src/cli/context/create.rs
@@ -56,34 +56,6 @@ impl CreateCommand {
             CreateCommand {
                 application_id: None,
                 watch: Some(path),
-                context_id: None,
-                metadata: Some(metadata),
-            } => {
-                let path = path.canonicalize_utf8()?;
-                let application_id =
-                    install_app(multiaddr, path.clone(), &client, Some(metadata.clone())).await?;
-                let context_id = create_context(multiaddr, application_id, &client, None).await?;
-
-                watch_app_and_update_context(multiaddr, context_id, path, &client, Some(metadata))
-                    .await?;
-            }
-            CreateCommand {
-                application_id: None,
-                watch: Some(path),
-                context_id: Some(context_id),
-                metadata: Some(metadata),
-            } => {
-                let path = path.canonicalize_utf8()?;
-                let application_id =
-                    install_app(multiaddr, path.clone(), &client, Some(metadata.clone())).await?;
-                let context_id =
-                    create_context(multiaddr, application_id, &client, Some(context_id)).await?;
-                watch_app_and_update_context(multiaddr, context_id, path, &client, Some(metadata))
-                    .await?;
-            }
-            CreateCommand {
-                application_id: None,
-                watch: Some(path),
                 context_id,
                 metadata,
             } => {
--- a/crates/meroctl/src/cli/context/create.rs
+++ b/crates/meroctl/src/cli/context/create.rs
@@ -1,3 +1,4 @@
+use calimero_primitives::context::ContextId;
 use camino::Utf8PathBuf;
 use clap::Parser;
 use libp2p::Multiaddr;
@@ -12,13 +13,17 @@ use crate::config_file::ConfigFile;
 #[derive(Debug, Parser)]
 pub struct CreateCommand {
     /// The application ID to attach to the context
-    #[clap(long, short = 'a', exclusive = true)]
+    #[clap(long, short = 'a')]
     application_id: Option<calimero_primitives::application::ApplicationId>,
 
     /// Path to the application file to watch and install locally
-    #[clap(long, short = 'w', exclusive = true)]
+    #[clap(long, short = 'w')]
     watch: Option<Utf8PathBuf>,
+
     metadata: Option<Vec<u8>>,
+
+    #[clap(long, short = 'c')]
+    context_id: Option<ContextId>,
 }
 
 impl CreateCommand {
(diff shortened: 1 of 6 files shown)


crates/meroctl/src/cli/context/create.rs
```diff
@@ -39,25 +44,45 @@ impl CreateCommand {
 
         let client = Client::new();
 
-        match self {
+        // Manually check for exclusivity between application_id and watch
+        if self.application_id.is_some() && self.watch.is_some() {
```
MatejVukosav: Throws strange error 
```
error: the argument '--watch <WATCH>' cannot be used with one or more of the other specified arguments

Usage: meroctl context create [OPTIONS] [METADATA]...
```

if `app_id` and watch are exclusive when I provide `watch` and `context`-id.

crates/meroctl/src/cli/context/create.rs
```diff
@@ -39,25 +44,45 @@ impl CreateCommand {
...
+                let context_id = create_context(multiaddr, application_id, &client, None).await?;
 
                 let application_id =
                     install_app(multiaddr, path.clone(), &client, metadata.clone()).await?;
 
-                let context_id = create_context(multiaddr, application_id, &client).await?;
+                watch_app_and_update_context(multiaddr, context_id, path, &client, Some(metadata))
+                    .await?;
+            }
+            CreateCommand {
```
MatejVukosav: I don't like this chaining but didn't find better option.

crates/meroctl/src/cli/context/create.rs
miraclx: Ah, yeah.. this is intentional though, right? there shouldn't be a use-case for activating `application_id` and `watch` together

Although, the specifier to clap is too restrictive..._

crates/meroctl/src/cli/context/create.rs
```diff
@@ -12,13 +13,17 @@ use crate::config_file::ConfigFile;
 #[derive(Debug, Parser)]
 pub struct CreateCommand {
     /// The application ID to attach to the context
-    #[clap(long, short = 'a', exclusive = true)]
+    #[clap(long, short = 'a')]
     application_id: Option<calimero_primitives::application::ApplicationId>,
 
     /// Path to the application file to watch and install locally
-    #[clap(long, short = 'w', exclusive = true)]
+    #[clap(long, short = 'w')]
```
miraclx: _...these should be [`conflicts_with`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with) instead

- `application_id` conflicts with `watch`
- `context_id` and `metadata` [requires](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) `watch`

crates/meroctl/src/cli/context/create.rs
```diff
@@ -39,29 +44,72 @@ impl CreateCommand {
 
         let client = Client::new();
 
-        match self {
+        // Manually check for exclusivity between application_id and watch
+        if self.application_id.is_some() && self.watch.is_some() {
+            eyre::bail!("Cannot use --application-id and --watch together");
```
miraclx: at which point, we can remove these

crates/meroctl/src/cli/context/create.rs
```diff
@@ -12,13 +13,17 @@ use crate::config_file::ConfigFile;
...
-    #[clap(long, short = 'a', exclusive = true)]
+    #[clap(long, short = 'a')]
     application_id: Option<calimero_primitives::application::ApplicationId>,
 
     /// Path to the application file to watch and install locally
-    #[clap(long, short = 'w', exclusive = true)]
+    #[clap(long, short = 'w')]
     watch: Option<Utf8PathBuf>,
+
     metadata: Option<Vec<u8>>,
```
miraclx, line 23: out of scope for this PR, just thinking about this.. maybe a `String` makes more sense, as vecs imply multiple specifications of the arg or alternatively, comma-separated

```console
$ meroctl context create --metadata 1 --metadata 2
$ meroctl context create --metadata 1,2,3
```

let's just make it a string, and before we send it to the node, make it a vec

crates/meroctl/src/cli/context/create.rs
miraclx: I think we should be able to collapse these and selectively apply defaults, seems we only need to match on context ID

```rust
            CreateCommand {
                application_id: None,
                watch: Some(path),
                context_id,
                metadata,
            } => {
                let path = path.canonicalize_utf8()?;
                let application_id =
                    install_app(multiaddr, path.clone(), &client, metadata.clone()).await?;
                let context_id = match context_id {
                    Some(context_id) => context_id,
                    None => create_context(multiaddr, application_id, &client, Some(context_id)).await?,
                };
                watch_app_and_update_context(multiaddr, context_id, path, &client, metadata)
                    .await?;
            }
```

crates/meroctl/src/cli/context/mod.rs
```diff
@@ -9,13 +9,13 @@ mod list;
...
-  $ meroctl --home data/ --node-name node1 context ls
+  $ cargo run -p meroctl -- --home data --node-name node1 context ls
 
   # Create a new context
-  $ meroctl --home data/ --node-name node1 context create --application-id my-app-id
+  $ cargo run -p meroctl -- --home data --node-name node1 context create --application-id <appId>
 
   # Create a new context in dev mode
-  $ meroctl --home data/ --node-name node1 context create --dev --path /path/to/app --version 1.0.0
+  $ cargo run -p meroctl -- --home data --node-name node1 context create --watch <path>
```
miraclx: ooh, this is the CLI 😅

we ship binaries, so we don't have any dependency on `cargo`

```suggestion
  $ meroctl -- --home data --node-name node1 context ls
  # Create a new context
  $ meroctl -- --home data --node-name node1 context create --application-id <appId>
  # Create a new context in dev mode
  $ meroctl -- --home data --node-name node1 context create --watch <path>
```

crates/node/src/lib.rs
```diff
@@ -512,12 +515,15 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
-                            let mut iter = args.split(' ');
-                            let application = iter.next()?;
-                            let private_key = iter.next();
-                            Some((application, private_key))
-                        }) else {
+                        let Some((application_id, private_key, context_id)) =
+                            args.and_then(|args| {
+                                let mut iter = args.split(' ');
+                                let application = iter.next()?;
+                                let context_id = iter.next();
```
miraclx: I don't think we can reasonably support context ID here, since `private_key` is also optional

let's maybe forbid context specification from here, at least for now

we can beef it up later to provide some disambiguation between these entries

cause right now, if I don't want to specify a context id but do a private key

I'll have no choice but to specify both or my private key will be used as the context ID

and subscribe will announce it on the network, yay 😅

crates/meroctl/src/cli/context/create.rs
MatejVukosav: Nope, I didn't use app id and watch together. I used only watch and context id

crates/node/src/lib.rs
MatejVukosav: I would then rather remove private key as we are not using it anyway

crates/meroctl/src/cli/context/create.rs
```diff
@@ -43,22 +48,52 @@ impl CreateCommand {
...
+                metadata: Some(metadata),
+            } => {
+                let path = path.canonicalize_utf8()?;
+                let application_id =
+                    install_app(multiaddr, path.clone(), &client, Some(metadata.clone())).await?;
+                let context_id =
+                    create_context(multiaddr, application_id, &client, Some(context_id)).await?;
+                watch_app_and_update_context(multiaddr, context_id, path, &client, Some(metadata))
+                    .await?;
+            }
```
miraclx: we can remove these now

crates/node/src/lib.rs
```diff
@@ -512,11 +515,11 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                     }
                     "create" => {
-                        let Some((application_id, private_key)) = args.and_then(|args| {
+                        let Some((application_id, context_id)) = args.and_then(|args| {
                             let mut iter = args.split(' ');
                             let application = iter.next()?;
-                            let private_key = iter.next();
-                            Some((application, private_key))
+                            let context_id = iter.next();
+                            Some((application, context_id))
```
miraclx, line 522: just checking, did you mean to remove the private key?

crates/node/src/lib.rs
miraclx: Ah missed this, just refreshed

crates/node/src/lib.rs
miraclx: yes 😅 

https://github.com/calimero-network/core/pull/611#discussion_r1724837747

crates/meroctl/src/cli/context/create.rs
MatejVukosav: i will skip it in this PR as you mentioned it is out of scope. can u create a proper issue so we can make it right at once?

miraclx: https://github.com/calimero-network/core/issues/615

miraclx (approved):

PR #620, opened 2024-08-21
fix: app install metadata

MatejVukosav:
Metadata field was mandatory and it only worked if something is sent at the end.

I have set now that version is required and metadata are optional.  Extending params to be passed with arg name requires too much work atm.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -8,7 +8,6 @@ use calimero_store::Store;
 use libp2p::gossipsub::{IdentTopic, TopicHash};
 use libp2p::identity as p2p_identity;
 use owo_colors::OwoColorize;
-use semver::Version;
 use tokio::io::AsyncBufReadExt;
 use tokio::sync::{broadcast, mpsc, oneshot};
 use tracing::{debug, error, info, warn};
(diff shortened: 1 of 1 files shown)


crates/node/src/lib.rs
```diff
@@ -382,14 +383,21 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                             }
                             "file" => {
                                 let path = camino::Utf8PathBuf::from(resource);
 
+                                let metadata = match metadata {
+                                    Some(metadata) => metadata.as_bytes().to_vec(),
+                                    None => {
+                                        vec![]
+                                    }
+                                };
```
miraclx: https://github.com/calimero-network/core/blob/7ae9d670b1b4b7e1ba89ec2cdd87cd439ce59e0a/crates/node/src/lib.rs#L561

crates/node/src/lib.rs
```diff
@@ -356,18 +357,18 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
-                            let version = iter.next();
-                            let metadata = iter.next()?.as_bytes().to_vec();
+                            let version = iter.next()?;
+                            let metadata = iter.next();
 
                             Some((type_, resource, version, metadata))
                         }) else {
                             println!(
-                                "{IND} Usage: application install <\"url\"|\"file\"> <resource> [version] <metadata>"
+                                "{IND} Usage: application install <\"url\"|\"file\"> <resource> [version] [metadata]"
```
```suggestion
                                "{IND} Usage: application install <\"url\"|\"file\"> <resource> <version> [metadata]"
```

crates/node/src/lib.rs
```diff
@@ -356,18 +357,18 @@ async fn handle_line(node: &mut Node, line: String) -> eyre::Result<()> {
...
                             );
                             break 'done;
                         };
 
-                        let Ok(version) = version.map(|v| v.parse()).transpose() else {
+                        let Ok(version) = version.parse::<Version>() else {
```
miraclx: you can avoid the import up top if this isn't explicitly defined, rust should be able to infer it properly from `install_application_from{url,path}`

miraclx (approved):

PR #622, opened 2024-08-21
feat: add /applications/:app_id endpoint

MatejVukosav:
Add new endpoint to fetch app details

### DIFF
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -1,15 +1,17 @@
+use axum::extract::Path;
 use std::error::Error;
 use std::fmt::{self, Display, Formatter};
-use std::str;
+use std::str::{self, FromStr};
 use std::sync::Arc;
 
 use axum::http::{header, HeaderMap, HeaderValue, StatusCode};
 use axum::response::IntoResponse;
 use axum::routing::{delete, get, post};
 use axum::{Extension, Json, Router};
+use calimero_primitives::application::ApplicationId;
 use calimero_server_primitives::admin::{
-    ApplicationInstallResult, ApplicationListResult, InstallApplicationResponse,
-    ListApplicationsResponse,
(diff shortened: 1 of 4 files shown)


crates/server/src/admin/service.rs
```diff
@@ -99,7 +105,7 @@ pub(crate) fn setup(
             "/identity/keys",
             delete(handlers::root_keys::delete_auth_keys_handler),
         )
-        .layer(middleware::auth::AuthSignatureLayer::new(store))
+        // .layer(middleware::auth::AuthSignatureLayer::new(store))
```
miraclx: ^

crates/server/src/admin/service.rs
```diff
@@ -262,6 +268,43 @@ async fn list_applications_handler(
     }
 }
 
+async fn get_application_details_handler(
+    Path(app_id): Path<String>,
+    Extension(state): Extension<Arc<AdminState>>,
+) -> impl IntoResponse {
+    let app_id_result = match ApplicationId::from_str(&app_id) {
```
miraclx: PS: you can write this as this:

```suggestion
    let app_id_result = match app_id.parse() {
```

or if you really want to be explicit about the type definition (though it's inferred correctly from it's use in `get_application` below):

```suggestion
    let app_id_result = match app_id.parse::<ApplicationId>() {
```

miraclx (approved):

crates/server/src/admin/service.rs
```diff
@@ -262,6 +268,43 @@ async fn list_applications_handler(
     }
 }
 
+async fn get_application_details_handler(
```
alenmestrov, line 270: shouldn't this be in src/admin/handlers/applications.rs?

MatejVukosav: Fixed https://github.com/calimero-network/core/pull/626

PR #624, opened 2024-08-22
feat(sdk): add setup context id input

frdomovic:
# feat(sdk): add setup context id input

## Summary
To make context id switching dynamic and not hardcoded in envs when deploying application; input field for context id is added on the setup page. ContextId Is needed because in the process of login the api for fetching context identities is called. 

Fixes # -

## Test plan

https://github.com/user-attachments/assets/4cc12385-69e3-4bdf-8cfa-94a7c8120d03


MatejVukosav (approved):
ty!

Issue #582, opened 2024-08-09
Context coordinator joining ceremony design


fbozic:
Write a small design document (in crate README.md) defining what the context coordinator joining ceremony looks like.

PR #625, opened 2024-08-22
docs(node): coordinator joining ceremony

fbozic:
- Document basic node struct.
- Create placeholders for the things that need to be documented.
- Document coordinator joining ceremony. 

Resolves: #582

### DIFF
--- a/crates/node/README.md
+++ b/crates/node/README.md
@@ -60,6 +60,7 @@ Setup:
 - Context creator sends a message to the topic requesting a coordinator.
 - Interested coordinators send a message to the topic replying with the intent to coordinate.
 - Context creator selects a coordinator from the interested coordinators.
+- Context creator stores `PeerId` of the selected coordinator.
 - Context creator sends a message to the topic with the information about the selected coordinator.
 - Selected coordinator joins the context topic, while other coordinators clear pending request.
 
@@ -82,6 +83,7 @@ sequenceDiagram
     Peer->>+Peer:Handle [Offer]
     alt If pending request exists
         Peer->>Coordinator: Send [OfferAccepted]
+        Peer->>Peer:Update context with the coordinator `PeerId`
         Peer->>Peer:Clear pending request
     else
         Peer->>-Coordinator: Send [OfferDeclined]


crates/node/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Coordinator joining ceremony
+
+Coorindator joining ceremony is the process of acquiring the coordinator node for the context.
+
+Setup:
+- Coordinators and peers are subscribed to the dedicated coordinator ceremony topic.
+- Context creator sends a message to the topic requesting a coordinator.
+- Interested coordinators send a message to the topic replying with the intent to coordinate.
+- Context creator selects a coordinator from the interested coordinators.
```
MatejVukosav, line 62: How will this look like?

fbozic: Initial implementation is accept the first one, and future implementation can be based on staked amount or similar.

MatejVukosav (approved):

crates/node/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+TODO: Write about the transaction handling process and draw sequence diagram
+
+### Coordinator joining ceremony
+
+Coorindator joining ceremony is the process of acquiring the coordinator node for the context.
+
+Setup:
+- Coordinators and peers are subscribed to the dedicated coordinator ceremony topic.
+- Context creator sends a message to the topic requesting a coordinator.
```
miraclx, line 60: is this at the point of context creation? otherwise will this require the context creator to always be online?

miraclx (approved):
lgtm, now we can deny random confirmations from other nodes attempting to take over the network

crates/node/README.md
fbozic: good question, I would say yes although I see the issues it brings.

do you maybe have some alternative approach?

miraclx: At the point of creation is fine, so everyone expects to use the same coordinator. And if they're offline, nothing happens.

So far as catchup sends the coordinator all the transactions in the pool when they come back online. We need to verify that's the case.

fbozic: I've linked this comment to an implementation issue so we don't forget about it.

Issue #115, opened 2024-04-10
Create a rust code style guide for the repository


chefsale:
We should have a style guide for the rust codebase.

- [ ] use husky to run pre-commit checks - pre-commit checks include: run cargo format - 1h
- [x] use short qualified names -  use cargo format to make sure used names are following convention - 1h 
- [ ] cargo command for death code check - detect orphans https://crates.io/crates/cargo-modules - include in husky
- [x] use tests in a separate file than source code (update existing 48 tests) - update 1h 
- [x] check if there is lint rule to avoid having tests in source code file - https://rust-lang.github.io/rust-clippy/master/index.html#/tests_outside_test_module - include in husky - included estimation above

Create document to explain style guide - 3h
- [ ] mention above rules
- [ ] explain rule: if there is an enum variant with value, we should use a kind object

Risk factor:
- husky - 2h

MatejVukosav:
Ideas for more clippy rules: https://github.com/danwilliams/standards-rs?tab=readme-ov-file#code-linting  @miraclx go through and see what we can include

PR #630, opened 2024-08-23
chore: Linting and cleanup

danwilliams:
## Summary

This addresses some of the work required by:

  - https://github.com/calimero-network/core/issues/115
  - https://github.com/calimero-network/core/issues/433

Notably:

  - Separated out test files
  - Implemented linting rules
  - Changed to use short names instead of fully-qualified
  - Added Rust toolchain

Each change has been made atomically to aid review. The linting is at a good state where we can enable more at key points going forward.

## Testing

At present, the main codebase has been updated and the tests need to be updated next. Those will follow shortly.

Note that when reviewing, it is not worth trying to review all the code changes - it is better to look at the Git history, which can then be reviewed quickly and easily, skipping over those with bulk updates that are not of particular interest.


crates/meroctl/src/cli/context/create.rs
```diff
@@ -110,44 +120,48 @@ async fn create_context(
...
+) -> EyreResult<()> {
     let (tx, mut rx) = mpsc::channel(1);
 
-    let handle = tokio::runtime::Handle::current();
+    let handle = Handle::current();
     let mut watcher = notify::recommended_watcher(move |evt| {
         handle.block_on(async {
-            let _ = tx.send(evt).await;
-        })
+            drop(tx.send(evt).await);
```
miraclx, line 155: why `drop` over `let _ =`?

crates/network/src/config.rs
```diff
@@ -1,8 +1,13 @@
-use std::{fmt, time};
+use core::fmt::Formatter;
+use core::time::Duration;
```
miraclx: why core?

crates/network/src/lib.rs
```diff
@@ -217,94 +231,96 @@ impl EventLoop {
...
-                let _ = sender.send(Ok(topic));
+                drop(sender.send(Ok(topic)));
             }
             Command::OpenStream { peer_id, sender } => {
-                let _ = sender.send(self.open_stream(peer_id).await.map_err(Into::into));
+                drop(sender.send(self.open_stream(peer_id).await.map_err(Into::into)));
             }
             Command::PeerCount { sender } => {
-                let _ = sender.send(self.swarm.connected_peers().count());
+                let _ignore = sender.send(self.swarm.connected_peers().count());
```
miraclx, line 308: why `let _ignore =` over `let _ =` and `drop`?

`#[must_use]`?

MatejVukosav (approved):
Merge only if User Guide tests are successful

crates/meroctl/src/cli/context/create.rs
```diff
@@ -22,32 +33,32 @@ pub struct CreateCommand {
...
         match self {
-            CreateCommand {
+            Self {
                 application_id: Some(app_id),
                 watch: None,
-                metadata: _,
+                ..
             } => {
-                create_context(multiaddr, app_id, &client).await?;
+                let _ = create_context(multiaddr, app_id, &client).await?;
```
miraclx: why `let _ =` over `drop` and `let _ignore =`?

crates/identity/src/tests/auth.rs
miraclx, line 1: seems we're tied on the poll, we need to break it. I don't like the fact that deleting `src/auth.rs` won't provide any hints to the user to remove `src/tests/auth.rs`.

crates/meroctl/src/cli/context.rs
```diff
@@ -32,13 +36,13 @@ pub struct ContextCommand {
...
-    Create(create::CreateCommand),
-    Join(join::JoinCommand),
+    List(ListCommand),
+    Create(CreateCommand),
+    Join(JoinCommand),
 }
 
 impl ContextCommand {
-    pub async fn run(self, args: RootArgs) -> eyre::Result<()> {
+    pub async fn run(self, args: RootArgs) -> EyreResult<()> {
```
miraclx, line 45: oof, I.. do not like this one..

crates/network/src/discovery/state.rs
```diff
@@ -1,90 +1,98 @@
...
             .insert(addr.clone());
     }
 
     pub(crate) fn remove_peer(&mut self, peer_id: &PeerId) {
-        self.peers.remove(peer_id);
-        self.relay_index.remove(peer_id);
-        self.rendezvous_index.remove(peer_id);
+        drop(self.peers.remove(peer_id));
+        let _ = self.relay_index.remove(peer_id);
+        let _ = self.rendezvous_index.remove(peer_id);
```
miraclx, line 42: other such cases

crates/meroctl/src/cli/context/create.rs
danwilliams: Good practice. For a type with a non-trivial `Drop` implementation, this causes it to be dropped right away instead of waiting until the end of scope. In longer functions this can have more of an effect, but enabling the lint means we do it everywhere. There's no downside - only upside.

https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#let-underscore-drop

crates/network/src/config.rs
danwilliams: Where possible, we should ideally look to import from `core` instead of `std`, which paves the way for `no_std` compatibility, which is something we should bear in mind given we are targeting Wasm (smaller builds).

https://rust-lang.github.io/rust-clippy/master/index.html#std_instead_of_core

crates/network/src/lib.rs
danwilliams: In this situation, as you have correctly observed, a `must_use` directive is in place. However, that alone requires use of e.g. `let _ =`.

https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_must_use

In a situation where `Drop` is implemented along with `must_use`, a name should be given.

https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#let-underscore-drop

crates/meroctl/src/cli/context/create.rs
danwilliams: In a situation where a trivial type, such as a boolean, is returned, and there is no `Drop`, there is nothing to drop, so `let _ =` is correct vs `drop()`, and a name is not required.

crates/meroctl/src/cli/context.rs
danwilliams: There is a reason 🙂 

The main reason is consistency - if we are applying this naming approach to all types then it unifies things that are named generically `Error`, `Result`, and the like, and prevents issues where people may otherwise `use eyre::Result` which would then silently override the standard `Result`. For such cases it's a good pattern to follow.

crates/network/src/discovery/state.rs
danwilliams: See explanations and links to lints given above 🙂

crates/identity/src/tests/auth.rs
danwilliams: The outcome was decided/agreed previously and so I've followed that. If the poll is tied then we need to go one way or the other. We can always change it later. Orphans can be detected using the cargo command we are due to add in, but note that orphans can occur in either way, so that point is not a differentiator.

crates/primitives/src/hash.rs
```diff
@@ -1,74 +1,90 @@
...
+#[path = "tests/hash.rs"]
+mod tests;
+
+use core::cmp::Ordering;
+use core::fmt::{self, Debug, Display, Formatter};
+use core::hash::{Hash as StdHash, Hasher};
+use core::mem::MaybeUninit;
+use core::ops::Deref;
+use core::str::{from_utf8, FromStr};
+use std::io::Result as IoResult;
```
miraclx, line 11: `io::Result` was only imported conditionally before, that's no longer the case ..._

crates/primitives/src/hash.rs
```diff
@@ -1,74 +1,90 @@
...
+        Ok(Self {
             bytes: hasher.finalize().into(),
             bs58: MaybeUninit::zeroed(),
         })
     }
 
     #[cfg(feature = "borsh")]
-    pub fn hash_borsh<T: borsh::BorshSerialize>(data: &T) -> std::io::Result<Self> {
-        let mut hasher = sha2::Sha256::default();
+    pub fn hash_borsh<T: BorshSerialize>(data: &T) -> IoResult<Self> {
```
miraclx, line 60: it's only use is here

crates/network/src/config.rs
miraclx: Idk, right now, it's not very valuable to do this. Seeing as these crates still use `std` unconditionally, there's no functional reason to do it at this time without introducing unnecessary confusion, WASM support will be investigated in a future effort, at which point (armed with functional purpose) we can focus on details like these.

Here in `network::discovery::state`, we're dependent on:

https://github.com/calimero-network/core/blob/db771010bbabbdf81dfc4fa5e412eebce4ae410e/crates/network/src/discovery/state.rs#L1-L8

It's also not all crates that will undergo this effort, parts like the server, runtime, rocks store etc won't be compiled to WASM.

crates/meroctl/src/cli/context.rs
miraclx: > and prevents issues where people may otherwise use eyre::Result which would then silently override the standard Result

and how is it "prevent"ed?

crates/store/blobs/src/lib.rs
```diff
@@ -74,8 +75,15 @@ impl BlobManager {
...
             let mut buf = {
-                let mut buf = Vec::with_capacity(CHUNK_SIZE);
+                let mut buf: Vec<MaybeUninit<u8>> = Vec::with_capacity(CHUNK_SIZE);
                 unsafe { buf.set_len(CHUNK_SIZE) };
+
+                let buf = unsafe {
+                    // Convert the Vec<MaybeUninit<u8>> into a Vec<u8> once it's
+                    // fully-initialised
+                    Vec::from_raw_parts(buf.as_mut_ptr().cast::<u8>(), buf.len(), buf.capacity())
+                };
```
miraclx: You just created two `Vec`s backed by the same memory, that'll double-free..

The current behavior was intentional, as this is a byte array, zero-filled or having arbitrary data, it doesn't matter, it's still a valid byte array. More complex types could be invalidly represented, but not this.

The logic that follows determines the legality/validity of slices of the buffer as we only read slices we previously wrote. It has nothing to do with it's initialization.

But to correct the patch above, either of these would've been correct:

```suggestion
let mut buf = vec![0; CHUNK_SIZE];
```

or, reverting back to what we had before, and finally zero-filling it after allocation:

```suggestion
let mut buf = Vec::with_capacity(CHUNK_SIZE);
unsafe { buf.set_len(CHUNK_SIZE) };
std::ptr::write_bytes(buf.as_mut_ptr(), 0, CHUNK_SIZE)
```

crates/primitives/src/hash.rs
danwilliams: If it's important, a condition can be added there 🙂

crates/network/src/config.rs
danwilliams: It's something that I've gotten into the habit of doing over the years. It's becoming increasingly common practice, and is a Rust recommendation. It causes no harm.

crates/meroctl/src/cli/context.rs
danwilliams: It's prevented because it's not overriding the standard `Result`.

crates/store/blobs/src/lib.rs
danwilliams: Basically what I already said in our chat on the subject 🙂 You are correct - I had initially simply initialised it, and then thought that by using `MaybeUninit` it would be a more elegant approach, and more performant. But I've gone back to the first fix, as it's just simpler and removes the unsafe code entirely.

crates/meroctl/src/cli/context.rs
miraclx: Yet, nothing stops you from overwriting the standard result.

Especially with the single-identifier enforcement.

You'd have to know about the alias of `EyreResult` elsewhere to do the same.

While you're not allowed to use `eyre::Result`, there's nothing to guide you to the appropriate conventional use.

crates/identity/src/tests/auth.rs
miraclx: At no point was it "decided", there was no such acknowledgement.

PR #660, opened 2024-08-26
fix: runtime/examples rps

MatejVukosav:
From the root 

`cargo run --package calimero-runtime --example rps -- <path>`

### DIFF
--- a/crates/runtime/examples/rps.rs
+++ b/crates/runtime/examples/rps.rs
@@ -60,7 +60,7 @@ fn main() -> EyreResult<()> {
     let path = &args[1];
     let path = Path::new(path);
     if !path.exists() {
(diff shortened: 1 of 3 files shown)


crates/runtime/examples/rps.rs
```diff
@@ -46,7 +51,19 @@ struct GameOver {
...
+    let args: Vec<String> = env::args().collect();
+    if args.len() != 2 {
+        println!("Usage: {:?} <path-to-wasm>", args);
+        return Ok(());
+    }
+
+    let path = &args[1];
+    let path = Path::new(path);
+    if !path.exists() {
+        return eyre::Result::Err(eyre::eyre!("RPS wasm file not found"));
```
miraclx: eyre::bail

miraclx (approved):

PR #664, opened 2024-08-26
ci: add simple rust workflow to build and run tests

MatejVukosav:
- Use caching for dependencies and build artifacts to avoid recompiling unchanged code.
- Split jobs into parallel tasks to optimize CI time.
- Leverage incremental builds during development.
- Consider sccache to cache compiled artifacts across different runs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,22 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --verbose
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,9 +2,9 @@ name: Rust
 
 on:
   push:
-    branches: [ "master" ]
+    branches: ['master']
   pull_request:
-    branches: [ "master" ]
+    branches: ['master', 'chore-cli']
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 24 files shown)


xilosada (approved):

.github/workflows/rust.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install sccache
+      run: |
+        wget https://github.com/mozilla/sccache/releases/download/v0.5.3/sccache-v0.5.3-x86_64-unknown-linux-musl.tar.gz
+        tar -xzf sccache-v0.5.3-x86_64-unknown-linux-musl.tar.gz
+        sudo mv sccache-v0.5.3-x86_64-unknown-linux-musl/sccache /usr/local/bin/
+        rm -rf sccache-v0.5.3-x86_64-unknown-linux-musl*
```
miraclx: this is [a year old](https://github.com/mozilla/sccache/releases/tag/v0.5.3), why aren't we using the github action?

https://github.com/marketplace/actions/sccache-action

.github/workflows/rust.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Cache Cargo and sccache
+      uses: actions/cache@v4
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+          ~/.cache/sccache
```
miraclx: we should be using https://github.com/Swatinem/rust-cache here

.github/workflows/rust.yml
```diff
@@ -14,34 +14,28 @@ jobs:
...
+      - name: Cache Cargo
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('rust-toolchain.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
```
miraclx: this is now handled by `rust-cache`

MatejVukosav: Wasn't that outcome of comment before?

miraclx: rust-cache will cache, and [knows the right things to include in the cache](https://github.com/Swatinem/rust-cache#cache-details), we don't need to duplicate it here

miraclx (approved):

Issue #583, opened 2024-08-09
Context coordinator joining ceremony types


fbozic:
Implement basic types used in context coordinator joining ceremony.

MatejVukosav:
Deprecating in favour of a new synchronization mechanism using CRDT

PR #665, opened 2024-08-26
feat: PeerId primitive and coordinator ceremony types

fbozic:
- implement `PeerId` primitive type with Borsh SerDe
- cleanup usage of `PeerId` across the code base
- remove `libp2p` dependency from `node` crate

Resolves: #583


crates/store/Cargo.toml
```diff
@@ -18,7 +18,7 @@ strum = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 thunderdome.workspace = true
 
-calimero-primitives = { path = "../primitives" }
+calimero-primitives = { path = "../primitives", features = ["borsh"] }
```
miraclx, line 21:
```suggestion
calimero-primitives = { path = "../primitives"  }
```

this is already selected for with the `datatypes` feature on line 29

PR #666, opened 2024-08-26
feat(cli): Added auth to dev mode in the cli

petarjuki7:
# Dev Mode Authentification

## Summary

When using the cli, the client signs the timestamp with the private key of the keypair of the node, and appends it to a header (the signature and the timestamp).
On the server side, it verifies the signature using the node's public key.

## Test plan

Tested commands
context create
context list
app list

Tried to do the same functionality just using curl, failing because missing authorization


crates/meroctl/src/common.rs
```diff
@@ -26,3 +29,28 @@ pub fn multiaddr_to_url(multiaddr: &Multiaddr, api_path: &str) -> EyreResult<Url
 
     Ok(url)
 }
+
+pub async fn get_response<T: Serialize>(
+    client: &Client,
+    url: Url,
+    request: Option<T>,
```
miraclx:
```suggestion
    body: Option<T>,
```

crates/meroctl/src/common.rs
```diff
@@ -26,3 +29,28 @@ pub fn multiaddr_to_url(multiaddr: &Multiaddr, api_path: &str) -> EyreResult<Url
...
+    let signature = keypair.sign(timestamp.as_bytes())?;
+
+    let mut builder = if request.is_some() {
+        client.post(url).json(&request)
+    } else {
+        client.get(url)
+    };
+
+    builder = builder
+        .header("X-Signature", hex::encode(signature))
```
miraclx: bs58?

crates/meroctl/src/cli/context/list.rs
```diff
@@ -25,7 +25,8 @@ impl ListCommand {
 
         let url = multiaddr_to_url(multiaddr, "admin-api/dev/contexts")?;
         let client = Client::new();
-        let response = client.get(url).send().await?;
+
+        let response = get_response::<()>(&client, url, None, &config.identity).await?;
```
miraclx:
```suggestion
        let response = get_response(&client, url, None::<()>, &config.identity).await?;
```

..._

crates/meroctl/src/common.rs
miraclx: _...

this reads like "get the response of a `T`", but it's not, how about the `impl Trait` syntax?

```suggestion
pub async fn get_response(
    client: &Client,
    url: Url,
    request: Option<impl Serialize>,
```

petarjuki7: I tried it like that but had trouble because None is not `Seralize`, this change: `None::<()>` should fix that, thanks.

crates/server/src/middleware/dev_auth.rs
```diff
@@ -0,0 +1,36 @@
...
+        .headers()
+        .get("X-Signature")
+        .and_then(|v| bs58::decode(v).into_vec().ok())
+        .ok_or(StatusCode::UNAUTHORIZED)?;
+
+    let timestamp = request
+        .headers()
+        .get("X-Timestamp")
+        .and_then(|v| Some(v.as_bytes()))
+        .ok_or(StatusCode::UNAUTHORIZED)?;
```
miraclx: I thought the use of timestamp here is to prevent replayability. There's no check for a timestamp threshold.

Maybe we permit a 5s threshold?

miraclx (commented):
rest lgtm

MatejVukosav (changes requested):
Update documentation and add in PR description how did u test it.

petarjuki7:
> Update documentation and add in PR description how did u test it.

Added new PR to docs

crates/server/src/admin/service.rs
```diff
@@ -35,8 +35,8 @@ use crate::admin::handlers::context::{
 use crate::admin::handlers::fetch_did::fetch_did_handler;
 use crate::admin::handlers::root_keys::{create_root_key_handler, delete_auth_keys_handler};
 use crate::config::ServerConfig;
+use crate::middleware;
 use crate::middleware::auth::AuthSignatureLayer;
-use crate::middleware::host::HostLayer;
```
miraclx, line 39:
```suggestion
#[cfg(feature = "host_layer")]
use crate::middleware::host::HostLayer;
```

PR #668, opened 2024-08-27
JWT Secret Management, Token Generation, and Refresh Token Functionality

alenmestrov:
**Overview:**
This PR introduces comprehensive functionality for JWT (JSON Web Token) management, including the generation and storage of a secure JWT key, the creation of access and refresh tokens, and the ability to refresh tokens securely. These enhancements ensure that the authentication process in the application is robust, secure, and easy to maintain.

**Key Features:**

1. **JWT Key Management:**
   - **Key Generation:** Implemented a mechanism to generate a new JWT key using a cryptographically secure random number generator if the key is not found in the store. This ensures that a unique, secure key is always available when the server starts.
   - **Key Storage and Retrieval:** The JWT key is stored in a persistent store using the `JwtTokenSecretEntry` structure. On startup, the server attempts to retrieve the secret; if it does not exist, a new secret is generated and stored, ensuring continuous availability of the secret for signing tokens.

2. **JWT Token Generation:**
   - **Access and Refresh Tokens:** Implemented functionality to generate JWT access tokens with a short lifespan and refresh tokens with a longer lifespan. These tokens are securely generated using the stored JWT secret and include claims necessary for validating the tokens' authenticity and intended use.
   - **Token Handling:** Added methods to insert and update refresh tokens in the store, ensuring that only valid tokens are used for authentication.

3. **Token Refresh Functionality:**
   - **Refreshing Tokens:** Introduced a feature to refresh JWT access tokens using a valid refresh token. This functionality checks the validity and expiration of the refresh token before issuing a new access token, ensuring secure and seamless token renewal.
   - **Validation and Error Handling:** Added robust validation to ensure that tokens are only refreshed if they are valid and not expired. Any errors during this process are handled gracefully, with clear error messages returned to the client.

4. **Error Handling Enhancements:**
   - **Graceful Degradation:** Improved error handling for scenarios where JWT keys or tokens cannot be retrieved or generated, ensuring that the application provides informative feedback and maintains security.
   - **Consistency in Response:** Standardized the error responses across all JWT-related functionalities, making it easier to diagnose and fix issues.

**Detailed Changes:**
- **`JwtTokenSecretEntry` and `JwtRefreshTokenEntry`:** Structures for handling the storage and retrieval of JWT keys and refresh tokens within the store.
- **`get_or_create_jwt_secret`:** Functionality to generate and insert a JWT secret into the store if it does not exist, ensuring the availability of a signing secret.
- **`generate_jwt_tokens`:** Function to generate and return both access and refresh tokens, using the securely stored JWT secret.
- **`refresh_access_token`:** Function to securely refresh the access token using a valid refresh token, ensuring continued access for authenticated clients.
- **API Endpoints:** 
  - **Token Generation Endpoint:** Added an endpoint to generate JWT tokens upon successful client authentication.
  - **Token Refresh Endpoint:** Added an endpoint to handle the refreshing of access tokens using a refresh token.

**Usage Example:**
- **Server Startup:** On server startup, the application checks for the existence of a JWT key, generating and storing a new one if necessary.
- **Token Generation:** When a client requests authentication, the server generates and returns both access and refresh tokens.
- **Token Refresh:** Clients can request a new access token using a valid refresh token, with the server handling the validation and issuance of the new token.

**Testing and Validation:**
- **Key and Token Generation:** Tested to ensure the JWT secret is generated, stored, and retrieved correctly, and that access and refresh tokens are generated securely.
- **Token Refresh:** Validated the token refresh functionality to ensure tokens are only refreshed when valid and within their expiration period.
- **Error Handling:** Simulated various failure scenarios to ensure the application handles errors gracefully and provides clear, actionable error messages.

https://github.com/user-attachments/assets/bc3189c3-03f2-43a1-b8aa-32a33ffb7e35


crates/server/src/admin/service.rs
```diff
@@ -118,6 +120,14 @@ pub(crate) fn setup(
             "/add-client-key",
             post(handlers::add_client_key::add_client_key_handler),
         )
+        .route(
+            "/refresh-jwt-token",
+            post(handlers::add_client_key::refresh_jwt_token_handler),
+        )
```
frdomovic, line 126: After you finish testing move this to protected routes as this will be called from the admin dashboard and it shouldn't be called from anywhere else

crates/server/src/admin/service.rs
```diff
@@ -118,6 +120,14 @@ pub(crate) fn setup(
...
+        .route(
+            "/generate-jwt-token",
+            post(handlers::add_client_key::generate_jwt_token_handler),
+        )
```
frdomovic: After you finish testing move this to protected routes as this will be called from the admin dashboard and it shouldn't be called from anywhere else

crates/server/src/admin/service.rs
```diff
@@ -118,6 +120,14 @@ pub(crate) fn setup(
             "/add-client-key",
             post(handlers::add_client_key::add_client_key_handler),
         )
+        .route(
+            "/refresh-jwt-token",
```
frdomovic: What happens if your send random values or valid jwt but not created in this node? 
Example we both running nodes and created a JWT. And I call refresh on your node with my JWT (I somehow know where your node is running)

alenmestrov: there is a JWT secret generated on server start that is unique (randomly generated) for each node so if you try to send JWT token (both access or refresh token) that wasn't created on that node you will get error message saying that JWT token couldn't be verified.

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -0,0 +1,76 @@
...
+// Method to get the JWT key from the store
+pub fn get_jwt_secret(store: Store) -> eyre::Result<Option<JwtTokenSecret>> {
+    let entry = JwtTokenSecretEntry::new();
+    let handle = store.handle();
+
+    match handle.get(&entry) {
+        Ok(Some(jwt_secret)) => Ok(Some(jwt_secret)),
+        Ok(None) => Ok(None),
+        Err(e) => Err(e.into()),
+    }
```
miraclx: in cases where you just want to transform the err, you can use:

```suggestion
    handle.get(&entry).map_err(Into::into)
```

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -0,0 +1,76 @@
...
+    let entry = JwtTokenSecretEntry::new();
+    let mut handle = store.handle();
+    match handle.put(&entry, &jwt_secret) {
+        Ok(_) => Ok(jwt_secret),
+        Err(e) => Err(e.into()),
+    }
+}
+
+// Method to get the JWT key from the store
+pub fn get_jwt_secret(store: Store) -> eyre::Result<Option<JwtTokenSecret>> {
```
miraclx:
```suggestion
pub fn get_jwt_secret(store: &Store) -> eyre::Result<Option<JwtTokenSecret>> {
```

crates/server/src/admin/utils/jwt.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    let context_id = req.context_id;
+    let executor_public_key = req.executor_public_key;
+    // Generate Access Token
+    let access_expiration = Utc::now() + Duration::hours(1);
+    let access_claims = Claims {
+        context_id: context_id.to_string(),
+        exec_pub_key: executor_public_key.to_string(),
+        exp: access_expiration.timestamp() as usize,
+        token_type: "access".to_string(),
```
miraclx:
```rust
#[derive(Serialize)]
#[serde(rename_all = "lowercase")]
enum TokenType {
    Access,
    Refresh
}
```

crates/server/src/admin/utils/jwt.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    };
+
+    let context_id = req.context_id;
+    let executor_public_key = req.executor_public_key;
+    // Generate Access Token
+    let access_expiration = Utc::now() + Duration::hours(1);
+    let access_claims = Claims {
+        context_id: context_id.to_string(),
+        exec_pub_key: executor_public_key.to_string(),
```
miraclx: `executor`, `identity`, `executor_public_key`

crates/server/src/admin/utils/jwt.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+        Err(err) => {
+            return Err(ApiError {
+                status_code: StatusCode::INTERNAL_SERVER_ERROR,
+                message: format!("Failed to get refresh token: {}", err),
+            });
+        }
+    };
+    // Check if the refresh token from the store matches the provided one
+    if refresh_token_db != refresh_token {
```
miraclx, line 191: in the case we have multiple clients, we'd want to be able to refresh multiple access tokens, so we can't store just one refresh token

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    fn key(&self) -> &Self::Key {
+        &self.key
+    }
+}
+
+impl JwtTokenSecretEntry {
+    fn new() -> Self {
+        Self {
+            key: Generic::new(*b"jwt_salt::server", [0; 32]),
```
miraclx, line 23: keep in mind you can store multiple entries under the fragment here, up to you to determine what's a good discriminant.

since the refresh token includes information about the executor and the expiry date, you can use that as the fragment specifier

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -0,0 +1,76 @@
...
+impl JwtTokenSecret {
+    pub fn jwt_secret(&self) -> &Vec<u8> {
+        &self.jwt_secret
+    }
+}
+
+// Method to generate a new JWT key
+fn generate_jwt_secret() -> Vec<u8> {
+    let mut rng = rand::thread_rng();
+    (0..32).map(|_| rng.gen()).collect()
```
miraclx:
```suggestion
fn generate_jwt_secret() -> [u8; 32] {
    rand::thread_rng().gen()
```

or

```suggestion
fn generate_jwt_secret() -> [u8; 32] {
    let mut rng = rand::thread_rng();
    let mut buf = [0; 32];
    rng.fill_bytes(&mut buf);
    buf
```

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn new() -> Self {
+        Self {
+            key: Generic::new(*b"jwt_salt::server", [0; 32]),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct JwtTokenSecret {
+    jwt_secret: Vec<u8>,
```
miraclx:
```suggestion
    jwt_secret: [u8; 32],
```

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl JwtTokenSecret {
+    pub fn jwt_secret(&self) -> &Vec<u8> {
```
miraclx:
```suggestion
    pub fn jwt_secret(&self) -> &[u8; 32] {
```

crates/server/src/admin/storage/jwt_token.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn new() -> Self {
+        Self {
+            key: Generic::new(*b"jwt_token:server", [0; 32]),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct JwtRefreshToken {
+    refresh_token: Vec<u8>,
```
miraclx, line 30: do we need to store the refresh token at all? we just need to ensure;

1. we can decode it from the user's request (so it was issued by us)
2. it has not expired

right?

MatejVukosav: Access token is short-lived token. Refresh token is used to obtain new access token

miraclx: Yeah, but do we need to store it?

MatejVukosav: You can use it to block some sessions. e.g. remove access from external location by deleting refresh key. Access token will be valid until expires but new one will not be issued

crates/server/src/admin/handlers/add_client_key.rs
```diff
@@ -72,16 +79,53 @@ pub async fn add_client_key_handler(
...
+            let data: String = "Client key stored".to_owned();
+            ApiResponse {
+                payload: AddClientKeyResponse { data },
+            }
+            .into_response()
+        })
+}
+
+//* Register client key to authenticate client requests  */
+pub async fn generate_jwt_token_handler(
```
MatejVukosav, line 110: return data:{} or error: {} parent objects

crates/server/src/admin/handlers/add_client_key.rs
```diff
@@ -94,6 +106,48 @@ pub async fn add_client_key_handler(
...
 }
 
+//* Register client key to authenticate client requests  */
+pub async fn generate_jwt_token_handler(
+    Extension(state): Extension<Arc<AdminState>>,
+    Json(req): Json<JwtTokenRequest>,
+) -> impl IntoResponse {
+    match generate_jwt_tokens(req, state.store.clone()) {
+        Ok(jwt_tokens) => {
+            let tokens = JwtTokens {
```
frdomovic, line 116: Can you add
```
#[derive(Debug, Deserialize, Serialize)]
struct JwtTokens {
access_token: String
refresh_token: String
}
```

To match return style that we have on other api endpoints. 
This is what i get on frontend and I would like it to be acessToken, refreshToken
Screenshot 2024-08-29 at 11 10 45

+do the same thing for refresh token

crates/server/src/admin/storage/jwt_secret.rs
```diff
@@ -25,34 +25,33 @@ impl JwtTokenSecretEntry {
...
+fn generate_jwt_secret() -> [u8; 32] {
+    rand::thread_rng().gen()
 }
 
 // Method to insert the JWT key if it doesn't exist
-pub fn get_or_create_jwt_secret(store: Store) -> eyre::Result<JwtTokenSecret> {
+pub fn get_or_create_jwt_secret(store: &Store) -> eyre::Result<JwtTokenSecret> {
     // Check if the key already exists
-    if let Some(existing_secret) = get_jwt_secret(store.clone())? {
+    if let Some(existing_secret) = get_jwt_secret(&store.clone())? {
```
miraclx:
```suggestion
    if let Some(existing_secret) = get_jwt_secret(&store)? {
```

crates/server/src/admin/utils/jwt.rs
```diff
@@ -1,5 +1,7 @@
 use std::str::from_utf8;
 
+use calimero_primitives::hash;
+// use calimero_primitives::hash;
```
miraclx: ^

crates/server/src/lib.rs
```diff
@@ -50,7 +50,7 @@ pub async fn start(
     let mut listeners = Vec::with_capacity(config.listen.len());
     let mut want_listeners = config.listen.into_iter().peekable();
 
-    if let Err(e) = get_or_create_jwt_secret(store.clone()) {
+    if let Err(e) = get_or_create_jwt_secret(&store.clone()) {
```
miraclx: ^^

```suggestion
    if let Err(e) = get_or_create_jwt_secret(&store) {
```

crates/server/src/admin/utils/jwt.rs
```diff
@@ -144,44 +195,18 @@ pub fn refresh_access_token(refresh_token: &str, store: Store) -> Result<JwtToke
...
-        });
-    }
-
-    let context_id = token_data.claims.context_id.clone();
-    let exec_pub_key = token_data.claims.exec_pub_key.clone();
-
     // Generate new Access Token
     let access_expiration = Utc::now() + Duration::hours(1);
     let access_claims = Claims {
         context_id: context_id.clone(),
```
miraclx: seems to me this can call `generate_jwt_tokens`, yeah?

crates/server/src/admin/handlers/add_client_key.rs
MatejVukosav: Every response should return either `data` or `error`. We have it like that in most endpoints except the first ones.

miraclx (approved):

PR #670, opened 2024-08-27
chore: add script to build all example apps

MatejVukosav:
Add script that will be called during tests

./scripts/build-all-apps.sh

```
./../apps/kv-store/build.sh executed successfully.
./../apps/only-peers/build.sh executed successfully.
./../apps/gen-ext/build.sh executed successfully.
```

### DIFF
--- a/scripts/build-all-apps.sh
+++ b/scripts/build-all-apps.sh
@@ -3,10 +3,12 @@
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
+cd "$(dirname $0)"
+
 BUILD_SCRIPTS=(
-    "./apps/gen-ext/build.sh"
-    "./apps/kv-store/build.sh"
-    "./apps/only-peers/build.sh"
+    "./../apps/gen-ext/build.sh"
+    "./../apps/kv-store/build.sh"
+    "./../apps/only-peers/build.sh"
 )
 
 run_script() {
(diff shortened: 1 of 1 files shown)


scripts/build-all-apps.sh
```diff
@@ -0,0 +1,34 @@
...
+        echo "$script executed successfully."
+    else
+        echo "Error: $script does not exist or is not executable."
+        exit 1
+    fi
+}
+
+# Iterate over each script in the array and run it in the background
+for script in "${BUILD_SCRIPTS[@]}"; do
+    run_script "$script" &
```
miraclx: overlapping stderr will make it tricky to determine which project caused an issue

MatejVukosav: What do you suggest then? To not build them in parallel?

MatejVukosav: I can also only show the result.

miraclx: > To not build them in parallel?

yeah

MatejVukosav: Now when I have removed logs and just show execution output it is easy to identify where was the issue. 

We should optimize overall execution time as much as possible.

miraclx: Hm, I don't think we get an edge building at the same time because they write to the same target folder, which gets locked for every build. What do you observe on your end? Does it go any faster?

miraclx (approved):

PR #671, opened 2024-08-27
chore: add test script

MatejVukosav:
Test:

chmod +x ./scripts/test.sh
./scripts/test.sh

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -3,11 +3,13 @@
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
+cd "$(dirname $0)"
+
 # prepare apps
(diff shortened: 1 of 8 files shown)


scripts/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Exit immediately if a command exits with a non-zero status.
+set -e
+
+# prepare apps
+./scripts/build-all-apps.sh
```
miraclx: This assumes we're in the root to execute this, yeah?

let's `cd "$(dirname $0)"`, so we can then call `./build-all-apps`

miraclx (approved):

PR #673, opened 2024-08-27
chore:  extend husky precheck

MatejVukosav:
Execute different commands on pre-commit based on changed files.

To enable husky run `pnpm install` from the root

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,10 @@
 pnpm format:md
+
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# Check for changes in the 'logic' directory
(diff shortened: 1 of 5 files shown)


.husky/pre-commit
```diff
@@ -1 +1,28 @@
...
+echo "Running pre-commit hook..."
+
+if git diff --cached  --name-only | grep -qE '\.mdx?$'; then
+  echo "Markdown or MDX files have been changed."
+  pnpm format:md
+fi
+
+
+# Check for change in Rust code
+if git diff --cached  --name-only | grep -vqE '^(node-ui/|packages/calimero-sdk/)'; then
```
miraclx: why not match for files ending in `.rs`?

miraclx (approved):

Issue #571, opened 2024-08-08
Secure Login Flow - Implement functionality in Calimero SDK for user login through admin dashboard utilizing JWT tokens


MatejVukosav:
Calimero SDK needs few new updated functionalities.
As the regular login is removed new login function is needed.
This login function should take in "applicationId" as parameter and then redirect the application to Admin Dashboard for the login confirmation.
The redirected url should get applicationId and callback url as query params.

Also its needed to update handling the login response (JWT tokens - access + refresh) - saving it to browser storage.

Third function is to add function to fetch saved JWT and append it to JSON rpc requests which are already in the SDK.

PR #677, opened 2024-08-28
feat(sdk): new client login

frdomovic:
# feat(sdk): new client login

## Summary
Components for new client login.
1. Main updates are in setup component where user needs to enter application id (previous context id). 
-> this application id is used to fetch context list and later on context identities which will be saved in JWT.
2. Updated the setup modal to use inline CSS because styled components cause issues.
source:
Screenshot 2024-08-28 at 15 09 13
3. Added login modal which just takes application id and redirects to admin dashboard + saves JWT on success return and redirect to logged in page.

Fixes # [Part of Issue](https://github.com/calimero-network/core/issues/571)

## Test plan
* Full test video when everything else is merged

https://github.com/user-attachments/assets/1896ff20-7550-4f5a-9489-a317c6648453


alenmestrov (approved):

packages/calimero-sdk/src/login/ClientLogin.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const callbackUrl = encodeURIComponent(window.location.href);
+    const redirectUrl = `${nodeUrl}/admin-dashboard/?application_id=${applicationId}&callback_url=${callbackUrl}`;
+
+    window.location.href = redirectUrl;
+  };
+
+  useEffect(() => {
+    const urlParams = new URLSearchParams(window.location.search);
+    const access_token = urlParams.get('access_token');
+    const refresh_token = urlParams.get('refresh_token');
```
frdomovic:
```suggestion
    const encodedAccesstToken = urlParams.get('access_token');
     const accessToken = dencodeURIComponent(encodedAccesstToken);
```

PR #680, opened 2024-08-30
feat: context config

miraclx:
Add context configuration contract

We decided to temporarily store strongly consistent data like application versions & member lists on-chain while we focus on eventual consistency for private data.

Features:

- Anybody can create a context so far as they can prove ownership of the private key that derived public key represented as the context ID.
	- This means signing the context creation payload. This payload includes a timestamp to mitigate replay attacks, the permitted threshold is configurable only by the contract account.
- Granular permissions, accounts can granted permissions to update the application or manage the member list (invite & remove).
- Paginated member list fetching.


xilosada:
missing test.sh

contracts/context-config/src/repr.rs
```diff
@@ -0,0 +1,145 @@
+use std::{fmt, ops::Deref};
+
+use near_sdk::{bs58, near};
+use thiserror::Error;
+
+#[derive(Eq, Ord, Copy, Clone, Debug, PartialEq, PartialOrd)]
+#[near(serializers = [borsh, json])]
+#[serde(transparent)]
+pub struct Repr<T> {
```
MatejVukosav, line 13: What does this mean? Is this "Representation"?

miraclx: `Repr` makes generic bytes representable over the wire, encoding them as `bs58`.

miraclx:
> missing test.sh

the one we have for package manager is broken, and has been for months, so I guess nobody uses it

but anyway, `cargo test` is all that's needed here, not sure a script that just calls `cargo test` is that useful, wdyt?

contracts/context-config/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+mod types;
+
+pub use repr::{Repr, ReprBytes};
+pub use types::{
+    ApplicationId, ApplicationSource, BlobId, ContextId, ContextIdentity, Guard, SignedPayload,
+};
+
+#[derive(Debug)]
+#[near(contract_state)]
+pub struct ContextConfig {
```
MatejVukosav: Structs hierarchy is confusing. Lets discuss it today

xilosada:
@miraclx then let's remove the file (in another PR) and ensure the tests run in the ci

contracts/context-config/src/lib.rs
MatejVukosav: `ContextsConfigs` or something similar to indicate more configurations

miraclx:
> and ensure the tests run in the ci

the tests are running

MatejVukosav (approved):

Issue #514, opened 2024-07-18
NEAR login refactor


MatejVukosav:
We should redesign wallet integrations as currently login with NEAR exposes the possibility of user losing access to his node if his signing key changes.

Quick fix is to ping near and check if new key is in the list and add it to root keys. 

Long term solution is to re-think usage of root keys and consider wallets as verifiable credentials instead of having function to unblock the node.

PR #683, opened 2024-09-02
fix: NEAR Login Refactor

alenmestrov:
### Description:

This PR implements a fix for issue [#514](https://github.com/calimero-network/core/issues/514), which addresses the problem of users potentially losing access to their nodes when their signing key changes.

#### Changes Made:
- **Key Validation**: The code now checks if a new signing key exists in the user's key list. If it does, it lets user through ensuring continuous access.

#### Testing:
- **Manual Testing**: Tested with various key changes to ensure that users retain access to their nodes after updating keys.

### Related Issues:
- [#514](https://github.com/calimero-network/core/issues/514) - NEAR Login Refactor

### How to test:
- Go to `packages/calimero-sdk/` and run `pnpm run build`
- Change in `/node-ui/package.json` version of `@calimero-is-near/calimero-p2p-sdk` to `../packages/calimero-sdk` so that it takes the latest changes applied in calimero-sdk.  Once these changes are merged it will automatically build a new version of calimero-sdk so that it corresponds to the version in `node-ui`
- Delete node_modules in `node-ui`, run `pnpm install`, and then `pnpm run dev`

https://github.com/user-attachments/assets/ba0a4f9d-a45c-42eb-9711-9e09c86ba7ff


packages/calimero-sdk/src/auth/headers.ts
```diff
@@ -30,9 +30,8 @@ export async function createAuthHeader(
   const signature = await privateKey.sign(hashArray);
   const signatureBase58 = bs58.encode(signature);
   const contentBase58 = bs58.encode(hashArray);
-
   const headers: Header = {
-    wallet_type: JSON.stringify(WalletType.NEAR),
+    wallet_type: JSON.stringify(WalletType.NEAR({ networkId: 'mainnet' })),
```
xilosada: why mainnet is hardcoded? The network should be passed to the sdk

xilosada:
I'm confused with the new flow. Why do we have to iterate over the stored keys? Ideally the public key is passed and verified in the node.

packages/calimero-sdk/src/auth/headers.ts
alenmestrov: yes, I'm doing that now.

alenmestrov:
Currently, when the user logs in for the first time his key is stored inside a node. If he wants to add more keys he needs to add them through UI. If the user already has an initial root key and wants to log in with a new wallet the application returns that the root key is missing and doesn't let him through.
In the case of NEAR login if the user changes device or browser and already has a root key he will not be able to log in since the new key is generated. That is why we are iterating through stored NEAR keys to see if they belong to the user's wallet.

packages/calimero-sdk/src/auth/headers.ts
```diff
@@ -11,7 +11,7 @@ export interface Header {
 
 export async function createAuthHeader(
   payload: string,
-  contextId: string | null = null,
+  nearNetworkType: string,
```
MatejVukosav: networkId ?

MatejVukosav (commented):
Why is this just a temp solution? What is the problem?

crates/server/src/verifywalletsignatures/near.rs
```diff
@@ -52,6 +95,66 @@ pub fn verify_near_signature(
     verify(public_key_str, &payload_hash, signature_base64).is_ok()
 }
 
+pub async fn check_for_near_account_key(
```
MatejVukosav: has_near_key(account, key..)

Or similar. Check indicates void and not response.

packages/calimero-sdk/src/auth/headers.ts
```diff
@@ -30,13 +30,11 @@ export async function createAuthHeader(
...
   const signatureBase58 = bs58.encode(signature);
   const contentBase58 = bs58.encode(hashArray);
-
   const headers: Header = {
-    wallet_type: JSON.stringify(WalletType.NEAR),
+    wallet_type: JSON.stringify(WalletType.NEAR({ networkId: networkId })),
     signing_key: signingKey,
     signature: signatureBase58,
     challenge: contentBase58,
-    context_id: contextId,
```
MatejVukosav, line 39: Why we don't use context anymore?

crates/server/src/admin/handlers/add_client_key.rs
```diff
@@ -164,20 +164,35 @@ pub fn store_client_key(
...
+        };
+
         let _ = store_root_key(
             req.wallet_metadata.verifying_key.clone(),
             req.wallet_metadata.wallet_type.clone(),
+            wallet_address,
             store,
-        )?;
+        )
+        .map_err(|err| {
```
MatejVukosav: )?

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                        return Err(ApiError {
+                            status_code: StatusCode::BAD_REQUEST,
+                            message: "Wallet address not present".to_string(),
+                        });
+                    }
+                };
+
+                // Get network_type and from it use correct rpc_url
+                let rpc_url = match network_id {
+                    NearNetworkId::Mainnet => "https://rpc.mainnet.near.org",
```
MatejVukosav: just open an issue to extract constants

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                if !check_for_near_account_key(
+                    &req.wallet_metadata.verifying_key,
+                    wallet_address,
+                    rpc_url,
+                )
+                .await?
+                {
+                    return Err(ApiError {
+                        status_code: StatusCode::BAD_REQUEST,
+                        message: "Given public key is not from wallet".into(),
```
MatejVukosav: Provided public key does not belong to account {#?}

crates/server/src/verifywalletsignatures/near.rs
```diff
@@ -52,6 +95,66 @@ pub fn verify_near_signature(
...
+    let client = Client::new();
+    // Check if root key belongs to the account
+    let body = json!({
+        "jsonrpc": "2.0",
+        "id": "dontcare",
+        "method": "query",
+        "params": {
+            "request_type": "view_access_key",
+            "finality": "final",
+            "account_id": account_name,
```
MatejVukosav: unify naming

MatejVukosav (commented):
resolve comments

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
     drop(match root_key_result {
         Some(root_key) => root_key,
         None => {
+            if let WalletType::NEAR { network_id } = &req.wallet_metadata.wallet_type {
```
danwilliams: To avoid this big code tree, it feels like this would benefit from being turned into a `let...else` with an early exit from the logic?

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                let wallet_address = match req.wallet_metadata.wallet_address.as_deref() {
+                    Some(address) => address,
+                    None => {
+                        return Err(ApiError {
```
danwilliams: For all the places here that `return`, it would be more idiomatic to simply return them implicitly, and then use e.g. `?` at the end of the match assignment.

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                    }
+                    Ok(keys) => keys,
+                    Err(err) => {
+                        info!("Error checking if near client key exists: {}", err);
+                        return Err(ApiError {
+                            status_code: StatusCode::INTERNAL_SERVER_ERROR,
+                            message: err.to_string(),
+                        });
+                    }
+                };
```
danwilliams: This might be more readable as a `map`... not sure.

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                    true => {
+                        let _ = store_root_key(
+                            req.wallet_metadata.verifying_key.clone(),
+                            req.wallet_metadata.wallet_type.clone(),
+                            wallet_address.to_string(),
+                            store,
+                        )
+                        .map_err(|err| {
+                            return err;
+                        })?;
```
danwilliams: This `map_err` is unnecessary - the `return` returns from the map closure, so that's the same thing as just doing `err`, and as `err` was passed in, nothing changes 🙂 You can just remove it, and keep the `?`.

crates/server/src/verifywalletsignatures/near.rs
```diff
@@ -7,8 +7,51 @@ use base64::engine::Engine;
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ResultDataWithPermission {
+    block_hash: String,
+    block_height: u64,
+    nonce: Option<u64>,
+    permission: Option<Permission>,
+    error: Option<String>, // Result-level error
+}
```
danwilliams, line 60: I think we should be adding Rustdocs to everything new, and this is a good example of where we would definitely benefit from the documentation! 😄

crates/server/src/verifywalletsignatures/near.rs
```diff
@@ -52,6 +95,66 @@ pub fn verify_near_signature(
...
+        "jsonrpc": "2.0",
+        "id": "dontcare",
+        "method": "query",
+        "params": {
+            "request_type": "view_access_key",
+            "finality": "final",
+            "account_id": account_name,
+            "public_key": current_near_root_key,
+        }
+    });
```
danwilliams: Out of interest, it might be nice to be formalising places like this into structs...

crates/server/src/verifywalletsignatures/near.rs
```diff
@@ -52,6 +95,66 @@ pub fn verify_near_signature(
...
+        .json::<NearJsonRpcResponse<ResultDataWithPermission>>()
+        .await
+        .map_err(|e| ApiError {
+            status_code: StatusCode::BAD_REQUEST,
+            message: format!("Failed to parse response: {}", e),
+        })?;
+
+    // Check if there is a top-level error
+    if let Some(ref error) = response.error {
+        println!("Top-level error found: {:?}", error);
```
danwilliams: There are a few `println`s around... should we be using logging here instead?

danwilliams (changes requested):
I can't speak to the functionality, but I found some approach nits that it would be good to sort out 🙂 Nothing major!

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                        status_code: StatusCode::BAD_REQUEST,
+                        message: format!(
+                            "Provided public key does not belong to account {:?}",
+                            wallet_address
+                        )
+                        .into(),
+                    });
+                }
+                // Check if the wallet_address has a NEAR account key from DB
+                match has_near_key(&near_keys, wallet_address, rpc_url).await? {
```
danwilliams: Matching on a boolean is not generally considered idiomatic, and may be clearer as an `if` 🙂

crates/server/src/admin/utils/auth.rs
```diff
@@ -283,6 +284,93 @@ pub fn validate_root_key_exists(
...
+                        let _ = store_root_key(
+                            req.wallet_metadata.verifying_key.clone(),
+                            req.wallet_metadata.wallet_type.clone(),
+                            wallet_address.to_string(),
+                            store,
+                        )
+                        .map_err(|err| {
+                            return err;
+                        })?;
+                        return Ok(req);
```
danwilliams: This `return` may not be necessary...

danwilliams (approved):
Nice work! 🙂 👍

PR #688, opened 2024-09-04
feat(cli): Added jsonrpc funcionality to cli

petarjuki7:
# CLI now supports jsonrpc commands

## Summary

With this new funcionality you can call functions on applications (contexts) installed on your node
Will add in documentation the usage of the CLI.

```
meroctl  --node-name node jsonrpc 
    --method <query or mutate> 
    --context-id <context_id> 
    --method get_user_info 
    --args-json '{"user_id": "12345"}' (defaults to {})
```

## Test plan
I installed on two nodes and a coordinator the app from [https://github.com/calimero-network/core-app-template/tree/master/logic](url). The only thing I changed is that the function `increase_count` always increases by `1`.

I run on node2 (the first non coordinator node):
`cargo run -p meroctl -- --node-name node2 json-rpc --call-type query --context-id EDwcrKJ9uhywwjFYssNmLzzJaj3AfDty6SxG9Pi2Juab --method get_count`

The response is:
`{"jsonrpc":"2.0","id":"dontcare","result":{"output":0}}`

I run two times:
`cargo run -p meroctl -- --node-name node2 json-rpc --call-type mutate --context-id EDwcrKJ9uhywwjFYssNmLzzJaj3AfDty6SxG9Pi2Juab --method increase_count`

I run again the same get_count method from before, 
response is: `{"jsonrpc":"2.0","id":"dontcare","result":{"output":10}}`


crates/meroctl/src/cli.rs
```diff
@@ -54,6 +56,8 @@ pub enum SubCommands {
     Run(RunCommand),
     Context(ContextCommand),
     App(AppCommand),
+    #[command(alias = "call")]
+    JsonRpc(JsonRpcCommand),
```
danwilliams: Anywhere we add stuff to lists, ideally we should do so in alphabetical order 🙂 Makes things much nicer over time.

crates/meroctl/src/cli/jsonrpc.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    #[arg(long)]
+    pub method: String,
+
+    #[arg(long, value_parser = value_parser!(Value), default_value = "{}")]
+    pub args_json: Value,
+
+    #[arg(long, default_value = "dontcare")]
+    pub id: String,
+}
```
danwilliams: I feel like this is a prime contender to have some documentation! 😄

crates/server-primitives/src/admin.rs
```diff
@@ -1,13 +1,23 @@
...
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use url::Url;
 
+#[derive(Debug)]
+pub struct AdminState {
+    pub store: Store,
+    pub keypair: Keypair,
+    pub ctx_manager: ContextManager,
+}
```
danwilliams: I think this should be marked as `#[non_exhaustive]`?

danwilliams (dismissed):
Looks good but some housekeeping we can do 🙂

crates/meroctl/src/cli.rs
MatejVukosav: https://github.com/calimero-network/core/pull/696#discussion_r1746859497

MatejVukosav:
@danwilliams can u unblock this PR?

crates/server-primitives/Cargo.toml
```diff
@@ -9,14 +9,17 @@ license.workspace = true
 [dependencies]
 camino.workspace = true
 eyre.workspace = true
+libp2p.workspace = true
```
miraclx: you can use `libp2p_identity` directly here, since this is a primitive, it's best to keep it light and the `libp2p` crate will introduce a lot more than just identity

crates/server-primitives/Cargo.toml
```diff
@@ -9,14 +9,17 @@ license.workspace = true
...
 semver = { workspace = true, features = ["serde"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 thiserror.workspace = true
 url = { workspace = true, features = ["serde"] }
 
+calimero-context = { path = "../context" }
 calimero-node-primitives = { path = "../node-primitives" }
 calimero-primitives = { path = "../primitives" }
+calimero-store = { path = "../store", features = ["serde"] }
```
miraclx: `calimero-store` depends on `rocksdb`, we don't want that anywhere near primitive crates ..._

crates/server-primitives/src/admin.rs
```diff
@@ -1,13 +1,34 @@
...
 use serde_json::Value;
 use url::Url;
 
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct AdminState {
+    pub store: Store,
+    pub keypair: Keypair,
+    pub ctx_manager: ContextManager,
+}
```
miraclx: _... This will only be used by `calimero-server`, it should be there instead.

Primitive crates are supposed to host structures/behaviour that aid interaction across crates.

danwilliams:
> @danwilliams can u unblock this PR?

Happy for it to be approved one @miraclx 's changes have been done.

miraclx (approved):
lgtm

crates/server-primitives/src/jsonrpc.rs
```diff
@@ -139,7 +139,7 @@ impl QueryRequest {
             context_id,
             method,
             args_json,
-            executor_public_key,
+            executor_public_key: executor_public_key.into(),
```
miraclx: require a `PublicKey` from the caller instead. change the params

PR #691, opened 2024-09-05
chore(context-config): reorganize

miraclx:
Decouple the context config contract and the library we can reuse in the node & Other QOL changes.


crates/server/src/admin/service.rs
```diff
@@ -39,6 +39,8 @@ use crate::admin::handlers::root_keys::{create_root_key_handler, delete_auth_key
 use crate::config::ServerConfig;
 use crate::middleware;
 use crate::middleware::auth::AuthSignatureLayer;
+#[cfg(feature = "host_layer")]
```
MatejVukosav: This should be separate PR

crates/context/config/src/types.rs
```diff
@@ -233,13 +290,44 @@ impl<T: serde::Serialize> Signed<T> {
...
+    }
+}
+
+impl<T, E> IntoResult<T> for Result<T, E> {
+    type Error = E;
+
+    fn into_result(self) -> Result<T, Self::Error> {
+        self
+    }
+}
```
danwilliams, line 314: This seems a bit... strange. And without documentation, I can't work out what the motivation is. All it does is wrap something in a `Result`... Hmmmm.

crates/context/config/src/types.rs
```diff
@@ -188,15 +214,43 @@ impl ReprBytes for Signature {
...
+
+    fn from_bytes<F>(f: F) -> repr::Result<Self, Self::Error>
+    where
+        F: FnOnce(&mut Self::DecodeBytes) -> bs58::decode::Result<usize>,
+    {
+        use VerificationKeyParseError::{InvalidVerificationKey, LengthMismatch};
+
+        let bytes = Self::DecodeBytes::from_bytes(f).map_err(|e| e.map(LengthMismatch))?;
+
+        Self::from_bytes(&bytes).map_err(|e| repr::Error::DecodeError(InvalidVerificationKey(e)))
```
danwilliams, line 243: Nice to see the careful error-handling 👍

crates/context/config/src/types.rs
```diff
@@ -160,8 +175,19 @@ impl ApplicationSource<'_> {
...
+    Ord,
+    Debug,
+    Default,
+    Clone,
+    PartialEq,
+    PartialOrd,
+    Serialize,
+    Deserialize,
+    BorshSerialize,
+    BorshDeserialize,
```
danwilliams, line 189: It would be nice if we could keep our `derive`s in alphabetical order - I went through and very carefully ordered them all a couple of weeks ago 😅

crates/context/config/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+    },
+    RemoveMembers {
+        members: Cow<'a, [Repr<ContextIdentity>]>,
+    },
+    Grant {
+        capabilities: Cow<'a, [(Repr<ContextIdentity>, Capability)]>,
+    },
+    Revoke {
+        capabilities: Cow<'a, [(Repr<ContextIdentity>, Capability)]>,
+    },
```
danwilliams, line 83: Where we have thing like this, it's so much nicer to maintain if they are kept in order 🙂

crates/context/config/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    pub signer_id: Repr<SignerId>,
+    pub timestamp_ms: Timestamp,
+}
+
+impl<'a> Request<'a> {
+    pub fn new(signer_id: SignerId, kind: RequestKind<'a>) -> Self {
+        let timestamp_ms = time::SystemTime::now()
+            .duration_since(time::UNIX_EPOCH)
+            .expect("system time is before epoch?")
```
danwilliams, line 31: Technically this isn't the only reason it could be rejected.

danwilliams (changes requested):
Looks mostly okay, but I found a couple of attention-to-detail points about keeping things in order, and also it would be good to be adding docs 🙂

crates/context/config/src/types.rs
miraclx: this is used in `Signed`, as a helper for the signing & verification processes, permitting a function to fail safely if it can as opposed to panicking, for example, while all byte slices are valid scalars in curve25519 to make up a `SigningKey`, not all of them are valid points in the curve to make up a `VerifyingKey`, but if you already had a `VerifyingKey` you won't have any errors in the key derivation process.

This is an accommodation for flexibility, if you have a result, return it, if not, return what you have and not have to worry about `rustc` asking for an `Err` type since we were too restrictive on the type system.

crates/context/config/src/lib.rs
miraclx: I find it more intuitive for related things to stay together

crates/context/config/src/lib.rs
miraclx: why else would it be rejected?

crates/context/config/src/types.rs
miraclx: I think there's more value to visibility & readability for derives, and as such, I prefer it to have a visual aid in the form of it ascending length-wise, with related traits grouped together.

crates/context/config/src/lib.rs
danwilliams: It could overflow.

miraclx: Hm, this isn't `chrono` btw, this is stdlib and it doesn't seem to

https://github.com/rust-lang/rust/blob/c2f74c3f928aeb503f15b4e9ef5778e77f3058b8/library/std/src/time.rs#L540-L542

https://github.com/rust-lang/rust/blob/c2f74c3f928aeb503f15b4e9ef5778e77f3058b8/library/std/src/sys/pal/unix/time.rs#L45-L47

https://github.com/rust-lang/rust/blob/c2f74c3f928aeb503f15b4e9ef5778e77f3058b8/library/std/src/sys/pal/unix/time.rs#L135-L166

crates/server/src/admin/service.rs
miraclx: https://github.com/calimero-network/core/pull/712

MatejVukosav (approved):

crates/context/config/src/types.rs
danwilliams: I didn't spot this before as it was closed, but just to say, the ordering by length is something I find highly weird... I am not sure how that helps code clarity at all.

danwilliams (approved):

PR #692, opened 2024-09-05
feat: intermediate stage for context config use in the node

miraclx:
This patch contains the initial iterations of the node update following #680, during which I noticed a number of things weren't what they should be and went ahead to fix them:

- Create context no longer takes a context ID, instead it takes a seed for a CSPRNG from which a context ID is derived.
	- The derived context ID is returned from the creation.
	- This impacts the CLI, for example as you can no longer pass `--context-id`, but now `--seed`
- `calimero_identity::identity` didn't do anything, it's been removed
- `blobstore` no longer keeps blocks in the the `apps` dir within the node's data dir, it's now, more appropriately `blobs`.
- Dropped `Keypair` in favor of `PrivateKey(Hash)` and `PublicKey(Hash)`
- Changed the storage format for the `ContextIdentity` entry, we're no longer storing the removed `Keypair` and rather, just the `PrivateKey`.
- The config file has the following changes:

	<table>
	<td>
	
	```toml
   [store]
   path = "data"
   
   [application]
   path = "apps"
	```
	
	</td>
	<td> -> </td>	
	<td>
	
	```toml
   [datastore]
   path = "data"
   
   [blobstore]
   path = "blobs"
   
   [context]
   relayer = "https://relayer.calimero.org/"
	```
	
	</td>
	</table>

- Server API changes:
	- create context takes a `context_seed` instead of `context_id`
	- join context returns just the context identity of the new user, no more the full context metadata


MatejVukosav (changes requested):
Please check `core-app-template` and `only-peers` as this can potentially break apps. 

Update docs so they reflect new way of context creation: https://calimero-network.github.io/build/dev-mode - check other places as well

Also, can you provide the link for the GH issue we discussed?

crates/server/src/admin/service.rs
```diff
@@ -121,7 +121,7 @@ pub(crate) fn setup(
         .route("/refresh-jwt-token", post(refresh_jwt_token_handler))
         .route("/generate-jwt-token", post(generate_jwt_token_handler))
         .layer(AuthSignatureLayer::new(store))
-        .layer(Extension(Arc::clone(&shared_state)));
+        .layer(Extension(shared_state.clone()));
```
danwilliams: I'm not sure why this change has been made? Is the `shared_state` no longer an `Arc`? It _should_ be wrapped in an `Arc` in order to be safe (and I'm pretty sure Axum requires it) so why the change? Note, using `.clone()` on reference-counted types is a code smell.

crates/context/src/config.rs
```diff
@@ -1,15 +1,7 @@
...
-
-impl ApplicationConfig {
-    #[must_use]
-    pub const fn new(dir: Utf8PathBuf) -> Self {
-        Self { dir }
-    }
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ContextConfig {
+    pub relayer: Url,
 }
```
danwilliams, line 7: No docs...

danwilliams (changes requested):
This looks good, and I only have a couple of nits. However, I feel the commits are not very useful - it would be nice to see more meaningful commit messages - and also, Rustdocs 😉

crates/context/src/config.rs
MatejVukosav: Can you provide an example of docs for this? Although I agree we need to document stuff, I am not sure what do you expect in this case

crates/server/src/admin/service.rs
miraclx: It's still an `Arc`, just thought we didn't need the verbose syntax. But, fair point. I'll change it back.

crates/context/src/config.rs
miraclx: There's no time for this, and these pieces are under active development (iterated on in successive PRs), documenting every iteration in code will only slow us down further, and I'm out next week.

crates/server/src/admin/service.rs
miraclx: Resolved

danwilliams (approved):

MatejVukosav:
> Please check `core-app-template` and `only-peers` as this can potentially break apps.
> 
> Update docs so it includes new way of context creation: https://calimero-network.github.io/build/dev-mode - check other places as well
> 
> Also, can you provide the link for the GH issue we discussed?

Fix for core-app-template https://github.com/calimero-network/core-app-template/pull/17

MatejVukosav:
> Member

Fix for only peers
https://github.com/calimero-network/only-peers-client/pull/73

PR #693, opened 2024-09-05
chore: remove `calimero-identity`

miraclx:
`calimero-identity` doesn't make sense as a crate, merged into `calimero-server`'s `admin` module


crates/server/src/admin/utils/auth.rs
```diff
@@ -292,3 +294,55 @@ pub fn validate_root_key_exists(
 
     Ok(req)
 }
+
+pub fn verify_near_public_key(public_key: &str, msg: &[u8], signature: &[u8]) -> EyreResult<bool> {
+    let public_key = bs58::decode(public_key)
+        .into_vec()
+        .map_err(|_| eyre!("Invalid public key: Base58 encoding error"))?;
+
+    let public_key = PublicKey::try_decode_protobuf(&public_key)
```
miraclx, line 440: why are we using protobuf?

MatejVukosav: I don't remember :) It was long time ago, but it worked

MatejVukosav (approved):

crates/server/src/admin/utils/auth.rs
```diff
@@ -292,3 +294,55 @@ pub fn validate_root_key_exists(
...
+    let message_hash = eth_message(message);
+    let recovery_id = i32::from(signature_bytes[64]).saturating_sub(27_i32);
+
+    // Attempt to recover the public key, returning false if recovery fails
+    match recover(&message_hash, &signature_bytes[..64], recovery_id) {
+        Ok(pubkey) => {
+            // Format the recovered public key as a hex string
+            let pubkey_hex = format!("{pubkey:02X?}");
+            // Compare the recovered public key with the account address in a case-insensitive manner
+            let result = account.eq_ignore_ascii_case(&pubkey_hex);
```
danwilliams, line 465: Always nice to see _useful_ explanatory comments! 👍

crates/server/src/admin/utils/auth.rs
```diff
@@ -292,3 +294,55 @@ pub fn validate_root_key_exists(
...
+}
+
+pub fn eth_message(message: &str) -> [u8; 32] {
+    keccak256(
+        format!(
+            "{}{}{}",
+            "\x19Ethereum Signed Message:\n",
+            message.len(),
+            message
+        )
```
danwilliams: This feels a bit icky. The use of a string to put inside a string... mmmm. Would it not be better as:

```rust
        format!("\x19Ethereum Signed Message:\n{}{message}", message.len())
```

danwilliams (approved):
One minor optional nit I spotted, otherwise seems good 🙂

crates/server/src/admin/utils/auth.rs
miraclx: these pieces we're moved, not introduced, but good point

crates/server/src/admin/utils/auth.rs
miraclx: Understood, and left as-is, if I understand correctly, those pieces will be succeeded by the JWT flow.

crates/server/src/admin/utils/auth.rs
miraclx: Resolved

crates/server/src/admin/utils/auth.rs
MatejVukosav: Not really. We are still using near as root key

PR #694, opened 2024-09-05
feat(context-config): add base client w/ relayer impl

miraclx:
Adds a foundational library for interacting with the context config contract à la #680.

This also introduces a relayer implementation. NEAR to come.

### DIFF
--- a/crates/context/config/src/client.rs
+++ b/crates/context/config/src/client.rs
@@ -223,7 +223,7 @@ impl<T: Transport> ClientRequest<'_, '_, T> {
 }
 
 impl<T: Transport> ContextConfigMutateClient<'_, T> {
-    pub async fn add_context<'a>(
+    pub fn add_context<'a>(
         &self,
         context_id: ContextId,
         author_id: ContextIdentity,
@@ -240,7 +240,7 @@ impl<T: Transport> ContextConfigMutateClient<'_, T> {
         ClientRequest { client: self, kind }
     }
 
-    pub async fn update_application<'a>(
+    pub fn update_application<'a>(
         &self,
         context_id: ContextId,
         application: Application<'a>,
@@ -253,7 +253,7 @@ impl<T: Transport> ContextConfigMutateClient<'_, T> {
         ClientRequest { client: self, kind }
     }
 
-    pub async fn add_members(
+    pub fn add_members(
         &self,
         context_id: ContextId,
         members: &[ContextIdentity],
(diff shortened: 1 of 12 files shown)


crates/context/config/src/client.rs
```diff
@@ -0,0 +1,323 @@
...
+        payload: Vec<u8>,
+    ) -> Result<Vec<u8>, Self::Error>;
+}
+
+#[derive(Debug)]
+pub struct TransportRequest<'a> {
+    pub network_id: Cow<'a, str>,
+    pub contract_id: Cow<'a, str>,
+    pub operation: Operation<'a>,
+}
```
danwilliams, line 33: It would be nice to see Rustdocs added to new things 🙂

crates/context/config/src/client.rs
```diff
@@ -0,0 +1,323 @@
...
+        offset: usize,
+        length: usize,
+    ) -> Result<Response<Vec<Repr<ContextIdentity>>>, Error<T>> {
+        self.read(
+            "members",
+            json!({
+                "context_id": Repr::new(context_id),
+                "offset": offset,
+                "length": length,
+            }),
```
danwilliams, line 163: I am curious... it feels like, in places like this, we would benefit from more formalisation. I.e. using structs, which are then converted to JSON as and when needed. That would, I think, be best/standard practice, and allow us to rely upon Rust's type system...? However, I'm sure there is a good reason why it's like this, and I likely don't know the whole context! 😄 Thoughts?

danwilliams (changes requested):
This is pretty straightforward, and all much the same in terms of what is being done. I have two overall observations:

  1. I feel we should consider formalising the data we're using JSON for, by using structs, and converting those to JSON where and when necessary.
  2. Lack of docs... 😉

crates/context/config/src/client.rs
miraclx: We're the only consumers of these at this time, with no near-term plans of publishing them to docs.rs, and these pieces are under active development (maybe iterated on in successive PRs), documenting every iteration in code will only slow us down further, and I'm out next week.

crates/context/config/src/client.rs
miraclx: Unlike the accommodations made for `mutate` and `set`, these query methods are transparently what `near_sdk`'s codegen produces, which maintains the structures these are deserialized to privately.

We could introduce a struct, but it would be single-use and private, at this point, there's no added benefit, with `json!` being more straightforward.

danwilliams (approved):

Issue #593, opened 2024-08-16
Application operations


petarjuki7:
Enable all endpoints currently available in the `server` regarding applications to be used through the `cli`

Estimated time: 2 days

Issue #585, opened 2024-08-12
Enable app install from the URL


MatejVukosav:
Support app installation so user can create context by taking app URL from admin dashboard. 

`meroctl app install –url <url>`

PR #695, opened 2024-09-05
Added all application functions to the cli

petarjuki7:
# Application control through the cli

## Summary

All apis regarding applications are now available through the cli, including installing an app with an URL
Solves #593 and #585 

## Usage

`cargo run -p meroctl -- --node-name node2 app get --method details --app-id <app-id>
`
`cargo run -p meroctl -- --node-name node2 app install --url <url> --metadata <metadata>`

### DIFF
--- a/crates/meroctl/src/cli/app.rs
+++ b/crates/meroctl/src/cli/app.rs
@@ -2,9 +2,11 @@ use clap::{Parser, Subcommand};
 use eyre::Result as EyreResult;
 
 use super::RootArgs;
+use crate::cli::app::get::GetCommand;
 use crate::cli::app::install::InstallCommand;
 use crate::cli::app::list::ListCommand;
 
+mod get;
 mod install;
 mod list;
 
@@ -19,6 +21,7 @@ pub enum AppSubCommands {
     Install(InstallCommand),
     #[command(alias = "ls")]
     List(ListCommand),
+    Get(GetCommand),
 }
 
 impl AppCommand {
@@ -26,6 +29,7 @@ impl AppCommand {
         match self.subcommand {
             AppSubCommands::Install(install) => install.run(args).await,
             AppSubCommands::List(list) => list.run(args).await,
+            AppSubCommands::Get(get) => get.run(args).await,
         }
     }
 }
--- a/crates/meroctl/src/cli/app.rs
+++ b/crates/meroctl/src/cli/app.rs
@@ -18,18 +18,18 @@ pub struct AppCommand {
 
 #[derive(Debug, Subcommand)]
 pub enum AppSubCommands {
+    Get(GetCommand),
     Install(InstallCommand),
     #[command(alias = "ls")]
     List(ListCommand),
-    Get(GetCommand),
 }
 
 impl AppCommand {
     pub async fn run(self, args: RootArgs) -> EyreResult<()> {
         match self.subcommand {
+            AppSubCommands::Get(get) => get.run(args).await,
             AppSubCommands::Install(install) => install.run(args).await,
             AppSubCommands::List(list) => list.run(args).await,
-            AppSubCommands::Get(get) => get.run(args).await,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


crates/meroctl/src/cli/app.rs
```diff
@@ -19,13 +21,15 @@ pub enum AppSubCommands {
...
     List(ListCommand),
+    Get(GetCommand),
 }
 
 impl AppCommand {
     pub async fn run(self, args: RootArgs) -> EyreResult<()> {
         match self.subcommand {
             AppSubCommands::Install(install) => install.run(args).await,
             AppSubCommands::List(list) => list.run(args).await,
+            AppSubCommands::Get(get) => get.run(args).await,
```
danwilliams: Nit: If we keep these things in alphabetical order as a general rule it helps overall and over time as things grow 🙂

crates/meroctl/src/cli/app/get.rs
```diff
@@ -0,0 +1,56 @@
...
+            bail!("Config file does not exist")
+        };
+
+        let Ok(config) = ConfigFile::load(&path) else {
+            bail!("Failed to load config file")
+        };
+
+        let Some(multiaddr) = config.network.server.listen.first() else {
+            bail!("No address.")
+        };
```
danwilliams, line 36: This is nice, good to see careful checking and informative output 👍

crates/meroctl/src/cli/app/get.rs
```diff
@@ -0,0 +1,56 @@
...
+        )?;
+
+        let response = get_response(&client, url, None::<()>, &config.identity).await?;
+
+        if !response.status().is_success() {
+            bail!("Request failed with status: {}", response.status())
+        }
+
+        let text = response.text().await?;
+        println!("{}", text);
```
danwilliams: Not important, but these two lines could be merged, i.e.

```rust
println!("{}", response.text().await?);
```

crates/meroctl/src/cli/app/get.rs
```diff
@@ -0,0 +1,56 @@
...
+    #[arg(long, short)]
+    pub app_id: String,
+}
+#[derive(ValueEnum, Debug, Clone)]
+pub enum GetValues {
+    Details,
+}
+
+impl GetCommand {
+    pub async fn run(self, args: RootArgs) -> EyreResult<()> {
```
danwilliams, line 23: This is not specific to this PR, but I am curious as to why we use Eyre in some places and have more formalised errors in others. I think in this place it kinda makes sense as they are one-off errors. Just thinking out loud...

crates/meroctl/src/cli/app/install.rs
```diff
@@ -13,13 +17,21 @@ use crate::config_file::ConfigFile;
...
 pub struct InstallCommand {
     /// Path to the application
-    #[arg(long, short)]
-    pub path: Utf8PathBuf,
+    #[arg(long, short, conflicts_with = "url")]
+    pub path: Option<Utf8PathBuf>,
+
+    /// Url of the application
+    #[clap(long, short, conflicts_with = "path", requires = "metadata")]
+    pub url: Option<String>,
```
danwilliams, line 25: Lovely to see some Rustdocs being written! 😄 👍 I'm hoping that it can happen everywhere... it would be good to add them to everything new that gets added 🙂

danwilliams (approved):
This looks good to me, I had a couple of very minor observations, but nothing blocking.

crates/meroctl/src/cli/app/install.rs
miraclx: > Lovely to see some Rustdocs being written!

these aren't just docs for rust docs though, as they serve dual-purpose and it's the second that pretty much mandates their presence, given that they're visible to the users of the CLI

Ideally, we prioritize documenting all public interfaces and work our way in over time.

crates/meroctl/src/cli/app/get.rs
miraclx: > I am curious as to why we use Eyre in some places and have more formalised errors in others

ideally, we formalize errors in all libs and use eyre in bins. for now, eyre is an escape-hatch, as error-handling can get pretty involved and we don't care much for pattern-matching what happened, we mostly just report it.

> we formalize errors in all libs and use eyre in bins

this formalization will come eventually

PR #696, opened 2024-09-06
feat(context-config): add NEAR transport impl

miraclx:
Adds a NEAR transport implementation in preparation for the relayer binary


crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+use url::Url;
+
+use super::{Operation, Transport, TransportRequest};
+
+#[derive(Debug)]
+pub struct NetworkConfig {
+    pub rpc_url: Url,
+    pub account_id: AccountId,
+    pub access_key: SecretKey,
+}
```
danwilliams, line 33: New stuff, we should be documenting... in this location most things are fairly self-explanatory, but still good to get into the habit? 🙂

crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+
+#[derive(Debug)]
+pub struct NearConfig<'a> {
+    pub networks: BTreeMap<Cow<'a, str>, NetworkConfig>,
```
danwilliams, line 37: Out of interest, what is the motivation for a copy-on-write string here? I am curious, because if it's a key it's unlikely to change? Perhaps we should make it a different type? Same with the other CoWs.

Note, this is an excellent example of where documentation would really help 😉

crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+    UnknownNetwork(String),
+    #[error("invalid response from RPC while {operation}")]
+    InvalidResponse { operation: String },
+    #[error("invalid contract ID: {0}")]
+    InvalidContractId(near_primitives::account::id::ParseAccountError),
+    #[error("failed while {operation}: {reason}")]
+    Custom {
+        operation: &'static str,
+        reason: String,
+    },
```
danwilliams: Minor nit - it would be useful to keep variants in alphabetical order so that as they grow they are easier to find and manage 🙂

crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+                    account_id: contract_id,
+                    method_name: method,
+                    args: FunctionArgs::from(args),
+                },
+            })
+            .await
+            .map_err(|err| Error::Custom {
+                operation: "querying contract",
+                reason: err.to_string(),
+            })?;
```
danwilliams: Do you think these could/should be formalised further here?

crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    async fn mutate(
+        &self,
+        contract_id: AccountId,
+        method: String,
+        args: Vec<u8>,
+    ) -> Result<Vec<u8>, Error> {
+        let response = self
+            .client
+            .call(methods::query::RpcQueryRequest {
```
danwilliams: There are a few naming path prefixes that it feels it would be good to drop throughout this function. I.e. to use `use`.

crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+            method_names,
+        } = permission
+        {
+            if receiver_id != contract_id {
+                return Err(Error::Custom {
+                    operation: "mutating contract",
+                    reason: format!(
+                        "access key does not have permission to call contract: {}",
+                        contract_id
+                    ),
```
danwilliams: It seems to me that a reason such as this is better formalised into an error enum variant...? Same with the others?

crates/context/config/src/client/near.rs
```diff
@@ -0,0 +1,287 @@
...
+            }),
+        }
+    }
+
+    async fn mutate(
+        &self,
+        contract_id: AccountId,
+        method: String,
+        args: Vec<u8>,
+    ) -> Result<Vec<u8>, Error> {
```
danwilliams, line 174: I am not overly fond of this function as it's quite long, and I wonder if it would benefit from being split up? There are a number of distinct stages and areas that could lend themselves well to being individual functions.

danwilliams (changes requested):
I think the error-handling could be improved, and a couple of places would benefit from slight reorganisation. Otherwise, the main thing is that we need docs! 😄

crates/context/config/src/client/near.rs
MatejVukosav: That should be the job of the formatting tool. It will be available soon https://github.com/rust-lang/rustfmt/pull/6313

crates/context/config/src/client/near.rs
miraclx: There's just no time for this, not as part of the development effort, which is why everything is self-explanatory. We'll revisit at a later time when everything comes together and works.

crates/context/config/src/client/near.rs
miraclx: `NearConfig` is a public struct, of which this `Cow` is a network ID, we don't keep this in the `NearTransport` that consumes the config, all we're saying is, give me a map between potential references to a string and a `NetworkConfig` in which case, different constructors could either have string slices or owned strings.

- `meroctl` will have owned strings, as a product of the config file deserialization
- the relayer, on the other hand, has static string slices

crates/context/config/src/client/near.rs
miraclx: We're the only consumers of this, and we're only calling `.to_string` anyway without pattern matching, so it's only valuable for the error's string representation to convey all information than for it to be conditionally matched over.

crates/context/config/src/client/near.rs
miraclx: Extracted out nonce determination

crates/context/config/src/client/near.rs
miraclx: These can and have been, the others.. not quite

crates/context/config/src/client/near.rs
miraclx: Resolved

danwilliams (approved):

PR #697, opened 2024-09-06
feat(context-config): add context config params to node config

miraclx:
generate random implicit accounts on node init, these are single-purpose accounts and are immediately functional once the node operator funds it with a small amount

this is provided as an alternative to the relayer, should they choose not to use it (at the user's convenience), it'll also help with testing before the relayer impl is complete & deployed

<table>

<thead>

<th>

old (at least since https://github.com/calimero-network/core/pull/692)

</th>

<th>
now
</th>

</thead>

<td>

```toml
[context]
relayer = "https://relayer.calimero.org/"
```

</td>

<td>

```toml
[context.config.new]
network = "testnet"
contract_id = "calimero-context-config.testnet"

[context.config.signer]
use = "relayer"

[context.config.signer.relayer]
url = "https://relayer.calimero.org/"

[context.config.signer.self.mainnet]
rpc_url = "https://rpc.mainnet.near.org/"
account_id = "a8102cc5a27e75168f8042492470e1c8b074c42f41fe2265b3d030bdf0fded6d"
public_key = "ed25519:CK3oJkMwBSe1BzpG5Tyw6a2CebHm5j8EuG1nm4AqefDv"
secret_key = "ed25519:5Lmw3BjqRXjP54DNYoptMogyCoufUfSoZercMQWG9PeYDAaXjonj1FdeqdYgN9n96oR8iny2fWvpZNEKiN5wbDNC"

[context.config.signer.self.testnet]
rpc_url = "https://rpc.testnet.near.org/"
account_id = "fc73352fccee37a5b0da5df0d461d5248bb721f758a6464e774a431ed15d3681"
public_key = "ed25519:HzTdYsSD52VnEL9Bpc2tmTNQatorpREZeSyeDeMaZBda"
secret_key = "ed25519:4cCu4U6Zsdzw8K4eXoqnP8RDBjocviYa4igpPhHCvbE9XHdDT5pgPNDCZ3T4tkpJ1nWvo1iBu73HKPNJDrovxSzk"

```

</td>

</table>

### DIFF
--- a/crates/context/config/src/client/config.rs
+++ b/crates/context/config/src/client/config.rs
@@ -6,34 +6,34 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct ContextConfigConfig {
-    pub signer: ContextConfigSigner,
+pub struct ContextConfigClientConfig {
+    pub signer: ContextConfigClientSigner,
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct ContextConfigSigner {
+pub struct ContextConfigClientSigner {
     #[serde(rename = "use")]
-    pub selected: ContextConfigSelectedSigner,
-    pub relayer: ContextConfigRelayerSigner,
+    pub selected: ContextConfigClientSelectedSigner,
+    pub relayer: ContextConfigClientRelayerSigner,
     #[serde(rename = "self")]
-    pub local: BTreeMap<String, ContextConfigLocalSigner>,
+    pub local: BTreeMap<String, ContextConfigClientLocalSigner>,
 }
 
 #[derive(Copy, Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "lowercase")]
-pub enum ContextConfigSelectedSigner {
+pub enum ContextConfigClientSelectedSigner {
     Relayer,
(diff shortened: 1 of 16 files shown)


crates/context/config/src/client/config.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct ContextConfigClientSigner {
+    #[serde(rename = "use")]
+    pub selected: ContextConfigClientSelectedSigner,
+    pub relayer: ContextConfigClientRelayerSigner,
+    #[serde(rename = "self")]
+    pub local: BTreeMap<String, ContextConfigClientLocalSigner>,
+}
```
danwilliams, line 27: It would be great to start adding in Rustdocs to everything new we add 🙂

danwilliams (changes requested):
This looks straightforward to me - I don't know what all of it does, but what I do know about looks sensible. My only nit is that we surely should be adding documentation now, at the very least to the new things we add in?

crates/context/config/src/client/config.rs
miraclx: There's no time for this, which is why the PR description includes a preview of what this does.

MatejVukosav (approved):

danwilliams (approved):

PR #700, opened 2024-09-09
Feat: Storage baseline

danwilliams:
# Purpose

This is the first of a series of PRs bringing in the storage and synchronisation functionality. This one creates the `storage` crate and adds the baseline of addressing in the storage system: through `Id` and `Path.

# Description of changes

  - Added new `storage` crate with README
  - Added base of storage `Id` and `Path` structs
      - Added a new `address` module to the `storage` crate.
      - Added base implementations of `Id` and `Path`.
  - Added additional functionality to storage `Path`
      - Added a range of additional foundation functions that will be useful to `Path` in the address module of the storage crate

# Notes and advisories


crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use calimero_storage::address::Path;
+    /// Path::new("::root::node::leaf").unwrap();
+    /// ```
+    ///
+    pub fn new<S: Into<String>>(path: S) -> Result<Self, PathError> {
+        let string = path.into();
```
miraclx: looks to me this can be a `&str`, avoiding the allocation

```suggestion
    pub fn new<S: AsRef<str>>(path: S) -> Result<Self, PathError> {
        let string = path.as_ref();
```

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+
+        if string.is_empty() {
+            return Err(PathError::Empty);
+        }
+        if !string.starts_with("::") {
+            return Err(PathError::NotAbsolute);
```
miraclx, line 135: is there any value to a non-absolute path? Is that why we have to accommodate the concept of an "absolute" one? if not, then perhaps we don't need the forward `::`?

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+        }
+
+        #[allow(clippy::string_slice)] // We know the string starts with `::`
+        let segments = string[2..].split("::").collect::<Vec<&str>>();
+
+        if segments.is_empty() {
```
miraclx, line 141: allocation here is avoidable

```suggestion
        let segments = string[2..].split("::").peekable();

        if segments.peek().is_none() {
```

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+#[derive(Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub struct Path {
+    /// A list of path segment offsets, where offset `0` is assumed, and not
+    /// stored.
+    offsets: Vec<u8>,
+
+    /// The path to the element. This is a string of up to 255 characters in
+    /// length, and is case-sensitive. Internally the segments are stored
+    /// without the separators.
+    path: Flexstr<255>,
```
miraclx: I think this is meant to be 256, holding up to 255 items since the first item is the length

miraclx: Hm, why are we using the `Flexstr` struct anyway, as the basic functionality we need is super simple (`buf: [u8; 255]`, `len: u8` & `str::from_utf8`), and most (~ 97%) of the value to the crate is unrealized here. As we're not pushing to the `Path` after it's been created, and we explicitly guard against a heap allocation as anything past 255 items is forbidden. Are there plans to expand it's use or that's about it? Just checking

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+                return Err(PathError::EmptySegment);
+            }
+            if str.len().saturating_add(segment.len()) > 255 {
+                return Err(PathError::Overflow);
+            }
+            if str.len() > 0 {
+                #[allow(clippy::cast_possible_truncation)] // Can't occur here
+                offsets.push(str.len() as u8);
+            }
+            let _: bool = str.push_str(segment);
```
miraclx, line 159: is the saved allocation of the input string in the `Path` worth the absence of `.as_str()` / efficiency of `Display`?

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+        }
+        let last_offset = 0_usize;
+
+        for &offset in &self.offsets {
+            #[allow(clippy::cast_sign_loss)] // Can't occur here
+            #[allow(trivial_numeric_casts)] // Not harmful here
+            if self.path[last_offset..offset as usize] != other.path[last_offset..offset as usize] {
+                return false;
+            }
+        }
```
miraclx: do you want to update `last_offset`?

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// ```rust
+    /// use calimero_storage::address::Path;
+    /// let path = Path::new("::root::node::leaf").unwrap();
+    /// assert_eq!(path.parent().unwrap().to_string(), "::root::node");
+    /// assert_eq!(Path::new("::root").unwrap().parent(), None);
+    /// ```
+    ///
+    #[must_use]
+    pub fn parent(&self) -> Option<Self> {
+        let mut clone = self.clone();
```
miraclx: Cloning all of `Self` is typically not the right thing to do when you don't need all of it as it will reallocate all of `Self`'s heap-backed capacity, which in certain cases won't be necessary.

Like going from `::a::b::c::d::e::f` (which would've allocated a `Vec` capable of holding 8 offsets) to `::a::b::c::d::e`, (whose offsets would fit comfortably in one capable of holding 4), you'll still have a `Vec` with capacity for 8.

It's fairly trivial, but from my understanding of how this module is written, it seems like optimization is a key piece, in which case, I figured I'd mention.

crates/storage/src/address.rs
```diff
@@ -0,0 +1,480 @@
...
+    ///
+    /// ```rust
+    /// use calimero_storage::address::Path;
+    /// let path = Path::new("::root::node::leaf").unwrap();
+    /// assert_eq!(path.segments(), vec!["root", "node", "leaf"]);
+    /// ```
+    ///
+    #[must_use]
+    pub fn segments(&self) -> Vec<&str> {
+        (0..=self.depth()).filter_map(|i| self.segment(i)).collect()
```
miraclx: how about return an iterator here and leave it to the caller to consume

```suggestion
    pub fn segments(&self) -> impl Iterator<Item = &str> {
        (0..=self.depth()).filter_map(|i| self.segment(i))
```

crates/storage/src/address.rs
danwilliams: Good question 🙂 Yes, there is some value - it's part of planned future functionality. Therefore, starting out enforcing that all paths are absolute at present means that we preserve this without concern.

Note by the way that this makes no difference to the size of the stored `Path` data.

crates/storage/src/address.rs
danwilliams: That's a good suggestion - I agree 👍

crates/storage/src/address.rs
danwilliams: That's a really good suggestion - peeking to see if there's a next element at the beginning instead of allocating is more efficient. Thanks! 🙂

crates/storage/src/address.rs
danwilliams: No, it is meant to be 255 🙂

danwilliams: Good question! Yes, it means we have an efficient `String`-like type to handle these things efficiently, so it prevents rolling some stuff ourselves, plus the wider crate gives us nice things like `fstr` which is extremely useful.

crates/storage/src/address.rs
danwilliams: You've lost me there 😅

crates/storage/src/address.rs
danwilliams: Yeah that's a good catch!

crates/storage/src/address.rs
danwilliams: That's insightful. I don't think the `clone()` is all that bad, as it needs to duplicate all but the last segment - but, it is technically unnecessary as we have the offsets and so could use them for the calculation 🤔

crates/storage/src/address.rs
danwilliams: Yeah that's not a bad shout, given that most use cases will involve iteration, and if we actually want a `Vec` we can just `collect()`, so nothing lost 👍

crates/storage/src/address.rs
miraclx: so you intend to store up to 254 items then? keep in mind, `Flexstr<N>` where N < 256 is a `tstr<N>` which is a `[u8; N]` with the first item representing the length. So `Flexstr<255>` will store 254 items.

crates/storage/src/address.rs
miraclx: so, from how this is defined, you will never be able to get a `&str` from a `Path`, even if you passed one in, but instead will always have to get back a heap-allocated `String`, so then the question becomes, is it perhaps more valuable to simply store that heap-allocated `String` in `Path` and keep the offsets, with which we can give out slices

also, since you're using the `String` representation of a `Path` to serialize it, that's another cost to factor in

crates/storage/src/address.rs
miraclx: > Good question 🙂 Yes, there is some value - it's part of planned future functionality. Therefore, starting out enforcing that all paths are absolute at present means that we preserve this without concern.

Understood, I'll keep an eye out for that

> Note by the way that this makes no difference to the size of the stored Path data.

Yes, that part is clear, just checking to make sure I understood the concept of an absolute path seeing as I didn't find any references to the concept of a relative one, so maybe the leading colons weren't necessary, but the clarification of future use is decent

danwilliams: Cool - just a note, it's in the docs, maybe that was missed 🙂 

```rust
/// Path segments are separated by a double-colon `::`, and all paths are
/// absolute, and should start with a leading separator to enforce the clarity
/// of this plus allow for future expansion of functionality.
```

crates/storage/src/address.rs
danwilliams: I've ended up leaving this as it was - although peeking would save an allocation, it doesn't actually help because we don't know the length, and so we end up with a worse allocation of the `Vec` later, or a double iteration to find the length. That speed trade-off is worse than the minor extra storage, and so I think it's fine as it is.

crates/storage/src/address.rs
danwilliams: Oh, I see what you mean now. Apologies, I had thought that `Flexstr` dealt with the internal length itself and that the `255` was the usable content length. You are correct - it should be 256. Nice catch!

crates/storage/src/address.rs
danwilliams: Okay, so having read through that a couple of times, I think I know what you mean.

Since I'm storing the segments in a `Flexstr`, and with a custom format, with the offsets separately in a `Vec<u8>`, I can't hand out borrowed slices (`&str`) directly from the `Path`. This means that every time we want to get the path or parts of it, we'd likely need to generate a `String` (heap-allocated) rather than simply handing out a reference to part of the underlying data (a `&str` slice). Am I right?

If so, I think that's acceptable, and is in line with the design criteria. The way I thought of it, the typical use case is using the whole `Path`, and we want to maintain efficiency. Segment manipulation is a less-common occurrence, and so the trade-off there should be okay.

Plus. I plan in future to adjust the serialisation to store the offsets and stripped path, but for now that's unnecessary work plus it's safer to round-trip it as the full string version for validation. But later, that will mean efficiency gains in terms of storage and skipping parsing.

As to whether after that it would be better to do it this way or a way that uses slices on a `String`, I am not sure. That feels like a premature micro-optimisation, which I'm against... we could also implement our own slicing if we really wanted to.

crates/storage/src/address.rs
miraclx:
```rust
let segments = string[2..].split("::").collect::<Vec<&str>>();
```

you only know the length because of the first allocation, at which point you didn't know the length, so you still have the "worse allocation of the `Vec`"

so you're still incurring the speed depreciation, and still having extra storage

worth noting that a `Vec<&str>`[ will also allocate 16x more memory](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1686d88b5458241995fe2b9d4c631ae5) than a `Vec<u8>`, though it's unnecessary here

crates/storage/src/address.rs
miraclx: > Segment manipulation is a less common occurrence, and so the trade-off there should be okay.

so from what I can see, there are two main uses for the `Path` - storing, and traversal. for traversal, you need segments (`&str`) and for storing (get/set) you need the full path (which is a string)

I don't think you would compromise segmentation if you also accommodated for the storing phase since you would still be slicing anyway, as you currently are

but at least you would no longer need the one-off heap allocation for the string representation of the `Path` during the encoding phase.

for efficiency gains, I'm not sure why the current model wins over the simpler one if while we have a `Path`, traversal should cost the same, and for storing, the current model would cost more since we're only dealing with one-off heap allocations

crates/storage/src/address.rs
danwilliams: I am not sure of the comparison to a `Vec<u8>`, as that's not something that is comparable here?

The best way to do this would be to manually loop over the string and work with slices, but I'm not interested in making micro-optimisations at this time... it is a distraction, and in an area that is insignificant. Hence I consider this "good enough", and certainly not bad.

crates/storage/src/address.rs
danwilliams: My approach to things like this is generally as follows:

  1. Plan out an initial design of the functionality
  2. Choose the most appropriate types, data structuring, and algorithmic approach to give a good foundation with known wins and close-to-hand functionality to call on
  3. Get the thing working
  4. Later on, as an when necessary, conduct benchmarks to show where areas might need to be optimised

We're working with Rust, it's fast, it's efficient, and we get to build on a lot of good stuff. I care about performance or I'd not be using Rust, but equally I am pragmatic and prefer not to pre-optimise. There is a law of diminishing returns, plus there is a risk of spending significant time micro-focusing only to then have the requirements change and find that time has been wasted.

My goal is to implement and iterate quickly, whilst maintaining good practices and approaches, and not to worry too much about things that don't matter until they matter 🙂 

There are a number of good reasons for the current design pattern, including the things yet to build on, and I don't want to code myself into a corner and have to undo things that might overfit the current shape. My time on this is limited, and to me this area of bikeshedding is in the "doesn't matter right now" category.

crates/storage/src/address.rs
miraclx: > I am not sure of the comparison to a Vec<u8>, as that's not something that is comparable here?

you need a `Vec<u8>`, which stores the offsets, but the current approach temporarily allocates a `Vec<&str>` to do that, which from what's been highlighted is unnecessary since the allocation of `Vec<&str>` incurs the same cost you mention you wanted to avoid

I don't think it's a micro-optimization, far from it, but certainly a non-blocker, approved

miraclx (approved):

PR #703, opened 2024-09-09
Feat/storage integration

danwilliams:
# Purpose

This is the fourth of a series of PRs bringing in the storage and synchronisation functionality. This one integrates the storage types with the store, and builds on the previous https://github.com/calimero-network/core/pull/702

# Description of changes

  - Added a `Storage` key to the store
      - Added a new `storage` module to `store::key`, with `Storage` and `StorageId` structs, and all of the implementations necessary to use these for storing the storage ID types.
  - Implemented (de)serialisation
      - Made `Element`/`Data`/`Metadata` (de)serialisable with Borsh.
      - Added manual Borsh (de)serialisation implementations for `Id` and `Path`.
  - Integrated `storage` types with `store` types
      - Added conversions between `Id` and `StorageKey` and related types.

# Notes and advisories

  - The previous PR https://github.com/calimero-network/core/pull/702 needs to merge to master first, then the target of this PR can change.

### DIFF
--- a/crates/storage/src/address.rs
+++ b/crates/storage/src/address.rs
@@ -484,11 +484,6 @@ impl Path {
 }
 
 impl BorshDeserialize for Path {
-    fn deserialize(buf: &mut &[u8]) -> Result<Self, IoError> {
-        Self::new(&String::deserialize(buf)?)
-            .map_err(|err| IoError::new(IoErrorKind::InvalidData, err))
-    }
-
     fn deserialize_reader<R: Read>(reader: &mut R) -> Result<Self, IoError> {
         Self::new(&String::deserialize_reader(reader)?)
             .map_err(|err| IoError::new(IoErrorKind::InvalidData, err))
--- a/crates/storage/src/address.rs
+++ b/crates/storage/src/address.rs
@@ -10,7 +10,9 @@
 mod tests;
 
 use core::fmt::{self, Debug, Display, Formatter};
+use std::io::{Error as IoError, ErrorKind as IoErrorKind, Read, Write};
 
+use borsh::{BorshDeserialize, BorshSerialize};
 use fixedstr::Flexstr;
 use thiserror::Error as ThisError;
 use uuid::{Bytes, Uuid};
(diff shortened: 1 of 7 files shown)


crates/store/src/key/storage.rs
miraclx, line 1: This is already accommodated for with `key::ContextState`, the difference then seems to be you need 16 bytes, instead of 32, you can change that.

These storage keys are also not scoped out by the context ID, so there's no partitioning.

crates/storage/src/address.rs
```diff
@@ -55,6 +58,34 @@ impl Id {
...
+                IoErrorKind::UnexpectedEof,
+                "Not enough bytes to deserialize Id",
+            ));
+        }
+        let (bytes, rest) = buf.split_at(16);
+        *buf = rest;
+        Ok(Self(Uuid::from_slice(bytes).map_err(|err| {
+            IoError::new(IoErrorKind::InvalidData, err)
+        })?))
+    }
```
miraclx, line 74: this can be omitted, the default implementation of deserialize will delegate to `deserialize_reader` which will return EOF since you're using `read_exact`

also, `Uuid::from_slice` does a length check internally

crates/store/src/key/storage.rs
danwilliams: I must admit the storage keys are not overly clear as to exactly what to do here, so your input is very much valued 👍 

However, in this case this is a semantic purpose of being used as a key for storage. That key might change in underlying type, and indeed is already doing so in changes I'm working on further down the line. I therefore don't think the `ContextState` would be appropriate to (re)use...?

I am not sure what you mean about the partitioning?

crates/storage/src/address.rs
danwilliams: Ah, that's good to know - I was unaware of that 🙂 I've not used Borsh before.

crates/store/src/key/storage.rs
miraclx: > I am not sure what you mean about the partitioning?

Contexts's state should be isolated from one another, which is why `ContextState` has a `ContextId` prefix before the `StateKey`

Given that this is supposed to merge to master while still WIP for iterative development, I think this is fine as-is

But once we're ready to port the current model to the CRDT, this will have to be resolved

`ContextState` uses the dedicated `State` column family, which is the appropriate thing to use, while this uses `Generic`, which is okay in the meantime while this is incomplete, but since these (the uuids) are random `[u8; 16]`, we have to keep in mind that the server stores DID, certificates etc inside the `Generic` key under specific keys, which must not be overridden

So for now, you can temporarily introduce a new column family while this is experimental and hasn't already replaced `ContextState`.

So when it's ready, we'll switch over to `ContextState` and update it as appropriate

danwilliams: > I think this is fine as-is

Okay, that's cool 🙂 I'm sure it will likely have further changes and refinements as things progress. For instance, as discussed, it won't be staying as a `[u8; 16]` but will be growing.

It sounds like there will be some kind of migration that you are implying, which is to be expected, and I agree that is likely the best time to resolve this question.

Does this mean that this PR should be approved now then?

miraclx (approved):

Issue #188, opened 2024-04-23
Task Manager Cache


chefsale:
Implement a task manager cache to optimize WASM runtime loading

PR #704, opened 2024-09-09
Feat: Storage persistence

danwilliams:
# Purpose

This is the fifth of a series of PRs bringing in the storage and synchronisation functionality. This one brings save and load functionality, and builds on the previous https://github.com/calimero-network/core/pull/703

# Description of changes

  - Added `create_test_store()`
      - Added a common test module for shared test functionality, and added a `create_test_store()` function to it to create a `Store` for testing purposes.
  - Added save and load functionality to storage interface
      - Added a `store` property to storage `Interface`, and updated the `new()` constructor to accept it.
      - Added initial basic implementations for `save()` and `find_by_id()`.

# Notes and advisories

  - The previous PR https://github.com/calimero-network/core/pull/703 needs to merge to master first, then the target of this PR can change.


crates/storage/src/interface.rs
```diff
@@ -44,8 +64,31 @@ impl Interface {
...
     ///
-    pub fn find_by_id(&self, _id: Id) -> Result<Option<Element>, StorageError> {
-        unimplemented!()
+    #[allow(clippy::significant_drop_tightening)]
+    pub fn find_by_id(&self, id: Id) -> Result<Option<Element>, StorageError> {
+        // TODO: It seems fairly bizarre/unexpected that the put() method is sync
+        // TODO: and not async. The reasons and intentions need checking here, in
+        // TODO: case this find() method should be async and wrap the blocking call
+        // TODO: with spawn_blocking(). However, this is not straightforward at
+        // TODO: present because Slice uses Rc internally for some reason.
```
miraclx, line 73: rocksdb doesn't expose an async interface, so to keep it simple, we didn't either

that, plus, async traits were only recently stabilized, and still lack proper support for dynamic dispatch, easy mitigation would've been to use the `async_trait` crate, but that comes with it's own issues

so, we kept it simple

as for the `Rc` bit, a mitigation to that would've made it in https://github.com/calimero-network/core/pull/519 ([src](https://github.com/calimero-network/core/compare/455fe09ca9be09df17046584a3ef6cd28564e01a~...f74f4f48e39a8ad2556881c4a0f599fba346f766)) but there was no time for the ripple effect of code change required, it's now planned at a later date.

crates/storage/src/interface.rs
```diff
@@ -13,21 +13,41 @@
...
 use crate::entities::Element;
 
 /// The primary interface for the storage system.
-#[derive(Debug, Default, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
-pub struct Interface;
+pub struct Interface {
+    /// The backing store to use for the storage interface.
+    store: Arc<RwLock<Store>>,
```
miraclx, line 36: `Store` is itself an `Arc<_>` https://github.com/calimero-network/core/blob/1db5a74d5afd88d294c10a5b758cdae383db0cbd/crates/store/src/lib.rs#L22-L24

and it doesn't need a mutual exclusion lock as none of the `Database` methods [do either](https://github.com/calimero-network/core/blob/1db5a74d5afd88d294c10a5b758cdae383db0cbd/crates/store/src/db.rs#L29-L46)

crates/storage/src/interface.rs
```diff
@@ -102,11 +145,23 @@ impl Interface {
...
+    ///
+    pub fn save(&self, id: Id, element: &Element) -> Result<(), StorageError> {
+        // TODO: It seems fairly bizarre/unexpected that the put() method is sync
+        // TODO: and not async. The reasons and intentions need checking here, in
+        // TODO: case this save() method should be async and wrap the blocking call
+        // TODO: with spawn_blocking(). However, this is not straightforward at
+        // TODO: present because Slice uses Rc internally for some reason.
+        self.store
+            .write()
+            .put(
```
miraclx, line 159: are we losing atomicity?

crates/storage/src/interface.rs
danwilliams: Okay, that makes sense. Just a couple of thoughts, though:

  - Although RocksDB is sync, we may want to consider making our calls to it async, to avoid blocking generally, as this will become a scalability concern.
  - Async traits stabilised fairly recently (end of last year from memory...?) but we've always had the functionality; we just had to roll the futures by hand or use the `async_trait` crate. So the lack of stabilisation in Rust core didn't really prevent anything here.
  - We could trivially change the `Rc` to an `Arc` couldn't we...?

I am wondering, do you think we should change the `Rc` to an `Arc` to allow `async` here, which then avoids thread blocking?

crates/storage/src/interface.rs
danwilliams: Ahhhh yes, thanks for pointing that out! That's a nice simplification to make 😄

crates/storage/src/interface.rs
danwilliams: How? 🤔

crates/storage/src/interface.rs
danwilliams: Unfortunately I can't make this change, and in coming to do it I remember now trying to do this before. The reason it's in an `Arc<RwLock>` is because the `Store` demands self-mut, which I don't want to propagate onto the `Interface`.

crates/storage/src/interface.rs
miraclx: So previously, if a transaction (execution request) fails, all of it's changes to storage are rolled back.

Just checking if this is behaviour you intend to retain under this model.

crates/storage/src/interface.rs
miraclx: > So the lack of stabilisation in Rust core didn't really prevent anything here.

https://github.com/calimero-network/core/blob/1db5a74d5afd88d294c10a5b758cdae383db0cbd/crates/store/src/lib.rs#L22-L24

`Store`'s is implemented with dynamic dispatch, neither of these are object safe since the returned future from different implementations might be different types:

```rust
trait Database {
    fn get(&self) -> impl Future<Output = ()>;
}
```

```rust
trait Database {
    async fn get(&self);
}
```

we'd have had to forcefully unify the return type to a known `Box<dyn Future<Output = ()>>` to abstract away that variance or introduce `async_trait`, and at the point of implementation, I decided against it, we'll do this in a future effort where we can incorporate https://github.com/calimero-network/core/compare/455fe09ca9be09df17046584a3ef6cd28564e01a~...f74f4f48e39a8ad2556881c4a0f599fba346f766

> I am wondering, do you think we should change the Rc to an Arc to allow async here, which then avoids thread blocking?

for now, you have to think about where this will eventually be used, as Wasmer, our VM, doesn't have an async interface, and as such, handling host calls from the guest cannot use any async interfaces without blocking anyway

unless there's some immediate, critical need to depend on something that only exports an async interface, I'd say we keep it simple and adopt the sync model instead

once we sort out async execution for the runtime, and the store, we'll do this too

I'd earmarked to resolve those issues during the implementation of the task manager ([ref](https://github.com/calimero-network/core/issues/188)), but there's a lot more questions to be answered surrounding that

crates/storage/src/interface.rs
miraclx: > demands self-mut

where?

crates/storage/src/interface.rs
danwilliams: That's absolutely fine, and I appreciate the insightful response 🙂

crates/storage/src/interface.rs
danwilliams: Ah, I see now. So at present there is no context of that at all, and it's unclear if there will be. There is only one operation right now, hence it is inherently atomic. If/when that expands to have multiple operations (which may well happen), then yes such things would naturally need to be wrapped in a transaction. So I don't think this applies here.

miraclx: Approving to unblock, but can you clarify what you mean by:

> There is only one operation right now

Would we only be saving one `Element` when the state changes?

miraclx (approved):

crates/storage/src/tests/common.rs
```diff
@@ -0,0 +1,30 @@
...
+/// This function creates a new temporary directory and opens a new store in it,
+/// returning the store and the directory. The directory and its test data will
+/// be cleaned up when the test completes.
+///
+pub fn create_test_store() -> (Store, TempDir) {
+    // Note: It would be nice to have a way to test against an in-memory store,
+    // but InMemoryDB is not integrated with Store and there's currently no way
+    // to support both. It may be that the Database trait is later applied to
+    // InMemoryDB as well as RocksDB, in which case the storage Interface could
+    // be changed to work against Database implementations and not just Store.
```
miraclx, line 17: this is the case, though:

https://github.com/calimero-network/core/blob/8ecf05ae597ef4ded21a1edd664ee482e16e03de/crates/store/src/db/memory.rs#L132

https://github.com/calimero-network/core/blob/8ecf05ae597ef4ded21a1edd664ee482e16e03de/crates/store/src/tests/db/memory.rs

PR #714, opened 2024-09-10
feat(blobstore): abort stream once a specified content size is hit

miraclx:
Since the eventual functionality is to request the blob from peers, and the only way to determine it's legit is to collect the entire thing, we have to limit the receipt to prevent malicious peers from infinitely flooding with useless data.

This is the foundation of further pieces that'll make it in as part of the general effort behind the other context config PRs.

We'll store the blob ID as well as the size of the blob on the contract, so you start with knowledge of how large the stream received from your peers is supposed to be.

For now, we don't immediately delete the received chunks, we'll eventually introduce garbage collection to a number of things, including this.

### DIFF
--- a/crates/store/blobs/src/lib.rs
+++ b/crates/store/blobs/src/lib.rs
@@ -7,7 +7,7 @@ use async_stream::try_stream;
 use calimero_primitives::blobs::BlobId;
 use calimero_primitives::hash::Hash;
 use calimero_store::key::BlobMeta as BlobMetaKey;
-use calimero_store::types::BlobMeta;
+use calimero_store::types::BlobMeta as BlobMetaValue;
 use calimero_store::Store as DataStore;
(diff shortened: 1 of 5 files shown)


crates/store/blobs/src/lib.rs
```diff
@@ -39,6 +43,32 @@ struct State {
     size: usize,
 }
 
+#[derive(Eq, Ord, Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub enum Size {
+    Hint(u64),
```
MatejVukosav, line 48: What is the purpose of Hint size? Why just not use exact?

miraclx: Allocation optimization, since we need to allocate a vec of blob IDs, having some idea of how large a stream is, even though inexact can be optimal depending on the source of the stream.

PR #721, opened 2024-09-11
chore: Implement Embedded Static Build Files for Admin Dashboard into binary

alenmestrov:
## Summary:
This PR introduces a solution for embedding static build files for the admin dashboard into the binary. The static assets from the build directory are embedded to ensure no external static files are required at runtime. Additionally, for any non-static file requests, the server falls back to serving `index.html`, allowing dynamic routes to be handled appropriately by the client-side logic.

## Changes Implemented:
1. **Static File Embedding**:
   - Embedded static assets from the `node-ui/build` directory using `include_dir!`, ensuring that all assets are included within the binary itself.
   - Requests for static files (e.g., JS, CSS, images) are served directly from the embedded directory.

2. **Fallback for Non-Static Routes**:
   - If a requested file is not found in the embedded directory, the server serves `index.html` by default. This allows client-side logic to handle the routing for dynamic paths (e.g., `/admin-dashboard/identity/root-key/near`).
   - This ensures that all dynamic routes are routed back to the client-side application.

3. **Error Handling**:
   - If the requested file or `index.html` cannot be found, the server returns a 404 error with a "File not found" message.
   - Proper error handling is implemented using the `Result` type and the `?` operator to propagate errors cleanly.

4. **Code Simplification**:
   - The route handling has been simplified using a single catch-all route (`"/*path"`) for both root and subpath requests, eliminating redundant logic.


miraclx:
Ideally we compress these assets yeah?

https://github.com/pyrossh/rust-embed

Cargo.toml
```diff
@@ -82,6 +82,7 @@ rand = "0.8.5"
 rand_chacha = "0.3.1"
 reqwest = "0.12.2"
 rocksdb = "0.22.0"
+rust-embed = { version = "8.5.0", features = ["mime-guess"] }
```
miraclx: just the version should suffice

```suggestion
rust-embed = "8.5.0"
```

crates/server/src/admin/service.rs
```diff
@@ -174,24 +178,96 @@ pub(crate) fn setup(
...
+/// - `Result<impl IntoResponse, StatusCode>`: If the requested file is found or the fallback to index.html
+///   succeeds, it returns an `Ok` with the response. If no file can be served, it returns an `Err` with
+///   a 404 NOT_FOUND status code.
+async fn serve_embedded_file(uri: Uri) -> Result<impl IntoResponse, StatusCode> {
+    // Extract the path from the URI, removing the "/admin-dashboard/" prefix and any leading slashes
+    let path = uri
+        .path()
+        .trim_start_matches("/admin-dashboard/")
+        .trim_start_matches('/')
+        .to_string();
```
miraclx: seems this can be skipped?

```suggestion
        .trim_start_matches('/');
```

crates/server/src/admin/service.rs
```diff
@@ -174,24 +178,96 @@ pub(crate) fn setup(
...
+///
+/// # Returns
+/// - `Result<impl IntoResponse, StatusCode>`: If the response is successfully built, it returns an `Ok`
+///   with the response. If there is an error building the response, it returns an `Err` with a
+///   500 INTERNAL_SERVER_ERROR status code.
+async fn serve_file(file: &EmbeddedFile) -> Result<impl IntoResponse, StatusCode> {
+    Response::builder()
+        .status(StatusCode::OK)
+        .header("Content-Type", file.metadata.mimetype())
+        .body(Body::from(file.data.to_vec()))
```
```suggestion
async fn serve_file(file: EmbeddedFile) -> Result<impl IntoResponse, StatusCode> {
    Response::builder()
        .status(StatusCode::OK)
        .header("Content-Type", file.metadata.mimetype())
        .body(Body::from(file.data.into_owned()))
```

miraclx (approved):

Cargo.toml
alenmestrov: feature mime-guess is optional feature in rust-embed so this needs to be present otherwise it throws error that mimtype() is not found.

miraclx: it's already accounted for in the `Cargo.toml` of `calimero-server`, the root manifest should only specify versions unless with good reason

alenmestrov: I tried it but if I remove it even from root Cargo.toml it throws the error

miraclx: `_` != `-`

https://github.com/calimero-network/core/blob/9cabb26747e98d9b194e12a6aa0c4a1e4f7d9758/Cargo.toml#L88

https://github.com/calimero-network/core/blob/9cabb26747e98d9b194e12a6aa0c4a1e4f7d9758/crates/server/Cargo.toml#L31

crates/server/src/admin/service.rs
```diff
@@ -266,7 +265,7 @@ async fn serve_file(file: &EmbeddedFile) -> Result<impl IntoResponse, StatusCode
     Response::builder()
         .status(StatusCode::OK)
         .header("Content-Type", file.metadata.mimetype())
-        .body(Body::from(file.data.to_vec()))
+        .body(Body::from(file.data.to_owned()))
```
miraclx: quick note:

[`to_owned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html#impl-ToOwned-for-%5BT%5D) will clone the slice into a new allocation, no different from [`to_vec`](https://doc.rust-lang.org/std/primitive.slice.html#method.to_vec).

[`into_owned`](https://doc.rust-lang.org/std/borrow/enum.Cow.html#method.into_owned) won't if you already have an owned allocation, which, following the code [is the case](https://docs.rs/rust-embed/latest/rust_embed/struct.EmbeddedFile.html#structfield.data) (notice `'static`)

I suspect `into_owned` didn't work because it needs to take ownership of `self`, which is why you fell back to `to_owned`, however, you do in fact have ownership of the `Cow` .._

crates/server/src/admin/service.rs
```diff
@@ -174,24 +178,95 @@ pub(crate) fn setup(
...
+        .path()
+        .trim_start_matches("/admin-dashboard/")
+        .trim_start_matches('/');
+
+    // Use "index.html" for empty paths (root requests)
+    let path = if path.is_empty() { "index.html" } else { &path };
+
+    // Attempt to serve the requested file
+    if let Some(file) = NodeUiStaticFiles::get(path) {
+        return serve_file(&file).await;
```
```suggestion
        return serve_file(file).await;
```

crates/server/src/admin/service.rs
```diff
@@ -174,24 +178,95 @@ pub(crate) fn setup(
...
+    }
+
+    // Fallback to index.html for SPA routing if the file wasn't found and it's not already "index.html"
+    if path != "index.html" {
+        if let Some(index_file) = NodeUiStaticFiles::get("index.html") {
+            return serve_file(&index_file).await;
```
```suggestion
            return serve_file(index_file).await;
```

crates/server/src/admin/service.rs
```diff
@@ -174,24 +178,95 @@ pub(crate) fn setup(
...
+/// and serves the file's content.
+///
+/// # Parameters
+/// - `file`: A reference to the `EmbeddedFile` to be served.
+///
+/// # Returns
+/// - `Result<impl IntoResponse, StatusCode>`: If the response is successfully built, it returns an `Ok`
+///   with the response. If there is an error building the response, it returns an `Err` with a
+///   500 INTERNAL_SERVER_ERROR status code.
+async fn serve_file(file: &EmbeddedFile) -> Result<impl IntoResponse, StatusCode> {
```
```suggestion
async fn serve_file(file: EmbeddedFile) -> Result<impl IntoResponse, StatusCode> {
```

miraclx (approved):

PR #723, opened 2024-09-11
chore(CI): fix CI

miraclx:
fix CI

### DIFF
--- a/crates/primitives/src/context.rs
+++ b/crates/primitives/src/context.rs
@@ -71,22 +71,27 @@ pub struct ContextInvitationPayload(Vec<u8>);
 
 impl fmt::Debug for ContextInvitationPayload {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let (context_id, invitee_id, network, contract_id) =
-            self.parts().map_err(|_| fmt::Error)?;
-
-        let is_alternate = f.alternate();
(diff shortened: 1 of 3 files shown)


MatejVukosav (approved):

miraclx:
Tests seem borked 🫠 

cc: @MatejVukosav, have you observed this kind of issue?

Tried to check to make sure we weren't stuck in an infinite loop, doesn't seem to be the case, but the logs are also gone

miraclx:
Merging anyway as a fix to master, will continue investigating the actions runner issue

MatejVukosav:
> Tests seem borked 🫠 Runs for 44m after which it's killed, or something?
> 
> cc: @MatejVukosav, have you observed this kind of issue?
> 
> Tried to check to make sure we weren't stuck in an infinite loop, doesn't seem to be the case, but the logs are also gone

Not so far. They were much much faster before these new changes

MatejVukosav:
Issue is with memory "You are running out of disk space. The runner will stop working when the machine runs out of disk space."

miraclx:
Yeah, I've nuked the cache, it was bordering on 4GiB of space, should work now.

miraclx:
Yep, we're back in the green

https://github.com/calimero-network/core/commit/cc3dbc6819e64619c260c72f7fafc3035f150971

Issue #705, opened 2024-09-09
implement p2p application sharing


miraclx:
During a catchup, sharing the metadata for the context may lead to the discovery of a node having an out-of-date application installed.

At which point, it has the application source, and can fetch it.

The problem is, that the source URL may be unavailable, private, or is perhaps, a file path.

An application is a blob ID, associated with some metadata. The Blob API returns a stream of the full content of a blob when the ID is presented.

So, we need a way for nodes in need of certain applications (if they couldn't get it from the URL directly or if it's a file), to request the blob from other peers (or perhaps the peer they're directly engaged in catchup with) to get a stream of the blob's content, after which they can populate their local blob store (perhaps leveraging part of the networking streaming functionality already implemented or some other form of this)

PR #728, opened 2024-09-13
feat(node): catchup application blob request/response types

fbozic:
Define application blob request/response types for the catchup process. 

Related to: #705

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -226,13 +226,6 @@ impl ContextManager {
     ) -> EyreResult<()> {
         let mut handle = self.store.handle();
 
-        handle.put(
-            &ContextIdentityKey::new(context.id, identity_secret.public_key()),
-            &ContextIdentityValue {
-                private_key: Some(*identity_secret),
-            },
-        )?;
-
         if is_new {
             handle.put(
                 &ContextConfigKey::new(context.id),
@@ -252,6 +245,13 @@ impl ContextManager {
 
             self.subscribe(&context.id).await?;
         }
+
+        handle.put(
+            &ContextIdentityKey::new(context.id, identity_secret.public_key()),
+            &ContextIdentityValue {
+                private_key: Some(*identity_secret),
+            },
+        )?;
 
         Ok(())
     }
@@ -322,9 +322,6 @@ impl ContextManager {
 
         let context_exists = handle.has(&ContextMetaKey::new(context_id))?;
 
-        self.add_context(&context, identity_secret, !context_exists)
-            .await?;
-
         if !self.is_application_installed(&context.application_id)? {
             let source = Url::parse(&application.source.0)?;
 
@@ -344,6 +341,9 @@ impl ContextManager {
                 bail!("application mismatch")
             }
         }
+
+        self.add_context(&context, identity_secret, !context_exists)
+            .await?;
 
         let _ = self.state.write().await.pending_catchup.insert(context_id);
 
@@ -678,6 +678,8 @@ impl ContextManager {
         path: Utf8PathBuf,
         metadata: Vec<u8>,
     ) -> EyreResult<ApplicationId> {
+        let path = path.canonicalize_utf8()?;
+
         let file = File::open(&path).await?;
 
         let meta = file.metadata().await?;


crates/node/src/catchup.rs
```diff
@@ -35,8 +36,11 @@ impl Node {
         };
 
         let request = match from_json_slice(&message?.data)? {
-            CatchupStreamMessage::Request(req) => req,
+            CatchupStreamMessage::TransactionsRequest(req) => req,
             message @ (CatchupStreamMessage::TransactionsBatch(_)
+            | CatchupStreamMessage::ApplicationBlobRequest(_)
+            | CatchupStreamMessage::ApplicationBlobSize(_)
```
miraclx: all peers already know this information, following #719, the client peer has that information too

`is_application_installed` returns `false` in either of these cases;
- the application is unavailable
- the blob of the application is unavailable

But where `is_application_installed` fails, you can still attempt to fetch the application via `ContextManager::get_application` to determine if it's just the blob that unavailable, and you can proceed to request it from your peer

crates/node/src/catchup.rs
```diff
@@ -35,8 +36,11 @@ impl Node {
...
+            | CatchupStreamMessage::ApplicationBlobChunk(_)
```
miraclx, line 42: I wonder if it's possible to lose chunks or them arriving out of order, since this isn't necessarily TCP, in your stream research, did you come across any guarantees for data resilience?

crates/node/src/types.rs
```diff
@@ -32,22 +32,41 @@ pub struct TransactionRejection {
...
+
+#[derive(Debug, Deserialize, Serialize)]
+#[non_exhaustive]
+pub struct CatchupApplicationBlobSize {
+    pub size: u64,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+#[non_exhaustive]
+pub struct CatchupApplicationBlobChunk(Box<[u8]>);
```
miraclx: unless you want to employ some additional chunking, it's worth noting that blobstore chunks are 1MiB in size

https://github.com/calimero-network/core/blob/64420b00cf96c27d3674a5810357f70a8a94332f/crates/store/blobs/src/lib.rs#L24

so this will be encoded as an array of a max of 1,048,576 elements, which isn't the best way to send it over the wire

which is fine for now, but I can also see that `Message::new` just takes in a `Vec<u8>`, so maybe we can change it to borsh serialization instead to be more compact

Eventually, it'll be good to use a `Cow<'a, [u8]>` instead of a `Box<u8>` to reduce the expectations from the constructor, since borsh serialization only needs a reference, but also fine for now, just noting the parts we'd have to employ some care in

crates/node/src/catchup.rs
fbozic: I don't fully understand what you mean, all peers do know the information about applicationId, but peers are not aware of blobId changes (as far as I can tell from the code). 

this is the application representation in the contract
```
pub struct Application<'a> {
    pub id: Repr<ApplicationId>,
    pub blob: Repr<BlobId>,
    #[serde(borrow)]
    pub source: ApplicationSource<'a>,
    pub metadata: ApplicationMetadata<'a>,
}
```
this is the application representation in the storage:
```
pub struct ApplicationMeta {
    // todo! impl proper entry reference count
    // pub refs: usize,
    pub blob: BlobMetaKey,
    pub version: Option<Box<str>>,
    pub source: Box<str>,
    pub metadata: Box<[u8]>,
}

pub struct BlobMetaKey(Key<BlobId>);
```

so I should check during the catchup what the current blobId in the contract is and if the application source is a path I should request that application blob from a peer. so I think `ApplicationBlobRequest` and `ApplictionBlobSize` are required.

crates/node/src/catchup.rs
fbozic: my understanding of all p2p protocols is that you don't have to worry about transport (since it is a different layer) and you care about semantics. so I would say that we should be okay from a resilience POV.

also this example echos 100 bytes back to the sender: https://github.com/libp2p/rust-libp2p/blob/3837e33cd4c40ae703138e6aed6f6c9d52928a80/examples/stream/src/main.rs#L120-L134
so I would assume bytes are in order

I've added as well `sequential_id` to the BlobChunk message so we can verify the order of chunks

crates/node/src/types.rs
fbozic: our codec (length delimited) has an 8MB limit, so I can collect 7 chunks and push them over the wire to reduce the number of messages. I can do this in the implementation PR.

also I've opened a issue for Stream message serialization improvement: https://github.com/calimero-network/core/issues/732

crates/node/src/catchup.rs
miraclx: Following the considerations and accommodations made in https://github.com/calimero-network/core/pull/714, I'll be adding blob size to the contract in ~a coming PR~ #738, as we cannot trust the size the peer gives us either.

fbozic: removed the application size message

miraclx (approved):

PR #730, opened 2024-09-13
fix: Fixed lots of linting

danwilliams:
- Upgraded toolchain to Rust 1.81.

  - Added additional lints that have recently been stabilised.

  - Added reasons for lint exceptions.


crates/store/blobs/src/lib.rs
```diff
@@ -56,6 +56,11 @@ pub enum Size {
 
 impl Size {
     const fn hint(&self) -> usize {
+        // TODO: Check this, as the incoming int is a u64
```
miraclx, line 59: saturation is fine here

crates/context/config/src/repr.rs
```diff
@@ -196,11 +200,7 @@ where
         F: FnOnce(&mut Self::DecodeBytes) -> Bs58Result<usize>,
     {
         let mut bytes = Vec::new();
-
-        let len = f(&mut bytes)?;
-
-        assert_eq!(len, bytes.len());
```
miraclx, line 197: you just removed a sanity check, making this a [`debug assertion`](https://doc.rust-lang.org/std/macro.debug_assert.html) is fine, but I'd rather keep it

crates/context/config/src/repr.rs
```diff
@@ -96,11 +97,14 @@ pub trait ReprTransmute<'a>: ReprBytes + 'a {
     fn rt<O: ReprBytes<DecodeBytes = Self::EncodeBytes<'a>>>(&'a self) -> Result<O, O::Error>;
 }
 
-impl<'a, T: 'a> ReprTransmute<'a> for T
+impl<'a, T> ReprTransmute<'a> for T
 where
-    T: ReprBytes<EncodeBytes<'a>: AsRef<[u8]>>,
+    T: 'a + ReprBytes<EncodeBytes<'a>: AsRef<[u8]>>,
```
miraclx, line 102: I think the lifetime bounds in the former representation are easier to read, and perhaps preferable. `stdlib` seems to have that opinion

- https://github.com/rust-lang/rust/blob/28e8f01c2a2f33fb4214925a704e3223b372cad5/library/core/src/iter/traits/iterator.rs#L3415-L3418
- https://github.com/rust-lang/rust/blob/28e8f01c2a2f33fb4214925a704e3223b372cad5/library/core/src/iter/traits/iterator.rs#L3463-L3466

are there any lints that suggest refactoring this?

contracts/context-config/tests/sandbox.rs
```diff
@@ -427,18 +427,18 @@ async fn main() -> eyre::Result<()> {
...
-                            metadata: Default::default(),
-                        },
+                    ContextRequestKind::UpdateApplication {
+                        application: Application::new(
+                            new_application_id,
+                            new_blob_id,
+                            0,
+                            Default::default(),
+                            Default::default(),
+                        ),
```
miraclx, line 439: I don't think these blanket `new` implementations are always better than the explicit construction syntax, I don't like that we're losing information that will impair readability.

There's nothing inherently wrong with the explicit construction syntax, and protection of internal fields should be a case-by-case consideration, otherwise, we have to provide mechanisms for reclaiming the constituents of the structure, or pre-emptively clone those entries to be used elsewhere.

Factor in that enum struct variants can't have a `new` method either, it'll still come down to the explicit construction syntax, we don't have to box ourselves in.

crates/context/config/src/client/near.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::exhaustive_structs, reason = "TODO: Allowed until reviewed")]
```
miraclx, line 1: you have a todo here

contracts/context-config/src/mutate.rs
```diff
@@ -1,4 +1,9 @@
-use std::time;
+#![allow(
+    clippy::multiple_inherent_impl,
+    reason = "Needed to separate NEAR functionality"
+)]
```
miraclx, line 4: why is this prevented anyway?

crates/context/config/src/client.rs
```diff
@@ -224,8 +234,10 @@ impl<'a, T: Transport> ContextConfigQueryClient<'a, T> {
         &self,
         context_id: ContextId,
         identities: &[ContextIdentity],
-    ) -> Result<Response<BTreeMap<Repr<SignerId>, Vec<Capability>>>, Error<T>> {
-        let identities = unsafe { mem::transmute::<_, &[Repr<ContextIdentity>]>(identities) };
+    ) -> Result<Response<BTreeMap<Repr<SignerId>, Vec<Capability>>>, ConfigError<T>> {
+        let identities = unsafe {
+            &*(ptr::from_ref::<[ContextIdentity]>(identities) as *const [Repr<ContextIdentity>])
+        };
```
miraclx, line 255: how's this any better than transmutation? looks to me it obscures what the intent is

crates/context/config/src/client/near.rs
```diff
@@ -267,36 +275,34 @@ impl Network {
...
+            return Err(NearError::InvalidResponse {
+                operation: ErrorOperation::FetchAccount,
+            });
         };
 
         if let AccessKeyPermissionView::FunctionCall {
-            allowance: _,
             receiver_id,
             method_names,
+            ..
```
miraclx, line 297: NEAR's API is unstable, I prefer to match on the exact state of things so when the API changes (while they retain the old behaviour for backwards compact), our code should break at the time we make an update, so we can make an informed decision on what the best course of action is. Use the new APIs, or stick with what we have and ignore the introductions.

`..` obscures that and auto-ignores any new introductions.

crates/context/config/src/client/near.rs
```diff
@@ -267,36 +275,34 @@ impl Network {
...
-            _ => {
-                return Err(Error::InvalidResponse {
-                    operation: ErrorOperation::FetchAccount,
-                })
-            }
+        let RpcQueryResponse {
+            kind: QueryResponseKind::AccessKey(AccessKeyView { nonce, permission }),
+            block_hash,
+            ..
+        } = response
```
miraclx, line 287: same here

crates/context/src/lib.rs
```diff
@@ -352,6 +350,7 @@ impl ContextManager {
         Ok(Some((context_id, invitee_id)))
     }
 
+    #[expect(clippy::similar_names, reason = "Different enough")]
```
miraclx, line 353: has this caught anything that wasn't a false positive?

crates/network/src/discovery.rs
```diff
@@ -31,9 +31,9 @@ impl EventLoop {
     // Sends rendezvous discovery requests to all rendezvous peers which are not throttled.
     // If rendezvous peer is not connected, it will be dialed which will trigger the discovery during identify exchange.
     // TODO: Consider splitting this function up to reduce complexity.
-    #[allow(clippy::cognitive_complexity)]
+    #[expect(clippy::cognitive_complexity, reason = "TODO: Will be refactored")]
     pub(crate) fn broadcast_rendezvous_discoveries(&mut self) {
-        #[allow(clippy::needless_collect)]
+        #[expect(clippy::needless_collect, reason = "Necessary here; false positive")]
```
miraclx, line 36: is it? seems to me it's valid

crates/network/src/discovery.rs
```diff
@@ -101,9 +101,9 @@ impl EventLoop {
     // Sends rendezvous registrations request to all rendezvous peers which require registration.
     // If rendezvous peer is not connected, it will be dialed which will trigger the registration during identify exchange.
     // TODO: Consider splitting this function up to reduce complexity.
-    #[allow(clippy::cognitive_complexity)]
+    #[expect(clippy::cognitive_complexity, reason = "TODO: Will be refactored")]
     pub(crate) fn broadcast_rendezvous_registrations(&mut self) {
-        #[allow(clippy::needless_collect)]
+        #[expect(clippy::needless_collect, reason = "Necessary here; false positive")]
```
miraclx, line 106: same here

miraclx, line 106: same here

crates/network/src/events.rs
```diff
@@ -1,3 +1,8 @@
+#![allow(
+    clippy::allow_attributes,
+    reason = "Needed for lints that don't follow expect"
```
miraclx, line 3: why doesn't `expect` work?

crates/primitives/src/context.rs
```diff
@@ -71,27 +71,26 @@ pub struct ContextInvitationPayload(Vec<u8>);
...
+            if !is_alternate {
                 return d.finish();
             }
         }
 
-        let _ = d.field("raw", &self.to_string());
+        let mut d = f.debug_struct("ContextInvitationPayload");
+        _ = d.field("raw", &self.to_string());
 
         d.finish()
```
miraclx, line 95: I don't think this is better than what we had before

crates/runtime/src/logic.rs
```diff
@@ -1,5 +1,5 @@
-#![allow(single_use_lifetimes, unused_lifetimes)]
-#![allow(clippy::mem_forget)]
+#![allow(single_use_lifetimes, unused_lifetimes, reason = "False positive")]
+#![allow(clippy::mem_forget, reason = "Safe for now")]
```
miraclx, line 2: are there multiple offenders? if not, can we inline these?

crates/runtime/src/logic/imports.rs
```diff
@@ -14,8 +14,7 @@ thread_local! {
 }
 
 impl VMLogic<'_> {
-    #[allow(clippy::too_many_arguments)]
-    #[allow(trivial_casts)]
+    #[expect(clippy::too_many_arguments, reason = "Acceptable here")]
     pub fn imports(&mut self, store: &mut Store) -> Imports {
```
miraclx, line 18: why is this even an issue, it's just one argument

crates/sdk/macros/src/logic.rs
```diff
@@ -42,7 +42,7 @@ impl<'a> TryFrom<LogicImplInput<'a>> for LogicImpl<'a> {
     type Error = Errors<'a, ItemImpl>;
 
     // TODO: This unwrap() call needs to be corrected to return an error.
-    #[allow(clippy::unwrap_in_result)]
+    #[expect(clippy::unwrap_in_result, reason = "TODO: This is temporary")]
```
miraclx, line 45: It's infallible

https://github.com/calimero-network/core/blob/d124aa3592c969d4aeb2879a46830d4ba1046ec4/crates/sdk/macros/src/sanitizer.rs#L251-L253

crates/sdk/macros/src/logic/ty.rs
```diff
@@ -28,7 +28,7 @@ impl<'a, 'b> TryFrom<LogicTyInput<'a, 'b>> for LogicTy {
     type Error = Errors<'a, Type>;
 
     // TODO: This unwrap() call needs to be corrected to return an error.
-    #[allow(clippy::unwrap_in_result)]
+    #[expect(clippy::unwrap_in_result, reason = "TODO: This is temporary")]
```
miraclx, line 31: same

crates/sdk/macros/src/sanitizer.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code, reason = "Will be used in future")]
```
miraclx, line 1: can we localize this to the offender?

crates/sdk/src/sys/types/buffer.rs
```diff
@@ -41,7 +41,7 @@ impl<'a, T> Slice<'a, T> {
     // TODO: but this function is meant to be infallible. That is a concern, as
     // TODO: we want to eliminate all potential panics. This needs future
     // TODO: assessment.
-    #[allow(clippy::cast_possible_truncation)]
+    #[expect(clippy::cast_possible_truncation, reason = "TODO: See above")]
```
miraclx, line 44: this executes on the guest, which at the time of writing is wasm32, so `usize` will be a 32-bits

We use `u64` to accommodate the potential future effort of supporting `wasm64` as well

And finally, `Buffer`s will only be constructed on the guest, never on the host, so it should always be safe to convert whatever made up `self.len` (which was originally a `usize`) back to a `usize`.

crates/sdk/src/sys/types/pointer.rs
```diff
@@ -21,7 +21,7 @@ impl PtrSizedInt {
     // TODO: but this function is meant to be infallible. That is a concern, as
     // TODO: we want to eliminate all potential panics. This needs future
     // TODO: assessment.
-    #[allow(clippy::cast_possible_truncation)]
+    #[expect(clippy::cast_possible_truncation, reason = "TODO: See above")]
```
miraclx, line 24: same here

crates/server-primitives/src/admin.rs
```diff
@@ -210,13 +210,19 @@ pub struct NearSignatureMessageMetadata {
 #[derive(Clone, Copy, Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 #[non_exhaustive]
-#[allow(clippy::empty_structs_with_brackets)]
+#[expect(
+    clippy::empty_structs_with_brackets,
+    reason = "Needed for serialisation"
+)]
```
miraclx, line 217: ideally, lints shouldn't distract from functionality, move the functional macro annotations closer to the struct

crates/store/blobs/src/lib.rs
```diff
@@ -235,7 +240,10 @@ impl Blob {
             return Ok(None);
         };
 
-        #[allow(clippy::semicolon_if_nothing_returned)]
+        #[expect(
+            clippy::semicolon_if_nothing_returned,
+            reason = "False positive; not possible with macro"
+        )]
```
miraclx, line 246: is this complaining about code generated by the macro?

crates/store/src/iter.rs
```diff
@@ -58,7 +58,10 @@ impl<'a, K, V> Iter<'a, K, V> {
 }
 
 // TODO: We should consider using std::iter::Iterator here
-#[allow(clippy::should_implement_trait)]
+#[expect(
+    clippy::should_implement_trait,
+    reason = "TODO: This should be implemented"
+)]
```
miraclx, line 64: uhm, why? we already have `.keys()` and `.entries()` that are iterators and represent the functionality we want

crates/store/blobs/src/lib.rs
danwilliams: Saturation isn't relevant, as this is a cast. It could overflow and be truncated, hence the comment. It's marked as a TODO for someone to either add proper conversion, or, add explanatory comments about why it's infallible or acceptable.

crates/context/config/src/repr.rs
danwilliams: The "sanity check" was for a situation that cannot technically occur. If you feel it can occur, it should have error handling and not panic.

crates/context/config/src/repr.rs
danwilliams: There was a reason for it being changed, but I don't know which lint triggered it. Feel free to experiment... I can't remember now after fixing hundreds of the things!

contracts/context-config/tests/sandbox.rs
danwilliams: The point is that anything exposed externally needs to be thought about. If something is truly exhaustive, that's fine and it can be marked as such... if not exhaustive then best practice is to mark it as `non_exhaustive` and that way consumers of the code know not to rely on it. One of the most important aspects of this is that it's then not possible to construct via struct syntax, although it is still possible to access public fields. As we stabilise the public interface we need to be thinking about these things and working in a semver-compatible manner, which is what this assists.

contracts/context-config/src/mutate.rs
danwilliams: Having multiple separate `impl`s is generally discouraged as it can be confusing. There are cases where it's okay.

crates/context/config/src/client/near.rs
danwilliams: Yup. That's for the code owner to pick up on and decide as to which should be exhaustive and which not.

crates/context/config/src/client.rs
danwilliams: It's safer. It's unsafe code, so the safer the better.

crates/context/config/src/client/near.rs
danwilliams: That's fair, I have no problem with that.

crates/context/src/lib.rs
danwilliams: It does for me from time to time, yes - there are some things that are visually similar and hence this can help avoid confusion.

crates/network/src/discovery.rs
danwilliams: Try it 🙂

crates/network/src/events.rs
danwilliams: Not all lints behave as "expect"ed with expect at present.

crates/primitives/src/context.rs
danwilliams: It's better than flagging a variable as unused and then using it...

crates/runtime/src/logic.rs
danwilliams: Feel free...

crates/runtime/src/logic/imports.rs
danwilliams: And yet it is 🙂

crates/sdk/macros/src/logic.rs
danwilliams: Feel free to follow up on the areas like this that are flagged for you

crates/sdk/macros/src/sanitizer.rs
danwilliams: Feel free to 🙂

crates/sdk/src/sys/types/buffer.rs
danwilliams: That would be great to note in the code 🙂

crates/store/blobs/src/lib.rs
danwilliams: Yup - as per the reason added 🙂

crates/store/src/iter.rs
danwilliams: Because of expectations is all... 🙂

MatejVukosav (approved):

PR #736, opened 2024-09-16
fix(node-ui): fix lint warnings

frdomovic:
# fix(node-ui): fix lint warnings

## Summary
As title says, lint warnings and unused code.


node-ui/src/App.tsx
```diff
@@ -31,6 +31,7 @@ export default function App() {
 
   useEffect(() => {
     setNodeUrlFromQuery(showServerDownPopup);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
frdomovic, line 34: To disable warnings but won't be adding suggested variables because of unexpected behaviour of useEffect function.

alenmestrov (approved):

PR #737, opened 2024-09-16
chore: remove dbg from store

MatejVukosav:


miraclx:
Nice catch! Though already included in #738.

PR #739, opened 2024-09-16
CLI application install

petarjuki7:
# Application install patch

## Summary

Metadata is optional and is passed to the CLI as a String


crates/meroctl/src/cli/app/install.rs
```diff
@@ -55,11 +55,13 @@ impl InstallCommand {
 
         let mut is_dev_installation = false;
 
+        let metadata = self.metadata.map(|m| m.into_bytes());
```
miraclx: you can move `unwrap_or_default` here as well

miraclx (approved):

MatejVukosav (approved):

PR #743, opened 2024-09-17
fix: canonicalize application path on installation

miraclx:
interactive CLI shouldn't require an absolute path input

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -678,6 +678,8 @@ impl ContextManager {
         path: Utf8PathBuf,
         metadata: Vec<u8>,
     ) -> EyreResult<ApplicationId> {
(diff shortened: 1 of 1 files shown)


frdomovic (approved):

crates/context/src/lib.rs
```diff
@@ -678,6 +678,8 @@ impl ContextManager {
         path: Utf8PathBuf,
         metadata: Vec<u8>,
     ) -> EyreResult<ApplicationId> {
+        let path = path.canonicalize_utf8()?;
```
frdomovic, line 681: Does this require any change in docs for install?

miraclx: No, the current behaviour is maintained

PR #750, opened 2024-09-18
feat(sdk): refresh jwt components

frdomovic:
# feat(sdk): refresh jwt components

## Summary
Added a wrapper which will be used in application (around all components) to check if token expired. If token expired it wil call admin api endpoint to get new token and save this token to storage.

As part of this PR , refresh jwt endpoint has been moved to public endpoints

### DIFF
--- a/packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
+++ b/packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
@@ -1,0 +1,66 @@
+import React, { useEffect } from 'react';
+import { getAccessToken, getRefreshToken } from '../storage';
+import { jwtDecode } from 'jwt-decode';
+import { getNewJwtToken } from './refreshToken';
+
+interface AccessTokenWrapperProps {
+  children: React.ReactNode;
+  getNodeUrl: () => string;
+}
+
+export const AccessTokenWrapper: React.FC<AccessTokenWrapperProps> = ({
+  children,
+  getNodeUrl,
+}) => {
+  const decodeToken = (token: string) => {
+    try {
+      return jwtDecode(token);
+    } catch (error) {
+      return null;
+    }
+  };
+
+  const isTokenExpiringSoon = (token: string) => {
+    const decodedToken = decodeToken(token);
+    if (!decodedToken || !decodedToken.exp) {
+      return true;
+    }
+
+    const currentTime = Math.floor(Date.now() / 1000);
+    const timeUntilExpiry = decodedToken.exp - currentTime;
+
+    return timeUntilExpiry <= 5 * 60;
+  };
+
+  const validateAccessToken = async () => {
+    const accessToken = getAccessToken();
+    const refreshToken = getRefreshToken();
+
+    if (!accessToken || !refreshToken) {
+      return;
+    }
+
+    if (isTokenExpiringSoon(accessToken)) {
+      try {
+        await getNewJwtToken({ refreshToken, getNodeUrl });
+      } catch (error) {
+        console.log(error);
+      }
+    }
+  };
+
+  useEffect(() => {
+    validateAccessToken();
+
+    const intervalId = setInterval(
+      () => {
+        validateAccessToken();
+      },
+      20 * 60 * 1000,
+    );
+
+    return () => clearInterval(intervalId);
+  }, [getNodeUrl]);
+
+  return <>{children}</>;
+};
--- a/packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
+++ b/packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
@@ -20,17 +20,20 @@ export const AccessTokenWrapper: React.FC<AccessTokenWrapperProps> = ({
     }
   }, []);
 
-  const isTokenExpiringSoon = useCallback((token: string) => {
-    const decodedToken = decodeToken(token);
-    if (!decodedToken || !decodedToken.exp) {
-      return true;
-    }
+  const isTokenExpiringSoon = useCallback(
+    (token: string) => {
+      const decodedToken = decodeToken(token);
+      if (!decodedToken || !decodedToken.exp) {
+        return true;
+      }
 
-    const currentTime = Math.floor(Date.now() / 1000);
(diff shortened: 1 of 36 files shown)


packages/calimero-sdk/src/login/refreshToken.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export const getNewJwtToken = async ({
+  refreshToken,
+  getNodeUrl,
+}: GetNewJwtTokenProps) => {
+  const tokenResponse: ResponseData<JwtTokenResponse> = await apiClient
+    .node()
+    .refreshToken(refreshToken, getNodeUrl());
+  if (tokenResponse.data) {
```
MatejVukosav: How do you handle error?

packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
alenmestrov, line 1: maybe wrap functions in useCallback to avoid unnecessary rerenders

packages/calimero-sdk/src/login/refreshToken.ts
frdomovic: Added proper error handling within this function and handler on client side.

packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
frdomovic: Updated

frdomovic:
How the testing was done:
1. Setup Admin Dashboard - install onlypeers application to be able to copy application id
2. Setup Only Peers Example
Env file used:
```
NEXT_PUBLIC_NEAR_ENV=testnet
NEXT_PUBLIC_CONTEXT_ID="ANY"
NEXT_PUBLIC_RPC_BASE_URL="http://localhost:2428"
NEXT_PUBLIC_RPC_PATH="/jsonrpc"
```
*NEXT_PUBLIC_CONTEXT_ID is env that will be removed in cleanup, right now not important
Go through the process of login and then try to fetch posts or create post. 
Fetch posts returns empty array while create posts returns no peers connected error which is expected behaviour when running only 1 not for testing.

Aditionally to test JWT update their expiration time from 1 hour to 30 seconds so things like expiry and not valid token can be tested.


Screenshot 2024-09-19 at 15 57 06


You might see error like this for forbidden which is caused by not updating logic in the wrapper - currently is set to work with hour values not seconds so this is expected in test mode.
Screenshot 2024-09-19 at 15 56 01

cc @alenmestrov

frdomovic:
Testing expired token / no token / invalid token (when refresh token can't be called asks to relogin)

https://github.com/user-attachments/assets/66a77ef5-7aae-4922-bcfe-4639a20f99c2


https://github.com/user-attachments/assets/3b2c5243-99cb-47c0-9370-9a68dbda15ab

packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    if (!accessToken || !refreshToken) {
+      return;
+    }
+
+    if (isTokenExpiringSoon(accessToken)) {
+      try {
+        await getNewJwtToken({ refreshToken, getNodeUrl });
+      } catch (error) {
+        console.log(error);
```
alenmestrov: console.error()

packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const AccessTokenWrapper: React.FC<AccessTokenWrapperProps> = ({
+  children,
+  getNodeUrl,
+}) => {
+  const decodeToken = useCallback((token: string) => {
+    try {
+      return jwtDecode(token);
+    } catch (error) {
+      return null;
```
alenmestrov, line 20: console.error()

packages/calimero-sdk/src/login/refreshToken.ts
```diff
@@ -0,0 +1,78 @@
...
+  error: RpcError,
+  getNodeUrl: () => string,
+): Promise<ErrorResponse> => {
+  const invalidSession = {
+    message: 'Your session is no longer valid. Please log in again.',
+    code: 401,
+  };
+  const expiredSession = {
+    message:
+      'Your session expired, but we have refreshed it. Please try again.',
```
alenmestrov: 401 or 403, 500 is for server error

packages/calimero-sdk/src/login/refreshToken.ts
```diff
@@ -0,0 +1,78 @@
...
+      error.type === 'RpcExecutionError'
+    ) {
+      return {
+        message: `Error: ${error?.inner?.data?.data?.type}`,
+        code: 500,
+      };
+    } else {
+      return unknownMessage;
+    }
+  }
```
alenmestrov: here we can use early returns. 
First check if error code is 401, then check for message otherwise return second exit.
```
if (error.code === 401) {
  if (error.message === 'Token expired.') {
    const refreshToken = getRefreshToken();
    try {
      const response = await getNewJwtToken({ refreshToken, getNodeUrl });
      if (response?.error) {
        clearJWT();
        return invalidSession;
      }
      return expiredSession;
    } catch (error) {
      clearJWT();
      return invalidSession;
    }
  }
  clearJWT();
  return invalidSession;
}
```

packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
frdomovic: updated

packages/calimero-sdk/src/login/AccessTokenWrapper.tsx
frdomovic: updated

packages/calimero-sdk/src/login/refreshToken.ts
frdomovic: updated

packages/calimero-sdk/src/login/refreshToken.ts
frdomovic: updated

alenmestrov (approved):

PR #752, opened 2024-09-20
feat(node): catchup application blob exchange

fbozic:
Resolves: #705


crates/context/src/lib.rs
```diff
@@ -768,6 +793,27 @@ impl ContextManager {
...
+            .await?;
+
+        let application = response.parse()?;
+
+        Ok(Application::new(
+            application.id.as_bytes().into(),
+            application.blob.as_bytes().into(),
+            application.size,
+            ApplicationSource::from_str(&application.source.0)?,
+            application.metadata.0.as_bytes().into(),
```
```suggestion
            application.metadata.0.into_inner().into_bytes(),
```

crates/node/src/catchup/blobs.rs
```diff
@@ -0,0 +1,121 @@
...
+            let remaining = &buffer[self.offset..];
+            let len = remaining.len().min(buf.len());
+            buf[..len].copy_from_slice(&remaining[..len]);
+            self.offset += len;
+            Poll::Ready(Ok(len))
+        } else {
+            Poll::Ready(Ok(0))
+        }
+    }
+}
```
miraclx: Looks like this is [BufReader](https://docs.rs/futures-util/latest/futures_util/io/struct.BufReader.html)?

crates/node/src/catchup/blobs.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub(crate) fn new(batch_size: u8, stream: Box<Stream>) -> Self {
+        Self {
+            batch_size,
+            batch: Vec::with_capacity((batch_size as usize) * BLOB_CHUNK_SIZE),
+            stream,
+            sequential_id: 0,
+        }
+    }
+
+    pub(crate) async fn send(&mut self, chunk: Box<[u8]>) -> EyreResult<()> {
```
miraclx:
```suggestion
    pub(crate) async fn send(&mut self, chunk: &[u8]) -> EyreResult<()> {
```

crates/node/src/catchup.rs
```diff
@@ -2,48 +2,96 @@ use std::collections::VecDeque;
...
+            ))?;
+            stream.send(Message::new(message)).await?;
+
+            return Ok(());
         };
 
+        // Stream messages are encoded with length delimited codec.
+        // Max message size is 8MB and blob chunk size 1MB (atm).
+        // So, collect 7 chunks and push them to over the stream.
+        let mut blob_sender = ApplicationBlobChunkSender::new(7, stream);
```
miraclx: can we `const`antly make this calculation? in case either of its dependencies change

(MAX_MESSAGE_SIZE / BLOB_CHUNK_SIZE ) - 1

crates/context/src/lib.rs
fbozic: This doesn't compile, error:
```
            application.metadata.0.into_iter().into_bytes(),
there is a method `into_iter` with a similar name: `into_iter`
        ))
```

crates/node/src/catchup/blobs.rs
fbozic: Removed buffering logic from AsyncRead and wrapped ApplicationBlobChunkStream into BufReader in the handler

crates/context/src/lib.rs
miraclx: `into_inner`, not `into_iter`

crates/node/src/catchup/blobs.rs
```diff
@@ -42,28 +36,21 @@ impl StreamFutures for ApplicationBlobChunkStream {
...
-            buf[..len].copy_from_slice(&remaining[..len]);
-            self.offset += len;
-            Poll::Ready(Ok(len))
-        } else {
-            Poll::Ready(Ok(0))
+        match Pin::new(&mut this.receiver).poll_recv(cx) {
+            Poll::Ready(Some(chunk)) => {
+                let data = &chunk.chunk;
+                let len = data.len().min(buf.len());
+                buf[..len].copy_from_slice(&data[..len]);
```
miraclx: `sequential_id` isn't checked?

miraclx: if the input buffer is less than what we read from the stream, this will truncate and discard the remaining bytes

we can use [`poll_fn`](https://docs.rs/futures-util/latest/futures_util/stream/fn.poll_fn.html) to poll [`poll_recv`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.poll_recv)

```rust
let stream = poll_fn(|cx| receiver.poll_recv(cx));
```

now we have a `Stream<Item = CatchupApplicationBlobChunk>` but to fix the issue mentioned above, we need to create a `BufReader` which requires the stream to be `AsyncRead`

so first, we can make the stream into a `Stream<Item = Result<Box<u8>, E>>` by:

```rust
let stream = stream.map(|chunk| {
    // check sequential_id, maybe return error?
    Ok(chunk.chunk)
});
```

and then we can make it into an `AsyncRead` with [`into_async_read`](https://docs.rs/futures-util/latest/futures_util/stream/trait.TryStreamExt.html#method.into_async_read)

```rust
let stream = stream.into_async_read();
```

At this point, we can pass it to [`BufReader::new`](https://docs.rs/futures-util/latest/futures_util/io/struct.BufReader.html#method.new)

```rust
let stream = BufReader::new(stream);
```

together it's about:

```rust
let stream = BufReader::new(
    poll_fn(|cx| receiver.poll_recv(cx))
        .map(|chunk| {
            // check sequential_id, maybe return error?
            Ok(chunk.chunk)
        })
        .into_async_read(),
);
```

..._

crates/node/src/catchup.rs
```diff
@@ -202,8 +259,123 @@ impl Node {
...
+            application_id: latest_application.id,
+        };
+
+        let data = to_json_vec(&request)?;
+
+        stream.send(Message::new(data)).await?;
+
+        let (tx, rx) = mpsc::channel(100);
+        let chunk_stream = ApplicationBlobChunkStream::new(rx);
+        let chunk_stream = BufReader::new(chunk_stream);
```
miraclx: _... which can be done here instead

crates/context/src/lib.rs
fbozic: sorry typo, I've tried this as well, but I get different error
```
            application.metadata.0.into_inner().into_bytes(),
items from traits can only be used if the trait is in scope
there is a method `bytes` with a similar name: `bytes`
        ))
```

And rustc is giving this hint: 
```
rustc: trait `IntoBytes` which provides `into_bytes` is implemented but not in scope; perhaps you want to import it: `use wasmer::into_bytes::IntoBytes;`
```

But that is not the trait we need. I'm not sure which trait I need here.

crates/node/src/catchup/blobs.rs
fbozic: amazing thank you

crates/context/src/lib.rs
miraclx: Ah, [`Cow::into_owned`](https://doc.rust-lang.org/nightly/alloc/borrow/enum.Cow.html#method.into_owned)

miraclx (approved):
lgtm ✨

crates/node/src/catchup/blobs.rs
```diff
@@ -0,0 +1,71 @@
...
+use calimero_blobstore::CHUNK_SIZE as BLOB_CHUNK_SIZE;
+use calimero_network::stream::{Message, Stream, MAX_MESSAGE_SIZE as MAX_STREAM_MESSAGE_SIZE};
+use eyre::{bail, Result as EyreResult};
+use futures_util::SinkExt;
+use serde_json::to_vec as to_json_vec;
+
+use crate::types::{CatchupApplicationBlobChunk, CatchupStreamMessage};
+
+pub struct ApplicationBlobChunkSender {
+    batch_size: u8,
```
miraclx: if it'll always be converted to a `usize` to be used, maybe just leave it as that?

miraclx (approved):

Issue #480, opened 2024-07-16
Improve rendezvous registration implementation


fbozic:
At the moment we register our external addresses at all rendezvous peers. Implement selection of rendezvous nodes to avoid network congestion.

One approach could be to select the first N peers as nominated peers. After our reservation has expired, we could nominate a new peer and perform registration.

PR #753, opened 2024-09-20
Improve rendezvous registration

petarjuki7:
# Improve rendezvous registration implementation

## Regarding #480 

> At the moment we register our external addresses at all rendezvous peers. Implement selection of rendezvous nodes to avoid network congestion.

>One approach could be to select the first N peers as nominated peers. After our reservation has expired, we could nominate a new peer and perform registration.


crates/network/src/discovery.rs
```diff
@@ -141,7 +141,14 @@ impl EventLoop {
...
             .wrap_err("Failed to get peer info")?;
 
-        if !peer_info.is_rendezvous_registration_required() {
+        if !peer_info.is_rendezvous_registration_required()
+            || self.discovery.state.rendezvous_nominated_peers_size()
+                >= self.discovery.rendezvous_config.nominated_peers_size
+            || self
+                .discovery
+                .state
+                .is_peer_rendezvous_nominated(rendezvous_peer)
```
fbozic: Maybe move this to the `is_rendezvous_registration_required` method since it will be required in the `broadcast_rendezvous_registrations`. 

Registration is required if we currently have less than `self.discovery.rendezvous_config.registrations_limit`

crates/network/src/config.rs
```diff
@@ -157,6 +157,21 @@ pub struct RendezvousConfig {
     pub discovery_rpm: f32,
 
     pub discovery_interval: Duration,
+
+    pub nominated_peers_size: usize,
```
fbozic: `registrations_limit` maybe?

crates/network/src/events/rendezvous.rs
```diff
@@ -89,10 +93,32 @@ impl EventHandler<Event> for EventLoop {
...
+                    false
+                });
+
+                if let Some(peer) = nominated_peer {
+                    if let Err(err) = self.rendezvous_register(&peer) {
+                        error!(%err, "Failed to update registration discovery");
+                    };
+                } else {
+                    error!("Couldn't find new peer to nominate for rendezvous registration.");
+                }
```
fbozic: I don't think this is an error. We just don't have a new candidate ATM.

crates/network/src/tests/discovery/state.rs
```diff
@@ -89,35 +89,35 @@ fn test_is_rendezvous_discovery_throttled() {
...
+//         ..Default::default()
+//     });
+//     assert!(peer_info.is_rendezvous_registration_required());
+
+//     peer_info.rendezvous = Some(PeerRendezvousInfo {
+//         registration_status: RendezvousRegistrationStatus::Expired,
+//         ..Default::default()
+//     });
+//     assert!(peer_info.is_rendezvous_registration_required());
+// }
```
petarjuki7: Because of the new logic, don't think this test makes sense, I can remove it or update it.

fbozic: Please update tests

crates/network/src/discovery/state.rs
```diff
@@ -162,6 +175,10 @@ impl DiscoveryState {
     pub(crate) fn is_peer_rendezvous(&self, peer_id: &PeerId) -> bool {
         self.rendezvous_index.contains(peer_id)
     }
+
+    pub(crate) fn is_rendezvous_registration_required(&self, max: usize) -> bool {
+        self.rendezvous_nominated_peers_size() < max
+    }
```
fbozic: Aren't we now only checking the number of nominated peers? I think this will lead to redundant registration calls since you are no longer checking if we have requested registration against a peer.

Thinking about it, it may make sense to drop a new index in the state because we can fairly simply check if we have less than `limit` registrations (registration status has to be in Requested or Registered).

crates/meroctl/src/cli/init.rs
```diff
@@ -77,6 +77,9 @@ pub struct InitCommand {
     #[clap(overrides_with("mdns"))]
     pub no_mdns: bool,
 
+    #[clap(long, default_value = "3")]
+    pub rendezvous_nominated_peers: usize,
```
fbozic: I think this should be called `rendezvous_registrations_limit` to match the name in the config

crates/network/src/discovery.rs
```diff
@@ -141,7 +143,9 @@ impl EventLoop {
             .get_peer_info(rendezvous_peer)
             .wrap_err("Failed to get peer info")?;
 
-        if !peer_info.is_rendezvous_registration_required() {
+        if !self.discovery.state.is_rendezvous_registration_required(
+            self.discovery.rendezvous_config.registrations_limit,
```
fbozic, line 147: Why are we passing the limit here, why not use `self...` inside the `is_rendezvous_registration_required`?

petarjuki7: Not sure how to get to the limit, the `is_rendezvous_registration_required` method is 'inside' the `DiscoveryState` struct which doesn't have knowledge of the `RendezvousConfig` struct

fbozic: ah yeah good point, sorry my mistake

PR #760, opened 2024-09-25
Implement CLI commands for interaction with WebSocket endpoint

### LINKED ISSUE #541: Implement CLI commands for interaction with WebSocket endpoint
Support consuming WebSocket stream from the CLI. Something like `meroctl context watch --context-id <123>` would be nice.

petarjuki7:
# MeroCLI now supports consuming WebSocket stream from the CLI

## Summary

Described in #541, you can "listen" to incoming data regarding a context 

Fixes #541

## Test plan

`cargo run -p meroctl -- --node-name <node-name> context watch --context-id <context-id>` starts the process

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,7 +104,7 @@ thiserror = "1.0.56"
 thunderdome = "0.6.1"
 tokio = "1.35.1"
 tokio-test = "0.4.4"
-tokio-tungstenite = { version = "0.24.0", features = ["native-tls"] }
+tokio-tungstenite = "0.24.0"
 tokio-util = "0.7.11"
 toml = "0.8.9"
 toml_edit = "0.22.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,7 @@ thiserror = "1.0.56"
 thunderdome = "0.6.1"
 tokio = "1.35.1"
 tokio-test = "0.4.4"
+tokio-tungstenite = { version = "0.24.0", features = ["native-tls"] }
 tokio-util = "0.7.11"
 toml = "0.8.9"
 toml_edit = "0.22.14"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -104,6 +104,7 @@ thiserror = "1.0.56"
 thunderdome = "0.6.1"
 tokio = "1.35.1"
 tokio-test = "0.4.4"
+tokio-tungstenite = { version = "0.24.0", features = ["native-tls"] }
```
fbozic: I think we shouldn't use features in top-level Cargo.toml

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -0,0 +1,67 @@
...
+            .await?;
+
+        println!("Subscribed to context {}", self.context_id);
+        println!("Streaming events (press Ctrl+C to stop):");
+
+        while let Some(message) = read.next().await {
+            match message {
+                Ok(msg) => {
+                    if let tokio_tungstenite::tungstenite::Message::Text(text) = msg {
+                        println!("{}", text);
```
miraclx, line 58: do you have an example of these being printed? is it in this format:

```json
{ "kind": "SomethingHappened", "data": "associative data" }
```

petarjuki7: `{"id":null,"result":{"contextIds":["B7KKoamV9uRfFW8B1VnRYPMhJ1vHBaHBiQJXauJod3na"]}}`

`{"id":null,"result":{"contextId":"B7KKoamV9uRfFW8B1VnRYPMhJ1vHBaHBiQJXauJod3na","data":{"events":[]},"type":"OutcomeEvent"}}`

miraclx: Alrighty, thought we could maybe prettify this to make it human-readable, but it's fine for now

miraclx (approved):

PR #763, opened 2024-09-30
feat: Finalised Data trait and associated design

danwilliams:
- Removed the temporary `Data` struct and implemented a proper `Data` trait in its place. This temporarily has an associated type called `Child`, to specify the (one-dimensional) type of children.

  - Added `element()`and `element_mut()` methods to obtain the `Element` that the `Data` type owns.

  - Added `id()` and `path()` methods for convenience to pass through from `Data` to `Element`.

  - Added documentation to explain the design choices behind the implemented structure.

  - Removed the `Element.data` property and matching `data()` method, as `Element` will no longer own or have any reference to `Data`; `Data` needs to own `Element` for the chosen design pattern.

  - Updated the `Element.update_data()` method to no longer accept `Data`, as it does not have access to it. Renamed it to simply `update()` to reflect this.

  - Added a `NoChildren` type to be used (for now) for those types that have no children.

  - Made some Interface functions generic around `D: Data` - these are `calculate_merkle_hash()`, `children_of()`, `find_by_id()`, `find_by_path()`, `find_children_by_id()`, and `save()`.

  - Updated `Interface.children_of()` to return types of `Data::Child`.

  - Updated tests and added new tests as appropriate.

### DIFF
--- a/crates/storage/src/address.rs
+++ b/crates/storage/src/address.rs
@@ -33,7 +33,7 @@ use uuid::{Bytes, Uuid};
 /// system operation. Abstracting the true type away provides a level of
 /// insulation that is useful for any future changes.
 ///
-#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct Id(Uuid);
 
 impl Id {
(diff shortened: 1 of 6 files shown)


miraclx (approved):
approving to unblock, though I'm not sure what `Paragraph` and `Page` are supposed to be

danwilliams:
> approving to unblock, though I'm not sure what `Paragraph` and `Page` are supposed to be

You mean the tests...? They are simply examples of data.

PR #764, opened 2024-09-30
feat: Added AtomicUnit macro and trait

danwilliams:
Added `AtomicUnit` macro and trait

  - Added a new `AtomicUnit` marker trait.

  - Added a new `storage-macros` crate, to contain the macros needed for presenting the high-level developer interface.

  - Added an `AtomicUnit` proc macro to automatically derive `AtomicUnit` and `Data` traits for a struct.

Added `test-utils` crate

  - Created a new `test-utils` crate, to contain common functions and data structures for the purpose of testing.

  - Moved common functionality that has a wider scope from `storage::common` to `tests-utils::storage`.

### DIFF
--- a/crates/storage-macros/src/lib.rs
+++ b/crates/storage-macros/src/lib.rs
@@ -1,0 +1,290 @@
+use borsh as _;
+/// For documentation links
+use calimero_storage as _;
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{parse_macro_input, parse_quote, Data, DeriveInput, Fields, Type};
+
+#[cfg(test)]
+mod integration_tests_package_usage {
+    use {borsh as _, calimero_storage as _, calimero_test_utils as _, trybuild as _};
+}
+
+/// Derives the [`AtomicUnit`](calimero_storage::entities::AtomicUnit) trait for
+/// a struct.
+///
+/// This macro automatically implements the [`AtomicUnit`](calimero_storage::entities::AtomicUnit)
+/// trait for a struct, which extends the [`Data`](calimero_storage::entities::Data)
+/// trait.
+///
+/// # Requirements
+///
+/// The following are mandatory requirements for the struct:
+///
+///   - A private `storage` field of type [`Element`](calimero_storage::entities::Element).
+///     This is needed as the [`Data`](calimero_storage::entities::Data)-based
+///     struct needs to own an [`Element`](calimero_storage::entities::Element).
+///
+/// # Generated implementations
+///
+/// This macro will generate the following implementations:
+///
+///   - [`Data`](calimero_storage::entities::Data) trait implementation.
+///   - [`AtomicUnit`](calimero_storage::entities::AtomicUnit) trait
+///     implementation.
+///   - Getter and setter methods for each field. These help to ensure that the
+///     access to the fields is controlled, and that any changes to the fields
+///     are reflected in the [`Element`](calimero_storage::entities::Element)'s
+///     state.
+///   - [`BorshSerialize`](borsh::BorshSerialize) and [`BorshDeserialize`](borsh::BorshDeserialize)
+///     will be implemented for the struct, so they should be omitted from the
+///     struct definition.
+///
+/// # Struct attributes
+///
+/// * `#[children]` - An optional attribute to specify the child type for the
+///                   struct, written as `#[children(ChildType)]`. If the
+///                   attribute or its value are omitted, then the child type
+///                   will default to [`NoChildren`](calimero_storage::entities::NoChildren).
+///
+/// # Field attributes
+///
+/// * `#[collection]` - Indicates that the field is a collection of other
+///                     [`Data`] types.
+/// * `#[private]`    - Designates fields that are local-only, and so should not
+///                     be shared with other nodes in the network. These fields
+///                     will not be serialised or included in the Merkle hash
+///                     calculation. Note that being local-only is not the same
+///                     as applying permissions via ACLs to share with only the
+///                     current user — these fields are not shared at all.
+/// * `#[skip]`       - Can be applied to fields that should not be serialised
+///                     or included in the Merkle hash calculation. These fields
+///                     will be completely ignored by the storage system, and
+///                     not even have getters and setters implemented.
+/// * `#[storage]`    - Indicates that the field is the storage element for the
+///                     struct. This is a mandatory field, and if it is missing,
+///                     there will be a panic during compilation. The name is
+///                     arbitrary, but the type has to be an [`Element`](calimero_storage::entities::Element).
+///
+/// Note that fields marked with `#[private]` or `#[skip]` must have [`Default`]
+/// implemented so that they can be initialised when deserialising.
+///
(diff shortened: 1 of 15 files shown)


crates/storage-macros/tests/compile_fail/atomic_unit.stderr
```diff
@@ -0,0 +1,10 @@
+error[E0308]: mismatched types
+  --> tests/compile_fail/atomic_unit.rs:28:30
+   |
+28 |         let _: Vec<Parent> = interface.children_of(&parent).unwrap();
+   |                -----------   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Vec<Parent>`, found `Vec<Child>`
+   |                |
+   |                expected due to this
+   |
+   = note: expected struct `Vec<Parent>`
+              found struct `Vec<Child>`
```
xilosada, line 10: 🤔

crates/storage-macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+/// #[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+/// struct Person {
+///     name: String,
+///     age: u32,
+///     #[private]
+///     secret: String,
+///     #[collection]
+///     friends: Vec<Person>,
+///     #[storage]
+///     storage: Element,
```
miraclx, line 132: so does this mean we're exposing `calimero_storage` to the WASM environment? ..._

crates/storage-macros/src/lib.rs
```diff
@@ -0,0 +1,290 @@
...
+            fn element(&self) -> &calimero_storage::entities::Element {
+                &self.#storage_ident
+            }
+
+            fn element_mut(&mut self) -> &mut calimero_storage::entities::Element {
+                &mut self.#storage_ident
+            }
+        }
+
+        impl calimero_storage::entities::AtomicUnit for #name {}
```
miraclx, line 282: _... This reads like that's the idea. `calimero_storage` won't compile to WASM, nor should we want it to.

But these macro definitions would be used by applications and be compiled to WASM, so you need to bridge between the guest and the host which manages storage.

crates/storage-macros/tests/compile_fail/atomic_unit.stderr
miraclx: yeah, should this be matched for? this isn't a macro-emitted error

danwilliams: Correct, it's not macro-emitted. This is a compilation error test, i.e. ensuring that something that is expecting to fails, fails. It's there to check that the wrong type is not accepted, to ensure that the constraints are correct.

crates/storage-macros/src/lib.rs
danwilliams: I don't know? I guess so? We kinda have to expose something.

crates/storage-macros/src/lib.rs
danwilliams: I guess that can be refined later as I don't really know anything about it 🙂

crates/storage-macros/src/lib.rs
xilosada: @danwilliams  it should behave in the same way than in the transactional model, but using CRDTs instead

crates/storage-macros/tests/compile_fail/atomic_unit.stderr
miraclx: Isn't that already accounted for with the first case?

```rust
let _: Vec<Child> = interface.children_of(&parent).unwrap();
```

if this compiles, it's guaranteed to be a `Vec<Child>` is it not?

crates/storage-macros/src/lib.rs
miraclx: That's one of the things I raised for us to clarify while designing the storage, what's needed to connect the guest to the host are storage entry identifiers, and you mentioned we'd have those so it seemed like a solved problem, now we have them in the form of `Path`s.

So all that's left is to expose `calimero_storage` to the `calimero_runtime`, and allow the guest address storage using `Path`s

So;

```
Guest{WASM} -> Host{Wasmer -> Storage}
```

crates/storage-macros/src/lib.rs
danwilliams: I don't know what that means 😅

crates/storage-macros/tests/compile_fail/atomic_unit.stderr
danwilliams: No, because `Child` implements certain traits and this is a constraint check.

crates/storage-macros/src/lib.rs
danwilliams: That sounds like something to look at later down the line...

crates/storage-macros/src/lib.rs
miraclx: https://github.com/calimero-network/core/pull/764#discussion_r1781443958

crates/storage-macros/src/lib.rs
miraclx: > That sounds like something to look at later down the line...

Well, not quite.. Maybe you can explain how you think of it, perhaps that's where the disconnect is. But the only purpose of this work is to redefine a foundation for the guest state that is easily reconcilable under the asynchronous, optimistically executed model, which affords us the transition from the transactional model.

So without these pieces - exposure to the guest & sync, we can't use it any other way.

danwilliams: So to confirm, exposure to the guest is not part of these changes, and nor is the network/context side.

crates/storage-macros/src/lib.rs
danwilliams: To confirm, neither the `Element` nor the `Collection` should cause any problems if exposed to the guest, and indeed are meant to be. That does not necessarily mean that we are exposing the `Interface`, but everything else is a grey area. The way in which the guest side works with this storage functionality would be a separate piece of work - this is storage layer only.

miraclx (approved):

crates/test-utils/src/storage.rs
miraclx, line 1: not convinced we need a separate module just for this, and seems unnecessary with the new iterations, correct?

danwilliams: That's a good question. I think there's value in having a standalone module for shared test data - it's a common pattern and I can see us expanding upon it. I am not sure how that is affected by the recent iterations, but if you decide at some point that you feel it is unnecessary, it can always be removed 🙂

miraclx: Sorry, just saw this.

While I agree we'll eventually have something like this in the future, potentially while implementing integration/simulation tests for the node, we don't need it introduced right now and it probably won't be an individual crate, but more likely a module in a testing crate.

And as for how it's affected by the recent iterations, I meant it's not used, not anymore, since storage now uses `calimero_sdk` as its storage backend, which either connects to an already initialized instance of rocksdb or for the examples, uses an in-memory store.

PR #765, opened 2024-09-30
feat: Added Collection macro and trait

danwilliams:
Added `Collection` macro and trait

  - Added a new `Collection` trait. This features a `Child: Data` type taken from the `Data` trait, plus `child_ids()` and `has_children()` methods taken from the `Element` struct.

  - Added a `Collection` proc macro to automatically derive the `Collection` trait for a struct.

  - Updated `Interface.children_of()` to accept a `Collection` instead of `Data` implementation.

  - Removed `Child: Data` from the `Data` trait, as this is now handled by the `Collection` trait.

  - Removed the `child_ids` property and the `child_ids()` and `has_children()` methods from `Element`, as these are now handled by the `Collection` trait.

  - Removed the `NoChildren` struct, as this is no longer necessary now that the `Collection` trait is available.

  - Removed the `#[children]` attribute from the `AtomicUnit` proc macro, as this is now handled through the `Collection` macro.

  - Added documentation to explain the design choices behind the implemented structure.

Updated Merkle hashing

  - Added `calculate_merkle_hash()` and `calculate_full_merkle_hash()` to the `Data` trait, abstracting the Merkle hash calculation out to the implementors. This is beneficial for control, but also allows the functionality to be constructed at compile-time with the knowledge that the proc macros have.

  - Updated `Interface.calculate_merkle_hash_for()` to call the new `calculate_full_merkle_hash()` function. This also has a new parameter to pass through, determining whether cached hashes should be used.

  - Updated the `AtomicUnit` macro to implement the new functions.

  - Amended the `AtomicUnit` macro to not update metadata upon change.


miraclx (approved):

PR #768, opened 2024-10-02
feat: Interface sync

danwilliams:
Added `ChildInfo`

  - Added a `ChildInfo` struct to encapsulate basic information about the children of an `Element`, which needs to be stored with the `Element`. At present this is the child `Element`'s ID and Merkle hash.

  - Renamed the `child_ids()` method and `#[child_ids]` field attribute to `child_info`.

  - Amended the `AtomicUnit` implementation of `Interface.calculate_full_merkle_hash()` to only load from the database if recursing, and to otherwise use the now-available stored hash.

Added `Data.collections()`

  - Added a `collections()` method to the `Data` trait, to return a list of all collections with `ChildInfo` for all their children.

Added `Interface.compare_trees()`

  - Added an `Action` enum to represent the various actions that can be required as a result of tree comparison, in order to sync.

  - Added `Interface.compare_trees()` to perform a comparison on tree data and emit local and remote action lists.

  - Added documentation to explain the design choices behind the implemented approach.

Added `Interface.apply_action()`

  - Added an `Interface.apply_action()` method, to carry out CRDT actions.

  - Added entity payload to `Action::Add` and `Action::Update`.

  - Added a new `Interface::find_by_id_raw()` function to find without deserialising, for efficient passthrough (prevents reserialising).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1114,6 +1114,7 @@ dependencies = [
  "sha2",
  "thiserror",
  "uuid 1.10.0",
+ "velcro",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


miraclx (approved):

PR #770, opened 2024-10-03
fix: find rendezvous peer when registration expires

petarjuki7:
## Improved rendezvous peer selection 

Reuse `Expired` peers when no other peers are available


crates/network/src/discovery.rs
```diff
@@ -181,6 +181,53 @@ impl EventLoop {
...
+                        RendezvousRegistrationStatus::Discovered => {
+                            // If we find a Discovered peer, return it right away
+                            return Some(peer_id);
+                        }
+                        RendezvousRegistrationStatus::Expired if candidate.is_none() => {
+                            match self.swarm.dial(peer_id) {
+                                Ok(_) => {
+                                    //Wait for the connection to be established
+                                    tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+                                    if !self.swarm.is_connected(&peer_id) {
```
fbozic: This is not the way it should be done. If you dial you need to finish with the current operation. Dial for unconnected peer will result in the discovery loop, which starts with this event: https://github.com/calimero-network/core/blob/76309531685b9c8bb6cec4d61107a69cdc257cad/crates/network/src/events.rs#L81

If you don't finish the current operation you will have to double process.

crates/network/src/discovery.rs
```diff
@@ -181,6 +181,53 @@ impl EventLoop {
...
+                    match rendezvous_info.registration_status() {
+                        RendezvousRegistrationStatus::Discovered => {
+                            // If we find a Discovered peer, return it right away
+                            return Some(peer_id);
+                        }
+                        RendezvousRegistrationStatus::Expired if candidate.is_none() => {
+                            match self.swarm.dial(peer_id) {
+                                Ok(_) => {
+                                    //Wait for the connection to be established
+                                    tokio::time::sleep(std::time::Duration::from_secs(3)).await;
```
fbozic: We shouldn't `sleep` here.

chefsale (approved):

PR #774, opened 2024-10-07
chore: Removed use of Interface from Data

danwilliams:
- Use of `Interface` has been removed from the `Data` trait. This is to allow a potential further decoupling whereby the `Data` trait might be supplied more independently (under investigation). This is therefore a precursor to a potential removal of hashing operations from the `Data` trait.

  - Removed `Data.calculate_full_merkle_hash()` and moved its logic to `Interface.calculate_merkle_hash_for()`, which previously called it.

  - `Data.calculate_merkle_hash()` has been retained for now, and may be removed in a future update. `Data.calculate_merkle_hash_for_child()` has been added to complement it.

  - Changed `Data.collections()` to return a `BTreeMap` due to the changes in logic, as order is now important in that logic due to the changes made.


miraclx (commented):
why can't we compute the merkle hash on the storage mechanism, why are we doing it as part of the `Data` trait?

`Element` includes `child_ids` which should suffice to recursively fetch all the children in a collection, so this should be accomplishable at that point.

We don't need to depend on the `Data` trait, and hence get into the type-hint problem.

miraclx:
> The purpose of this PR is literally to move the hashing functionality from Data to Interface...

yes, but that's not done

danwilliams:
The blocker raised was about the presence of `Interface` in the `Data` trait. This has been removed, eliminating that blocker. If that is a correct direction then we can also explore how far the actual hashing can be moved. Bear in mind that there is a reliance upon knowledge that the `Data` side of the equation has.

danwilliams:
Also, picking up on this:

> `Element` includes `child_ids` which should suffice to recursively fetch all the children in a collection, so this should be accomplishable at that point.
>
> We don't need to depend on the `Data` trait, and hence get into the type-hint problem.

That's not the case at present, as we need to deserialise in order to use that information.

miraclx:
> Bear in mind that there is a reliance upon knowledge that the Data side of the equation has.

let's start from "what knowledge is needed", and try to itemize why we can't have that at the storage level?

miraclx (approved):

PR #776, opened 2024-10-07
fix(node-ui): context list and details page

frdomovic:
# fix(node-ui): context list and details page

## Summary
As the way we install applications and start context was changed, these pages did not display proper data.
Now we fetch application metadata which has true contract application id saved. 
With this ID we fetch metadata like package / release info.

* Note - the application id we get from node is hash of the "Application" object hashed so it's not the real value we need.

https://github.com/user-attachments/assets/8edefcd0-d570-40c8-998f-4fe3d686a052

Screenshot 2024-10-08 at 00 03 40


frdomovic:
@alenmestrov Build pushed by mistake but will be merged as current workflow doesn't work properly and won't recognise changes.

alenmestrov (approved):

PR #780, opened 2024-10-09
feat(node+storage): use CRDT-backed storage in the runtime

miraclx:
This demonstrates the expectation of the behavior of the `Data` stored in an `Element`, for which a hash can be computed without needing any type hints, using the `sys::storage_{read,write}` methods via the runtime-compat layer to serialize the app state.

In this model, the application is unchanged and serialized as a single atomic unit. I'll demonstrate collections in a coming PR using the new bridge interface discussed in Slack.

### DIFF
--- a/crates/node/src/runtime_compat_v2.rs
+++ b/crates/node/src/runtime_compat_v2.rs
@@ -1,0 +1,53 @@
+use calimero_runtime::store::{Key, Storage, Value};
+use calimero_storage::address::{Id, Path};
+use calimero_storage::entities::Element;
+use calimero_storage::interface::Interface;
+
+#[derive(Debug)]
+pub struct RuntimeCompatStore {
+    pub interface: Interface,
+}
+
+impl RuntimeCompatStore {
+    fn key_as_id(&self, key: &Key) -> Option<Id> {
+        let mut id = [0; 16];
+
+        (key.len() == id.len()).then_some(())?;
+
+        id.copy_from_slice(&key);
+
+        Some(id.into())
+    }
+}
+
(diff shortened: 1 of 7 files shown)


crates/node/src/runtime_compat_v2.rs
```diff
@@ -0,0 +1,53 @@
+use calimero_runtime::store::{Key, Storage, Value};
+use calimero_storage::address::{Id, Path};
+use calimero_storage::entities::Element;
+use calimero_storage::interface::Interface;
+
+#[derive(Debug)]
+pub struct RuntimeCompatStore {
+    pub interface: Interface,
```
fbozic, line 8: Can we please not use `Interface` or `interface`? It provides absolutely no information.

miraclx: I agree, but that's the name as defined in `calimero-storage`, and this is an experiment to see if we can maintain the same programming model for application developers on top of the new CRDT-backed storage.

This won't merge as is

miraclx:
> For now, this doesn't reconcile? It just overrides the current value with the new value?

Yes, this identifies that `calimero-storage` has no temporal store, and so does not prepare a batch of mutations that can be broacasted to other peers, so we can either introduce those bits, or define some other sort of reconciliation strategy.

Anyway, these are experiments, and in my v2 integration approach, I'm mocking out some of those pieces. The goal at the end is to maintain the same app development model but swap out the core storage mechanism. Keep low-level pieces, low-level.

chefsale:
It this PR still valid or can we close it? @miraclx

PR #781, opened 2024-10-09
feat: Presented storage through SDK

danwilliams:
- Changed the `storage` crate to be accessible through the SDK for direct use by guest runtimes.

  - Removed the `store` parameter from `Interface::new()`, as it will now use the central store provided to the guest and available through `env` functions.

  - Removed conversion to/from `StorageKey` for `Id`, as this is no longer accessible due to the move to the SDK.

  - Added mock storage to use during non-WASM testing, where the guest env's storage is not available.

  - Added non-WASM calls into the mocked storage from `env`.

  - Made `Interface` fully static; `Interface` no longer needs to be instantiated, and so use of `self` has been removed, along with the `new()` constructor.

  - Updated storage to use WASM-compatible functions for time and UUID.

### DIFF
--- a/crates/sdk/src/env.rs
+++ b/crates/sdk/src/env.rs
@@ -6,6 +6,8 @@ use borsh::{from_slice as from_borsh_slice, to_vec as to_borsh_vec};
 use uuid::Uuid;
 
 use crate::event::AppEvent;
+#[cfg(not(target_arch = "wasm32"))]
+use crate::mocks::{mock_storage_read, mock_storage_remove, mock_storage_write};
 use crate::state::AppState;
 use crate::sys;
 use crate::sys::{
@@ -154,12 +156,36 @@ pub fn emit<T: AppEvent>(event: &T) {
     unsafe { sys::emit(Event::new(&kind, &data)) }
 }
 
+#[cfg(target_arch = "wasm32")]
 #[inline]
 pub fn storage_read(key: &[u8]) -> Option<Vec<u8>> {
     unsafe { sys::storage_read(Buffer::from(key), DATA_REGISTER) }
         .try_into()
         .unwrap_or_else(expected_boolean::<bool>)
         .then(|| read_register(DATA_REGISTER).unwrap_or_else(expected_register))
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+#[expect(clippy::print_stdout, reason = "Only used in testing")]
+#[must_use]
+pub fn storage_read(key: &[u8]) -> Option<Vec<u8>> {
+    println!("storage_read: {:?}", hex::encode(key));
+    mock_storage_read(key)
+}
+
+#[cfg(target_arch = "wasm32")]
+#[inline]
+pub fn storage_remove(key: &[u8]) -> bool {
+    unsafe { sys::storage_remove(Buffer::from(key), DATA_REGISTER).try_into() }
+        .unwrap_or_else(expected_boolean)
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+#[expect(clippy::print_stdout, reason = "Only used in testing")]
+#[must_use]
+pub fn storage_remove(key: &[u8]) -> bool {
+    println!("storage_remove: {:?}", hex::encode(key));
+    mock_storage_remove(key)
 }
 
 #[must_use]
@@ -171,10 +197,19 @@ pub fn state_read<T: AppState>() -> Option<T> {
     }
 }
 
+#[cfg(target_arch = "wasm32")]
 #[inline]
 pub fn storage_write(key: &[u8], value: &[u8]) -> bool {
     unsafe { sys::storage_write(Buffer::from(key), Buffer::from(value), DATA_REGISTER).try_into() }
         .unwrap_or_else(expected_boolean)
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+#[expect(clippy::print_stdout, reason = "Only used in testing")]
+#[must_use]
+pub fn storage_write(key: &[u8], value: &[u8]) -> bool {
+    println!("storage_write: {:?}", hex::encode(key));
+    mock_storage_write(key, value)
 }
 
 pub fn state_write<T: AppState>(state: &T) {
@@ -182,7 +217,7 @@ pub fn state_write<T: AppState>(state: &T) {
         Ok(data) => data,
         Err(err) => panic_str(&format!("Cannot serialize app state: {err:?}")),
     };
-    let _ = storage_write(STATE_KEY, &data);
+    _ = storage_write(STATE_KEY, &data);
 }
 
 /// Generate a new random UUID v4.
--- a/crates/sdk/src/tests/mocks.rs
+++ b/crates/sdk/src/tests/mocks.rs
@@ -1,0 +1,24 @@
+use std::collections::HashMap;
+use std::sync::{OnceLock, RwLock};
+
+static MOCK_STORAGE: OnceLock<RwLock<HashMap<Vec<u8>, Vec<u8>>>> = OnceLock::new();
+
+fn get_mock_storage() -> &'static RwLock<HashMap<Vec<u8>, Vec<u8>>> {
+    MOCK_STORAGE.get_or_init(|| RwLock::new(HashMap::new()))
+}
+
+pub fn mock_storage_read(key: &[u8]) -> Option<Vec<u8>> {
+    get_mock_storage().read().unwrap().get(key).cloned()
+}
+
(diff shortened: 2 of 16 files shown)


crates/sdk/src/tests/mocks.rs
miraclx, line 1: mocking is a larger and much more involved effort, these pieces can't be reliably used in this way, let's keep them out

crates/sdk/src/env.rs
miraclx, line 1: if you need to test the SDK, you'd have to run it through the runtime

for example: https://github.com/calimero-network/core/blob/master/crates/runtime/examples/demo.rs

crates/storage/src/interface.rs
```diff
@@ -215,16 +215,12 @@ mod tests;
...
 use std::io::Error as IoError;
-use std::sync::Arc;
 
 use borsh::{to_vec, BorshDeserialize, BorshSerialize};
-use calimero_store::key::Storage as StorageKey;
-use calimero_store::layer::{ReadLayer, WriteLayer};
-use calimero_store::slice::Slice;
-use calimero_store::Store;
+use calimero_sdk::env;
+use env::{storage_read, storage_remove, storage_write};
```
miraclx:
```suggestion
use calimero_sdk::env::{storage_read, storage_remove, storage_write};
```

crates/storage/src/interface.rs
```diff
@@ -606,22 +574,8 @@ impl Interface {
...
-        // TODO:     self.store.read()
-        // TODO:         .get(&StorageKey::new((*id).into()))
-        // TODO:         .map_err(StorageError::StoreError)
-        // TODO: }).await.map_err(|err| StorageError::DispatchError(err.to_string()))??;
-        let store = self.store.read();
-        let value = store
-            .get(&StorageKey::new(id.into()))
-            .map_err(StorageError::StoreError)?;
+    pub fn find_by_id<D: Data>(id: Id) -> Result<Option<D>, StorageError> {
+        let value = storage_read(id.to_string().as_bytes());
```
miraclx, line 577: why `to_string` first?

`storage_read` expects a `&[u8]`, `Id::to_bytes` returns a `&[u8; 16]`

miraclx (changes requested):

crates/sdk/src/tests/mocks.rs
danwilliams: These are very, very important. It's not possible to test against the same interface as the WASM has provided to it, so we need an equivalent. Without these mocks, we cannot carry out tests. Therefore, unless there is an alternative way of being able to do this, they need to stay in?

crates/sdk/src/env.rs
danwilliams: Unfortunately, we do need to have unit tests for the storage functionality, and I'm not aware of any other way to do this. Unless you know of one?

crates/storage/src/interface.rs
danwilliams: The key I added to the store previously allows for 16 - this was added in order to cater for these IDs. (Previously as in, previously merged, in #703).

crates/sdk/src/tests/mocks.rs
miraclx: > It's not possible to test

you can run `calimero_runtime::run` in a test, with the assertions inside the guest code

crates/sdk/src/env.rs
miraclx: Write an app that uses these methods and includes assertions for appropriate behaviour, compile it to wasm, run it in the runtime, if the app panics, the runtime will report that, at which point you can handle the panic on the host

crates/storage/src/interface.rs
miraclx: `Id`s are just `Uuid`s which are just `[u8; 16]`, am I missing anything?

crates/sdk/src/tests/mocks.rs
danwilliams: Hmmmm, unless I've misunderstood, that would not be a unit test? It would be an integration test, and one somewhat at arm's length. That would be highly problematic, as the library code needs to test internally, i.e. unit tests.

crates/sdk/src/env.rs
danwilliams: Yes... that's not unit tests, and would not be suitable to use here 🙂 We need tightly-associated unit tests to test the specific functionality that is here. That's not desirable to do through an extended dependency test, and neither is it fully possible (as we want to test internals).

crates/storage/src/interface.rs
danwilliams: Perhaps I misunderstood the question here - you said:

> storage_read expects a &[u8], Id::to_bytes returns a &[u8; 16]

I thought you were asking about the key type...?

crates/sdk/src/tests/mocks.rs
miraclx: yes, the only way to test calimero applications is to test them as calimero applications

crates/sdk/src/env.rs
miraclx: Testing an app that's supposed to be run as a wasm application on the host as a native binary, is pretty much useless.

It doesn't give you any relevant information or test for what really matters.

crates/storage/src/interface.rs
miraclx: I'm saying it can simply be

```suggestion
        let value = storage_read(id.as_bytes());
```

crates/sdk/src/tests/mocks.rs
danwilliams: Okay. That won't be suitable here - we need unit tests for this module's functionality, and this mock is supplying what we cannot get elsewhere. It needs to stay.

crates/sdk/src/env.rs
danwilliams: That's not the point - the point is that the storage layer calls out to the underlying store, using this functionality, and in the absence of the actual store provided to the WASM, we need a way to still run tests and verify the storage layer's functionality.

Unless I have perhaps misunderstood what you are objecting to...? But it seems that you want to remove the mock store, and that's necessary for us to unit test this module.

crates/storage/src/interface.rs
danwilliams: Ah, you wrote two separate things. I don't have a problem with the question raised about that. It was the second part with a seemingly-separate observation on the type and/or length that I was replying to.

crates/sdk/src/env.rs
miraclx: > Write an app that uses these methods and includes assertions for appropriate behaviour, compile it to wasm, run it in the runtime, if the app panics, the runtime will report that, at which point you can handle the panic on the host

crates/storage/src/interface.rs
miraclx: > It was the second part with a seemingly-separate observation on the type and/or length that I was replying to.

that was in reply to yours, indicating that there's no reason why `.to_string()` can't be skipped

> The key I added to the store previously allows for 16 - this was added in order to cater for these IDs. (Previously as in, previously merged, in https://github.com/calimero-network/core/pull/703).

crates/sdk/src/tests/mocks.rs
miraclx: In that case, localize your mocks to `storage`, I saw somewhere else you had a trait for `storage_{read,write,delete}` accesses, use those for your unit test, but we cannot have these mocks in the form you've defined them for the SDK

crates/sdk/src/env.rs
miraclx: Should be resolved by https://github.com/calimero-network/core/pull/781#discussion_r1801073836

crates/storage/src/interface.rs
danwilliams: But we weren't discussing `.to_string()` at any point.

crates/sdk/src/tests/mocks.rs
danwilliams: Can you suggest appropriate changes? We need to ensure that we have proper representation so that we can conduct full unit tests.

crates/storage/src/interface.rs
miraclx: Can you reread this thread?

danwilliams: There has been zero discussion about `.to_string()` 🙂

crates/sdk/src/tests/mocks.rs
miraclx: `StorageAdapter`, which you introduced in https://github.com/calimero-network/core/pull/792, moves in a direction that solves this problem

danwilliams: If you are happy with that later change, great. That means this is resolved here.

miraclx (approved):
approving to unblock, pending comments will be resolved in coming patches

PR #782, opened 2024-10-09
feat: Interface remaining integrations

danwilliams:
- Added `Interface::type_of()`

  - Integrated the new storage system


crates/node/src/types.rs
```diff
@@ -78,17 +61,20 @@ pub enum CatchupError {
     InternalError,
 }
 
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
 #[non_exhaustive]
-pub struct TransactionWithStatus {
```
chefsale, line 83: This should probably stay, as we dont want clients to submit actions and the runtime to generate the actions for the CRDT. This would make the client to generate actions while this has be done inside the guest from my understanding. cc: @miraclx

crates/runtime/examples/crdt.rs
```diff
@@ -0,0 +1,237 @@
...
+    //         "path": "::library"
+    //     }
+    // });
+    let library_data = Library {
+        books: Books {},
+        storage: Element::new(&EntityPath::new("::library")?),
+    };
+    let library_id = library_data.storage.id();
+    
+    let action = Action::Add {
```
chefsale: Should be deleted and runtime should generate the action.

crates/runtime/examples/crdt.rs
```diff
@@ -0,0 +1,237 @@
+#![allow(unused_crate_dependencies, reason = "Not actually unused")]
```
chefsale: This example should be deleted and we should make sure that the kv value example stays the same while working with this new functionality.

crates/node/src/types.rs
danwilliams: Okay so my current understanding is that this should be removed - that seems to have been confirmed as correct 🙂

crates/runtime/examples/crdt.rs
danwilliams: This sounds perfectly sensible to me - when it comes to the examples, I am not entirely sure of what they are meant to be doing, and so moving things around sounds fine 👍

crates/runtime/examples/crdt.rs
danwilliams: I don't think this example should be deleted, as we do need to provide an example of hierarchical CRDT-based data, and that's what this shows modelled out. Otherwise, we don't have anything like this illustrated.

I don't disagree that the other examples should be updated as necessary.

crates/node/src/types.rs
miraclx: Sandi's concern is valid and we've since discussed this on a call.

For posterity, while we should remove the concept of "transactions", clients should maintain the same model for requesting execution (method name like `set` for kv-store or `create_post` for only-peers, for example), during which the "actions" should be generated by the storage subsystem following storage interactions during the execution of the client request.

crates/node/src/action_pool.rs
miraclx, line 1: why do we need this?

crates/node/src/catchup.rs
```diff
@@ -247,87 +144,36 @@ impl Node {
...
-
-                                drop(self.tx_pool.remove(&transaction_hash));
-                            }
-                            NodeType::Coordinator => {
-                                self.persist_transaction(context, transaction, transaction_hash)?;
-                            }
-                            _ => bail!("Unexpected node type"),
-                        },
+                    // TODO: Not clear if this is needed. Emulating the old behaviour for now,
+                    // TODO: but it's likely that an action pool is unnecessary.
```
miraclx: yeah.. let's just apply them directly

crates/node/src/lib.rs
```diff
@@ -1234,21 +1146,15 @@ impl Node {
...
 
     async fn call_mutate(
-        &mut self,
+        &self,
         context: &Context,
-        transaction: Transaction,
-        outcome_sender: oneshot::Sender<Result<Outcome, MutateCallError>>,
-    ) -> Result<Hash, MutateCallError> {
-        if context.id != transaction.context_id {
+        action_list: ActionMessage,
```
miraclx: as mentioned later in the call, actions won't come from the client, so this would just be the (method, args & executor ID)

crates/node/src/lib.rs
```diff
@@ -1268,120 +1174,94 @@ impl Node {
...
-                error!(%err, "Failed to insert transaction into the pool.");
-                MutateCallError::InternalError
-            })?;
-
-        Ok(tx_hash)
+        Ok(())
     }
 
+    // TODO: Should we be using this somewhere in the CRDT approach, or not?
     async fn execute_in_context(
```
miraclx: this was supposed to be called after receiving a coordinator confirmation, it's no longer needed, can be removed

crates/runtime/Cargo.toml
```diff
@@ -17,6 +18,8 @@ ureq.workspace = true
 wasmer.workspace = true
 wasmer-types.workspace = true
 
+calimero-storage = { path = "../storage" }
+
```
miraclx: `calimero-storage` should not be a dependency of the runtime, if you need it for the example, it should be a `dev-dep` instead

crates/runtime/examples/crdt.rs
```diff
@@ -0,0 +1,237 @@
...
+    
+    let action = Action::Add {
+        id: Id::from(library_id),
+        type_id: 11,
+        // data: serde_json::to_vec(&library_data)?,
+        data: borsh::to_vec(&library_data)?,
+        ancestors: Vec::new(),
+    };
+    let serialized_action = serde_json::to_string(&action)?;
+    let input = std::fmt::format(format_args!("{{\"action\": {}}}", serialized_action));
```
miraclx:
```suggestion
    let input = format!("{{\"action\": {}}}", serialized_action);
```

or better yet

```suggestion
    let input = json!({ "action": serialized_action });
```

crates/runtime/examples/crdt.rs
```diff
@@ -0,0 +1,237 @@
...
+    println!("Input: {}", input);
+
+    println!("Action: {serialized_action}");
+    match run(
+        &file,
+        "apply_action",
+        VMContext::new(input.as_bytes().to_owned(), [0; 32]),
```
miraclx: no need to copy

```suggestion
        VMContext::new(input.into_bytes(), [0; 32]),
```

crates/runtime/examples/crdt.rs
chefsale: Im fine with keeping it, but main priority should be making only peers and kv store work with crdts.

crates/node/src/action_pool.rs
danwilliams: I'm not sure we do. I added comments to that file to that effect (as you've noted below). I am not 100% sure of all the changes needed, and wanted to be conservative in what I removed.

crates/node/src/catchup.rs
danwilliams: That sounds fine to me.

crates/node/src/lib.rs
danwilliams: Feel free to make changes as you see fit 🙂

crates/node/src/lib.rs
danwilliams: Sounds good, I agree with that 👍 I wanted to be careful.

crates/runtime/examples/crdt.rs
danwilliams: No issues there - but I will hold off from making any changes here given that I'm not currently working on the examples. I don't want to overlap. The examples are pretty raw and are rough test code only. This PR is still in draft. Ideally review at this stage is for overall approach and direction of changes, and not things at a line level.

crates/node/src/types.rs
danwilliams: That's a mix of different things.

We don't have transactions anymore, which is why they have been removed. They have been replaced by the operations needed for CRDTs. This has been confirmed as correct.

Separately, the method names in examples, i.e. things used by clients, are arbitrary from my perspective and I don't see any issue maintaining those. If that mechanism is required to be a certain way, that's fine. However, that's with you at present I believe.

Yesterday you said that the `Action`s should be created via this `set()` method - that sounds fine to me; we appeared to be on the same page about how and where that should happen. However, if that needs to be moved or refined, I have no objection - feel free! 🙂

crates/runtime/examples/crdt.rs
danwilliams: Just to comment further, from the benefit of having gained additional perspective from the evolving conversation:

As explained, the things happening in this file are manually-created tests. I would not expect the developer to write code like this. Creation of `Action`s is clearly an automated process in the system, but these examples are artificially creating and injecting data for test purposes (only because I was picking up on what Filip was doing and continuing it in hopefully the correct and intended way). This is all a guess on my part, so anything in the examples should change in whatever manner is desired.

crates/node/src/lib.rs
danwilliams: I've removed this.

crates/node/src/catchup.rs
danwilliams: I've removed this.

crates/node/src/action_pool.rs
danwilliams: I've removed this.

crates/node/src/catchup.rs
```diff
@@ -193,24 +86,28 @@ impl Node {
         info!(%context_id, %peer_id, "Interval triggered catchup successfully finished");
     }
 
+    // TODO: Is this even needed now? Can it be removed? In theory, a sync will
+    // TODO: take place so long as there is a comparison - i.e. it will send
+    // TODO: everything back and forth until everything matches. But, for e.g. a
+    // TODO: first-time sync, that would be slower than just sending everything
+    // TODO: all at once. So... could this be utilised for that?
```
danwilliams, line 93: @miraclx just want to make sure this wasn't missed.

crates/node/src/catchup/batch.rs
```diff
@@ -5,13 +5,12 @@ use eyre::Result as EyreResult;
 use futures_util::SinkExt;
 use serde_json::to_vec as to_json_vec;
 
-use crate::types::{
-    CatchupError, CatchupStreamMessage, CatchupTransactionsBatch, TransactionWithStatus,
-};
+use crate::types::{ActionMessage, CatchupActionsBatch, CatchupError, CatchupStreamMessage};
 
+// TODO: Should we be using this somewhere in the CRDT approach, or not?
```
danwilliams, line 10: @miraclx just want to make sure this wasn't missed.

crates/node/src/lib.rs
```diff
@@ -1132,82 +1077,24 @@ impl Node {
...
-                Err(err) => {
-                    drop(request.outcome_sender.send(Err(CallError::Query(err))));
-                }
-            };
-        }
+            CallError::Mutate(MutateCallError::InternalError)
+        })));
     }
 
+    // TODO: Should we be using this somewhere in the CRDT approach, or not?
```
danwilliams: @miraclx just want to make sure this wasn't missed.

miraclx: Seems resolved

crates/node/src/catchup.rs
miraclx: Yeah, I think this makes sense to retain.

> In theory, a sync will take place so long as there is a comparison

this is what initiates that comparison, right? that's what it did in the transactional model anyway

crates/node/src/catchup/batch.rs
miraclx: Yeah, this makes sense to retain.

crates/runtime/src/logic.rs
```diff
@@ -335,6 +336,26 @@ impl VMHostFunctions<'_> {
...
+    ///
+    pub fn send_action(&mut self, action_ptr: u64, action_len: u64) -> VMLogicResult<()> {
+        let action_bytes = self.read_guest_memory(action_ptr, action_len)?;
+        let action: Action = BorshDeserialize::deserialize(&mut &action_bytes[..])
+            .map_err(|_err| HostError::DeserializationError)?;
+
+        println!("Action received: {:?}", action);
+
+        // TODO: What should this now call in order to propagate the Action to other
+        // TODO: nodes in the network? We need to get it into Node::push_action()
```
danwilliams: @miraclx

crates/storage/src/interface.rs
```diff
@@ -1053,6 +1053,11 @@ impl<S: StorageAdaptor> MainInterface<S> {
             }
         };
 
+        // We have to serialise here rather than send the Action itself, as the SDK
+        // has no knowledge of the Action type, and cannot use it as it would be a
+        // circular dependency.
```
miraclx, line 1062: ideally, we define `calimero-storage-primitives` at this point, but this is fine too

crates/runtime/src/logic.rs
miraclx: The rest of the flow would work the way `value_return` / events / logs work.

For example:

https://github.com/calimero-network/core/blob/13cfbd431eeddb4542bb08e32953ac88b408b41f/crates/runtime/src/logic.rs#L302

and eventually, makes it to the `Outcome`

https://github.com/calimero-network/core/blob/13cfbd431eeddb4542bb08e32953ac88b408b41f/crates/runtime/src/logic.rs#L169-L173

which is already the output of `call_query`, for example

https://github.com/calimero-network/core/blob/13cfbd431eeddb4542bb08e32953ac88b408b41f/crates/node/src/lib.rs#L522-L528

miraclx (approved):
Approving, all the necessary pieces to get "local" execution working are in place.

Here's what's left, will merge to #802:

- update the apps to use the new CRDT foundation
	- kv-store: #823 
- publish the root hash following a series of mutations to the node to aid catchup: #824
- refine the node's execution flow: #825
- update the catchup process: https://github.com/calimero-network/core/issues/831
- update JSONRPC API: https://github.com/calimero-network/core/issues/832
- improve developer experience: https://github.com/calimero-network/core/issues/833

miraclx (approved):

miraclx (approved):

miraclx:
oops, lag

PR #792, opened 2024-10-10
wip: Added indexing

danwilliams:
Added Merkle tree indexing

  - Added `index` module to the `storage` crate, to take care of all index management. For now this is registration of the IDs and hashes of nodes in the tree. For any nodes that have children, their child collections are stored in the index data - this means it no longer has to be stored in the `Data` side, reducing the amount of serialised data in the `Action` payloads, and also opening this up for various performance optimisations that can take place in the `Index` without affecting the data side.

  - Added an `Index` struct acting as an interface to the indexing, along with an `EntityIndex` struct for representing the data that gets stored. This is currently the ID, parent ID, child information (the IDs and hashes of all children, by collection), plus the hash of the node-local data and the full hash of the node's data plus its children. This should be sufficient for offloading many operations onto the index, but does not replace primary production of the data hashes, which still needs to occur on the `Data`-implementing type. Equally, knowledge of child collections also still has to come from the type knowledge.

  - Updated `Interface::children_of()` to take a parent ID, to hook in to the new `Index` functionality.

  - Implemented `Interface::find_children_by_id()` now that the `Index` is present to allow it nicely.

  - Added `add_child_to()`, `child_info_for()`, `has_children()`, `parent_of()`, `generate_comparison_data()`, `remove_child_from()`, and `root()` functions to `Interface`, to map through to use the new `Index` functionality.

  - Removed `Collection.child_info()` and `has_children()`, as these are now handled through the `Interface` as planned.

  - Removed `Interface::calculate_merkle_hash_for()` as this is now handled by the `Index`.

  - Added `Data::is_root()` to indicate if a type represents the root of the modelled hierarchy.

  - Added `#[root]` designator to the `AtomicUnit` macro.

  - Added a `ComparisonData` struct to encapsulate the information that gets exchanged between nodes to perform a comparison, now that this is no longer available from the entity directly.

Added `ForeignInterface` functionality

  - Added a `StorageAdaptor` trait to handle abstraction of the storage source, allowing it to be changed.

  - Added a `MainStorage` struct, used by the main Interface, now itself called `MainInterface`, but with a type alias to provide ease of use as `Interface`.

  - Added a `ForeignInterface` with `ForeignStorage`, for use in tests, to allow modelling of a separate data hierarchy representing another node's data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1113,6 +1113,7 @@ dependencies = [
  "fixedstr",
  "hex",
  "indexmap 2.6.0",
+ "serde",
  "sha2",
  "thiserror",
  "uuid 1.10.0",
@@ -1124,6 +1125,7 @@ name = "calimero-storage-macros"
 version = "0.1.0"
 dependencies = [
  "borsh",
+ "calimero-sdk",
  "calimero-storage",
  "quote",
  "syn 2.0.72",
--- a/crates/sdk/src/tests/mocks.rs
+++ b/crates/sdk/src/tests/mocks.rs
@@ -1,24 +1,24 @@
+use core::cell::RefCell;
 use std::collections::HashMap;
-use std::sync::{OnceLock, RwLock};
+use std::thread_local;
 
-static MOCK_STORAGE: OnceLock<RwLock<HashMap<Vec<u8>, Vec<u8>>>> = OnceLock::new();
-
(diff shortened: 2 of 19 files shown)


miraclx (approved):

danwilliams:
This was merged wrongly.

chefsale:
Just revert it?

danwilliams:
@chefsale I fixed it manually

PR #793, opened 2024-10-10
chore: removed duplicated of config and multiaddr verification from t…

chefsale:
Run:

```
cargo run -p meroctl -- --node-name coordinator app install --help
```

or any other command to make sure stuff compiles.

### DIFF
--- a/crates/meroctl/src/cli/context/watch.rs
+++ b/crates/meroctl/src/cli/context/watch.rs
@@ -1,13 +1,12 @@
 use calimero_primitives::context::ContextId;
 use calimero_server_primitives::ws::{RequestPayload, SubscribeRequest};
 use clap::Parser;
-use eyre::{bail, Result as EyreResult};
+use eyre::Result as EyreResult;
 use futures_util::{SinkExt, StreamExt};
 use tokio_tungstenite::connect_async;
 
 use super::RootArgs;
 use crate::common::multiaddr_to_url;
-use crate::config_file::ConfigFile;
 
 #[derive(Debug, Parser)]
 pub struct WatchCommand {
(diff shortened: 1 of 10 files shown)


crates/meroctl/src/common.rs
```diff
@@ -59,6 +61,27 @@ where
...
+    Ok(config)
+}
+
+pub fn load_multiaddr(config: &ConfigFile) -> EyreResult<Multiaddr> {
+    let Some(multiaddr) = config.network.server.listen.first() else {
+        bail!("No address.")
+    };
+
+    Ok(multiaddr.clone())
+}
```
miraclx: I think we can skip this one, especially since the function name isn't very descriptive of what it aids, but if we chose to keep it, we can rename it and change the return type

```suggestion
pub fn load_multiaddr(config: &ConfigFile) -> EyreResult<&Multiaddr> {
    let Some(multiaddr) = config.network.server.listen.first() else {
        bail!("No address.")
    };

    Ok(multiaddr)
```

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -19,18 +18,10 @@ pub struct WatchCommand {
 impl WatchCommand {
     pub async fn run(self, root_args: RootArgs) -> EyreResult<()> {
         let path = root_args.home.join(&root_args.node_name);
+        let config = crate::common::load_config(&path)?;
+        let multiaddr = crate::common::load_multiaddr(&config)?;
```
miraclx: `crate::common::` we can import

miraclx (approved):

xilosada (approved):

crates/meroctl/src/common.rs
chefsale: yes, I will clean this up in the following PRs of method naming, for new lets leave it be until i go through rest of code. Changed to return refernce.

PR #799, opened 2024-10-13
feat: WASM extras

danwilliams:
- Added `storage_remove()`, which was missing.

  - Added a `generate_uuid()` host function. This is necessary because the guest does not have randomness available.

  - Added a `time_now()` host function. This is necessary because the guest does not have randomness available.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,7 +117,6 @@ tracing-subscriber = "0.3.17"
 trybuild = "1.0"
 ureq = "2.9.7"
 url = "2.5.2"
-uuid = { version = "1.10.0", features = ["v4"] }
 velcro = "0.5.4"
 wasmer = "4.2.5"
 wasmer-types = "4.2.5"
@@ -127,6 +126,12 @@ rcgen = "0.13.1"
 x509-parser = "0.16.0"
 axum-server = { version = "0.6", features = ["tls-rustls"] }
 axum-server-dual-protocol = "0.6.0"
+
+[workspace.target.'cfg(target_arch = "wasm32")'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v5"] }
+
+[workspace.target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
 
 [profile.release]
 strip = "symbols"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,6 +117,7 @@ tracing-subscriber = "0.3.17"
 trybuild = "1.0"
 ureq = "2.9.7"
 url = "2.5.2"
+uuid = { version = "1.10.0", features = ["serde"] }
 velcro = "0.5.4"
 wasmer = "4.2.5"
 wasmer-types = "4.2.5"
@@ -126,12 +127,6 @@ rcgen = "0.13.1"
 x509-parser = "0.16.0"
 axum-server = { version = "0.6", features = ["tls-rustls"] }
 axum-server-dual-protocol = "0.6.0"
-
-[workspace.target.'cfg(target_arch = "wasm32")'.dependencies]
-uuid = { version = "1.10.0", features = ["serde", "v5"] }
-
-[workspace.target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-uuid = { version = "1.10.0", features = ["serde", "v4"] }
 
 [profile.release]
 strip = "symbols"
--- a/crates/runtime/Cargo.toml
+++ b/crates/runtime/Cargo.toml
@@ -14,14 +14,9 @@ owo-colors = { workspace = true, optional = true }
 serde = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
 ureq.workspace = true
+uuid = { workspace = true, features = ["serde", "v4"] }
 wasmer.workspace = true
 wasmer-types.workspace = true
-
-[target.'cfg(target_arch = "wasm32")'.dependencies]
-uuid = { version = "1.10.0", features = ["serde", "v4"] }
-
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-uuid = { version = "1.10.0", features = ["serde", "v4"] }
 
 [[example]]
 name = "demo"
--- a/crates/runtime/Cargo.toml
+++ b/crates/runtime/Cargo.toml
@@ -17,6 +17,12 @@ ureq.workspace = true
 wasmer.workspace = true
 wasmer-types.workspace = true
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
+
 [[example]]
 name = "demo"
 
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -2,6 +2,7 @@
 #![allow(clippy::mem_forget, reason = "Safe for now")]
 
 use core::num::NonZeroU64;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use borsh::from_slice as from_borsh_slice;
 use ouroboros::self_referencing;
@@ -150,6 +151,25 @@ impl<'a> VMLogic<'a> {
     #[must_use]
     pub fn generate_uuid(&self) -> Uuid {
         Uuid::new_v4()
+    }
+
+    /// Gets the current time.
+    ///
+    /// This function obtains the current time as a nanosecond timestamp, as
+    /// [`SystemTime`] is not available inside the guest runtime. Therefore the
+    /// guest needs to request this from the host.
+    ///
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "Impossible to overflow in normal circumstances"
+    )]
+    #[expect(clippy::expect_used, reason = "Effectively infallible here")]
+    #[must_use]
+    pub fn time_now(&self) -> u64 {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards to before the Unix epoch!")
+            .as_nanos() as u64
     }
 }
 
@@ -477,6 +497,22 @@ impl VMHostFunctions<'_> {
             logic
                 .registers
                 .set(logic.limits, register_id, *uuid.as_bytes())
+        })?;
+        Ok(())
+    }
+
+    /// Gets the current time.
+    ///
+    /// This function obtains the current time as a nanosecond timestamp, as
+    /// [`SystemTime`] is not available inside the guest runtime. Therefore the
+    /// guest needs to request this from the host.
+    ///
+    pub fn time_now(&mut self, register_id: u64) -> VMLogicResult<()> {
+        let now = self.borrow_logic().time_now();
+        self.with_logic_mut(|logic| {
+            logic
+                .registers
+                .set(logic.limits, register_id, now.to_le_bytes())
         })?;
         Ok(())
     }
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -6,6 +6,7 @@ use core::num::NonZeroU64;
 use borsh::from_slice as from_borsh_slice;
 use ouroboros::self_referencing;
 use serde::Serialize;
+use uuid::Uuid;
 
 use crate::constraint::{Constrained, MaxU64};
 use crate::errors::{FunctionCallError, HostError, Location, PanicContext};
@@ -138,6 +139,17 @@ impl<'a> VMLogic<'a> {
     pub fn get_executor_identity(&mut self, register_id: u64) -> VMLogicResult<()> {
         self.registers
             .set(self.limits, register_id, self.context.executor_public_key)
+    }
+
+    /// Generates a new UUID using v4 standard.
+    ///
+    /// This function generates a new UUID using the v4 standard. The UUID is
+    /// dependent on randomness, which is not available inside the guest
+    /// runtime. Therefore the guest needs to request this from the host.
+    ///
+    #[must_use]
+    pub fn generate_uuid(&self) -> Uuid {
+        Uuid::new_v4()
     }
 }
 
@@ -451,5 +463,21 @@ impl VMHostFunctions<'_> {
 
         self.with_logic_mut(|logic| logic.registers.set(logic.limits, register_id, data))?;
         Ok(status)
+    }
+
+    /// Generates a new UUID using v4 standard.
+    ///
+    /// This function generates a new UUID by calling the host. The UUID is
+    /// dependent on randomness, which is not available inside the guest
+    /// runtime. Therefore the guest needs to request this from the host.
+    ///
+    pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
+        let uuid = self.borrow_logic().generate_uuid();
+        self.with_logic_mut(|logic| {
+            logic
+                .registers
+                .set(logic.limits, register_id, *uuid.as_bytes())
+        })?;
+        Ok(())
     }
 }
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -152,25 +152,6 @@ impl<'a> VMLogic<'a> {
     pub fn generate_uuid(&self) -> Uuid {
         Uuid::new_v4()
     }
-
-    /// Gets the current time.
-    ///
-    /// This function obtains the current time as a nanosecond timestamp, as
-    /// [`SystemTime`] is not available inside the guest runtime. Therefore the
-    /// guest needs to request this from the host.
-    ///
-    #[expect(
-        clippy::cast_possible_truncation,
-        reason = "Impossible to overflow in normal circumstances"
-    )]
-    #[expect(clippy::expect_used, reason = "Effectively infallible here")]
-    #[must_use]
-    pub fn time_now(&self) -> u64 {
-        SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("Time went backwards to before the Unix epoch!")
-            .as_nanos() as u64
-    }
 }
 
 #[derive(Debug, Serialize)]
@@ -492,11 +473,11 @@ impl VMHostFunctions<'_> {
     /// runtime. Therefore the guest needs to request this from the host.
     ///
     pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
-        let uuid = self.borrow_logic().generate_uuid();
         self.with_logic_mut(|logic| {
+            let uuid = logic.generate_uuid();
             logic
                 .registers
-                .set(logic.limits, register_id, *uuid.as_bytes())
+                .set(logic.limits, register_id, uuid.into_bytes())
         })?;
         Ok(())
     }
@@ -507,8 +488,20 @@ impl VMHostFunctions<'_> {
     /// [`SystemTime`] is not available inside the guest runtime. Therefore the
     /// guest needs to request this from the host.
     ///
+    #[expect(
+        clippy::cast_possible_truncation,
+        reason = "Impossible to overflow in normal circumstances"
+    )]
+    #[expect(
+        clippy::expect_used,
+        clippy::unwrap_in_result,
+        reason = "Effectively infallible here"
+    )]
     pub fn time_now(&mut self, register_id: u64) -> VMLogicResult<()> {
-        let now = self.borrow_logic().time_now();
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards to before the Unix epoch!")
+            .as_nanos() as u64;
         self.with_logic_mut(|logic| {
             logic
                 .registers
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -345,6 +345,32 @@ impl VMHostFunctions<'_> {
         Ok(0)
     }
 
+    pub fn storage_remove(
+        &mut self,
+        key_ptr: u64,
+        key_len: u64,
+        register_id: u64,
+    ) -> VMLogicResult<u32> {
+        let logic = self.borrow_logic();
+
+        if key_len > logic.limits.max_storage_key_size.get() {
+            return Err(HostError::KeyLengthOverflow.into());
+        }
+
+        let key = self.read_guest_memory(key_ptr, key_len)?;
+
+        if let Some(value) = logic.storage.get(&key) {
+            self.with_logic_mut(|logic| {
+                drop(logic.storage.remove(&key));
+                logic.registers.set(logic.limits, register_id, value)
+            })?;
+
+            return Ok(1);
+        }
+
+        Ok(0)
+    }
+
     pub fn storage_write(
         &mut self,
         key_ptr: u64,
--- a/crates/runtime/src/logic/imports.rs
+++ b/crates/runtime/src/logic/imports.rs
@@ -60,6 +60,8 @@ impl VMLogic<'_> {
                 body_len: u64,
                 register_id: u64
             ) -> u32;
+
+            fn generate_uuid(register_id: u64);
         }
     }
 }
--- a/crates/runtime/src/logic/imports.rs
+++ b/crates/runtime/src/logic/imports.rs
@@ -47,6 +47,7 @@ impl VMLogic<'_> {
                 register_id: u64,
             ) -> u32;
             fn storage_read(key_ptr: u64, key_len: u64, register_id: u64) -> u32;
+            fn storage_remove(key_ptr: u64, key_len: u64, register_id: u64) -> u32;
 
             fn fetch(
                 url_ptr: u64,
--- a/crates/runtime/src/logic/imports.rs
+++ b/crates/runtime/src/logic/imports.rs
@@ -62,6 +62,7 @@ impl VMLogic<'_> {
             ) -> u32;
 
             fn generate_uuid(register_id: u64);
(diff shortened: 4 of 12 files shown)


crates/runtime/Cargo.toml
```diff
@@ -17,6 +17,12 @@ ureq.workspace = true
 wasmer.workspace = true
 wasmer-types.workspace = true
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
+
```
miraclx: these directives are no longer necessary

crates/runtime/src/logic.rs
```diff
@@ -426,4 +484,36 @@ impl VMHostFunctions<'_> {
...
+
+    /// Generates a new UUID using v4 standard.
+    ///
+    /// This function generates a new UUID by calling the host. The UUID is
+    /// dependent on randomness, which is not available inside the guest
+    /// runtime. Therefore the guest needs to request this from the host.
+    ///
+    pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
+        let uuid = self.borrow_logic().generate_uuid();
+        self.with_logic_mut(|logic| {
```
miraclx: there's no fundamental reason to have a non-mut logic reference in this case, can be simplified to

```suggestion
        self.with_logic_mut(|logic| {
            let uuid = logic.generate_uuid();
```

crates/runtime/src/logic.rs
```diff
@@ -426,4 +484,36 @@ impl VMHostFunctions<'_> {
...
     }
+
+    /// Generates a new UUID using v4 standard.
+    ///
+    /// This function generates a new UUID by calling the host. The UUID is
+    /// dependent on randomness, which is not available inside the guest
+    /// runtime. Therefore the guest needs to request this from the host.
+    ///
+    pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
+        let uuid = self.borrow_logic().generate_uuid();
```
miraclx: although, there's also no reason to have the `VMLogic::generate_uuid` convenience method since this specialized functionality will only be used by this specialized function

```suggestion
        let uuid = Uuid::new_v4();
        self.with_logic_mut(|logic| {
```

crates/runtime/src/logic.rs
```diff
@@ -426,4 +484,36 @@ impl VMHostFunctions<'_> {
...
+    }
+
+    /// Gets the current time.
+    ///
+    /// This function obtains the current time as a nanosecond timestamp, as
+    /// [`SystemTime`] is not available inside the guest runtime. Therefore the
+    /// guest needs to request this from the host.
+    ///
+    pub fn time_now(&mut self, register_id: u64) -> VMLogicResult<()> {
+        let now = self.borrow_logic().time_now();
```
miraclx: same here

```suggestion
        let now = SystemTime::now()
            .duration_since(UNIX_EPOCH)
            .expect("Time went backwards to before the Unix epoch!")
            .as_nanos() as u64
```

crates/runtime/src/logic/imports.rs
```diff
@@ -59,6 +60,9 @@ impl VMLogic<'_> {
                 body_len: u64,
                 register_id: u64
             ) -> u32;
+
+            fn generate_uuid(register_id: u64);
```
miraclx, line 64: `generate_uuid` can be further generalized to `random_bytes::<N>` that returns random bytes of arbitrary length

Knowledge of the fact that it's a `Uuid` stays with storage, neither the SDK nor runtime need that information, and can be constructed by storage from a `[u8; 16]` via [`uuid::Builder::from_random_bytes`](https://docs.rs/uuid/latest/uuid/struct.Builder.html#method.from_random_bytes)

crates/sdk/Cargo.toml
```diff
@@ -14,6 +14,12 @@ serde_json.workspace = true
 
 calimero-sdk-macros = { path = "./macros" }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v5"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
+
```
miraclx: ^

crates/runtime/src/logic.rs
```diff
@@ -426,4 +484,36 @@ impl VMHostFunctions<'_> {
...
+    /// This function generates a new UUID by calling the host. The UUID is
+    /// dependent on randomness, which is not available inside the guest
+    /// runtime. Therefore the guest needs to request this from the host.
+    ///
+    pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
+        let uuid = self.borrow_logic().generate_uuid();
+        self.with_logic_mut(|logic| {
+            logic
+                .registers
+                .set(logic.limits, register_id, *uuid.as_bytes())
```
miraclx:
```suggestion
                .set(logic.limits, register_id, uuid.into_bytes())
```

crates/sdk/src/env.rs
```diff
@@ -181,3 +184,47 @@ pub fn state_write<T: AppState>(state: &T) {
...
+    #[cfg(target_arch = "wasm32")]
+    {
+        unsafe { sys::generate_uuid(DATA_REGISTER) };
+        Uuid::from_slice(&read_register_sized::<16>(DATA_REGISTER).expect("Must have UUID"))
+            .expect("UUID must be valid")
+    }
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        Uuid::new_v4()
+    }
```
miraclx, line 201: these are not necessary, anything that would use `generate_uuid` would at least have already called `state_read -> storage_read`, which would according to https://github.com/calimero-network/core/blob/54336ece92deffa11447c54e1f420fb959f7d727/crates/sdk/src/sys.rs#L48-L50 simply say "not supported"

the same would be maintained of `sys::generate_uuid`, so the mocks are not needed

crates/sdk/src/env.rs
```diff
@@ -181,3 +184,47 @@ pub fn state_write<T: AppState>(state: &T) {
...
+        clippy::cast_possible_truncation,
+        reason = "Impossible to overflow in normal circumstances"
+    )]
+    #[expect(clippy::expect_used, reason = "Effectively infallible here")]
+    {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Time went backwards to before the Unix epoch!")
+            .as_nanos() as u64
+    }
```
miraclx, line 229: same here

crates/storage/Cargo.toml
```diff
@@ -15,10 +15,15 @@ indexmap.workspace = true
...
-uuid.workspace = true
 
 calimero-store = { path = "../store", features = ["datatypes"] }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v5"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+uuid = { version = "1.10.0", features = ["serde", "v4"] }
+
```
miraclx: ^

miraclx (changes requested):

crates/runtime/Cargo.toml
danwilliams: Could you explain? 🙂 At present these are here to ensure that we can achieve the same functionality both through WASM and not. If these were removed, what would we do instead?

crates/runtime/src/logic.rs
danwilliams: That's a fair observation 👍 Done! 🙂

crates/runtime/src/logic.rs
danwilliams: Ahhhh - no; this is actually used a little more widely. I think it's best we keep this calling `generate_uuid()` as that is then consistent with the rest of the usage.

crates/runtime/src/logic.rs
danwilliams: Agreed - calling `logic.time_now()` here is better. Additionally, in this location we can indeed subsume the `logic.time_now()` call into this location, as it's the only time we use it. So although it was added to try to preserve consistency with the other, similar implementations, there is no special need for a separate function here. So this has been removed 🙂 👍

crates/runtime/src/logic/imports.rs
danwilliams: That's an interesting thought. Yes, that could potentially be done, and I can see your reasoning for that. However, in this situation we don't have a need for arbitrary length, and also this would not fit neatly with the UUID functionality. It's definitely a potential for later, but I've left this as-is for now 🙂

crates/sdk/Cargo.toml
danwilliams: I will need to know a little more about your intentions here 🙂

crates/sdk/src/env.rs
danwilliams: So I think the purpose here has been misunderstood... and the premise seems incorrect. If I remove this, then the build script no longer works:

```
error[E0432]: unresolved import `calimero_sdk::env::time_now`
   --> crates/storage/src/entities.rs:219:5
    |
219 | use calimero_sdk::env::time_now;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `time_now` in `env`
    |
note: found an item that was configured out
   --> /Users/dan/Code/calimero/crates/sdk/src/env.rs:233:8
    |
233 | pub fn time_now() -> u64 {
    |        ^^^^^^^^
note: the item is gated behind the `wasm32` feature
   --> /Users/dan/Code/calimero/crates/sdk/src/env.rs:230:7
    |
230 | #[cfg(target_arch = "wasm32")]
    |       ^^^^^^^^^^^^^^^^^^^^^^
```

It may be possible to do what you're suggesting in a different way, but I'm not sure how, and what's in place currently works.

crates/sdk/src/env.rs
danwilliams: See below example with `time_now()`, as that's unaffected by the wider question around UUID type. For this specific context, the same applies - i.e. cannot remove this.

crates/runtime/Cargo.toml
miraclx: they're the same thing, just merge it into `[dependencies]`

crates/runtime/src/logic.rs
miraclx: > this is actually used a little more widely

by?

crates/runtime/src/logic.rs
miraclx: > Additionally, in this location we can indeed subsume the logic.time_now() call into this location

not necessarily, I like to think of these like `unsafe {}` (it technically is behind the scenes too), limit the scope of the closure to the use of `logic` and time doesn't use `logic`, so it can be outside

the same way unsafe things should be outside unsafe blocks even though unsafe things use their outputs

crates/runtime/src/logic/imports.rs
miraclx: The concern is, by the time this ships, we'll have a contract to uphold between built developer apps and the interfaces the runtime supports, where there's no reason to.

The ideal approach to designing a bridge is to be liberal within boundaries defined with good reason, and in this case, the generic approach doesn't incur any extra cost.

Randomness has always been a feature we wanted to introduce, it just had to be done safely which is only possible by optimistic execution. As this is the effort that introduces optimistic execution, and a component of its - `uuid` needs randomness, then we have to provide the fundamental facility for randomness.

Otherwise, what's likely to happen after this merges as is, is developers need random numbers and abuse `generate_uuid` to get that. Calling `generate_uuid` N times to get `16N` random bytes and not using it at all for uuids.

crates/sdk/Cargo.toml
miraclx: same as the snippet in `runtime`, this directive can be merged into `[dependencies]`, although following https://github.com/calimero-network/core/pull/799#discussion_r1800168580, it won't be necessary to have the dependency in the SDK at all

crates/sdk/src/env.rs
miraclx: I'm confused, the point of this comment is to highlight that feature gating for wasm32 is unnecessary.

crates/sdk/src/env.rs
miraclx: Replied below

crates/storage/Cargo.toml
miraclx: move to `[dependencies]`, unless you actually need that `v5` feature activated, in which case, I can't find any examples of it's use

crates/runtime/Cargo.toml
danwilliams: Ahhh, I see. Yes - I had overlooked that in this location they ended up being the same. Well-spotted!

crates/runtime/src/logic.rs
danwilliams: If you search the codebase for the function, you'll see it's used in a fair few places 🙂

crates/sdk/Cargo.toml
danwilliams: Ahhh, I see. No - they are not the same thing. The WASM code cannot have the v4. In the other snippet, they were indeed the same, and I had overlooked that they had become so. Which means, that was a valid observation. But here, they are not the same.

crates/sdk/src/env.rs
danwilliams: Yes, I understand the point of the comment, but as illustrated above, that's incorrect. If you attempt what you have described, then the build fails. Therefore this is necessary.

crates/storage/Cargo.toml
danwilliams: As per above, they are not the same - the WASM cannot have the v4.

crates/runtime/src/logic.rs
danwilliams: > > Additionally, in this location we can indeed subsume the logic.time_now() call into this location
> 
> not necessarily

I am confused - this was your suggestion? 😄 I was agreeing with you that this could indeed happen here, and I've made that change.

If the change is incorrect, can you explain what you intend more clearly?

If the focus is about the logic, then the move inside agrees with your comments.

If the focus is about what's unsafe or not, then that's the way I had previously approached it, in both locations.

crates/runtime/src/logic/imports.rs
danwilliams: Hmmmm, I am not sure that there is actually any effect from that, i.e. I do not think the outcome varies.

Let's consider: In a situation where you ask for random bytes, returning `[u8; 16]`, then you get random bytes and can do that as many times as you like. Ergo, the situation you describe as being undesirable. It's the same outcome.

However, in this case we are using semantic identification to recognise that this is a UUID. So it's more constrained than a random number generator of arbitrary length, but also is specific to a purpose of providing UUIDs.

Yes they could call it umpteen times - but they could do this in both cases. What's the difference?

Therefore, on balance, it seems better to retain this as a UUID type, because:

  - There is no difference in randomness or ability to call it multiple times
  - There is semantic identification of the type in use
  - The approach taken to generate a UUID is a specific algorithm and not directly equivalent to an arbitrary randomness generator
  - The exact approach we take to generate UUIDs is abstracted away, and we are free to change it in future without it being tied to arbitrary randomness

Hence it would appear that using a UUID here, i.e. the `Uuid` type, is the correct choice - and even more importantly, that using the UUID functionality for generation and not an arbitrary randomness generator is also correct.

From a contract perspective, the only contract we are formally providing is that an `Id` performs in a particular way. We are free to change `Id` internals later without breaking that contract.

crates/runtime/src/logic.rs
miraclx: I did, and I couldn't find any references, which is why I asked, `VMLogic` isn't to be used by anything but `calimero_runtime::run` anyway, so it can't be used by anything else

crates/runtime/src/logic.rs
miraclx: > If the change is incorrect
> If the focus is about the logic, then the move inside agrees with your comments.

so `logic` is just a variable, `with_logic_mut` is a safe way of fetching it because it's self-referencing and that requires `unsafe`

so it's not incorrect, it just makes sense to limit the scope of the closure to things that _actually need to use the variable_

crates/runtime/src/logic/imports.rs
miraclx: > In a situation where you ask for random bytes, returning [u8; 16], then you get random bytes and can do that as many times as you like. Ergo, the situation you describe as being undesirable. It's the same outcome.

as mentioned earlier:

> `generate_uuid` can be further generalized to `random_bytes::<N>` that returns random bytes of arbitrary length

so it's not the same

> The approach taken to generate a UUID is a specific algorithm and not directly equivalent to an arbitrary randomness generator

there is no "specific algorithm" for UUID v4, the only requirement is you have sufficiently random bytes, and appropriately set the version and variant when making those random bytes into a UUID

> The exact approach we take to generate UUIDs is abstracted away, and we are free to change it in future without it being tied to arbitrary randomness

ideally, if you don't trust a randomness source to be cryptographically secure, you'd pass that as a seed to a CSPRNG like ChaCha. That's what [`StdRng`](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html) does for example.

> From a contract perspective, the only contract we are formally providing is that an Id performs in a particular way. We are free to change Id internals later without breaking that contract.

that's maintained

crates/sdk/Cargo.toml
miraclx: it doesn't need the `v5` either, it just needs to call `generate_uuid` which returns a UUID, so you don't need this directive

just make it

```toml
[dependencies]
uuid.workspace = true
```

but IMO (https://github.com/calimero-network/core/pull/799#discussion_r1800168580), this dependency doesn't even need to exist in the SDK at all

crates/sdk/src/env.rs
miraclx: The build fails because you still have a feature flag 🤷🏽‍♂️

crates/storage/Cargo.toml
miraclx: Replied above, it doesn't need to, yet this still stands

crates/runtime/src/logic.rs
danwilliams: Okay, so does that mean that this is resolved...? I am not sure 😅

crates/runtime/src/logic/imports.rs
danwilliams: > there is no "specific algorithm" for UUID v4, the only requirement is you have sufficiently random bytes, and appropriately set the version and variant when making those random bytes into a UUID

That's self-contradicting 🙂 I'm glad you agree with me that v4 UUIDs are not the same as a set of random bytes. As you've described, they have a version and variant built in, which makes them functionally different.

I believe there's no change to make here.

crates/runtime/src/logic.rs
miraclx: Localizing the call to this location without involving `VMLogic`, yes.

However, the `let now = _` statement should be outside the closure.

crates/runtime/src/logic/imports.rs
miraclx: > That's self-contradicting 🙂 I'm glad you agree with me that v4 UUIDs are not the same as a set of random bytes. As you've described, they have a version and variant built in, which makes them functionally different.

it's not, as mentioned earlier, to construct a v4 uuid with self-provided random bytes, you'd use [`uuid::Builder::from_random_bytes`](https://docs.rs/uuid/latest/uuid/struct.Builder.html#method.from_random_bytes) which proceeds to set the version and variant.

Everything else is random. That's what matters.

crates/sdk/Cargo.toml
danwilliams: It's not possible to do `uuid.workspace = true` here, as it doesn't exist in the workspace dependencies - it can't, due to this exact issue of arch targeting.

No problem removing the v5, although that was consciously there for use in tests, but that's not strictly necessary and not currently used.

crates/sdk/src/env.rs
danwilliams: That's the point, isn't it? In order to build, this needs to be available 🤔

crates/runtime/src/logic.rs
danwilliams: Thank you for confirming 🙂

crates/runtime/src/logic/imports.rs
danwilliams: In order to do that, you still need to expose a source of random bytes, though, even if you are then feeding them into UUID.

Initially, I thought you were talking about returning a `[u8; 16]` of random bytes. I objected to that for obvious reasons.

However, if you are saying you are constructing an actual UUID type then you still need that type available, and you also still need a source of randomness.

I don't see any change to make here - what is already present works, and does what it should, and I'm not sure I even understand fully what you are intending to do, as your objectives are not totally clear to me.

crates/sdk/Cargo.toml
miraclx: > It's not possible to do uuid.workspace = true here, as it doesn't exist in the workspace dependencies

it can be introduced

> it can't, due to this exact issue of arch targeting.

not true, with the version specified in the root manifest, you could do this in the local crate's

```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
uuid = { workspace = true, features = ["serde", "v5"] }

[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
uuid = { workspace = true, features = ["serde", "v4"] }
```

crates/sdk/src/env.rs
miraclx: and that's the point, it doesn't need it, you don't need to accommodate for anything but wasm

crates/runtime/src/logic/imports.rs
miraclx: > In order to do that, you still need to expose a source of random bytes, though, even if you are then feeding them into UUID.

that's the idea, as mentioned in the original comment

> However, if you are saying you are constructing an actual UUID type then you still need that type available, and you also still need a source of randomness.

as mentioned, not only are we NOT constructing a UUID type at this level, we do not need to even have that dependency in the runtime nor SDK

> I don't see any change to make here

`random_bytes::<N>` for runtime & SDK, `uuid::Builder::from_random_bytes` for storage

> and I'm not sure I even understand fully what you are intending to do, as your objectives are not totally clear to me.

https://github.com/calimero-network/core/pull/799#discussion_r1800312350, not sure how else to explain

danwilliams: Okay, so I think I should recap my previous points:

> Therefore, on balance, it seems better to retain this as a UUID type, because:
> 
>   - There is no difference in randomness or ability to call it multiple times
>   - There is semantic identification of the type in use
>   - The approach taken to generate a UUID is a specific algorithm and not directly equivalent to an arbitrary randomness generator
>   - The exact approach we take to generate UUIDs is abstracted away, and we are free to change it in future without it being tied to arbitrary randomness
> 
> Hence it would appear that using a UUID here, i.e. the Uuid type, is the correct choice - and even more importantly, that using the UUID functionality for generation and not an arbitrary randomness generator is also correct.

I think the best path here is to merge this and then any subsequent changes can be carried out as you see fit - I do not see any harm in UUID being available; the user is able to do the same thing in both cases, and having the actual type present is important and beneficial.

crates/sdk/Cargo.toml
danwilliams: Okay, so that's fair about introducing it into the main workspace - my preference was not to do that, as I like it being obvious to have the arch matches together, but it's also fine having it in the main list and then overridden 👍

crates/sdk/src/env.rs
danwilliams: For whatever reason, if that is removed, the build script fails.

miraclx: what does it look like when it's removed?

Cargo.toml
```diff
@@ -117,6 +117,7 @@ tracing-subscriber = "0.3.17"
 trybuild = "1.0"
 ureq = "2.9.7"
 url = "2.5.2"
+uuid = { version = "1.10.0", features = ["serde"] }
```
miraclx, line 120: the root manifest should just define versions which can be inherited with no default feature activations unless with good reason (security or other similar constraint)

Cargo.toml
```diff
@@ -127,11 +128,8 @@ x509-parser = "0.16.0"
 axum-server = { version = "0.6", features = ["tls-rustls"] }
 axum-server-dual-protocol = "0.6.0"
 
-[workspace.target.'cfg(target_arch = "wasm32")'.dependencies]
-uuid = { version = "1.10.0", features = ["serde", "v5"] }
-
 [workspace.target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-uuid = { version = "1.10.0", features = ["serde", "v4"] }
+uuid = { workspace = true, features = ["serde", "v4"] }
```
miraclx: we don't need this in the root

crates/sdk/src/env.rs
danwilliams: You know more about this side than I do - I've only recently started fiddling with it, in order to try to test this stuff. I wrote what I did in order for it to fully work. Removing what you've suggested breaks it in the way described. I'd say the best path forward at this stage is for you to make the changes that you are thinking of, as it may be that I've misunderstood you.

Cargo.toml
danwilliams: The root manifest here defines the features that all usages of this crate need, i.e. serde. That is used everywhere that the crate is used, and is in place as a default.

fbozic: I agree with mira on this one, root manifest should just define versions

fbozic: Could you also clean up `fixedstr`, `uuid` and `axum-server` deps? I just saw they are not following the pattern. Thanks

miraclx (approved):
approving to unblock, pending comments will be resolved in coming patches

PR #800, opened 2024-10-14
feat: Add ancestor hashes and type IDs

danwilliams:
Added type ID to entities

  - Added a `type_id()` method to the `Data` trait, to provide a unique type ID per implemented `Data` type, formalising the approach to type identification.

  - Added a `type_id` parameter to `Index::add_child_to()` and `Index::root()`.

Added ancestor hashes to `Action`

  - Added ancestor information to `Action::Add`, `Delete`, and `Update`. This allows the ancestor hashes to be checked after applying an action, to see if further syncing is needed.

  - Added an `ancestors` field to `ComparisonData`. This will be useful for checking the higher tree levels in some cases.

  - Added `get_ancestors_of()` and `get_type_id()` to `Index`. The type ID is now formalised as a `u8`.

  - Changed the `Action` `id` field to be a `u8`, to match `type_id`.

  - Made the Action variants into struct variants for greater clarity.

Changed `apply_action()` to check ancestors

  - Amended the behaviour of `Interface::apply_action()` to check the recalculated ancestor hashes of an updated entity after applying the update, and indicate if any of the ancestors need re-syncing as a result (based on the expectation that the hashes will all match).

### DIFF
--- a/crates/storage/src/entities.rs
+++ b/crates/storage/src/entities.rs
@@ -237,6 +237,7 @@ use crate::interface::StorageError;
 /// use calimero_storage_macros::AtomicUnit;
 ///
 /// #[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+/// #[type_id(43)]
 /// struct Page {
 ///     title: String,
 ///     #[private]
@@ -261,6 +262,7 @@ pub trait AtomicUnit: Data {}
 /// use calimero_storage::entities::{ChildInfo, Data, Element};
 ///
 /// #[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+/// #[type_id(42)]
 /// struct Book {
 ///     title: String,
 ///     pages: Pages,
@@ -273,6 +275,7 @@ pub trait AtomicUnit: Data {}
 /// struct Pages;
 ///
 /// #[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+/// #[type_id(43)]
 /// struct Page {
 ///     content: String,
 ///     #[storage]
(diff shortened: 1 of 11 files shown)


crates/storage/src/entities.rs
```diff
@@ -261,6 +262,7 @@ pub trait AtomicUnit: Data {}
 /// use calimero_storage::entities::{ChildInfo, Data, Element};
 ///
 /// #[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+/// #[type_id(42)]
```
miraclx, line 265: developers have to manage this count themselves?

crates/storage/src/index.rs
```diff
@@ -33,6 +33,12 @@ struct EntityIndex {
     /// Merkle hash of the entity's immediate data only. This gets combined with
     /// the hashes of its children to form the full hash.
     own_hash: [u8; 32],
+
+    /// Type identifier of the entity. This is noted so that the entity can be
+    /// deserialised correctly in the absence of other semantic information. It
+    /// is intended that the [`Path`](crate::address::Path) will be used to help
+    /// with this at some point, but at present paths are not fully utilised.
```
miraclx, line 40: I was going to mention that, paths are almost entirely useless atm, `find_by_path` is still unimplemented and it's unclear how to use paths to do anything

miraclx (commented):
having these user-managed is a questionable pattern

crates/storage/src/entities.rs
danwilliams: At present yes. As I've indicated in the code, some areas are longhand for now, to establish working functionality, and can be tided into macros later. However, this part will likely never become fully automatic, as there needs to be some independent consistency in order to maintain the same association throughout various code changes. So I think this will be as it currently is, or an approach that is similar enough, i.e. the type ID needs to be statically established somewhere.

crates/storage/src/index.rs
danwilliams: You are correct. Paths are a fundamental part of the design, and important to certain aspects of the design, but as yet there are certain key questions about their usage that are outstanding, to be resolved. As such, given the fact that we've decided to go with a primarily ID-based approach in the first instance, the paths are sat waiting in the wings somewhat.

miraclx (approved):
approving to unblock, will test these out in the rollup PR

PR #801, opened 2024-10-14
Added JSON output to cli commands

petarjuki7:
You can now add the `--test` or `-t` flag to `meroctl` commands and receive outputs in JSON format.


miraclx (commented):
why not `--json`, `--format json` or something of the sort?

petarjuki7:
> why not `--json`, `--format json` or something of the sort?

Sure, makes sense

crates/meroctl/src/cli/app/get.rs
```diff
@@ -12,6 +13,9 @@ pub struct GetCommand {
 
     #[arg(long, short)]
     pub app_id: String,
+
+    #[arg(long, short)]
+    pub test: bool,
```
fbozic: Can we make this global flag for all commands maybe? It would also enforce future commands to have output. Ideally each command should return structured output and single handler would serialize output based on the global flag.

fbozic: This should also standardize the output of all commands across all formats.

chefsale:
Status on this PR, is it still valid?

chefsale:
To rephrase this question then: Is this still in progress and whats the ETA? :)

petarjuki7:
This PR increased in scope a bit so I'm going to explain the changes.

1.  Added `context update` command
2.  The functions in the `context create` file now use the commands from the `app` part of the cli, it made sense to me to reuse the code because the way it was before seemed like duplicating a lot of the code
3. Most important change, removed `eyre` from most of the crates, now we have a custom error type. Influenced by the way it was done in [ws.rs](https://github.com/calimero-network/core/blob/master/crates/server/src/ws.rs#L351-L377) and [jsonrpc.rs](https://github.com/calimero-network/core/blob/master/crates/server/src/jsonrpc.rs#L114-L140), we push the responses to the top level and then print out just one JSON

Open to feedback on `error` naming and different kinds of errors, the way it is right now is mostly influenced by the `ws.rs` file, so everything may not fit perfectly into the cli

cc: @miraclx  @fbozic

crates/meroctl/src/cli/app/install.rs
```diff
@@ -30,22 +30,34 @@ pub struct InstallCommand {
...
+            let install_dev_request = InstallDevApplicationRequest::new(
+                app_path
+                    .canonicalize_utf8()
+                    .map_err(|_| CliError::InternalError(format!("Canonicalize path failed")))?,
+                metadata,
+            );
             is_dev_installation = true;
-            serde_json::to_value(install_dev_request)?
+            serde_json::to_value(install_dev_request)
+                .map_err(|e| CliError::InternalError(format!("{}", e.to_string())))?
```
miraclx, line 47:
```suggestion
                .map_err(|e| CliError::InternalError(e.to_string()))?
```

et al.

crates/meroctl/src/cli/app/install.rs
```diff
@@ -30,22 +30,34 @@ pub struct InstallCommand {
...
         let mut is_dev_installation = false;
         let metadata = self.metadata.map(String::into_bytes).unwrap_or_default();
 
         let install_request = if let Some(app_path) = self.path {
-            let install_dev_request =
-                InstallDevApplicationRequest::new(app_path.canonicalize_utf8()?, metadata);
+            let install_dev_request = InstallDevApplicationRequest::new(
+                app_path
+                    .canonicalize_utf8()
+                    .map_err(|_| CliError::InternalError(format!("Canonicalize path failed")))?,
```
miraclx, line 42:
```suggestion
                    .map_err(|_| CliError::InternalError("Canonicalize path failed".to_owned()))?,
```

et al.

crates/meroctl/src/cli/context/create.rs
```diff
@@ -76,56 +80,55 @@ impl CreateCommand {
...
     context_seed: Option<Hash>,
     application_id: ApplicationId,
     params: Option<String>,
     keypair: &Keypair,
-) -> EyreResult<ContextId> {
-    if !app_installed(base_multiaddr, &application_id, client, keypair).await? {
-        bail!("Application is not installed on node.")
+    args: &RootArgs,
+) -> Result<CreateContextResponse, CliError> {
+    if app_installed(&application_id, &args).await.is_err() {
```
miraclx, line 128: quick check: is it supposed to return an error if the app doesn't exist?

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -15,39 +15,49 @@ pub struct WatchCommand {
...
+            .await
+            .map_err(|e| CliError::MethodCallError(e.to_string()))?;
 
         while let Some(message) = read.next().await {
             match message {
                 Ok(msg) => {
                     if let tokio_tungstenite::tungstenite::Message::Text(text) = msg {
-                        println!("{}", text);
+                        println!(
+                            "{:#?}",
```
miraclx, line 50: bummer, these won't fit the model

miraclx (commented):
`context watch` not being able to conform to the rest of the app suggests perhaps a better model would be similar to how rust does it

```
$ cargo build -p meroctl --message-format json
{ json message 1 }
{ json message 2 }
{ json message 3 }
```

anything that produces a console report would emit a report instead of returning one

depending on the presence of the `--format` flag, we can either catch that report and print a pretty console output or print the JSON serialized form of the individual messages

https://docs.rs/cargo_metadata/latest/cargo_metadata/index.html

crates/meroctl/src/cli/context/create.rs
petarjuki7: Yes, if the user passes an `app_id` when creating a context, it should already be installed so we return an error

MatejVukosav:
Please add this to user docs when ready

PR #807, opened 2024-10-16
fix: load config should factor in node's home dir

miraclx:
The `load_config` refactor in #793 lost some information, notably no longer loading the config from the node's home dir, the effect of which only works when the executor's CWD is inside the node's home dir

### DIFF
--- a/crates/meroctl/src/cli.rs
+++ b/crates/meroctl/src/cli.rs
@@ -60,7 +60,7 @@ pub struct RootArgs {
 
     /// Name of node
     #[arg(short, long, value_name = "NAME")]
-    pub node_name: Utf8PathBuf,
+    pub node_name: String,
 }
 
 impl RootCommand {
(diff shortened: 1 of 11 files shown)


crates/meroctl/src/cli.rs
```diff
@@ -60,7 +60,7 @@ pub struct RootArgs {
 
     /// Name of node
     #[arg(short, long, value_name = "NAME")]
-    pub node_name: Utf8PathBuf,
+    pub node_name: String,
```
miraclx, line 63: are there any arguments in favor of this remaining a path?

frdomovic:
Verifing this works.

`cargo run -p meroctl -- --home data --node-name node1 context create --watch ./apps/only-peers/res/only_peers.wasm`
Screenshot 2024-10-16 at 11 50 31

crates/meroctl/src/cli.rs
frdomovic: nope

frdomovic (approved):

PR #808, opened 2024-10-16
fix: incorporate review feedback from #799

miraclx:
1. remove `VMLogic::generate_uuid` - [ref](https://github.com/calimero-network/core/pull/799#discussion_r1800159587)
2. remove `uuid` dep from runtime & SDK - [ref](https://github.com/calimero-network/core/pull/799#discussion_r1800313072)
3. generalize `sys::generate_uuid` to `sys::random_bytes` - [ref](https://github.com/calimero-network/core/pull/799#discussion_r1800168580)
4. remove mocks for SDK functionality - ref [1](https://github.com/calimero-network/core/pull/799#discussion_r1800172131) & [2](https://github.com/calimero-network/core/pull/799#discussion_r1800172227)
	- as mentioned in https://github.com/calimero-network/core/pull/781#discussion_r1801073836, this will be accommodated inside the storage crate
5. clean up `uuid` in root manifest - [ref](https://github.com/calimero-network/core/pull/799#discussion_r1801168500)
	- the others - `fixedstr`, `axum-server` will be addressed in a separate PR targeting `master`

additionally, this includes optimizations for sized data exchanged between the runtime and the SDK, skipping register heap allocations & copying

### DIFF
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -6,8 +6,8 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use borsh::from_slice as from_borsh_slice;
 use ouroboros::self_referencing;
+use rand::RngCore;
 use serde::Serialize;
-use uuid::Uuid;
 
 use crate::constraint::{Constrained, MaxU64};
 use crate::errors::{FunctionCallError, HostError, Location, PanicContext};
@@ -458,20 +458,14 @@ impl VMHostFunctions<'_> {
         Ok(status)
     }
 
-    /// Generates a new UUID using v4 standard.
-    ///
-    /// This function generates a new UUID by calling the host. The UUID is
-    /// dependent on randomness, which is not available inside the guest
-    /// runtime. Therefore the guest needs to request this from the host.
-    ///
-    pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
-        let uuid = Uuid::new_v4();
+    pub fn random_bytes(&mut self, len: u64, register_id: u64) -> VMLogicResult<()> {
+        let len = usize::try_from(len).map_err(|_| HostError::IntegerOverflow)?;
 
-        self.with_logic_mut(|logic| {
-            logic
-                .registers
-                .set(logic.limits, register_id, uuid.into_bytes())
-        })?;
+        let mut buf = vec![0; len];
+
+        rand::thread_rng().fill_bytes(&mut buf);
+
+        self.with_logic_mut(|logic| logic.registers.set(logic.limits, register_id, buf))?;
 
         Ok(())
     }
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -458,14 +458,14 @@ impl VMHostFunctions<'_> {
         Ok(status)
     }
 
-    pub fn random_bytes(&mut self, len: u64, register_id: u64) -> VMLogicResult<()> {
+    pub fn random_bytes(&mut self, ptr: u64, len: u64) -> VMLogicResult<()> {
         let len = usize::try_from(len).map_err(|_| HostError::IntegerOverflow)?;
 
         let mut buf = vec![0; len];
 
         rand::thread_rng().fill_bytes(&mut buf);
 
-        self.with_logic_mut(|logic| logic.registers.set(logic.limits, register_id, buf))?;
+        self.borrow_memory().write(ptr, &buf)?;
 
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


chefsale (approved):

crates/runtime/src/logic.rs
```diff
@@ -466,19 +455,15 @@ impl VMHostFunctions<'_> {
...
-    pub fn generate_uuid(&mut self, register_id: u64) -> VMLogicResult<()> {
-        self.with_logic_mut(|logic| {
-            let uuid = logic.generate_uuid();
-            logic
-                .registers
-                .set(logic.limits, register_id, uuid.into_bytes())
-        })?;
+    pub fn random_bytes(&mut self, ptr: u64, len: u64) -> VMLogicResult<()> {
+        let len = usize::try_from(len).map_err(|_| HostError::IntegerOverflow)?;
+
```
chefsale: remove the spaces. looks weird.

crates/sdk/src/env.rs
```diff
@@ -109,9 +106,11 @@ pub fn read_register(register_id: RegisterId) -> Option<Vec<u8>> {
 #[inline]
 fn read_register_sized<const N: usize>(register_id: RegisterId) -> Option<[u8; N]> {
     let len = register_len(register_id)?;
-    let buffer = [0; N];
+
+    let mut buffer = [0; N];
+
```
chefsale, line 106: same here :)

crates/sdk/src/env.rs
```diff
@@ -185,46 +184,20 @@ pub fn state_write<T: AppState>(state: &T) {
...
-    )]
-    #[expect(clippy::expect_used, reason = "Effectively infallible here")]
-    {
-        SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("Time went backwards to before the Unix epoch!")
-            .as_nanos() as u64
-    }
+    let mut bytes = [0; 8];
+
```
chefsale, line 236: same here

danwilliams (approved):

crates/sdk/src/env.rs
miraclx: spaces help readability, which in turn makes it easier to maintain, especially with `unsafe` things, and doesn't affect the code

crates/runtime/src/logic.rs
miraclx: combined with the vec initialization to match

https://github.com/calimero-network/core/blob/402f5cfd1242b9bb4efe88f07b9492312bde3538/crates/runtime/src/logic.rs#L186-L192

chefsale: Also ugly imo :D

crates/sdk/src/env.rs
danwilliams: I don't personally like the space either. It's quite uncommon, and unidiomatic in Rust, to insert blank lines like this between assignments.

danwilliams: Rust is generally kept compact, as the language is fairly terse, and brevity is sought-after. Plus, as the rustfmt standards lead to a lot of vertical code, it is undesirable to add more height.

I tend to add blank lines to differentiate between semantically-different blocks of code, which does aid clarity when there are a lot of lines. In short functions it provides no benefit and actually detracts.

miraclx: > It's quite uncommon and unidiomatic in Rust to insert blank lines like this between assignments.

There's no convention - https://github.com/rust-lang/style-team/issues/57

- https://github.com/rust-lang/rust/blob/28e8f01c2a2f33fb4214925a704e3223b372cad5/library/alloc/src/raw_vec.rs#L620
- https://github.com/sharkdp/bat/blob/eca6b8a3768ec3931a41330c42b138107b1786b5/src/diff.rs#L19

As you mentioned

> I tend to add blank lines to differentiate between semantically-different blocks of code, which does aid clarity when there are a lot of lines. In short functions it provides no benefit and actually detracts.

Below this, there's a macro annotation, merging them would impair clarity.

danwilliams: I am not sure how it affects clarity, but even if wanting space between this assignment and the macro, I don't see why there's also a space between this assignment and the one above. It seems unnecessary.

I think there are more important things to worry about, but I tend to see these things as things I clean up from time to time in codebase hygiene sweeps. So I'm with @chefsale on this one.

crates/runtime/src/logic.rs
miraclx: Shrunk

crates/sdk/src/env.rs
miraclx: Shrunk

crates/sdk/src/env.rs
miraclx: Shrunk

chefsale (approved):

PR #811, opened 2024-10-16
fix: incorporate review feedback from #781 (approach 1)

miraclx:
Defined functional mocks in the storage crate as opposed to the SDK - ref [1](https://github.com/calimero-network/core/pull/781#discussion_r1800174373) & [2](https://github.com/calimero-network/core/pull/781#discussion_r1800175002).

This is one approach to resolving it, continuing the approach that was defined in #792, which turned out to be far more complicated than the alternative which conditionally includes `calimero_sdk`, using it when we're targeting wasm, and mocking it out when we aren't.

This is incomplete btw, figured it wasn't worth continuing seeing as #812 was the better approach, to finish it, the macro codegen would need to be updated to know which environment to be run in (due to `Id::new` calls).


chefsale:
Closing this as we agreed on approach 2.

PR #812, opened 2024-10-16
fix: incorporate review feedback from #781 (approach 2)

miraclx:
Defined functional mocks in the storage crate as opposed to the SDK - ref [1](https://github.com/calimero-network/core/pull/781#discussion_r1800174373) & [2](https://github.com/calimero-network/core/pull/781#discussion_r1800175002).

This is one approach to resolving it which conditionally includes `calimero_sdk`, using it when we're targeting wasm, and mocking it out when we aren't.

An alternative, presented in #811 is to continue the approach that was defined in #792, but it introduced far more complexity than was needed.


crates/storage/src/address.rs
```diff
@@ -49,7 +50,13 @@ impl Id {
     ///
     #[must_use]
     pub fn new() -> Self {
-        Self(generate_uuid())
+        let mut bytes = [0; 16];
+
```
chefsale: again 😂

miraclx: Shrunk

chefsale (approved):
stamp

PR #813, opened 2024-10-17
chore: seperated the meroctl commands and interactive cli commands in…

chefsale:
NOT compiling yet.


crates/node/src/interactive_cli.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        },
+        Err(err) => {
+            println!("Failed to parse command: {}", err);
+            return Ok(());
+        }
+    };
+
+    Ok(())
+}
```
miraclx: 1. Looks like no `Err`s are returned, or bubbled up via `?`
2. We should match and exit early for try-parse  `Err`
3. We can compress these

```suggestion
pub async fn handle_line(node: &mut Node, line: String) {
    // IMPORTANT: Parser needs first string to be binary name
    let mut args = vec![""];
    args.extend(line.split_whitespace());
    println!("args: {:?}", args);

    let command = match RootCommand::try_parse_from(args) {
        Ok(command) => command,
        Err(err) => {
            println!("Failed to parse command: {}", err);
            return;
        }
    };
    
    let result = match command.action {
        SubCommands::Application(application) => application.run(node).await,
        SubCommands::Call(call) => call.run(node).await,
        SubCommands::Context(context) => context.run(node).await,
        SubCommands::Gc(gc) => gc.run(node).await,
        SubCommands::Identity(identity) => identity.run(node).await,
        SubCommands::Peers(peers) => peers.run(node.network_client.clone().into()).await,
        SubCommands::Pool(pool) => pool.run(node).await,
        SubCommands::State(state) => state.run(node).await,
        SubCommands::Store(store) => store.run(node).await,
        SubCommands::Transactions(transactions) => transactions.run(node).await,
    };
        
    if let Err(err) = result {
        println!("Error running command: {}", e);
    }
}
```

crates/node/src/interactive_cli/applications.rs
```diff
@@ -0,0 +1,91 @@
...
+                                    .map(|x| x.as_bytes().to_owned())
+                                    .unwrap_or_default(),
+                            )
+                            .await
+                            .ok()
+                        {
+                            application_id
+                        } else {
+                            println!("{ind} Failed to install application from path");
+                            return Err(eyre::eyre!("Failed to install application from path"));
```
miraclx:
```suggestion
                            eyre::bail!("Failed to install application from path");
```

miraclx (approved):

crates/node/src/interactive_cli.rs
```diff
@@ -0,0 +1,108 @@
...
+    Store(store::StoreCommand),
+    State(state::StateCommand),
+    Transactions(transactions::TransactionsCommand),
+}
+
+pub async fn handle_line(node: &mut Node, line: String) -> Result<(), Report> {
+    // IMPORTANT: Parser needs first string to be binary name
+    let mut args = vec![""];
+    args.extend(line.split_whitespace());
+    println!("args: {:?}", args);
```
miraclx: don't forget this

crates/node/src/interactive_cli/call.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub async fn run(self, node: &mut Node) -> Result<()> {
+        let ind = ">>".blue();
+        let executor_key = match PublicKey::from_str(&self.executor_key) {
+            Ok(key) => key,
+            Err(_) => {
+                println!("{} invalid executor key: {}", ind, self.executor_key);
+                return Ok(());
+            }
+        };
+        let payload = serde_json::from_str::<Value>(&self.payload)?;
```
miraclx: you can put this in the command args

crates/node/src/interactive_cli/call.rs
```diff
@@ -0,0 +1,130 @@
...
+use tokio::sync::oneshot;
+
+use crate::Node;
+use eyre::Result;
+
+#[derive(Debug, Parser)]
+pub struct CallCommand {
+    context_id: String,
+    method: String,
+    payload: String,
```
miraclx:
```suggestion
    payload: Value,
```

crates/node/src/interactive_cli/call.rs
```diff
@@ -0,0 +1,130 @@
...
+    executor_key: String,
+}
+
+impl CallCommand {
+    pub async fn run(self, node: &mut Node) -> Result<()> {
+        let ind = ">>".blue();
+        let executor_key = match PublicKey::from_str(&self.executor_key) {
```
miraclx: same here

miraclx (commented):
we can merge as-is, but ideally, we should validate inputs with clap instead of secondary parsing, but that can be a separate PR for sure

PR #814, opened 2024-10-17
fix(node-ui): local storage clearing in login process

frdomovic:
# fix(node-ui): local storage clearing in login process

## Summary
If local storage gets manually removed in the process of login added now checks to redirect to setup page again.
This is the only way that this can be recreated (by clearning local storage) in other examples it will give you error under login card of each network.


node-ui/src/hooks/useIcp.tsx
```diff
@@ -189,6 +190,10 @@ export function useIcp(): useIcpReturn {
 
   const signMessageAndLogin = useCallback(
     async (isLogin: boolean, setErrorMessage: (msg: string) => void) => {
+      if (!getAppEndpointKey()) {
+        navigate('/?node_url_error=true');
```
MatejVukosav: why don't you just navigate to setup page?

MatejVukosav (changes requested):

node-ui/src/hooks/useIcp.tsx
frdomovic: Because its a bad UX redirecting user and not knowing why you got redirected. You press sign message and then youre back at setup again, would look like its a bug

MatejVukosav: But how can local storage lose the necessary data? There must be action before it.

frdomovic: > This is the only way that this can be recreated (by clearning local storage) in other examples it will give you error under login card of each network.

frdomovic: Removed popup and redirect back to setup page.

alenmestrov (approved):

PR #815, opened 2024-10-18
adds docker and docker-compose to run coordinater and app nodes

elliotBraem:
I was having issues running the node on my machine via the `cargo run` in docs, so this PR introduces a Dockerfile and docker-compose to create an image and run the two nodes.

Please, feel free to make any changes to it -- it builds a 5.78 GB image that is reused across the two instances. The `init` could probably be better and not in the compose.

Also -- it doesn't come without an issue... I'm able to successfully build the image and run docker-image (proof below).

Screenshot 2024-10-17 at 9 29 24 PM

Inside the docker container, I can access the dashboard:

`docker exec -it core-app_node_run-1 /bin/sh`

then inside the bash, I download the certificate:

```
curl -k -o /certs/node_certificate.crt https://localhost:2428/admin-api/certificate
```

/certs is shared with my local machine, so I see it in finder, double click, add to MacOS keychain (according to docs, proof)

Screenshot 2024-10-17 at 9 32 51 PM

And I always trust, refresh browser. 

**I still cannot connect to node's admin-dashboard from my device**

Although I CAN successfully ping it.

## Checklist

- [x] I have performed a self-review of my own code
- [ ] I have added a test plan that prove my fix is effective or that my feature
      works

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,46 @@
+# Use an official Rust image as the base
+FROM rust:latest
+
+# Install system dependencies
+RUN  apt-get update && apt-get install -y \
+    zlib1g-dev \
+    libsnappy-dev \
+    libbz2-dev \
+    liblz4-dev \
+    libzstd-dev \
+    clang \
+    libclang-dev \
+    curl \
+    build-essential \
+    pkg-config \
+    jq
+
+# Install Node.js (version 20) and pnpm
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
+    apt-get install -y nodejs && \
+    npm install -g pnpm
+
+# Set the working directory
+WORKDIR /app
+
+# Copy only necessary files for building dependencies
+COPY Cargo.toml Cargo.lock ./
+COPY crates ./crates
+COPY contracts ./contracts
+COPY apps ./apps
+COPY node-ui ./node-ui
+COPY packages ./packages
+
+# Build the node UI
+WORKDIR /app/node-ui
+RUN pnpm install && pnpm run build
+
+# Build the merod binary
+WORKDIR /app
+RUN cargo build --release -p merod
+
+# Set the binary as the entrypoint
+ENTRYPOINT ["/app/target/release/merod"]
+
+# Default command (can be overridden in docker-compose)
+CMD ["--help"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,41 @@
+version: '3.8'
+
+services:
+  coordinator_init:
+    image: merod-image:latest
+    command: >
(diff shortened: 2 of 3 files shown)


chefsale:
@fbozic can you take the optimization of the docker image as a task to fix this?

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: '3.8'
+
+services:
+  coordinator_init:
+    image: merod-image:latest
+    command: >
+      --node-name coordinator --home /data init --server-port 2427 --swarm-port 2527
```
chefsale: probably should be customizable (coordinator vs node)

chefsale:
Also, thanks for the contribution. 💯

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+    ports:
+      - "2428:2428"
+      - "2528:2528"
+    command: >
+      --node-name node1 --home /data run
+    volumes:
+      - ./data:/data
+      - ./certs:/certs
+    depends_on:
+      - app_node_init
```
fbozic: ~nit: Missing new line

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# Build the merod binary
+WORKDIR /app
+RUN cargo build --release -p merod
+
+# Set the binary as the entrypoint
+ENTRYPOINT ["/app/target/release/merod"]
+
+# Default command (can be overridden in docker-compose)
+CMD ["--help"]
```
fbozic: ~nit: Missing new line

docker-compose.yml
fbozic: When running inside the container server host should be 0.0.0.0. To configure that pass `--server-host 0.0.0.0`.

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+
+  app_node_init:
+    image: merod-image:latest
+    command: >
+      --node-name node1 --home /data init --server-port 2428 --swarm-port 2528
+    volumes:
+      - ./data:/data
+
+  app_node_run:
+    image: merod-image:latest
```
fbozic: Can we drop the `-image` suffix?

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+      --node-name coordinator --home /data run --node-type coordinator
+    volumes:
+      - ./data:/data
+    depends_on:
+      - coordinator_init
+
+  app_node_init:
+    image: merod-image:latest
+    command: >
+      --node-name node1 --home /data init --server-port 2428 --swarm-port 2528
```
fbozic: And the same here, set the server host to 0.0.0.0

PR #819, opened 2024-10-18
chore: improved interactive cli context iteration

chefsale:
Node builds and runs.

### DIFF
--- a/crates/primitives/src/context.rs
+++ b/crates/primitives/src/context.rs
@@ -20,6 +20,12 @@ impl From<[u8; 32]> for ContextId {
     }
 }
 
+impl From<String> for ContextId {
+    fn from(s: String) -> Self {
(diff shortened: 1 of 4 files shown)


crates/primitives/src/context.rs
```diff
@@ -20,6 +20,12 @@ impl From<[u8; 32]> for ContextId {
     }
 }
 
+impl From<String> for ContextId {
+    fn from(s: String) -> Self {
+        s.parse().expect("Invalid ContextId string")
+    }
+}
```
miraclx: panicking is discouraged, and calling `.parse()` at source is the better approach anyway

chefsale: will fix it, still fighting with the compile error above.

miraclx (approved):

MatejVukosav:
merge?

Issue #842, opened 2024-10-22
Remove use of deprecated query and mutate calls


chefsale:
- [ ] Examples should be updated to use execute method instead
- [ ] Documentation  should be updated to use execute method instead
- [ ] In one of the following breaking releases 1.0.0 these methods should deleted from the SDK

miraclx:
@frdomovic, can you take a look. All is resolved here except the Client SDK, can we update that?

frdomovic:
Resolved [here](https://github.com/calimero-network/calimero-client-js/pull/34)

PR #843, opened 2024-10-22
chore: merged mutate and query into one query method

chefsale:
Added a task to fully deleted the mutate and execute calls in the future https://github.com/calimero-network/core/issues/842, tagged them @depracated for now to keep current things working. Also, we should have a breaking SDK and core release to push changes like this. Ideally before 1.0.0.


crates/server-primitives/src/jsonrpc.rs
```diff
@@ -233,9 +175,8 @@ impl MutateResponse {
...
 #[serde(tag = "type", content = "data")]
 #[non_exhaustive]
-pub enum MutateError {
+pub enum ExecuteError {
     SerdeError { message: String },
     CallError(CallError),
     FunctionCallError(String),
-}
-// *************************************************************************
+}
```
frdomovic: eof

frdomovic (approved):
Looks good, was new sdk version tested with one of the examples?

crates/server-primitives/src/jsonrpc.rs
chefsale: wdym? :)

frdomovic:
```suggestion
}

```
end of file. To match other files in project where all end with space

chefsale: Oh 😂

PR #854, opened 2024-10-23
feat: setup proxy lib foundation

xilosada:
Setup proxy lib structure for testing
In this initial commit I just added the foundation of the NEAR lib. It implements a basic smartcontract to request the member list from the config contract. Tests and code will change a lot in the following PRs. 
NOTE:
This PR also downgrade near-workspaces version to support darwin x86-64


xilosada:
@miraclx I'll replace the errors with eyre in the next commit

crates/context/config/src/repr.rs
```diff
@@ -50,6 +51,13 @@ impl<T: ReprBytes> Display for Repr<T> {
     }
 }
 
+impl<T: ReprBytes> Hash for Repr<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let encoded = bs58::encode(self.inner.as_bytes()).into_string();
+        encoded.hash(state);
+    }
+}
```
miraclx: I think deriving `Hash` for `Repr<T>` is best so it defers to `T: Hash` ..._

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[near(contract_state)]
+#[derive(PanicOnDefault)]
+pub struct ProxyContract {
+    pub context_id: Repr<ContextId>,
+    pub context_config_account_id: AccountId,
+    pub num_approvals: u32,
+    pub proposal_nonce: ProposalId,
+    pub proposals: IterableMap<ProposalId, ProposalWithSigner>,
+    pub approvals: IterableMap<ProposalId, HashSet<Repr<SignerId>>>,
```
miraclx: _... although it looks like `Repr<T>: Hash` is only needed because of this.

`Repr` is only necessary for wire serialization (JSON). For state serialization `SignerId` already implements `Borsh{Se,Dese}rialize`.

```suggestion
    pub approvals: IterableMap<ProposalId, HashSet<SignerId>>,
```

contracts/context-config/src/query.rs
```diff
@@ -44,6 +44,15 @@ impl ContextConfigs {
         members
     }
 
+    pub fn has_member(&self, context_id: Repr<ContextId>, identity: Repr<ContextIdentity>) -> bool {
+        let context = self
+            .contexts
+            .get(&context_id)
+            .expect("context does not exist");
```
danwilliams, line 51: We may want to think about returning a `Result` here.

contracts/proxy-lib/src/ext_config.rs
```diff
@@ -0,0 +1,11 @@
+use near_sdk::ext_contract;
+use calimero_context_config::repr::Repr;
+use calimero_context_config::types::{SignerId, ContextId};
+
+#[ext_contract(config_contract)]
+trait Configcontract {
```
danwilliams: Should this be `ConfigContract`?

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(Clone, Debug)]
+#[near(serializers = [borsh])]
+pub struct FunctionCallPermission {
+    allowance: Option<U128>,
+    receiver_id: AccountId,
+    method_names: Vec<String>,
+}
+
+// An internal request wrapped with the signer_pk and added timestamp to determine num_requests_pk and prevent against malicious key holder gas attacks
```
danwilliams: Ideally we would turn this into Rustdoc documentation, i.e. using `///` and then following the standard formatting approach for function documentation 🙂

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct Proposal {
+    pub receiver_id: AccountId,
+    pub actions: Vec<ProposalAction>,
+}
+
+#[near]
+impl ProxyContract {
+    #[init]
+    pub fn init(context_id: Repr<ContextId>, context_config_account_id: AccountId) -> Self {
+        assert!(!env::state_exists(), "Already initialized");
```
miraclx: the `#[init]` annotation expands to this:

```rust
#[cfg(target_arch = "wasm32")]
#[no_mangle]
pub extern "C" fn init() {
    ::near_sdk::env::setup_panic_hook();
    if ::near_sdk::env::attached_deposit().as_yoctonear() != 0 {
        ::near_sdk::env::panic_str("Method init doesn't accept deposit");
    }
    if ::near_sdk::env::state_exists() {
        ::near_sdk::env::panic_str("The contract has already been initialized");
    }
    let contract = ContextConfigs::init();
    ::near_sdk::env::state_write(&contract);
}
```

so it does this already

contracts/proxy-lib/tests/config_helper.rs
```diff
@@ -0,0 +1,104 @@
...
+                let kind = RequestKind::Context(ContextRequest::new(
+                    context_id,
+                    ContextRequestKind::Add {
+                        author_id,
+                        application: Application::new(
+                            application_id,
+                            blob_id,
+                            0,
+                            Default::default(),
+                            Default::default(),
```
danwilliams: More idiomatic to use the specific type names here, instead of `Default`.

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    pub fn create_proposal(&mut self, proposal: Signed<ProposalWithSigner>) -> ProposalId {
+        // Verify the signature corresponds to the signer_id
+        let proposal = proposal
+            .parse(|i| *i.signer_id)
+            .expect("failed to parse input");
+
+        let singer_id = &proposal.signer_id;
```
miraclx:
```suggestion
        let signer_id = proposal.signer_id;
```

also, `SignerId: Copy`, so you can copy it here (i.e remove `&`) and reference everywhere else

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        let num_proposals = self.num_proposals_pk.get(singer_id).unwrap_or(&0) + 1;
+        assert!(
+            num_proposals <= self.active_proposals_limit,
+            "Account has too many active proposals. Confirm or delete some."
+        );
+        self.num_proposals_pk
+            .insert(singer_id.clone(), num_proposals);
```
miraclx: `SignerId: Copy`, so:

```suggestion
        self.num_proposals_pk
            .insert(*singer_id, num_proposals);
```

or (if you removed `&` in the earlier comment)

```suggestion
        self.num_proposals_pk
            .insert(singer_id, num_proposals);
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.approvals.insert(self.proposal_nonce, HashSet::new());
+        self.proposal_nonce += 1;
+        self.proposal_nonce - 1
+    }
+
+    pub fn create_and_approve_proposal(&mut self, proposal: Signed<ProposalWithSigner>) -> Promise {
+        let proposal_id = self.create_proposal(proposal.clone());
+        let proposal = proposal
+            .parse(|i| *i.signer_id)
+            .expect("failed to parse input");
```
miraclx: `create_proposal` will call `.parse` and so are you here.

you can introduce a non-`pub`* method that creates a proposal from a pre-parsed proposal.

*: I didn't say "private" since `near-sdk` has a concept of `#[private]` and that's not what I mean

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+            proposal_nonce: 0,
+            proposals: IterableMap::new(b"r".to_vec()),
+            approvals: IterableMap::new(b"c".to_vec()),
+            num_proposals_pk: IterableMap::new(b"k".to_vec()),
+            num_approvals: 2,
+            active_proposals_limit: 10,
+        }
+    }
+
+    pub fn create_proposal(&mut self, proposal: Signed<ProposalWithSigner>) -> ProposalId {
```
miraclx: do we want anybody to be able submit a proposal even if they're not a context member?

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    pub fn approve(&mut self, request: Signed<ConfirmationRequestWithSigner>) -> Promise {
+        let request = request
+            .parse(|i| *i.signer_id)
+            .expect("failed to parse input");
+        return self.approve_by_member(request.signer_id, request.proposal_id);
+    }
+
+    fn internal_confirm(&mut self, request_id: ProposalId, signer_id: Repr<SignerId>) -> () {
```
miraclx:
```suggestion
    fn internal_confirm(&mut self, request_id: ProposalId, signer_id: Repr<SignerId>) {
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+                Self::ext(env::current_account_id()).internal_process_members(identity, request_id),
+            )
+    }
+
+    pub fn requests(&self, offset: usize, length: usize) -> Vec<(&u32, &ProposalWithSigner)> {
+        let mut requests = Vec::with_capacity(length);
+        for request in self.proposals.iter().skip(offset).take(length) {
+            requests.push(request);
+        }
+        requests
```
miraclx: nit, shouldn't matter much:

```suggestion
        let mut requests = Vec::with_capacity((self.proposals.len() - offset).min(length));
        for request in self.proposals.iter().skip(offset).take(length) {
            requests.push(request);
        }
        requests
```

or leave it to `Iterator` which appropriately derive the exact size from `Iterator::size_hint`

```suggestion
        self.proposals.iter().skip(offset).take(length).collect()
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    pub fn get_confirmations_count(&self, proposal_id: ProposalId) -> ProposalWithApprovals {
+        let size = self
+            .approvals
+            .get(&proposal_id)
+            .unwrap_or(&HashSet::new())
+            .len();
```
miraclx: shouldn't we return an option here if the proposal doesn't exist?

```suggestion
    pub fn get_confirmations_count(&self, proposal_id: ProposalId) -> Option<ProposalWithApprovals> {
        let size = self
            .approvals
            .get(&proposal_id)?
            .len();
```

contracts/proxy-lib/src/ext_config.rs
```diff
@@ -0,0 +1,11 @@
...
+    fn has_member(&self,
```
miraclx: has this been implemented in the ctx config contract?

contracts/context-config/src/query.rs
xilosada: I followed the structure of the other functions in the module. I had to add this helper to faciliate the cx contract call. cc: @miraclx

contracts/proxy-lib/src/ext_config.rs
xilosada: Fixed

contracts/proxy-lib/src/lib.rs
xilosada: Will do it in a following commit, leaving this open till that

contracts/proxy-lib/tests/config_helper.rs
xilosada: I'm afraid this is coming from the config-contract, I'll create an issue to address it in a different commit. Leaving this open until the issue is created

contracts/context-config/src/query.rs
miraclx: yeah, so NEAR contracts will only roll back a failed transaction if execution fails, and they fail only when the contract panics, unlike ours.

Returning a result will serialize the result as a success value, even if it's an `Err`, 🫠 

So the spent gas won't be refunded, state mutations will hold, etc..

contracts/proxy-lib/src/lib.rs
xilosada: +1 it must panic and keep the map empty. Will work on it

contracts/proxy-lib/src/lib.rs
xilosada: Fair

contracts/proxy-lib/src/ext_config.rs
xilosada: here it goes https://github.com/calimero-network/core/pull/854/files#diff-22355c320575851613e81b59302954d94433115e28c09199b70079772934a535R47

miraclx: Ah, missed that, noice.

contracts/proxy-lib/src/lib.rs
```diff
@@ -94,35 +89,18 @@ impl ProxyContract {
...
-    pub fn create_proposal(&mut self, proposal: Signed<ProposalWithSigner>) -> ProposalId {
+    pub fn create_and_approve_proposal(&self, proposal: Signed<Proposal>) -> Promise {
         // Verify the signature corresponds to the signer_id
-        let proposal = proposal
-            .parse(|i| *i.signer_id)
-            .expect("failed to parse input");
+        let proposal = proposal.parse(|i| *i.author_id).expect("failed to parse input");
 
-        let singer_id = &proposal.signer_id;
+        let author_id = proposal.author_id.rt().expect("Invalid signer");
```
miraclx: it's `SignerId` on both sides, so no need to transmute

```suggestion
        let author_id = *proposal.author_id;
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -191,4 +179,34 @@ impl ProxyContract {
...
+
+    #[private]
+    pub fn internal_create_proposal(
+        &mut self,
+        proposal: Proposal,
+        num_proposals: u32,
+        #[callback_result] call_result: Result<bool, PromiseError>, // Match the return type
+    ) -> ProposalWithApprovals {
+        assert!(call_result.is_ok(), "Error: Membership check failed");
+        assert!(call_result.unwrap(), "Error: Is not a member");
```
miraclx:
```suggestion
        let has_member = call_result.expect("Membership check failed");
        assert!(has_member, "Is not a member");
```

we typically don't need the `Error:` prefix, as that's already evident to whoever's handling it

contracts/proxy-lib/src/lib.rs
```diff
@@ -155,7 +133,17 @@ impl ProxyContract {
             )
     }
 
-    pub fn requests(&self, offset: usize, length: usize) -> Vec<(&u32, &ProposalWithSigner)> {
+    fn create_by_member(&self, proposal: Proposal, num_proposals: u32) -> Promise {
+        log!("Starting fetch_members...");
+        config_contract::ext(self.context_config_account_id.clone())
+            .with_static_gas(Gas::from_tgas(5))
```
miraclx: Maybe we need to make this more dynamic?

```suggestion
            .with_static_gas(env::prepaid_gas())
```

or perhaps

```suggestion
            .with_static_gas(env::prepaid_gas().saturating_sub(env::used_gas()))
```

though I'd imagine the manual subtraction shouldn't be necessary since that delta would be different at the end of the execution anyway, so we may never be able to determine the exact amount of unused gas

does removing the `with_static_gas` definition fail?

contracts/proxy-lib/src/lib.rs
```diff
@@ -191,4 +179,34 @@ impl ProxyContract {
...
+        self.num_proposals_pk
+            .insert(*proposal.author_id, num_proposals);
+        
+        let proposal_id = self.proposal_nonce;
+        self.proposal_nonce += 1;
+
+        self.proposals.insert(proposal_id, proposal.clone());
+        self.approvals.insert(proposal_id, HashSet::new());
+
+        self.internal_confirm(proposal_id, proposal.author_id);
```
miraclx: it's cheaper to copy `SignerId` than to clone `Proposal`

```suggestion
        let author_id = *proposal.author_id;

        self.proposals.insert(proposal_id, proposal);
        self.approvals.insert(proposal_id, HashSet::new());

        self.internal_confirm(proposal_id, author_id);
```

contracts/proxy-lib/tests/common/proxy_lib_helper.rs
```diff
@@ -40,35 +40,35 @@ impl ProxyContractHelper {
...
 
+    pub fn create_proposal(
+        &self,
+        author: &SigningKey,
+        receiver: &Account,
+        actions: Vec<ProposalAction>,
+    ) -> Result<Signed<Proposal>> {
+        let proposal = Proposal {
+            receiver_id: receiver.id().clone(),
+            author_id: author.verifying_key().to_bytes().rt().expect("Invalid signer"),
```
miraclx: [`VerifyingKey: ReprBytes`](https://github.com/calimero-network/core/blob/fd4fdc9c6a8f52baf501ae5f013dbe698064b6e1/crates/context/config/src/types.rs#L255), so:

```suggestion
            author_id: author.verifying_key().rt().expect("Invalid signer"),
```

contracts/proxy-lib/tests/common/proxy_lib_helper.rs
```diff
@@ -40,35 +40,35 @@ impl ProxyContractHelper {
...
+            actions
+        };
+        let signed = Signed::new(&{proposal}, |p| author.sign(p))?;
```
miraclx:
```suggestion
        let signed = Signed::new(&proposal, |p| author.sign(p))?;
```

contracts/proxy-lib/tests/common/proxy_lib_helper.rs
```diff
@@ -40,35 +40,35 @@ impl ProxyContractHelper {
...
             .call(self.proxy_contract.id(), "create_and_approve_proposal")
             .args_json(json!({
-                "proposal": args
+                "proposal": proposal
             }))
             .max_gas()
             .transact()
-            .await
+            .await;
+        Ok(call?)
```
miraclx: interesting model 😅 

typically we do something like:

```suggestion
            .await;
        call.map_err(Into::into)
```

but here you can just bubble it up at the point you got the result, so it's clearer

```suggestion
            .await?;
        Ok(call)
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    fn internal_confirm(&mut self, request_id: ProposalId, signer_id: Repr<SignerId>) -> () {
+        let confirmations = self.approvals.get_mut(&request_id).unwrap();
+        assert!(
+            !confirmations.contains(&signer_id),
+            "Already confirmed this request with this key"
+        );
+        confirmations.insert(signer_id);
+
+        // Check if the number of confirmations is enough to execute the request
```
miraclx: As discussed, you'll be introducing a mechanism for re-validating the approvals, yeah?

So `has_member` will need to become `has_members` so we can pass the list of approvers to it, yeah?

Or maybe pass a list of IDs, and return a list of non-members. So we can identify which approvals to invalidate.

Or instead of returning a list of non-members, it just returned a count of non-members. So knowing when to execute the request would be:

`if threshold == (number_of_members_checked - number_of_non_members_received) {}`

which also means if someone leaves the context and rejoins before the execution under the same identity, their approval stays valid

contracts/proxy-lib/src/ext_config.rs
```diff
@@ -0,0 +1,11 @@
+use near_sdk::ext_contract;
+use calimero_context_config::repr::Repr;
+use calimero_context_config::types::{SignerId, ContextId};
+
+#[ext_contract(config_contract)]
+pub trait ConfigContract {
```
miraclx, line 6: also, btw, you can use this directly from the crate if you added it as a dependency

```rust
use calimero_context_config_near::ContextConfigs;

let promise = ContextConfigs::ext(self.context_config_account_id)
    .with_static_gas(..)
    .members(context_id, offset, length)
    .then(..);
```

contracts/proxy-lib/src/lib.rs
xilosada: What about this?

crates/context/config/src/repr.rs
xilosada: derived Hash for Identity and SignerId

contracts/proxy-lib/src/lib.rs
xilosada: Done

contracts/proxy-lib/src/lib.rs
xilosada: Removed

contracts/proxy-lib/tests/common/proxy_lib_helper.rs
xilosada: :D

contracts/proxy-lib/src/ext_config.rs
xilosada: Will keep this until I change the config contract with the new apis

contracts/proxy-lib/src/lib.rs
xilosada: discussed internally. This will be part of a new PR

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+pub struct ProxyContract {
+    pub context_id: ContextId,
+    pub context_config_account_id: AccountId,
+    pub num_approvals: u32,
+    pub proposal_nonce: ProposalId,
+    pub proposals: IterableMap<ProposalId, Proposal>,
+    pub approvals: IterableMap<ProposalId, HashSet<SignerId>>,
+    pub num_proposals_pk: IterableMap<SignerId, u32>,
+    pub active_proposals_limit: u32,
+    pub context_storage: LookupMap<Box<[u8]>, Box<[u8]>>,
```
xilosada: to IterableMap

contracts/proxy-lib/tests/common/proxy_lib_helper.rs
```diff
@@ -42,11 +42,11 @@ impl ProxyContractHelper {
     pub fn create_proposal(
         &self,
         author: &SigningKey,
-        actions: Vec<ProposalAction>,
+        actions: &Vec<ProposalAction>,
```
miraclx, line 46:
```suggestion
        actions: &[ProposalAction],
```

contracts/proxy-lib/tests/common/counter_helper.rs
```diff
@@ -0,0 +1,34 @@
...
+use super::deploy_contract;
+
+const COUNTER_WASM: &str = "../test-counter/res/test_counter_near.wasm";
+
+#[derive(Clone)]
+pub struct CounterContracttHelper {
+    pub counter_contract: Contract,
+}
+
+impl CounterContracttHelper {
```
MatejVukosav:
```suggestion
impl CounterContractHelper {
```

contracts/proxy-lib/tests/common/counter_helper.rs
```diff
@@ -0,0 +1,34 @@
+use eyre::Result;
+use near_workspaces::{network::Sandbox, Contract, Worker};
+
+use super::deploy_contract;
+
+const COUNTER_WASM: &str = "../test-counter/res/test_counter_near.wasm";
+
+#[derive(Clone)]
+pub struct CounterContracttHelper {
```
MatejVukosav:
```suggestion
pub struct CounterContractHelper {
```

contracts/proxy-lib/tests/common/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+use near_workspaces::{network::Sandbox, types::NearToken, Account, Contract, Worker};
+use rand::Rng;
+
+pub mod config_helper;
+pub mod counter_helper;
+pub mod proxy_lib_helper;
+
+pub async fn deploy_contract(worker: &Worker<Sandbox>, wasm_path: &str) -> Result<Contract> {
+    let wasm = std::fs::read(wasm_path)?;
+    let contract = worker.dev_deploy(&wasm).await?;
```
MatejVukosav, line 14: Can this fail?

xilosada: yes and it will make the test fail

contracts/proxy-lib/tests/common/counter_helper.rs
xilosada: good catch

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+use calimero_context_config::types::{ContextId, Signed, SignerId};
+use near_sdk::json_types::{Base64VecU8, U128};
+use near_sdk::store::IterableMap;
+use near_sdk::{
+    env, near, AccountId, Gas, NearToken, PanicOnDefault, Promise, PromiseError, PromiseOrValue,
+    PromiseResult,
+};
+
+pub mod ext_config;
+pub use crate::ext_config::config_contract;
```
miraclx, line 15: `ext_config` is already in scope

```suggestion
pub use ext_config::config_contract;
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    allowance: Option<U128>,
+    receiver_id: AccountId,
+    method_names: Vec<String>,
+}
+
+#[derive(Clone, PartialEq)]
+#[near(serializers = [json, borsh])]
+pub struct ProposalApprovalWithSigner {
+    pub proposal_id: ProposalId,
+    pub signer_id: Repr<SignerId>,
```
miraclx, line 62: you shouldn't need `Repr` in state at all

```suggestion
    pub signer_id: SignerId,
```

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+            .expect("failed to parse input");
+
+        let author_id = proposal.author_id;
+
+        let num_proposals = self.num_proposals_pk.get(&author_id).unwrap_or(&0) + 1;
+        assert!(
+            num_proposals <= self.active_proposals_limit,
+            "Account has too many active proposals. Confirm or delete some."
+        );
+        return self.perform_action_by_member(MemberAction::Create {
```
miraclx, line 130: Hm, they can take advantage of this only being processed in the next block to submit multiple proposals at once, hoping they'll be included in the same block, at which point they'll subvert this limitation

we should increment this number before the config check

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+        proposal: Proposal,
+        num_proposals: u32,
+        #[callback_result] call_result: Result<bool, PromiseError>, // Match the return type
+    ) -> Option<ProposalWithApprovals> {
+        assert_membership(call_result);
+
+        self.num_proposals_pk
+            .insert(*proposal.author_id, num_proposals);
+
+        let proposal_id = self.proposal_nonce;
```
miraclx, line 236: to match SDK expectations, proposal ID should be random and determined by the runtime, but this can come in a separate PR, just noting it

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    ) -> Option<Box<[u8]>> {
+        let val = self.context_storage.insert(key.clone(), value);
+        val
+    }
+
+    pub fn context_storage_entries(
+        &self,
+        offset: usize,
+        length: usize,
+    ) -> Vec<(&Box<[u8]>, &Box<[u8]>)> {
```
miraclx, line 381:
```suggestion
    ) -> Vec<(&[u8], &[u8])> {
```

crates/context/config/src/types.rs
miraclx, line 1: we shouldn't need these ..._

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+#[near(contract_state)]
+#[derive(PanicOnDefault)]
+pub struct ProxyContract {
+    pub context_id: ContextId,
+    pub context_config_account_id: AccountId,
+    pub num_approvals: u32,
+    pub proposal_nonce: ProposalId,
+    pub proposals: IterableMap<ProposalId, Proposal>,
+    pub approvals: IterableMap<ProposalId, HashSet<SignerId>>,
```
miraclx, line 44: _... if this was a `BTreeSet`

```suggestion
    pub approvals: IterableMap<ProposalId, BTreeSet<SignerId>>,
```

are you concerned about rebalancing on insertion since `SignerId` is random? 32-bytes is pretty small comparison

contracts/proxy-lib/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+    #[init]
+    pub fn init(context_id: Repr<ContextId>, context_config_account_id: AccountId) -> Self {
+        Self {
+            context_id: context_id.rt().expect("Invalid context id"),
+            context_config_account_id,
+            proposal_nonce: 0,
+            proposals: IterableMap::new(b"r".to_vec()),
+            approvals: IterableMap::new(b"c".to_vec()),
+            num_proposals_pk: IterableMap::new(b"k".to_vec()),
+            num_approvals: 3,
```
miraclx, line 111: should we take this as an initialization parameter?

contracts/proxy-lib/src/lib.rs
xilosada: yes

contracts/proxy-lib/src/lib.rs
xilosada: @petarjuki7 let's add it once we have the proxy-contract integration in the node

miraclx (approved):

contracts/proxy-lib/src/lib.rs
xilosada: @danwilliams plese change it once you are generating the Proposal Id

contracts/proxy-lib/src/lib.rs
xilosada: will do later

PR #855, opened 2024-10-23
feat(cli): init node protocol option

### LINKED ISSUE #828: Starknet Integration - starknet json rpc query method
Following Near integration example, add Starknet json rpc integration for query method.

frdomovic:
# feat(cli): init node protocol option

## Summary
Added protocol flag (--protocol (optional - default value (near)) to update protocol values for different networks.
For starknet added mainnnet and sepolia with RPC values while contract and relayer accounts are now default values (will be updated with rest of implementation)

Fixes #828 

## Test plan
one node with protocol near another with protocol starknet
Screenshot 2024-10-23 at 16 04 02


MatejVukosav:
But user can participate in different contexts where each context can use different chain? How do you apply that to 
`context.config.signer.self.mainnet` ?

crates/merod/src/cli.rs
```diff
@@ -65,6 +65,7 @@ pub struct RootArgs {
     /// Name of node
     #[arg(short, long, value_name = "NAME")]
     pub node_name: Utf8PathBuf,
+
```
frdomovic:
```suggestion
```

crates/merod/src/cli/init.rs
```diff
@@ -165,6 +170,12 @@ impl InitCommand {
             .relayer_url
             .unwrap_or_else(defaults::default_relayer_url);
 
+        let protocol_value = match self.protocol.as_str() {
+            "near" => config::Protocol::Near,
+            "starknet" => config::Protocol::Starknet,
+            _ => config::Protocol::UnknownNetwork,
+        };
```
miraclx: if `Protocol` impls `FromStr`, you can move it to the args itself, so clap does the validation

crates/merod/src/cli/init.rs
```diff
@@ -70,6 +71,10 @@ pub struct InitCommand {
     #[clap(long, value_name = "URL")]
     pub relayer_url: Option<Url>,
 
+    /// Name of protocol
+    #[clap(long, value_name = "PROTOCOL", default_value = "near")]
+    pub protocol: String,
```
miraclx:
```suggestion
    pub protocol: Protocol,
```

frdomovic: Fixed

PR #863, opened 2024-10-25
feat: introduce a secret key for encryption type into context identity

chefsale:
WIP, waiting for @miraclx and catchup.


crates/context/src/lib.rs
```diff
@@ -294,6 +298,7 @@ impl ContextManager {
     pub async fn join_context(
         &self,
         identity_secret: PrivateKey,
+        sender_secret: PrivateKey,
```
miraclx: same (user shouldn't pass it in, it should be random)

crates/context/src/lib.rs
```diff
@@ -144,10 +144,11 @@ impl ContextManager {
         seed: Option<[u8; 32]>,
         application_id: ApplicationId,
         identity_secret: Option<PrivateKey>,
+        sender_secret: Option<PrivateKey>,
```
miraclx: the user doesn't need to know this, it should always be random

crates/context/src/lib.rs
```diff
@@ -135,7 +135,7 @@ impl ContextManager {
     }
 
     #[must_use]
-    pub fn new_identity(&self) -> PrivateKey {
+    pub fn new_pk(&self) -> PrivateKey {
```
miraclx: `pk` can be `private_key` or `public_key`

crates/context/src/lib.rs
miraclx: this was the third comment in the review, somehow it got broken up, crap internet 😪

chefsale:
@petarjuki7 to take over this PR

crates/context/src/lib.rs
```diff
@@ -748,6 +758,28 @@ impl ContextManager {
...
+        &self,
+        context_id: &ContextId,
+        own_public_key: &PublicKey,
+    ) -> EyreResult<SigningKey> {
+        let handle = self.store.handle();
+        let key = handle
+            .get(&ContextIdentityKey::new(*context_id, *own_public_key))?
+            .unwrap() // we are sure it exists because we own the public_key
+            .sender_key
+            .unwrap();
```
petarjuki7: Maybe some assumptions that are invalid?

miraclx: use `ok_or_eyre` to convert it to an error

crates/node/src/sync.rs
```diff
@@ -23,25 +26,38 @@ pub struct SyncConfig {
...
-
+async fn send(
+    stream: &mut Stream,
+    message: &StreamMessage<'_>,
+    shared_key: Option<SharedKey>,
+) -> EyreResult<()> {
+    let base_data = borsh::to_vec(message)?;
+    let data = shared_key
+        .and_then(|key| key.encrypt(base_data.clone(), [0; aead::NONCE_LEN]))
+        .unwrap_or(base_data);
```
miraclx: rewrite with `match` to avoid the `.clone`

crates/node/src/sync.rs
```diff
@@ -23,25 +26,38 @@ pub struct SyncConfig {
...
     let Some(message) = timeout(duration, stream.next()).await? else {
         return Ok(None);
     };
 
-    let message = borsh::from_slice(&message?.data)?;
+    let message_data = message?.data.to_vec();
+
+    let data = shared_key
+        .and_then(|key| key.decrypt(message_data.clone(), [0; aead::NONCE_LEN]))
+        .unwrap_or(message_data);
```
miraclx: same

crates/node/src/sync/blobs.rs
```diff
@@ -63,6 +66,15 @@ impl Node {
             }
         };
 
+        let our_sending_key = self
+            .ctx_manager
+            .get_own_signing_key(&context.id, &our_identity)?;
+
+        let shared_key = SharedKey::new(
+            &our_sending_key,
+            &VerifyingKey::from_bytes(&their_identity)?,
```
miraclx: I'd change the parameters to `::new` to not require `ed25519_dalek::*` types, but instead `calimero_primitives::identity`

should save bubbling up that dependency through `node` and `context`

crates/node/src/sync/state.rs
```diff
@@ -71,6 +75,15 @@ impl Node {
 
         let mut sqx_out = Sequencer::default();
 
+        let our_sending_key = self
+            .ctx_manager
+            .get_own_signing_key(&context.id, &our_identity)?;
+
+        let shared_key = SharedKey::new(
+            &our_sending_key,
+            &VerifyingKey::from_bytes(their_identity.deref())?,
```
```suggestion
            &VerifyingKey::from_bytes(&*their_identity)?,
```

should suffice

crates/node/src/sync/state.rs
```diff
@@ -31,11 +34,12 @@ impl Node {
...
                 },
             },
+            None,
         )
         .await?;
 
-        let Some(ack) = recv(stream, self.sync_config.timeout).await? else {
-            bail!("connection closed while awaiting state sync handshake");
+        let Some(ack) = recv(stream, self.sync_config.timeout, None).await? else {
+            bail!("no response to state sync request");
```
miraclx: bad merge?

crates/node/src/types.rs
```diff
@@ -18,6 +18,7 @@ pub enum BroadcastMessage<'a> {
         author_id: PublicKey,
         root_hash: Hash,
         artifact: Cow<'a, [u8]>,
+        sending_identity: PublicKey,
```
miraclx: we already have the `author_id`

crates/node/src/lib.rs
```diff
@@ -301,15 +305,31 @@ impl Node {
...
+                };
+
+                let our_sending_key = self
+                    .ctx_manager
+                    .get_own_signing_key(&context_id, &our_identity)?;
+
+                let shared_key = SharedKey::new(
+                    &our_sending_key,
+                    &VerifyingKey::from_bytes(&sending_identity)?,
+                );
```
miraclx: when receiving a state delta broadcast, we'll be using the sender key as opposed to deriving a shared key

crates/node/src/lib.rs
```diff
@@ -366,11 +386,32 @@ impl Node {
...
+            };
+
+            let our_sending_key = self
+                .ctx_manager
+                .get_own_signing_key(&context.id, &our_identity)?;
+
+            let shared_key = SharedKey::new(
+                &our_sending_key,
+                &VerifyingKey::from_bytes(&executor_public_key)?,
+            );
```
miraclx: we already have `author_id`, shared key should just be the sender key. You can add a method to derive a `SharedKey` without DHKE (since it's already pre-shared)

crates/node/src/interactive_cli/identity.rs
```diff
@@ -69,9 +69,9 @@ impl IdentityCommand {
             }
             IdentitySubcommands::New => {
                 // Handle the "new" subcommand
-                let identity = node.ctx_manager.new_identity();
-                println!("{ind} Private Key: {}", identity.cyan());
-                println!("{ind} Public Key: {}", identity.public_key().cyan());
+                let identity = node.ctx_manager.new_private_key();
+                println!("Private Key: {}", identity.cyan());
+                println!("Public Key: {}", identity.public_key().cyan());
```
miraclx: bad merge?

crates/node/src/sync/blobs.rs
petarjuki7: Left the `ed25519_dalek::SigningKey` for now, can remove later in a cleanup if everything else is okay.

crates/context/src/lib.rs
```diff
@@ -768,7 +768,7 @@ impl ContextManager {
             .get(&ContextIdentityKey::new(*context_id, *own_public_key))?
             .unwrap() // we are sure it exists because we own the public_key
             .sender_key
-            .unwrap();
+            .ok_or_eyre(eyre!("Non existant SenderKey"))?;
```
```suggestion
            .ok_or_eyre("Non existant SenderKey")?;
```

crates/node/src/lib.rs
miraclx: this is not resolved

crates/node/src/lib.rs
```diff
@@ -302,14 +305,26 @@ impl Node {
...
-                .await?;
+                let identities = self.ctx_manager.get_context_owned_identities(context_id)?;
+
+                let Some(our_identity) = identities.into_iter().choose(&mut thread_rng()) else {
+                    bail!("no identities found for context: {}", context_id);
+                };
+
+                let our_sending_key = self
+                    .ctx_manager
+                    .get_own_signing_key(&context_id, &our_identity)?;
```
miraclx: we're not decrypting with "our sender key"

we're decrypting with "their sender key", so you should be passing `author_id` to `get_own_signing_key`

also, that method isn't named right

crates/node/src/lib.rs
```diff
@@ -396,22 +391,18 @@ impl Node {
...
-            );
+            let shared_key = SharedKey::from_sk(&our_sending_key);
 
             let artifact_encrypted = shared_key
                 .encrypt(outcome.artifact.clone(), [0; aead::NONCE_LEN])
                 .unwrap();
 
             let message = to_vec(&BroadcastMessage::StateDelta {
-                // We are encrypting only the artifact so we can use the context_id and own_id for decrypting
+                // We are encrypting only the artifact so we can use the context_id decrypting
```
miraclx: but we're not, right? not using the context id to decrypt

crates/node/src/lib.rs
```diff
@@ -302,14 +305,26 @@ impl Node {
...
+                    .ctx_manager
+                    .get_own_signing_key(&context_id, &our_identity)?;
+
+                let shared_key = SharedKey::from_sk(&our_sending_key);
+
+                let artifact = from_slice::<Vec<u8>>(
+                    &shared_key
+                        .decrypt(artifact.into_owned(), [0; aead::NONCE_LEN])
+                        .ok_or_else(|| eyre!("Failed to decrypt message"))?,
+                )?;
```
miraclx: why `from_slice` here?

crates/node/src/lib.rs
```diff
@@ -366,11 +381,28 @@ impl Node {
             .await
             != 0
         {
+            let identities = self.ctx_manager.get_context_owned_identities(context.id)?;
+
+            let Some(our_identity) = identities.into_iter().choose(&mut thread_rng()) else {
+                bail!("no identities found for context: {}", context.id);
+            };
```
miraclx: we already have an identity here, `executor_public_key`

crates/node/src/lib.rs
```diff
@@ -366,11 +381,28 @@ impl Node {
...
+
+            let our_sending_key = self
+                .ctx_manager
+                .get_own_signing_key(&context.id, &our_identity)?;
+
+            let shared_key = SharedKey::from_sk(&our_sending_key);
+
+            let artifact_encrypted = shared_key
+                .encrypt(outcome.artifact.clone(), [0; aead::NONCE_LEN])
+                .unwrap();
```
miraclx: don't unwrap please

crates/node/src/sync.rs
```diff
@@ -23,25 +25,47 @@ pub struct SyncConfig {
...
     stream: &mut Stream,
     duration: Duration,
+    shared_key: Option<SharedKey>,
 ) -> EyreResult<Option<StreamMessage<'static>>> {
     let Some(message) = timeout(duration, stream.next()).await? else {
         return Ok(None);
     };
 
-    let message = borsh::from_slice(&message?.data)?;
+    let message_data = message?.data.to_vec();
```
miraclx: `to_vec` will copy to a new allocation. whereas messages received are already owned.

```suggestion
    let message_data = message?.data.into_owned();
```

crates/node/src/sync.rs
```diff
@@ -166,9 +190,21 @@ impl Node {
             }
         }
 
+        let identities = self.ctx_manager.get_context_owned_identities(context.id)?;
+
+        let Some(our_identity) = identities.into_iter().choose(&mut thread_rng()) else {
+            bail!("no identities found for context: {}", context.id);
+        };
```
miraclx: this already happens inside `handle_blob_share_request` and `handle_state_sync_request`

crates/context/src/lib.rs
```diff
@@ -748,6 +758,28 @@ impl ContextManager {
         Ok(ids)
     }
 
+    pub fn get_own_signing_key(
```
miraclx: rename this method, we need to be able to fetch "sender" (not signing) keys for any member in the context

crates/context/src/lib.rs
```diff
@@ -748,6 +758,28 @@ impl ContextManager {
...
+        &self,
+        context_id: &ContextId,
+        own_public_key: &PublicKey,
+    ) -> EyreResult<SigningKey> {
```
```suggestion
    ) -> EyreResult<Option<SigningKey>> {
```

we won't always have it, and when we don't, we want to initiate sync

crates/context/src/lib.rs
```diff
@@ -748,6 +758,28 @@ impl ContextManager {
...
+        let handle = self.store.handle();
+        let key = handle
+            .get(&ContextIdentityKey::new(*context_id, *own_public_key))?
+            .unwrap() // we are sure it exists because we own the public_key
```
miraclx: it won't always exist, because we won't only want to fetch the sender key for our identity

crates/context/src/lib.rs
miraclx: but ideally, don't return an error on this.

crates/node/Cargo.toml
```diff
@@ -9,24 +9,27 @@ license.workspace = true
 [dependencies]
 borsh.workspace = true
 camino = { workspace = true, features = ["serde1"] }
-clap = { workspace = true, features = ["derive"] }
+clap.workspace = true
```
miraclx: why was this removed? I'm pretty sure we use this for interactive CLI

petarjuki7: clippy told me it's unused, but true, I see it's used in the interactive CLI, I'll bring it back

crates/node/src/lib.rs
petarjuki7: Yes, the comment should be better worded as we use the `context_id` so we can fetch the `SenderKey`

crates/context/src/lib.rs
```diff
@@ -758,26 +756,20 @@ impl ContextManager {
...
-            temp[32..].copy_from_slice(&key);
-            temp
-        };
+            .and_then(|ctx_identity| ctx_identity.sender_key);
 
-        Ok(SigningKey::from_keypair_bytes(&combined)?)
+        match key {
+            Some(key) => Ok(Some(PrivateKey::from(key))),
+            None => Ok(None),
+        }
```
miraclx: alt:

```suggestion
        Ok(key.map(PrivateKey::from))
```

crates/crypto/src/lib.rs
```diff
@@ -1,11 +1,12 @@
-use calimero_primitives::identity::PublicKey;
-use ed25519_dalek::SigningKey;
+use calimero_primitives::identity::{PrivateKey, PublicKey};
+use ed25519_dalek::hazmat::ExpandedSecretKey;
```
miraclx: I'd [avoid using this directly](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/hazmat/index.html#-warning-hazmat)

crates/crypto/src/lib.rs
```diff
@@ -110,10 +115,14 @@ mod tests {
         let verifier = SigningKey::generate(&mut csprng);
         let invalid = SigningKey::generate(&mut csprng);
 
-        let signer_shared_key =
-            SharedKey::new(&signer, &(*verifier.verifying_key().as_bytes()).into());
-        let invalid_shared_key =
-            SharedKey::new(&invalid, &(*invalid.verifying_key().as_bytes()).into());
+        let signer_shared_key = SharedKey::new(
+            &PrivateKey::from(signer.to_bytes()),
```
miraclx: use `PrivateKey::random` above instead

crates/node/src/lib.rs
```diff
@@ -381,24 +383,22 @@ impl Node {
...
+            let possible_sending_key = self
+                .ctx_manager
+                .get_sender_key(&context.id, &executor_public_key)?;
 
-            let Some(our_identity) = identities.into_iter().choose(&mut thread_rng()) else {
-                bail!("no identities found for context: {}", context.id);
+            let sending_key = match possible_sending_key {
+                Some(key) => key,
+                None => todo!(), // initiate sync
             };
```
```suggestion
            let sender_key = self
                .ctx_manager
                .get_sender_key(&context.id, &executor_public_key)?
                .ok_or_eyre("expected own identity to have sender key");
```

crates/node/src/lib.rs
```diff
@@ -381,24 +383,22 @@ impl Node {
...
-                .ctx_manager
-                .get_own_signing_key(&context.id, &our_identity)?;
-
-            let shared_key = SharedKey::from_sk(&our_sending_key);
+            let shared_key = SharedKey::from_sk(&sending_key);
 
             let artifact_encrypted = shared_key
                 .encrypt(outcome.artifact.clone(), [0; aead::NONCE_LEN])
-                .unwrap();
+                .ok_or(eyre!("Encryption failed"))?;
```
```suggestion
                .ok_or_eyre("encryption failed")?;
```

crates/node/src/lib.rs
```diff
@@ -305,26 +305,28 @@ impl Node {
...
             } => {
-                let identities = self.ctx_manager.get_context_owned_identities(context_id)?;
+                let possible_sending_key =
+                    self.ctx_manager.get_sender_key(&context_id, &author_id)?;
 
-                let Some(our_identity) = identities.into_iter().choose(&mut thread_rng()) else {
-                    bail!("no identities found for context: {}", context_id);
+                let sending_key = match possible_sending_key {
+                    Some(key) => key,
+                    None => todo!(), //initiate sync
```
miraclx: for example:

https://github.com/calimero-network/core/blob/db644002b9093b0363e2af55347ebb849161e07e/crates/node/src/lib.rs#L350-L354

I'd do

```rust
let Some(sender_key) = self
    .ctx_manager
    .get_sender_key(&context.id, &executor_public_key)?
else {
    return self.initiate_sync(context_id, source).await;
};
```

leave the rest of it to sync

crates/node/src/lib.rs
```diff
@@ -305,26 +305,28 @@ impl Node {
...
 
-                let our_sending_key = self
-                    .ctx_manager
-                    .get_own_signing_key(&context_id, &our_identity)?;
+                let shared_key = SharedKey::from_sk(&sending_key);
 
-                let shared_key = SharedKey::from_sk(&our_sending_key);
+                let artifact = &shared_key
+                    .decrypt(artifact.into_owned(), [0; aead::NONCE_LEN])
+                    .ok_or_else(|| eyre!("Failed to decrypt message"))?;
```
```suggestion
                    .ok_or_eyre("failed to decrypt message")?;
```

crates/node/src/sync/blobs.rs
```diff
@@ -65,11 +65,16 @@ impl Node {
...
-            .get_own_signing_key(&context.id, &our_identity)?;
+            .get_sender_key(&context.id, &our_identity)?;
 
-        let shared_key = SharedKey::new(&our_sending_key, &their_identity);
+        let sending_key = match possible_sending_key {
+            Some(key) => key,
+            None => todo!(),
+        };
+
+        let shared_key = SharedKey::new(&sending_key, &their_identity);
```
miraclx: we are not to use sender keys for sync, but DHKE

petarjuki7: Not sure what you mean, check if fixed in new commit.

miraclx: the rest are resolved, but here it shouldn't be our "sending key" x their identity, but our private key x their public key.

sync uses diffie-hellman key exchange, not sender keys, since it's a direct interaction between 2 peers

it's broadcast that uses sender keys, since it's 1 to many.

crates/node/Cargo.toml
```diff
@@ -9,24 +9,27 @@ license.workspace = true
...
 eyre.workspace = true
 futures-util = { workspace = true, features = ["io"] }
 libp2p.workspace = true
 owo-colors.workspace = true
 rand.workspace = true
+ring.workspace = true
```
miraclx: you should be able to save this dependency, we don't need to import `ring::aead` to determine the size of the statically-sized array

crates/node/Cargo.toml
```diff
@@ -9,24 +9,27 @@ license.workspace = true
...
 
 calimero-context = { path = "../context" }
+calimero-crypto = { path = "../crypto" }
 calimero-blobstore = { path = "../store/blobs" }
 calimero-network = { path = "../network" }
 calimero-node-primitives = { path = "../node-primitives" }
 calimero-primitives = { path = "../primitives" }
 calimero-runtime = { path = "../runtime" }
 calimero-server = { path = "../server", features = ["jsonrpc", "websocket", "admin"] }
+calimero-storage = { path = "../storage" }
```
miraclx: ^

crates/crypto/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+#[derive(Debug)]
+pub struct Record {
+    pub token: Vec<u8>,
+    pub nonce: [u8; aead::NONCE_LEN],
+}
+
+impl SharedKey {
+    pub fn new(sk: &PrivateKey, pk: &PublicKey) -> Self {
+        SharedKey {
+            key: (Scalar::from_bytes_mod_order(**sk)
```
miraclx: do the tests pass? I'd expect `Scalar::from_bytes_mod_order(sha512(bytes)[..32])`, so it shouldn't work.

any reason we're not using [`SigningKey::from_bytes`](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/struct.SigningKey.html#method.from_bytes)?

crates/node/Cargo.toml
miraclx: this was marked as resolved?

crates/node/src/lib.rs
```diff
@@ -302,12 +305,23 @@ impl Node {
...
+                    .decrypt(artifact.into_owned(), [0; aead::NONCE_LEN])
+                    .ok_or_eyre("failed to decrypt message")?;
+
                 self.handle_state_delta(
                     source,
                     context_id,
                     author_id,
                     root_hash,
-                    artifact.into_owned(),
+                    artifact.to_vec(),
```
miraclx: you shouldn't need this

```suggestion
                    artifact,
```

crates/node/src/lib.rs
```diff
@@ -302,12 +305,23 @@ impl Node {
...
+                let Some(sender_key) = self.ctx_manager.get_sender_key(&context_id, &author_id)?
+                else {
+                    return self.initiate_sync(context_id, source).await;
+                };
+
+                let shared_key = SharedKey::from_sk(&sender_key);
+
+                let artifact = &shared_key
+                    .decrypt(artifact.into_owned(), [0; aead::NONCE_LEN])
+                    .ok_or_eyre("failed to decrypt message")?;
```
miraclx: better yet, just move all of these to `handle_state_delta`

crates/node/src/sync.rs
```diff
@@ -23,25 +25,47 @@ pub struct SyncConfig {
...
+    stream: &mut Stream,
+    message: &StreamMessage<'_>,
+    shared_key: Option<SharedKey>,
+) -> EyreResult<()> {
+    let base_data = borsh::to_vec(message)?;
+
+    let data = match shared_key {
+        Some(key) => match key.encrypt(base_data, [0; aead::NONCE_LEN]) {
+            Some(data) => data,
+            None => bail!("Encryption failed"),
```
miraclx: use lowercase for errors

chefsale:
What is the status on this one? @miraclx @petarjuki7

petarjuki7:
> What is the status on this one? @miraclx @petarjuki7

This should be the last commit before merge

crates/node/src/sync/state.rs
```diff
@@ -71,6 +73,17 @@ impl Node {
...
         let mut sqx_out = Sequencer::default();
 
+        let possible_sending_key = self
+            .ctx_manager
+            .get_sender_key(&context.id, &our_identity)?;
+
+        let sending_key = match possible_sending_key {
+            Some(key) => key,
+            None => todo!(),
+        };
```
miraclx: this should match what you did in `blobs`, use the private key, not the sender key

crates/node/src/sync/state.rs
```diff
@@ -110,6 +131,17 @@ impl Node {
...
         };
 
+        let possible_sending_key = self
+            .ctx_manager
+            .get_sender_key(&context.id, &our_identity)?;
+
+        let sending_key = match possible_sending_key {
+            Some(key) => key,
+            None => todo!(),
+        };
```
miraclx: same

crates/node/src/sync/blobs.rs
```diff
@@ -139,13 +148,25 @@ impl Node {
...
         };
 
+        let possible_sending_key = self
+            .ctx_manager
+            .get_sender_key(&context.id, &our_identity)?;
+
+        let sending_key = match possible_sending_key {
+            Some(key) => key,
+            None => todo!(),
+        };
```
miraclx: use private key

crates/node/src/sync.rs
```diff
@@ -23,25 +24,47 @@ pub struct SyncConfig {
...
         return Ok(None);
     };
 
-    let message = borsh::from_slice(&message?.data)?;
+    let message_data = message?.data.into_owned();
+
+    let data = match shared_key {
+        Some(key) => match key.decrypt(message_data, [0; 12]) {
+            Some(data) => data,
+            None => bail!("Encryption failed"),
```
miraclx:
```suggestion
            None => bail!("decryption failed"),
```

crates/context/src/lib.rs
```diff
@@ -400,6 +401,7 @@ impl ContextManager {
 
         let Some(ContextIdentityValue {
             private_key: Some(requester_secret),
+            sender_key: None,
```
miraclx: we shouldn't care about `sender_key` here, and if we did it's almost a 100% certainty that when `private_key` is `Some`, `sender_key` will be too.

```suggestion
            ..
```

crates/context/src/lib.rs
```diff
@@ -309,9 +309,10 @@ impl ContextManager {
         }
 
         handle.put(
-            &ContextIdentityKey::new(context.id, identity_secret.public_key()),
+            &ContextIdentityKey::new(context.id, identity_secret.public_key()), //local node saving its local identity
```
miraclx: artifact of development iteration? or you think this is useful because the code is unreadable?

```suggestion
            &ContextIdentityKey::new(context.id, identity_secret.public_key()),
```

miraclx:
> What is the status on this one?

nearly there

crates/node/src/lib.rs
```diff
@@ -335,6 +336,16 @@ impl Node {
...
         }
 
+        let Some(sender_key) = self.ctx_manager.get_sender_key(&context_id, &author_id)? else {
+            return self.initiate_sync(context_id, source).await;
+        };
+
+        let shared_key = SharedKey::from_sk(&sender_key);
+
+        let artifact = &shared_key
+            .decrypt(artifact, [0; 12])
```
miraclx, line 346: thought I'd mention.. ideally, IVs shouldn't be constant as that impairs the effectiveness of GCM encryption and this can be passed in plain text, but that's an easy fix for a later PR, we can merge with this

crates/context/src/lib.rs
petarjuki7: Something weird is happening, I keep commiting things I don't actually have locally...

miraclx (commented):
lgtm, please test it out and make sure blob sync works as expected, once that is confirmed, I'll stamp it

miraclx (commented):
ah, key sharing is missing

crates/node/src/interactive_cli.rs
miraclx, line 1: ^

crates/node/src/sync.rs
```diff
@@ -16,6 +16,7 @@ use crate::types::{InitPayload, StreamMessage};
 use crate::Node;
 
 mod blobs;
+mod key_share;
```
miraclx: `key` is fine, blobs could've been `blob_share` as well, but isn't

crates/node/src/sync.rs
```diff
@@ -223,6 +227,10 @@ impl Node {
                 )
                 .await?
             }
+            InitPayload::KeyShare {} => {
```
miraclx:
```suggestion
            InitPayload::KeyShare => {
```

crates/node/src/sync/key_share.rs
```diff
@@ -0,0 +1,92 @@
...
+        context: &mut Context,
+        our_identity: PublicKey,
+        stream: &mut Stream,
+    ) -> eyre::Result<()> {
+        send(
+            stream,
+            &StreamMessage::Init {
+                context_id: context.id,
+                party_id: our_identity,
+                payload: InitPayload::KeyShare {},
```
miraclx:
```suggestion
                payload: InitPayload::KeyShare,
```

crates/node/src/sync/key_share.rs
```diff
@@ -0,0 +1,92 @@
...
+                ..
+            } => sender_key,
+            unexpected @ (StreamMessage::Init { .. }
+            | StreamMessage::Message { .. }
+            | StreamMessage::OpaqueError) => {
+                bail!("unexpected message: {:?}", unexpected)
+            }
+        };
+
+        // Do I store "his" SenderKey somewhere?
```
miraclx: yessir

crates/node/src/sync/key_share.rs
```diff
@@ -0,0 +1,92 @@
...
+            StreamMessage::Message {
+                payload: MessagePayload::KeyShare { sender_key },
+                ..
+            } => sender_key,
+            unexpected @ (StreamMessage::Init { .. }
+            | StreamMessage::Message { .. }
+            | StreamMessage::OpaqueError) => {
+                bail!("unexpected message: {:?}", unexpected)
+            }
+        };
```
miraclx: this should be "their end of the handshake", which we expect to include their public key.

the same as blob and state sync

crates/node/src/sync/key_share.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        send(
+            stream,
+            &StreamMessage::Message {
+                sequence_id: sequencer.next(),
+                payload: MessagePayload::KeyShare { sender_key },
+            },
+            Some(shared_key), // or None?
+        )
+        .await?;
```
miraclx: announce yourself first, via the unencrypted `Init`

then share the sender key over the encrypted `Message`

crates/node/src/sync/state.rs
```diff
@@ -144,7 +144,8 @@ impl Node {
                     application_id: context.application_id,
                 },
             },
-            Some(shared_key),
+            None, // I think it should be None here,
+                  // because the first recv in the function above has to have some way of decrypting it?
```
miraclx: `Init` is always unencrypted, yes

crates/node/src/types.rs
```diff
@@ -47,12 +47,14 @@ pub enum InitPayload {
         root_hash: Hash,
         application_id: ApplicationId,
     },
+    KeyShare {},
 }
-
+// this I was encrypting
```
miraclx:
```suggestion
    KeyShare,
}

```

file.patch
miraclx, line 1: ^

crates/context/src/lib.rs
```diff
@@ -799,6 +799,24 @@ impl ContextManager {
...
+        sender_key: &PrivateKey,
+    ) -> EyreResult<()> {
+        let mut handle = self.store.handle();
+        handle.put(
+            &ContextIdentityKey::new(*context_id, *public_key),
+            &ContextIdentityValue {
+                private_key: None,
+                sender_key: Some(**sender_key),
+            },
+        )?;
```
miraclx: let's not assume to not have the private key (in which case it would be overridden)

read from storage, edit, and write back

crates/node/src/sync/key.rs
```diff
@@ -33,6 +33,30 @@ impl Node {
...
+            .get_private_key(context.id, our_identity)?
+            .ok_or_eyre("expected own identity to have private key")?;
+
+        let shared_key = SharedKey::new(&private_key, &their_identity);
+
+        let Some(ack) = recv(stream, self.sync_config.timeout, Some(shared_key)).await? else {
+            bail!("connection closed while awaiting state sync handshake");
+        };
+
         let sender_key = match ack {
```
miraclx: we're done with the handshake, now we're sending the messages

```suggestion
        let Some(msg) = recv(stream, self.sync_config.timeout, Some(shared_key)).await? else {
            bail!("connection closed while awaiting key share");
        };

        let sender_key = match msg {
```

crates/node/src/sync/key.rs
```diff
@@ -45,7 +69,25 @@ impl Node {
...
+                        &their_identity,
+                        &sender_key,
+                    )?;
+                }
+            }
+            None => {
+                self.ctx_manager
+                    .update_sender_key(&context.id, &their_identity, &sender_key)?;
+            }
+        }
```
miraclx: unconditionally call `self.ctx_manager.update_sender_key`

crates/node/src/sync.rs
```diff
@@ -200,6 +227,10 @@ impl Node {
                 )
                 .await?
             }
+            InitPayload::KeyShare => {
+                self.handle_key_share_request(context, their_identity, stream)
+                    .await?
+            }
```
miraclx: move this to the first match so it matches the order on the sender side

crates/context/src/lib.rs
```diff
@@ -806,10 +806,14 @@ impl ContextManager {
...
+            .get(&ContextIdentityKey::new(*context_id, *public_key))?
+            .and_then(|ctx_identity| ctx_identity.private_key);
+
         handle.put(
             &ContextIdentityKey::new(*context_id, *public_key),
             &ContextIdentityValue {
-                private_key: None,
+                private_key: private_key,
                 sender_key: Some(**sender_key),
             },
```
miraclx: why not

```suggestion
        let mut identity = handle
            .get(&ContextIdentityKey::new(*context_id, *public_key))?
            .ok_or_eyre("unknown identity");

        identity.sender_key = Some(**sender_key);

        handle.put(
            &ContextIdentityKey::new(*context_id, *public_key),
            &identity,
```

crates/node/src/sync/key.rs
```diff
@@ -53,11 +53,11 @@ impl Node {
 
         let shared_key = SharedKey::new(&private_key, &their_identity);
 
-        let Some(ack) = recv(stream, self.sync_config.timeout, Some(shared_key)).await? else {
-            bail!("connection closed while awaiting state sync handshake");
+        let Some(msg) = recv(stream, self.sync_config.timeout, Some(shared_key)).await? else {
+            bail!("connection closed while awaiting key share");
```
miraclx: I think key share sync should be bi-directional, so before receiving, we should share ours

crates/crypto/src/lib.rs
```diff
@@ -0,0 +1,126 @@
+use calimero_primitives::identity::{PrivateKey, PublicKey};
+use ed25519_dalek::{SecretKey, SigningKey};
+use ring::aead;
+use serde::{Deserialize, Serialize};
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub struct SharedKey {
```
miraclx: is this right? why are we serializing this?

miraclx (approved):
aside this, rest looks good to me

PR #868, opened 2024-10-26
chore: improved the descriptions and usage of meroctl commands

chefsale:
Improved the meroctl command descriptions and usage.

### DIFF
--- a/crates/meroctl/src/cli.rs
+++ b/crates/meroctl/src/cli.rs
@@ -11,11 +11,11 @@ mod jsonrpc;
 
 use app::AppCommand;
 use context::ContextCommand;
-use jsonrpc::JsonRpcCommand;
+use jsonrpc::CallCommand;
 
 pub const EXAMPLES: &str = r"
   # List all applications
-  $ meroctl -- --home data --node-name node1 app ls
+  $ meroctl -- --node-name node1 app ls
 
   # List all contexts
   $ meroctl -- --home data --node-name node1 context ls
@@ -41,8 +41,7 @@ pub struct RootCommand {
 pub enum SubCommands {
     Context(ContextCommand),
     App(AppCommand),
-    #[command(name = "call")]
-    JsonRpc(JsonRpcCommand),
+    JsonRpc(CallCommand),
 }
 
 #[derive(Debug, Parser)]
--- a/crates/meroctl/src/cli.rs
+++ b/crates/meroctl/src/cli.rs
@@ -14,17 +14,11 @@ use context::ContextCommand;
 use jsonrpc::JsonRpcCommand;
 
 pub const EXAMPLES: &str = r"
-  # Initialize a new node
-  $ meroctl --home data/ --node-name node1 init
+  # List all applications
+  $ meroctl -- --home data --node-name node1 app ls
 
-  # Configure an existing node
-  $ meroctl --home data/ --node-name node1 config --server-host 143.34.182.202 --server-port 3000
-
-  # Run a node as a peer
-  $ meroctl --home data/ --node-name node1 run
-
-  # Run a node as a coordinator
-  $ meroctl --home data/ --node-name node1 run --node-type coordinator
+  # List all contexts
+  $ meroctl -- --home data --node-name node1 context ls
 ";
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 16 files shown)


crates/meroctl/src/cli/context/get.rs
```diff
@@ -3,20 +3,25 @@ use eyre::{bail, Result as EyreResult};
...
 
 #[derive(Parser, Debug)]
+#[command(about = "Fetch details about the context")]
 pub struct GetCommand {
-    #[clap(long, short)]
+    #[arg(
+        value_name = "METHOD",
+        help = "Method to fetch details",
+        value_parser = clap::builder::EnumValueParser::<GetRequest>::new()
+    )]
```
miraclx: `value_enum` should do the trick

```suggestion
    #[arg(
        value_name = "METHOD",
        help = "Method to fetch details",
        value_enum,
    )]
```

crates/meroctl/src/cli/context/get.rs
```diff
@@ -3,20 +3,25 @@ use eyre::{bail, Result as EyreResult};
 use libp2p::identity::Keypair;
 use libp2p::Multiaddr;
 use reqwest::Client;
+use strum::VariantNames;
```
miraclx: we don't need strum here

crates/meroctl/src/cli.rs
```diff
@@ -47,7 +41,7 @@ pub struct RootCommand {
 pub enum SubCommands {
     Context(ContextCommand),
     App(AppCommand),
-    #[command(alias = "call")]
+    #[command(name = "call")]
     JsonRpc(JsonRpcCommand),
```
miraclx: alternatively, we can rename the command to `CallCommand`

crates/meroctl/src/cli/app.rs
```diff
@@ -10,7 +11,20 @@ mod get;
 mod install;
 mod list;
 
+pub const EXAMPLES: &str = r"
+  # List all applications
+  $ meroctl -- --home data --node-name node1 application ls
```
miraclx: I think we should skip `--home data` in all these examples, and let it use the default

crates/meroctl/src/cli/app.rs
```diff
@@ -10,7 +11,20 @@ mod get;
...
+
+  # Get details of an application
+  $ meroctl app get <APP_ID>
```
miraclx: `--node-name ____`

crates/meroctl/src/cli/context/create.rs
```diff
@@ -26,22 +26,37 @@ use crate::cli::RootArgs;
...
-    #[clap(long, short = 'a')]
+    #[clap(
+        long,
+        short = 'a',
+        help = "The application ID to attach to the context"
+    )]
     application_id: Option<ApplicationId>,
 
-    #[clap(long, short = 'p')]
+    #[clap(long, short = 'p', help = "The parameters to pass to the application")]
```
miraclx: let's indicate these are initialization parameters

crates/meroctl/src/cli/context/create.rs
```diff
@@ -26,22 +26,37 @@ use crate::cli::RootArgs;
...
 #[derive(Debug, Parser)]
+#[command(about = "Create a new context")]
 pub struct CreateCommand {
-    /// The application ID to attach to the context
-    #[clap(long, short = 'a')]
+    #[clap(
+        long,
+        short = 'a',
+        help = "The application ID to attach to the context"
+    )]
```
miraclx, line 35: any reason rustdoc syntax was swapped out for `help = "__"`? imo, rustdocs are better.

crates/meroctl/src/cli/jsonrpc.rs
```diff
@@ -4,32 +4,43 @@ use calimero_server_primitives::jsonrpc::{
...
     pub args_json: String,
 
-    /// Id of the JsonRpc execute call
-    #[arg(long, default_value = "dontcare")]
+    #[arg(
+        long,
+        default_value = "dontcare",
+        help = "Id of the JsonRpc execute call"
+    )]
     pub id: String,
```
miraclx: this shouldn't even be here, ideally, this is auto defined, or maybe, we make it an option?

crates/meroctl/src/cli/context/create.rs
chefsale: because we had a mix of both and I prefered it to be defined here as people either forget rust docs or remove them.

miraclx (approved):

PR #886, opened 2024-10-28
[DO NOT MERGE/ Reference]feat: expose access to the Proxy Contract

petarjuki7:
# Refactor Context Client API for Proxy Contract Support

Abstracts query and mutate methods through a shared `Environment` trait, enabling support for both config and proxy contracts. This allows for foundation for future proxy contract integration

Key changes:
- Added Environment trait with associated Query/Mutate types
- Implemented trait for both ContextConfigClient and ContextProxyClient
- Added query methods needed for the rpc in the context manager

### DIFF
--- a/crates/context/config/src/types.rs
+++ b/crates/context/config/src/types.rs
@@ -401,3 +401,5 @@ pub enum ProposalAction {
         value: Box<[u8]>,
     },
 }
+
+pub type ProposalId = u32;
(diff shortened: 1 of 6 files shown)


crates/context/config/src/types.rs
```diff
@@ -372,3 +374,30 @@ impl<'a, T: Deserialize<'a>> Signed<T> {
...
+        num_approvals: u32,
+    },
+    SetActiveRequestsLimit {
+        active_proposals_limit: u32,
+    },
+    SetContextValue {
+        key: Box<[u8]>,
+        value: Box<[u8]>,
+    },
+}
```
miraclx, line 403: would've be better if the context-proxy contract had a base crate that defines primitives

xilosada:
Closing this PR as it's not relevant anymore

PR #887, opened 2024-10-29
feat: blockchain developer interface

danwilliams:
# Purpose

This brings in the developer interface that's needed to support the blockchain integration, with the WASM bridge.

  - It has moved from the originally-intended closure registration approach to using a predefined function that must be implemented in the guest code.
  - `handle_proposal()` on `Node` is how the host calls into the guest implementation.
  - `modify_value()` is still needed as it's the bridge function that lets guest code modify blockchain values.

The path for `modify_value()` goes from:

  1. Guest code calls it via `sdk/src/env.rs`,
  2. Which uses the syscall in `sdk/src/sys.rs`,
  3. Which gets handled by `VMHostFunctions`/`VMLogic` through `runtime/src/logic.rs` and `imports.rs`.

This relies on the standard method calling pattern used elsewhere in the codebase.

# Description of changes

  - Added a `modify_value()` function to the guest WASM and host functions in order for values on the blockchain to be modified. The intention is that this should then, from the host VMLogic side, call through to the appropriate blockchain function.

  - Added a `handle_proposal()` function to Node, to call a corresponding function on the guest (this is expected to exist).

# How to test

This cannot really be tested until it's hooked up to the wider blockchain integration.

# Notes / advisories

  - It needs hooking in to the wider blockchain integration, likely at the places indicated.
  - The parameters to pass to the guest function need to be defined, along with the exact nature of the data it should return.

### DIFF
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -108,7 +108,6 @@ pub struct VMLogic<'a> {
     events: Vec<Event>,
     actions: Vec<Vec<u8>>,
     root_hash: Option<[u8; 32]>,
-    proposal_function: Option<Box<dyn FnMut() -> Vec<u8>>>,
 }
 
 impl<'a> VMLogic<'a> {
@@ -124,7 +123,6 @@ impl<'a> VMLogic<'a> {
             events: vec![],
             actions: vec![],
             root_hash: None,
-            proposal_function: None,
         }
     }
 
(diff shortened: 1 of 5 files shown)


crates/runtime/src/logic.rs
```diff
@@ -142,6 +142,17 @@ impl<'a> VMLogic<'a> {
...
+    /// Modifies a value through the blockchain bridge
+    ///
+    /// # Parameters
+    ///
+    /// * `value` - The value to be modified on the blockchain.
+    ///
+    pub fn modify_value(&mut self, _value: Vec<u8>) -> VMLogicResult<()> {
+        // TODO: Implement actual blockchain call
+        Ok(())
+    }
```
xilosada, line 155: Here we should parse the [proxy contract actions](https://github.com/calimero-network/core/pull/854/files#diff-d7aee4a31d8be649fe544572f6a08467e3bbfa6780d448eebb3047346a9d1e9cR68), not just `modify_value`

xilosada: this should be submitProposal(actions)

danwilliams:
Closing as this is superseded by https://github.com/calimero-network/core/pull/899

Issue #772, opened 2024-10-07
Implement machine-consumable CLI output


fbozic:
Implement machine-consumable CLI output so it can be used during E2E tests. Probably Json is the best candidate.

petarjuki7:
Is being worked on here: #801

PR #888, opened 2024-10-30
feat(meroctl): standardise command output

fbozic:
Resolves: #772

### DIFF
--- a/crates/meroctl/src/cli.rs
+++ b/crates/meroctl/src/cli.rs
@@ -92,6 +92,7 @@ impl RootCommand {
                 Err(err) => CliError::Other(err),
             };
             environment.output.write(&err);
+            return Err(err);
         }
 
         return Ok(());


crates/meroctl/src/cli/context.rs
```diff
@@ -53,14 +53,14 @@ pub enum ContextSubCommands {
 }
 
 impl ContextCommand {
-    pub async fn run(self, args: RootArgs) -> EyreResult<()> {
+    pub async fn run(self, context: CommandContext) -> EyreResult<()> {
```
miraclx: `ContextCommand` / `CommandContext` 😅

crates/meroctl/src/cli.rs
```diff
@@ -54,14 +55,31 @@ pub struct RootArgs {
     /// Name of node
     #[arg(short, long, value_name = "NAME")]
     pub node_name: String,
+
+    #[arg(long, value_name = "FORMAT")]
+    pub output_format: Format,
+}
+
+pub struct CommandContext {
```
miraclx: `Environment` maybe? what do you think?

crates/meroctl/src/common.rs
```diff
@@ -88,3 +88,11 @@ pub enum RequestType {
...
     Delete,
 }
+pub async fn craft_failed_request_message(response: Response, message: &str) -> EyreResult<String> {
+    let status = response.status();
+    let error_text = response.text().await?;
+    Ok(format!(
+        "{} - Status: {}, Error: {}",
+        message, status, error_text
+    ))
+}
```
miraclx: we should define a standardized error structure which will be serialized to JSON

something like

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "type", content = "data")]
enum Error<T> {
    ApiError { code: __, data: __ },
    CliError(T),
}
```

crates/meroctl/src/cli/app/get.rs
```diff
@@ -12,10 +20,29 @@ pub struct GetCommand {
...
+#[derive(ValueEnum, Debug, Clone)]
+pub enum GetValues {
+    Details,
+}
+
+#[derive(Debug, Serialize)]
+struct OutputReport(GetApplicationResponse);
+
+impl Display for OutputReport {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
miraclx: Alternatively, you can define a new trait?

```rust
trait Report {
    fn report(&self) {}
}

impl Report for GetApplicationResponse {
    fn report(&self) {
        // here, you can use `println!` directly
    }
}
```

crates/meroctl/src/common.rs
fbozic: I am not sure if I agree with this ATM, plus this would require refactoring how we handle return values (instead of just bailing).
for example, kubectl will ignore the output type if there is an internal error and it will return just the text string
Example 
```
kgp doesnt-exist-123 -o yaml
Error from server (NotFound): pods "doesnt-exist-123" not found
```

miraclx: > plus this would require refactoring how we handle return values (instead of just bailing).

not necessarily.. it can be `bail!(Error::from(response))`

and then at the root, that can be downcasted from `eyre::Report`, passed to `Output`, and then the process exits

> for example, kubectl will ignore the output type if there is an internal error and it will return just the text string

it's fine, just depends on if we want structured errors, cause without it, failures will be more involved to parse from stderr, unless we always want users to "see" exactly what we report here (implying the presence of a terminal).

chefsale (approved):

crates/meroctl/src/cli.rs
```diff
@@ -60,26 +64,78 @@ pub struct RootArgs {
...
+}
+
+impl Report for CliError {
+    fn report(&self) {
+        println!("{}", self);
+    }
+}
+
+#[derive(Debug, Serialize, ThisError)]
+#[error("ApiError")]
```
miraclx: let's discourage these patterns, to clarify we only used these in meroctl when displayability of the error was unimportant, if that's the same case here, then it's fine, otherwise, this should be the `Display` format for this error

```suggestion
#[error("{status_code}: {message}")]
```

or something like that

crates/meroctl/src/cli.rs
```diff
@@ -60,26 +64,78 @@ pub struct RootArgs {
...
+pub struct ApiError {
+    pub status_code: u16,
+    pub message: String,
+}
+
+fn serialize_eyre_report<S>(report: &EyreReport, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    serializer.serialize_str(&report.to_string())
```
miraclx: use [Serializer::collect_str](https://docs.rs/serde/latest/serde/ser/trait.Serializer.html#method.collect_str) instead, delegating to the `eyre::Report: Display`

```suggestion
    serializer.collect_str(&report)
```

crates/meroctl/src/cli/context/delete.rs
```diff
@@ -14,33 +14,32 @@ pub struct DeleteCommand {
...
 
-impl DeleteCommand {
-    pub async fn run(self, context: CommandContext) -> EyreResult<()> {
-        let config = load_config(&context.args.home, &context.args.node_name)?;
-
-        self.delete_context(fetch_multiaddr(&config)?, &Client::new(), &config.identity)
-            .await
+impl Report for DeleteContextResponse {
+    fn report(&self) {
+        println!("is_deleted: {}", self.data.is_deleted);
```
miraclx, line 19: ideally, pretty reports should be pretty, but this is fine for this iteration, we can do better in another

crates/meroctl/src/output.rs
```diff
@@ -34,3 +34,21 @@ impl Output {
...
     }
 }
+
+#[derive(Clone, Debug, Serialize)]
+pub struct InfoLine<'a>(pub &'a str);
+
+impl Report for InfoLine<'_> {
+    fn report(&self) {
+        println!("{}", self.0);
+    }
```
miraclx: will this be prefixed? otherwise should we impl `Report` for `String`?

crates/meroctl/src/output.rs
```diff
@@ -34,3 +34,21 @@ impl Output {
...
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ErrorLine<'a>(pub &'a str);
+
+impl Report for ErrorLine<'_> {
+    fn report(&self) {
+        println!("[ERROR] {}", self.0);
```
miraclx:
```suggestion
        println!("[{}] {}", "ERROR".red(), self.0);
```

crates/meroctl/src/cli/context/delete.rs
fbozic: yeah I agree, here is the ticket for the pretty print: https://github.com/calimero-network/core/issues/902

crates/meroctl/src/output.rs
fbozic: added [INFO] prefix

miraclx (approved):

PR #899, opened 2024-10-31
feat: Blockchain WASM integration

danwilliams:
# Purpose

This brings in the developer interface that's needed to support the blockchain integration, with the WASM bridge.

  - It has changed quite a bit from the original design that was part of https://github.com/calimero-network/core/pull/887.
  - `send_proposal()` is the bridge function that lets guest code pass actions to end up in a proposal.

The path for `send_proposal()` is:

  1. Guest code calls it via `sdk/src/env.rs`,
  2. Which uses the syscall in `sdk/src/sys.rs`,
  3. Which gets handled by `VMHostFunctions`/`VMLogic` through `runtime/src/logic.rs` and `imports.rs`.

This relies on the standard method calling pattern used elsewhere in the codebase.

# Description of changes

  - Added a `send_proposal()` function to the guest WASM and host functions in order for values on the blockchain to be modified. The intention is that this should then, from the host VMLogic side, call through to the appropriate blockchain function.

# How to test

This cannot really be tested until it's hooked up to the wider blockchain integration.

# Notes / advisories

  - It needs hooking in to the wider blockchain integration, likely at the places indicated.
  - The `ProposalAction` enum is needed, at which point this code will be updated in a couple of places to use it.


crates/runtime/src/logic.rs
```diff
@@ -227,9 +239,7 @@ impl VMHostFunctions<'_> {
 
         String::from_utf8(buf).map_err(|_| HostError::BadUTF8.into())
     }
-}
 
-impl VMHostFunctions<'_> {
```
miraclx: this is semantic separation, please restore it

the above are utility methods, not host functions

and below are host functions

crates/runtime/src/logic.rs
```diff
@@ -556,4 +566,24 @@ impl VMHostFunctions<'_> {
...
+    ///
+    /// # Parameters
+    ///
+    /// * `ptr` - Pointer to the start of the action data in WASM memory.
+    /// * `len` - Length of the action data.
+    ///
+    pub fn send_proposal(&mut self, ptr: u64, len: u64) -> VMLogicResult<()> {
+        // TODO: This will need to deserialise into ProposalActions when available
+        // TODO: So this will become Vec<ProposalAction>
+        let actions: Vec<Vec<u8>> = from_borsh_slice(&self.read_guest_memory(ptr, len)?)
```
miraclx: the runtime shouldn't care about the type being returned here, just receive a `Vec<u8>` and push it to a list of proposals on `VMLogic`, like we did with `send_action`

crates/runtime/src/logic.rs
```diff
@@ -142,6 +142,18 @@ impl<'a> VMLogic<'a> {
...
+    ///
+    /// # Parameters
+    ///
+    /// * `actions` - The actions that make up the proposal.
+    ///               TODO: This will be a `Vec<ProposalAction>` when available.
+    ///
+    pub fn send_proposal(&mut self, _actions: Vec<Vec<u8>>) -> VMLogicResult<()> {
+        // TODO: Implement actual blockchain call
+        Ok(())
+    }
```
miraclx: we shouldn't need a function here, `send_proposal` should follow the model of accommodations made for `send_action`

crates/sdk/src/env.rs
```diff
@@ -208,3 +208,13 @@ pub fn time_now() -> u64 {
...
     u64::from_le_bytes(bytes)
 }
+
+/// Call the contract's `send_proposal()` function through the bridge.
+///
+/// # Parameters
+///
+/// * `actions` - The proposal actions, sent as raw data.
+///
+pub fn send_proposal(actions: &[u8]) {
```
miraclx: this should return the `ProposalId` as well, check with Xabi what the ideal type for that is

```suggestion
pub fn send_proposal(actions: &[u8]) -> ProposalId {
```

this will be determined by the runtime (random) and mapped there

`VMLogic` should have a `Map<ProposalId, Vec<u8>>`. Which eventually, should make it to `Outcome`.

Also ideally, `ProposalId` should be constant-size.

crates/runtime/src/logic.rs
danwilliams: Apologies - it's a little unorthodox and so I thought it was a mistake 😅 Restored! 👍

crates/runtime/src/logic.rs
danwilliams: Ahhhh, okay - I think that makes sense. Changed!

crates/runtime/src/logic.rs
danwilliams: Gotcha. I've added a `proposals` field and am now pushing it in the same way.

crates/sdk/src/env.rs
danwilliams: This is not clear to me, as I'm not sure where the proposal ID will come from. I've added the type as a `u64` for now, until the actual type is available to use, and also added a TODO... can you give me a hint as to where that ID will come from? Perhaps that will become clearer once the functionality using it is present?

crates/runtime/src/logic.rs
```diff
@@ -556,4 +558,28 @@ impl VMHostFunctions<'_> {
...
+    /// * `len` - Length of the action data.
+    ///
+    // TODO: The u64 below will need to become a ProposalId or similar, once
+    // TODO: available.
+    pub fn send_proposal(&mut self, ptr: u64, len: u64) -> VMLogicResult<u64> {
+        let actions_bytes: Vec<u8> = self.read_guest_memory(ptr, len)?;
+
+        self.with_logic_mut(|logic| logic.proposals.push(actions_bytes));
+
+        // TODO: It's not clear how this ProposalID will be obtained...
```
xilosada: You have to generate it and pass it to the contract and back to the guest

crates/sdk/src/env.rs
```diff
@@ -208,3 +208,15 @@ pub fn time_now() -> u64 {
...
+///
+/// # Parameters
+///
+/// * `actions` - The proposal actions, sent as raw data.
+///
+// TODO: The u64 below will need to become a ProposalId or similar, once
+// TODO: available.
+pub fn send_proposal(actions: &[u8]) -> u64 {
+    unsafe { sys::send_proposal(Buffer::from(actions)) }
+}
```
xilosada: can you make it available? This is part of the PR

crates/runtime/src/logic.rs
```diff
@@ -108,6 +108,7 @@ pub struct VMLogic<'a> {
     events: Vec<Event>,
     actions: Vec<Vec<u8>>,
     root_hash: Option<[u8; 32]>,
+    proposals: Vec<Vec<u8>>,
```
miraclx: so once the ID has been assigned (by `VMHostFunctions::send_proposal`), it needs to be here too

so this will become a `Map<[u8; 32], Vec<u8>>`

crates/sdk/src/env.rs
```diff
@@ -208,3 +208,15 @@ pub fn time_now() -> u64 {
...
+/// Call the contract's `send_proposal()` function through the bridge.
+///
+/// # Parameters
+///
+/// * `actions` - The proposal actions, sent as raw data.
+///
+// TODO: The u64 below will need to become a ProposalId or similar, once
+// TODO: available.
+pub fn send_proposal(actions: &[u8]) -> u64 {
+    unsafe { sys::send_proposal(Buffer::from(actions)) }
```
miraclx: receive the proposal ID from the runtime, and return it

crates/runtime/src/logic.rs
miraclx: `random_bytes`, must be fixed sized

crates/sdk/src/env/ext.rs
```diff
@@ -1,8 +1,120 @@
...
+    }
+
+    /// Finalise the proposal and send it to the blockchain.
+    #[must_use]
+    pub fn send(self) -> ProposalId {
+        #[expect(unused_mut, reason = "Needed by the unsafe code")]
+        let mut buf = [0; 32];
+        let actions = to_vec(&self.actions).unwrap();
+
+        unsafe { sys::send_proposal(Buffer::from(actions.as_slice()), BufferMut::new(buf)) }
```
miraclx:
```suggestion
        unsafe { sys::send_proposal(Buffer::from(&actions), BufferMut::new(&mut buf)) }
```

crates/sdk/macros/src/state.rs
```diff
@@ -50,6 +50,12 @@ impl ToTokens for StateImpl<'_> {
             impl #impl_generics ::calimero_sdk::state::AppState for #ident #ty_generics #where_clause {
                 type Event<#lifetime> = #event;
             }
+
+            impl #impl_generics #ident #ty_generics {
```
miraclx:
```suggestion
            impl #impl_generics #ident #ty_generics #where_clause {
```

crates/sdk/src/env/ext.rs
```diff
@@ -1,8 +1,120 @@
...
+
+/// Interface for interacting with external proposals for blockchain actions.
+#[derive(Clone, Copy, Debug)]
+pub struct External;
+
+impl External {
+    /// Create a new proposal. This will initially be a draft, until sent.
+    #[must_use]
+    pub fn propose(self) -> DraftProposal {
+        DraftProposal::default()
```
miraclx: can be `const` if a `const` `DraftProposal::new` was introduced

crates/runtime/src/logic.rs
```diff
@@ -556,4 +559,34 @@ impl VMHostFunctions<'_> {
...
+    ///                   memory.
+    /// * `actions_len` - Length of the action data.
+    /// * `actions_ptr` - Pointer to the start of the id data in WASM memory.
+    ///                   This should always be exactly 32 bytes in length.
+    ///
+    pub fn send_proposal(
+        &mut self,
+        actions_ptr: u64,
+        actions_len: u64,
+        id_ptr: u64,
```
miraclx, line 585: follow the model defined for `time_now`

```suggestion
        id_ptr: u64,
        id_len: u64,
```

crates/runtime/src/logic.rs
```diff
@@ -556,4 +559,34 @@ impl VMHostFunctions<'_> {
...
+    ) -> VMLogicResult<()> {
+        let actions_bytes: Vec<u8> = self.read_guest_memory(actions_ptr, actions_len)?;
+        let mut proposal_id = [0; 32];
+        rand::thread_rng().fill_bytes(&mut proposal_id);
+        drop(self.with_logic_mut(|logic| logic.proposals.insert(proposal_id, actions_bytes)));
```
miraclx, line 596:
```suggestion
        self.with_logic_mut(|logic| logic.proposals.insert(proposal_id, actions_bytes))?;
```

also please space out this block where you can, makes it easier to read and maintain, thanks

crates/runtime/src/logic.rs
```diff
@@ -123,6 +125,7 @@ impl<'a> VMLogic<'a> {
             events: vec![],
             actions: vec![],
             root_hash: None,
+            proposals: BTreeMap::new(),
```
miraclx, line 128: it still needs to make it out through `Outcome` so you need to tweak `VMLogic::finish`

crates/runtime/src/logic.rs
danwilliams: This is an `Option` here, not a `Result`...

crates/sdk/src/env/ext.rs
danwilliams: Not needed as this is already `&mut`.

danwilliams: Not needed as this is already `&mut`.

crates/runtime/src/logic.rs
miraclx: Ah, right.. okay

crates/sdk/src/env/ext.rs
miraclx: `buf`? it's `[u8; 32]`

danwilliams: So we can't do `&actions` there as a suitable `From` doesn't exist, hence the `.as_slice()` 🙂 

Meanwhile the `&mut` in `&mut buf` is unnecessary as it's already defined above as `&mut`.

miraclx: > So we can't do &actions there as a suitable From doesn't exist, hence the .as_slice() 🙂

Ah.. okay, was trying to satisfy `Buffer: From<&[u8]>`, thought it would auto-deref, anyway  `&*actions` would've done the trick.

> as it's already defined above as `&mut`.

you've declared it as mutable, but you've not gotten a mutable reference to it

Clippy will complain, yes.. it doesn't know better, here's other such cases: https://github.com/calimero-network/core/blob/5df5176b91708d89b6544a30370276f1d474a4c7/crates/sdk/src/env.rs#L201-L204

danwilliams: When I put it in, the compiler tells me to take it out as it's already present 🤔

miraclx: just.. follow [other](https://github.com/calimero-network/core/blob/5df5176b91708d89b6544a30370276f1d474a4c7/crates/sdk/src/env.rs#L101-L104) [references](https://github.com/calimero-network/core/blob/5df5176b91708d89b6544a30370276f1d474a4c7/crates/sdk/src/env.rs#L201-L204)

danwilliams: Ahhh, I'm with you now 😄 Sorry, I hadn't made the link

miraclx (approved):
lgtm ✨

crates/sdk/src/env/ext.rs
```diff
@@ -1,8 +1,133 @@
...
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
+            actions: Vec::new(),
+        }
+    }
+
+    /// Add an action to transfer tokens to an account.
+    #[must_use]
+    pub fn transfer(mut self, receiver: AccountId, amount: u64) -> Self {
```
miraclx, line 90: ah, we need to expose `function_call` as well

@xilosada, wdyt about the rest, keep them out for now?

crates/sdk/src/env/ext.rs
```diff
@@ -1,8 +1,133 @@
...
-use crate::sys::Buffer;
+use crate::sys::{Buffer, BufferMut};
+
+/// A blockchain proposal action.
+///
+/// This enum represents the different actions that can be executed against a
+/// blockchain, and combined into a proposal.
+///
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug, Eq, PartialEq)]
+pub enum ProposalAction {
```
miraclx, line 13: this should not need to be directly exposed (yet)

```suggestion
enum ProposalAction {
```

crates/sdk/macros/src/state.rs
```diff
@@ -50,6 +50,12 @@ impl ToTokens for StateImpl<'_> {
             impl #impl_generics ::calimero_sdk::state::AppState for #ident #ty_generics #where_clause {
                 type Event<#lifetime> = #event;
             }
+
+            impl #impl_generics #ident #ty_generics #where_clause {
+                fn external() -> ::calimero_sdk::env::ext::External {
+                    ::calimero_sdk::env::ext::External {}
```
miraclx, line 56: while this works, `External` is defined without them, so maybe this?

```suggestion
                    ::calimero_sdk::env::ext::External
```

MatejVukosav:
Can you update description with usage example?

PR #903, opened 2024-10-31
feat(context-client): generalize implementation

petarjuki7:
### Part of refactoring of the contract clients

This does not compile yet


crates/context/config/src/client/protocol/near.rs
```diff
@@ -26,7 +24,10 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use url::Url;
 
-use super::{Operation, Transport, TransportRequest};
+use super::{private, Near};
```
miraclx: `Near` should be in this module instead

crates/context/config/src/client/protocol/near.rs
```diff
@@ -26,7 +24,10 @@ use serde::{Deserialize, Serialize};
...
+use crate::client::{Operation, Transport, TransportRequest};
+
+impl private::Protocol for Near {}
```
miraclx: `Protocol` no longer needs to be private, we can remove that

crates/context/src/lib.rs
```diff
@@ -73,7 +73,7 @@ impl ContextManager {
         network_client: NetworkClient,
     ) -> EyreResult<Self> {
         let client_config = config.client.clone();
-        let config_client = ContextConfigClient::from_config(&client_config);
+        let config_client = Client::from_config(&client_config);
```
miraclx: alias this to `ExternalClient` or something like that

from my reorganization comment, this should later be in a module called

`calimero_external::Client`

crates/context/config/src/client/env/config.rs
```diff
@@ -1,7 +1,19 @@
+use std::collections::BTreeMap;
+
+use query::appilcation_revision::{ApplicationRevision, Revision};
```
chefsale: Appilcation?

crates/context/config/src/client/env/config/mutate/context.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub(crate) kind: RequestKind<'a>,
+}
+
+impl<'a> Method<Mutate<'a>> for Near {
+    const METHOD: &'static str = "mutate";
+
+    type Returns = ();
+
+    fn encode(params: &Mutate) -> eyre::Result<Vec<u8>> {
+        let signed = Signed::new(&Request::new(todo!(), params.kind), |b| {
```
petarjuki7: Was not sure what is the signer_id and what is the secret_key, tried to figure out from what parameters are passed to it from src/lib.rs but coudln't figure it out.
Help appreciated.
cc: @miraclx

crates/context/config/src/client/env/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod config;
+// pub mod proxy;
```
xilosada: I removed this to reduce the scope of the PR, `crates/context/config/src/client/env/proxy.rs` should be addressed independently

crates/context/config/src/client/env/proxy.rs
```diff
@@ -0,0 +1,101 @@
+use crate::client::protocol::near::Near;
```
xilosada: I left the file just for future reference

petarjuki7:
Is this fine to be merged now?
cc: @miraclx  @xilosada

miraclx (approved):

PR #907, opened 2024-10-31
Id replace

chefsale:
WIP

### DIFF
--- a/crates/storage-macros/src/lib.rs
+++ b/crates/storage-macros/src/lib.rs
@@ -352,7 +352,7 @@ pub fn atomic_unit_derive(input: TokenStream) -> TokenStream {
             fn calculate_merkle_hash(&self) -> Result<[u8; 32], calimero_storage::interface::StorageError> {
                 use calimero_storage::exports::Digest;
                 let mut hasher = calimero_storage::exports::Sha256::new();
-                hasher.update(self.element().id().as_bytes());
+                hasher.update(self.element().id().bytes);
                 #(
                     hasher.update(
                         &calimero_sdk::borsh::to_vec(&self.#regular_fields)
--- a/crates/storage/src/address.rs
+++ b/crates/storage/src/address.rs
@@ -16,7 +16,6 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use calimero_sdk::serde::{Deserialize, Serialize};
 use fixedstr::Flexstr;
 use thiserror::Error as ThisError;
-use uuid::{Bytes, Uuid};
 
 use crate::env::random_bytes;
 
@@ -35,9 +34,11 @@ use crate::env::random_bytes;
 /// system operation. Abstracting the true type away provides a level of
 /// insulation that is useful for any future changes.
 ///
-#[derive(Copy, Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(BorshSerialize, BorshDeserialize, Copy, Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
 #[serde(crate = "calimero_sdk::serde")]
-pub struct Id(Uuid);
+pub struct Id {
+    pub bytes: [u8; 32],
+}
 
 impl Id {
     /// Creates a new globally-unique identifier.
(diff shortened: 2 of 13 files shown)


crates/storage-macros/src/lib.rs
```diff
@@ -352,7 +352,7 @@ pub fn atomic_unit_derive(input: TokenStream) -> TokenStream {
             fn calculate_merkle_hash(&self) -> Result<[u8; 32], calimero_storage::interface::StorageError> {
                 use calimero_storage::exports::Digest;
                 let mut hasher = calimero_storage::exports::Sha256::new();
-                hasher.update(self.element().id().as_bytes());
+                hasher.update(self.element().id().bytes);
```
miraclx, line 355: use the method instead

crates/storage/src/address.rs
```diff
@@ -35,9 +34,11 @@ use crate::env::random_bytes;
 /// system operation. Abstracting the true type away provides a level of
 /// insulation that is useful for any future changes.
 ///
-#[derive(Copy, Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+#[derive(BorshSerialize, BorshDeserialize, Copy, Clone, Debug, Deserialize, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
 #[serde(crate = "calimero_sdk::serde")]
-pub struct Id(Uuid);
+pub struct Id {
+    pub bytes: [u8; 32],
```
miraclx, line 40: keep it private, or at least impl `Deref` so `*id` works.

but we shouldn't allow direct access to the internal field

crates/storage/src/address.rs
```diff
@@ -50,86 +51,52 @@ impl Id {
...
     }
 }
 
 impl Default for Id {
     fn default() -> Self {
-        Self::new()
-    }
+            Self::new([0; 32])
+        }
 }
```
miraclx, line 77: let's take this out, and keep `Id` generation explicit, either through `new`, `root` or `random`

crates/storage/src/address.rs
```diff
@@ -50,86 +51,52 @@ impl Id {
...
-        Self(Uuid::from_bytes(*bytes))
-    }
-}
-
-impl From<Id> for [u8; 16] {
-    fn from(id: Id) -> Self {
-        *id.0.as_bytes()
+impl From<&[u8; 32]> for Id {
+    fn from(bytes: &[u8; 32]) -> Self {
+        Self { bytes: *bytes }
```
miraclx, line 93:
```suggestion
impl From<[u8; 32]> for Id {
    fn from(bytes: [u8; 32]) -> Self {
        Self::new(bytes)
```

crates/storage/src/address.rs
```diff
@@ -50,86 +51,52 @@ impl Id {
...
     }
 }
 
-impl From<Id> for Uuid {
+impl From<Id> for [u8; 32] {
     fn from(id: Id) -> Self {
-        id.0
+        id.bytes
     }
```
miraclx, line 100: addressing my earlier point, you can do `id.into()` since this is here, but keep `.bytes` private

crates/storage/src/address.rs
```diff
@@ -50,86 +51,52 @@ impl Id {
...
-            IoError::new(IoErrorKind::InvalidData, err)
-        })?))
-    }
-
-    fn deserialize_reader<R: Read>(reader: &mut R) -> Result<Self, IoError> {
-        let mut bytes = [0_u8; 16];
-        reader.read_exact(&mut bytes)?;
-        Ok(Self(Uuid::from_bytes(bytes)))
+    pub fn root() -> Self {
+        Id::new([0; 32])
```
miraclx, line 62:
```suggestion
        Id::new(env::context_id())
```

crates/storage/src/interface.rs
```diff
@@ -944,8 +944,8 @@ impl<S: StorageAdaptor> MainInterface<S> {
     /// If an error occurs when interacting with the storage system, an error
     /// will be returned.
     ///
-    pub fn root<D: Data>(context_id: &[u8; 16]) -> Result<Option<D>, StorageError> {
-        Self::find_by_id(context_id.into())
+    pub fn root<D: Data>() -> Result<Option<D>, StorageError> {
+        Self::find_by_id(env::context_id().into())
```
miraclx, line 948:
```suggestion
        Self::find_by_id(Id::root())
```

crates/storage/src/address.rs
miraclx: in my [later comment](https://github.com/calimero-network/core/pull/907#discussion_r1824465398), `From<Id> for [u8; 32]` is fine, no need for `Deref`

PR #911, opened 2024-10-31
feat: expose proposal api to calimero sdk

MatejVukosav:
Exposed wasm `create_proposal` function which creates proposal in proxy contract


crates/server/src/admin/service.rs
```diff
@@ -125,6 +129,26 @@ pub(crate) fn setup(
...
+            get(get_proposal_approvers_handler),
+        )
+        .route(
+            "/contexts/:context_id/proposals/count",
+            get(get_number_of_active_proposals_handler),
+        )
+        .route(
+            "/contexts/:context_id/proposals",
+            get(get_proposals_handler),
+        )
```
xilosada, line 154: why not just return the size in proposals? If you have all you just use `len` and if they come paginated it should come in the response

xilosada (commented):
:shipit:

crates/server/src/admin/handlers/proposals.rs
```diff
@@ -0,0 +1,250 @@
...
+
+pub async fn get_proposals_handler(
+    session: Session,
+    Extension(state): Extension<Arc<AdminState>>,
+    Json(req): Json<GetProposalsResponse>,
+) -> impl IntoResponse {
+    let sample_action = Action::ExternalFunctionCall(ExternalFunctionCall {
+        receiver_id: get_mock_user(),
+        method_name: "sampleMethod".to_string(),
+        args: "sampleArgs".to_string(),
```
xilosada: why not `Json` instead?

crates/server/src/admin/service.rs
MatejVukosav: pagination will be second iteration

crates/server/src/admin/service.rs
```diff
@@ -19,6 +19,10 @@ use tower_sessions::{MemoryStore, SessionManagerLayer};
 use tracing::info;
 
 use super::handlers::did::delete_did_handler;
+use super::handlers::proposals::{
+    get_number_of_active_proposals_handler, get_number_of_proposal_approvals_handler,
+    get_proposal_approvers_handler, get_proposal_handler, get_proposals_handler,
+};
```
xilosada, line 25: TBH I don't understand how this lint rules works 😓

MatejVukosav:
test

crates/server/src/admin/handlers/proposals.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+pub async fn get_proposals_handler(
+    session: Session,
+    Extension(state): Extension<Arc<AdminState>>,
+    Json(req): Json<GetProposalsResponse>,
+) -> impl IntoResponse {
+    let sample_action = Action::ExternalFunctionCall(ExternalFunctionCall {
+        receiver_id: get_mock_user(),
+        method_name: "sampleMethod".to_owned(),
```
petarjuki7, line 124: I think `to_string()` should be everywhere where you are converting a `&str` to `String`, since it's more clear what you are trying to do, but might be just a nitpick, not sure about performance (cc: @miraclx)

crates/server/src/admin/handlers/proposals.rs
```diff
@@ -0,0 +1,250 @@
...
+        args: serde_json::json!({"example": "value"}),
+        deposit: "100".to_owned(),
+        gas: "10".to_owned(),
+    });
+
+    let proposals = vec![Proposal {
+        id: "proposal_1".to_owned(),
```
petarjuki7, line 131: same here

crates/server/src/admin/handlers/proposals.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+pub async fn get_proposal_handler(
+    Path(proposal_id): Path<String>,
+    session: Session,
+    Extension(state): Extension<Arc<AdminState>>,
+    Json(req): Json<GetProposalResponse>,
+) -> impl IntoResponse {
+    let proposal = Proposal {
+        id: "proposal_1".to_owned(),
```
petarjuki7, line 151: same here... and the others

crates/server/src/admin/handlers/proposals.rs
```diff
@@ -0,0 +1,250 @@
...
+use axum::response::IntoResponse;
+use axum::{Extension, Json};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use tower_sessions::Session;
+
+use crate::admin::service::ApiResponse;
+use crate::AdminState;
+
+//todo split it up into separate files
```
petarjuki7, line 14: Should be split into separate files, and most of the structs into the `primitives` crate

petarjuki7 (changes requested):

crates/server/src/admin/handlers/proposals.rs
MatejVukosav: It was before clippy complaint

crates/server/src/admin/handlers/proposals.rs
MatejVukosav: Thats what todo says 😄

petarjuki7:
I added CLI capability, you can test it with:
`cargo run -p meroctl -- --node-name <node_name> --home data --output-format json proxy get num-proposal-approvals <context_id> <proposal_id>`

Other options are:
`proposal`
`proposals`
`num-active-proposals`
`proposal-approvers`

Also had to make some changes in the handlers for the get APIs.

crates/sdk/src/env/ext.rs
```diff
@@ -126,9 +127,27 @@ impl External {
...
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    Serialize,
+    Deserialize,
+)]
 pub struct ProposalId(pub [u8; 32]);
 
+impl AsRef<[u8]> for ProposalId {
```
MatejVukosav, line 151: AsRef

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn send_proposal_messages(
+        &mut self,
+        // request: SendProposalMessageRequest, I cannot to this?? How to use camelCase?
+        proposal_id: String,
+        message: Message,
+    ) -> Result<bool, Error> {
+        let proposal_id = env::ext::ProposalId(Self::string_to_u8_32(proposal_id.as_str()));
+
+        println!("Send message to proposal: {:?}", proposal_id);
+        let proposal_messages = self.messages.get(&proposal_id).unwrap();
```
MatejVukosav, line 119: missed asref

packages/calimero-sdk/src/rpc/jsonrpc.ts
```diff
@@ -53,33 +51,33 @@ export class JsonRpcClient implements RpcClient {
...
-    params: RpcQueryParams<Args>,
-    config?: RequestConfig,
-  ): Promise<RpcResult<RpcQueryResponse<Output>>> {
-    return await this.request<RpcQueryParams<Args>, RpcQueryResponse<Output>>(
-      'execute',
-      params,
-      config,
-    );
-  }
+  // /**
```
MatejVukosav: jsonrpc

chefsale (approved):

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn approve_proposal(&mut self, _proposal_id: env::ext::ProposalId) -> Result<bool, Error> {
+        println!("Approve proposal: {:?}", _proposal_id);
+        // Self::external()
+        Ok(true)
+    }
+
+    // Messages (discussion)
+    pub fn get_proposal_messages(
+        &self,
+        // request: GetProposalMessagesRequest, I cannot to this??
```
petarjuki7, line 94: Why not?
```suggestion
        request: GetProposalMessagesRequest, //I cannot to this??
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        proposal_id: String,
+    ) -> Result<Vec<Message>, Error> {
+        let proposal_id = env::ext::ProposalId(Self::string_to_u8_32(proposal_id.as_str()));
```
petarjuki7:
```suggestion
        let proposal_id = env::ext::ProposalId(Self::string_to_u8_32(request.proposal_id.as_str()));
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        println!("Approve proposal: {:?}", _proposal_id);
+        // Self::external()
+        Ok(true)
+    }
+
+    // Messages (discussion)
+    pub fn get_proposal_messages(
+        &self,
+        // request: GetProposalMessagesRequest, I cannot to this??
+        proposal_id: String,
```
petarjuki7, line 95:
```suggestion
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+            Some(messages) => Ok(messages.clone()),
+            None => Ok(vec![]),
+        }
+    }
+
+    pub fn send_proposal_messages(
+        &mut self,
+        // request: SendProposalMessageRequest, I cannot to this?? How to use camelCase?
+        proposal_id: String,
+        message: Message,
```
petarjuki7, line 112:
```suggestion
        request: SendProposalMessageRequest, //I cannot to this?? How to use camelCase?
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    ) -> Result<bool, Error> {
+        let proposal_id = env::ext::ProposalId(Self::string_to_u8_32(proposal_id.as_str()));
```
petarjuki7, line 116:
```suggestion
        let proposal_id = env::ext::ProposalId(Self::string_to_u8_32(request.proposal_id.as_str()));
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        println!("Send message to proposal: {:?}", proposal_id);
+        let proposal_messages = self.messages.get(&proposal_id).unwrap();
+        match proposal_messages {
+            Some(mut messages) => {
+                messages.push(message);
```
petarjuki7, line 122:
```suggestion
                messages.push(request.message);
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+                self.messages.insert(proposal_id, messages)?;
+            }
+            None => {
+                let messages = vec![message];
```
petarjuki7, line 126:
```suggestion
                let messages = vec![request.message];
```

petarjuki7 (changes requested):

apps/blockchain/src/lib.rs
MatejVukosav: this does not work

apps/blockchain/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+calimero-sdk = { path = "../../crates/sdk" }
+calimero-storage = { path = "../../crates/storage" }
+serde.workspace = true
```
miraclx: `serde` is reexported through `calimero_sdk` as `calimero_sdk::serde`

apps/blockchain/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[profile.app-release]
+inherits = "release"
+codegen-units = 1
+opt-level = "z"
+lto = true
+debug = false
+panic = "abort"
+overflow-checks = true
```
miraclx, line 21: rust should warn on this, profiles should only be in the root manifest

Cargo.toml
```diff
@@ -32,6 +32,7 @@ members = [
     "./apps/only-peers",
     "./apps/gen-ext",
     "./apps/visited",
+    "./apps/blockchain",
```
miraclx, line 35: can we try a better name maybe?

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
+use calimero_sdk::app;
+use calimero_sdk::borsh::{BorshDeserialize, BorshSerialize};
+use calimero_sdk::env::{self};
+use calimero_sdk::types::Error;
+use calimero_storage::collections::UnorderedMap;
+use calimero_storage::entities::Element;
+use calimero_storage::AtomicUnit;
+use serde::{Deserialize, Serialize};
```
miraclx:
```suggestion
use calimero_sdk::serde::{Deserialize, Serialize};
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Clone, Debug, PartialEq, PartialOrd, Deserialize)]
+#[serde(crate = "calimero_sdk::serde", rename_all = "camelCase")]
+pub struct SendProposalMessageRequest {
+    proposal_id: String,
+    message: Message,
+}
+
+#[app::event]
+pub enum Event {
+    ProposalCreated(),
```
miraclx, line 29: this?

```suggestion
    ProposalCreated { id: ProposalId },
```

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn init() -> AppState {
+        AppState {
+            count: 0,
+            storage: Element::root(),
+            messages: UnorderedMap::new().unwrap(),
+        }
+    }
+
+    pub fn create_new_proposal(&mut self, _request: CreateProposalRequest) -> Result<bool, Error> {
+        println!("Create new proposal: {:?}", _request);
```
miraclx: this doesn't work, if you want to log, use `env::log`

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    #[app::init]
+    pub fn init() -> AppState {
+        AppState {
+            count: 0,
+            storage: Element::root(),
+            messages: UnorderedMap::new().unwrap(),
+        }
+    }
+
+    pub fn create_new_proposal(&mut self, _request: CreateProposalRequest) -> Result<bool, Error> {
```
miraclx: wanna try returning `ProposalId` instead?

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        println!("Approve proposal: {:?}", proposal_id);
+        let _ = Self::external().approve().approve(proposal_id);
+        Ok(true)
+    }
+
+    // Messages (discussion)
+    pub fn get_proposal_messages(
+        &self,
+        // request: GetProposalMessagesRequest, I cannot to this??
+        proposal_id: String,
```
miraclx, line 95: why is it a struct?

apps/blockchain/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+            }
+            None => {
+                let messages = vec![message];
+                self.messages.insert(proposal_id, messages)?;
+            }
+        }
+        Ok(true)
+    }
+
+    fn string_to_u8_32(s: &str) -> [u8; 32] {
```
miraclx, line 134: there's no guarantee a proposal Id will be safely encodable as utf8, use bs58 instead

crates/context/config/src/client/env/proxy/mutate/methods.rs
```diff
@@ -24,15 +24,15 @@ impl<'a, T> ContextProxyMutate<'a, T> {
 
     pub fn approve(
         self,
-        signer_id: Repr<SignerId>,
+        signer_id: SignerId,
```
miraclx, line 27: 👍🏽

crates/context/src/lib.rs
```diff
@@ -1009,4 +1011,73 @@ impl ContextManager {
...
+        let Some(context_config) = handle.get(&ContextConfigKey::new(context_id))? else {
+            //handle
+            return Ok(());
+        };
+
+        //How to get signer id?
+        let signer_id = SignerId::from_bytes(|buffer: &mut [u8; 32]| {
+            *buffer = [0; 32];
+            Ok(32)
+        })?;
```
```suggestion
        let signer_id = signer_id.rt()?;
```

you can read `signing_key` from the `ContextIdentity` record in store

for example, see how this is done when inviting members to a context

crates/node/src/lib.rs
```diff
@@ -35,6 +35,7 @@ use camino::Utf8PathBuf;
 use eyre::{bail, eyre, Result as EyreResult};
 use libp2p::gossipsub::{IdentTopic, Message, TopicHash};
 use libp2p::identity::Keypair;
+use libp2p::relay::client::new;
```
miraclx: ?

crates/node/src/lib.rs
```diff
@@ -427,6 +428,25 @@ impl Node {
...
         };
 
+        for proposal in &outcome.proposals {
+            let _ = self
+                .ctx_manager
+                .propose(
+                    context_id,
+                    executor_public_key,
+                    proposal.0.clone(),
+                    proposal.1.clone(),
```
```suggestion
        for (proposal_id, actions) in &outcome.proposals {
            let _ = self
                .ctx_manager
                .propose(
                    context_id,
                    executor_public_key,
                    proposal_id,
                    actions,
```

crates/context/src/lib.rs
```diff
@@ -1009,4 +1011,73 @@ impl ContextManager {
     pub fn is_application_blob_installed(&self, blob_id: BlobId) -> EyreResult<bool> {
         self.blob_manager.has(blob_id)
     }
+
+    pub async fn propose(
+        &self,
+        context_id: ContextId,
+        signing_key: PublicKey,
+        proposal_id: ProposalId,
+        actions: Vec<u8>,
```
miraclx: since it's going to be serialized, take a `&[u8]` here

crates/sdk/src/env.rs
```diff
@@ -188,6 +188,14 @@ pub fn random_bytes(buf: &mut [u8]) {
...
 }
 
+#[inline]
+pub fn send_proposal(value: &[u8], buf: &mut [u8]) {
+    unsafe { sys::send_proposal(Buffer::new(value), BufferMut::new(buf)) }
+}
+#[inline]
+pub fn approve_proposal(buf: &mut [u8]) {
+    unsafe { sys::approve_proposal(BufferMut::new(buf)) }
+}
```
miraclx: no need for these

crates/sdk/src/sys.rs
```diff
@@ -64,3 +65,5 @@ macro_rules! wasm_imports {
 }
 
 use wasm_imports;
+
+use crate::env::ext::ProposalId;
```
miraclx: you shouldn't need these either

crates/sdk/src/env/ext.rs
```diff
@@ -123,12 +133,34 @@ impl External {
     pub const fn propose(self) -> DraftProposal {
         DraftProposal::new()
     }
+
+    pub const fn approve(self) -> DraftProposal {
+        DraftProposal::new()
```
miraclx:
```suggestion
    pub const fn approve(self, proposal_id: ProposalId) {
        unsafe { sys::approve_proposal(BufferMut::new(&proposal_id)) }
```

crates/sdk/src/env/ext.rs
```diff
@@ -111,6 +114,13 @@ impl DraftProposal {
 
         ProposalId(buf)
     }
+
+    /// Finalise the proposal and send it to the blockchain.
+    #[must_use]
+    pub fn approve(self, proposal_id: ProposalId) -> ProposalId {
+        unsafe { sys::approve_proposal(BufferMut::new(&proposal_id)) }
+        proposal_id
+    }
```
miraclx: `DraftProposal::approve` is implied on `send`, no need for this here

crates/server/src/admin/handlers/proposals.rs
miraclx: `str::to_string` [calls](https://github.com/rust-lang/rust/blob/27e38f8fc7efc57b75e9a763d7a0ee44822cd5f7/library/alloc/src/string.rs#L2795) `String::from(&str)` which [calls](https://github.com/rust-lang/rust/blob/27e38f8fc7efc57b75e9a763d7a0ee44822cd5f7/library/alloc/src/string.rs#L2880) `str::to_owned`, all of which are inlined, so in most cases, it should compile to the same code.

I personally prefer just calling `.to_owned()`

Cargo.toml
MatejVukosav: that will be in separate repo. here just for testing

crates/context/src/lib.rs
```diff
@@ -1009,4 +1011,73 @@ impl ContextManager {
...
+    ) -> EyreResult<()> {
+        let handle = self.store.handle();
+
+        let actions: Vec<ProposalAction> = vec![]; // - How to deserialize actions?
```
MatejVukosav, line 1024: q

apps/blockchain/src/lib.rs
MatejVukosav: To match client request as it is cleaner to change request objects instead of method signature.

crates/sdk/src/env.rs
MatejVukosav: why?

apps/blockchain/Cargo.toml
frdomovic: Removed

apps/blockchain/src/lib.rs
frdomovic: Updated

crates/sdk/src/env.rs
frdomovic: removed

crates/sdk/src/env/ext.rs
frdomovic: Removed

crates/node/src/lib.rs
```diff
@@ -427,10 +429,18 @@ impl Node {
...
             });
         };
+        let args_str = str::from_utf8(&payload).map_err(|e| {
+            error!(%e, "Failed to convert payload to UTF-8.");
+            CallError::InternalError
+        })?;
+        let args_json: ProposalRequest = serde_json::from_str(args_str).map_err(|e| {
+            error!(%e, "Failed to parse proposal request.");
+            CallError::InternalError
+        })?;
```
miraclx: this should be from `outcome.proposals` which is the outcome of the execution, not `payload` (which is the argument to the function call)

`outcome.proposals` is a `Map<Id, Vec<u8>>`, the `proposal` payload should be deserialized `Vec<ProposalAction>`, and a `Proposal` can be constructed.

a similar path for `outcome.approvals`

miraclx: ah, this is trying to determine the `sender` and `receiver`

`sender` is `executor_public_key`, and `receiver` is determined by the action construction inside the app.

~all that needs to happen here is proposal construction (wrapping already constructed actions)~

I see that `ctx_manager.propose` takes `Vec<Action>`, so this is further simplified

crates/sdk/src/env.rs
miraclx: the `External` API is how developers will interface with these.

what gets passed here is delicate since it'll attempt to be deserialized on the node, so controlling low-level access is ideal

crates/sdk/src/env/ext.rs
```diff
@@ -74,6 +75,7 @@ pub struct DraftProposal {
     /// The actions to be executed by the proposal. One proposal can contain
     /// multiple actions to execute.
     actions: Vec<ProposalAction>,
+    approval: Option<ProposalId>,
```
miraclx, line 78: this is needed here

apps/blockchain/src/lib.rs
MatejVukosav: app will be extracted so tbd

crates/context/config/src/client/protocol/near.rs
```diff
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::collections::BTreeMap;
-use std::time;
+use std::{time, vec};
```
miraclx, line 3: ?

crates/context/config/src/client/protocol/near.rs
```diff
@@ -272,8 +272,8 @@ impl Network {
             actions: vec![Action::FunctionCall(Box::new(FunctionCallAction {
                 method_name: method,
                 args,
-                gas: 100_000_000_000_000, // 100 TeraGas
-                deposit: 0,
+                gas: 300_000_000_000_000,
+                deposit: 5_000_000_000_000_000_000_000_000,
```
miraclx, line 276: highlighting this to come back to it

crates/context/src/lib.rs
```diff
@@ -1021,4 +1023,77 @@ impl ContextManager {
...
+        };
+
+        let _ = self
+            .config_client
+            .mutate::<ContextProxy>(
+                context_config.protocol.as_ref().into(),
+                context_config.network.as_ref().into(),
+                context_config.contract.as_ref().into(),
+            )
+            .propose(proposal_id, signer_id.rt().unwrap(), actions)
```
miraclx, line 1060:
```suggestion
            .propose(proposal_id, signer_id.rt().expect("infallible conversion"), actions)
```

while we want to avoid panic points, any thing that is "safe" to use unwrap, must define why that's the case

crates/context/src/lib.rs
```diff
@@ -1021,4 +1023,77 @@ impl ContextManager {
...
+        };
+
+        let _ = self
+            .config_client
+            .mutate::<ContextProxy>(
+                context_config.protocol.as_ref().into(),
+                context_config.network.as_ref().into(),
+                context_config.contract.as_ref().into(),
+            )
+            .approve(signer_id.rt().unwrap(), proposal_id)
```
miraclx, line 1093: same here

crates/meroctl/src/cli.rs
```diff
@@ -49,6 +51,8 @@ pub enum SubCommands {
     App(AppCommand),
     Context(ContextCommand),
     Identity(IdentityCommand),
+    JsonRpc(CallCommand),
```
miraclx, line 54: bad merge? this reintroduces `JsonRpc`

crates/node/Cargo.toml
miraclx, line 1: nit: order

crates/node/src/lib.rs
```diff
@@ -6,13 +6,15 @@
 
 use core::future::{pending, Future};
 use core::pin::Pin;
+use core::str;
```
miraclx, line 9: ?

crates/node/src/lib.rs
```diff
@@ -424,7 +426,7 @@ impl Node {
         }
 
         let outcome_option = self
-            .execute(&mut context, method, payload, executor_public_key)
+            .execute(&mut context, method, payload.clone(), executor_public_key)
```
miraclx, line 429: this shouldn't be `clone`d

crates/node/src/lib.rs
```diff
@@ -436,6 +438,33 @@ impl Node {
...
+            drop(
+                self.ctx_manager
+                    .propose(
+                        context_id,
+                        executor_public_key,
+                        proposal_id.clone(),
+                        actions,
+                    )
+                    .await,
+            );
```
miraclx, line 458: handle the result instead of `drop`ping.

crates/node/src/lib.rs
```diff
@@ -436,6 +438,33 @@ impl Node {
...
+        }
+
+        for approval in &outcome.approvals {
+            drop(
+                self.ctx_manager
+                    .approve(context_id, executor_public_key, approval.clone())
+                    .await,
+            );
```
miraclx, line 466: same here

crates/node/src/types.rs
```diff
@@ -53,3 +54,9 @@ pub enum MessagePayload<'a> {
     StateSync { artifact: Cow<'a, [u8]> },
     BlobShare { chunk: Cow<'a, [u8]> },
 }
+
+#[derive(Deserialize)]
+pub struct ProposalRequest {
+    pub sender: String,
+    pub receiver: String,
+}
```
miraclx, line 62: now unused?

crates/sdk/src/env/ext.rs
```diff
@@ -35,7 +36,7 @@ pub enum ProposalAction {
         receiver_id: AccountId,
 
         /// The amount of tokens to transfer.
-        amount: u64,
+        amount: u128,
```
miraclx, line 39: deposit in function call should be `u128` too

crates/sdk/src/env/ext.rs
miraclx: 🤦🏽 not needed

PR #912, opened 2024-10-31
feat: increase the id bytes size and use context_id as root

chefsale:
tests are passing.

### DIFF
--- a/crates/storage/src/address.rs
+++ b/crates/storage/src/address.rs
@@ -16,9 +16,8 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use calimero_sdk::serde::{Deserialize, Serialize};
 use fixedstr::Flexstr;
 use thiserror::Error as ThisError;
-use uuid::{Bytes, Uuid};
 
-use crate::env::random_bytes;
+use crate::env::{context_id, random_bytes};
 
 /// Globally-unique identifier for an [`Element`](crate::entities::Element).
 ///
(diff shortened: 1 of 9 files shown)


crates/storage/src/address.rs
```diff
@@ -35,101 +34,95 @@ use crate::env::random_bytes;
...
 impl Display for Id {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.0)
-    }
-}
-
-impl From<[u8; 16]> for Id {
-    fn from(bytes: [u8; 16]) -> Self {
-        Self(Uuid::from_bytes(bytes))
+        write!(f, "{:?}", self.bytes)
```
miraclx: ideally this should be `bs58`, but we can skip the extra dep, in the meantime, defer to `Id: Debug`, not `[u8; 32]: Debug`

```suggestion
        write!(f, "{:?}", self)
```

miraclx (approved):

crates/storage/src/address.rs
```diff
@@ -35,101 +34,95 @@ use crate::env::random_bytes;
...
+    fn from(bytes: [u8; 32]) -> Self {
+        Self::new(bytes)
     }
 }
 
-impl From<Id> for Uuid {
+impl From<Id> for [u8; 32] {
     fn from(id: Id) -> Self {
-        id.0
+        id.into()
```
miraclx:
```suggestion
        id.bytes
```

PR #914, opened 2024-10-31
feat: introduce basic vector and set collections to storage 

chefsale:
./scripts/test.sh all passing.


crates/storage/src/entities.rs
```diff
@@ -644,17 +644,19 @@ impl Element {
...
+    /// * `id` - The id of the [`Element`] in the hierarchy of the storage.
     ///
     /// # Panics
     ///
     /// This method can technically panic if the system time goes backwards, to
     /// before the Unix epoch, which should never ever happen!
     ///
     #[must_use]
-    pub fn new(path: &Path) -> Self {
+    pub fn new(path: &Path, id: Option<Id>) -> Self {
```
xilosada, line 655: I see we are extensively using id = `None`, would it be better to create another constructor?

crates/runtime/examples/rps.rs
```diff
@@ -96,15 +96,15 @@ fn main() -> EyreResult<()> {
         [0; 32],
     );
     let create_keypair_outcome = run(&file, "create_keypair", cx, &mut storage, &limits)?;
-    dbg!(&create_keypair_outcome);
+    &create_keypair_outcome;
```
xilosada: is this intended? 🤔

chefsale: great clippy or whatever was introduced

chefsale: will revert all the changes not related to my code

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,373 @@
...
+use crate::interface::Interface;
+use crate::{AtomicUnit, Collection};
+
+/// A map collection that stores key-value pairs.
+#[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+#[type_id(255)]
+#[root]
+pub struct UnorderedMap<K, V> {
+    /// The prefix used for the map's entries.
+    prefix: String,
```
miraclx: so this.. should be the ID of the map

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,373 @@
...
+{
+    /// Create a new map collection.
+    ///
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn new(prefix: &str) -> Result<Self, StoreError> {
```
miraclx: we shouldn't need to take in a prefix here, pick a default non-variable dependent prefix

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,373 @@
...
+
+        let _ = Interface::save(&mut this)?;
+
+        Ok(this)
+    }
+
+    /// Compute the ID for a key.
+    fn compute_id(&self, key: &[u8]) -> Id {
+        let mut hasher = Sha256::new();
+        hasher.update(self.prefix.as_bytes());
```
miraclx:
```suggestion
        hasher.update(self.id().as_bytes());
```

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn insert(&mut self, key: K, value: V) -> Result<(), StoreError> {
+        if self.contains(&key)? {
+            let id = self.compute_id(key.as_ref());
+            if let Some(mut entry) = Interface::find_by_id::<Entry<K, V>>(id)? {
```
miraclx: `self.contains(..)` calls `compute_id` and `find_by_id`, and we're doing it again here, let's flatten this out.

we can introduce a private `get_raw` method that returns `Option<Entry>`

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn remove<Q>(&mut self, key: &Q) -> Result<bool, StoreError>
+    where
+        K: Borrow<Q>,
+        Q: PartialEq + AsRef<[u8]> + ?Sized,
+    {
+        let entry = Element::new(&self.path(), Some(self.compute_id(key.as_ref())));
```
miraclx, line 210: you shouldn't need `entries` here

crates/storage/src/entities.rs
chefsale: Its legacy code from Dan which has to be removed, but in next PRs.

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// Helper to associate the generic types with the collection.
+    _priv: PhantomData<V>,
+}
+
+/// An entry in a vector.
+#[derive(AtomicUnit, Clone, Debug, Eq, PartialEq, PartialOrd)]
+#[type_id(250)]
+pub struct Entry<V> {
+    /// The index for the entry.
+    index: usize,
```
miraclx: we can skip `index` here, rely on the CRDT system for it's determination

crates/runtime/examples/rps.rs
chefsale: reverted

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,329 @@
...
+    ///
+    pub fn pop(&mut self) -> Result<Option<V>, StoreError> {
+        let len = self.len()?;
+        if len == 0 {
+            return Ok(None);
+        }
+
+        let index = len - 1;
+        let id = self.compute_id(index);
+        let value = self.get(index)?;
```
miraclx: `child_info_for` returns a list of `Id`s, we can get the ID of that last item from there

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    /// Compute the ID for a element in the vector based on the prefix and index.
+    fn compute_id(&self, index: usize) -> Id {
+        let mut hasher = Sha256::new();
+        hasher.update(self.prefix.as_bytes());
+        hasher.update(index.to_le_bytes());
+        Id::new(hasher.finalize().into())
+    }
+
```
miraclx: we shouldn't need compute ID for vec at all

crates/storage/src/entities.rs
miraclx, line 655: this should always be specified, the caller should call `Id::random()`

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,373 @@
...
+
+    /// Insert a key-value pair into the map.
+    ///
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn insert(&mut self, key: K, value: V) -> Result<(), StoreError> {
```
xilosada: Shouldn't this return a bool? 
https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert
https://docs.rs/near-sdk/latest/near_sdk/collections/struct.UnorderedSet.html#method.insert

crates/storage/src/entities.rs
chefsale: will remove path and make it specified in a later PR

crates/storage/src/collections/unordered_map.rs
chefsale: will change it.

crates/storage/src/collections/unordered_map.rs
chefsale: not sure what do you mean here?

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,401 @@
...
+        Ok(this)
+    }
+
+    /// Compute the ID for a key.
+    fn compute_id(&self, key: &[u8]) -> Id {
+        let mut hasher = Sha256::new();
+        hasher.update(self.id.as_bytes());
+        hasher.update(key);
+        Id::new(hasher.finalize().into())
+    }
```
miraclx, line 80: how about you move this into the `collections` module, so both `unordered_{map,set}` can use it instead of duplicating

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,401 @@
...
+    key: K,
+    /// The value for the entry.
+    value: V,
+    /// The storage element for the entry.
+    #[storage]
+    storage: Element,
+}
+
+impl<
+        K: BorshSerialize + BorshDeserialize + AsRef<[u8]> + PartialEq,
```
miraclx: localize these extra bounds to the function that requires them, otherwise they seem like arbitrary requirements

```suggestion
        K: BorshSerialize + BorshDeserialize,
```

crates/storage/src/collections/unordered_map.rs
miraclx: Just saw the linked snippets are for sets, the map should return the pre-existing item

https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert
https://docs.rs/near-sdk/latest/near_sdk/store/lookup_map/struct.LookupMap.html#method.insert

crates/storage/src/collections/unordered_set.rs
```diff
@@ -0,0 +1,267 @@
...
+        if self.contains(&value)? {
+            return Ok(false);
+        } else {
+            let storage = Element::new(&path, Some(self.compute_id(value.as_ref())));
+            let _ = Interface::add_child_to(
+                self.storage.id(),
+                &mut self.entries,
+                &mut Entry { value, storage },
+            )?;
+        }
```
miraclx: early return helps save on indentation

```suggestion
            return Ok(false);
        }

        let storage = Element::new(&path, Some(self.compute_id(value.as_ref())));
        let _ = Interface::add_child_to(
            self.storage.id(),
            &mut self.entries,
            &mut Entry { value, storage },
        )?;
```

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Add a value to the end of the vector.
+    ///
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn push(&mut self, value: V) -> Result<bool, StoreError> {
```
miraclx: do we need to return a `bool` here?

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    /// Get the raw storage entry at a specific index in the vector.
+    ///
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn get_raw(&self, index: usize) -> Result<Option<Entry<V>>, StoreError> {
```
miraclx:
```suggestion
    fn get_raw(&self, index: usize) -> Result<Option<Entry<V>>, StoreError> {
```

crates/storage/src/collections/unordered_map.rs
```diff
@@ -0,0 +1,401 @@
...
+            let _ = Interface::add_child_to(
+                self.storage.id(),
+                &mut self.entries,
+                &mut Entry {
+                    key,
+                    value,
+                    storage,
+                },
+            )?;
+        }
```
miraclx:
```suggestion
            return Ok(false);
        }

        let path = self.path();
        let storage = Element::new(&path, Some(self.compute_id(key.as_ref())));
        let _ = Interface::add_child_to(
            self.storage.id(),
            &mut self.entries,
            &mut Entry {
                key,
                value,
                storage,
            },
        )?;
```

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn pop(&mut self) -> Result<Option<V>, StoreError> {
+        let id = match Interface::child_info_for(self.id(), &self.entries)?
+            .into_iter()
+            .last()
```
miraclx: this will be `O(N)` even though the length is known. `Vec::pop` is more appropriate here.

crates/storage/src/collections/vector.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system, or a child
+    /// [`Element`](crate::entities::Element) cannot be found, an error will be
+    /// returned.
+    ///
+    pub fn get_raw(&self, index: usize) -> Result<Option<Entry<V>>, StoreError> {
+        let id = match Interface::child_info_for(self.id(), &self.entries)?
+            .into_iter()
+            .nth(index)
```
miraclx:
```suggestion
        let id = match Interface::child_info_for(self.id(), &self.entries)?
            .get(index)
```

crates/storage/src/collections/unordered_map.rs
chefsale: Will move into common.rs in new PR

crates/storage/src/collections/vector.rs
chefsale: Implemented to be same as rust std collection

miraclx: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push

PR #917, opened 2024-11-01
remove warnings from clippy

chefsale:


.github/actions/style/action.yml
```diff
@@ -41,7 +41,7 @@ runs:
             echo "Rust code is not properly formatted."
             exit 1  # Exit with a non-zero status code if formatting fails
           fi
-          if ! cargo clippy -- -D warnings; then
+          if ! cargo clippy; then
```
miraclx: so they don't clutter CI

```suggestion
          if ! cargo clippy -- -A warnings; then
```

xilosada (approved):

PR #918, opened 2024-11-01
feat: implement context invite & identity new command

fbozic:
- implement missing commands in the meroctl for the e2e tests
  -  context invite
  - identity generate
- cleanup up server admin primitives
  - force `camelCase` on all request/response types 
  - expose `new` method only on Response types, not ResponseData types
---
I'm pretty certain this will break some node-ui since I've enforced `camelCase` on all request/response types

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,7 +34,7 @@ dependencies = [
  "pin-project-lite",
  "smallvec",
  "tokio",
- "tokio-util 0.7.12",
+ "tokio-util 0.7.11",
 ]
 
 [[package]]
@@ -44,7 +44,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.86",
+ "syn 2.0.72",
 ]
 
 [[package]]
@@ -65,7 +65,7 @@ checksum = "b6ac1e58cded18cb28ddc17143c4dea5345b3ad575e14f32f66e4054a56eb271"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.86",
+ "syn 2.0.72",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


chefsale (approved):
🚢

Cargo.lock
miraclx, line 1: weird.. no dependency changes, but we've got a massive bump here

did you `cargo update`?

crates/server-primitives/src/admin.rs
```diff
@@ -286,116 +368,78 @@ impl JoinContextRequest {
...
+
+// -------------------------------------------- Identity API ----------------------------------------
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
-#[non_exhaustive]
-pub struct GetContextStorageResponseData {
-    pub size_in_bytes: u64,
+pub struct GenerateContextIdentityResponseData {
+    pub public_key: PublicKey,
+    pub private_key: PrivateKey,
```
miraclx, line 423: we really shouldn't expose private keys at all through the server, I thought we could pre-save unassociated private keys to some record, say `ContextIdentity { context_id: [0;32] }` so when the invitation payload comes back, we'll have the private key, but that's not implemented yet

Cargo.lock
fbozic: nope, there were some merge conflicts so I've deleted the lock and let cargo recreate it. let me see if I can revert.

fbozic: reverted, c/p from master branch

crates/server-primitives/src/admin.rs
fbozic: follow up issue for this: https://github.com/calimero-network/core/issues/919

MatejVukosav:
Did u test admin dashboard after changes?

chefsale:
Needs fixing in node-ui and sdk

https://github.com/calimero-network/core/blob/master/packages/calimero-sdk/src/api/nodeApi.ts
https://github.com/calimero-network/core/blob/master/node-ui/src/api/dataSource/NodeDataSource.ts

PR #921, opened 2024-11-01
feat: added a simple app with nested map and set

chefsale:
Added a simple showcase app of set and map working together being nested. 

Wasm binary is building.

### DIFF
--- a/apps/visited/src/lib.rs
+++ b/apps/visited/src/lib.rs
@@ -1,0 +1,52 @@
+#![allow(clippy::len_without_is_empty)]
+
+use std::result::Result;
+
+use calimero_sdk::app;
+use calimero_sdk::types::Error;
+use calimero_storage::collections::{UnorderedMap, UnorderedSet};
+use calimero_storage::entities::Element;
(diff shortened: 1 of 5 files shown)


apps/visited/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[app::init]
+    pub fn init() -> VisitedCities {
+        VisitedCities {
+            visited: UnorderedMap::new().unwrap(),
+            storage: Element::root(),
+        }
+    }
+
+    pub fn add_person(&mut self, person: String) -> Result<bool, Error> {
+        Ok(self.visited.insert(person, UnorderedSet::new().unwrap())?)
```
miraclx, line 32: avoid `unwrap` everywhere but `init`, which has technical limitations for why we can't return a result

on perhaps we make it so `UnorderedMap::new` doesn't return an error, and panics internally if it can't be created

chefsale: will fix in follow up PR this is just an example to get master int o good state

PR #927, opened 2024-11-03
refactor: extract proxy entities

xilosada:
Expose proxy contract entities in the context-config
Refactor proxy contract to use a similar deserialization than config contract (using a mutate pattern)


crates/context/config/src/client/env/proxy/query/proposal.rs
```diff
@@ -0,0 +1,42 @@
+use serde::Serialize;
```
xilosada: this fileis not part of this PR... I'll remove it

crates/context/config/src/lib.rs
```diff
@@ -101,6 +102,78 @@ pub enum ContextRequestKind<'a> {
...
+}
+
+#[derive(PartialEq, Serialize, Deserialize, Copy, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+#[expect(clippy::exhaustive_enums, reason = "Considered to be exhaustive")]
+pub struct ProposalWithApprovals {
+    pub proposal_id: ProposalId,
+    pub num_approvals: usize,
+}
+
```
xilosada: I'll move the types below the SystemRequest

chefsale (approved):

Issue #265, opened 2024-05-09
Tracking issue for CI testing


miraclx:
Items (in no particular order):

- [ ] Rust - 3d
	- [x] `cargo test` all crates that have unit tests
	- [x] `cargo fmt` to ensure styling consistency
	- [x] `cargo clippy` to lint
	- ^~~ fundamental pieces (tracked by https://github.com/calimero-network/core/issues/586)
	- [ ] `cargo check` to ensure the `HEAD` builds right
		- [ ] We need to walk the dep tree, creating a dependency graph and build each crate in order of dependence so we can guarantee all crates build independently as well as when linked. (See https://github.com/calimero-network/core/pull/258#discussion_r1593846781)
		- [ ] Don't forget to build examples
	- [ ] `cargo udeps` /& `cargo machete`  to flag unused deps in each crate's manifest
		- [ ] Perhaps even instructing `rustc` to warn on `unused-crate-dependencies`  by default (`.cargo/config.toml`) and denying warnings in CI so they're hard errors.
		- The key concern here will be false positives due to feature flags or any dependencies of codegen.
	- [x] `cargo deny` to govern dependency deduplication & licenses
	- [ ] cleanup all the `cargo deny` deduplication & licenses warnings and issues which make sense
	- [ ] `cargo audit` to flag security vulnerabilities
	- [ ] Eventually, some testing code coverage
- [ ] Research - 1d
	- [ ] Some way of guaranteeing primitives are lightweight 🤷🏽 (estimate is for research only)
	- [ ] Code compliance enforcer with a tool like or comparable to [themis](https://github.com/near/nearcore/blob/9932e95c5c49262189bc4737366ce46bb2b953c8/tools/themis/src/rules.rs) (estimate is for research only)

PR #928, opened 2024-11-03
chore: run only rust tests affected with changed files

MatejVukosav:
Tested with changed files
https://github.com/calimero-network/core/actions/runs/11698775125 

`./scripts/test.sh now accepts --all or --local {changed_files}`

Docs https://github.com/calimero-network/calimero-network.github.io/pull/48


chefsale (approved):
LGTM

scripts/test.sh
```diff
@@ -5,13 +5,161 @@ set -ex
...
+            duplicate=true
+            break
+        fi
+    done
+    # If item is not a duplicate, add it to unique_array
+    if ! $duplicate; then
+        unique_changed_crates+=("$item")
+    fi
+done
+echo "Unique crates from changed files" "${unique_changed_crates[@]}"
```
miraclx, line 93: `uniq` command?

scripts/test.sh
```diff
@@ -5,13 +5,161 @@ set -ex
...
+    for dep in "${dependencies[@]}"; do
+        # Compare dependency with list of crates from workspace to skip external dependencies
+        calimero_dep_crate_path="${dep/calimero-/crates/}"
+        for crate in "${all_crates[@]}"; do
+            if [[ "$crate" == "$calimero_dep_crate_path" ]]; then
+                echo "Found matching dependency: $calimero_dep_crate_path"
+                dep_arr+=("$calimero_dep_crate_path")
+            fi
+        done
+    done
```
miraclx, line 117: we shouldn't need to test dependencies of changed crates, but "dependents" of changed crates.

`calimero-node` depends on `calimero-context` which depends on `calimero-primitives`, if `calimero-context` changes we need to test that as well as `calimero-node`, but not necessarily `calimero-primitives`

scripts/test.sh
```diff
@@ -5,13 +5,161 @@ set -ex
...
+
+# Run tests for each module and its dependencies
+echo "Running tests for affected modules and their dependencies..."
+echo "Crates to run tests:" "${crates_to_test[@]}"
+for crate in "${crates_to_test[@]}"; do
+    # Convert from folder name to package name
+    crate_package_name=("${crate/crates\//calimero-}")
+    if [[  "$crate_package_name" == "calimero-merod" ]]; then
+        echo "Testing crate merod"
+        cargo +nightly test -p "merod"
```
miraclx, line 157: why test with nightly?

miraclx (commented):
A lot of the crates don't compile on their own, for example `calimero-server`, `calimero-node-primitives`, `calimero-config` etc, relying instead for `merod` & `meroctl` to maximally define all the minimum feature requirements to satisfy all the crates.

Not an ideal situation, but quite the effort to untangle. It's tracked by #265 under

> 	- [ ] `cargo check` to ensure the `HEAD` builds right
>		- [ ] We need to walk the dep tree, creating a dependency graph and build each crate in order of dependence so we can guarantee all crates build independently as well as when linked. (See https://github.com/calimero-network/core/pull/258#discussion_r1593846781)
>		- [ ] Don't forget to build examples

This PR attempts to test each crate independently, it will break in cases where `cargo build` for a specific crate doesn't work.

if mode is not "all", this will not test apps and contracts? they don't have `calimero-` prefix, nor are they in `crates/*` folder.

We should revert this.

scripts/test.sh
MatejVukosav: It tests dependencies which imports changed crate

PR #929, opened 2024-11-03
feat: proxy client impl

xilosada:
First iteration for the Proxy Client


crates/context/config/src/client/env/proxy/mutate.rs
```diff
@@ -1,8 +1,86 @@
...
+pub struct ContextProxyMutateRequest<'a, T> {
     client: CallClient<'a, T>,
+    raw_request: ProxyMutateRequest,
+}
+
+#[derive(Debug)]
+struct Mutate<'a> {
+    pub(crate) signing_key: [u8; 32],
+    pub(crate) raw_request: ProxyMutateRequest,
+    _marker: PhantomData<&'a ()>,
```
xilosada: @miraclx I'll need some guidance here. Not sure what is the best way to handle the lifetime

miraclx: you don't need one here

crates/context/config/src/client/env/proxy/mutate.rs
```diff
@@ -1,8 +1,86 @@
...
+}
+
+impl<'a> Method<Near> for Mutate<'a> {
```
petarjuki7: I don't think you need lifetimes here.

crates/context/config/src/client/env/proxy/mutate.rs
```diff
@@ -1,8 +1,86 @@
...
+
+        Ok(encoded)
+    }
+
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        serde_json::from_slice(&response).map_err(Into::into)
+    }
+}
+
+impl<'a> Method<Starknet> for Mutate<'a> {
```
petarjuki7: And here

crates/context/config/src/client/env/proxy/mutate.rs
```diff
@@ -1,8 +1,86 @@
...
+}
+
+impl<'a, T: Transport> ContextProxyMutateRequest<'a, T> {
+    pub async fn send(
+        self,
+        signing_key: [u8; 32],
+    ) -> Result<Option<ProposalWithApprovals>, ClientError<T>> {
+        let request = Mutate {
+            signing_key,
+            _marker: PhantomData,
```
petarjuki7: No need for a a `_marker`

petarjuki7 (dismissed):

crates/context/config/src/client/env/proxy/mutate.rs
xilosada: Removed

crates/context/config/src/client/env/proxy.rs
```diff
@@ -7,17 +7,17 @@ use mutate::ContextProxyMutate;
 use query::ContextProxyQuery;
 
 #[derive(Copy, Clone, Debug)]
-pub enum ContextProxy {}
+pub enum ContextConfig {}
```
miraclx, line 10: this is the wrong change. it should be `ContextProxy`

PR #944, opened 2024-11-06
feat(node): add proxy contract node logic

frdomovic:


crates/context/src/lib.rs
```diff
@@ -1050,12 +1050,22 @@ impl ContextManager {
...
 
+        let proxy_contract: String = self
+            .config_client
+            .query::<ContextConfigEnv>(
+                context_config.protocol.as_ref().into(),
+                context_config.network.as_ref().into(),
+                context_config.contract.as_ref().into(),
+            )
+            .get_proxy_contract(context_id.rt().unwrap())
+            .await?;
```
miraclx: we should read this from the contract only once for the lifetime of the context, and store it in the context config of the local store

same as we do with `protocol`, `network`, `config_contract`, etc..

frdomovic: -> Now in process of creating context / joining context it fetches proxy contract id , saves it and this are only times this is being called. 

In propose functions now it takes the argument from config

crates/context/src/lib.rs
```diff
@@ -183,18 +196,6 @@ impl ContextManager {
...
-            identity_secret,
-            Some(ContextConfigParams {
-                protocol: self.client_config.new.protocol.as_str().into(),
-                network_id: self.client_config.new.network.as_str().into(),
-                contract_id: self.client_config.new.contract_id.as_str().into(),
-                application_revision: 0,
-                members_revision: 0,
-            }),
-        )?;
-
```
miraclx: when calling `init`, this "must've" been made available. so this can't be moved.

at this point, you can just make it empty.

crates/context/src/lib.rs
```diff
@@ -332,10 +352,13 @@ impl ContextManager {
 
         let context_exists = handle.has(&ContextMetaKey::new(context_id))?;
 
+        let proxy_contract = self.get_proxy_contract(context_id).await;
+
         let mut config = (!context_exists).then(|| ContextConfigParams {
             protocol: protocol.into(),
             network_id: network_id.into(),
             contract_id: contract_id.into(),
+            proxy_contract: proxy_contract.unwrap().into(),
```
```suggestion
        let proxy_contract = self.get_proxy_contract(context_id).await?;

        let mut config = (!context_exists).then(|| ContextConfigParams {
            protocol: protocol.into(),
            network_id: network_id.into(),
            contract_id: contract_id.into(),
            proxy_contract: proxy_contract.into(),
```

crates/context/src/lib.rs
```diff
@@ -244,6 +245,24 @@ impl ContextManager {
 
         let this = self.clone();
         let context_id = context.id;
+        let _ignored = tokio::spawn(async move {
+            let proxy_contract = this.get_proxy_contract(context_id).await;
```
miraclx: when creating a context, this will always fail, because the context doesn't yet exist on the context config

there's no need for a separate async task here, just call it after `add_context` in the `finalizer` task

crates/context/src/lib.rs
```diff
@@ -332,10 +352,13 @@ impl ContextManager {
 
         let context_exists = handle.has(&ContextMetaKey::new(context_id))?;
 
+        let proxy_contract = self.get_proxy_contract(context_id).await;
+
         let mut config = (!context_exists).then(|| ContextConfigParams {
```
miraclx: I'd move this into the block, since we only need to call it when the context doesn't exist

this will require a bit of a refactor

miraclx (commented):
I'd move th

crates/context/src/lib.rs
```diff
@@ -323,22 +339,26 @@ impl ContextManager {
...
+            Some(ContextConfigParams {
+                protocol: protocol.into(),
+                network_id: network_id.into(),
+                contract_id: contract_id.into(),
+                proxy_contract: proxy_contract.into(),
+                application_revision: 0,
+                members_revision: 0,
+            })
+        } else {
+            //should this be existing config?
```
MatejVukosav: q

miraclx: if the context already exists, the config already exists as well.

crates/context/src/lib.rs
```diff
@@ -323,22 +339,26 @@ impl ContextManager {
...
 
         let handle = self.store.handle();
-
         let identity_key = ContextIdentityKey::new(context_id, invitee_id);
-
         if handle.has(&identity_key)? {
             return Ok(None);
         }
 
+        let proxy_contract = self.get_proxy_contract(context_id).await?;
```
miraclx: move this to the `!context_exists` branch

crates/context/src/lib.rs
```diff
@@ -323,22 +339,26 @@ impl ContextManager {
...
+                protocol: protocol.into(),
+                network_id: network_id.into(),
+                contract_id: contract_id.into(),
+                proxy_contract: proxy_contract.into(),
+                application_revision: 0,
+                members_revision: 0,
+            })
+        } else {
+            //should this be existing config?
+            return Ok(None);
```
```suggestion
            None
```

crates/context/src/lib.rs
```diff
@@ -239,11 +240,25 @@ impl ContextManager {
                 .send(*context_secret)
                 .await?;
 
+            let proxy_contract = this.get_proxy_contract(context.id).await?;
+            drop(this.add_context(
```
miraclx: why drop the result?

miraclx: also, as mentioned earlier, state initialization (which is above) requires `add_context` to have been called

so restore [this](https://github.com/calimero-network/core/pull/944/files#r1832510769)

and here, instead of `add_context`, fetch the context config params and just update the proxy contract.

crates/context/src/lib.rs
```diff
@@ -323,22 +339,25 @@ impl ContextManager {
         }
 
         let handle = self.store.handle();
-
         let identity_key = ContextIdentityKey::new(context_id, invitee_id);
-
         if handle.has(&identity_key)? {
```
miraclx: these are complex pieces, let's not impair readability

frdomovic: Reverted

crates/context/src/lib.rs
frdomovic: Reverted + proxy_contract empty string

crates/node/src/lib.rs
```diff
@@ -438,14 +438,12 @@ impl Node {
...
-            };
-            let actions = vec![action];
 
+        for (proposal_id, actions) in &outcome.proposals {
+            let actions: Vec<ProposalAction> = from_slice(&actions).map_err(|e| {
+                error!(%e, "Failed to deserialize proposal actions.");
+                CallError::InternalError
+            })?;
             drop(
                 self.ctx_manager
```
miraclx: why are we dropping? I suspect that's a result? it should be handled

same as below

MatejVukosav: updated

crates/node/src/lib.rs
```diff
@@ -446,24 +446,28 @@ impl Node {
...
+            self.ctx_manager
+                .propose(
+                    context_id,
+                    executor_public_key,
+                    proposal_id.clone(),
+                    actions.clone(),
+                )
+                .await
+                .map_err(|e| {
+                    error!(%e, "Failed to create proposal {:?} with actions {:?} in context {:?} by identity {:?}", proposal_id, actions, context_id, executor_public_key);
```
miraclx: just failed to create proposal, I think is fine (to avoid spamming the console).. the error (from the contract) will be in `e`

crates/node/src/lib.rs
```diff
@@ -446,24 +446,28 @@ impl Node {
...
-            drop(
-                self.ctx_manager
-                    .approve(context_id, executor_public_key, approval.clone())
-                    .await,
-            );
+            self.ctx_manager
+                .approve(context_id, executor_public_key, approval.clone())
+                .await
+                .map_err(|e| {
+                    error!(%e, "Failed to approve proposal {:?}", approval);
```
```suggestion
        for proposal_id in &outcome.approvals {
            self.ctx_manager
                .approve(context_id, executor_public_key, proposal_id)
                .await
                .map_err(|e| {
                    error!(%e, "Failed to approve proposal {:?}", proposal_id);
```

crates/context/src/lib.rs
```diff
@@ -145,6 +145,19 @@ impl ContextManager {
...
 
+    async fn get_proxy_contract(&self, context_id: ContextId) -> EyreResult<String> {
+        let proxy_contract = self
+            .config_client
+            .query::<ContextConfigEnv>(
+                self.client_config.new.protocol.as_str().into(),
+                self.client_config.new.network.as_str().into(),
+                self.client_config.new.contract_id.as_str().into(),
+            )
+            .get_proxy_contract(context_id.rt().unwrap())
```
```suggestion
            .get_proxy_contract(context_id.rt().expect("infallible conversion"))
```

miraclx (approved):
factoring in #953, lgtm

PR #945, opened 2024-11-06
feat(contracts): update sandbox tests and logic

frdomovic:
# feat(contracts): update sandbox tests and logic

## Summary
Added a guard to deploy proxy contract + fix tests


frdomovic:
@alenmestrov is this is outdated please close it

PR #954, opened 2024-11-08
fix: proxy contract selection

MatejVukosav:


crates/context/src/lib.rs
```diff
@@ -1078,14 +1078,19 @@ impl ContextManager {
...
         };
 
+        let proxy_contract = self.get_proxy_contract(context_id).await?;
         let _ = self
             .config_client
             .mutate::<ContextProxy>(
                 context_config.protocol.as_ref().into(),
                 context_config.network.as_ref().into(),
-                context_config.contract.as_ref().into(),
+                proxy_contract.into(),
```
frdomovic:
```suggestion
                context_config.proxy_contract.as_ref().into(), 
```

crates/context/src/lib.rs
```diff
@@ -1078,14 +1078,19 @@ impl ContextManager {
             bail!("No private key found for signer");
         };
 
+        let proxy_contract = self.get_proxy_contract(context_id).await?;
```
frdomovic:
```suggestion
```

crates/context/src/lib.rs
frdomovic: We already have it saved in config we should not refetch this @miraclx

crates/context/src/lib.rs
```diff
@@ -1111,14 +1116,15 @@ impl ContextManager {
...
```
frdomovic:
```suggestion
```

crates/context/src/lib.rs
```diff
@@ -1111,14 +1116,15 @@ impl ContextManager {
...
         let _ = self
             .config_client
             .mutate::<ContextProxy>(
                 context_config.protocol.as_ref().into(),
                 context_config.network.as_ref().into(),
-                context_config.contract.as_ref().into(),
+                proxy_contract.into(),
```
frdomovic:
```suggestion
                 context_config.proxy_contract.as_ref().into(), 
```

frdomovic (approved):

miraclx (approved):

PR #955, opened 2024-11-08
fix:  join context with proxy

MatejVukosav:
Joining context throws

```
 ERROR calimero_node: Failed handling user command: 
   0: transport error: relayer response (500 Internal Server Error): left error: error while querying contract: handler error: [Function call returned an error: wasm execution failed with error: MethodResolveError(MethodNotFound)]

```

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -354,7 +354,10 @@ impl ContextManager {
 
         let context_exists = handle.has(&ContextMetaKey::new(context_id))?;
         let mut config = if !context_exists {
-            let proxy_contract = self.get_proxy_contract(context_id).await?;
+            let proxy_contract = self
+                .get_proxy_contract(context_id)
+                .await
+                .unwrap_or_else(|_| "".to_owned());
             Some(ContextConfigParams {
                 protocol: protocol.into(),
                 network_id: network_id.into(),
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -1083,9 +1083,13 @@ impl ContextManager {
             .mutate::<ContextProxy>(
                 context_config.protocol.as_ref().into(),
                 context_config.network.as_ref().into(),
-                context_config.contract.as_ref().into(),
+                context_config.proxy_contract.as_ref().into(),
             )
-            .propose(proposal_id, signer_id.rt().unwrap(), actions)
+            .propose(
+                proposal_id,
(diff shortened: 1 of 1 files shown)


frdomovic (approved):

crates/context/src/lib.rs
```diff
@@ -354,7 +354,10 @@ impl ContextManager {
 
         let context_exists = handle.has(&ContextMetaKey::new(context_id))?;
         let mut config = if !context_exists {
-            let proxy_contract = self.get_proxy_contract(context_id).await?;
+            let proxy_contract = self
+                .get_proxy_contract(context_id)
+                .await
+                .unwrap_or_else(|_| "".to_owned());
```
miraclx, line 360: if there's an error, we should handle it

MatejVukosav: Error is that it does not exists.

miraclx: if it doesn't exist, then there's a deeper issue, the config contract didn't record the context, and if that's the case, we shouldn't be here at all (context creation should've failed).

Perhaps we need to tweak near tx submission to await cross contract calls, so we can appropriately determine if the whole chain succeeded.

MatejVukosav: It exists on owner side but it does not exists for invited peer

miraclx: > It exists on owner side but it does not exists for invited peer

`get_proxy_contract` calls the contract, and it would only return that error if the context doesn't exist on the config contract

miraclx: > It exists on owner side but it does not exists for invited peer

interesting that there's an invitation payload to begin with, `add_member` should've failed as well if the context doesn't exist.

MatejVukosav: Invited peer joined context but catchup failed.

miraclx: what's the error?

MatejVukosav: in description

miraclx: doesn't explain why ignoring the error is the solution

the `propose` method wasn't found because it was calling the wrong contract, which this PR fixes

but this line, you said was because the context didn't exist, in which case, we shouldn't be ignoring the error but figuring out why that's the case.

but if the error no longer happens, then we don't need this line, and any errors that occur here should be handled.

PR #956, opened 2024-11-08
Context and Proxy Contracts Improvements

alenmestrov:
### Changed logic in the process of creating a subaccount responsible for Proxy Contract:
- Now we are deriving subaccount from Context contract account and because of that we don't need to deposit anything in mutate method when calling Context contract
- Gas needed for deploying the contract is now determined by the size of Proxy contract, Proxy contract init arguments and minimum required funds for generating and seeding accounts instead of being static
- Unused gas is returned to Context contract account
- Optimized tests in Context config to reflect new process

### Fixed return types for callbacks
- Using `#[callback_result]` for callback functions

- Fixed minor issues with code such us double json encoding, duplicates in tests, taking references, async functions declarations

### DIFF
--- a/contracts/context-config/src/mutate.rs
+++ b/contracts/context-config/src/mutate.rs
@@ -8,7 +8,7 @@ use calimero_context_config::{ContextRequest, ContextRequestKind, Request, Reque
 use near_sdk::serde_json::{self, json};
 use near_sdk::store::IterableSet;
 use near_sdk::{
-    env, near, require, AccountId, Gas, NearToken, Promise, PromiseError, PromiseResult,
+    env, near, require, AccountId, Gas, NearToken, Promise, PromiseError, PromiseOrValue,
 };
 
 use super::{
@@ -18,7 +18,6 @@ use super::{
 
 #[near]
 impl ContextConfigs {
-    #[payable]
     pub fn mutate(&mut self) {
         parse_input!(request: Signed<Request<'_>>);
 
@@ -125,9 +124,11 @@ impl ContextConfigs {
             env::panic_str("context already exists");
         }
 
-        // Initiate proxy contract deployment with a callback for when it completes
+        env::log_str(&format!("Context `{}` added", context_id));
+
+        // Deploy proxy contract
         self.deploy_proxy_contract(context_id, account_id)
-            .then(Self::ext(env::current_account_id()).add_context_callback(context_id))
+            .then(Self::ext(env::current_account_id()).add_context_callback())
     }
 
     fn update_application(
@@ -306,20 +307,48 @@ impl ContextConfigs {
         context_id: Repr<ContextId>,
         account_id: AccountId,
     ) -> Promise {
-        // Deploy and initialize the proxy contract
-        Promise::new(account_id.clone())
+        // Known constants from NEAR protocol
+        //Use near tokens instead of yoctoNEAR
+        const ACCOUNT_CREATION_COST: u128 = 1_000_000_000_000_000_000_000;    // 0.001 NEAR
+        const MIN_ACCOUNT_BALANCE: u128 = 3_500_000_000_000_000_000_000;      // 0.0035 NEAR
+
+        // Calculate storage needs
+        let contract_bytes = self.proxy_code.get().clone().unwrap();
+        let storage_cost = (contract_bytes.len() as u128) * env::storage_byte_cost().as_yoctonear();
+        
+        // Calculate required deposit
+        let required_deposit = NearToken::from_yoctonear(
+            ACCOUNT_CREATION_COST +    // Cost to create account
+            MIN_ACCOUNT_BALANCE +      // Minimum balance required
+            storage_cost +             // Storage cost for contract
+            10_000_000_000_000_000_000 // Additional deposit for storage
+        );
+
+        require!(
+            env::account_balance() >= required_deposit,
+            "Insufficient contract balance for deployment"
+        );
+
+        // Calculate init gas dynamically
+        let init_args = serde_json::to_vec(&json!({
+            "context_id": context_id,
+            "context_config_account_id": env::current_account_id()
+        })).unwrap();
+
+        let init_gas = Gas::from_gas(
+            Gas::from_tgas(20).as_gas() +
+            ((init_args.len() as u64) + contract_bytes.len() as u64) * 100_000
+        );
+
+        Promise::new(account_id)
             .create_account()
-            .transfer(env::attached_deposit())
-            .deploy_contract(self.proxy_code.get().clone().unwrap())
+            .transfer(required_deposit)
+            .deploy_contract(contract_bytes)
             .function_call(
                 "init".to_owned(),
-                serde_json::to_vec(&json!({
-                    "context_id": context_id,
-                    "context_config_account_id": env::current_account_id()
-                }))
-                .unwrap(),
+                init_args,
                 NearToken::from_near(0),
-                Gas::from_tgas(30),
+                init_gas,
             )
             .then(Self::ext(env::current_account_id()).proxy_deployment_callback())
     }
@@ -360,42 +389,43 @@ impl ContextConfigs {
     pub fn proxy_deployment_callback(
         &mut self,
         #[callback_result] call_result: Result<(), PromiseError>,
-    ) {
-        if let Err(e) = call_result {
-            panic!("Failed to deploy proxy contract: {:?}", e);
+    ) -> PromiseOrValue<()> {
+        if let Ok(_) = call_result {
+            // Calculate actual storage used and refund excess
+            let actual_storage_used = env::storage_usage();
+            let actual_cost = (actual_storage_used as u128) * env::storage_byte_cost().as_yoctonear();
+            let deposit_used = env::attached_deposit().as_yoctonear();
+            
+            env::log_str("Successfully deployed proxy contract");
+            
+            if actual_cost < deposit_used {
+                Promise::new(env::current_account_id())
+                    .transfer(NearToken::from_yoctonear(deposit_used - actual_cost)).into()
+            } else {
+                PromiseOrValue::Value(())
+            }
+        } else {
+            env::panic_str("Failed to deploy proxy contract");
         }
     }
 
-    pub fn add_context_callback(&mut self, context_id: Repr<ContextId>) {
-        require!(
-            env::promise_results_count() == 1,
-            "Expected 1 promise result"
-        );
-
-        match env::promise_result(0) {
-            PromiseResult::Successful(_) => {
-                env::log_str(&format!("Context `{context_id}` added"));
-            }
-            _ => {
-                panic!("Failed to deploy proxy contract for context");
-            }
+    pub fn add_context_callback(
+        &mut self,
+        #[callback_result] call_result: Result<(), PromiseError>,
+    ) {
+        if let Err(e) = call_result {
+            panic!("Failed to deploy proxy contract for context: {:?}", e);
         }
     }
 
     #[private]
     #[handle_result]
-    pub fn update_contract_callback(&mut self) -> Result<(), &'static str> {
-        require!(
-            env::promise_results_count() == 1,
-            "Expected 1 promise result"
-        );
+    pub fn update_contract_callback(&mut self, #[callback_result] call_result: Result<(), PromiseError>) -> Result<(), &'static str> {
 
-        match env::promise_result(0) {
-            PromiseResult::Successful(_) => {
-                env::log_str("Successfully updated proxy contract code");
-                Ok(())
-            }
-            _ => Err("Failed to update proxy contract code"),
+        if let Err(e) = call_result {
+            panic!("Failed to update proxy contract: {:?}", e);
         }
+
+        Ok(())
     }
 }
--- a/contracts/context-config/src/mutate.rs
+++ b/contracts/context-config/src/mutate.rs
@@ -309,19 +309,19 @@ impl ContextConfigs {
     ) -> Promise {
         // Known constants from NEAR protocol
         //Use near tokens instead of yoctoNEAR
-        const ACCOUNT_CREATION_COST: u128 = 1_000_000_000_000_000_000_000;    // 0.001 NEAR
-        const MIN_ACCOUNT_BALANCE: u128 = 3_500_000_000_000_000_000_000;      // 0.0035 NEAR
+        const ACCOUNT_CREATION_COST: u128 = 1_000_000_000_000_000_000_000; // 0.001 NEAR
+        const MIN_ACCOUNT_BALANCE: u128 = 3_500_000_000_000_000_000_000; // 0.0035 NEAR
 
         // Calculate storage needs
         let contract_bytes = self.proxy_code.get().clone().unwrap();
         let storage_cost = (contract_bytes.len() as u128) * env::storage_byte_cost().as_yoctonear();
-        
+
         // Calculate required deposit
         let required_deposit = NearToken::from_yoctonear(
             ACCOUNT_CREATION_COST +    // Cost to create account
             MIN_ACCOUNT_BALANCE +      // Minimum balance required
             storage_cost +             // Storage cost for contract
-            10_000_000_000_000_000_000 // Additional deposit for storage
+            10_000_000_000_000_000_000, // Additional deposit for storage
         );
 
         require!(
@@ -333,11 +333,12 @@ impl ContextConfigs {
         let init_args = serde_json::to_vec(&json!({
             "context_id": context_id,
             "context_config_account_id": env::current_account_id()
-        })).unwrap();
+        }))
+        .unwrap();
 
         let init_gas = Gas::from_gas(
-            Gas::from_tgas(20).as_gas() +
-            ((init_args.len() as u64) + contract_bytes.len() as u64) * 100_000
+            Gas::from_tgas(20).as_gas()
+                + ((init_args.len() as u64) + contract_bytes.len() as u64) * 100_000,
         );
 
         Promise::new(account_id)
@@ -393,14 +394,16 @@ impl ContextConfigs {
         if let Ok(_) = call_result {
             // Calculate actual storage used and refund excess
             let actual_storage_used = env::storage_usage();
-            let actual_cost = (actual_storage_used as u128) * env::storage_byte_cost().as_yoctonear();
+            let actual_cost =
+                (actual_storage_used as u128) * env::storage_byte_cost().as_yoctonear();
             let deposit_used = env::attached_deposit().as_yoctonear();
-            
+
             env::log_str("Successfully deployed proxy contract");
-            
+
             if actual_cost < deposit_used {
                 Promise::new(env::current_account_id())
-                    .transfer(NearToken::from_yoctonear(deposit_used - actual_cost)).into()
+                    .transfer(NearToken::from_yoctonear(deposit_used - actual_cost))
+                    .into()
             } else {
                 PromiseOrValue::Value(())
             }
@@ -420,8 +423,10 @@ impl ContextConfigs {
 
     #[private]
     #[handle_result]
-    pub fn update_contract_callback(&mut self, #[callback_result] call_result: Result<(), PromiseError>) -> Result<(), &'static str> {
-
+    pub fn update_contract_callback(
+        &mut self,
+        #[callback_result] call_result: Result<(), PromiseError>,
+    ) -> Result<(), &'static str> {
         if let Err(e) = call_result {
             panic!("Failed to update proxy contract: {:?}", e);
         }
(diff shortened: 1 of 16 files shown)


contracts/context-config/src/mutate.rs
```diff
@@ -125,9 +124,11 @@ impl ContextConfigs {
             env::panic_str("context already exists");
         }
 
-        // Initiate proxy contract deployment with a callback for when it completes
+        env::log_str(&format!("Context `{}` added", context_id));
+
+        // Deploy proxy contract
         self.deploy_proxy_contract(context_id, account_id)
```
miraclx: the function says it 😅 

```suggestion
        self.deploy_proxy_contract(context_id, account_id)
```

contracts/context-config/src/mutate.rs
```diff
@@ -306,20 +307,49 @@ impl ContextConfigs {
         context_id: Repr<ContextId>,
         account_id: AccountId,
     ) -> Promise {
-        // Deploy and initialize the proxy contract
-        Promise::new(account_id.clone())
+        // Known constants from NEAR protocol
+        //Use near tokens instead of yoctoNEAR
+        const ACCOUNT_CREATION_COST: u128 = 1_000_000_000_000_000_000_000; // 0.001 NEAR
+        const MIN_ACCOUNT_BALANCE: u128 = 3_500_000_000_000_000_000_000; // 0.0035 NEAR
```
miraclx:
```suggestion
        const ACCOUNT_CREATION_COST: NearToken = NearToken::from_millinear(1); // 0.001 NEAR
        const MIN_ACCOUNT_BALANCE: NearToken = NearToken::from_millinear(35).saturating_div(10); // 0.0035 NEAR
        const STORAGE_TIP: NearToken = NearToken::from_millinear(1).saturating_div(100); // 0.00001 NEAR
```

contracts/context-config/src/mutate.rs
```diff
@@ -306,20 +307,49 @@ impl ContextConfigs {
...
+
+        // Calculate storage needs
+        let contract_bytes = self.proxy_code.get().clone().unwrap();
+        let storage_cost = (contract_bytes.len() as u128) * env::storage_byte_cost().as_yoctonear();
```
miraclx:
```suggestion
        let storage_cost = env::storage_byte_cost().saturating_mul(contract_bytes.len() as u128);
```

contracts/context-config/src/mutate.rs
```diff
@@ -306,20 +307,49 @@ impl ContextConfigs {
...
+
+        // Calculate required deposit
+        let required_deposit = NearToken::from_yoctonear(
+            ACCOUNT_CREATION_COST +    // Cost to create account
+            MIN_ACCOUNT_BALANCE +      // Minimum balance required
+            storage_cost +             // Storage cost for contract
+            10_000_000_000_000_000_000, // Additional deposit for storage
```
miraclx: `0.00001 NEAR`, this is intentional? just checking

contracts/context-config/src/mutate.rs
```diff
@@ -306,20 +307,49 @@ impl ContextConfigs {
...
+        );
```
miraclx:
```suggestion
        let required_deposit = ACCOUNT_CREATION_COST
            .saturating_add(MIN_ACCOUNT_BALANCE)
            .saturating_add(STORAGE_TIP)
            .saturating_add(storage_cost);
```

contracts/context-config/src/mutate.rs
```diff
@@ -306,20 +307,49 @@ impl ContextConfigs {
...
+
+        require!(
+            env::account_balance() >= required_deposit,
+            "Insufficient contract balance for deployment"
+        );
+
+        // Calculate init gas dynamically
```
miraclx: comment on wrong snippet?

contracts/context-config/src/mutate.rs
```diff
@@ -306,20 +307,49 @@ impl ContextConfigs {
...
+        let init_args = serde_json::to_vec(&json!({
+            "context_id": context_id,
+            "context_config_account_id": env::current_account_id()
+        }))
+        .unwrap();
+
+        let init_gas = Gas::from_gas(
+            Gas::from_tgas(20).as_gas()
+                + ((init_args.len() as u64) + contract_bytes.len() as u64) * 100_000,
```
miraclx: will `contract_bytes` be necessary here? since `init_gas` is only useful for the function call (which is already after deployment)

contracts/context-config/src/mutate.rs
```diff
@@ -360,42 +390,47 @@ impl ContextConfigs {
     pub fn proxy_deployment_callback(
         &mut self,
         #[callback_result] call_result: Result<(), PromiseError>,
-    ) {
-        if let Err(e) = call_result {
-            panic!("Failed to deploy proxy contract: {:?}", e);
+    ) -> PromiseOrValue<()> {
+        if let Ok(_) = call_result {
```
miraclx: let's flatten this out, `if call_result.is_err() {

contracts/context-config/src/mutate.rs
```diff
@@ -360,42 +390,47 @@ impl ContextConfigs {
...
+            let actual_storage_used = env::storage_usage();
+            let actual_cost =
+                (actual_storage_used as u128) * env::storage_byte_cost().as_yoctonear();
+            let deposit_used = env::attached_deposit().as_yoctonear();
+
+            env::log_str("Successfully deployed proxy contract");
+
+            if actual_cost < deposit_used {
+                Promise::new(env::current_account_id())
+                    .transfer(NearToken::from_yoctonear(deposit_used - actual_cost))
```
miraclx: this is transferring to itself? should be in the proxy contract, no?

contracts/context-config/src/mutate.rs
```diff
@@ -360,42 +390,47 @@ impl ContextConfigs {
...
-    ) {
-        if let Err(e) = call_result {
-            panic!("Failed to deploy proxy contract: {:?}", e);
+    ) -> PromiseOrValue<()> {
+        if let Ok(_) = call_result {
+            // Calculate actual storage used and refund excess
+            let actual_storage_used = env::storage_usage();
+            let actual_cost =
+                (actual_storage_used as u128) * env::storage_byte_cost().as_yoctonear();
+            let deposit_used = env::attached_deposit().as_yoctonear();
```
miraclx: use `NearToken`, we shouldn't need to convert to raw yocto

https://docs.rs/near-sdk/latest/near_sdk/struct.NearToken.html#method.saturating_sub

contracts/context-config/src/mutate.rs
```diff
@@ -360,42 +390,47 @@ impl ContextConfigs {
...
-                panic!("Failed to deploy proxy contract for context");
-            }
+    pub fn add_context_callback(
+        &mut self,
+        #[callback_result] call_result: Result<(), PromiseError>,
+    ) {
+        if let Err(e) = call_result {
+            panic!("Failed to deploy proxy contract for context: {:?}", e);
         }
     }
```
miraclx: this is already accounted for with `proxy_deployment_callback` right?

contracts/context-config/src/mutate.rs
```diff
@@ -360,42 +390,47 @@ impl ContextConfigs {
...
-        match env::promise_result(0) {
-            PromiseResult::Successful(_) => {
-                env::log_str("Successfully updated proxy contract code");
-                Ok(())
-            }
-            _ => Err("Failed to update proxy contract code"),
+    pub fn update_contract_callback(
+        &mut self,
+        #[callback_result] call_result: Result<(), PromiseError>,
+    ) -> Result<(), &'static str> {
```
miraclx: we shouldn't need to return a result here, right?

contracts/context-config/src/sys.rs
```diff
@@ -36,8 +36,12 @@ impl ContextConfigs {
         for (_, context) in self.contexts.drain() {
             drop(context.application.into_inner());
             context.members.into_inner().clear();
+            // let mut proxy = context.proxy.into_inner();
+            // proxy = AccountId::try_from("".to_string()).unwrap();
```
miraclx:
```suggestion
            context.proxy.into_inner();
```

should suffice, I think

contracts/context-config/src/sys.rs
```diff
@@ -36,8 +36,12 @@ impl ContextConfigs {
...
         }
 
+        self.proxy_code = LazyOption::new(b"p", None);
```
miraclx:
```suggestion
        self.proxy_code.set(None);
```

contracts/context-config/tests/sandbox.rs
```diff
@@ -931,19 +919,21 @@ async fn test_deploy() -> eyre::Result<()> {
     let context_public = context_secret.verifying_key();
     let context_id = context_public.to_bytes().rt()?;
 
+    // Fund node1 just for gas fees
     drop(
         root_account
-            .transfer_near(node1.id(), NearToken::from_near(100))
+            .transfer_near(node1.id(), NearToken::from_near(500))
             .await,
     );
```
miraclx: `create_subaccount(..).initial_balance`

contracts/context-config/tests/sandbox.rs
```diff
@@ -931,19 +919,21 @@ async fn test_deploy() -> eyre::Result<()> {
...
 
-    // Also transfer NEAR to the contract to cover proxy deployment costs
+    // Fund the contract with enough NEAR for proxy deployments
     drop(
         root_account
-            .transfer_near(contract.id(), NearToken::from_near(10))
+            .transfer_near(contract.id(), NearToken::from_near(100))
             .await,
     );
```
miraclx: don't drop the result

contracts/context-config/tests/sandbox.rs
```diff
@@ -979,13 +970,11 @@ async fn test_deploy() -> eyre::Result<()> {
...
-        .deposit(NearToken::from_near(10))
         .transact()
-        .await?
-        .into_result()?;
+        .await?;
 
-    // Uncomment to print the context creation result
-    // println!("Result of mutate: {:?}", res);
+    // println!("Execution result: {:#?}", res);
+    // println!("Logs: {:#?}", res.logs());  // Print the logs
```
miraclx: should we be asserting here?

contracts/context-config/tests/sandbox.rs
```diff
@@ -1036,7 +1025,7 @@ async fn test_deploy() -> eyre::Result<()> {
     );
 
     // Create proposal
-    let proposal_id = rng.gen();
+    let proposal_id = rand::thread_rng().gen();
```
miraclx: any reason this is changed?

contracts/context-config/tests/sandbox.rs
```diff
@@ -1053,21 +1042,40 @@ async fn test_deploy() -> eyre::Result<()> {
         },
     };
     let signed = Signed::new(&request, |p| alice_cx_sk.sign(p))?;
-
+    // println!("signed {:?}", signed);
     let res = node1
         .call(&proxy_address, "mutate")
-        .args_json(json!(signed))
+        .args_json(signed.clone())
```
miraclx: `&T: Serialize` [where](https://docs.rs/serde/latest/serde/trait.Serialize.html#impl-Serialize-for-%26T) `T: Serialize`, so you should only need this:

```suggestion
        .args_json(&signed)
```

contracts/context-config/tests/sandbox.rs
```diff
@@ -931,19 +919,21 @@ async fn test_deploy() -> eyre::Result<()> {
...
+            .transfer_near(node1.id(), NearToken::from_near(500))
             .await,
     );
 
-    // Also transfer NEAR to the contract to cover proxy deployment costs
+    // Fund the contract with enough NEAR for proxy deployments
     drop(
         root_account
-            .transfer_near(contract.id(), NearToken::from_near(10))
+            .transfer_near(contract.id(), NearToken::from_near(100))
```
miraclx: `create_subaccount(..).initial_balance`

contracts/context-config/tests/sandbox.rs
```diff
@@ -1053,21 +1042,40 @@ async fn test_deploy() -> eyre::Result<()> {
...
+        "Expected storage balance error, got: {}",
+        err
+    );
+
+    // Now fund the account
+    drop(
+        root_account
+            .transfer_near(&proxy_address, NearToken::from_near(5))
+            .await,
+    );
```
miraclx: don't drop the result

contracts/proxy-lib/tests/sandbox.rs
```diff
@@ -52,6 +52,13 @@ async fn setup_test(
         .expect("Contract not found");
 
     let proxy_id: AccountId = contract_id_str.parse()?;
+
+    drop(
```
miraclx: same here

crates/context/config/src/client/protocol/near.rs
```diff
@@ -273,7 +273,7 @@ impl Network {
                 method_name: method,
                 args,
                 gas: 300_000_000_000_000,
-                deposit: 5_000_000_000_000_000_000_000_000,
+                deposit: 0,
```
miraclx, line 276: 🎉

crates/context/config/src/lib.rs
```diff
@@ -186,7 +186,7 @@ pub struct ProposalApprovalWithSigner {
     pub added_timestamp: u64,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
miraclx: we can remove this now

contracts/context-config/tests/sandbox.rs
alenmestrov: hmm, I'm not sure if I'm following this. This is to fund the Context Contract account, subaccount is created in the contract itself.

miraclx: yeah.. still, funding the context contract doesn't have to be an explicit "transfer near" call since we `dev_deploy` above.

Also, to amend my comment, we shouldn't need this line at all.

Since `dev_` accounts (which is the context contract at this point), are already auto-funded with `100 N`

ref: https://github.com/near/near-workspaces-rs/blob/0f44e0b425777615370b0ef51dd8ef4626ccdeee/workspaces/src/network/sandbox.rs#L22

contracts/context-config/src/mutate.rs
```diff
@@ -392,21 +390,8 @@ impl ContextConfigs {
         #[callback_result] call_result: Result<(), PromiseError>,
     ) -> PromiseOrValue<()> {
```
miraclx: since this is a callback and doesn't actually construct any promises or return anything, I don't think a return type is required here

contracts/context-config/src/mutate.rs
```diff
@@ -423,15 +408,14 @@ impl ContextConfigs {
 
     #[private]
     #[handle_result]
-    pub fn update_contract_callback(
+    pub fn update_proxy_callback(
```
miraclx: any chance we can reuse the same callback for initial deployment and update?

the only specialization seems to be the contents of the log?

contracts/context-config/tests/sandbox.rs
```diff
@@ -920,18 +920,15 @@ async fn test_deploy() -> eyre::Result<()> {
...
     // Fund node1 just for gas fees
-    drop(
-        root_account
-            .transfer_near(node1.id(), NearToken::from_near(500))
-            .await,
-    );
+
+    let _unused = root_account
+        .transfer_near(node1.id(), NearToken::from_near(500))
+        .await;
```
miraclx: we're still not handling the results

```suggestion
    let _unused = root_account
        .transfer_near(node1.id(), NearToken::from_near(500))
        .await?
        .into_result()?;
```

same applies to the other cases

contracts/context-config/tests/sandbox.rs
```diff
@@ -1075,7 +1049,10 @@ async fn test_deploy() -> eyre::Result<()> {
         .await?;
 
     // Assert proposal creation result
-    assert!(res.is_success(), "Transaction failed: {:#?}", res);
+    let success_value = res.raw_bytes()?;
+    let proposal_result: ProposalWithApprovals = serde_json::from_slice(&success_value)?;
```
miraclx: how about `res.json::<ProposalWithApprovals>()`?

contracts/proxy-lib/src/mutate.rs
```diff
@@ -207,22 +207,34 @@ impl ProxyContract {
...
         Promise::new(env::current_account_id())
             .deploy_contract(new_code)
-            .then(Self::ext(env::current_account_id()).update_contract_callback())
+            .then(Self::ext(env::current_account_id()).update_contract_callback(storage_before))
     }
 
     #[private]
     #[handle_result]
-    pub fn update_contract_callback(&mut self) -> Result<(), &'static str> {
+    pub fn update_contract_callback(&mut self, storage_before: u64) -> Result<(), &'static str> {
```
miraclx: `#[callback_result]`

contracts/proxy-lib/src/mutate.rs
```diff
@@ -207,22 +207,34 @@ impl ProxyContract {
...
             "Expected 1 promise result"
         );
 
         match env::promise_result(0) {
             PromiseResult::Successful(_) => {
+                // Calculate storage difference and refund if needed
+                let storage_after = env::storage_usage();
+                if storage_after < storage_before {
+                    let refund = (storage_before - storage_after) as u128
+                        * env::storage_byte_cost().as_yoctonear();
```
miraclx: I like the idea of refunding the difference if the new contract is smaller, but since no deposit is attached when calling update from the context config, if the contract is larger and the proxy contract lacks the balance for storage, that will fail, no?

So first, we should compute the storage usage necessary for the update, attach that, and after the deployment is complete, refund this:

```text
(deposit + (SCB * old_SU)) - (SCB * new_SU)

where;
- SCB = storage cost per byte
- *_SU = storage usage
```

which should allow larger updates when the proxy lib lacks the balance for it, and refund excess when the wasm ends up smaller than before

also, I'm very curious, can you test that the difference between the new storage usage and old storage usage IS the new contract size - old contract size? (I wonder if they factor in precompilation size, even though they shouldn't as it could break determinism, good to check to be sure anyway)

but if that's the case, then the formula won't need to account for storage usage difference at all (since that can be exploited to steal from the context config)

how? they could submit large blobs to modify the state in-between the blocks of this deployment finalization, causing the proxy lib to refund less than the cost for the code, sneaky

we could easily mitigate this by storing the current code size in the proxy lib state, and using that to compute the difference when updating

contracts/context-config/tests/sandbox.rs
```diff
@@ -776,7 +770,7 @@ async fn main() -> eyre::Result<()> {
     // assert_eq!(state.len(), 1);
     // assert_eq!(state.get(&b"STATE"[..]).map(|v| v.len()), Some(24));
 
-    // After contract deployment
+    // // After contract deployment
     // let state_size = worker
     //     .view(contract.id(), "get_state_size")  // We'd need to add this method to the contract
```
miraclx: Just saw this, if `erase` works correctly, we should be able to restore `view_state` above. we shouldn't need a method for getting state size.

contracts/context-config/tests/sandbox.rs
```diff
@@ -920,15 +914,10 @@ async fn test_deploy() -> eyre::Result<()> {
...
-        .transfer_near(node1.id(), NearToken::from_near(500))
-        .await;
-
-    // Fund the contract with enough NEAR for proxy deployments
-    let _unused = root_account
-        .transfer_near(contract.id(), NearToken::from_near(100))
-        .await;
+        .transfer_near(node1.id(), NearToken::from_near(50))
+        .await?
+        .into_result();
```
miraclx: `create_subaccount().initial_balance()`

contracts/proxy-lib/src/lib.rs
```diff
@@ -56,6 +57,7 @@ impl ProxyContract {
             num_approvals: 3,
             active_proposals_limit: 10,
             context_storage: IterableMap::new(b"l"),
+            code_size: env::storage_usage() as u64,
```
miraclx: ah, at initialization is `storage_usage` == `code size`? I'd think it shouldn't be, right? Should vary by a few bytes, but yeah this is fine.

contracts/proxy-lib/src/mutate.rs
```diff
@@ -207,39 +207,41 @@ impl ProxyContract {
...
-    pub fn update_contract_callback(&mut self, storage_before: u64) -> Result<(), &'static str> {
+    pub fn update_contract_callback(
+        &mut self,
+        new_code_size: u64,
+        #[callback_result] call_result: Result<(), PromiseError>,
+    ) {
         require!(
             env::promise_results_count() == 1,
             "Expected 1 promise result"
         );
```
miraclx: still need this?

miraclx (approved):
lgtm ✨

Issue #916, opened 2024-11-01
fix: updating application should check the context config


miraclx:
Updating the application for a context should first submit it to the context config contract

Here's the relevant snippet:

https://github.com/calimero-network/core/blob/f1ddc4c2542e078ab3b95b52505a74a166d0ccab/crates/context/src/lib.rs#L685-L703

miraclx:
Resolved in #957

PR #948, opened 2024-11-07
chore: futher fixes and cleanups

miraclx:
- mitigate executor impersonation
- fix args parsing in CLI
- other touchups

PR #957, opened 2024-11-08
fix: update app id in contract

MatejVukosav:
Fix for https://github.com/calimero-network/core/issues/916

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -850,9 +850,7 @@ impl ContextManager {
             bail!("Context not found")
         };
 
-        let this = self.clone();
-
-        let Some(application) = this.get_application(&application_id)? else {
+        let Some(application) = self.get_application(&application_id)? else {
             bail!("Application with id {:?} not found", application_id)
         };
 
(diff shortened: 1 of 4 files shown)


crates/meroctl/src/cli/context/create.rs
```diff
@@ -260,7 +260,9 @@ async fn update_context_application(
         &format!("admin-api/dev/contexts/{context_id}/application"),
     )?;
 
-    let request = UpdateContextApplicationRequest::new(application_id);
+    //How to make conversion from libp2p to public key
+    let public_key = keypair.public();
+    let request = UpdateContextApplicationRequest::new(application_id, public_key);
```
miraclx: no, it shouldn't be the libp2p keypair, this should be the `member_public_key` from the create context response

this was wrong about `context call` as well, which I [fixed some days ago](https://github.com/calimero-network/core/pull/948/files#diff-f5d45fa7e6c0662f1afc369ed781cd633d630781f205ce957203e02437621722L77-R97)

crates/context/src/lib.rs
```diff
@@ -849,6 +850,43 @@ impl ContextManager {
             bail!("Context not found")
         };
 
+        let this = self.clone();
```
miraclx: why clone `self`?

crates/context/src/lib.rs
```diff
@@ -834,10 +834,11 @@ impl ContextManager {
...
     }
 
-    pub fn update_application_id(
+    pub async fn update_application_id(
         &self,
         context_id: ContextId,
         application_id: ApplicationId,
+        signer_id: PublicKey,
     ) -> EyreResult<()> {
         // todo! use context config
```
miraclx: we can remove this now

crates/context/src/lib.rs
```diff
@@ -849,6 +850,43 @@ impl ContextManager {
...
+        let Some(application) = this.get_application(&application_id)? else {
+            bail!("Application with id {:?} not found", application_id)
+        };
+
+        let Some(ContextIdentityValue {
+            private_key: Some(requester_secret),
+        }) = handle.get(&ContextIdentityKey::new(context_id, signer_id))?
+        else {
+            bail!(
+                "Identity {:?} not found in context {:?}",
```
```suggestion
                "'{}' is not a member of '{}'",
```

crates/context/src/lib.rs
```diff
@@ -849,6 +850,43 @@ impl ContextManager {
...
+                context_id
+            )
+        };
+
+        let _ = this
+            .config_client
+            .mutate::<ContextConfigEnv>(
+                this.client_config.new.protocol.as_str().into(),
+                this.client_config.new.network.as_str().into(),
+                this.client_config.new.contract_id.as_str().into(),
```
miraclx: we should read the context config from the `ContextConfig` store record

for example

https://github.com/calimero-network/core/blob/db644002b9093b0363e2af55347ebb849161e07e/crates/context/src/lib.rs#L397-L399

crates/context/src/lib.rs
```diff
@@ -849,6 +850,43 @@ impl ContextManager {
...
+            )
+            .update_application(
+                context_id.rt().expect("bla"),
```
```suggestion
                context_id.rt().expect("infallible conversion"),
```

crates/context/src/lib.rs
MatejVukosav: updated

miraclx (approved):

PR #958, opened 2024-11-08
fix: correct casting of returned proposal ID

alenmestrov:
Fixed issue that occured during tests in PR reviews. 

Correctly casted proposal id to [u8; 32] instead of u8.

### DIFF
--- a/contracts/context-config/tests/sandbox.rs
+++ b/contracts/context-config/tests/sandbox.rs
@@ -6,8 +6,7 @@ use calimero_context_config::types::{
     Application, Capability, ContextIdentity, Revision, Signed, SignerId,
 };
 use calimero_context_config::{
-    ContextRequest, ContextRequestKind, Proposal, ProposalAction, ProxyMutateRequest, Request,
-    RequestKind, SystemRequest,
+    ContextRequest, ContextRequestKind, Proposal, ProposalAction, ProposalWithApprovals, ProxyMutateRequest, Request, RequestKind, SystemRequest
 };
 use ed25519_dalek::{Signer, SigningKey};
 use eyre::Ok;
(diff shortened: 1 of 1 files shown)


contracts/context-config/tests/sandbox.rs
```diff
@@ -1066,7 +1066,14 @@ async fn test_deploy() -> eyre::Result<()> {
...
     assert_eq!(proposal_result["num_approvals"], 1);
-    let created_proposal_id = proposal_result["proposal_id"].as_u64().unwrap();
+    let created_proposal_id: [u8; 32] = proposal_result["proposal_id"]
+        .as_array()
+        .expect("proposal_id should be a byte array")
+        .iter()
+        .map(|v| v.as_u64().unwrap() as u8)
+        .collect::<Vec<u8>>()
+        .try_into()
+        .expect("proposal_id should be 32 bytes");
```
miraclx: `proposal_id` really should've been represented as a new type`ProposalId(Identity)`, like the others in `config/types`

then we'd have been able to use `Repr<ProposalId>`

but for now, here, I think leaning in to `serde_json` should suffice

```suggestion
    let created_proposal_id: [u8; 32] = serde_json::from_value(proposal_result["proposal_id"].clone())?;
```

or alternatively, and perhaps preferably, deserialize to the return type `ProposalWithApprovals`

contracts/context-config/tests/sandbox.rs
```diff
@@ -1036,7 +1036,7 @@ async fn test_deploy() -> eyre::Result<()> {
     );
 
     // Create proposal
-    let proposal_id = rng.gen();
+    let proposal_id = rand::thread_rng().gen();
```
miraclx, line 1039: hm, was there a reason for this?

alenmestrov: just because of consistency since it was done in that way in proxy_lib tests

MatejVukosav (approved):

miraclx (approved):

PR #960, opened 2024-11-10
Added update command

petarjuki7:
App update command

### DIFF
--- a/crates/meroctl/src/cli/context/update.rs
+++ b/crates/meroctl/src/cli/context/update.rs
@@ -40,7 +40,11 @@ pub struct UpdateCommand {
     )]
     path: Option<Utf8PathBuf>,
 
-    #[clap(long, help = "Metadata needed for the application installation")]
+    #[clap(
+        long,
+        conflicts_with = "application_id",
+        help = "Metadata needed for the application installation"
+    )]
     metadata: Option<String>,
 
     #[clap(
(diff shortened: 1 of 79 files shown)


MatejVukosav:
Can you add to docs how to use it?

petarjuki7:
`cargo run -p meroctl -- --node-name <node_name> context update -a  <app_id> --context-id <contrext_id>  -p <member_public_key> --path <path_to_app>`

miraclx (commented):
so.. we need the following:

```console
$ meroctl context update --context-id <context-id> -a <application-id> --as <identity>
# one-off update

$ meroctl context update --context-id <context-id> --path ./path/to/file --as <identity>
# one-off update

$ meroctl context update --context-id <context-id> --path ./path/to/file --watch --as <identity>
# recurring update
```

see the `call` command for how identity is managed

`<application-id>` and `<path>` are mutually exclusive, and `--watch` should only be allowed when `--path` is present.

crates/meroctl/src/cli/context/update.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    #[clap(
+        long,
+        conflicts_with = "application_id",
+        help = "Path to the application file to watch and install locally"
+    )]
+    path: Option<Utf8PathBuf>,
+
+    #[clap(long, help = "Metadata needed for the application installation")]
+    metadata: Option<String>,
```
miraclx: this should also conflict with `application_id`

crates/meroctl/src/cli/context/update.rs
```diff
@@ -0,0 +1,282 @@
...
+                    &config.identity,
+                    executor_public_key,
+                )
+                .await?;
+            }
+            Self {
+                context_id,
+                application_id: None,
+                path: Some(path),
+                metadata: None,
```
miraclx:
```suggestion
                metadata,
```

crates/meroctl/src/cli/context/update.rs
```diff
@@ -0,0 +1,282 @@
...
+                    &client,
+                    multiaddr,
+                    context_id,
+                    path,
+                    metadata.map(String::into_bytes),
+                    &config.identity,
+                    executor_public_key,
+                )
+                .await?;
+            }
```
miraclx: the only difference between these two would be when `watch` is true, call `watch_app*`

but they should both "install" first

crates/meroctl/src/cli/context/update.rs
```diff
@@ -81,88 +85,58 @@ impl UpdateCommand {
...
-                    path,
-                    metadata.map(String::into_bytes),
+                    path.clone(),
+                    metadata.clone(),
                     &config.identity,
-                    executor_public_key,
                 )
                 .await?;
+                if self.watch {
+                    watch_app_and_update_context(
```
miraclx, line 120: this won't update the context unless the application changes

we should update and then watch, see `context create` for reference

miraclx (approved):

PR #962, opened 2024-11-11
feat(merod): implement generic config edit

fbozic:
Implement a more generic config edit solution. Interface similar to docker build-arg pattern.  
Examples of string, number, and bool args:
```
cargo run -p merod -- --node-name node1 config \
  --arg=context.config.new.contract_id='"myamestrov.testnet"' \
  --arg=sync.timeout_ms=20000 \
  --arg=discovery.mdns=false
```

### DIFF
--- a/contracts/proxy-lib/tests/common/proxy_lib_helper.rs
+++ b/contracts/proxy-lib/tests/common/proxy_lib_helper.rs
@@ -6,7 +6,7 @@ use calimero_context_config::{
 use ed25519_dalek::{Signer, SigningKey};
 use near_sdk::AccountId;
 use near_workspaces::result::{ExecutionFinalResult, ViewResultDetails};
-use near_workspaces::{Account, Contract};
+use near_workspaces::Account;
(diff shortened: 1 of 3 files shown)


MatejVukosav (approved):

crates/merod/src/cli/config.rs
```diff
@@ -1,209 +1,76 @@
...
-    pub server_host: Vec<IpAddr>,
-
-    /// Port to listen on for RPC
-    #[arg(long, value_name = "PORT")]
-    pub server_port: Option<u16>,
-
-    #[command(flatten)]
-    pub mdns: Option<MdnsArgs>,
+    /// Key-value pairs to be added or updated in the TOML file
+    #[clap(short, long, value_parser = value_parser!(KeyValuePair))]
```
miraclx, line 20:
```suggestion
```

You should be able to skip `value_parser` as `T: FromStr` is implied. Also, can we skip the `--arg` flag?

PR #966, opened 2024-11-13
fix: prevent external mutation on execution failure

miraclx:
Failed transactions should not;

1. Apply their changes to storage
2. Broadcast a state delta
3. Have any effect on external systems (blockchain)


alenmestrov (approved):
LGTM

PR #969, opened 2024-11-15
fix: state sync

miraclx:
Remove `type_id` from `calimero_storage`, which unblocks sync.

Address other classes of misuse of the crdt subsystem and questionable design

### DIFF
--- a/apps/kv-store/src/__private.rs
+++ b/apps/kv-store/src/__private.rs
@@ -1,7 +1,6 @@
-use calimero_sdk::borsh::{from_slice, to_vec};
+use calimero_sdk::borsh::from_slice;
 use calimero_sdk::{app, env};
-use calimero_storage::collections::unordered_map::{Entry, UnorderedMap};
-use calimero_storage::entities::Data;
+use calimero_storage::address::Id;
 use calimero_storage::integration::Comparison;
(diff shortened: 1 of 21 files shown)


MatejVukosav:
Can you elaborate more on misuse and questionable design?

how removing type_id unblocks the sync?

chefsale (approved):
ship it

miraclx:
> Can you elaborate more on misuse and questionable design? What was wrong?

Misuse: `Interface`'s functions were unintuitive and our attempts to abstract over it to provide easier-to-use APIs through `Unordered{Set,Map} & Vector` still required calling `Interface::save` on the collection after all mutations happen. When to call that and by whom was an open question, which led to one class of issues.

Questionable design:

- Many functions in `index.rs` call `get_index` multiple times to extract bits of unmutated information, this is still the case, but it's been reduced to an extent.
- Structures contained information like creation, update, merkle hash, etc.. that would be sent over the wire unnecessarily.
- Many functions in `interface.rs` required a definition of `T: Data` which led to the `type_id` requirement. 

> how removing `type_id` unblocks the sync?

- Just look at `__private.rs` from before. Since functions required knowledge of which type had which data updated, we had to exhaustively be aware of all types that constitute the state. That was unwieldy and led us down the path of `__private.rs`.
That is no longer the case.

PR #977, opened 2024-11-20
chore:  add rust husky

MatejVukosav:
- use husky-rs instead of regular husky
- use dprint instead of prettier
- fix md files based on dprint
- remove unused eslint files

> Note
> After this is merged; execute `cargo clean `and `cargo build` to setup husky


crates/merod/Cargo.toml
```diff
@@ -36,6 +36,7 @@ calimero-node = { path = "../node" }
 calimero-network = { path = "../network" }
 calimero-server = { path = "../server", features = ["jsonrpc", "websocket", "admin"] }
 calimero-store = { path = "../store" }
+husky-rs.workspace = true # activate husky hooks
```
miraclx, line 39: shouldn't it be `[dev-dependencies]`?

.prettierignore
miraclx, line 1: are these accounted for someway with dprint?

Cargo.lock
miraclx, line 1: a lot more changed here than needed to, try this

```console
```

to restore the lockfile version from master, then `cargo build` on top of that

Wholesale dependency upgrades which `cargo update` does must be handled more consciously, ideally a separate effort.

miraclx: restore & build also seems to do a wholesale update in this case, this appears to be the minimum lockfile requirement 👍🏽

MatejVukosav: yes, it works,  just pushed. tnx!

MatejVukosav:
Bump

PR #978, opened 2024-11-20
Added nonce handling

petarjuki7:
Nonces are created in the `SharedKey`s constructor.

Sent as a part of the message in the `send()` funciton and extracted in the `recv()` function.

### DIFF
--- a/crates/crypto/src/lib.rs
+++ b/crates/crypto/src/lib.rs
@@ -1,6 +1,9 @@
 use calimero_primitives::identity::{PrivateKey, PublicKey};
 use ed25519_dalek::{SecretKey, SigningKey};
+use rand::{thread_rng, Rng};
 use ring::aead;
+
+pub const NONCE_LEN: usize = 12;
 
 #[derive(Copy, Clone, Debug)]
 pub struct SharedKey {
@@ -10,26 +13,31 @@ pub struct SharedKey {
 #[derive(Debug)]
 pub struct Record {
     pub token: Vec<u8>,
-    pub nonce: [u8; 12],
+    pub nonce: [u8; NONCE_LEN],
 }
 
 impl SharedKey {
-    pub fn new(sk: &PrivateKey, pk: &PublicKey) -> Self {
-        SharedKey {
-            key: (SigningKey::from_bytes(sk).to_scalar()
-                * curve25519_dalek::edwards::CompressedEdwardsY(**pk)
-                    .decompress()
-                    .expect("pk should be guaranteed to be the y coordinate"))
-            .compress()
-            .to_bytes(),
-        }
+    pub fn new(sk: &PrivateKey, pk: &PublicKey) -> (Self, [u8; NONCE_LEN]) {
+        let nonce = thread_rng().gen::<[u8; NONCE_LEN]>();
+        (
+            SharedKey {
+                key: (SigningKey::from_bytes(sk).to_scalar()
+                    * curve25519_dalek::edwards::CompressedEdwardsY(**pk)
+                        .decompress()
+                        .expect("pk should be guaranteed to be the y coordinate"))
+                .compress()
+                .to_bytes(),
+            },
+            nonce,
+        )
     }
 
-    pub fn from_sk(sk: &PrivateKey) -> Self {
-        SharedKey { key: **sk }
+    pub fn from_sk(sk: &PrivateKey) -> (Self, [u8; NONCE_LEN]) {
+        let nonce = thread_rng().gen::<[u8; NONCE_LEN]>();
+        (SharedKey { key: **sk }, nonce)
     }
 
-    pub fn encrypt(&self, payload: Vec<u8>, nonce: [u8; 12]) -> Option<Vec<u8>> {
+    pub fn encrypt(&self, payload: Vec<u8>, nonce: [u8; NONCE_LEN]) -> Option<Vec<u8>> {
         let encryption_key =
             aead::LessSafeKey::new(aead::UnboundKey::new(&aead::AES_256_GCM, &self.key).ok()?);
 
@@ -45,7 +53,7 @@ impl SharedKey {
         Some(cipher_text)
     }
 
-    pub fn decrypt(&self, cipher_text: Vec<u8>, nonce: [u8; aead::NONCE_LEN]) -> Option<Vec<u8>> {
+    pub fn decrypt(&self, cipher_text: Vec<u8>, nonce: [u8; NONCE_LEN]) -> Option<Vec<u8>> {
         let decryption_key =
             aead::LessSafeKey::new(aead::UnboundKey::new(&aead::AES_256_GCM, &self.key).ok()?);
 
@@ -73,16 +81,15 @@ mod tests {
 
     #[test]
     fn test_encrypt_decrypt() -> eyre::Result<()> {
-        let mut csprng = rand::thread_rng();
+        let mut csprng = thread_rng();
 
         let signer = PrivateKey::random(&mut csprng);
         let verifier = PrivateKey::random(&mut csprng);
 
-        let signer_shared_key = SharedKey::new(&signer, &verifier.public_key());
-        let verifier_shared_key = SharedKey::new(&verifier, &signer.public_key());
+        let (signer_shared_key, nonce) = SharedKey::new(&signer, &verifier.public_key());
+        let (verifier_shared_key, _nonce) = SharedKey::new(&verifier, &signer.public_key());
 
         let payload = b"privacy is important";
-        let nonce = [0u8; aead::NONCE_LEN];
 
         let encrypted_payload = signer_shared_key
             .encrypt(payload.to_vec(), nonce)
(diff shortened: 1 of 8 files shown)


crates/node/src/sync/blobs.rs
```diff
@@ -187,6 +189,7 @@ impl Node {
                     },
                 },
                 Some(shared_key),
+                Some(nonce),
```
petarjuki7: Should we generate a new `nonce` each time in the `while` loop?

crates/crypto/src/lib.rs
```diff
@@ -10,26 +13,31 @@ pub struct SharedKey {
 #[derive(Debug)]
 pub struct Record {
     pub token: Vec<u8>,
-    pub nonce: [u8; 12],
+    pub nonce: [u8; NONCE_LEN],
 }
```
miraclx: what is this used for?

crates/node/src/sync/blobs.rs
miraclx: we can put the next nonce in the last message?

crates/crypto/src/lib.rs
```diff
@@ -10,26 +13,31 @@ pub struct SharedKey {
...
-    pub fn new(sk: &PrivateKey, pk: &PublicKey) -> Self {
-        SharedKey {
-            key: (SigningKey::from_bytes(sk).to_scalar()
-                * curve25519_dalek::edwards::CompressedEdwardsY(**pk)
-                    .decompress()
-                    .expect("pk should be guaranteed to be the y coordinate"))
-            .compress()
-            .to_bytes(),
-        }
+    pub fn new(sk: &PrivateKey, pk: &PublicKey) -> (Self, [u8; NONCE_LEN]) {
```
miraclx: why not just leave it to the caller to generate the nonce?

crates/node/src/lib.rs
```diff
@@ -385,17 +388,18 @@ impl Node {
                 .get_sender_key(&context.id, &executor_public_key)?
                 .ok_or_eyre("expected own identity to have sender key")?;
 
-            let shared_key = SharedKey::from_sk(&sender_key);
+            let (shared_key, nonce) = SharedKey::from_sk(&sender_key);
```
miraclx: I think you should invert the process, instead of the nonce coming from `crypto`, it should always come from the caller.

`from_sk` and `new`, stay the same as before, only `encrypt` and `decrypt` need to change

crates/node/src/sync.rs
```diff
@@ -29,15 +29,19 @@ async fn send(
     stream: &mut Stream,
     message: &StreamMessage<'_>,
     shared_key: Option<SharedKey>,
+    nonce: Option<[u8; NONCE_LEN]>,
```
miraclx:
```suggestion
    shared_key: Option<(SharedKey, Nonce)>,
```

crates/node/src/sync.rs
```diff
@@ -56,10 +60,18 @@ async fn recv(
     let message_data = message?.data.into_owned();
 
     let data = match shared_key {
-        Some(key) => match key.decrypt(message_data, [0; 12]) {
-            Some(data) => data,
-            None => bail!("decryption failed"),
-        },
+        Some(key) => {
+            let (message, nonce) = message_data.split_at(message_data.len() - NONCE_LEN);
```
miraclx: what's the issue with receiving the nonce from the caller?

miraclx (commented):
might make sense to introduce a `Nonce` type to `crypto` at this point.

also, `StreamMessage::Init` should have a `nonce` field.

and `StreamMessage::Message` can have a `nonce` field for the next message.

crates/node/src/sync/key.rs
```diff
@@ -103,7 +128,8 @@ impl Node {
             .get_private_key(context.id, our_identity)?
             .ok_or_eyre("expected own identity to have private key")?;
 
-        let (shared_key, nonce) = SharedKey::new(&private_key, &their_identity);
+        let shared_key = SharedKey::new(&private_key, &their_identity);
+        let new_nonce = thread_rng().gen::<Nonce>();
```
miraclx: we don't need one here, right?

crates/node/src/types.rs
```diff
@@ -30,10 +30,12 @@ pub enum StreamMessage<'a> {
         party_id: PublicKey,
         // nonce: usize,
```
miraclx: this was here already

also, let's use the `next_` prefix so it's easier to follow

crates/node/src/sync/blobs.rs
```diff
@@ -178,26 +190,33 @@ impl Node {
...
         while let Some(chunk) = blob.try_next().await? {
+            let new_nonce = thread_rng().gen::<Nonce>();
             send(
                 stream,
                 &StreamMessage::Message {
                     sequence_id: sequencer.next(),
                     payload: MessagePayload::BlobShare {
                         chunk: chunk.into_vec().into(),
                     },
+                    next_nonce: new_nonce,
```
miraclx:
```suggestion
            let next_nonce = thread_rng().gen::<Nonce>();

            send(
                stream,
                &StreamMessage::Message {
                    sequence_id: sequencer.next(),
                    payload: MessagePayload::BlobShare {
                        chunk: chunk.into_vec().into(),
                    },
                    next_nonce,
```

crates/node/src/sync/blobs.rs
```diff
@@ -178,26 +190,33 @@ impl Node {
...
         }
 
+        let new_nonce = thread_rng().gen::<Nonce>();
+
         send(
             stream,
             &StreamMessage::Message {
                 sequence_id: sequencer.next(),
                 payload: MessagePayload::BlobShare { chunk: b"".into() },
+                next_nonce: new_nonce,
```
miraclx: we can skip generating a new nonce here since we won't actually be sending any further messages

```suggestion
        send(
            stream,
            &StreamMessage::Message {
                sequence_id: sequencer.next(),
                payload: MessagePayload::BlobShare { chunk: b"".into() },
                next_nonce: [0; 16],
```

crates/node/src/sync/key.rs
```diff
@@ -60,26 +72,39 @@ impl Node {
         our_identity: PublicKey,
         their_identity: PublicKey,
         stream: &mut Stream,
+        nonce: Nonce,
```
miraclx:
```suggestion
        their_nonce: Nonce,
```

crates/node/src/sync/key.rs
```diff
@@ -60,26 +72,39 @@ impl Node {
...
     ) -> eyre::Result<()> {
         debug!(
             context_id=%context.id,
             their_identity=%their_identity,
             "Received key share request",
         );
 
+        let their_nonce = nonce;
+
+        let nonce = thread_rng().gen::<Nonce>();
```
miraclx: we can use `our_` prefix here, making it easier to read and maintain

crates/node/src/sync/key.rs
```diff
@@ -88,6 +113,8 @@ impl Node {
         our_identity: PublicKey,
         their_identity: PublicKey,
         stream: &mut Stream,
+        sending_nonce: Nonce,
+        receiving_nonce: Nonce,
```
miraclx: `our_`, `their_`?

crates/node/src/sync/key.rs
```diff
@@ -115,19 +142,27 @@ impl Node {
             &StreamMessage::Message {
                 sequence_id: sqx_out.next(),
                 payload: MessagePayload::KeyShare { sender_key },
+                next_nonce: sending_nonce, // the next nonce will not be used
```
miraclx: code is clear enough

```suggestion
                next_nonce: sending_nonce,
```

crates/node/src/sync/blobs.rs
```diff
@@ -190,31 +194,29 @@ impl Node {
...
                     },
-                    next_nonce: new_nonce,
+                    next_nonce,
                 },
                 Some((shared_key, nonce)),
             )
             .await?;
 
-            nonce = new_nonce;
+            nonce = next_nonce;
```
miraclx: let's use `our_nonce` anyway for consistency, even though there's no "their_` variant

crates/node/src/sync/state.rs
```diff
@@ -268,17 +265,17 @@ impl Node {
 
         let mut sqx_in = Sequencer::default();
 
-        let mut sending_nonce = sending_nonce;
-        let mut receiving_nonce = receiving_nonce;
+        let mut our_nonce = our_nonce;
+        let mut their_nonce = their_nonce;
```
```suggestion
```

crates/node/src/sync/state.rs
```diff
@@ -290,7 +287,7 @@ impl Node {
                 }
             };
 
-            receiving_nonce = new_receiving_nonce;
+            their_nonce = new_their_nonce;
```
```suggestion
            their_nonce = their_new_nonce;
```

crates/node/src/sync/state.rs
```diff
@@ -314,7 +311,7 @@ impl Node {
                 "State sync outcome",
             );
 
-            let new_sending_nonce = thread_rng().gen::<Nonce>();
+            let new_our_nonce = thread_rng().gen::<Nonce>();
```
```suggestion
            let our_new_nonce = thread_rng().gen::<Nonce>();
```

miraclx (commented):
can you appropriately name the PR and have you tested it?

PR #979, opened 2024-11-21
fix: run proxy tests with a single worker instance

xilosada:
* Introdued a single shared Worker instance across all tests to avoid conflicts caused by reusing ports.

### DIFF
--- a/contracts/proxy-lib/tests/common/mod.rs
+++ b/contracts/proxy-lib/tests/common/mod.rs
@@ -35,3 +35,17 @@ pub async fn create_account_with_balance(
         .into_result()?;
     Ok(account)
 }
+
+pub fn generate_near_account_id() -> Result<String> {
+    let random_account_id: String = rand::thread_rng()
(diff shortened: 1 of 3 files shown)


contracts/proxy-lib/tests/common/mod.rs
```diff
@@ -35,3 +35,17 @@ pub async fn create_account_with_balance(
...
+        .take(6)
+        .chain(
+            rand::thread_rng()
+                .sample_iter(&rand::distributions::Uniform::new_inclusive('0', '9'))
+                .take(2),
+        ) // Append digits
+        .collect();
+
+    Ok(random_account_id)
+}
```
miraclx: can we `Worker::dev_create_account` instead?

https://docs.rs/near-workspaces/latest/near_workspaces/struct.Worker.html#method.dev_create_account

miraclx (commented):
one case where [`#[rstest]`](https://docs.rs/rstest) could come in handy, but best not to overcomplicate it, this is okay!

PR #981, opened 2024-11-21
feat: Context and Proxy Starknet contract integration into Calimero

alenmestrov:
# Integrate Starknet Blockchain integration into Calimero

## Summary

Implemented logic for communicating with Starknet smart contracts (both Context and Proxy contracts).
The implementation was done based on NEAR integration. For some cases we needed to include custom logic for casting from NEAR to Starknet and vice versa. 

## Test plan
**Setup of devnet environment and preparation of contracts can be found in these two repositories:**
[Context contract](https://github.com/calimero-network/starknet-proxy-contract)
[Proxy contract](https://github.com/calimero-network/starknet-proxy-contract)

1. Initialize the node with additional flag `--protocol starknet`

```
cargo run -p merod -- --node-name starknet --home data init --server-port 2428 --swarm-port 2528 --protocol starknet
```

2. After initialization go to `Config.toml` file in node directory and adjust the values there to use devnet environment. 

```
[context.config.new]
protocol = "starknet"
network = "sepolia"
contract_id = "HERE_INSERT_CONTEXT_CONTRACT_ADDRESS"
```

3. Set `context.config.signer` `use` value to `self`

```
[context.config.signer]
use = "self"
```

4. Adjust signer details:

```
[context.config.signer.self.starknet.sepolia]
rpc_url = "http://127.0.0.1:5050/rpc"
account_id = "USE_ONE_OF_DEVNET_ACCOUNT_ADDRESSES"
public_key = "ACCOUNT_PUBLIC_KEY"
secret_key = "ACCOUNT_SECRET_KEY"
```

5. Start the node
```
cargo run -p merod -- --node-name starknet --home data run
```

6. Setup Demo Blockchain application
[Repository](https://github.com/calimero-network/demo-blockchain-integrations)

7. In first node install application wasm, create context, then initalize one more node on different ports, and add that node to context created in first node. 
Steps are explained [here](https://calimero-network.github.io/getting-started/setup)

8. Start Demo blockchain app frontend, connect to nodes through Admin Dashboard

9. Create demo proposal and approve proposal on different node

10. Results should be visible on both nodes


Cargo.toml
```diff
@@ -105,7 +105,7 @@ semver = "1.0.22"
 serde = "1.0.196"
 serde_json = "1.0.113"
 serde_with = "3.9.0"
-starknet = "0.12.0"
+starknet = { git = "https://github.com/xJonathanLEI/starknet-rs" }
```
xilosada: can you add a comment?

crates/node/src/sync.rs
```diff
@@ -106,7 +106,7 @@ impl Node {
         chosen_peer: PeerId,
     ) -> EyreResult<()> {
         let mut context = self.ctx_manager.sync_context_config(context_id).await?;
-
+        println!("initiating sync for context: {:?}", context);
```
xilosada: clean

crates/context/config/src/client/protocol/starknet.rs
```diff
@@ -314,7 +314,40 @@ impl Network {
...
+                    starknet::core::types::ExecutionResult::Succeeded => {
+                        for event in invoke_receipt.events.iter() {
+                            if event.from_address == contract_id {
+                                let result = StarknetProposalWithApprovals::decode(&event.data)
+                                    .map_err(|e| StarknetError::Custom {
+                                        operation: ErrorOperation::Query,
+                                        reason: format!("Failed to decode event: {:?}", e),
+                                    })?;
+                                
+                                println!("Event decoded: {:?}", result);
```
xilosada: nit

Cargo.toml
miraclx: ideally, use a dependency override

https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

for example:

https://github.com/calimero-archive/PortalApps/blob/93d73e572049e0855b9400e0750dd82dfe1b32f1/Cargo.toml#L62

crates/context/Cargo.toml
miraclx, line 1: these are unused?

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -56,28 +62,76 @@ impl<'a> Method<Near> for Mutate<'a> {
...
+        let signing_key = StarknetSigningKey::from_secret_scalar(secret_scalar);
+        let verifying_key = signing_key.verifying_key().scalar();
+        let verifying_key_bytes = verifying_key.to_bytes_be();
+
+        // Derive ed25519 key from private key
+        let user_key = SigningKey::from_bytes(&self.signing_key).verifying_key();
+        let user_key_bytes = user_key.to_bytes();
+
+        // Create Repr wrapped ContextIdentity instances
+        let signer_id = Repr::new(ContextIdentity::from_bytes(|bytes| {
```
miraclx: you should never need to call `from_bytes` manually

```suggestion
        let signer_id = verifying_key_bytes.rt().expect(..);
```

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -56,28 +62,76 @@ impl<'a> Method<Near> for Mutate<'a> {
...
+            bytes.copy_from_slice(&verifying_key_bytes);
+            Ok(32)
+        })?);
+
+        let user_id = Repr::new(ContextIdentity::from_bytes(|bytes| {
```
miraclx: same

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +42,69 @@ impl Method<Starknet> for ApplicationRequest {
...
 
-    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        if response.is_empty() {
+            return Err(eyre::eyre!("No application found"));
+        }
+
+        // Check if it's a None response (single zero Felt)
+        if response.len() == 32 && response.iter().all(|&x| x == 0) {
```
miraclx:
```suggestion
        if response == [0; 32] {
```

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +42,69 @@ impl Method<Starknet> for ApplicationRequest {
...
+        // Convert both Felts to bytes and concatenate
+        let mut result = Vec::new();
+        result.extend_from_slice(&high_felt.to_bytes_be());
+        result.extend_from_slice(&low_felt.to_bytes_be());
+        Ok(result)
     }
 
-    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
```
miraclx, line 55: we really should've implemented `SNApplication: Decode`, as this will suffer in maintainership, and is a likely candidate for bugs. but we can resolve that in a later iteration.

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +42,69 @@ impl Method<Starknet> for ApplicationRequest {
...
+        id_bytes[16..].copy_from_slice(&response[48..64]); // low part
+        let id = Repr::new(id_bytes.rt()?);
+
+        // Next two Felts are blob id (high/low)
+        let mut blob_bytes = [0u8; 32];
+        blob_bytes[..16].copy_from_slice(&response[80..96]); // high part
+        blob_bytes[16..].copy_from_slice(&response[112..128]); // low part
+        let blob = Repr::new(blob_bytes.rt()?);
+
+        // Next Felt is size (0x1af25)
```
miraclx: why is this the case? `0x1af25`?

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +42,69 @@ impl Method<Starknet> for ApplicationRequest {
...
+            let chunk = &response[i..];
+            if chunk.iter().take(32).all(|&b| b == 0) {
+                break;
+            }
+            source_bytes.extend(chunk.iter().take(32).filter(|&&b| b != 0));
+            i += 32;
+        }
+        let source = ApplicationSource(String::from_utf8(source_bytes)?.into());
+
+        // Find metadata after source string (look for 0.0.1)
```
miraclx: any alternative to hardcoding this in?

crates/context/config/src/client/env/config/query/proxy_contract.rs
```diff
@@ -31,10 +32,35 @@ impl Method<Starknet> for ProxyContractRequest {
...
+            return Err(eyre::eyre!("No proxy contract found"));
+        }
+
+        // Check if it's a None response (single zero Felt)
+        if response.iter().all(|&x| x == 0) {
+            return Err(eyre::eyre!("No proxy contract found"));
+        }
+
+        // Convert the Felt to a hex string representing the contract address
+        let hex_string = format!("0x{}", hex::encode(&response));
```
miraclx: alternatively, we can parse to `Felt`, and then `format!("{:x}", __)` on that

crates/context/config/src/client/env/config/types/starknet.rs
miraclx, line 1: I'd really like to see less logic duplication. We can take a leaf from config/types.rs.

https://github.com/calimero-network/core/blob/8b26910ebb39d24c22e694ebed8847ba5c7923f3/crates/context/config/src/types.rs#L84

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
...
+        let (high_bytes, low_bytes) = bytes.split_at(bytes.len() / 2);
+        ContextId {
+            high: Felt::from_bytes_be_slice(high_bytes),
+            low: Felt::from_bytes_be_slice(low_bytes),
+        }
+    }
+}
+
+// Add this implementation for Repr wrapped type
+impl From<Repr<crate::ContextId>> for ContextId {
```
miraclx: `Repr<T>: Deref<Target = T>`, so calling `.into()` on a `Repr<ContextId>` will convert to this `ContextId` as well.

So, the `From<Repr<T>>` impls are not needed.

And in places where you're not using `.into`, and instead opt for `T::from`, you'd have to use `T::from(*__)`

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
+use hex;
+use starknet::core::codec::{Encode, Error, FeltWriter};
+use starknet::core::types::Felt;
+
+use crate::repr::{Repr, ReprBytes};
+use crate::types::SignerId;
+
+#[derive(Debug, Clone, Encode)]
```
miraclx: impl `Copy` for all types entirely composed of primitives, like here, it's all just `Felt`s, which is essentially a fixed-size array. So it's `Copy`-able

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
...
+
+#[derive(Debug, Encode)]
+pub struct Signed {
+    pub payload: Vec<Felt>,
+    pub signature_r: Felt,
+    pub signature_s: Felt,
+}
+
+impl<T> From<crate::types::Signed<T>> for Signed {
+    fn from(value: crate::types::Signed<T>) -> Self {
```
miraclx: this should not be necessary right? we never construct a `Signed<T>` in the path to SN

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
...
+            .payload
+            .into_inner()
+            .chunks_exact(32)
+            .map(|chunk| {
+                let chunk_array: [u8; 32] = chunk.try_into().expect("chunk should be 32 bytes");
+                Felt::from_bytes_be(&chunk_array)
+            })
+            .collect();
+        // Extract r and s from the signature
+        let sig_bytes = value.signature.as_bytes();
```
miraclx: this, also, would be an ed25519 signature, so it's incompatible.

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
...
+
+#[derive(Debug, Encode)]
+pub struct Request {
+    pub kind: RequestKind,
+    pub signer_id: ContextIdentity,
+    pub user_id: ContextIdentity,
+    pub nonce: u64,
+}
+
+impl From<crate::Request<'_>> for Request {
```
miraclx: this also is not used

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct EncodableString(pub String);
+
+impl From<crate::types::ApplicationSource<'_>> for EncodableString {
+    fn from(value: crate::types::ApplicationSource<'_>) -> Self {
+        EncodableString(value.0.into_owned())
+    }
+}
```
miraclx, line 334: this shouldn't need to be a `String`, we can't guarantee it's utf-8, it's pretty much just arbitrary bytes.

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,484 @@
...
+
+        // Write number of full words
+        writer.write(Felt::from(full_words_count));
+
+        // Write full words (31 chars each)
+        for i in 0..full_words_count {
+            let start = i * WORD_SIZE;
+            let word_bytes = &bytes[start..start + WORD_SIZE];
+            let word_hex = hex::encode(word_bytes);
+            writer.write(Felt::from_hex(&format!("0x{}", word_hex)).unwrap());
```
miraclx: why `bytes -> hex -> felt` when we can `bytes -> felt`? is it because it's 31 bytes? does it work if we prepend or append a zeroed 32nd byte?

crates/context/config/src/client/env/proxy/query/proposals.rs
```diff
@@ -31,10 +34,74 @@ impl Method<Starknet> for ProposalsRequest {
...
+                4 => {
+                    // SetContextValue
+                    let key_len = felts[offset + 5].to_bytes_be()[31] as usize;
+                    let value_len = felts[offset + 6 + key_len].to_bytes_be()[31] as usize;
+                    offset + 7 + key_len + value_len // base + key_len + value_len + key + value
+                }
+                _ => return Err(eyre::eyre!("Unknown action variant: {}", variant)),
+            };
+
+            let proposal = StarknetProposal::decode(&felts[offset..proposal_end])
```
miraclx: does the `Decode` impl of `StarknetProposal` require the iterator to be fully exhausted?

if it doesn't why do we need the hardcoded logic for the proposal_end?

if I understand correctly and the issue is about progressively advancing through the input felts, then make an iterator ahead of time (outside the consuming loop) and pass it to `decode` on every iteration

crates/context/config/src/client/protocol/starknet.rs
```diff
@@ -248,7 +250,9 @@ impl Network {
             |result| {
                 Ok(result
                     .into_iter()
-                    .flat_map(|felt| felt.to_bytes_be().to_vec())
+                    // Remove the skip(1) here
+                    .map(|felt| felt.to_bytes_be())
+                    .flatten()
```
miraclx: is it still coming or it's already gone?

crates/context/config/src/client/protocol/starknet.rs
```diff
@@ -312,9 +313,76 @@ impl Network {
...
-        let transaction_hash: Vec<u8> = vec![response.transaction_hash.to_bytes_be()[0]];
-        Ok(transaction_hash)
+        // Process the receipt
+        match receipt.receipt {
+            starknet::core::types::TransactionReceipt::Invoke(invoke_receipt) => {
+                match invoke_receipt.execution_result {
+                    ExecutionResult::Succeeded => {
+                        // Process events and return result
+                        for event in invoke_receipt.events.iter() {
+                            if event.from_address == contract_id {
```
miraclx: again, this event thing, not ideal. did you find an alternative? something along the lines of progressively awaiting the chain of execution until all x-calls are done?

crates/context/config/src/lib.rs
miraclx, line 1: can't think of why these would be needed, are they?

crates/context/config/src/types.rs
miraclx, line 1: or these. I suspect it's for the `SNSigned: From<Signed<T>>`, which as earlier mentioned shouldn't be needed

crates/context/src/lib.rs
miraclx, line 1: development iterations?

crates/context/config/src/lib.rs
alenmestrov: we have need for Clone because of this line:
crate::RequestKind::Context(ctx_req) => RequestKind::Context(ctx_req.to_owned().into())

in crates/context/config/src/client/env/config/types/starknet.rs:144:82


I removed Clone trait from RequestKind

crates/context/src/lib.rs
alenmestrov: reverted back to the original

crates/context/config/src/client/protocol/starknet.rs
alenmestrov: What do you mean? 
The idea was that we wait for receipt to be updated and then check the emitted events. 
This is the transaction lifecycle: https://docs.starknet.io/architecture-and-concepts/network-architecture/transaction-life-cycle/#transaction_status

crates/context/config/src/client/protocol/starknet.rs
alenmestrov: updated

crates/context/config/src/client/env/proxy/query/proposals.rs
alenmestrov: updated

crates/context/config/src/client/env/config/types/starknet.rs
alenmestrov: hmm, I'm a little confused here to be honest, I thought the idea was for metadata and source to be "long Strings" in terms of types for Starknet since that's the only type that can fit, that is why this was implemented like this.

crates/context/config/src/client/env/config/types/starknet.rs
alenmestrov: yes, the thing was that if we have 32bytes it will overflow, same like we had for other variables that needed to be split into two parts. Since there is no decode function out-of-the-box for ByteArray/String type I needed to introduce this.

crates/context/config/src/lib.rs
miraclx: > we have need for Clone because of this line

hasn't that been removed?

crates/context/config/src/client/protocol/starknet.rs
miraclx: so this transport will be used for more than just the proxy contract, but you're exclusively decoding `StarknetProposalWithApprovals`, which won't always be valid.

crates/context/config/src/client/env/config/types/starknet.rs
miraclx: so.. application source is a string, metadata, isn't. but on the starknet side, it's pretty much the same, dynamic array of felts.

but on the rust side, metadata is `Vec<u8>`

crates/context/config/src/client/protocol/starknet.rs
alenmestrov: Yes, added checks for this specific event

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -56,28 +62,75 @@ impl<'a> Method<Near> for Mutate<'a> {
...
+        let user_id = user_key_bytes
+            .rt::<ContextIdentity>()
+            .map_err(|e| eyre::eyre!("Failed to convert user key: {}", e))?;
+        let user_id = Repr::new(user_id);
+
+        // Create the Request structure using into() conversions
+        let request = StarknetRequest {
+            signer_id: signer_id.into(),
+            user_id: user_id.into(),
+            nonce: 0,
```
miraclx: `self.nonce`?

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -56,28 +62,75 @@ impl<'a> Method<Near> for Mutate<'a> {
...
 
-impl<'a, T: Transport> ContextConfigMutateRequest<'a, T> {
+impl<'a, T: Transport + Debug> ContextConfigMutateRequest<'a, T> {
     pub async fn send(self, signing_key: [u8; 32]) -> Result<(), ClientError<T>> {
         let request = Mutate {
             signing_key,
+            nonce: 0,
             // todo! when nonces are implemented in context
             // todo! config contract, we fetch it here first
             // nonce: _,
```
miraclx:
```suggestion
            // todo! when nonces are implemented in context
            // todo! config contract, we fetch it here first
            nonce: 0,
```

SN contract doesn't validate against incremental nonces?

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -0,0 +1,545 @@
...
+
+#[derive(Debug, Clone, Copy, Encode, Decode)]
+pub struct ApplicationId(FeltPair);
+
+#[derive(Debug, Clone, Copy, Encode, Decode)]
+pub struct ApplicationBlob(FeltPair);
+
+// Single conversion trait
+pub trait IntoFeltPair {
+    fn into_felt_pair(self) -> (Felt, Felt);
```
miraclx, line 32: for every `crate::types::{T}` here, you can impl `From<{T}>` for `FeltPair`, should come down to less code overall

the for the types here, it'll be for example

```rust
impl From<crate::types::ContextId> for ContextId {
    fn from(value: crate::types::ContextId) -> Self {
        Self(value.into())
    }
}
```

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +44,45 @@ impl Method<Starknet> for ApplicationRequest {
...
+        let bytes = self.context_id.as_bytes();
+        let mid_point = bytes
+            .len()
+            .checked_div(2)
+            .ok_or_else(|| eyre::eyre!("Length should be even"))?;
+        let (high_bytes, low_bytes) = bytes.split_at(mid_point);
+
+        // Convert to Felts
+        let high_felt = Felt::from_bytes_be_slice(high_bytes);
+        let low_felt = Felt::from_bytes_be_slice(low_bytes);
```
miraclx: this "could've" used `FeltPair` conversion

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +44,45 @@ impl Method<Starknet> for ApplicationRequest {
...
+
+        // Convert both Felts to bytes and concatenate
+        let mut result = Vec::new();
+        result.extend_from_slice(&high_felt.to_bytes_be());
+        result.extend_from_slice(&low_felt.to_bytes_be());
+        Ok(result)
```
miraclx: and here, `FeltPair::encode`

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -41,10 +44,45 @@ impl Method<Starknet> for ApplicationRequest {
...
-        todo!()
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        if response.is_empty() {
+            return Err(eyre::eyre!("No application found"));
+        }
+
+        // Convert bytes to Felts
+        let mut felts = Vec::new();
+        for chunk in response.chunks(32) {
+            if chunk.len() == 32 {
```
miraclx: `chunk_exact`? and if there are any chunks that aren't 32 bytes, what does that mean? is it safe to ignore?

crates/context/config/src/client/env/config/query/application_revision.rs
```diff
@@ -31,10 +32,34 @@ impl Method<Starknet> for ApplicationRevisionRequest {
...
+        let (high_bytes, low_bytes) = bytes.split_at(mid_point);
+
+        // Convert to Felts
+        let high_felt = Felt::from_bytes_be_slice(high_bytes);
+        let low_felt = Felt::from_bytes_be_slice(low_bytes);
+
+        // Convert both Felts to bytes and concatenate
+        let mut result = Vec::new();
+        result.extend_from_slice(&high_felt.to_bytes_be());
+        result.extend_from_slice(&low_felt.to_bytes_be());
```
miraclx: `FeltPair`

crates/context/config/src/client/env/config/query/application_revision.rs
```diff
@@ -31,10 +32,34 @@ impl Method<Starknet> for ApplicationRevisionRequest {
...
 
-    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        if response.is_empty() {
+            return Err(eyre::eyre!("Empty response"));
+        }
+
+        // Response should be a single u64 in the last 8 bytes of a felt
+        let revision_bytes = &response[24..32]; // Take last 8 bytes
```
miraclx: should we check we have 32 instead

```suggestion
        if response.len() != 32 {
            return Err(eyre::eyre!("Empty response"));
        }

        // Response should be a single u64 in the last 8 bytes of a felt
        let revision_bytes = &response[24..32]; // Take last 8 bytes
```

because of the index which may panic

crates/context/config/src/client/env/config/query/has_member.rs
```diff
@@ -32,10 +33,50 @@ impl Method<Starknet> for HasMemberRequest {
...
+            .len()
+            .checked_div(2)
+            .ok_or_else(|| eyre::eyre!("Length should be even"))?;
+        let (identity_high, identity_low) = identity_bytes.split_at(mid_point);
+
+        // Convert to Felts and add to result
+        let identity_high_felt = Felt::from_bytes_be_slice(identity_high);
+        let identity_low_felt = Felt::from_bytes_be_slice(identity_low);
+        result.extend_from_slice(&identity_high_felt.to_bytes_be());
+        result.extend_from_slice(&identity_low_felt.to_bytes_be());
```
miraclx: `FeltPair`

crates/context/config/src/client/env/config/query/members.rs
```diff
@@ -1,18 +1,21 @@
...
 
 #[derive(Copy, Clone, Debug, Serialize)]
-pub(super) struct MembersRequest {
-    pub(super) context_id: Repr<ContextId>,
-    pub(super) offset: usize,
-    pub(super) length: usize,
+pub struct MembersRequest {
+    pub context_id: Repr<ContextId>,
+    pub offset: usize,
+    pub length: usize,
```
miraclx:
```suggestion
pub(crate) struct MembersRequest {
    pub(crate) context_id: Repr<ContextId>,
    pub(crate) offset: usize,
    pub(crate) length: usize,
```

crates/context/config/src/client/env/config/query/members_revision.rs
```diff
@@ -31,10 +32,35 @@ impl Method<Starknet> for MembersRevisionRequest {
...
+        let (high_bytes, low_bytes) = bytes.split_at(mid_point);
+
+        // Convert to Felts
+        let high_felt = Felt::from_bytes_be_slice(high_bytes);
+        let low_felt = Felt::from_bytes_be_slice(low_bytes);
+
+        // Convert both Felts to bytes and concatenate
+        let mut result = Vec::new();
+        result.extend_from_slice(&high_felt.to_bytes_be());
+        result.extend_from_slice(&low_felt.to_bytes_be());
```
miraclx: `FeltPair`

crates/context/config/src/client/env/config/query/members_revision.rs
```diff
@@ -31,10 +32,35 @@ impl Method<Starknet> for MembersRevisionRequest {
...
     }
 
-    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        if response.is_empty() {
+            return Err(eyre::eyre!("Empty response"));
+        }
+
+        // Response should be a single u64 in the last 8 bytes of a felt
```
miraclx: check if 32

crates/context/config/src/client/env/config/query/privileges.rs
```diff
@@ -63,10 +66,62 @@ impl<'a> Method<Starknet> for PrivilegesRequest<'a> {
...
+        let mid_point = context_bytes
+            .len()
+            .checked_div(2)
+            .ok_or_else(|| eyre::eyre!("Length should be even"))?;
+        let (high_bytes, low_bytes) = context_bytes.split_at(mid_point);
+
+        // Convert to Felts and then to bytes
+        let mut result = Vec::new();
+        result.extend_from_slice(&Felt::from_bytes_be_slice(high_bytes).to_bytes_be());
+        result.extend_from_slice(&Felt::from_bytes_be_slice(low_bytes).to_bytes_be());
```
miraclx: `FeltPair` -- and other cases, idk if there would be more at this point, but I won't point them out anymore to not clutter the review

you can maybe search for `checked_div`

crates/context/config/src/client/env/proxy/query/proposal_approvers.rs
```diff
@@ -44,10 +48,44 @@ impl Method<Starknet> for ProposalApproversRequest {
...
+        let starknet_id = StarknetProposalId(FeltPair {
+            high: Felt::from_bytes_be(&high),
+            low: Felt::from_bytes_be(&low),
+        });
+
+        // Encode exactly as in mutate response
+        let mut encoded = Vec::new();
+        encoded.extend_from_slice(&starknet_id.0.high.to_bytes_be());
+        encoded.extend_from_slice(&starknet_id.0.low.to_bytes_be());
+        Ok(encoded)
```
miraclx: this could've been:

```suggestion
        let starknet_id = StarknetProposalId::from(self.proposal_id);

        let mut encoded = Vec::new();
        starknet_id.encode(&mut encoded);

        Ok(encoded)
```

crates/context/config/src/client/env/proxy/query/proposals.rs
```diff
@@ -31,10 +34,34 @@ impl Method<Starknet> for ProposalsRequest {
...
-        todo!()
+        let mut bytes = Vec::with_capacity(64); // 2 * 32 bytes for two u32 parameters
+
+        // First parameter (offset): pad to 32 bytes
+        bytes.extend_from_slice(&[0; 28]); // 28 bytes of zeros
+        bytes.extend_from_slice(&(self.offset as u32).to_be_bytes()); // 4 bytes of actual value
+
+        // Second parameter (length): pad to 32 bytes
+        bytes.extend_from_slice(&[0; 28]); // 28 bytes of zeros
+        bytes.extend_from_slice(&(self.length as u32).to_be_bytes()); // 4 bytes of actual value
```
miraclx: surely we can leave this to a derive impl of `Encode` on `StarknetProposal`, right?

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+impl From<StarknetProposals> for Vec<Proposal> {
+    fn from(value: StarknetProposals) -> Self {
+        value.proposals.into_iter().map(Into::into).collect()
+    }
+}
+
+// Conversions for StarknetIdentity
+impl From<Repr<SignerId>> for StarknetIdentity {
```
miraclx: `From<Repr<T>>` shouldn't be necessary, looks like we can benefit from `FeltPair` here

crates/context/config/src/lib.rs
miraclx, line 1: did you verify these?

crates/context/src/lib.rs
miraclx, line 1: can we restore these?

crates/primitives/src/context.rs
```diff
@@ -204,7 +204,7 @@ impl Context {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
miraclx: seems unnecessary?

crates/context/config/src/client/env/config/mutate.rs
alenmestrov: It does but its in draft, since we are waiting for NEAR implementation

crates/context/src/lib.rs
alenmestrov: restored

crates/context/config/src/client/env/config/types/starknet.rs
```diff
@@ -543,3 +582,17 @@ impl From<ContextIdentity> for crate::types::ContextIdentity {
...
+        value.into_inner().into()
+    }
+}
+
+// Add From implementation for Repr<ContextIdentity>
+impl From<Repr<crate::types::ContextIdentity>> for FeltPair {
+    fn from(value: Repr<crate::types::ContextIdentity>) -> Self {
+        value.into_inner().into()
+    }
+}
```
miraclx, line 598: https://github.com/calimero-network/core/pull/981#discussion_r1858319020

crates/context/config/src/client/env/proxy/query/proposal.rs
```diff
@@ -35,43 +36,54 @@ impl Method<Starknet> for ProposalRequest {
...
         let starknet_id: StarknetProposalId = self.proposal_id.into();
-
-        // Encode both parts (0 is FeltPair)
-        let mut encoded = Vec::new();
-        encoded.extend_from_slice(&starknet_id.0.high.to_bytes_be());
-        encoded.extend_from_slice(&starknet_id.0.low.to_bytes_be());
-
-        Ok(encoded)
+        let mut call_data = CallData::default();
+        starknet_id.0.encode(&mut call_data)?;
```
miraclx: doesn't `StarknetProposalId: Encode`?

```suggestion
        starknet_id.encode(&mut call_data)?;
```

crates/context/config/src/client/env/config/query/privileges.rs
```diff
@@ -66,59 +67,57 @@ impl<'a> Method<Starknet> for PrivilegesRequest<'a> {
...
-        // Convert to Felts and then to bytes
-        let mut result = Vec::new();
-        result.extend_from_slice(&Felt::from_bytes_be_slice(high_bytes).to_bytes_be());
-        result.extend_from_slice(&Felt::from_bytes_be_slice(low_bytes).to_bytes_be());
-
+        let mut call_data = CallData::default();
+        
+        // Encode context_id
+        let context_pair: FeltPair = self.context_id.into();
+        context_pair.encode(&mut call_data)?;
```
miraclx: why not to `StarknetContextId`?

crates/context/config/src/client/env/config/query/members_revision.rs
```diff
@@ -32,33 +33,27 @@ impl Method<Starknet> for MembersRevisionRequest {
...
-        // Convert both Felts to bytes and concatenate
-        let mut result = Vec::new();
-        result.extend_from_slice(&high_felt.to_bytes_be());
-        result.extend_from_slice(&low_felt.to_bytes_be());
-
-        Ok(result)
+        let felt_pair: FeltPair = self.context_id.into();
+        let mut call_data = CallData::default();
+        felt_pair.encode(&mut call_data)?;
+        Ok(call_data.0)
```
miraclx: same

crates/context/config/src/client/env/proxy/query/proposal_approvals.rs
```diff
@@ -41,22 +42,37 @@ impl Method<Starknet> for ProposalApprovalsRequest {
         let starknet_id: StarknetProposalId = self.proposal_id.into();
 
         // Encode both high and low parts
-        let mut encoded = Vec::new();
-        encoded.extend_from_slice(&starknet_id.0.high.to_bytes_be());
-        encoded.extend_from_slice(&starknet_id.0.low.to_bytes_be());
-
-        Ok(encoded)
+        let mut call_data = CallData::default();
+        starknet_id.0.encode(&mut call_data)?;
```
miraclx: `.0`

miraclx (approved):

crates/context/config/src/client/env/config/types/starknet.rs
alenmestrov: I missed it... 
Updated!

PR #982, opened 2024-11-21
fix(CI): advance after transactions are fully finalized

miraclx:
cross contract calls settled across blocks won't have been resolved by the time `transact()` returns

https://github.com/near/near-workspaces-rs/pull/379


contracts/context-config/tests/sandbox.rs
```diff
@@ -1155,11 +1154,15 @@ async fn test_storage_usage_matches_code_size() -> eyre::Result<()> {
     let expected_log = format!("Context `{}` added", context_id);
     assert!(res.logs().iter().any(|log| log == &expected_log));
 
+    time::sleep(time::Duration::from_secs(1)).await;
```
xilosada: worker.fast_forward?

chefsale (approved):

PR #984, opened 2024-11-26
feat: add install script

xilosada:
# [product] short description

## Summary

Please include a summary of the change and which issue is fixed. Please also
include relevant motivation and context. List any dependencies that are required
for this change.

## Test plan

instructions so we can reproduce. Please also list any relevant details for your
test configuration. Add screenshots or videos for changes in the user-interface.

## Checklist

- [ ] I have added a test plan that prove my fix is effective or that my feature
      works


chefsale (approved):

scripts/install.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
fbozic, line 1: Shellcheck recommends one change:
```
❯ shellcheck ./scripts/install.sh

In ./scripts/install.sh line 52:
  echo 'export PATH="$HOME/.local/bin:$PATH"' >> "$HOME/.bashrc"
       ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

For more information:
  https://www.shellcheck.net/wiki/SC2016 -- Expressions don't expand in singl...
```

.github/workflows/cross-platform-install-test.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test-install:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        arch: [x86_64, aarch64]
```
fbozic, line 17: Nice, didn't know that arm instances have landed in the open-source GA. We probably might want to do the same for the build instead of emulating arm builds on amd.

scripts/install.sh
fbozic: although not 100% sure if he is correct

miraclx: in this case, I think the expected outcome is `$HOME` and `$PATH` gets expanded at runtime, so it seems fine to me

miraclx (approved):
ideally, we eventually incorporate https://github.com/axodotdev/cargo-dist like [near-cli-rs](https://github.com/near/near-cli-rs/releases/tag/v0.16.0)

xilosada:
Will check cargo dist

PR #986, opened 2024-11-26
feat(node): icp integration

frdomovic:
# feat(node): icp integration

## Summary
Complete ICP integration including:
- new protocol in the process of node initialisation (--protocol icp) 
- generic ICP code for query and mutate 
- updated for context config contract query + mutate methods (types encode and decode for ICP)
- updated for proxy contract query + mutate methods (types encode and decode for ICP)


crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -126,7 +128,37 @@ impl<'a> Method<Starknet> for Mutate<'a> {
...
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        if response.is_empty() {
+            // Empty response means success
+            Ok(())
+        } else {
+            // Non-empty response means there was an error message
+            let error_msg = String::from_utf8(response)
+                .map_err(|e| eyre::eyre!("Invalid UTF-8 in error message: {}", e))?;
+            eyre::bail!("{}", error_msg)
+        }
```
miraclx: the error message here would just be the response, without any indication on why it's wrong.

crates/context/config/src/client/env/config/query/members_revision.rs
```diff
@@ -63,3 +66,19 @@ impl Method<Starknet> for MembersRevisionRequest {
...
     }
 }
+
+impl Method<Icp> for MembersRevisionRequest {
+    type Returns = Revision;
+
+    const METHOD: &'static str = "members_revision";
+
+    fn encode(self) -> eyre::Result<Vec<u8>> {
+        let context_id: ICContextId = (*self.context_id).rt()?;
```
miraclx:
```suggestion
        let context_id: ICContextId = self.context_id.rt()?;
```

are these initial derefs necessary?

crates/context/config/src/client/env/config/types/icp.rs
miraclx, line 1: we really shouldn't be duplicating any of these

crates/context/config/src/client/env/proxy/types/icp.rs
miraclx, line 1: same here

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        Self { networks }
+    }
+}
+
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum IcpError {
+    #[error("unsupported protocol `{0}`")]
+    UnsupportedProtocol(String),
```
miraclx:
```suggestion
```

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -0,0 +1,262 @@
...
+                // Now try to decode the actual data
+                match candid::decode_one::<Result<(), String>>(&data) {
+                    Ok(decoded) => {
+                        match decoded {
+                            Ok(()) => Ok(vec![]),                     // Return empty vec for success
+                            Err(err_msg) => Ok(err_msg.into_bytes()), // Return error message as bytes
+                        }
+                    }
+                    Err(e) => {
+                        println!("Failed to decode: {}", e);
```
miraclx: ^

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -0,0 +1,262 @@
...
+        match response {
+            Ok(CallResponse::Response((data, _))) => {
+                // Now try to decode the actual data
+                match candid::decode_one::<Result<(), String>>(&data) {
+                    Ok(decoded) => {
+                        match decoded {
+                            Ok(()) => Ok(vec![]),                     // Return empty vec for success
+                            Err(err_msg) => Ok(err_msg.into_bytes()), // Return error message as bytes
+                        }
+                    }
```
miraclx: this only encodes errors as the response? how about the actual return data, can we make that into `Vec<u8>`, without the type information that ICP typically infuses?

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -0,0 +1,262 @@
...
+                    Err(e) => {
+                        println!("Failed to decode: {}", e);
+                        Ok(e.to_string().into_bytes()) // Return decode error as bytes
+                    }
+                }
+            }
+            Ok(CallResponse::Poll(_)) => Ok("Unexpected polling response".as_bytes().to_vec()),
```
miraclx: this concerns me, should we be checking the status in a loop?

crates/context/config/src/types.rs
miraclx, line 1: you shouldn't need any of the changes here.

anything that currently does should call `.as_bytes()` instead

crates/merod/Cargo.toml
```diff
@@ -14,13 +14,17 @@ camino = { workspace = true, features = ["serde1"] }
...
 dirs.workspace = true
+ed25519-consensus.workspace = true
 eyre.workspace = true
 futures-util.workspace = true
 hex.workspace = true
 libp2p.workspace = true
 multiaddr.workspace = true
 near-crypto.workspace = true
+rand.workspace = true
+ring.workspace = true
```
miraclx: `ring` shouldn't be here, right?

crates/context/config/src/client/env/config/query/members_revision.rs
frdomovic: Works, updated in all cases

crates/context/config/src/client/protocol/icp.rs
frdomovic: Updated

crates/merod/Cargo.toml
frdomovic: Updated

crates/context/config/src/client/protocol/icp.rs
frdomovic: Updated that now returns vec<u8> of data

crates/context/config/src/client/protocol/icp.rs
frdomovic: Removed

crates/context/config/src/client/env/config/mutate.rs
frdomovic: Updated

crates/context/config/src/client/protocol/icp.rs
frdomovic: This is just default behaviour and way of handling responses from their docs

miraclx: what does it mean to get this response though? can you link me to the doc?

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -239,17 +239,14 @@ impl Network {
...
-                match candid::decode_one::<Result<(), String>>(&data) {
-                    Ok(decoded) => {
-                        match decoded {
-                            Ok(()) => Ok(vec![]),                     // Return empty vec for success
-                            Err(err_msg) => Ok(err_msg.into_bytes()), // Return error message as bytes
-                        }
-                    }
+                match candid::decode_one::<Result<Vec<u8>, String>>(&data) {
+                    Ok(decoded) => match decoded {
+                        Ok(return_data) => Ok(return_data),
```
miraclx: @alenmestrov, won't this mean the canister must've returned a `Vec<u8>`?

have we tested context creation? did it work?

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -239,17 +239,14 @@ impl Network {
...
+                        Err(err_msg) => Ok(err_msg.into_bytes()),
+                    },
                     Err(e) => {
                         println!("Failed to decode: {}", e);
-                        Ok(e.to_string().into_bytes()) // Return decode error as bytes
+                        Ok(e.to_string().into_bytes())
```
miraclx: any reason we're encoding the error into the `Ok` variant? why not `Err`?

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -152,8 +152,11 @@ impl<'a> Method<Icp> for Mutate<'a> {
         } else {
             // Non-empty response means there was an error message
             let error_msg = String::from_utf8(response)
-                .map_err(|e| eyre::eyre!("Invalid UTF-8 in error message: {}", e))?;
-            eyre::bail!("{}", error_msg)
+                .map_err(|e| eyre::eyre!("Invalid UTF-8 in the error response: {}", e))?;
+            eyre::bail!(
+                "Error response received from the system: '{}'",
+                error_msg
+            );
```
miraclx: Errors shouldn't come here at all. The transport should've reported the error directly to the client, at which point it will short-circuit and be returned to the user.

crates/context/config/src/client/protocol/icp.rs
frdomovic: https://internetcomputer.org/docs/current/developer-docs/developer-tools/off-chain/agents/rust-agent#example 
https://internetcomputer.org/docs/current/references/ingress-messages#recommendations-and-best-practices-for-secure-update-call-handling
https://internetcomputer.org/docs/current/references/ic-interface-spec#http-async-call-overview

"Related to the asynchronous nature of update calls on the Internet Computer. This response indicates that the call is still being processed and requires further polling to get the final result."

Okay so maybe update this to use ".call_and_wait()" and then remove this option  
cc @alenmestrov ?

crates/context/config/src/client/protocol/icp.rs
frdomovic: Tested context creation, conrext invite, context join

crates/context/config/src/client/protocol/icp.rs
frdomovic: cc @alenmestrov

crates/context/config/src/client/protocol/icp.rs
alenmestrov: sure, try with that

frdomovic: Update to remove polling

crates/context/config/src/client/protocol/icp.rs
frdomovic: update into Err

crates/context/config/src/client/protocol/icp.rs
miraclx: Eh, yeah.. the return values for mutate on context config are null, we should care instead about proxy, which returns `ICProposalWithApprovals`. Try that.

Otherwise, I'd expect this to pretty much look the same as above with query. No need for `decode_one`.

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -242,9 +242,15 @@ impl Network {
                     Ok(return_data) => Ok(return_data),
                     Err(err_msg) => Ok(err_msg.into_bytes()),
```
miraclx: here as well, but imo we shouldn't be decoding here

crates/context/config/src/client/protocol/icp.rs
```diff
@@ -0,0 +1,256 @@
...
+        &self,
+        canister_id: &Principal,
+        method: String,
+        args: Vec<u8>,
+    ) -> Result<Vec<u8>, IcpError> {
+        self.client
+            .fetch_root_key()
+            .await
+            .map_err(|_| IcpError::Custom {
+                operation: ErrorOperation::Query,
```
miraclx:
```suggestion
                operation: ErrorOperation::Mutate,
```

frdomovic: Updated

crates/context/config/src/client/protocol/icp.rs
frdomovic: Updated

crates/context/config/src/client/env/config/mutate.rs
frdomovic: So updated this and used same logic as we have in near

crates/merod/src/cli/relay.rs
miraclx, line 1: this doesn't have the ICP transport for the relayer

miraclx (commented):
rest looks good

miraclx (approved):
lgtm

frdomovic:
@miraclx @alenmestrov reverted proxy query to Principal + updated encoding , tested so merging this PR. 
Screenshot 2024-12-07 at 02 35 07

PR #987, opened 2024-11-26
feat: transport generalization for recursive declaration

petarjuki7:
The goal is to further generalize the external client to support recursive nesting of structures like:

```rust
type All = Either<Relayer, Both<Near, Both<Starknet, Both<InternetComputer, Both<Ethereum, Solana>>>>>
```

or even

```rust
type NearAndStarknet =
    Both<Near, Starknet>;

type EthereumAndSolana =
    Both<Ethereum, Solana>;

type PolkadotAndInternetComputer =
    Both<Polkadot, InternetComputer>;

type NearAndStarknetOrEthereumAndSolana =
    Either<NearAndStarknet, EthereumAndSolana>;

type NearAndStarknetOrEthereumAndSolanaAndPolkadotAndInternetComputer =
    Both<NearAndStarknetOrEthereumAndSolana, PolkadotAndInternetComputer>;

type RelayerOrNearAndStarknetOrEthereumAndSolanaAndPolkadotAndInternetComputer =
    Either<Relayer, NearAndStarknetOrEthereumAndSolanaAndPolkadotAndInternetComputer>;
```

Which means we don't have to expand `Both` to contain more than 2 transports, or add new structures for each new accommodation.


miraclx (approved):
had some thoughts, needed to experiment to be sure they'd work, compiled in #991

miraclx:
don't forget to fix the PR title

PR #989, opened 2024-11-27
fix(near): handle request timeout

miraclx:
`TimeoutError` on `TxStatusError` is [effectively a dud](https://github.com/near/nearcore/blob/29f388f058a6fc908902b79018844c79eb95baa6/chain/rosetta-rpc/src/errors.rs#L52-L54), and from semantic analysis, it's never actually constructed by the node.

We should, instead.. match for the RPC timeout represented as the `408 Request Timeout` status code.

I've retained the check for `TxStatusError::TimeoutError` to be more exhaustive anyway.


frdomovic (approved):
ok

alenmestrov (approved):
LGTM

PR #999, opened 2024-12-03
ci: define file paths on events and fix gh release view args

fbozic:


.github/workflows/rust_binaries_release.yml
```diff
@@ -4,12 +4,20 @@ on:
   push:
     branches:
       - master
-      - 'releases/**'
-      - 'hotfix/**'
```
miraclx, line 8: can we restore this check for `pull_request`?

fbozic: I can't filter the source branch like this, I could implement a step to check if PR + release branch at the begging and skip `build` and `release` steps conditionally.

miraclx: yeah, that's what I meant.

```yml
    if: startsWith(github.head_ref, 'releases/')
```

or something like that, just to maintain the behaviour

fbozic: done

miraclx (approved):

PR #1001, opened 2024-12-04
fix: update docker compose to remove coordinator

MatejVukosav:


MatejVukosav:
Let's create an issue for that.

PR #1012, opened 2024-12-10
Fix  update proxy contracts with delete

alenmestrov:
Fixed queries and parsing of responses for ICP Proxy contract, introduced delete proposal action for all contracts


crates/context/config/src/client/env/proxy/query/active_proposals.rs
```diff
@@ -65,7 +65,7 @@ impl Method<Icp> for ActiveProposalRequest {
     }
 
     fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        let value = Decode!(&response, Self::Returns)?;
-        Ok(value)
+        let value = Decode!(&response, u32)?;
```
frdomovic, line 68: @miraclx its okay for now as we are not hitting limit for u32 -> u16 conversion ?

frdomovic (approved):
lgtm

crates/context/config/src/client/env/proxy/query/active_proposals.rs
miraclx: but why is it `u16` anyway, if all the contracts return `u32`?

crates/context/config/src/client/env/proxy/query/proposal_approvers.rs
```diff
@@ -104,15 +104,17 @@ impl Method<Icp> for ProposalApproversRequest {
...
 
         // safety: `ICRepr<T>` is a transparent wrapper around `T`
         #[expect(
             clippy::transmute_undefined_repr,
             reason = "ICRepr<T> is a transparent wrapper around T"
         )]
         let identities = unsafe {
-            mem::transmute::<Vec<ICRepr<ContextIdentity>>, Vec<ContextIdentity>>(identities)
+            mem::transmute::<Vec<ICRepr<ContextIdentity>>, Vec<ContextIdentity>>(
+                identities.expect("error unwrapping identities"),
```
miraclx: when `None` is a valid return value, it needs to be propagated, not panicked.

```suggestion
        let Some(identities) = Decode!(&response, Option<Vec<ICRepr<ContextIdentity>>>)? else {
            return Ok(None);
        };

        // safety: `ICRepr<T>` is a transparent wrapper around `T`
        #[expect(
            clippy::transmute_undefined_repr,
            reason = "ICRepr<T> is a transparent wrapper around T"
        )]
        let identities = unsafe {
            mem::transmute::<Vec<ICRepr<ContextIdentity>>, Vec<ContextIdentity>>(identities
```

crates/sdk/src/env/ext.rs
```diff
@@ -147,6 +153,14 @@ impl DraftProposal {
         self
     }
 
+    /// Add an action to delete a proposal.
+    #[must_use]
+    pub fn delete_proposal(mut self, proposal_id: ProposalId) -> Self {
```
miraclx: can we just call it `delete` since the path would already be `Self::external().propose().delete(..)`

miraclx (approved):

PR #1019, opened 2024-12-18
feat: optimized starknet arguments conversion

alenmestrov:
Optimized conversion for arguments in CrossFunctionCall action in proposals for Starknet


crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -538,3 +531,38 @@ impl From<(Vec<Felt>, Vec<Felt>)> for ContextStorageEntry {
...
+        serde_json::Value::Array(arr) => {
+            let json_str = serde_json::to_string(&arr)
+                .map_err(|e| anyhow!("Failed to serialize array: {}", e))?;
+            Ok(Felt::from_bytes_be_slice(json_str.as_bytes()))
+        }
+        serde_json::Value::Object(obj) => {
+            let json_str = serde_json::to_string(&obj)
+                .map_err(|e| anyhow!("Failed to serialize object: {}", e))?;
+            Ok(Felt::from_bytes_be_slice(json_str.as_bytes()))
+        }
```
miraclx: Unsure about this bit, looks like we'll need to recursively encode all the values right?

The starknet contracts were making calls to won't deserialize json

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -1,3 +1,4 @@
+use eyre::anyhow;
```
miraclx: use `eyre!` instead, for consistency

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -289,19 +331,26 @@ impl From<Vec<ProposalAction>> for StarknetProposalActionWithArgs {
...
-                }
-            }
+            StarknetProposalActionWithArgs::ExternalFunctionCall(
+                contract,
+                selector,
+                amount,
+                calldata,
+            ) => ProposalAction::ExternalFunctionCall {
+                receiver_id: format!("0x{}", hex::encode(contract.to_bytes_be())),
+                method_name: format!("0x{}", hex::encode(selector.to_bytes_be())),
```
miraclx: these are the same as `.to_string()`, right? just checking

```suggestion
                receiver_id: contract.to_string(),
                method_name: selector.to_string(),
```

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -397,3 +446,143 @@ impl From<StarknetProposalId> for ProposalId {
...
+                    let chunks: Vec<_> = s.as_bytes().chunks(chunk_size).collect();
+
+                    // Write number of chunks first
+                    writer.write(Felt::from(chunks.len()));
+
+                    // Write each chunk as a Felt
+                    for chunk in chunks {
+                        writer.write(Felt::from_bytes_be_slice(chunk));
+                    }
+                }
```
miraclx: just in case we have a string that starts with 0x and isn't a felt

```suggestion
                if s.starts_with("0x") {
                    // Attempt to handle hex string directly as a single Felt
                    if let Ok(felt) = Felt::from_hex(&s) {
                        writer.write(felt);
                        return Ok(());
                    }
                }

                // Regular string - split into chunks
                let chunk_size = 31;
                let chunks: Vec<_> = s.as_bytes().chunks(chunk_size).collect();

                // Write number of chunks first
                writer.write(Felt::from(chunks.len()));

                // Write each chunk as a Felt
                for chunk in chunks {
                    writer.write(Felt::from_bytes_be_slice(chunk));
                }
```

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -397,3 +446,143 @@ impl From<StarknetProposalId> for ProposalId {
...
+                    let chunk_size = 31;
+                    let chunks: Vec<_> = s.as_bytes().chunks(chunk_size).collect();
+
+                    // Write number of chunks first
+                    writer.write(Felt::from(chunks.len()));
+
+                    // Write each chunk as a Felt
+                    for chunk in chunks {
+                        writer.write(Felt::from_bytes_be_slice(chunk));
+                    }
```
miraclx:
```suggestion
                    let chunk_size = 31;
                    let bytes = s.as_bytes();

                    // Write number of chunks first
                    writer.write(Felt::from(bytes.len() / chunk_size));

                    // Write each chunk as a Felt
                    for chunk in bytes.chunks(chunk_size) {
                        writer.write(Felt::from_bytes_be_slice(chunk));
                    }
```

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -397,3 +446,143 @@ impl From<StarknetProposalId> for ProposalId {
...
+                    // For other numbers, convert to string and encode as string
+                    let s = n.to_string();
+                    ValueCodec(&serde_json::Value::String(s)).encode(writer)?;
+                }
+                Ok(())
+            }
+            serde_json::Value::Object(obj) => {
+                writer.write(Felt::from(obj.len()));
+                for (key, value) in obj {
+                    ValueCodec(&serde_json::Value::String(key.clone())).encode(writer)?;
```
miraclx: to avoid the clone here, you can move the logic for the `String` case into a function which you can call here.

signature can be something like `fn(&str, &mut Vec<Felt>)`, since string encoding is (or rather should be), infallible

miraclx (commented):
everything else looks good

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -289,19 +331,26 @@ impl From<Vec<ProposalAction>> for StarknetProposalActionWithArgs {
...
+                contract,
+                selector,
+                amount,
+                calldata,
+            ) => ProposalAction::ExternalFunctionCall {
+                receiver_id: contract.to_string(),
+                method_name: selector.to_string(),
+                args: calldata
+                    .iter()
+                    .map(|felt| format!("0x{}", hex::encode(felt.to_bytes_be())))
```
miraclx, line 344: to_string, and other cases, where present.

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -397,3 +446,146 @@ impl From<StarknetProposalId> for ProposalId {
...
+                writer.write(Felt::from(*b as u64));
+                Ok(())
+            }
+            serde_json::Value::Number(n) => {
+                if let Some(n) = n.as_u64() {
+                    writer.write(Felt::from(n));
+                } else {
+                    // For other numbers, convert to string and encode as string
+                    let s = n.to_string();
+                    ValueCodec(&serde_json::Value::String(s)).encode(writer)?;
```
miraclx: This will only fail when the number is negative or a float, how are those typically represented in Felt?

Is it first converted to string?

miraclx (approved):
Approving to unblock, but there's one thing I'm unclear about.

alenmestrov:
bump

crates/context/config/src/client/env/proxy/types/starknet.rs
frdomovic: Not sure what exactly should be done here. There is [Cairy Utils](https://cairo-utils-web.vercel.app/) website that shows you can input negative or decimal numbers and get output felt number. 
Conversion in [docs](https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/)

Screenshot 2025-01-07 at 11 24 46
Screenshot 2025-01-07 at 11 24 56
Screenshot 2025-01-07 at 11 25 17
Screenshot 2025-01-07 at 11 25 25

miraclx: Looks like floats aren't a concept in SN, but negative numbers are supported.

So we need one more `else if` for `as_i64`, which can then be converted into a `Felt`

I'm not convinced we'd have any need to account for floats, and I'm tempted to remove the string encoding to avoid implicitly introducing an encoding issue.

When the need arises, we'll figure out what the best accomodation is.

frdomovic:
bump

crates/context/src/lib.rs
```diff
@@ -1421,4 +1421,15 @@ impl ContextManager {
...
         Ok(response)
     }
+
+    pub async fn get_proxy_id(&self, context_id: ContextId) -> EyreResult<String> {
+        let handle = self.store.handle();
+        let Some(context_config) = handle.get(&ContextConfigKey::new(context_id))? else {
+            bail!("Context not found");
+        };
+
+        let proxy_contract = context_config.proxy_contract.as_ref().into();
```
miraclx: here you want to convert `Box<str>` -> `String`

`String` [impls](https://doc.rust-lang.org/std/string/struct.String.html#impl-From%3CBox%3Cstr%3E%3E-for-String) `From<Box<str>>`


```suggestion
        let proxy_contract = context_config.proxy_contract.into();
```

the `.as_ref` would've made it `Box<str>` -> `&str` -> `String`, of which the `&str` -> `String` part would've involved an allocation

miraclx (approved):
lgtm

PR #1021, opened 2024-12-18
feat: add nonce logic to contracts

frdomovic:
# feat: add nonce logic to contracts

## Summary
Add nonce logic , update tests, remove timestamp checks


frdomovic:
tbd - sandbox, config_helper, testing

crates/context/config/src/lib.rs
```diff
@@ -89,18 +89,23 @@ pub enum ContextRequestKind<'a> {
...
+        nonce: u64,
     },
     Grant {
         capabilities: Cow<'a, [(Repr<ContextIdentity>, Capability)]>,
+        nonce: u64,
     },
     Revoke {
         capabilities: Cow<'a, [(Repr<ContextIdentity>, Capability)]>,
+        nonce: u64,
     },
```
miraclx: you can deduplicate these, though.. if you moved it to `ContextRequest`

crates/context/src/lib.rs
```diff
@@ -217,6 +218,23 @@ impl ContextManager {
                 )
             }
 
+            let context_id: IdentityContextId = context.id.rt().expect("infallible conversion");
+            let member_id = identity_secret
+                .public_key()
+                .rt()
+                .expect("infallible conversion");
```
miraclx: you won't need the `IdentityContextId` import if you localized the transmutation to the calling position.

crates/context/src/lib.rs
```diff
@@ -217,6 +218,23 @@ impl ContextManager {
...
+
+            let nonce: u64 = this
+                .config_client
+                .query::<ContextConfigEnv>(
+                    this.client_config.new.protocol.as_str().into(),
+                    this.client_config.new.network.as_str().into(),
+                    this.client_config.new.contract_id.as_str().into(),
+                )
+                .fetch_nonce(context_id, member_id)
```
miraclx: here

crates/context/src/lib.rs
```diff
@@ -217,6 +218,23 @@ impl ContextManager {
...
+                .await?;
+            let nonce = nonce + 1;
```
miraclx: no need to add 1, the contract does that already

crates/context/src/lib.rs
```diff
@@ -928,6 +959,20 @@ impl ContextManager {
                 context_id
             );
         };
+
+        let member_id = signer_id.rt().expect("infallible conversion");
```
miraclx: localize as much of these as possible, to the call position, (so long as it's not reused)

transmutation is mostly useful for specialization, so in most cases, it can be reduced to just the caller

crates/context/src/lib.rs
frdomovic: Updated

crates/context/src/lib.rs
frdomovic: Updated

crates/context/src/lib.rs
frdomovic: Updated

crates/context/config/src/client/env/proxy/types/starknet.rs
```diff
@@ -353,22 +353,20 @@ impl From<StarknetProposalActionWithArgs> for ProposalAction {
...
-                amount,
                 calldata,
+                _,
             ) => ProposalAction::ExternalFunctionCall {
                 receiver_id: format!("0x{}", hex::encode(contract.to_bytes_be())),
                 method_name: format!("0x{}", hex::encode(selector.to_bytes_be())),
                 args: calldata
-                    .iter()
-                    .map(|felt| format!("0x{}", hex::encode(felt.to_bytes_be())))
+                    .0
```
frdomovic: cc @alenmestrov not sure what goes here

frdomovic: Reverted

crates/context/config/src/lib.rs
frdomovic: Updated

contracts/near/context-config/src/mutate.rs
```diff
@@ -41,18 +35,23 @@ impl ContextConfigs {
...
                 }
                 ContextRequestKind::Grant { capabilities } => {
+                    self.check_and_increment_nonce(&request.nonce, &request.signer_id, context_id);
                     self.grant(&request.signer_id, context_id, capabilities.into_owned());
                 }
                 ContextRequestKind::Revoke { capabilities } => {
+                    self.check_and_increment_nonce(&request.nonce, &request.signer_id, context_id);
                     self.revoke(&request.signer_id, context_id, capabilities.into_owned());
                 }
                 ContextRequestKind::UpdateProxyContract => {
```
miraclx: this is missing `check_and_increment_nonce`

contracts/near/context-config/src/mutate.rs
```diff
@@ -183,7 +202,9 @@ impl ContextConfigs {
         for member in members {
             env::log_str(&format!("Added `{member}` as a member of `{context_id}`"));
 
-            let _ = ctx_members.insert(*member);
+            let _ignored = context.member_nonces.insert(*member, 0);
```
miraclx: this will then be a way to "reset the nonce" back to 0, we don't want to insert here for an existing member.

contracts/near/context-config/src/mutate.rs
```diff
@@ -205,7 +226,7 @@ impl ContextConfigs {
             .get_mut();
 
         for member in members {
-            let _ = ctx_members.remove(&member);
+            let _ignored = ctx_members.remove(&member);
             let member = member.rt().expect("infallible conversion");
```
miraclx, line 237: we want to remove the member from the nonce map too

miraclx (approved):

contracts/near/context-config/src/lib.rs
```diff
@@ -37,6 +37,7 @@ struct Config {
 struct Context {
     pub application: Guard<Application<'static>>,
     pub members: Guard<IterableSet<ContextIdentity>>,
+    pub member_nonces: IterableMap<ContextIdentity, u64>,
     pub proxy: Guard<AccountId>,
 }
```
miraclx, line 42: this requires a migration

frdomovic:
bump

PR #1025, opened 2024-12-19
chore: update merod command instructions

MatejVukosav:


crates/merod/src/cli.rs
```diff
@@ -16,17 +16,22 @@ use relay::RelayCommand;
...
-  $ merod --home data/ --node-name node1 config --server-host 143.34.182.202 --server-port 3000
+  $ merod --node-name node1 config --server-host 143.34.182.202 --server-port 3000
+
+  # Run a node
+  $ merod --node-name node1 run
 
-  # Run a node as a peer
-  $ merod --home data/ --node-name node1 run
+  # Relay incoming requests to specific location
+  $ merod --node-name node2 relay --listen 127.0.0.1:3333
```
miraclx: we shouldn't need this, at least not in this form. this is a super specialized use-case that the majority won't need to care about.

MatejVukosav: Which form do you suggest?

miraclx: I mean to say, we don't need examples of `merod relay` in this report. Unless it's hosted by `merod relay --help` itself.

MatejVukosav:
bump

crates/merod/src/cli.rs
```diff
@@ -16,17 +16,19 @@ use relay::RelayCommand;
...
   $ merod --home data/ --node-name node1 init
 
   # Configure an existing node
-  $ merod --home data/ --node-name node1 config --server-host 143.34.182.202 --server-port 3000
+  $ merod --node-name node1 config --server-host 143.34.182.202 --server-port 3000
 
-  # Run a node as a peer
-  $ merod --home data/ --node-name node1 run
+  # Run a node
+  $ merod --node-name node1 run
```
miraclx: There's an extra line here

miraclx (approved):

frdomovic:
Whats left for this PR to be merged? cc @miraclx @MatejVukosav

PR #1030, opened 2024-12-20
feat(node): set ICP init config

frdomovic:
# feat: set ICP init config

## Summary
Set default values for ICP as local network and self signer. This is because we decided not to go with the relayer solution for now

Screenshot 2024-12-20 at 15 38 34


alenmestrov (approved):

miraclx (approved):
lgtm so long as `http://127.0.0.1:4943` is the rpc url for local dfx.

PR #1031, opened 2024-12-20
Feat introduced  icp nonces for context contract

alenmestrov:
Introduced nonce logic for mutate requests in icp context contract.

Optimized context contract for less spending.
Optimized proxy contract for cross-contract calls that include deposit


contracts/icp/context-config/src/mutate.rs
```diff
@@ -356,3 +372,19 @@ async fn update_proxy_contract(
...
     Ok(())
 }
+
+fn check_and_increment_nonce(
+    context: &mut Context,
+    nonce: u64,
+    signer_id: &SignerId,
+) -> Result<(), String> {
+    let context_identity = signer_id.rt().expect("infallible conversion");
+    let current_nonce = *context.member_nonces.get(&context_identity).unwrap_or(&0);
```
miraclx: `get_mut`

contracts/icp/context-config/src/lib.rs
```diff
@@ -23,6 +23,7 @@ pub struct Context {
     pub application: Guard<ICApplication>,
     pub members: Guard<BTreeSet<ICRepr<ContextIdentity>>>,
     pub proxy: Guard<Principal>,
+    pub member_nonces: BTreeMap<ICRepr<ContextIdentity>, u64>,
```
miraclx, line 26: can merge `members` and `member_nonces`?

contracts/icp/context-config/src/mutate.rs
```diff
@@ -169,22 +167,25 @@ fn update_application(
...
+        check_and_increment_nonce(context, nonce, signer_id)?;
+
+        // Rest of the function...
         let guard_ref = context.members.get(signer_id).map_err(|e| e.to_string())?;
         let mut ctx_members = guard_ref.get_mut();
 
-        // Add each member
         for member in members {
             ctx_members.insert(member);
+            context.member_nonces.insert(member, 0);
```
miraclx: if the member existed before, we want to skip

```suggestion
            let _ignored = context.member_nonces.entry(member).or_default();
```

contracts/icp/context-config/src/mutate.rs
```diff
@@ -194,15 +195,18 @@ fn add_members(
 fn remove_members(
     signer_id: &SignerId,
     context_id: &ContextId,
+    nonce: u64,
```
miraclx: this should remove the member from the other map (if you're using another map)

contracts/icp/context-config/src/mutate.rs
```diff
@@ -356,3 +372,19 @@ async fn update_proxy_contract(
...
+    signer_id: &SignerId,
+) -> Result<(), String> {
+    let context_identity = signer_id.rt().expect("infallible conversion");
+    let current_nonce = *context.member_nonces.get(&context_identity).unwrap_or(&0);
+
+    if current_nonce != nonce {
+        return Err("invalid nonce".into());
+    }
+
+    context.member_nonces.insert(context_identity, nonce + 1);
```
miraclx:
```suggestion
    *current_nonce += 1;
```

miraclx (commented):
we can deduplicate the nonce increment, check the near impl

contracts/icp/context-proxy/src/mutate.rs
```diff
@@ -106,13 +106,24 @@ async fn execute_proposal(proposal_id: &ProposalId) -> Result<(), String> {
                         created_at_time: None,
                     };
 
-                    let _: (Result<u64, TransferError>,) =
+                    // Call transfer and explicitly handle the result
+                    let transfer_result: (Result<u64, TransferError>,) =
                         ic_cdk::call(Principal::from(ledger_id), "transfer", (transfer_args,))
                             .await
-                            .map_err(|e| format!("Transfer failed: {:?}", e))?;
+                            .map_err(|e| format!("Transfer call failed: {:?}", e))?;
```
miraclx: ah, I thought this was meant to be approve?

contracts/icp/context-config/src/mutate.rs
```diff
@@ -169,22 +163,26 @@ fn update_application(
...
+
+        // Rest of the function...
         let guard_ref = context.members.get(signer_id).map_err(|e| e.to_string())?;
         let mut ctx_members = guard_ref.get_mut();
 
-        // Add each member
         for member in members {
-            ctx_members.insert(member);
+            if !ctx_members.contains_key(&member) {
+                ctx_members.insert(member, 0); // Only insert if member doesn't exist
```
miraclx: `contains_key` will do a lookup, and `insert` will do another (to return the old value which doesn't exist)

`entry(..).or_default()` works too, and avoids that situation.

contracts/icp/context-config/src/mutate.rs
```diff
@@ -356,3 +359,23 @@ async fn update_proxy_contract(
...
+        .member_nonces
+        .get(&context_identity)
+        .copied()
+        .unwrap_or(0);
+
+    if current_nonce != nonce {
+        return Err("invalid nonce".into());
+    }
+
+    context.member_nonces.insert(context_identity, nonce + 1);
```
miraclx: Use `get_mut` instead of reinserting

miraclx (approved):

contracts/icp/context-config/src/query.rs
```diff
@@ -134,3 +134,14 @@ fn privileges(
...
 }
+
+#[ic_cdk::query]
+fn fetch_nonce(context_id: ICRepr<ContextId>, member_id: ICRepr<ContextIdentity>) -> Option<u64> {
+    with_state(|configs| {
+        configs
+            .contexts
+            .get(&context_id)
+            .and_then(|context| context.member_nonces.get(&member_id))
+            .copied()
```
miraclx, line 145: for consistency, we should probably panic if the context does not exist

same as we have it above

miraclx:
And.. migrations

contracts/icp/context-config/src/query.rs
miraclx: but should we be panicking in ICP at all? how do they handle that? since the state is in-memory, do we lose data?

contracts/icp/context-proxy/mock/external/src/lib.rs
```diff
@@ -1,20 +1,76 @@
...
+        amount: Tokens::from_e8s(100_000_000), // Example amount, in practice this would be parsed from args
+        fee: Tokens::from_e8s(10_000),
+        from_subaccount: Some(Subaccount::from(caller)),
+        to: AccountIdentifier::new(&self_id, &Subaccount([0; 32])),
+        created_at_time: None,
+    };
+
+    // Execute the transfer with proper type annotations
+    let transfer_result: Result<(Result<u64, TransferError>,), _> =
+        ic_cdk::call(ledger_id, "transfer", (transfer_args,)).await;
```
miraclx, line 37: `icrc2`'s `transfer_from`? or does this work too

miraclx (approved):

contracts/icp/context-proxy/mock/external/src/lib.rs
alenmestrov: this works :)

PR #1034, opened 2025-01-09
fix(node+contracts): remove gas, option to update proxy contract, proposal approvals

frdomovic:
# fix(node+contracts): remove gas, option to update proxy contract, proposal approvals

## Summary
In this PR gas field was removed from all structs. This is needed in order to make Blockchain demo application work.
Also proposal approval logic is updated as before it still needed 2 approvals even if threshold was set to 1 because creator approval was not included

#Tests:
Tested with ICP dev net, custom greet smart contract deployed ICP and blockchain demo application to execute cross contract call to custom contract.


miraclx (approved):
lgtm

PR #1037, opened 2025-01-13
fix: updated II login process so that it doesnt listen to MM

alenmestrov:
Updated login process with II to bypass the Metamask and other wallets Extensions messages.


node-ui/src/hooks/useIcp.tsx
```diff
@@ -351,6 +355,7 @@ export function useIcp(): useIcpReturn {
       // Wait for the II response and update the local state
       const responsePromise = new Promise<MessageEvent>((resolve, reject) => {
         const responseHandler = (e: MessageEvent) => {
+          console.log('responseHandler', e);
```
MatejVukosav: Do we need this in the code?

alenmestrov: this was placed just for testing (to see if the messages are passing) after you login and if it works I will remove it before merging.

frdomovic (approved):

PR #1038, opened 2025-01-13
feat: add blockchain integration wasm

frdomovic:
# feat: add blockchain integration wasm

## Summary
As part of ICP tutorial I used wasm from this folder in the process of installation and contract creation.
img15 blockchain demo app

If we want users to have 1:1 with tutorial we can add it here or mention to use it from other repository.

Also to note this is latest blockchain demo application and there shouldn't be more updates until ICP hackaton ends


alenmestrov (commented):
I would remove this from core and directly use demo-blockchain-app repo for this

frdomovic:
Decision is we use mock contract instead.

PR #1043, opened 2025-01-17
fix: switched from zig to gcc compiler and simplified build command

alenmestrov:
Switched linux aarch64 compiler from zig to gcc while maintaining OpenSSL configuration with simplified build command

### DIFF
--- a/.github/workflows/rust_binaries_release.yml
+++ b/.github/workflows/rust_binaries_release.yml
@@ -147,7 +147,6 @@ jobs:
           PKG_CONFIG_SYSROOT_DIR: ${{ env.PKG_CONFIG_SYSROOT_DIR }}
           OPENSSL_STATIC: ${{ env.OPENSSL_STATIC }}
         run: |
-          # Remove cargo-zigbuild
           cargo build ${{ steps.cargo_args.outputs.args }}
 
       - name: Build binaries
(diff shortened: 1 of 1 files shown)


.github/workflows/rust_binaries_release.yml
```diff
@@ -121,31 +133,22 @@ jobs:
...
           OPENSSL_LIB_DIR: ${{ env.OPENSSL_LIB_DIR }}
           OPENSSL_INCLUDE_DIR: ${{ env.OPENSSL_INCLUDE_DIR }}
           PKG_CONFIG_PATH: ${{ env.PKG_CONFIG_PATH }}
           PKG_CONFIG_ALLOW_CROSS: ${{ env.PKG_CONFIG_ALLOW_CROSS }}
           PKG_CONFIG_SYSROOT_DIR: ${{ env.PKG_CONFIG_SYSROOT_DIR }}
           OPENSSL_STATIC: ${{ env.OPENSSL_STATIC }}
-          RUSTFLAGS: '-C link-arg=-lstdc++ -C link-arg=-lpthread -C link-arg=-lc'
-        run: cargo zigbuild ${{ steps.cargo_args.outputs.args }}
+        run: |
+          # Remove cargo-zigbuild
```
fbozic: I think this comment is not needed right?

alenmestrov: updated

PR #1050, opened 2025-01-23
Update meroctl commands

petarjuki7:
# PR short description

Now all the functionalities from the` interactive cli` are included in the `meroctl`.
Updates from this PR include:

1.  `context get identities` (`--owned`) command - list all the identities in a context, the `--owned` command is to show only the owned identities
2. `peers` command - show the number of connected peers

## Test plan

`get identities` tested by creating a context, inviting one other peer and then verifying the output of the `identities` and `identities --owned` commands.
`peers` command tested by creating three nodes, and verifying that the output is `3`

## Documentation update

No documentation update is necessary right now, although when we move form interactive cli to meroctl in our public docs it should be mentioned

### DIFF
--- a/crates/meroctl/src/cli/context/get.rs
+++ b/crates/meroctl/src/cli/context/get.rs
@@ -2,7 +2,7 @@ use calimero_server_primitives::admin::{
     GetContextClientKeysResponse, GetContextIdentitiesResponse, GetContextResponse,
     GetContextStorageResponse, GetContextUsersResponse,
 };
-use clap::{Parser, ValueEnum};
+use clap::Parser;
 use eyre::Result as EyreResult;
 use libp2p::identity::Keypair;
 use libp2p::Multiaddr;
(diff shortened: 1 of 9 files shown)


MatejVukosav:
Can you add a new subcommand in meroctl page:  https://calimero-network.github.io/developer-tools/CLI/meroctl

crates/meroctl/src/cli/context/get.rs
```diff
@@ -17,20 +17,32 @@ use crate::output::Report;
...
     Users,
+
+    #[command(about = "Get client keys")]
     ClientKeys,
+
+    #[command(about = "Get storage information")]
     Storage,
-    Identities,
+
+    #[command(about = "Get identities")]
```
MatejVukosav, line 41: In interactive CLI this action is under `identity` and not under `context`. We should have that unified

miraclx (approved):

crates/meroctl/src/cli/context/get.rs
petarjuki7: Identites are related to contexts, so I think it makes sense for them to be under `context`.
In that sense I believe we should change the `interactive cli`, but also now that the `meroctl` is "complete", I think it's best to remove all mention of the `interactive cli` in our documentation. It can be used by us / core developers for quick "hacking" and testing, but we don't guarantee it's stable-ness.
cc: @miraclx

frdomovic (approved):

crates/meroctl/src/cli/context/get.rs
```diff
@@ -17,20 +17,32 @@ use crate::output::Report;
...
     #[arg(value_name = "CONTEXT_ID", help = "context_id of the context")]
     pub context_id: String,
 }
 
-#[derive(Clone, Debug, ValueEnum)]
-pub enum GetRequest {
+#[derive(Debug, Parser)]
+pub enum GetSubcommand {
+    #[command(about = "Get context details")]
     Context,
```
miraclx, line 30: this will be

```
meroctl context get context --context-id ___
```

something else like `info` is more ideal or even `details`

crates/meroctl/src/cli/context/get.rs
```diff
@@ -17,20 +17,32 @@ use crate::output::Report;
...
+
+    #[command(about = "Get context users")]
     Users,
```
miraclx, line 33: what's the difference between users and identities

miraclx: looks like `/users` doesn't do anything

https://github.com/calimero-network/core/blob/bcf74fb0108a38e2ee46d1927e1771b57e8005a8/crates/server/src/admin/handlers/context/get_context_users.rs#L16

can we remove it please?

PR #1054, opened 2025-01-24
Fix node-ui protocol value options and build node-ui

alenmestrov:
Fixed protocol values and labels in node-ui on context creation.


frdomovic (approved):

miraclx:
Please update the PR name

PR #1058, opened 2025-01-27
feat(meroctl): add named identities

petarjuki7:
# Create named identities

With this PR, when you generate a new identity, you can give it a name.
Then, when joining a `context`, inviting someone to a `context` or executing a `call` to a `context`, you don't have to copy-paste your `PublicKey`, you can just input the name of the identity you want to use.
Also, when creating a context, you can give a name to the identity that will be associated to that context. So also, when sending an invite, you can just name your identity, instead of copy-pasting your `PublicKey`.

## Test plan

Created two nodes, node1 and node2.
Installed an app on node 1.
Created a context on node1 with: 
`cargo run -p meroctl -- --node-name node1 context create -a HpJVYNf564xy7qsG5eXcNTjRfUvovKPj8VeNnBix3bsL --protocol near --as mero`

Created a new identity on node2 with:
`cargo run -p meroctl -- --home data/ --node-name node2 identity generate --identity-name pero
`

Invited node2 to the context I created from node1 with:
`cargo run -p meroctl -- --node-name node1 context invite 9AsZrrbQzRnjEVz9Dj1XsT57zQykRr6ho4hWXDkBmfNs (context_id) 826j3ZLzaLHgj241Po4TxMEvRSKxTqJovHA9Ejf1AvC4 (public key of node2) pero (your identity name)`

Joined the context on node2 with:
` cargo run -p meroctl -- --home data/ --node-name node2 context join s6WhFwWqEQHZpPdvU8Vtoh4sdqtwkycmkxRHzS4eAUqc5GAQzDnB6Z5bNKjeUEKvFXo4spUo2oxoMUyvNJ97S94CmHp87gAaFRiuvqjcAYQyRKdQtSjNV6maN5U2c9DTP6juKHi1YE4hDunU3TqPF4WuBwbWdFB5M (invitation payload) J4jph27MZtdnyHqV7soGAp9UZJUW9Xf8tFMU4CHz4s15 (private key)`

Verified that `node2` has joined the `context`.

## Documentation update

When this is merged, the documentation as a whole is ready for a large overhaul. Removing the mention of the `interactive cli`, and showing the new, simpler workflow with `meroctl`.


petarjuki7:
End-to-end test will probably fail, will need to update.

crates/meroctl/src/cli/context/invite.rs
```diff
@@ -46,12 +59,17 @@ impl InviteCommand {
     pub async fn invite(&self, environment: &Environment) -> EyreResult<ContextInvitationPayload> {
         let config = load_config(&environment.args.home, &environment.args.node_name)?;
 
+        let my_public_key = match self.inviter_id {
+            Some(id) => id,
+            None => open_identity(environment, self.identity_name.as_ref().unwrap())?.public_key,
```
petarjuki7: The `.unwrap(`) should be fine because `clap` guarantees it exists?

miraclx: if we had to do it this way, we'd ideally rewrite it to something like

https://docs.rs/clap/latest/clap/_derive/index.html#arggroup-attributes

https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#argument-relations

crates/meroctl/src/cli/call.rs
```diff
@@ -34,8 +35,20 @@ pub struct CallCommand {
...
+    )]
+    pub executor: Option<PublicKey>,
+
+    #[clap(
+        short = 'i',
+        long,
+        value_name = "IDENTITY_NAME",
+        help = "Name of the identity which you want to use as executor"
+    )]
+    identity_name: Option<String>,
```
miraclx: Hm, I don't see why we can't have a single interface for identity specification.

`--as miraclx` or `--as HBdeL2KPRHyacHWf15QpPVUSMPnJW8wtS3V7HcmfPpJC`

first we should attempt to decode into a `PublicKey` if that fails, we do an alias lookup

crates/meroctl/src/identity.rs
```diff
@@ -0,0 +1,57 @@
...
+use calimero_primitives::identity::{PrivateKey, PublicKey};
+use eyre::{eyre, Result as EyreResult};
+use serde::{Deserialize, Serialize};
+
+use crate::cli::Environment;
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct Identity {
+    pub(crate) public_key: PublicKey,
+    pub(crate) private_key: Option<PrivateKey>,
```
miraclx: why is private key here?

crates/meroctl/src/cli/call.rs
petarjuki7: Makes sense.
Do you think we could do `--as` for the others also, like `context invite` and `context join`?
So it's unified everywhere you need to input either your `PublicKey`, `PrivateKey` or your "name".

crates/meroctl/src/identity.rs
petarjuki7: When you do `context join` you have to input your `PrivateKey` also, so I included it in the struct.

crates/meroctl/src/cli/call.rs
miraclx: > Do you think we could do --as for the others also, like context invite and context join?

yeah, I think so..

crates/meroctl/src/identity.rs
miraclx: but it's useless in the file, we only want to store public key & alias mappings

crates/meroctl/src/identity.rs
miraclx, line 1: does the identity file need to be JSON encoded?

since it's just the public key, that's all we need

petarjuki7:
I'd close this PR now, it served as a good proof of concept for starting work with aliases.

PR #1063, opened 2025-01-29
chore: update bug template

MatejVukosav:
New template for collecting bug info.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,17 +6,35 @@ labels: ''
 assignees: ''
 ---
 
-## Describe the bug
+## Desktop (please complete the following information)
+
+- OS: [e.g. mac M2 Sonoma, Linux Ubuntu 24.10]
```
xilosada: Here you are combining OS(sonoma) and arch(M2), however for ubuntu we don't put arch in the placeholder. Can we make it clear that we need both?

MatejVukosav: Clarified with example instructions

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,17 +6,56 @@ labels: ''
...
+
+Linux:
+
+```bash
+echo "OS Version: $(lsb_release -d | cut -f2)"
+echo "CPU Architecture: $(uname -m)"
+echo "Processor Model: $(grep -m 1 'model name' /proc/cpuinfo | cut -d: -f2-)"
+```
+
+- Browser [e.g. chrome , safari]
```
miraclx: this is optional, right?

miraclx (approved):

.github/ISSUE_TEMPLATE/bug_report.md
MatejVukosav: Yes. Added optional mark

PR #1075, opened 2025-01-30
fix(core): stellar rpc mutate function + types update

frdomovic:
# fix(core): stellar rpc mutate function + types update

Fixing mutate encode function to properly pass byte arguments to context config contract.
Updating stellar types for proper casting from Stellar struct to Calimero types.

## Test plan
Setup node from docs and run following
```
application install file <path>
> application_id
context create <application_id> --protocol stellar
> should get error in different place and its regarding proxy which is WIP in other pr
context ls
> context should be create
```

## Documentation update

-

### DIFF
--- a/crates/context/config/src/client/env/config/mutate.rs
+++ b/crates/context/config/src/client/env/config/mutate.rs
@@ -180,15 +180,12 @@ impl<'a> Method<Stellar> for Mutate<'a> {
         let signer_sk = SigningKey::from_bytes(&self.signing_key);
 
         let request = StellarRequest::new(
-          signer_sk.verifying_key().rt()?,
-          StellarRequestKind::from_with_env(self.kind, &env),
-          self.nonce,
+            signer_sk.verifying_key().rt()?,
+            StellarRequestKind::from_with_env(self.kind, &env),
+            self.nonce,
         );
-        let signed_request = StellarSignedRequest::new(
-            &env,
-            request,
-            |b| Ok(signer_sk.sign(b))
-        ).map_err(|e| eyre::eyre!("Failed to sign request: {:?}", e))?;
+        let signed_request = StellarSignedRequest::new(&env, request, |b| Ok(signer_sk.sign(b)))
+            .map_err(|e| eyre::eyre!("Failed to sign request: {:?}", e))?;
 
         let bytes: Vec<u8> = signed_request.to_xdr(&env).into_iter().collect();
 
--- a/crates/context/config/src/client/env/config/mutate.rs
+++ b/crates/context/config/src/client/env/config/mutate.rs
@@ -2,7 +2,7 @@ use std::fmt::Debug;
 
 use ed25519_dalek::{Signer, SigningKey};
 use soroban_sdk::xdr::ToXdr;
-use soroban_sdk::{Bytes, BytesN, Env, IntoVal};
+use soroban_sdk::Env;
 use starknet::core::codec::Encode as StarknetEncode;
 use starknet::signers::SigningKey as StarknetSigningKey;
 use starknet_crypto::{poseidon_hash_many, Felt};
@@ -17,12 +17,9 @@ use crate::client::transport::Transport;
 use crate::client::{CallClient, ClientError, Operation};
 use crate::icp::types::{ICRequest, ICSigned};
 use crate::repr::{Repr, ReprTransmute};
-use crate::stellar::stellar_types::{
-    StellarApplication, StellarContextRequest, StellarContextRequestKind, StellarRequest,
-    StellarRequestKind, StellarSignedRequest,
-};
+use crate::stellar::stellar_types::{StellarRequest, StellarRequestKind, StellarSignedRequest};
 use crate::types::Signed;
-use crate::{ContextIdentity, ContextRequestKind, Request, RequestKind};
+use crate::{ContextIdentity, Request, RequestKind};
 pub mod methods;
 
 use crate::stellar::stellar_types::FromWithEnv;
@@ -192,8 +189,8 @@ impl<'a> Method<Stellar> for Mutate<'a> {
         Ok(bytes)
     }
 
-    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        Ok(())
     }
 }
 
--- a/crates/context/config/src/client/env/config/mutate.rs
+++ b/crates/context/config/src/client/env/config/mutate.rs
@@ -2,7 +2,7 @@ use std::fmt::Debug;
 
 use ed25519_dalek::{Signer, SigningKey};
 use soroban_sdk::xdr::ToXdr;
-use soroban_sdk::Env;
+use soroban_sdk::{Bytes, BytesN, Env, IntoVal};
 use starknet::core::codec::Encode as StarknetEncode;
 use starknet::signers::SigningKey as StarknetSigningKey;
 use starknet_crypto::{poseidon_hash_many, Felt};
@@ -18,10 +18,11 @@ use crate::client::{CallClient, ClientError, Operation};
 use crate::icp::types::{ICRequest, ICSigned};
 use crate::repr::{Repr, ReprTransmute};
 use crate::stellar::stellar_types::{
-    StellarContextRequest, StellarRequest, StellarRequestKind, StellarSignedRequest,
+    StellarApplication, StellarContextRequest, StellarContextRequestKind, StellarRequest,
+    StellarRequestKind, StellarSignedRequest,
 };
 use crate::types::Signed;
-use crate::{ContextIdentity, Request, RequestKind};
+use crate::{ContextIdentity, ContextRequestKind, Request, RequestKind};
 pub mod methods;
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


crates/context/config/src/stellar/stellar_types.rs
```diff
@@ -60,40 +114,44 @@ pub enum StellarContextRequestKind {
...
             ContextRequestKind::UpdateApplication { application } => {
-                StellarContextRequestKind::UpdateApplication(application.into())
+                StellarContextRequestKind::UpdateApplication(StellarApplication::from_with_env(
+                    application,
+                    env,
+                ))
             }
             ContextRequestKind::AddMembers { members } => {
-                let mut vec = Vec::new(&env);
+                let mut vec = Vec::new(&Env::default());
```
frdomovic: change envDefault with env

crates/context/config/src/stellar/stellar_types.rs
```diff
@@ -60,40 +114,44 @@ pub enum StellarContextRequestKind {
...
                 for member in members.into_owned() {
-                    vec.push_back(BytesN::from_array(&env, &member.as_bytes()));
+                    vec.push_back(BytesN::from_array(&Env::default(), &member.as_bytes()));
```
frdomovic: +

crates/context/src/lib.rs
```diff
@@ -250,6 +250,8 @@ impl ContextManager {
                 .send(*context_secret, 0)
                 .await?;
 
+            println!("Got to proxy contract");
```
frdomovic: remove

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -175,25 +175,22 @@ impl<'a> Method<Stellar> for Mutate<'a> {
...
-        // Convert to bytes using XDR
         let bytes: Vec<u8> = signed_request.to_xdr(&env).into_iter().collect();
 
         Ok(bytes)
     }
 
-    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        Ok(())
```
frdomovic, line 193: Same as impl for Starknet. Add context method doesn't return anything (vec![]) if successful and if there is error it breaks before getting to decode in stellar.rs lines 410+

crates/context/config/src/stellar/stellar_types.rs
```diff
@@ -22,34 +26,83 @@ pub struct StellarApplication {
...
 }
 
+impl<'a> FromWithEnv<Application<'a>> for StellarApplication {
+    fn from_with_env(value: Application<'a>, env: &Env) -> Self {
+        StellarApplication {
+            id: BytesN::from_array(env, &value.id.rt().expect("infallible conversion")),
+            blob: BytesN::from_array(env, &value.blob.rt().expect("infallible conversion")),
+            size: value.size,
+            source: String::from_str(env, &value.source.0.into_owned()),
+            metadata: Bytes::from_slice(env, &value.metadata.0.as_bytes().to_vec()),
```
frdomovic: Updated metadata to casting

crates/context/config/src/stellar/stellar_types.rs
```diff
@@ -22,34 +26,83 @@ pub struct StellarApplication {
...
+        }
+    }
+}
+
+impl<'a> From<StellarApplication> for Application<'a> {
+    fn from(value: StellarApplication) -> Self {
+        // Convert Soroban String to std::String
+        let mut bytes = vec![0u8; value.source.len() as usize];
```
frdomovic, line 43: Couldn't find a way to optimize as soroban string didn't have needed methods and other solution was pretty much same amount of code and no visible improvement code + execution

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -209,12 +212,26 @@ impl Network {
...
+                })?;
+            let val: Val = signed_request
+                .try_into_val(&env)
+                .map_err(|_| StellarError::Custom {
+                    operation: ErrorOperation::Mutate,
+                    reason: "Failed to convert to Val".to_owned(),
+                })?;
+            let sc_val: ScVal = val.try_into_val(&env).map_err(|_| StellarError::Custom {
+                operation: ErrorOperation::Mutate,
+                reason: "Failed to convert to ScVal".to_owned(),
```
miraclx: Looking at the macro expansion of `#[contracttype]`, I noticed this:

```rust
impl TryFrom<&StellarSignedRequest> for soroban_sdk::xdr::ScVal {
    type Error = soroban_sdk::xdr::Error;
    #[inline(always)]
    fn try_from(val: &StellarSignedRequest) -> Result<Self, soroban_sdk::xdr::Error> {
        Ok(soroban_sdk::xdr::ScVal::Map(Some(val.try_into()?)))
    }
}
impl TryFrom<StellarSignedRequest> for soroban_sdk::xdr::ScVal {
    type Error = soroban_sdk::xdr::Error;
    #[inline(always)]
    fn try_from(val: StellarSignedRequest) -> Result<Self, soroban_sdk::xdr::Error> {
        (&val).try_into()
    }
}
```

which means we don't need `StellarSignedRequest` -> `Val` -> `ScVal`, we can go directly with either `request.try_into()` or `ScVal::try_from(request)`

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -280,12 +297,26 @@ impl Network {
...
+                    reason: "Failed to convert to Val".to_owned(),
+                })?;
+            let sc_val: ScVal = val.try_into_val(&env).map_err(|_| StellarError::Custom {
                 operation: ErrorOperation::Mutate,
-                reason: e.to_string(),
+                reason: "Failed to convert to ScVal".to_owned(),
             })?;
-            let scval_bytes = ScVal::Bytes(sc_bytes);
-            Some(vec![scval_bytes])
+            Some(vec![sc_val])
```
miraclx: why do this conversion before either `query` / `mutate` is called, so we don't have to duplicate the logic

You can pass the `ScVal` into `query` & `mutate` instead of `Vec<u8>`

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -295,6 +326,23 @@ impl Network {
...
+        let simulation_result = self
+            .client
+            .simulate_transaction(transaction.clone(), None)
+            .await;
+
+        match simulation_result {
+            Ok(_) => {
+                // Simulation succeeded, continue with execution
+            }
+            Err(e) => {
```
miraclx: `if let Err(err) = simulation_result {}`

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -280,12 +297,26 @@ impl Network {
         let args = if args.is_empty() {
             None
         } else {
-            let sc_bytes = ScBytes::try_from(args).map_err(|e| StellarError::Custom {
+            let env = Env::default();
```
miraclx:
```rust
let mut args = None;

if !args.is_empty() {
    let env = Env::default();
    ...
    args = Some(vec![request.try_into()]);
}
```

crates/context/config/src/stellar/stellar_types.rs
miraclx: `soroban_sdk::String` [impls](https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.String.html#impl-ToString-for-String) `ToString`

crates/context/config/src/stellar/stellar_types.rs
miraclx: this doesn't work?

```suggestion
            metadata: Bytes::from_slice(env, &value.metadata.0),
```

crates/context/config/src/stellar/stellar_types.rs
```diff
@@ -22,34 +26,83 @@ pub struct StellarApplication {
     pub metadata: Bytes,
 }
 
+impl<'a> FromWithEnv<Application<'a>> for StellarApplication {
+    fn from_with_env(value: Application<'a>, env: &Env) -> Self {
+        StellarApplication {
+            id: BytesN::from_array(env, &value.id.rt().expect("infallible conversion")),
+            blob: BytesN::from_array(env, &value.blob.rt().expect("infallible conversion")),
+            size: value.size,
+            source: String::from_str(env, &value.source.0.into_owned()),
```
miraclx:
```suggestion
            source: String::from_str(env, &value.source.0),
```

crates/context/config/src/stellar/stellar_types.rs
```diff
@@ -22,34 +26,83 @@ pub struct StellarApplication {
...
+        let mut bytes = vec![0u8; value.source.len() as usize];
+        value.source.copy_into_slice(&mut bytes);
+        let std_string = std::string::String::from_utf8(bytes).expect("valid utf8");
+
+        Application::new(
+            value.id.rt().expect("infallible conversion"),
+            value.blob.rt().expect("infallible conversion"),
+            value.size,
+            ApplicationSource(Cow::Owned(std_string)),
+            ApplicationMetadata(Repr::new(Cow::Owned(value.metadata.into_iter().collect()))),
```
miraclx: [`to_alloc_vec`](https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Bytes.html#method.to_alloc_vec)?

```suggestion
            ApplicationMetadata(Repr::new(Cow::Owned(value.metadata.to_alloc_vec()))),
```

frdomovic: Works

crates/context/config/src/stellar/stellar_types.rs
frdomovic: works

crates/context/config/src/client/protocol/stellar.rs
frdomovic: updated with different naming as args is name of query function param:
```
async fn query(
        &self,
        contract: &Contracts,
        method: &str,
        args: Vec<u8>,
    ) -> Result<Vec<u8>, StellarError>
```

crates/context/config/src/client/protocol/stellar.rs
frdomovic: Updated

crates/context/config/src/client/protocol/stellar.rs
frdomovic: This as mentioned didn't work.

crates/context/config/src/stellar/stellar_types.rs
frdomovic: Works

crates/context/config/src/stellar/stellar_types.rs
frdomovic: Works

crates/context/config/src/client/protocol/stellar.rs
frdomovic: Updated, thanks for catch!

crates/context/config/src/client/protocol/stellar.rs
frdomovic: Actually updated like this as was extracted outside of query / mutate.

crates/context/config/src/stellar/stellar_types.rs
frdomovic: Actually when updated throw in CI this error so reverted:
Screenshot 2025-01-30 at 14 27 54

alenmestrov (approved):

miraclx (approved):

crates/context/config/src/client/protocol/stellar.rs
miraclx: we needed to activate a feature flag

cleanshot_2025-01-30_at_15 39 04_2x

PR #1084, opened 2025-01-30
fix: release workflow clean changes after creating PR

MatejVukosav:
# [product] short description
Clean state after PR is created for installation scripts so each loop does not use changes from previous loop.

## Test plan
-

## Documentation update
None

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -375,6 +375,7 @@ jobs:
             if [ "${release_required}" == "true" ]; then
               echo "Updating installation script for ${binary}, version: ${version}"
 
+              git checkout master
               target_branch="chore/bump-${binary}-installation-script-to-v${version}"
               git fetch origin "${target_branch}" || true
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -375,6 +375,7 @@ jobs:
             if [ "${release_required}" == "true" ]; then
               echo "Updating installation script for ${binary}, version: ${version}"
 
+              git checkout master
```
miraclx, line 378: aren't we already at master here?

MatejVukosav: Only in the first iteration

miraclx: With https://github.com/calimero-network/core/pull/1089, we'll be releasing them together. We can benefit from using one PR to update both files.

MatejVukosav: Cool, I will close this one as it improves the code that will be deleted.

PR #1087, opened 2025-01-30
feat: query gh api to determine latest version in install script

fbozic:


.github/workflows/test-install-script.yml
```diff
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        arch: [x86_64, aarch64]
```
fbozic, line 17: This was not used, and it can not be tested on managed runners (we would need actual arm runners)

Issue #1091, opened 2025-02-01
fix: add description for `meroctl peers` in meroctl --help`


552020:
<img width="958" alt="Image" src="https://github.com/user-attachments/assets/1a15e5e7-1747-4cb2-8b6b-31e1ec85640e" />

petarjuki7:
Fixed with #1092

PR #1092, opened 2025-02-02
patch(cli): add help to peers command

petarjuki7:
# [product] Help output to the peers command in the `meroctl`

Solves #1091 

## Test plan

Verified the output with ` cargo run -p meroctl -- --home data/ --node-name node1 --help`

## Documentation update

No need.

### DIFF
--- a/crates/meroctl/src/cli/peers.rs
+++ b/crates/meroctl/src/cli/peers.rs
@@ -1,5 +1,6 @@
 use calimero_server_primitives::admin::GetPeersCountResponse;
 use clap::Parser;
+use const_format::concatcp;
 use eyre::Result as EyreResult;
 use reqwest::Client;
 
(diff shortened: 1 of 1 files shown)


MatejVukosav (approved):

MatejVukosav (approved):

xilosada (approved):

552020:
Thanks for fixing this. 

I don't know if the double dash before an option is a convention in the help pages, but since it seems to be a literal command it doesn't work in meroctl this way. I mean the example `meroctl -- --node-name node1 peers`.

image

crates/meroctl/src/cli/peers.rs
```diff
@@ -1,13 +1,24 @@
...
+use const_format::concatcp;
 use eyre::Result as EyreResult;
 use reqwest::Client;
 
 use crate::cli::Environment;
 use crate::common::{do_request, fetch_multiaddr, load_config, multiaddr_to_url, RequestType};
 use crate::output::Report;
 
+pub const EXAMPLES: &str = r"
+  #
```
miraclx, line 12: empty line here

miraclx:
> I don't know if the double dash before an option is a convention in the help pages, but since it seems to be a literal command it doesn't work in meroctl this way. I mean the example `meroctl -- --node-name node1 peers`.

You're right on that, our docs previously showed examples of executing `meroctl` via cargo, which required `--`, since they've been updated to use the binary directly, they're no longer necessary. Thanks for spotting this, the docs will be updated.

crates/meroctl/src/cli/peers.rs
```diff
@@ -1,13 +1,24 @@
...
+  $ meroctl -- --node-name node1 peers
```
chefsale, line 13: shouldn't this be: meroctl --node-name node1 peers

PR #1107, opened 2025-02-11
chore: remove install scripts, set release_name during release upload

fbozic:
Moved install scripts out of the repo: https://github.com/calimero-network/install-sh
Updated docs: https://github.com/calimero-network/calimero-network.github.io/pull/112


.github/workflows/release.yml
```diff
@@ -322,3 +328,5 @@ jobs:
           git push origin "${target_branch}"
 
           gh pr create -f || true
+
+# 0.3.1: Feat Stellar context config contract implementation (#1056)
```
MatejVukosav: This looks like it shouldn't be here

fbozic: oh yeah good point

MatejVukosav (approved):

PR #1110, opened 2025-02-12
Feat e2e stellar

alenmestrov:
Introduced e2e tests for stellar

Implemented `--protocol` flag for running e2e tests with following rules:
Multiple protocols can be defined with flag:
`--protocol near icp`
If protocol flag is set then it will run tests for that/those protocol
If protocol flag is not set then it will run tests for all protocols

Created separate test for Stellar e2e while changing original test to run only NEAR and ICP

## Test plan

- Navigate to `contracts/icp/context-config/`
- Run `deploy_devnet.sh` to start icp devnet
- Navigate to `contracts/stellar/context-config/`
- Run `deploy_devnet_env.sh` to deploy the Stellar devnet
- Copy Output values of CONTRACT_ID, ACCOUNT_ADDRESS, SECRET_KEY
- Update values in `e2e-tests/config/config.json` in Stellar section

- Run command: 
`cargo build -p merod -p meroctl`
`cargo run -p e2e-tests -- --input-dir ./e2e-tests/config --output-dir ./e2e-tests/corpus --merod-binary ./target/debug/merod --meroctl-binary ./target/debug/meroctl`

## Documentation update
There is no documentation update for this PR


.github/workflows/e2e_tests.yml
```diff
@@ -92,6 +92,50 @@ jobs:
...
           cd ../../..
 
+      - name: Deploy Stellar local devnet
+        id: deploy_stellar
+        run: |
+          set -e
+          
+          # Deploy contract and capture output
+          cd ./contracts/stellar/context-config
+          chmod +x ./deploy_devnet_env.sh
```
fbozic: `chmod` changes are reflected in the filesystem and hence included in git as well. run it once locally and remove this line.

contracts/stellar/context-config/deploy_devnet_env.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Invoke the contract to set proxy code
+stellar contract invoke \
+    --id "$CONTRACT_ID" \
+    --source local \
+    --network local \
+    -- \
+    set_proxy_code \
+    --proxy-wasm-file-path "$SCRIPT_DIR/../context-proxy/res/calimero_context_proxy_stellar.wasm" \
+    --owner "$ACCOUNT_ADDRESS"
```
fbozic: ~nit: missing new line

.github/workflows/e2e_tests.yml
```diff
@@ -13,52 +13,76 @@ on:
...
+        with:
+          name: resources
+          path: .
+
+      - name: Set Permissions
         run: |
-          cargo build -p meroctl -p merod -p e2e-tests
+          chmod +x target/debug/e2e-tests target/debug/merod target/debug/meroctl
+          chmod +x apps/kv-store/res/kv_store.wasm
+          chmod +x e2e-tests/config/config.json
```
fbozic: I think file permissions should be preserved, no need to chmod again.

.github/workflows/e2e_tests.yml
```diff
@@ -13,52 +13,76 @@ on:
...
+      - name: Upload Build Artifacts (Binaries)
+        uses: actions/upload-artifact@v4
+        with:
+          name: resources
+          path: |
+            target/debug/meroctl
+            target/debug/merod
+            target/debug/e2e-tests
+            apps/kv-store/res/kv_store.wasm
+            e2e-tests/config/config.json
```
miraclx: this isn't code generated, doesn't need to be part of the artifacts

.github/workflows/e2e_tests.yml
miraclx: we shouldn't need these two

.github/workflows/e2e_tests.yml
```diff
@@ -69,43 +93,91 @@ jobs:
...
         env:
           NO_COLOR: '1'
-          # RUST_LOG: calimero_node=debug,calimero_network=debug
+        continue-on-error: true
         run: |
           export SWARM_HOST=$(ifconfig | grep 'inet ' | grep -v '127.0.0.1' | awk '{print $2}' | head -n 1)
+          echo "Running e2e tests for ${{ matrix.protocol }}"
+          chmod +x ./target/debug/e2e-tests
+          chmod +x ./target/debug/merod
+          chmod +x ./target/debug/meroctl
```
miraclx: you did this earlier with the "set permissions" step

.github/workflows/e2e_tests.yml
```diff
@@ -69,43 +93,91 @@ jobs:
...
+            ./deploy_devnet.sh
+            cd ../../..
+
+            jq --arg icpPort "$ICP_PORT" \
+              '.protocolSandboxes[1].config.rpcUrl = "http://127.0.0.1:\($icpPort)"' \
+              e2e-tests/config/config.json > tmp.json && mv tmp.json e2e-tests/config/config.json
+
+          elif [[ "${{ matrix.protocol }}" == "stellar" ]]; then
+            chmod +x ./contracts/stellar/context-config/deploy_devnet_env.sh
+            chmod +x e2e-tests/config/config.json
```
miraclx: `json` file doesn't need to be executable

.github/workflows/e2e_tests.yml
```diff
@@ -121,20 +193,18 @@ jobs:
             --base master \
             --json number \
             -q '.[0].number')" >> $GITHUB_OUTPUT
-
+      
       - name: Update pull request comment
         if: (success() || failure()) && steps.pr_number.outputs.PR_NUMBER != ''
         uses: thollander/actions-comment-pull-request@v3
         with:
           file-path: ./e2e-tests/corpus/report.md
```
miraclx: Doesn't have to come in this PR, but it would be nice to combine the reports into one comment.

.stellar/identity/alice.toml
miraclx, line 1: do these need to be here? statically defined?

contracts/stellar/context-config/deploy_devnet_env.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+set -e
+
+# Get the directory where the script is located
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Install stellar CLI
+cargo install --locked stellar-cli@22.2.0 --features opt
```
miraclx: this has a noticeable cost on CI runtime, taking 11m 27s to compile

</img>


Get the prebuilt binary instead - https://github.com/stellar/stellar-cli/releases/tag/v22.2.0

.github/workflows/e2e_tests.yml
```diff
@@ -121,20 +193,18 @@ jobs:
             --base master \
             --json number \
             -q '.[0].number')" >> $GITHUB_OUTPUT
-
+      
       - name: Update pull request comment
         if: (success() || failure()) && steps.pr_number.outputs.PR_NUMBER != ''
```
fbozic: We don't need `(success() || failure())` since we are now continuing on an error in the test step.

contracts/stellar/context-config/deploy_devnet_env.sh
fbozic: candid-extractor has impact on CI runtime as well

miraclx: I'd recommend not downloading binaries in `deploy_devnet`. Assume the caller of the script has the dependencies. If they don't let them install it themselves, instead of choosing for them.

And in our case, with CI, we can download the deps in an earlier step.

We can always check if the binary is available like we do for `wasm-opt` in app's `build.sh`:

https://github.com/calimero-network/core/blob/5545d8dd6d39b3d9a796523e10a1c2df27f5bbfa/apps/kv-store/build.sh#L16-L18

alenmestrov: ok, for candid-extractor I added a step that saves it to cache and if the protocol is ICP and the if there is no candid-extractor in cache, then it will install it again, for ICP I removed the x arg

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -260,6 +260,8 @@ impl Network {
             .simulate_transaction(transaction.clone(), None)
             .await;
 
+        println!("simulation_result: {:?}", simulation_result);
```
fbozic: Is this leftover?

fbozic: there are few printlns left around

e2e-tests/src/meroctl.rs
```diff
@@ -68,7 +68,7 @@ impl Meroctl {
                 ],
             )
             .await?;
-
+        println!("json: {:?}", json);
```
fbozic: Is this leftover as well?

contracts/stellar/context-config/deploy_devnet_env.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
```
fbozic: Can we maybe call this `deploy_devnet.sh` to be consistent with the `icp` naming (or update `icp` script name)?

contracts/stellar/context-config/deploy_devnet_env.sh
miraclx: `candid-extractor` [has binaries](https://github.com/dfinity/candid-extractor/releases/tag/0.1.5), we shouldn't need to cache or even build it ourselves.

contracts/stellar/context-config/deploy_devnet_env.sh
```diff
@@ -0,0 +1,106 @@
...
+
+# Add network
+stellar network add local \
+    --rpc-url http://localhost:8000/soroban/rpc \
+    --network-passphrase "Standalone Network ; February 2017"
+
+# Generate and fund keys
+stellar keys generate local --network local --fund
+ACCOUNT_ADDRESS=$(stellar keys address local)
+echo "Account address: $ACCOUNT_ADDRESS"
```
fbozic: Why don't we write output to json file and use jq to parse it afterwards?

crates/context/config/src/client/protocol/stellar.rs
alenmestrov: yes, good catch.

contracts/stellar/context-config/deploy_devnet_env.sh
alenmestrov: We can write it into a JSON file, is there some specific difference that that approach makes?

.github/workflows/e2e_tests.yml
```diff
@@ -182,10 +182,52 @@ jobs:
...
+        run: |
+          mkdir -p reports # Ensure directory exists
+          
+          echo "## E2E Test Results" > combined_report.md
+          echo "" >> combined_report.md
+          echo "### Scenario: kv-store" >> combined_report.md
+          echo "" >> combined_report.md
+          
+          # Add table header
+          echo "| Protocol/Step | 0. app install (AllMembers) | 1. ctx create | 2. call (set, Inviter) | 3. call (get, Inviter) | 4. ctx invite-join | 5. call (get, AllMembers) |" >> combined_report.md
```
miraclx: we need the header/body separator after the header

.github/workflows/e2e_tests.yml
```diff
@@ -69,46 +90,144 @@ jobs:
...
+        if: matrix.protocol == 'icp'
+        uses: actions/cache@v3
+        with:
+          path: ~/.cargo/bin/candid-extractor
+          key: ${{ runner.os }}-candid-extractor
+
+      - name: Install candid-extractor
+        id: install-candid
+        if: steps.cache-candid.outputs.cache-hit != 'true' && matrix.protocol == 'icp'
+        run: cargo install candid-extractor
```
miraclx: any reason we're building `candid-extractor` from source instead of pulling prebuilt binaries?

https://github.com/dfinity/candid-extractor/releases/tag/0.1.5

.github/workflows/e2e_tests.yml
```diff
@@ -69,46 +90,144 @@ jobs:
...
-          echo "ICP_PORT=$ICP_PORT"
-          cargo install candid-extractor
-          cd ./contracts/icp/context-config
-          ./deploy_devnet.sh
-          cd ../../..
+          if [[ "${{ matrix.protocol }}" == "icp" ]]; then
+            echo "ICP_PORT=$ICP_PORT"
+            cd ./contracts/icp/context-config
+            ./deploy_devnet.sh
+            cd ../../..
```
miraclx: Tracking `../../..` is easy to miss when things move around

```suggestion
            pushd ./contracts/icp/context-config
            ./deploy_devnet.sh
            popd
```

contracts/stellar/context-config/deploy_devnet.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
+set -e
+
+# Get the directory where the script is located
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
miraclx: is there anything in this script that needs the CWD where it's called from?

otherwise, you can skip saving `SCRIPT_DIR` if you just did

```
cd "$(dirname $0)"
```

at the start of this file

.github/workflows/e2e_tests.yml
```diff
@@ -121,20 +240,11 @@ jobs:
             --base master \
             --json number \
             -q '.[0].number')" >> $GITHUB_OUTPUT
-
+
```
miraclx:
```suggestion

```

contracts/stellar/context-config/deploy_devnet.sh
```diff
@@ -0,0 +1,106 @@
...
+    
+    echo "Downloading $BINARY_NAME..."
+    wget "https://github.com/stellar/stellar-cli/releases/download/v22.2.0/${BINARY_NAME}"
+    tar xzf "${BINARY_NAME}"
+    chmod +x stellar
+    sudo mv stellar /usr/local/bin/
+    rm "${BINARY_NAME}"
+else
+    echo "stellar is already installed"
+fi
```
miraclx: https://github.com/calimero-network/core/pull/1110#discussion_r1955958423

if it's not installed, print an error and return. if you want to add some generalized way of downloading a supported stellar version, you can add this to `scripts/` dir, and call it in a step in our CI.

Our scripts should not be invasive. Especially that use of `sudo`.

contracts/stellar/context-config/deploy_devnet.sh
```diff
@@ -0,0 +1,106 @@
...
+            echo "Unsupported platform: $OS-$ARCH"
+            exit 1
+            ;;
+    esac
+    
+    echo "Downloading $BINARY_NAME..."
+    wget "https://github.com/stellar/stellar-cli/releases/download/v22.2.0/${BINARY_NAME}"
+    tar xzf "${BINARY_NAME}"
+    chmod +x stellar
+    sudo mv stellar /usr/local/bin/
```
miraclx: we shouldn't require elevated permissions for this. Use `$HOME/.local/bin` instead?

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -329,7 +335,7 @@ impl Network {
                 })
             }
         };
-
+        println!("result: {:?}", result);
```
fbozic: println

crates/context/config/src/client/protocol/stellar.rs
```diff
@@ -290,6 +292,8 @@ impl Network {
                     let status = response.base.status;
                     let start = Instant::now();
 
+                    println!("response {:?}", response.error_result);
```
miraclx: these are still here

contracts/stellar/context-config/deploy_devnet.sh
```diff
@@ -0,0 +1,106 @@
...
+    rm "${BINARY_NAME}"
```
fbozic: I agree with Mira, we should not install tools just check and fail if missing.

e2e-tests/src/driver.rs
```diff
@@ -83,7 +84,27 @@ impl Driver {
...
+                ProtocolSandboxConfig::Near(_) => "near",
+                ProtocolSandboxConfig::Icp(_) => "icp",
+            };
+
+            if !self.environment.protocols.is_empty()
+                && !self
+                    .environment
+                    .protocols
+                    .iter()
+                    .any(|p| p.to_lowercase() == protocol_name)
```
miraclx: [an empty iterator returns false](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any)

```suggestion
            if !self
                .environment
                .protocols
                .iter()
                .any(|p| p.to_lowercase() == protocol_name)
```

contracts/stellar/context-config/deploy_devnet.sh
```diff
@@ -0,0 +1,106 @@
...
+            BINARY_NAME="stellar-cli-22.2.0-aarch64-unknown-linux-gnu.tar.gz"
+            ;;
+        *)
+            echo "Unsupported platform: $OS-$ARCH"
+            exit 1
+            ;;
+    esac
+    
+    echo "Downloading $BINARY_NAME..."
+    wget "https://github.com/stellar/stellar-cli/releases/download/v22.2.0/${BINARY_NAME}"
```
miraclx: let's not download to the CWD

`mktemp -d` && `wget -P $tmpdir`

contracts/stellar/context-config/deploy_devnet_env.sh
alenmestrov: The idea behind dependency installation was that the user sets up everything with one command. 

In the case of ICP, the deploy_devnet.sh script is used for the general setup of environment and usage, while in the case of Stellar, the script is only used for CI since we are using testnet environment in general usage. 

Based on previous experience with hackathon participants I suggest leaving the automatic installation of the dependencies if they are missing, at least in the case of ICP.

contracts/stellar/context-config/deploy_devnet.sh
alenmestrov: The idea behind dependency installation was that the user sets up everything with one command.

In the case of ICP, the deploy_devnet.sh script is used for the general setup of environment and usage, while in the case of Stellar, the script is only used for CI since we are using testnet environment in general usage.

Based on previous experience with hackathon participants I suggest leaving the automatic installation of the dependencies if they are missing, at least in the case of ICP.

I have removed the installation part from the Stellar script.

.github/workflows/e2e_tests.yml
```diff
@@ -115,34 +115,76 @@ jobs:
...
+              *)
+                echo "Unsupported platform"
+                exit 1
+                ;;
+            esac
+            
+            wget -P "$TMPDIR" "https://github.com/dfinity/candid-extractor/releases/download/${VERSION}/${BINARY}"
+            tar xzf "$TMPDIR/$BINARY" -C "$TMPDIR"
+            mkdir -p ~/.cargo/bin
+            mv "$TMPDIR/candid-extractor" ~/.cargo/bin/
```
miraclx: `$HOME/.local/bin`?

PR #1114, opened 2025-02-17
feat: associate type information with alias

miraclx:
Retain type information, even though we're allowing aliases, which should be less error-prone

Some missing pieces, that will come in later PRs

1. `alias list`
2. `use` for a default selected alias
3. `application alias`


crates/store/src/key.rs
```diff
@@ -20,7 +20,7 @@ mod component;
 mod context;
 mod generic;
 
-pub use alias::{Alias, Kind};
+pub use alias::{Alias, Aliasable, DefaultScope as DefaultAliasScope, ScopeIsh, StoreScopeCompat};
```
miraclx: Remove DefaultScope & Scopish re-exports.

crates/server/src/admin/service.rs
```diff
@@ -140,9 +140,7 @@ pub(crate) fn setup(
         .route("/identity/keys", delete(delete_auth_keys_handler))
         .route("/generate-jwt-token", post(generate_jwt_token_handler))
         .route("/peers", get(get_peers_count_handler))
-        .route("/add-alias", post(create_identity_alias::handler))
-        .route("/get-alias", post(get_identity_alias::handler))
-        .route("/remove-alias", post(delete_identity_alias::handler))
+        .nest("/alias", alias::service())
```
miraclx, line 143: Document API changes

crates/primitives/src/alias.rs
```diff
@@ -2,93 +2,147 @@
...
-#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
-pub struct Alias(String);
+use crate::application::ApplicationId;
+use crate::context::ContextId;
+use crate::identity::PublicKey;
 
-#[derive(Clone, Debug, Error)]
-#[error("Invalid alias: {0}")]
-pub struct InvalidAlias(String);
+const MAX_LENGTH: usize = 50;
```
miraclx, line 16: Consider reducing this to 30, which is less than the minimum valid length for base58-encoded 32-byte hashes.

crates/meroctl/src/cli/proxy/get.rs
```diff
@@ -22,11 +23,11 @@ pub struct GetCommand {
...
     pub method: GetRequest,
 
-    #[arg(value_name = "CONTEXT_ID", help = "context_id of the context")]
-    pub context_id: String,
+    #[arg(value_name = "CONTEXT", help = "Context for which to query")]
+    pub context: Alias<ContextId>,
 
     #[arg(value_name = "PROPOSAL_ID", help = "proposal_id of the proposal")]
-    pub proposal_id: String,
+    pub proposal_id: Hash,
```
miraclx, line 30: ProposalId?

petarjuki7 (approved):

chefsale:
Can we create issues and make them potential bounties maybe even for the mentioned follow up PRs?

petarjuki7:
> 3. `context list --context <CONTEXT>` is now `context list <CONTEXT>`

Did you mean `context` **identity** `list --context <CONTEXT>` is now `context` **identity** ` list <CONTEXT>`?

miraclx:
> Can we create issues and make them potential bounties maybe even for the mentioned follow up PRs?

yep!

alenmestrov (approved):

crates/meroctl/src/cli/proxy/get.rs
miraclx: `ProposalId` isn't in `calimero-primitives`, this is fine.

crates/primitives/src/alias.rs
miraclx: Leaving this for now. Will revisit.

crates/node/src/interactive_cli/peers.rs
```diff
@@ -23,11 +22,11 @@ impl PeersCommand {
...
 
-        let context_id: Option<ContextId> = self
-            .context_id
-            .map(|context_inner| resolve_identifier(node, &context_inner, Kind::Context, None))
+        let context_id = self
+            .context
+            .map(|context| node.ctx_manager.resolve_alias(context, None))
             .transpose()?
-            .map(|hash| hash.into());
+            .ok_or_eyre("unable to resolve")?;
```
petarjuki7, line 29: Won't this make it fail if we didn't provide an Alias / ContextId?

crates/context/src/lib.rs
```diff
@@ -1493,21 +1496,116 @@ impl ContextManager {
...
+
+    pub fn resolve_alias<T>(
+        &self,
+        alias: Alias<T>,
+        scope: Option<T::Scope>,
+    ) -> EyreResult<Option<T>>
+    where
+        T: Aliasable<Scope: StoreScopeCompat> + From<Hash> + FromStr<Err: Into<eyre::Report>>,
+    {
+        if let Some(value) = self.lookup_alias(alias, scope)? {
```
petarjuki7, line 1560: Doesn't this last ? return the Error too soon?

crates/node/src/interactive_cli/peers.rs
miraclx: Fixed in https://github.com/calimero-network/core/pull/1128

PR #1126, opened 2025-02-18
fix(meroctl): Fix error handling

petarjuki7:
# Meroctl error handling fix

Regarding last PRs on Aliases, there were some cases of returning `Errors` too soon, which would cause problems if the user didn't provide an alias.
They are now fixed, and also some formating enhancements on `meroctl`.

## Test plan

Tested the `peers` command in interactive cli.
Tested one "round" of creating a context, inviting and joining with two nodes both on the interactive cli and meroctl.

## Documentation update

No need.

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -1557,8 +1557,10 @@ impl ContextManager {
     where
         T: Aliasable<Scope: StoreScopeCompat> + From<Hash> + FromStr<Err: Into<eyre::Report>>,
     {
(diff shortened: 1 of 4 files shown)


alenmestrov (approved):

miraclx (commented):
~I'm not sure I understand this, can you provide an example of such a case where an error is returned too soon that shouldn't have been?~

The parts this patch affects are fine, the problematic bits are smaller scale and are fixed in https://github.com/calimero-network/core/pull/1128.

Issue #482, opened 2024-07-16
Integrate AutoNAT protocol


fbozic:
At the moment we assume all nodes are running in a private network behind NAT. We should integrate the AutoNAT protocol so we can dynamically determine the node's status. If we are running behind the NAT we should request a relay reservation. If we are not running behind the NAT we should register our external addresses at the rendezvous peers.

Our boot nodes already support the server part of AutoNAT protocol: https://github.com/calimero-network/boot-node/blob/83d1f6487f80d9614034fb63d579d8e763cbdb88/src/main.rs#L18. But we should run more than 1 boot node to get a reliable status. 

Linked to: https://github.com/calimero-network/core/issues/486

chefsale:
Can we close this one? @fbozic

PR #1137, opened 2025-03-06
feat(networking): implement autonat protocol

petarjuki7:
# Implement Autonat protocol in networking

Refrence issue: #482 

Node dinamically determines if its public or private, and then either registers itself at a rendezvous node or requests a relay reservation.

## Test plan

TBD

## Documentation update

Maybe mention somewhere in our networking docs that we also use autonat now.


crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,34 @@
...
+                        .update_autonat_confidence(self.swarm.behaviour().autonat.confidence());
+                }
+                libp2p::autonat::OutboundProbeEvent::Error { .. } => {
+                    error!("Outbound probe failed")
+                }
+            },
+            Event::StatusChanged { old, new } => {
+                debug!("NAT status changed from {:?} to {:?}", old, new);
+                // Should we unregister from rendezvous here if we became private?
+                self.discovery.state.update_autonat_status(new);
```
petarjuki7: cc: @miraclx

miraclx: I don't have full coverage of the involved pieces to evaluate this atm

cc: @fbozic

crates/network/src/config.rs
```diff
@@ -111,6 +116,16 @@ impl BootstrapNodes {
...
+
+    #[must_use]
+    pub fn autonat() -> Self {
+        Self {
+            list: AUTONAT_BOOT_NODES
+                .iter()
+                .map(|addr| addr.parse().expect("invalid multiaddr"))
+                .collect(),
+        }
+    }
```
miraclx: if autonat nodes are not configurable, we don't need this function here, either use the `const` directly or separate this helper method from `BootstrapNodes`

`autonat` nodes aren't necessarily boot nodes

crates/network/src/lib.rs
```diff
@@ -100,6 +102,19 @@ fn init(
         peers
     };
 
+    let autonat_bootstrap_peers = {
+        let mut autonat_peers = vec![];
+
+        for mut addr in BootstrapNodes::autonat().list {
```
miraclx: here, although I don't see why we can't make this configurable

if you do make them, use a separate struct

miraclx (commented):
Or perhaps a better way of doing this autonat server connection instead of separating them from the boot nodes would be:

when a peer `identify`-ies itself as speaking the autonat protocol, the same way we do for relay and rendezvous

So by default this applies to the selected boot nodes.

crates/network/src/discovery.rs
```diff
@@ -177,6 +183,28 @@ impl EventLoop {
         Ok(())
     }
 
+    pub(crate) fn rendezvous_unregister(&mut self, rendezvous_peer: &PeerId) -> EyreResult<()> {
```
fbozic, line 188: Can you add docs for this method, all methods in this mod have documentation

petarjuki7:
Will remove the comments from code when the PR is approved, keeping them for now for some reasoning behind certain choices

crates/network/src/discovery/state.rs
```diff
@@ -78,9 +98,31 @@ impl DiscoveryState {
...
+                        Entry::Vacant(entry) => {
+                            let _ = entry.insert(PeerInfo {
+                                addrs: HashSet::default(),
+                                discoveries: HashSet::default(),
+                                relay: None,
+                                rendezvous: None,
+                                autonat: Some(PeerAutonatInfo::default()),
+                            });
+                        }
+                    }
```
fbozic: This looks like it is under `if protocol == &RENDEZVOUS_PROTOCOL_NAME`, I think it should be at the same level as other protocol checks.

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use super::{EventHandler, EventLoop};
+
+impl EventHandler<Event> for EventLoop {
+    async fn handle(&mut self, event: Event) {
+        debug!("{}: {:?}", "autonat".yellow(), event);
+
+        match event {
+            Event::OutboundProbe(outbound_probe_event) => match outbound_probe_event {
+                libp2p::autonat::OutboundProbeEvent::Request { peer, .. } => {
```
fbozic: Shouldn't we import `OutboundProbeEvent` and avoid using a fully qualified name?

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+            _ => {}
+        }
+    }
+}
+
+#[allow(
+    clippy::multiple_inherent_impl,
+    reason = "Currently necessary due to code structure"
+)]
```
fbozic: Let's try to avoid this pattern since it will have to go away with the actor setup anyway. Those two methods are quite small we can probably inline them simply.

crates/network/src/events/identify.rs
```diff
@@ -13,21 +13,35 @@ impl EventHandler<Event> for EventLoop {
...
+                if let Err(err) = self.add_autonat_server(&peer_id) {
                     error!(%err, "Failed to handle relay reservation");
                 };
             }
 
-            if self.discovery.state.is_peer_rendezvous(&peer_id) {
-                if let Err(err) = self.rendezvous_discover(&peer_id) {
-                    error!(%err, "Failed to perform rendezvous discovery");
-                };
+            // this logic was moved to the autonat handler
```
fbozic: I don't think we should move the logic to autonat handler, duplicating logic maybe but we might lose behavior if we just move it, or am I missing something?

crates/network/src/discovery.rs
```diff
@@ -251,4 +282,29 @@ impl EventLoop {
 
         Ok(())
     }
+
+    pub(crate) fn add_autonat_server(&mut self, autonat_peer: &PeerId) -> EyreResult<()> {
```
fbozic, line 288: And docs for this method since it is in documented module

crates/network/src/discovery/state.rs
```diff
@@ -78,10 +98,32 @@ impl DiscoveryState {
...
+                    Entry::Vacant(entry) => {
+                        let _ = entry.insert(PeerInfo {
+                            addrs: HashSet::default(),
+                            discoveries: HashSet::default(),
+                            relay: None,
+                            rendezvous: None,
+                            autonat: Some(PeerAutonatInfo::default()),
+                        });
+                    }
+                }
```
miraclx: all of this should be collapsible to

```suggestion
                let peer_info = self.peers.entry(*peer_id).or_insert_with(Default::default);

                let _ignored = peer_info.autonat.get_or_insert_with(Default::default);
```

the other protocols can benefit from this as well

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,82 @@
...
+    async fn handle(&mut self, event: Event) {
+        debug!("{}: {:?}", "autonat".yellow(), event);
+
+        match event {
+            Event::OutboundProbe(outbound_probe_event) => match outbound_probe_event {
+                OutboundProbeEvent::Request { peer, .. } => {
+                    debug!("Sent probe request to: {:?}", peer)
+                }
+                OutboundProbeEvent::Response { peer, address, .. } => {
+                    debug!("Peer: {} determined external address {}", peer, address);
```
miraclx:
```suggestion
                    debug!(%peer, %address, "Peer determined our external address");
```

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,82 @@
...
+use super::{EventHandler, EventLoop};
+
+impl EventHandler<Event> for EventLoop {
+    async fn handle(&mut self, event: Event) {
+        debug!("{}: {:?}", "autonat".yellow(), event);
+
+        match event {
+            Event::OutboundProbe(outbound_probe_event) => match outbound_probe_event {
+                OutboundProbeEvent::Request { peer, .. } => {
+                    debug!("Sent probe request to: {:?}", peer)
```
miraclx:
```suggestion
                    debug!(%peer, "Sent probe request");
```

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,76 @@
...
+                    debug!("Peer: {} determined external address {}", peer, address);
+
+                    let rendezvous_peers: Vec<PeerId> =
+                        self.discovery.state.get_rendezvous_peer_ids().collect();
+                    let relay_peers: Vec<PeerId> =
+                        self.discovery.state.get_relay_peer_ids().collect();
+
+                    if self.discovery.state.is_autonat_status_public()
+                        && self.swarm.behaviour().autonat.confidence()
+                            >= self.discovery.autonat_config.confidence_threshold
```
miraclx: move this check to L19, short-circuiting

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,76 @@
...
+                            }
+                            if let Err(err) = self.rendezvous_register(peer_id) {
+                                error!(%err, "Failed to register with rendezvous");
+                            }
+                        }
+                    }
+
+                    if self.discovery.state.is_autonat_status_private()
+                        && self.swarm.behaviour().autonat.confidence()
+                            >= self.discovery.autonat_config.confidence_threshold
```
miraclx: so you can remove it from here too

crates/network/src/events/autonat.rs
```diff
@@ -0,0 +1,76 @@
...
+                    }
+                }
+                OutboundProbeEvent::Error { .. } => {
+                    error!("Outbound probe failed")
+                }
+            },
+            Event::StatusChanged { old, new } => {
+                debug!("NAT status changed from {:?} to {:?}", old, new);
+
+                self.discovery.state.update_autonat_status(new.clone());
```
miraclx:
```suggestion
                self.discovery.state.update_autonat_status(new);
```

crates/network/src/discovery.rs
```diff
@@ -251,4 +282,32 @@ impl EventLoop {
...
+        );
+
+        self.swarm
+            .behaviour_mut()
+            .autonat
+            .add_server(*autonat_peer, None);
+
+        self.discovery
+            .state
+            .update_autonat_connection_status(autonat_peer, AutonatServerStatus::Connected);
```
miraclx: is this useful?

crates/network/src/discovery/state.rs
```diff
@@ -147,6 +188,17 @@ impl DiscoveryState {
...
+    pub(crate) fn update_autonat_connection_status(
+        &mut self,
+        autonat_server_peer: &PeerId,
+        status: AutonatServerStatus,
+    ) {
+        let _ = self
+            .peers
+            .entry(*autonat_server_peer)
+            .and_modify(|info| info.update_autonat_server_connection_status(status));
+    }
```
miraclx: do we need to track which peer we've connected with?

crates/network/src/events/autonat.rs
miraclx: Reevaluating this, I'd say yes.

crates/network/src/discovery/state.rs
```diff
@@ -26,12 +26,11 @@ pub struct DiscoveryState {
     relay_index: BTreeSet<PeerId>,
     rendezvous_index: BTreeSet<PeerId>,
     autonat_index: BTreeSet<PeerId>,
-    autonat: AutonatStatus,
 }
 #[derive(Debug)]
 pub struct AutonatStatus {
-    status: NatStatus,
-    last_status_public: bool,
```
miraclx: `autonat` is a discovery mechanism tho, it makes sense to keep it in discovery

petarjuki7: cc: @fbozic

crates/network/src/discovery/state.rs
petarjuki7: Honestly, I think we can remove the whole `PeerAutonatInfo` struct from `PeerInfo`.

Maybe wait for a comment from @fbozic.

crates/network/src/events/autonat.rs
```diff
@@ -12,51 +12,49 @@ impl EventHandler<Event> for EventLoop {
...
+                    if self.swarm.behaviour().autonat.confidence()
+                        >= self.discovery.autonat_config.confidence_threshold
                     {
-                        for peer_id in &rendezvous_peers {
-                            if let Err(err) = self.rendezvous_discover(peer_id) {
-                                error!(%err, "Failed to perform rendezvous discovery");
-                            }
-                            if let Err(err) = self.rendezvous_register(peer_id) {
-                                error!(%err, "Failed to register with rendezvous");
+                        if self.is_autonat_status_public() {
```
miraclx: instead of indenting, short-circuit

```suggestion
                    if self.swarm.behaviour().autonat.confidence()
                        < self.discovery.autonat_config.confidence_threshold
                    {
                        return;
                    }
                    
                    if self.is_autonat_status_public() {
```

crates/network/src/discovery/state.rs
fbozic: I yeah should have dropped a comment instead of chat in dm. So discovery state was intended to track info about other peers not about us. But if you think it makes sense to keep it here we can keep it here as well.

crates/network/src/discovery/state.rs
fbozic: yeah we don't need to track connections from us to the autonat server, we just need to track autonat server info in the discovery state in case we need it ever in the future

crates/network/src/discovery/state.rs
petarjuki7: I'm going to leave this as it is now if no objections, I can change it later if needed

miraclx: > discovery state was intended to track info about other peers not about us

oh, the name's not very indicative of that, as-is it's an umbrella for everything that facilitates discovery, which I think is fine

no strong objections

fbozic: > as-is it's an umbrella for everything that facilitates discovery

let's go with this since u both independently came to same conclusion 😄 
sorry Petar for back and forth

miraclx: Ah wait.. docs says it.

https://github.com/calimero-network/core/blob/f6efd7e530e8fe3721b660283e294ed26f388c98/crates/network/src/discovery/state.rs#L19-L23

Maybe move it to `Discovery` idk?

petarjuki7: yeah, I put it back

Issue #1122, opened 2025-02-18
Consider alias replacement in execution args


miraclx:
When an alias is created for a context identity, we may want to consider allowing alias substitution.

For example:

```console
context call chat_app --as jane send_message '{"recipient": "%emily%", "message": "Hi emily"}'
```

with `%emily%` being replaced with emily's public key in the context

rena-ertam028:
AI tried to solve the issue:

Added functionality to substitute aliases enclosed in %% in command arguments with their corresponding public keys in the context. This enables using aliases like %emily% which gets replaced with the actual public key during execution.

You can review changes in this commit: https://github.com/rena-ertam028/calimero-network-core-1122/commit/c18313ff33e6e485249dc895c4280c60488fd165.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

Afoxcute:
gmgm, Pls i would like to work on this.

chefsale:
@Afoxcute thanks for the interest. Can you join our discord and send the same message in dev-general and I will reach out via DM.

Afoxcute:
@chefsale Please send the new discord invite link.

chefsale:
There you go: https://discord.gg/QryD7VqE

Afoxcute:
@chefsale Does Calimero org have contribution guide? If yes, Can you point me to it?

petarjuki7:
> [@chefsale](https://github.com/chefsale) Does Calimero org have contribution guide? If yes, Can you point me to it?

You can find them [here](https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md)

Afoxcute:
This is well descriptive enough to complete the implementation, I will review and revert back.

chefsale:
Assigned, good luck.

PR #1141, opened 2025-03-08
Add Alias Substitution for Context Parameters

Afoxcute:
**Add alias substitution in context parameters using %alias% syntax**

**Features Added**

- Added support for alias substitution in both context create and context call commands
- Implemented a regex-based parser to identify and replace %alias% patterns with their corresponding public keys
- Added informative messages about successful substitutions and warnings for failed lookups
- Ensured backward compatibility with existing commands

**Example Usage**

# Create a context with alias substitution in parameters
meroctl context create --application-id app123 --protocol chat --params '{"admin": "%admin%", "users": ["%alice%", "%bob%"]}'

# Call a context function with alias substitution
meroctl context call chat_app --as alice send_message '{"recipient": "%bob%", "message": "Hello Bob!"}'

**Implementation Details**

- Added a substitute_aliases function
- Modified the create_context and call_context functions to process parameters before sending them to the application
- Added appropriate error handling and user feedback

**Testing**

- Parameters with single and multiple alias substitutions
- Parameters with non-existent aliases
- Parameters with no substitution needed
- JSON and non-JSON parameter formats


Afoxcute:
@chefsale Could you review pls?

crates/meroctl/src/cli/context/call.rs
```diff
@@ -0,0 +1,185 @@
...
+                        context_id
+                    } else {
+                        bail!("Context alias '{}' not found", self.context);
+                    }
+                }
+                Err(e) => bail!("Error looking up context alias '{}': {}", self.context, e),
+            }
+        } else {
+            bail!("Invalid context ID or alias format: {}", self.context);
+        };
```
petarjuki7: There are examples in our repo how to do this more cleanly, for example:
https://github.com/calimero-network/core/blob/master/crates/meroctl/src/cli/context/get.rs#L82-#L86 
```suggestion
       let context_id = resolve_alias(multiaddr, &config.identity, self.context, None)
            .await?
            .value()
            .cloned()
            .ok_or_eyre("unable to resolve")?;
```

crates/meroctl/src/cli/context/call.rs
```diff
@@ -0,0 +1,185 @@
...
+                    Some(public_key)
+                } else {
+                    bail!("Identity alias not found");
+                }
+            }
+            Err(e) => bail!("Error looking up identity alias: {}", e),
+        }
+    } else {
+        None
+    };
```
petarjuki7: Same here, use the `resolve_alias()` function

crates/meroctl/src/cli/context/call.rs
```diff
@@ -0,0 +1,185 @@
...
+    let response: CallContextResponse =
+        do_request(client, url, Some(request), keypair, RequestType::Post).await?;
+
+    environment.output.write(&response);
+
+    Ok(())
+}
+
+/// Substitutes aliases in the format %alias% with their corresponding public keys
+async fn substitute_aliases(
```
petarjuki7: If you are going to use this function in more than one place, put it in `crates/meroctl/src/common.rs`.

crates/meroctl/src/cli/context/create.rs
```diff
@@ -223,6 +231,55 @@ pub async fn create_context(
     Ok((response.data.context_id, response.data.member_public_key))
 }
 
+/// Substitutes aliases in the format %alias% with their corresponding public keys
+async fn substitute_aliases(
```
petarjuki7: As far as I can see, it's the same function as above? Better put it into `crates/meroctl/src/common.rs`

crates/meroctl/src/cli/context/mod.rs
petarjuki7, line 1: We don't need a `mod.rs` here, we handle all `context` meroctl commands from the `crates/meroctl/src/cli/context.rs` file.

crates/meroctl/src/cli/context/remove.rs
```diff
@@ -0,0 +1,58 @@
...
+                        context_id
+                    } else {
+                        bail!("Context alias '{}' not found", self.context);
+                    }
+                }
+                Err(e) => bail!("Error looking up context alias '{}': {}", self.context, e),
+            }
+        } else {
+            bail!("Invalid context ID or alias format: {}", self.context);
+        };
```
petarjuki7: Use `resolve_alias()`

crates/meroctl/src/cli/mod.rs
petarjuki7, line 1: I don't think we need this `mod.rs` file? We handle all top level `meroctl` commands from the `crates/meroctl/src/cli.rs`

crates/meroctl/src/cli/context/remove.rs
petarjuki7, line 1: We already have an implemented `context delete` command (see: [here](https://github.com/calimero-network/core/blob/master/crates/meroctl/src/cli/context/delete.rs))

crates/meroctl/src/cli/context/call.rs
petarjuki7, line 1: We already have a `call` command implemented here: `crates/meroctl/src/cli/call.rs`.
Better to just update that than to create a new one.

crates/meroctl/src/cli/context/list.rs
```diff
@@ -1,4 +1,5 @@
-use calimero_server_primitives::admin::GetContextsResponse;
+use calimero_primitives::context::ContextId;
+use calimero_server_primitives::admin::ListContextsResponse;
```
petarjuki7: `ListContextsResponse` struct doesn't exist in `calimero_server_primitives::admin`... I feel it's better to stick with `GetContextsResponse`.

petarjuki7 (changes requested):

Afoxcute:
@petarjuki7 i will make the changes you requested.

crates/meroctl/src/cli/context/call.rs
```diff
@@ -0,0 +1,185 @@
...
+    params: &str,
+) -> EyreResult<String> {
+    let re = Regex::new(r"%([^%]+)%")?;
+    let mut result = params.to_string();
+    
+    for cap in re.captures_iter(params) {
+        if let Some(alias_match) = cap.get(1) {
+            let alias_str = alias_match.as_str();
+            // Parse the alias string into an Alias<PublicKey>
+            if let Ok(alias) = alias_str.parse::<Alias<PublicKey>>() {
```
miraclx: when using `if let..else`, with the consequent block (if) extending beyond just a couple of lines and the alternative (else) effectively bails, consider the [`let..else`](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) syntax.

```rust
if let OK(val) = thing {
    func(val);
    do_ok_1;
    do_ok_2;
    do_ok_3;
} else {
    do_else_1;
    return;
}
```

```rust
let Ok(val) = thing else {
    do_else_1;
    return;
};

func(val);
do_ok_1;
do_ok_2;
do_ok_3;
```

this syntax helps flatten out indentation, and keeps short-circuits closer to the condition that triggered them, making it easier to read without needing to jump over the code

miraclx (commented):
Hi Afoxcute, thanks for taking a stab at this!

We just discussed this at the engineering sync, and came to this resolution for the ideal implementation of this feature: https://github.com/calimero-network/core/issues/1122#issuecomment-2710975001

Can you have a look and let us know what you think?

Also, seconding the points @petarjuki7 made! Things like not using `mod.rs` are design choices and should be mentioned in the contributions guide, we'll be sure to update that.

PR #1142, opened 2025-03-10
EVM Implementation 

alenmestrov:
## Description

Introduced EVM into Calimero. Enabled reading and writing to context contract. 

## Test plan

For testing you will need to install anvil for local devnet of EVM environment, follow setup steps in this [repo](https://github.com/calimero-network/calimero-evm-contracts) to setup the contracts.

Then start anvil, deploy context contract and call setProxyCode method with following commands:

*Run this from root dir of contract folder:

```bash
forge script script/ContextConfig.s.sol \             
    --rpc-url http://localhost:8545 \
    --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
    --broadcast   
```
In src/script/ContextConfig.s.sol you will see line with adding owner (address) as constructor argument to contract, change it to one of the accounts received when running `anvil` command.

- http://localhost:8545 - URL of devnet - in output of anvil command
- 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 - private key from anvil

```bash
cast send 0x5FbDB2315678afecb367f032d93F642f64180aa3 \
    "setProxyCode(bytes)" \
    $(forge inspect ContextProxy bytecode) \
    --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
    --rpc-url http://localhost:8545
```

- 0x5FbDB2315678afecb367f032d93F642f64180aa3 - contract address received when deploying contract

Try to create a context 

## Documentation update
Documentation PR is [here](https://github.com/calimero-network/calimero-network.github.io/pull/120) and will be merged after this PR and new release is published.


crates/context/config/src/client/config.rs
```diff
@@ -71,5 +72,13 @@ pub enum Credentials {
     Near(NearCredentials),
     Starknet(StarknetCredentials),
     Icp(IcpCredentials),
-    Stellar(StellarCredentials),
+    Raw(RawCredentials),
```
frdomovic, line 75: Stellar and Evm imported but not used anywhere? How does this update work? If they are same creds object to we need these two imports then?

crates/context/config/src/client.rs
```diff
@@ -201,13 +203,55 @@ impl Client<AnyTransport> {
...
                 right: Both {
                     left: icp_transport,
-                    right: stellar_transport,
+                    right: Both {
+                        left: stellar_transport,
+                        right: evm_transport,
+                    },
                 },
             },
         };
```
miraclx: would be interesting to introduce a macro to simplify this at some point

```rust
transport! {
    pub type LocalTransports<'a> = (
        MaybeNear<'a>,
        MaybeStarknet<'a>,
        MaybeIcp<'a>,
        MaybeStellar<'a>,
    );
};
```

and

```rust
let all_transport = transport!(
    near_transport,
    starknet_transport,
    icp_transport,
    stellar_transport,
);
```

crates/context/config/src/client/env/config/types/evm.rs
```diff
@@ -0,0 +1,289 @@
...
+            // If your contract expects an array of members and an array of capabilities:
+            let members: Vec<FixedBytes<32>> = capabilities
+                .iter()
+                .map(|(member, _)| FixedBytes::from(member.as_bytes()))
+                .collect();
+
+            let capability_values: Vec<u8> = capabilities
+                .iter()
+                .map(|(_, capability)| capability.to_sol() as u8)
+                .collect();
```
miraclx: [`Iterator::unzip`](https://doc.rust-lang.org/beta/std/iter/trait.Iterator.html#method.unzip) or [`.collect::<(Vec<_>, Vec<_>)>`](https://doc.rust-lang.org/beta/std/iter/trait.FromIterator.html#impl-FromIterator%3C(EA,)%3E-for-(A,))

crates/node/src/interactive_cli/context.rs
```diff
@@ -25,6 +25,7 @@ enum Protocol {
     Starknet,
     Icp,
     Stellar,
+    Evm,
```
miraclx: this should be `ethereum`

crates/merod/src/cli/init.rs
```diff
@@ -331,6 +332,34 @@ impl InitCommand {
                 .insert("stellar".to_owned(), local_config);
         }
 
+        {
+            let _ignored = client_params.insert(
+                "evm".to_owned(),
+                ClientConfigParams {
+                    network: "sepolia".into(),
+                    protocol: "evm".into(),
```
miraclx: the protocol name isn't `evm`

miraclx (commented):
the rest of the comments are covered in the [other PR](https://github.com/calimero-network/core/pull/1146)

crates/merod/src/cli/init.rs
miraclx: but the name can be revised in a follow-up PR, not to block this one.

crates/context/config/src/client.rs
miraclx: Filed in #1150.

frdomovic (approved):

Cargo.lock
miraclx, line 1: bad merge here, restore master and rebuild

```bash

$ cargo build -p merod -p calimero-context-config --features client
```

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -24,6 +24,14 @@ use crate::types::Signed;
 use crate::{ContextIdentity, Request, RequestKind};
 pub mod methods;
 
+use alloy::primitives::{keccak256, B256};
+use alloy::signers::local::PrivateKeySigner;
+use alloy::signers::{Signature, SignerSync};
+use alloy_sol_types::SolValue;
```
miraclx: move these to L9

crates/context/config/src/client/env/config/mutate.rs
```diff
@@ -24,6 +24,14 @@ use crate::types::Signed;
 use crate::{ContextIdentity, Request, RequestKind};
 pub mod methods;
```
miraclx: move this before L10

crates/context/config/src/client/env/config/query/privileges.rs
```diff
@@ -219,3 +222,52 @@ impl<'a> Method<Stellar> for PrivilegesRequest<'a> {
...
+
+            let capabilities = user_cap
+                .capabilities
+                .into_iter()
+                .map(|cap| match cap {
+                    SolCapability::ManageApplication => Capability::ManageApplication,
+                    SolCapability::ManageMembers => Capability::ManageMembers,
+                    SolCapability::Proxy => Capability::Proxy,
+                    SolCapability::__Invalid => {
+                        panic!("Invalid capability encountered in response")
```
miraclx: `bail!`

crates/context/config/src/client/env/config/query/privileges.rs
```diff
@@ -219,3 +222,52 @@ impl<'a> Method<Stellar> for PrivilegesRequest<'a> {
...
+                    }
+                })
+                .collect();
+
+            assert!(
+                result.insert(user_id, capabilities).is_none(),
+                "Duplicate user ID in response"
+            );
```
miraclx:
```rust
if result.insert(user_id, capabilities).is_some() {
    bail!(..)
}
```

miraclx (approved):
lgtm, though we still have some `SolValue::abi_encode(&value)` instead of `value.abi_encode()`

approving to unblock

Cargo.lock
frdomovic: Updated with these exact instructions.

frdomovic:
Before merging @petarjuki7 said he wants to test it all (as someone who did not work on this to maybe find potential bugs we overlooked) , so after his approval we can merge.
cc @alenmestrov @miraclx

.github/workflows/e2e-tests.yml
```diff
@@ -12,7 +12,7 @@ on:
       - '.github/workflows/e2e_tests.yml'
 
 env:
-  CALIMERO_CONTRACTS_VERSION: '0.4.0' # Or you can use 'prerelease-{PR-NUMBER}'
+  CALIMERO_CONTRACTS_VERSION: 'latest'
```
miraclx, line 15: Can we keep the comment? Maybe one line above so it's not mistakenly removed when this line is modified.

.github/workflows/e2e-tests.yml
```diff
@@ -71,27 +74,21 @@ jobs:
...
-          SWARM_PORT="${random_numbers[0]}"
-          SERVER_PORT="${random_numbers[1]}"
-          ICP_PORT="${random_numbers[2]}"
+          get_random_port() {
+            python3 -c 'import socket; s = socket.socket(); s.bind(("", 0)); port = s.getsockname()[1]; s.close(); print(port)'
+          }
+          
+          SWARM_PORT=$(get_random_port)
+          SERVER_PORT=$(get_random_port)
+          ICP_PORT=$(get_random_port)
```
miraclx, line 83: this doesn't guarantee that multiple invocations of `get_random_port` don't return the same port

.github/workflows/e2e-tests.yml
```diff
@@ -71,27 +74,21 @@ jobs:
...
           SWARM_HOST=$(ifconfig | grep 'inet ' | grep -v '127.0.0.1' | awk '{print $2}' | head -n 1)
-
+          
+          # Set outputs for other steps
           echo "SWARM_PORT=$SWARM_PORT" >> $GITHUB_OUTPUT
           echo "SERVER_PORT=$SERVER_PORT" >> $GITHUB_OUTPUT
           echo "ICP_PORT=$ICP_PORT" >> $GITHUB_OUTPUT
           echo "SWARM_HOST=$SWARM_HOST" >> $GITHUB_OUTPUT
-
+
```
miraclx, line 91: extraneous whitespace

crates/meroctl/src/cli/proxy/get.rs
```diff
@@ -27,7 +28,14 @@ pub struct GetCommand {
     pub context: Alias<ContextId>,
 
     #[arg(value_name = "PROPOSAL_ID", help = "proposal_id of the proposal")]
-    pub proposal_id: Hash,
+    pub proposal_id: Option<Hash>,
+
+    #[arg(long, value_parser = serde_value, help = "JSON arguments to pass to the method (e.g., {\"offset\": 0, \"limit\": 10})")]
+    pub args: Option<Value>,
+}
```
miraclx, line 35: we need to redesign this proxy subcommand entirely, but out of scope

when to use `proposal_id` over `args` is unclear, and vice-versa

e2e-tests/config/scenarios/kv-store/test.json
miraclx, line 1: we shouldn't remove this, this is the simple test case, integrations are a separate thing too

e2e-tests/src/driver.rs
```diff
@@ -585,23 +631,16 @@ impl PortBinding {
...
             }
+            Err(e) => bail!("Failed to bind to random port: {}", e),
         }
-
-        bail!(
-            "unable to select a port in range {}..={}",
-            *port - 100,
-            *port - 1
-        );
     }
```
miraclx, line 645: why did this change?

e2e-tests/src/main.rs
```diff
@@ -92,9 +92,20 @@ pub struct RootArgs {
...
+}
+
+fn parse_scenario(s: &str) -> Result<String, String> {
+    match s {
+        "ethereum" | "near" | "stellar" | "icp" => Ok(s.to_owned()),
+        _ => Err(format!(
+            "Invalid scenario. Must be one of: ethereum, near, stellar, icp. Got: {}",
+            s
+        )),
+    }
```
miraclx, line 108: could've been an enum with `arg(value_enum)`

e2e-tests/src/main.rs
```diff
@@ -92,9 +92,20 @@ pub struct RootArgs {
     #[arg(env = "E2E_OUTPUT_FORMAT", hide_env_values = true)]
     pub output_format: OutputFormat,
 
-    /// Filter tests by protocols (e.g., "stellar near icp")
-    #[arg(long, value_name = "PROTOCOLS", num_args = 0..)]
-    pub protocols: Vec<String>,
+    /// Scenario to run ("ethereum", "near", "stellar", "icp")
+    #[arg(long, value_name = "SCENARIO")]
+    #[arg(value_parser = parse_scenario)]
+    pub scenario: String,
```
miraclx, line 98: any reason this can no longer operate on a list of scenarios?

e2e-tests/src/meroctl.rs
```diff
@@ -159,6 +159,30 @@ impl Meroctl {
         Ok((public_key, private_key))
     }
 
+    pub async fn get_proposals(
+        &self,
+        node_name: &str,
+        context_id: &str,
+        args: &serde_json::Value,
+    ) -> EyreResult<serde_json::Value> {
+        let args_json = serde_json::to_string(args).unwrap();
```
miraclx, line 168: thought it technically should never fail

```suggestion
    ) -> EyreResult<serde_json::Value> {
        let args_json = serde_json::to_string(args)?;
```

e2e-tests/src/meroctl.rs
```diff
@@ -174,7 +198,7 @@ impl Meroctl {
             [
                 "call",
                 context_id,
-                method_name,
+                &method_name,
```
miraclx, line 201: this shouldn't be necessary

e2e-tests/src/steps/application_install.rs
```diff
@@ -96,9 +104,49 @@ impl Test for ApplicationInstallStep {
...
+                let bytes = response.bytes().await?;
+
+                let decoded_bytes = if url.ends_with(".gz") {
+                    let mut decoder = GzDecoder::new(&bytes[..]);
+                    let mut decompressed = Vec::new();
+                    decoder.read_to_end(&mut decompressed)?;
+                    decompressed
+                } else {
+                    bytes.to_vec()
+                };
```
miraclx, line 139: consider using [`async_compression`](https://docs.rs/async-compression/latest/async_compression/futures/bufread/struct.GzipDecoder.html) to prevent buffering the whole file in memory

then you can call [`into_async_read()`](https://docs.rs/futures-util/latest/futures_util/stream/trait.TryStreamExt.html#method.into_async_read) on the [`bytes_stream`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes_stream) (after mapping the error of course)

You can then [`split`](https://docs.rs/futures-util/latest/futures_util/io/trait.AsyncReadExt.html#method.split) the `AsyncRead` into read/write halves.

And then join the task that writes from the request to the decoder write half and the task that writes from the decoder read half to the file.

e2e-tests/src/steps/get_proposals.rs
```diff
@@ -0,0 +1,60 @@
...
+                    ids.push(id.to_owned());
+                }
+            }
+        }
+
+        if ids.is_empty() {
+            bail!("No proposal IDs found in response: {:?}", proposals);
+        }
+
+        ctx.proposal_id = Some(ids.first().unwrap().clone());
```
miraclx, line 53: Why keep all the IDs if you only need the first one?

use an Option, and break from the loop when the value is set

also, indexing `serde_json::Value` [doesn't panic](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#impl-Index%3CI%3E-for-Value), it'll return `Value::Null` instead

```suggestion
        let mut proposal_id = None;

        if let Some(proposals) = proposals["data"].as_array() {
            for proposal in proposals {
                if let Some(id) = proposal["id"].as_str() {
                    proposal_id = Some(id);
                    break;
                }
            }
        }

        let Some(proposal_id) = proposal_id else {
            bail!("No proposal IDs found in response: {:?}", proposals);
        };

        ctx.proposal_id = Some(proposal_id);
```

PR #1143, opened 2025-03-10
feat(e2e): add testing aliases to e2e tests

petarjuki7:
# Add e2e test for testing context aliases

Create an alias for a context on the inviter node named `my_context`, and use that alias instead of a `contextId` in subsequent jsonrpc calls.

## Documentation update

N/A

### DIFF
--- a/e2e-tests/src/driver.rs
+++ b/e2e-tests/src/driver.rs
@@ -32,6 +32,7 @@ pub struct TestContext<'a> {
     pub invitees_public_keys: HashMap<String, String>,
     pub protocol_name: &'a str,
     pub output_writer: OutputWriter,
+    pub context_alias: Option<String>,
 }
 
 pub trait Test {
@@ -57,6 +58,7 @@ impl<'a> TestContext<'a> {
             inviter_public_key: None,
             invitees_public_keys: HashMap::new(),
             output_writer,
+            context_alias: None,
         }
     }
 }
--- a/e2e-tests/src/steps/jsonrpc_call.rs
+++ b/e2e-tests/src/steps/jsonrpc_call.rs
@@ -27,11 +27,7 @@ impl Test for CallStep {
     }
 
     async fn run_assert(&self, ctx: &mut TestContext<'_>) -> EyreResult<()> {
-        let mut context_id = if let Some(ref alias) = ctx.context_alias {
-            alias
-        } else {
-            bail!("Context ID or Alias is required for JsonRpcExecuteStep");
-        };
+        let context_id;
 
         let mut public_keys = HashMap::new();
         if let Some(ref inviter_public_key) = ctx.inviter_public_key {
@@ -41,7 +37,13 @@ impl Test for CallStep {
         }
 
         match self.target {
-            CallTarget::Inviter => {}
+            CallTarget::Inviter => {
+                if let Some(ref alias) = ctx.context_alias {
+                    context_id = alias;
+                } else {
+                    bail!("Alias is required for JsonRpcExecuteStep on the Inviter node");
+                };
+            }
             CallTarget::AllMembers => {
                 if let Some(ref id) = ctx.context_id {
                     context_id = id;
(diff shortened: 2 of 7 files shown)


e2e-tests/src/driver.rs
```diff
@@ -32,6 +32,7 @@ pub struct TestContext<'a> {
     pub invitees_public_keys: HashMap<String, String>,
     pub protocol_name: &'a str,
     pub output_writer: OutputWriter,
+    pub context_alias: Option<String>,
```
miraclx, line 35: why not merge this with context? it's already a `String`

e2e-tests/src/steps/jsonrpc_call.rs
```diff
@@ -27,8 +27,10 @@ impl Test for CallStep {
     }
 
     async fn run_assert(&self, ctx: &mut TestContext<'_>) -> EyreResult<()> {
-        let Some(ref context_id) = ctx.context_id else {
-            bail!("Context ID is required for JsonRpcExecuteStep");
+        let mut context_id = if let Some(ref alias) = ctx.context_alias {
+            alias
+        } else {
+            bail!("Context ID or Alias is required for JsonRpcExecuteStep");
```
miraclx: would also solve the issue of there not being a `context_alias`, but we have the `context_id`

petarjuki7: True, will fix

petarjuki7: Ah, my idea was that only the `inviter` does calls with the alias, because I create it only on the `inviter`. So I can still use `context_id` on the other node.

e2e-tests/src/driver.rs
fbozic: I'm not sure if I understand this comment fully, it is already part of `TestContext` struct.

fbozic: okay I got it, just put it under `context_id` please. or if you create the same alias on all nodes we don't need context_id anymore. but it is fine as it is now.

e2e-tests/src/steps/jsonrpc_call.rs
miraclx: this is not yet resolved

miraclx (changes requested):

miraclx (approved):

e2e-tests/src/steps/jsonrpc_call.rs
miraclx: this message seems to suggest we should've been checking the presence of the context ID, though we're not, so it seemed incomplete

read through and it holds up, can we at least change the message

And for maximum clarity, the issue here was it would've made `ContextCreateAliasStep` **REQUIRED** for `CallStep`, even though it's only required for the inviter

e2e-tests/src/steps/jsonrpc_call.rs
```diff
@@ -41,7 +37,13 @@ impl Test for CallStep {
         }
 
         match self.target {
-            CallTarget::Inviter => {}
+            CallTarget::Inviter => {
+                if let Some(ref alias) = ctx.context_alias {
+                    context_id = alias;
+                } else {
+                    bail!("Alias is required for JsonRpcExecuteStep on the Inviter node");
```
miraclx, line 44: ah you have, noice!

e2e-tests/src/steps/jsonrpc_call.rs
```diff
@@ -27,11 +27,7 @@ impl Test for CallStep {
     }
 
     async fn run_assert(&self, ctx: &mut TestContext<'_>) -> EyreResult<()> {
-        let mut context_id = if let Some(ref alias) = ctx.context_alias {
-            alias
-        } else {
-            bail!("Context ID or Alias is required for JsonRpcExecuteStep");
-        };
+        let context_id;
```
miraclx, line 30: ideally, this should be localized to the definition at L39 right?

miraclx (approved):

PR #1146, opened 2025-03-11
feat(node): EVM integration + cleanup

frdomovic:
# feat(node): EVM integration + cleanup

Wrapping up EVM integration + code cleanup.
- replace unsigned mutate transaction with alloy signed transaction
- protocol file cleanup
- query + mutate encode decode for proxy contract

## Test plan
Test #1 - create context with evm protocol flag
Screenshot 2025-03-11 at 13 38 20

Test #2 - invite user to created context
Test #3 - join invited context
Test #4 - create change min approvals proposal
Test #5 - approve from node2 node3 - proposal executed
Test #6 - create transfer proposal
Test #7 - create cross contract call proposal with deposit + 2 arguments
Test #8 - create cross contract call proposal with no deposit + 2 arguments
Test #9 - create cross contract call proposal with deposit + 6 arguments

## Documentation update
TBD


crates/context/config/src/client/env/config/query/privileges.rs
```diff
@@ -227,7 +229,16 @@ impl<'a> Method<Evm> for PrivilegesRequest<'a> {
...
     fn encode(self) -> eyre::Result<Vec<u8>> {
-        todo!()
+        let context_id: [u8; 32] = self.context_id.rt().expect("infallible conversion");
+        let context_id_val = B256::from_slice(&context_id);
+
+        let mut identities_val = Vec::new();
+        for identity in self.identities.iter() {
+            let identity_raw: [u8; 32] = identity.rt().expect("infallible conversion");
+            identities_val.push(B256::from_slice(&identity_raw));
+        }
```
alenmestrov:
```rust
let identities: Vec<B256> = self.identities
            .into_iter()
            .map(|id| {
                let bytes: [u8; 32] = id.rt().expect("infallible conversion");
                B256::from_slice(&bytes)
            })
            .collect();
```

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
```diff
@@ -154,13 +156,27 @@ impl Method<Stellar> for ContextStorageEntriesRequest {
...
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        let decoded: Vec<B256> = SolValue::abi_decode(&response, false)?;
+
+        let context_storage_entries = decoded
+            .chunks(2)
+            .map(|chunk| ContextStorageEntry {
+                key: chunk[0].to_vec(),
+                value: chunk[1].to_vec(),
+            })
+            .collect();
```
miraclx: won't this mean `key` and `value` are limited to 32 bytes?

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -204,7 +199,16 @@ impl Network {
...
+        let signer: PrivateKeySigner = PrivateKeySigner::from_str(&self.secret_key).unwrap();
+        let wallet = EthereumWallet::from(signer);
+
+        let provider =
+            ProviderBuilder::new()
+                .wallet(wallet)
+                .on_http(Url::parse(&self.rpc_url).map_err(|e| EvmError::Custom {
+                    operation: ErrorOperation::Mutate,
+                    reason: e.to_string(),
+                })?);
```
miraclx: move these to `EvmTransport::new`

and store `Provider::builder().wallet(..).on_http(..)` instead of `RpcClient` in the state

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -159,35 +162,27 @@ impl Network {
...
                 "data": Bytes::from(call_data)
             }),
             json!("latest"),
-        ]; // Added "latest" block parameter
+        ];
 
-        // Send the eth_call and get the response
         let response: Bytes =
-            client
+            self.client
```
miraclx: once you have the provider in `&self`, you can call [`self.provider.client().request(..)`](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.client)

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -159,35 +162,27 @@ impl Network {
...
                 .request("eth_call", params)
```
miraclx: were these considered:

1. [`Provider::call`](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.call)
2. [`TransactionRequest::to`](https://docs.rs/alloy/latest/alloy/rpc/types/struct.TransactionRequest.html#method.to) & [`TransactionRequest::input`](https://docs.rs/alloy/latest/alloy/rpc/types/struct.TransactionRequest.html#method.input)
3. [`TransactionInput: From<Bytes>`](https://docs.rs/alloy/latest/alloy/rpc/types/struct.TransactionInput.html#impl-From%3CBytes%3E-for-TransactionInput)

should come down to something like this:

```rust
let request = TransactionRequest::default()
    .to(address)
    .input(Bytes::from(bytes).into());

let bytes = self.provider.call(request).await.map_err(...)?;
```

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -215,37 +219,36 @@ impl Network {
...
         let mut receipt = None;
+
+        // Wait for the transaction to be mined
         for _ in 0..30 {
-            // Try for 30 attempts with 2-second intervals
             let receipt_params = vec![json!(tx_hash)];
-            let result: Option<serde_json::Value> = client
+            let result: Option<serde_json::Value> = self
+                .client
                 .request("eth_getTransactionReceipt", receipt_params)
```
miraclx: use [`Provider::get_transaction_receipt`](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.get_transaction_receipt)

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
frdomovic: Updated, now using vec<u8> not limited to 32

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -73,9 +75,16 @@ pub struct EvmConfig<'a> {
...
+    provider: FillProvider<
+        JoinFill<
+            JoinFill<
+                Identity,
+                JoinFill<GasFiller, JoinFill<BlobGasFiller, JoinFill<NonceFiller, ChainIdFiller>>>,
+            >,
+            WalletFiller<EthereumWallet>,
+        >,
+        RootProvider,
+    >,
```
miraclx: Ah right, didn't account for `ProviderBuilder::wallet`, thought it would just be `RootProvider`

To simplify, we can call [`Provider::erased`](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.erased) to make this a `DynProvider<Ethereum>`.

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -89,14 +98,15 @@ impl<'a> EvmTransport<'a> {
...
-            let _ignored = networks.insert(
-                network_id.clone(),
-                Network {
-                    client: ClientBuilder::default().http(network_config.rpc_url.clone()),
-                    rpc_url: network_config.rpc_url.clone().to_string(),
-                    secret_key: network_config.access_key.clone(),
-                },
-            );
+            let signer: PrivateKeySigner =
+                PrivateKeySigner::from_str(&network_config.access_key).unwrap();
```
miraclx: we shouldn't be panicking here, pre-parse `access_key` before `NetworkConfig` is constructed

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -89,14 +98,15 @@ impl<'a> EvmTransport<'a> {
...
+            let wallet = EthereumWallet::from(signer);
+
+            let provider = ProviderBuilder::new()
```
miraclx:
```suggestion
            let provider = Provider::builder()
```

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
```diff
@@ -167,13 +166,14 @@ impl Method<Evm> for ContextStorageEntriesRequest {
...
 
         let context_storage_entries = decoded
             .chunks(2)
             .map(|chunk| ContextStorageEntry {
-                key: chunk[0].to_vec(),
-                value: chunk[1].to_vec(),
+                key: chunk[0].clone(),
+                value: chunk[1].clone(),
             })
             .collect();
```
miraclx:
```rust
let decoded: Vec<Vec<u8>> = decoded.into_iter().map(|b| b.to_vec()).collect();
```

^^^ this will reallocate for every `Bytes`

```rust
key: chunk[0].clone(),
value: chunk[1].clone(),
```

^^^ this will reallocate again for those bytes

[`Bytes: Into<Vec<u8>>`](https://docs.rs/alloy/latest/alloy/primitives/struct.Bytes.html#impl-From%3CBytes%3E-for-Vec%3Cu8%3E), so instead we can write this:

```suggestion
        let mut decoded = decoded.into_iter();

        let mut entries = Vec::with_capacity(decoded.len() / 2);

        while let Some(key) = decoded.next() {
            let value = decoded.next().ok_or_eyre("missing value for storage entry")?;
            
            entries.push(ContextStorageEntry {
                key: key.into(),
                value: key.into(),
            ));
        }
```

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -183,25 +182,20 @@ impl Network {
...
+        let request = TransactionRequest::default()
+            .to(address)
+            .input(Bytes::from(call_data).into());
 
-        let response: Bytes = self
+        let bytes = self
             .provider
-            .client()
-            .request("eth_call", params)
+            .call(&request)
```
miraclx, line 179:
```suggestion
            .call(&request)
            .block(BlockId::latest())
```

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -183,25 +182,20 @@ impl Network {
...
             .await
             .map_err(|e| EvmError::Custom {
                 operation: ErrorOperation::Query,
                 reason: format!("Failed to execute eth_call: {}", e),
             })?;
 
-        Ok(response.to_vec())
+        Ok(bytes.to_vec())
```
miraclx: [`Bytes: Into<Vec<u8>>`](https://docs.rs/alloy/latest/alloy/primitives/struct.Bytes.html#impl-From%3CBytes%3E-for-Vec%3Cu8%3E)

```suggestion
        Ok(bytes.into())
```

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -215,38 +213,37 @@ impl Network {
...
-            // Try for 30 attempts with 2-second intervals
-            let receipt_params = vec![json!(tx_hash)];
-            let result: Option<serde_json::Value> = client
-                .request("eth_getTransactionReceipt", receipt_params)
+            let result = self
+                .provider
+                .get_transaction_receipt(*tx_hash)
                 .await
                 .map_err(|e| EvmError::Custom {
                     operation: ErrorOperation::Mutate,
```
miraclx: also, why aren't we awaiting the [`PendingTransactionBuilder`](https://docs.rs/alloy/latest/alloy/providers/struct.PendingTransactionBuilder.html) returned by [`send_transaction`](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.send_transaction)?

why do we have a custom loop that's checking the transaction receipt

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -261,36 +258,20 @@ impl Network {
             std::thread::sleep(std::time::Duration::from_secs(2));
```
miraclx: also, this will block the thread

crates/context/config/src/client/protocol/evm.rs
miraclx: Maybe cc: @alenmestrov

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/protocol/evm.rs
frdomovic: Was throwing issue so didn't update

crates/context/config/src/client/protocol/evm.rs
frdomovic: updated

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/protocol/evm.rs
frdomovic: _In many cases, sending a transaction doesn't return the transaction receipt directly; it returns a pending transaction object that can be used to track the transaction until it's mined_ ~ as it says

crates/context/config/src/client/protocol/evm.rs
frdomovic: replaced with tokio

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

crates/context/config/src/client/protocol/evm.rs
miraclx: yes, but isn't that what we want? waiting for the transaction to be finalized?

crates/context/config/src/client/protocol/evm.rs
miraclx: what was the error? you probably needed to specify `N` in `Provider<N>`

```rust
Provider::<Ethereum>::builder()
```

but that's me guessing, either way, as you have it is alright

crates/context/config/src/client/protocol/evm.rs
miraclx: resolving to be cleaned up by the comment above this, we shouldn't need to be manually polling for receipt finalization IMO if `send_transaction` already returns a future that waits for the transaction to be resolved.

crates/context/config/Cargo.toml
```diff
@@ -32,8 +33,10 @@ starknet = { workspace = true, optional = true }
 starknet-crypto = { workspace = true, optional = true }
 starknet-types-core = { workspace = true, optional = true }
 thiserror.workspace = true
+tokio = { workspace = true, features = ["sync", "macros"] }
 url = { workspace = true, optional = true }
 
+
```
miraclx: extraneous space

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -197,21 +200,108 @@ impl From<ProposalAction> for SolProposalAction {
...
+                        ethabi::StateMutability::Payable
+                    } else {
+                        ethabi::StateMutability::NonPayable
+                    };
+                    let amount = U256::from(deposit);
 
-                    let mut call_data = Vec::with_capacity(4 + args.len());
-                    call_data.extend_from_slice(&selector);
-                    call_data.extend_from_slice(&args.as_bytes());
+                    #[allow(deprecated)]
```
miraclx: localize to L294, since it's only the `constant` field that is deprecated

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -197,21 +200,108 @@ impl From<ProposalAction> for SolProposalAction {
...
+                                        ParamType::Array(Box::new(ParamType::Int(256)))
+                                    }
+                                    "array(address)" => {
+                                        ParamType::Array(Box::new(ParamType::Address))
+                                    }
+                                    "array(bool)" => ParamType::Array(Box::new(ParamType::Bool)),
+                                    "array(bytes)" => ParamType::Array(Box::new(ParamType::Bytes)),
+                                    "tuple" => ParamType::Tuple(vec![]),
+                                    _ => panic!("Unsupported parameter type: {}", key),
+                                };
```
miraclx: instead of two iterations, you can generate both `tokens` and `inputs` from a single iteration if you used either [`Iterator::unzip`](https://doc.rust-lang.org/beta/std/iter/trait.Iterator.html#method.unzip) or [`.collect::<(Vec<_>, Vec<_>)>()`](https://doc.rust-lang.org/beta/std/iter/trait.FromIterator.html#impl-FromIterator%3C(EA,)%3E-for-(A,)).

You'd just have to return a tuple for each item during your iteration.

crates/context/config/src/client.rs
```diff
@@ -228,12 +230,17 @@ impl Client<AnyTransport> {
                         eyre::bail!("missing account id for `{}` signer", network);
                     };
 
+                    let access_key: PrivateKeySigner = PrivateKeySigner::from_str(
+                        &credentials.secret_key.clone(),
+                    )
+                    .map_err(|_| eyre::eyre!("failed to convert secret key to PrivateKeySigner"))?;
```
miraclx: try [`.wrap_err`](https://docs.rs/eyre/latest/eyre/trait.WrapErr.html#tymethod.wrap_err) instead, so we don't lose the reason why the `secret_key` isn't valid

```suggestion
                    .wrap_err("failed to convert secret key to PrivateKeySigner")?;
```

crates/context/config/src/client/env/proxy/query/proposal.rs
```diff
@@ -164,10 +165,13 @@ impl Method<Stellar> for ProposalRequest {
...
 
-    const METHOD: &'static str = "proposal";
+    const METHOD: &'static str = "getProposal(bytes32)";
 
     fn encode(self) -> eyre::Result<Vec<u8>> {
-        todo!()
+        let proposal_id: [u8; 32] = self.proposal_id.rt().expect("infallible conversion");
+        let proposal_id_val = B256::from_slice(&proposal_id);
+
+        Ok(SolValue::abi_encode(&(proposal_id_val)))
```
miraclx: Looks like [`[u8; 32]: SolValue`](https://docs.rs/alloy/latest/alloy/sol_types/trait.SolValue.html#impl-SolValue-for-%5Bu8;+N%5D).

which means we can skip L172, so this can be simplified to:

```suggestion
        let proposal_id: [u8; 32] = self.proposal_id.rt().expect("infallible conversion");

        Ok(proposal_id.abi_encode())
```

crates/context/config/src/client/env/proxy/query/proposal_approvers.rs
```diff
@@ -176,13 +178,25 @@ impl Method<Stellar> for ProposalApproversRequest {
...
 
-    const METHOD: &'static str = "proposal_approvers";
+    const METHOD: &'static str = "proposalApprovers(bytes32)";
 
     fn encode(self) -> eyre::Result<Vec<u8>> {
-        todo!()
+        let proposal_id: [u8; 32] = self.proposal_id.rt().expect("infallible conversion");
+        let proposal_id_val = B256::from_slice(&proposal_id);
+
+        Ok(SolValue::abi_encode(&(proposal_id_val)))
```
miraclx: here too, and others..

crates/context/config/src/client/env/proxy/query/proposal_approvers.rs
```diff
@@ -176,13 +178,25 @@ impl Method<Stellar> for ProposalApproversRequest {
...
-    fn decode(_response: Vec<u8>) -> eyre::Result<Self::Returns> {
-        todo!()
+    fn decode(response: Vec<u8>) -> eyre::Result<Self::Returns> {
+        let decoded: Vec<B256> = SolValue::abi_decode(&response, false)?;
+
+        let identities: Vec<[u8; 32]> = decoded.into_iter().map(|val| val.into()).collect();
+
+        let context_identities: Vec<ContextIdentity> = identities
+            .into_iter()
+            .map(|bytes| bytes.rt().expect("infallible conversion"))
```
miraclx: surely you don't need an intermediate allocation

```suggestion
        let context_identities: Vec<ContextIdentity> = decoded
            .into_iter()
            .map(|bytes| bytes.rt().expect("infallible conversion"))
```

crates/context/config/Cargo.toml
```diff
@@ -82,7 +85,8 @@ stellar_client = [
 
 evm = [
```
miraclx, line 85: should this be `ethereum` instead?

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +189,191 @@ impl From<SolProposalAction> for ProposalAction {
...
+                    let data = SolValue::abi_encode(&(contract_address, call_data, amount));
+
+                    data.into()
+                }
+                ProposalAction::Transfer {
+                    receiver_id,
+                    amount,
+                } => {
+                    let data = TransferData {
+                        recipient: receiver_id.parse().expect("Invalid address"),
```
miraclx: isn't this unsanitized input from the app (untrusted) (which will likely be from some already untrusted frontend)? this would mean a malicious app can crash the node 😕

crates/context/config/Cargo.toml
miraclx: https://github.com/calimero-network/core/pull/1142#discussion_r1995920234

we'll revise the name in a separate PR

crates/context/config/src/client/env/proxy/types/evm.rs
frdomovic: Updated with unwrap_or_else

crates/context/config/src/client.rs
```diff
@@ -5,6 +5,7 @@ use std::str::FromStr;
 
 use alloy::signers::local::PrivateKeySigner;
 use env::Method;
+use eyre::Context;
```
miraclx, line 7: move to line 18, since it's a module not a crate

crates/context/config/src/client/env/config/query/application.rs
```diff
@@ -161,9 +160,8 @@ impl Method<Evm> for ApplicationRequest {
 
     fn encode(self) -> eyre::Result<Vec<u8>> {
         let context_id: [u8; 32] = self.context_id.rt().expect("infallible conversion");
-        let context_id_val = B256::from_slice(&context_id);
 
-        Ok(SolValue::abi_encode(&(context_id_val)))
+        Ok(SolValue::abi_encode(&(context_id)))
```
miraclx: why not this?

```suggestion
        Ok(context_id.abi_encode())
```

crates/context/config/src/client/env/config/query/application_revision.rs
```diff
@@ -119,9 +118,8 @@ impl Method<Evm> for ApplicationRevisionRequest {
 
     fn encode(self) -> eyre::Result<Vec<u8>> {
         let context_id: [u8; 32] = self.context_id.rt().expect("infallible conversion");
-        let context_id_val = B256::from_slice(&context_id);
 
-        Ok(SolValue::abi_encode(&(context_id_val)))
+        Ok(SolValue::abi_encode(&(context_id)))
```
miraclx: and others?

crates/context/config/src/client/env/proxy/types/evm.rs
miraclx: there are others, see `SolProposalAction: From<ProposalAction>`

if there are legitimate cases where the conversion can fail (with no reasonable default), consider making it in to a `TryFrom` instead, so you can return a result

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -308,7 +311,8 @@ impl From<ProposalAction> for SolProposalAction {
                     amount,
                 } => {
                     let data = TransferData {
-                        recipient: receiver_id.parse().expect("Invalid address"),
+                        recipient: Address::from_str(&receiver_id)
```
miraclx: like this.... potentially dangerous, because this will cause the money to be burnt, whenever the address is invalid, which is questionable

crates/context/config/src/client/env/proxy/types/evm.rs
frdomovic: Tried few things but localisation did not work.

crates/context/config/src/client.rs
miraclx: I should've been clearer, we should move `env::Method` not `eyre::Context`

crates/context/config/src/client/env/config/query/application.rs
frdomovic: Updated all queries to this, initial version was following already existing encoding style.

crates/context/config/src/client.rs
```diff
@@ -228,12 +231,16 @@ impl Client<AnyTransport> {
                         eyre::bail!("missing account id for `{}` signer", network);
                     };
 
+                    let access_key: PrivateKeySigner =
+                        PrivateKeySigner::from_str(&credentials.secret_key.clone())
```
miraclx: shouldn't need to clone here

```suggestion
                        PrivateKeySigner::from_str(&credentials.secret_key)
```

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -258,39 +258,23 @@ impl Network {
...
             });
+        };
+
+        if receipt.status() {
+            return Ok(return_data.to_vec());
         }
+
+        Err(EvmError::Custom {
+            operation: ErrorOperation::Mutate,
+            reason: format!("Transaction failed"),
```
miraclx: use `.to_owned()` / `.to_string()` when there's no variables present

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated by @alenmestrov

crates/context/config/src/client/env/proxy/types/evm.rs
frdomovic: Updated with TryFrom for 2 conversions with error handling

crates/context/config/src/client/env/proxy/types/evm.rs
frdomovic: Added error exit if it fails to convert to address

crates/context/config/src/client/env/proxy/mutate.rs
```diff
@@ -216,7 +216,9 @@ impl Method<Evm> for Mutate {
         let address = signer.address();
         let ecdsa_public_key = address.into_word();
 
-        let proxy_request_data: Vec<u8> = (&self.raw_request).into();
+        let proxy_request_data: Vec<u8> = (&self.raw_request)
+            .try_into()
```
miraclx: so you don't have to clone unnecessarily in the `TryFrom` impl

```suggestion
        let proxy_request_data: Vec<u8> = self.raw_request
            .try_into()
```

I see you probably NEED this because of L226, just move it above this line (to L218)

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -188,152 +188,182 @@ impl From<SolProposalAction> for ProposalAction {
...
+                        let param = Param {
+                            name: format!("param{}", index),
+                            kind: param_type,
+                            internal_type: None,
+                        };
+                        Ok((token, param))
+                    })
+                    .collect::<Result<Vec<_>, _>>()?
+                    .into_iter()
+                    .unzip();
```
miraclx: did you try

```suggestion
                    .collect::<Result<(Vec<_>, Vec<_>), _>>()?;
```

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -347,35 +377,55 @@ impl From<&ProxyMutateRequest> for SolRequestKind {
...
+                let signer_id: [u8; 32] = approval
+                    .signer_id
+                    .rt()
+                    .map_err(|_| "Invalid signer ID".to_string())?;
                 let proposal_approval = SolProposalApprovalWithSigner {
                     proposalId: B256::from(proposal_id),
                     userId: B256::from(signer_id),
                 };
-                SolValue::abi_encode(&proposal_approval)
+                Ok(SolValue::abi_encode(&proposal_approval))
```
miraclx: `proposal_approval.abi_encode()` and others in this file

crates/context/config/src/client/protocol/evm.rs
```diff
@@ -215,81 +202,81 @@ impl Network {
...
-                    json!("latest"),
-                ];
-
-                return Err(EvmError::Custom {
-                    operation: ErrorOperation::Mutate,
-                    reason: "Transaction failed".to_string(),
-                });
+            match call_result {
+                Ok(_) => println!("eth_call succeeded but transaction failed - strange!"),
+                Err(e) => println!("eth_call error details: {}", e),
```
miraclx: instead of 3 separate calls since we can't guarantee the simulation will accurately represent the transaction nor the call after since they'd target different blocks, why not:

1. construct transaction (since `send_transaction` only needs a reference)
2. call `send_transaction`, returning the receipt
3. call `call` with `block_id` from the receipt to get the result

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +188,245 @@ impl From<SolProposalAction> for ProposalAction {
...
-        key: Vec<u8>,
-        value: Vec<u8>,
-    },
-    DeleteProposal(ProposalId),
+impl TryFrom<ProposalAction> for SolProposalAction {
+    type Error = String;
+
+    fn try_from(action: ProposalAction) -> Result<Self, Self::Error> {
+        let kind: <SolProposalActionKind as SolType>::RustType = action.clone().into();
+        let data: Result<Vec<u8>, Self::Error> = match action {
```
miraclx: we shouldn't need to keep a result variable since you return result, use `?`

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +188,245 @@ impl From<SolProposalAction> for ProposalAction {
...
+
+                let call_data = function
+                    .encode_input(&tokens)
+                    .map_err(|_| "Encoding error".to_string())?;
+
+                let contract_address =
+                    Address::from_str(&receiver_id).map_err(|_| "Invalid address".to_string())?;
+                let data = SolValue::abi_encode(&(contract_address, call_data, amount));
+
+                Ok(data.into())
```
miraclx:
```suggestion
                (contract_address, call_data, amount).abi_encode()
```

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +188,245 @@ impl From<SolProposalAction> for ProposalAction {
...
+                    name: method_name,
+                    inputs,
+                    outputs: vec![],
+                    constant: Some(false),
+                    state_mutability,
+                };
+
+                let call_data = function
+                    .encode_input(&tokens)
+                    .map_err(|_| "Encoding error".to_string())?;
```
miraclx: consider using `.wrap_err` since we now return `eyre::Report` instead of `.map_err` throughout these `TryFrom` methods so we don't lose the cause of the error

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +188,245 @@ impl From<SolProposalAction> for ProposalAction {
...
+                                ParamType::Array(Box::new(ParamType::Uint(256))),
+                            ),
+                            "tuple" => (
+                                Token::Tuple(
+                                    serde_json::from_str(value)
+                                        .map_err(|_| "Invalid tuple".to_string())?,
+                                ),
+                                ParamType::Tuple(vec![]),
+                            ),
+                            _ => return Err(format!("Unsupported type: {}", key)),
```
miraclx: this can be `bail!` after the refactor

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +188,245 @@ impl From<SolProposalAction> for ProposalAction {
...
+                        };
+                        let param = Param {
+                            name: format!("param{}", index),
+                            kind: param_type,
+                            internal_type: None,
+                        };
+                        Ok((token, param))
+                    })
+                    .collect::<Result<Vec<_>, _>>()?
```
miraclx: if `bail!` doesn't work, as I suspect, just make this

```suggestion
                    .collect::<eyre::Result<_>>()?
```

crates/context/config/src/client/env/proxy/mutate.rs
frdomovic: Not sure exactly what to do here as proposed change doesn't work, moved proposal kind to L217

crates/context/config/src/client/env/proxy/types/evm.rs
frdomovic: Updated

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -1,12 +1,16 @@
...
+use alloy_sol_types::{sol, SolType, SolValue};
+use ethabi::{Function, Param, ParamType, Token};
+use eyre::Context;
+
+use crate::repr::{Repr, ReprTransmute};
 use crate::types::{Identity, ProposalId, SignerId};
-use crate::{Proposal, ProposalAction};
+use crate::{Proposal, ProposalAction, ProxyMutateRequest};
 
 sol! {
```
miraclx, line 11: standardize the indentation here, should it be 2 or 4?

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -63,9 +136,8 @@ impl From<SolProposalAction> for ProposalAction {
                     SolValue::abi_decode(&action.data, false).expect("Invalid external call data");
```
miraclx: we have an `expect` here, make this `TryFrom`

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +189,218 @@ impl From<SolProposalAction> for ProposalAction {
...
-        value: Vec<u8>,
-    },
-    DeleteProposal(ProposalId),
+impl TryFrom<ProposalAction> for SolProposalAction {
+    type Error = eyre::Report;
+
+    fn try_from(action: ProposalAction) -> Result<Self, Self::Error> {
+        let kind: <SolProposalActionKind as SolType>::RustType = action.clone().into();
+
+        let data: Vec<u8> = match action {
```
miraclx:
```suggestion
        let data = match action {
```

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +189,218 @@ impl From<SolProposalAction> for ProposalAction {
...
+                    constant: Some(false),
+                    state_mutability,
+                };
+
+                let call_data = function.encode_input(&tokens).wrap_err("Encoding error")?;
+
+                let contract_address =
+                    Address::from_str(&receiver_id).wrap_err("Invalid address")?;
+
+                (contract_address, call_data, amount).abi_encode().into()
```
miraclx: `abi_encode` already returns `Vec<u8>`

```suggestion
                (contract_address, call_data, amount).abi_encode()
```

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +189,218 @@ impl From<SolProposalAction> for ProposalAction {
...
+            ProposalAction::Transfer {
+                receiver_id,
+                amount,
+            } => {
+                let data = TransferData {
+                    recipient: Address::from_str(&receiver_id)
+                        .wrap_err(format!("Invalid receiver address format"))?,
+                    amount: U256::from(amount),
+                };
+                data.abi_encode().into()
```
miraclx: and others

crates/context/config/src/client/env/proxy/types/evm.rs
```diff
@@ -117,23 +189,218 @@ impl From<SolProposalAction> for ProposalAction {
...
+            }
+            ProposalAction::SetNumApprovals { num_approvals } => num_approvals.abi_encode().into(),
+            ProposalAction::SetActiveProposalsLimit {
+                active_proposals_limit,
+            } => active_proposals_limit.abi_encode().into(),
+            ProposalAction::SetContextValue { key, value } => {
+                let data = ContextValueData {
+                    key: key.to_vec().into(),
+                    value: value.to_vec().into(),
```
miraclx: `Bytes: From<Box<[u8]>>`

```suggestion
                    key: key.into(),
                    value: value.into(),
```

crates/context/config/src/client/env/proxy/mutate.rs
miraclx: It wouldn't work because the `TryFrom` impl was targeting `&T` instead of `T`, that needed to change too

Resolved in [`e208515` (#1146)](https://github.com/calimero-network/core/pull/1146/commits/e2085156928cf52b1b948a713145390549f277f3)

crates/context/config/src/client/env/proxy/types/evm.rs
frdomovic: Screenshot 2025-03-18 at 13 40 25

crates/context/config/src/client/protocol/evm.rs
frdomovic: Updated

miraclx (approved):

miraclx:
Please check to make sure there's no dangling `dbg!` / `println!` statements.

PR #1158, opened 2025-03-19
feat(runtime): implement actor model

petarjuki7:
# Actor Model - Runtime

Actor refactor of the Runtime crate.

## Test plan

Tested locally, no e2e test updates needed

## Documentation update
N/A


crates/runtime/src/lib.rs
```diff
@@ -15,6 +22,64 @@ pub use constraint::Constraint;
...
+pub struct ExecuteMsg {
+    pub blob: Vec<u8>,
+    pub method_name: String,
+    pub context: VMContext,
+    pub limits: VMLimits,
+    pub context_id: ContextId,
+}
+
+pub struct RuntimeManager {
+    pub tasks: BTreeMap<ContextId, Notify>,
```
miraclx:
```suggestion
    pub tasks: BTreeMap<ContextId, Arc<Notify>>,
```

should fix your lifetime issues

but now that I think about it, let's make this a `Mutex<()>` instead of `Notify`

crates/runtime/src/lib.rs
```diff
@@ -15,6 +22,64 @@ pub use constraint::Constraint;
...
 pub type RuntimeResult<T, E = VMRuntimeError> = Result<T, E>;
 
+#[derive(Message)]
+#[rtype(result = "RuntimeResult<(Outcome, InMemoryStorage)>")]
+pub struct ExecuteMsg {
+    pub blob: Vec<u8>,
+    pub method_name: String,
+    pub context: VMContext,
+    pub limits: VMLimits,
+    pub context_id: ContextId,
```
miraclx: `context_id` is already in `VMContext`

crates/runtime/src/lib.rs
```diff
@@ -15,6 +22,64 @@ pub use constraint::Constraint;
 
 pub type RuntimeResult<T, E = VMRuntimeError> = Result<T, E>;
 
+#[derive(Message)]
+#[rtype(result = "RuntimeResult<(Outcome, InMemoryStorage)>")]
+pub struct ExecuteMsg {
+    pub blob: Vec<u8>,
+    pub method_name: String,
+    pub context: VMContext,
+    pub limits: VMLimits,
```
miraclx: I'd move `VMLimits` to `RuntimeManager`'s constructor

miraclx (dismissed):

crates/runtime/src/lib.rs
```diff
@@ -22,61 +23,61 @@ pub use constraint::Constraint;
...
 }
 
 impl Actor for RuntimeManager {
     type Context = Context<Self>;
 }
 
+impl RuntimeManager {
+    pub fn new(tasks: BTreeMap<ContextId, Arc<Mutex<()>>>, limits: VMLimits) -> Self {
+        RuntimeManager { tasks, limits }
```
miraclx: surely we don't need the constructor to pass the BTreeMap instance, create an empty one here instead

crates/runtime/src/lib.rs
```diff
@@ -22,61 +23,61 @@ pub use constraint::Constraint;
...
 
-    fn handle(&mut self, msg: ExecuteMsg, ctx: &mut Self::Context) -> Self::Result {
-        let notify = self
+    fn handle(&mut self, msg: ExecuteMsg, _ctx: &mut Self::Context) -> Self::Result {
+        let mutex = self
             .tasks
-            .entry(msg.context_id)
-            .or_insert_with(|| Notify::new());
+            .entry(msg.context.context_id.into())
+            .or_insert_with(|| Arc::new(Mutex::new(())))
```
miraclx:
```suggestion
            .or_default()
```

crates/runtime/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+use std::collections::BTreeMap;
+use std::sync::Arc;
+
+use actix::prelude::*;
+use calimero_primitives::context::ContextId;
+use calimero_utils_actix::global_runtime;
+use store::InMemoryStorage;
```
miraclx: move this line to L22 (since it's not an external crate), and if you can do the same for the `use crate::` ones on L11-14, that'll be great

you can remove the `crate::` prefix as well, we're already at the root, it's unnecessary, thanks

crates/node/src/lib.rs
```diff
@@ -40,6 +42,7 @@ use eyre::{bail, eyre, OptionExt, Result as EyreResult};
 use libp2p::gossipsub::{IdentTopic, Message, TopicHash};
 use libp2p::identity::Keypair;
 use rand::{thread_rng, Rng};
+use runtime_compat::RuntimeCompatStore;
```
miraclx: move this back to L54

crates/node/src/lib.rs
```diff
@@ -529,15 +543,22 @@ impl Node {
 
         let mut store = self.store.clone();
 
-        let mut storage = RuntimeCompatStore::new(&mut store, context.id);
+        let storage = Box::new(RuntimeCompatStore::new(&mut store, context.id));
+
+        let exec_msg = ExecuteMsg {
```
miraclx: can we rename this? both the variable and the struct

crates/node/src/lib.rs
```diff
@@ -557,6 +578,7 @@ impl Node {
                 self.ctx_manager.save_context(context)?;
             }
 
+            // InMemoryStorage doesn't have .is_empty() or .commit() ?
             if !storage.is_empty() {
                 storage.commit()?;
             }
```
miraclx: after you change `Handler<Execute>` to also return the `Box<dyn Storage>` that was passed in

we'll need to downcast this back to `RuntimeCompatStore`, you can make [`Reflect`](https://github.com/calimero-network/core/blob/0cc661f4fd4d2f2576d3fc9b76178db2ade8f214/crates/primitives/src/reflect.rs#L34) a supertrait of `Storage`

which should let you call

```rust
storage.downcast_box::<RuntimeCompatStore>()
```

crates/runtime/Cargo.toml
```diff
@@ -7,9 +7,11 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
+actix.workspace = true
 borsh = { workspace = true, features = ["derive"] }
 fragile.workspace = true
 ouroboros.workspace = true
+tokio.workspace = true
```
miraclx: ~nit: order

crates/runtime/src/lib.rs
```diff
@@ -12,14 +14,74 @@ mod memory;
...
+        Box::pin(future)
+    }
+}
+
 pub fn run(
     code: &[u8],
     method_name: &str,
     context: VMContext,
-    storage: &mut dyn Storage,
+    storage: Box<dyn Storage>,
```
miraclx: you shouldn't need this change, we only need to `Box` because of the `'static` requirement of [`Addr::send`](https://docs.rs/actix/latest/actix/struct.Addr.html#method.send)

crates/runtime/src/logic.rs
```diff
@@ -65,7 +65,7 @@ pub struct VMLimits {
 
 #[derive(Debug)]
 pub struct VMLogic<'a> {
-    storage: &'a mut dyn Storage,
+    storage: Box<dyn Storage>,
```
miraclx: or any of these

crates/runtime/src/logic.rs
```diff
@@ -80,7 +80,7 @@ pub struct VMLogic<'a> {
 }
 
 impl<'a> VMLogic<'a> {
-    pub fn new(storage: &'a mut dyn Storage, context: VMContext, limits: &'a VMLimits) -> Self {
+    pub fn new(storage: Box<dyn Storage>, context: VMContext, limits: &'a VMLimits) -> Self {
```
miraclx: ^

crates/runtime/src/lib.rs
```diff
@@ -41,37 +40,34 @@ impl Actor for RuntimeManager {
...
     fn handle(&mut self, msg: ExecuteMsg, _ctx: &mut Self::Context) -> Self::Result {
         let mutex = self
             .tasks
             .entry(msg.context.context_id.into())
-            .or_insert_with(|| Arc::new(Mutex::new(())))
+            .or_default()
             .clone();
 
-        let mut storage = InMemoryStorage::default();
+        let storage = msg.storage;
```
miraclx: use `&msg.store` on L69 instead

crates/runtime/src/lib.rs
```diff
@@ -4,31 +4,30 @@ use std::sync::Arc;
...
+use errors::{FunctionCallError, VMRuntimeError};
+use logic::{Outcome, VMContext, VMLimits, VMLogic, VMLogicError};
+use memory::WasmerTunables;
+use store::Storage;
 
 pub type RuntimeResult<T, E = VMRuntimeError> = Result<T, E>;
 
 #[derive(Message, Debug)]
-#[rtype(result = "(RuntimeResult<Outcome>, InMemoryStorage)")]
+#[rtype(result = "RuntimeResult<Outcome>")]
```
miraclx:
```suggestion
#[rtype(result = "(RuntimeResult<Outcome>, Box<dyn Storage>)")]
```

crates/node/src/lib.rs
```diff
@@ -581,7 +582,11 @@ impl Node {
                 self.ctx_manager.save_context(context)?;
             }
 
-            // InMemoryStorage doesn't have .is_empty() or .commit() ?
+            let storage = storage
+                .downcast_box::<RuntimeCompatStore>()
+                .ok()
+                .ok_or_eyre("fatal: received store we didn't send??")?;
+
```
petarjuki7: Don't we have this two times now? Also in lines from L563 - L565

miraclx (approved):
Tested locally, works. Thanks!

PR #1159, opened 2025-03-20
chore(e2e): test ports

miraclx:
add a CI check to get a feel for how ports behave in CI


miraclx:
bump

miraclx:
This is so weird..

The port check passes, but still spinning up the node doesn't work. Which means this PR still doesn't account for the cause of the error.

Also, e2e tests doesn't exit when the node could not be spun up, leading to it waiting infinitely, we should fix that.

miraclx:
Let's get this in now, while we continue the investigation.

chefsale (approved):

PR #1160, opened 2025-03-21
feat(workflows): add pr description check workflow

frdomovic:
# Feat(workflows): PR description verification

This workflow should notify when PR's are not following engineering agreements and block PR's from being merged.
Each PR should include:
- title
- description
- test plan
- documentation update

With this updates each PR should have clear description of changes and how to test introduced changes.

## Test plan

Tested with [this](https://github.com/frdomovic/stellar-json-rpc) repository. 
Screenshot 2025-03-21 at 12 35 07

## Documentation update

No updates.

### DIFF
--- a/.github/workflows/pr-description-check.yml
+++ b/.github/workflows/pr-description-check.yml
@@ -1,0 +1,45 @@
+name: PR Description Check
+on:
+  pull_request:
+    types: [opened, edited]
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  check-description:
+    permissions: write-all
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR Description
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const requiredSections = [
+              "## Test plan",
+              "## Documentation update"
+            ];
+            
+            const prBody = context.payload.pull_request.body || "";
+
+            function isValidDescription(body) {
+              return requiredSections.every(section => 
+                body.includes(section) && 
+                body.split(section)[1].trim().length > 10
+              );
+            }
+
+            if (!isValidDescription(prBody)) {
+              github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.payload.pull_request.number,
+                body: "PR description does not follow the required template. Please update it to include:\n\n" +
+                      "`## Description`\n" +
+                      "`## Test plan`\n" +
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-description-check.yml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    types: [opened, edited]
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  check-description:
+    permissions: write-all
```
fbozic, line 13: Add `name: Check Description` here

.github/workflows/pr-description-check.yml
```diff
@@ -0,0 +1,45 @@
+name: PR Description Check
```
fbozic: Maybe just `name: PR` so we can add more checks as jobs in the future for PR validation

miraclx (approved):

.github/workflows/pr-description-check.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR Description
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const requiredSections = [
+              "## Test plan",
+              "## Documentation update"
+            ];
```
miraclx, line 23: there's no `## Description` here, even though the report mentions it

.github/workflows/pr-description-check.yml
```diff
@@ -0,0 +1,46 @@
...
+              github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.payload.pull_request.number,
+                body: "PR description does not follow the required template. Please update it to include:\n\n" +
+                      "`## Description`\n" +
+                      "`## Test plan`\n" +
+                      "`## Documentation update`\n\n" +
+                      "❌ PR merge is blocked until this is fixed."
+              });
```
miraclx, line 44: this won't remove the old comment, just add a new one

Ideally we reuse patterns already established for this:

https://github.com/calimero-network/core/blob/bd5d080dc056db3b026d0f89653cdfe5414e3940/.github/workflows/e2e-tests.yml#L267-L274

Deleting the comment once all is good.

miraclx: Also, don't duplicate the list of required sections. Try putting them in a variable that both the validate and report steps use.

PR #1161, opened 2025-03-24
ci(pr-description-check): only 1 comment per PR and comment cleanup

frdomovic:
## Description
Updated comments mentioned [here](https://github.com/calimero-network/core/pull/1160). Introduced new changes which include:
- proper required items 
- only 1 comment per PR with no commenting on each edit (previous behaviour)
- deleting comment after description edit workflow passed  

## Test plan

Was tested with private repository.

## Documentation update

No documentation update needed here.

### DIFF
--- a/.github/workflows/pr-description-check.yml
+++ b/.github/workflows/pr-description-check.yml
@@ -33,7 +33,7 @@ jobs:
               core.setOutput("comment_body", 
                 `**PR description is missing required sections!**\n\n` +
                 requiredSections.map(s => `${s}`).join("\n") +
-                `\n\n**PR merge is blocked until this is fixed.**`
+                `\n\n**❌ PR merge is blocked until this is fixed.**`
               );
             } else {
               core.setOutput("is_invalid", "false");
--- a/.github/workflows/pr-description-check.yml
+++ b/.github/workflows/pr-description-check.yml
@@ -31,9 +31,11 @@ jobs:
 
           if [[ ${#missing_sections[@]} -gt 0 ]]; then
             echo "comment_action=recreate" >> "$GITHUB_OUTPUT"
+            echo "is_valid=false" >> "$GITHUB_OUTPUT"
             echo "comment_content=**PR is missing required sections:** <br><br>${missing_sections[*]}.<br><br>Please update the PR description.<br><br>**❌ PR merge is blocked until this is fixed.**" >> "$GITHUB_OUTPUT"
           else
             echo "comment_action=delete" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-description-check.yml
```diff
@@ -10,37 +11,45 @@ permissions:
...
-                body: "PR description does not follow the required template. Please update it to include:\n\n" +
-                      "`## Description`\n" +
-                      "`## Test plan`\n" +
-                      "`## Documentation update`\n\n" +
-                      "❌ PR merge is blocked until this is fixed."
-              });
-              core.setFailed("PR description is missing required sections.");
-            }
+      - name: Fail job if PR description is invalid
+        if: needs.check-description.outputs.comment_action == 'recreate'
```
fbozic: Let's maybe add additional output called `is_valid` or something like that, instead of checking if `== recreate`

frdomovic: Added output is_valid and checks "== 'false'" as GITHUB output is string not a boolean so cant be short check. (as tested  [here](https://github.com/frdomovic/stellar-json-rpc).

miraclx (approved):

.github/workflows/pr-description-check.yml
```diff
@@ -10,37 +11,47 @@ permissions:
...
-            const requiredSections = [
-              "## Test plan",
-              "## Documentation update"
-            ];
-            
-            const prBody = context.payload.pull_request.body || "";
+      - name: Validate Description
+        id: validate_description
+        run: |
+          required_sections=("## Description" "## Test plan" "## Documentation update")
```
miraclx: would be nice if this was broken out elsewhere for visibility, like an env var

https://stackoverflow.com/a/73274812/9806233

Can even break it up in to multiple lines

https://stackoverflow.com/a/69811126/9806233

.github/workflows/pr-description-check.yml
```diff
@@ -10,37 +11,47 @@ permissions:
 jobs:
   check-description:
     name: Check Description
-    permissions: write-all
     runs-on: ubuntu-latest
+    outputs:
+      comment_action: ${{ steps.validate_description.outputs.comment_action }}
+      comment_content: ${{ steps.validate_description.outputs.comment_content }}
```
fbozic, line 17: aren't we missing here `is_valid`?

.github/workflows/pr-description-check.yml
frdomovic: Split into env

.github/workflows/pr-description-check.yml
frdomovic: updated

PR #1163, opened 2025-03-26
feat: Implemented the application watch subcommand

cy4n1d3-p1x3l:
# Description

This PR implements the feature to continuously install the path if it is being watched.

Resolves #706.

## Usage 

The command is available as ```meroctl --node-name <node-name> app install --watch --path <path-to-the-file>```
It also logs that a new file is in use now.

PR #1164, opened 2025-03-26
ci(pr-description-check): escape special characters

frdomovic:
## Description

Issue noticed in [this](https://github.com/calimero-network/core/pull/1163) PR where "` ` `" is used and it breaks the workflow

## Test plan

Tested in private repository [here](https://github.com/frdomovic/stellar-json-rpc/pull/21)

## Documentation update

N/A

### DIFF
--- a/.github/workflows/pr-description-check.yml
+++ b/.github/workflows/pr-description-check.yml
@@ -27,23 +27,26 @@ jobs:
         run: |
           # Read required sections from env var, one per line
           readarray -t required_sections <<< "$REQUIRED_SECTIONS"
-          pr_body="${{ github.event.pull_request.body }}"
+          
+          # Encode PR body in base64 to avoid issues with special characters
+          pr_body_base64=$(echo -n "${{ github.event.pull_request.body }}" | base64 -w0)
           missing_sections=()
 
           for section in "${required_sections[@]}"; do
-            if [[ -z $(echo "$pr_body" | grep -i "$section") ]]; then
+            # Decode base64 and search for the section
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-description-check.yml
```diff
@@ -25,25 +25,24 @@ jobs:
...
         run: |
-          # Read required sections from env var, one per line
           readarray -t required_sections <<< "$REQUIRED_SECTIONS"
-          pr_body="${{ github.event.pull_request.body }}"
+          pr_body_base64=$(echo -n "${{ github.event.pull_request.body }}" | base64 -w0)
           missing_sections=()
 
           for section in "${required_sections[@]}"; do
-            if [[ -z $(echo "$pr_body" | grep -i "$section") ]]; then
+            if ! echo "$pr_body_base64" | base64 --decode | grep -q -i "$section"; then
```
miraclx, line 33: if the issue was substitution, base64-encoding shouldn't be necessary, right?

```suggestion
          pr_body="${{ github.event.pull_request.body }}"
          missing_sections=()

          for section in "${required_sections[@]}"; do
            if ! echo "$pr_body" | grep -q -i "$section"; then
```

frdomovic: Tested this in private repository check [here](https://github.com/frdomovic/stellar-json-rpc/actions/runs/14197121277/job/39774810647?pr=22). It breaks again because it tries to execute the command written in description

miraclx (approved):

.github/workflows/pr-description-check.yml
miraclx: weird

PR #1165, opened 2025-03-26
docs(contributing): add rust style guide

petarjuki7:
# Update the contributing guide

## Description

Add a rust style section.

## Test plan

N/A

## Documentation update

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ There are several ways you can contribute:
 Contributions are managed via Issues and Pull Requests (PRs). Here are some
 general guidelines:
 
-- Read the Rust style guide [STYLE]!
+- Read our [Rust style guide][STYLE] to ensure compatibility with the rest of our codebase!
 
 - Before creating a new Issue or PR, search for [existing ones][Issues].
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ There are several ways you can contribute:
 Contributions are managed via Issues and Pull Requests (PRs). Here are some
 general guidelines:
 
-- Read the Rust style guide bellow!
+- Read the Rust style guide [STYLE]!
 
 - Before creating a new Issue or PR, search for [existing ones][Issues].
 
@@ -75,10 +75,6 @@ would like to work on an issue, please follow these steps:
 
 By following this process, we can avoid duplication of efforts and ensure clear
 communication among all contributors.
-
-## Rust style guide
-
-### Refer to [STYLE] for the complete Rust style conventions used in our codebase
 
 ## Creating a New Issue
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,10 +77,69 @@ communication among all contributors.
 
 ## Rust Style Guide
 
-### Formatting
+## Formatting
 
-- Use rustfmt with nightly features to maintain consistent code formatting.
+- Use rustfmt with nightly features to maintain consistent code formatting:
+  
+  ```bash
+  cargo +nightly fmt
+  ```
+
 - Sort Cargo.toml dependencies alphabetically.
+
+- Organize imports according to the `StdExternalCrate` pattern:
+  1. Standard library
+  2. External crates
+  3. Symbols from local crate & parent module
+  4. Local modules definition
+  5. Symbols from local modules (optional)
+
+```rust
+// Standard library
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::{mem, time};
+
+// External crates
+use serde::{Serialize, Deserialize};
+use tokio::sync::{oneshot, RwLock};
+
+// Symbols from local crate & parent module
+use crate::{common, Node};
+use super::Shared;
+
+// Local modules definition
+mod config;
+mod types;
+
+// Symbols from local modules (optional)
+use config::ContextConfig;
+use types::BroadcastMessage;
+```
+
+- Use module import granularity. This means not grouping imports from the same crate together.
+
+**NOT ALLOWED:**
+
+```rust
+use core::{
+    future::{pending, Future},
+    pin::Pin,
+    str,
+};
+use std::{thread, time::Duration};
+```
+
+**ALLOWED:**
+
+```rust
+use core::future::pending;
+use core::future::Future;
+use core::pin::Pin;
+use core::str;
+use std::thread;
+use std::time::Duration;
+```
 
 ### Module Organization
 
@@ -111,6 +170,15 @@ crates/meroctl/src/cli/app/list.rs
 
 ### Error Handling
 
+- We use the eyre crate extensively in our code, and import it as follows:
+  
+  ```rust
+  use eyre::{Result as EyreResult};
+  ```
+
+- Almost no unwrapping (acceptable in tests and possibly when dealing with thread join handlers).
+- If unwrapping is absolutely necessary, explain why with a comment.
+- On values that may return errors, use `.map_err()` to map the error into the appropriate Error type used in that crate/module.
 - Almost no unwrapping (acceptable in tests and possibly when dealing with thread join handlers).
 - If unwrapping is absolutely necessary, explain why with a comment.
 - On values that may return errors, use `.map_err()` to map the error into the appropriate Error type used in that crate/module.
@@ -124,42 +192,35 @@ crates/meroctl/src/cli/app/list.rs
 // NOT RECOMMENDED:
 if some_condition {
     // ... (lots of code)
+} else {
+    return Err(YourError::Something);
 }
-
 // RECOMMENDED:
 if !some_condition {
     return Err(YourError::Something);
 }
 // Continue with main code path...
-```
-
-- Extract values from options using `if let` when possible:
-
-```rust
-// RECOMMENDED:
-if let Some(value) = optional_value {
-    // Use value directly
-}
-
-// INSTEAD OF:
-match optional_value {
-    Some(value) => {
-        // Use value
-    },
-    None => {},
-}
-```
 
 ### Code Organization
 
 - Break functions into smaller parts if they become too large.
 - Place reusable functions in a `commons.rs` file or similar.
 - Put structs needed by multiple parts of the codebase into `primitives` files.
+- We try to avoid using fully qualified names, prefer using imports.
+
+```rust
+// NOT RECOMMENDED:
+std::fs::File::open("data.txt")
+
+// RECOMMENDED:
+use std::fs::File;
+File::open("data.txt")
+```
 
 ### Naming Conventions
 
-- Types shall be `UpperCamelCase`
-- Enum variants shall be `UpperCamelCase`
+- Types shall be `PascalCase`
+- Enum variants shall be `PascalCase`
 - Struct fields shall be `snake_case`
 - Function and method names shall be `snake_case`
 - Local variables shall be `snake_case`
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -73,7 +75,102 @@ would like to work on an issue, please follow these steps:
 By following this process, we can avoid duplication of efforts and ensure clear
 communication among all contributors.
 
-### Creating a New Issue
+## Rust Style Guide
+
+### Formatting
+
+- Use rustfmt with nightly features to maintain consistent code formatting.
+- Sort Cargo.toml dependencies alphabetically.
```
fbozic: Can we add command somewhere `cargo +nightly fmt`?

CONTRIBUTING.md
```diff
@@ -73,7 +75,102 @@ would like to work on an issue, please follow these steps:
...
+
+And we would have these individual files:
+
+```
+crates/meroctl/src/cli/app/get.rs
+crates/meroctl/src/cli/app/install.rs
+crates/meroctl/src/cli/app/list.rs
+```
+
+### Error Handling
```
fbozic: Should we note that we use Eyre everywhere and how we import EyreResult with `as`?

CONTRIBUTING.md
```diff
@@ -73,7 +75,102 @@ would like to work on an issue, please follow these steps:
...
+// INSTEAD OF:
+match optional_value {
+    Some(value) => {
+        // Use value
+    },
+    None => {},
+}
+```
+
+### Code Organization
```
fbozic: Can we mention that we try to avoid FQN and we prefer imports?

CONTRIBUTING.md
petarjuki7: Is that true in general?

fbozic: Probably for runtime and a few other components it is not true but we still use it a lot, but I still think it is worth mentioning.

CONTRIBUTING.md
```diff
@@ -73,7 +75,102 @@ would like to work on an issue, please follow these steps:
...
+
+- Break functions into smaller parts if they become too large.
+- Place reusable functions in a `commons.rs` file or similar.
+- Put structs needed by multiple parts of the codebase into `primitives` files.
+
+### Naming Conventions
+
+- Types shall be `UpperCamelCase`
+- Enum variants shall be `UpperCamelCase`
```
miraclx: `PascalCase` not `UpperCamelCase`

CONTRIBUTING.md
```diff
@@ -73,7 +75,102 @@ would like to work on an issue, please follow these steps:
...
+if some_condition {
+    // ... (lots of code)
+}
+
+// RECOMMENDED:
+if !some_condition {
+    return Err(YourError::Something);
+}
+// Continue with main code path...
+```
```
miraclx: it's only an issue when there's an `else` case

````suggestion
```rust
// NOT RECOMMENDED:
if some_condition {
    // ... (lots of code)
} else {
    return Err(YourError::Something);
}

// RECOMMENDED:
if !some_condition {
    return Err(YourError::Something);
}
// Continue with main code path...
```
````

File: CONTRIBUTING.md [hunk 6]
```diff
@@ -73,7 +75,102 @@ would like to work on an issue, please follow these steps:
...
+}
+
+// INSTEAD OF:
+match optional_value {
+    Some(value) => {
+        // Use value
+    },
+    None => {},
+}
+```
```
miraclx: I don't think we should have this here, match is often times preferable, depending on the case.

the only think I can think of that closely resembles this, is `let..else`, but that's only for the short-circuiting case.

CONTRIBUTING.md
```diff
@@ -104,12 +201,13 @@ We use the ["fork-and-pull"][GitPR] Git workflow:
 
 7. Tag a maintainer to review your PR.
 
+8. Make sure your PR follows our PR template (has to consist of a `Description`, `Test plan` and `Documentation update`)
```
miraclx:
```suggestion
8. Make sure your PR follows our PR template (has to consist of a `Description`, `Test plan` and `Documentation update` sections)
```

miraclx (commented):
also consider including imports organization, for which we use [`StdExternalCrate`](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#group_imports)

```
standard library

external crates

symbols from local crate & parent module

local modules definition

symbols from local modules (optional)
```

for example

```rust
use std::collections::HashMap;
use std::sync::Arc;
use std::{mem, time};

use serde::{Serialize, Deserialize};
use tokio::sync::{oneshot, RwLock};

use crate::{common, Node};
use super::Shared;

mod config;
mod types;

use config::ContextConfig;
use types::BroadcastMessage;
```

and we use an [import granularity](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=granular#imports_granularity) of module so this is not allowed:

```rust
use core::{
    future::{pending, Future},
    pin::Pin,
    str,
};
use std::{thread, time::Duration};
```

CONTRIBUTING.md
```diff
@@ -77,11 +77,70 @@ communication among all contributors.
...
+    str,
+};
+use std::{thread, time::Duration};
+```
+
+**ALLOWED:**
+
+```rust
+use core::future::pending;
+use core::future::Future;
```
miraclx:
```suggestion
use core::future::{pending, Future};
```

this is allowed

CONTRIBUTING.md
```diff
@@ -77,11 +77,70 @@ communication among all contributors.
 
 ## Rust Style Guide
 
-### Formatting
+## Formatting
```
miraclx:
```suggestion

## Rust Style Guide

### Formatting
```

CONTRIBUTING.md
```diff
@@ -77,11 +77,70 @@ communication among all contributors.
...
+
+- Use rustfmt with nightly features to maintain consistent code formatting:
```
miraclx: use headers for these so we can link to them

```suggestion
#### Use rustfmt with nightly features to maintain consistent code formatting:
```

also helps the weird indentation

CONTRIBUTING.md
```diff
@@ -111,6 +170,15 @@ crates/meroctl/src/cli/app/list.rs
 
 ### Error Handling
 
+- We use the eyre crate extensively in our code, and import it as follows:
+  
+  ```rust
+  use eyre::{Result as EyreResult};
+  ```
+
+- Almost no unwrapping (acceptable in tests and possibly when dealing with thread join handlers).
```
miraclx: any joining should be a handle-able issue, it's only acceptable when it's absolutely certain it won't fail, like integration's `ReprTransmute` where we have a bunch of `.expect("infallible conversion")`

CONTRIBUTING.md
```diff
@@ -111,6 +170,15 @@ crates/meroctl/src/cli/app/list.rs
...
+- If unwrapping is absolutely necessary, explain why with a comment.
```
miraclx: it's never necessary, just that the author has a lot of confidence in it never panicking

CONTRIBUTING.md
```diff
@@ -124,42 +192,35 @@ crates/meroctl/src/cli/app/list.rs
...
-    // Use value directly
-}
-
-// INSTEAD OF:
-match optional_value {
-    Some(value) => {
-        // Use value
-    },
-    None => {},
-}
```
miraclx: Suggest `let..else` over `if let..else`

https://github.com/calimero-network/core/pull/1165#discussion_r2016603244

https://github.com/calimero-network/core/pull/1141#discussion_r1987152728

miraclx:
Consider breaking up the rust style guide into a separate `STYLE.md` file, with `CONTRIBUTING.md` linking to it.

STYLE.md
```diff
@@ -0,0 +1,197 @@
+# Rust Style Guide
+
+## Formatting
+
+### Use rustfmt with nightly features to maintain consistent code formatting
+  
+  ```bash
+  cargo +nightly fmt
+  ```
```
miraclx, line 9: we no longer need indentation here

STYLE.md
```diff
@@ -0,0 +1,197 @@
...
+  use eyre::{Result as EyreResult};
+  ```
+
+### Almost no unwrapping (acceptable in tests and possibly when dealing with thread join handlers)
+
+### If unwrapping is absolutely necessary, explain why with a comment
+
+### On values that may return errors, use `.map_err()` to map the error into the appropriate Error type used in that crate/module
+
+### Almost no unwrapping (it's only acceptable when the author is absolutely certain it won't fail (explain why with a comment))
```
miraclx: now we have two "almost no unwrapping"

try something like

"employ maximum caution dealing with panic points (`.unwrap()`, `.expect(..)`, `assert!`, `panic!`, etc). Only introduce them when you have maximum confidence it will NEVER panic, or if it does, it's a fatal error from which we cannot recover and aborting the node is the best thing to do. In this case, introduce a comment stating why"

STYLE.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Struct fields shall be `snake_case`
+
+### Function and method names shall be `snake_case`
+
+### Local variables shall be `snake_case`
+
+### Macro names shall be `snake_case`
+
+### Constants (consts and immutable statics) shall be `SCREAMING_SNAKE_CASE`
```
miraclx: these can bullet points, since the whole section is rather small, and there are no examples, linking to the "Naming Conventions" header is good enough

STYLE.md
```diff
@@ -0,0 +1,197 @@
...
+crates/meroctl/src/cli/app/list.rs
+```
+
+## Error Handling
+
+### We use the eyre crate extensively in our code, and import it as follows
+  
+  ```rust
+  use eyre::{Result as EyreResult};
+  ```
```
miraclx: unindent

STYLE.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Formatting
+
+### Use rustfmt with nightly features to maintain consistent code formatting
+  
+  ```bash
+  cargo +nightly fmt
+  ```
+
+### Sort Cargo.toml dependencies alphabetically
```
miraclx: <code>\`Cargo.toml\`</code>

STYLE.md
```diff
@@ -0,0 +1,197 @@
...
+// RECOMMENDED:
+let Ok(val) = thing else {
+    do_else_1;
+    return;
+};
+func(val);
+do_ok_1;
+do_ok_2;
+do_ok_3;
+```
```
miraclx:
```suggestion
// NOT RECOMMENDED:
if let Ok(val) = thing {
    func(val);
    do_ok_1;
    do_ok_2;
    do_ok_3;
} else {
    do_else_1;
    return;
}

// RECOMMENDED:
let Ok(val) = thing else {
    do_else_1;
    return;
};

func(val);
do_ok_1;
do_ok_2;
do_ok_3;
```

CONTRIBUTING.md
```diff
@@ -73,7 +76,11 @@ would like to work on an issue, please follow these steps:
 By following this process, we can avoid duplication of efforts and ensure clear
 communication among all contributors.
 
-### Creating a New Issue
+## Rust style guide
+
+### Refer to [STYLE] for the complete Rust style conventions used in our codebase
```
miraclx: we shouldn't need a special section just to link to the style guide .._

CONTRIBUTING.md
```diff
@@ -29,6 +30,8 @@ There are several ways you can contribute:
 Contributions are managed via Issues and Pull Requests (PRs). Here are some
 general guidelines:
 
+- Read the Rust style guide bellow!
```
miraclx: we can link it here.

STYLE.md
petarjuki7: it already is, right?

CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ There are several ways you can contribute:
 Contributions are managed via Issues and Pull Requests (PRs). Here are some
 general guidelines:
 
-- Read the Rust style guide bellow!
+- Read the Rust style guide [STYLE]!
```
miraclx: when in preview mode, this line reads a bit weird, consider moving the link

STYLE.md
```diff
@@ -96,18 +96,16 @@ crates/meroctl/src/cli/app/list.rs
...
   
-  ```rust
-  use eyre::{Result as EyreResult};
-  ```
+```rust
+use eyre::{Result as EyreResult};
+```
 
-### Almost no unwrapping (acceptable in tests and possibly when dealing with thread join handlers)
+### Employ maximum caution dealing with panic points (.unwrap(), .expect(..), assert!, panic!, etc). Only introduce them when you have maximum confidence it will NEVER panic, or if it does, it's a fatal error from which we cannot recover and aborting the node is the best thing to do. In this case, introduce a comment stating why
```
miraclx:
```suggestion
### Employ maximum caution dealing with panic points (.unwrap(), .expect(..), assert!, panic!, etc).

Only introduce them when you have maximum confidence it will NEVER panic, or if it does, it's a fatal error from which we cannot recover and aborting the node is the best thing to do. In this case, introduce a comment stating why
```

CONTRIBUTING.md
petarjuki7: I think it should be fixed now

miraclx (approved):

STYLE.md
fbozic: still looks indented

STYLE.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Sort `Cargo.toml` dependencies alphabetically
+
+### Organize imports according to the `StdExternalCrate` pattern
+
+  1. Standard library
+  2. External crates
+  3. Symbols from local crate & parent module
+  4. Local modules definition
+  5. Symbols from local modules (optional)
```
fbozic: and this looks indented as well unnecessarily

PR #1168, opened 2025-03-31
Fix  General e2e improvements

alenmestrov:
## Description

- Introduced functionality to run multiple protocols at the same time
- Optimized application install
- Fixed context_storage_entries endpoint
- Brought back kv-store scenario
- Resolved [PR](https://github.com/calimero-network/core/pull/1142) comments 

## Test plan

Run e2e tests locally with one of the enabled protocols flag

## Documentation update

N/A

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -209,7 +209,7 @@ jobs:
             --output-dir ./e2e-tests/corpus \
             --merod-binary ./target/debug/merod \
             --meroctl-binary ./target/debug/meroctl \
-            --scenario ${{ matrix.protocol }}
+            --scenarios ${{ matrix.protocol }}
 
       - name: Upload Test Artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -72,27 +72,6 @@ jobs:
         run: |
           chmod +x target/debug/e2e-tests target/debug/merod target/debug/meroctl
 
-      - name: Generate ports
-        id: generate_ports
-        run: |
-          # Generate unique random numbers
-          random_numbers=()
-          while [ ${#random_numbers[@]} -lt 3 ]; do
-            num=$((RANDOM%37001 + 3000))
-            if [[ ! " ${random_numbers[@]} " =~ " ${num} " ]]; then
-              random_numbers+=($num)
-            fi
-          done
-          echo "Random numbers: ${random_numbers[@]}"
-          SWARM_PORT="${random_numbers[0]}"
-          SERVER_PORT="${random_numbers[1]}"
-          ICP_PORT="${random_numbers[2]}"
-          
-          # Set outputs for other steps
-          echo "SWARM_PORT=$SWARM_PORT" >> $GITHUB_OUTPUT
-          echo "SERVER_PORT=$SERVER_PORT" >> $GITHUB_OUTPUT
-          echo "ICP_PORT=$ICP_PORT" >> $GITHUB_OUTPUT
-
       - name: Download Contracts
         run: |
           echo "Downloading contracts, target version: $CALIMERO_CONTRACTS_VERSION"
@@ -108,7 +87,7 @@ jobs:
       - name: Deploy ICP Devnet
         if: matrix.protocol == 'icp'
         env:
-          ICP_PORT: ${{ steps.generate_ports.outputs.ICP_PORT }}
+          ICP_PORT: 4943
         run: |
           echo "ICP_PORT=$ICP_PORT"
           ./scripts/icp/deploy-devnet.sh
@@ -120,7 +99,7 @@ jobs:
       - name: Deploy Stellar Devnet
         if: matrix.protocol == 'stellar'
         env:
-          ICP_PORT: ${{ steps.generate_ports.outputs.ICP_PORT }}
+          ICP_PORT: 4943
         run: |
           echo "Installing stellar CLI..."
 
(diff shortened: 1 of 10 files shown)


frdomovic (approved):
lgtm

.github/workflows/e2e-tests.yml
```diff
@@ -74,37 +75,23 @@ jobs:
...
-          echo "SWARM_HOST=$SWARM_HOST" >> $GITHUB_OUTPUT
-          
-          # Update config.json
-          jq --arg swarmPort "$SWARM_PORT" \
-            --arg serverPort "$SERVER_PORT" \
-            --arg swarmHost "$SWARM_HOST" \
-            '.network.swarmHost = ($swarmHost) |
-            .network.startSwarmPort = ($swarmPort | tonumber) |
-            .network.startServerPort = ($serverPort | tonumber)' \
-            e2e-tests/config/config.json > updated_config.json
```
miraclx, line 99: we're not setting these anymore?

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
```diff
@@ -1,8 +1,9 @@
 use std::io::Cursor;
 
+use alloy::dyn_abi::{DynSolType, DynSolValue};
 use alloy_sol_types::SolValue;
 use candid::{Decode, Encode};
-use eyre::OptionExt;
+use eyre::{eyre, Result};
```
miraclx: if you're going to import `eyre::Result` into scope, don't shadow the std `Result`, use `EyreResult` instead. Otherwise, using `eyre::Result` is fine.

e2e-tests/src/steps/application_install.rs
```diff
@@ -127,24 +130,50 @@ impl ApplicationSource {
...
+                        let mut stream = stream;
+                        while let Some(chunk) = stream.next().await {
+                            let chunk =
+                                chunk.map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+                            tx.send(Ok(chunk))
+                                .await
+                                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+                        }
+                        Ok::<_, io::Error>(())
+                    });
```
miraclx: we shouldn't need to `spawn` or use a `channel` here, try [`split`](https://docs.rs/futures-util/latest/futures_util/io/trait.AsyncReadExt.html#method.split)-ting the `GzipDecoder` and joining two [`copy`](https://docs.rs/futures-util/latest/futures_util/io/fn.copy.html) tasks

```
task 1 := [ response       -> gzip write half ]
task 2 := [ gzip read half -> file            ]
```

.github/workflows/e2e-tests.yml
miraclx: while I agree they shouldn't be "necessary" (since we will pick an open port), we still print the ports in a later job, what should we do instead?

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
```diff
@@ -158,29 +158,51 @@ impl Method<Ethereum> for ContextStorageEntriesRequest {
...
+            Ok(entries
+                .into_iter()
+                .map(|entry| {
+                    if let DynSolValue::Tuple(fields) = entry {
+                        let all_bytes = fields[1]
+                            .as_bytes()
+                            .ok_or_else(|| eyre!("Failed to get bytes from field"))?;
+
+                        // Get key
+                        let key_len = all_bytes[31] as usize;
```
miraclx: wonder why they waste 30 bytes in front

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
```diff
@@ -158,29 +158,51 @@ impl Method<Ethereum> for ContextStorageEntriesRequest {
...
+        // Define the struct type as a tuple
+        let struct_type = "tuple(bytes,bytes)[]".parse::<DynSolType>()?;
+        // Decode using dynamic ABI decoder
+        let decoded = struct_type.abi_decode(&response)?;
+        // Convert the decoded value to our type
+        if let DynSolValue::Array(entries) = decoded {
+            Ok(entries
+                .into_iter()
+                .map(|entry| {
+                    if let DynSolValue::Tuple(fields) = entry {
```
miraclx: use `let..else` so we bail early

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
```diff
@@ -158,29 +158,51 @@ impl Method<Ethereum> for ContextStorageEntriesRequest {
...
-        }
 
-        Ok(entries)
+    fn decode(response: Vec<u8>) -> Result<Self::Returns> {
+        // Define the struct type as a tuple
+        let struct_type = "tuple(bytes,bytes)[]".parse::<DynSolType>()?;
+        // Decode using dynamic ABI decoder
+        let decoded = struct_type.abi_decode(&response)?;
+        // Convert the decoded value to our type
+        if let DynSolValue::Array(entries) = decoded {
```
miraclx: same here

https://github.com/calimero-network/core/blob/petarjuki7/contributing-guide-update/STYLE.md#use-letelse-for-deep-conditionals

e2e-tests/Cargo.toml
```diff
@@ -9,31 +9,34 @@ license.workspace = true
...
 nix = { version = "0.29.0", features = ["signal"] }
 rand.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 tokio = { workspace = true, features = ["fs", "io-util", "macros", "process", "rt", "rt-multi-thread", "time"] }
-url = { workspace = true }
-reqwest = { workspace = true }
+reqwest.workspace = true
+tokio-util.workspace = true
+tokio-stream.workspace = true
```
miraclx: ~nit: order

crates/context/config/src/client/env/proxy/query/context_storage_entries.rs
miraclx: that was what caught us off guard too, leading to us thinking we had bad data, looked into the raw data and what we expected was in there, just padded with wasted bytes we couldn't explain

alenmestrov: padding I guess

e2e-tests/src/steps/application_install.rs
```diff
@@ -131,37 +129,18 @@ impl ApplicationSource {
...
-                        }
-                        Ok::<_, io::Error>(())
-                    });
-
-                    let (write_result, process_result) = tokio::join!(write_to_file, process_input);
-                    write_result??;
-                    process_result??;
+                    let stream = response
+                        .bytes_stream()
+                        .map_err(|e| io::Error::new(io::ErrorKind::Other, e));
```
miraclx:
```suggestion
                        .map_err(io::Error::other);
```

miraclx (approved):

Issue #1118, opened 2025-02-18
Default selections for CLI variables


miraclx:
A lot of operations could benefit from having the ability to select defaults.

A default context, default identity in a context, etc;

- `context identity list <context-id>` instead of `context identity list <context-id>`
- `call <method>` instead of `call <context> <method> --as <who>` etc..

We can take advantage of the Alias API introduced in #1094 & #1114.

The alias name can be `default`.

The API for defining defaults can be;

```console
context alias add chat-app DJaNFHW9HSt23AJj4R2YA9cERZrAEwuRiBXp8cAAQ1U8
context use chat-app

context identity alias add jake CCG3HwvSFeVgsGkzkv1KLXfwoGKZky9m9FdKWa6vZGS
context identity use jake
```

but of course, the `alias` step need not be defined

```console
context use DJaNFHW9HSt23AJj4R2YA9cERZrAEwuRiBXp8cAAQ1U8

context identity use CCG3HwvSFeVgsGkzkv1KLXfwoGKZky9m9FdKWa6vZGS
```

and when a context has not been aliased or set as default, the required arguments to `identity use` becomes:

```console
context identity use CCG3HwvSFeVgsGkzkv1KLXfwoGKZky9m9FdKWa6vZGS --context DJaNFHW9HSt23AJj4R2YA9cERZrAEwuRiBXp8cAAQ1U8
```

And listing (#1117), should take this into account, showing `default`

chefsale:
Althought we haven’t officially launched the bounty program, discussed this issue with @rtb-12 and we agreed he will be implementing the default selections for CLI variables

Once the PRs are merged and KYC has been passed we will be providing an 50 USDC bounty for this issue.

This issue should be several hours of work and ideally implemented in the next month in worst case scenario, but would love to see it earlier if possible.

Good luck and looking forward to being part of the community and contributing more.

Thanks, Sandi.

cc: @miraclx @xilosada

Issue #1117, opened 2025-02-18
Add ability to list aliases


miraclx:
#1094 & #1114 introduced aliases to the node, and exposed them via

```console
meroctl context alias [ add | get | rm ]
meroctl context identity alias [ add | get | rm ]
```

#1094 previously introduced the ability to see associated aliases when you do `context list` or `context identity list`, but #1114 removed them due to searching inefficiency.

#1114 already introduced `ContextManager::list_aliases` which returns all the aliases, as well as the values they alias.

That should be used in future PRs to;

1. Reintroduce listing as it's own specialized `context alias list`.
2. Expose an alias listing API over the server. Which should help `meroctl` also have that feature.

miraclx:
https://github.com/calimero-network/core/pull/1114 later introduced alias listing to `merod`'s interactive CI. What's left is `meroctl`.

cy4n1d3-p1x3l:
Can i pick up this issue?

chefsale:
Go for it.

miraclx:
Further context on the behaviour of `alias list` (following the merge of #1171) and as mentioned in https://github.com/calimero-network/core/pull/1171#discussion_r2035783487:

```console
context identity list
    (lists identities in `default` context)

context identity list --all
    (lists identities in all contexts, but this has to conflict with "context" being specified)

context identity list -c 2XGCNdL88mGVRVnhJpxAM32J9PBiQ1pw4s3BgeFf2juw
    (lists identities in one specific context)
```

antonpaisov:
hey @cy4n1d3-p1x3l ! as the PR above is merged, do you want to give it a go now?

PR #1171, opened 2025-04-01
feat(merod, meroctl): introduce 'use' command for default aliases

rtb-12:
# Add Default Context Support with `use` Command

Added a `context use` command that sets a default context, allowing users to omit the context parameter in subsequent commands. Commands now automatically use the default context when none is specified, improving usability when working with the same context repeatedly. 

Fixes issue #1118


rtb-12:
@miraclx kindly review the PR

crates/meroctl/src/cli/call.rs
```diff
@@ -12,7 +12,7 @@ use serde_json::{json, Value};
 
 use crate::cli::Environment;
 use crate::common::{
-    do_request, fetch_multiaddr, load_config, multiaddr_to_url, resolve_alias, RequestType,
+    do_request, fetch_multiaddr, load_config, multiaddr_to_url, resolve_context,resolve_alias, RequestType,
```
miraclx: please run `cargo +nightly fmt`

crates/meroctl/src/cli/context/use_command.rs
miraclx, line 1: can we merge this with the alias command? so we don't need to break convention with the `_command` suffix.

plus, it pretty much pertains to aliases

crates/meroctl/src/common.rs
```diff
@@ -401,3 +401,27 @@ where
 
     Ok(ResolveResponse { alias, value })
 }
+
+
+pub async fn resolve_context(
```
miraclx: So we don't only need to `resolve_context`s, but also identity aliases, and eventually application aliases.

So we need a generalized form of this method. Use the design of `resolve_alias`' generics to guide.

crates/node/src/interactive_cli/identity.rs
```diff
@@ -43,9 +44,9 @@ enum AliasSubcommands {
         name: Alias<PublicKey>,
         /// The identity to create an alias for
         identity: PublicKey,
-        /// The context that the identity is a member of
-        #[arg(long, short)]
-        context: Alias<ContextId>,
+        /// The context that the identity is a member of (omit to use default context)
+        #[clap(long = "context", short = 'c')]
+        context: Option<Alias<ContextId>>,
```
miraclx:
```suggestion
        /// The context that the identity is a member of (omit to use default context)
        #[clap(long, short)]
        context: Option<Alias<ContextId>>,
```

crates/node/src/interactive_cli/identity.rs
```diff
@@ -55,21 +56,22 @@ enum AliasSubcommands {
     Remove {
         /// Name of the alias to remove
         identity: Alias<PublicKey>,
-        /// The context that the identity is a member of
-        #[arg(long, short)]
-        context: Alias<ContextId>,
+        /// The context that the identity is a member of (omit to use default context)
```
miraclx: How is this represented in the `--help` report?

Maybe we should break this into a separate line?

Otherwise, we can also omit this, and cover it in the documentation.

crates/node/src/interactive_cli/context.rs
```diff
@@ -60,8 +60,9 @@ enum Commands {
     },
     /// Invite a user to a context
     Invite {
-        /// The context to invite the user to
-        context: Alias<ContextId>,
+        /// The context to invite the user to (omit to use default context)
+        #[clap(long = "context", short = 'c')]
+        context: Option<Alias<ContextId>>,
```
miraclx: Hm, can we try using clap's `default_value` instead?

```suggestion
        /// The context to invite the user to (omit to use default context)
        #[clap(long, short, default_value = "default")]
        context: Alias<ContextId>,
```

what do you think? (let's see the `--help` report)

miraclx:
Hi @rtb-12, thanks for tackling this.

crates/node/src/interactive_cli/context.rs
miraclx: PS: if you did this, you wouldn't need a `resolve_alias` impl that defaults to `"default"`, let's just compare both approaches. I just like the `clap` approach since it's less complicated to implement.

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +64,34 @@ impl ContextAliasCommand {
         Ok(())
     }
 }
+
+#[derive(Debug, Parser)]
+#[command(name = "use", about = "Set the default context")]
+pub struct UseCommand {
+    /// The context id to set as default
+    pub context_id: ContextId,
```
miraclx: this should be `Alias<ContextId>`

which should permit either of these:

```console
context use H1phzc8eoGu8N8XJuRi99z29aRYr9KustsdTKjPQ1pwQ
```

or

```console
context use chat-app
```

crates/meroctl/src/common.rs
```diff
@@ -401,27 +401,47 @@ where
...
+// Keeping resolve_context function as a convenience wrapper
+pub async fn resolve_context(
+    multiaddr: &Multiaddr,
+    keypair: &Keypair,
+    context: Option<Alias<ContextId>>,
+) -> EyreResult<ContextId> {
+    resolve_entity(multiaddr, keypair, context, None)
+        .await
+        .map_err(|e| eyre!("Context resolution error: {}", e))
+}
```
miraclx: this really is not necessary, if the only thing we save by having it is `None`

crates/node/src/interactive_cli/identity.rs
```diff
@@ -44,8 +44,8 @@ enum AliasSubcommands {
         name: Alias<PublicKey>,
         /// The identity to create an alias for
         identity: PublicKey,
-        /// The context that the identity is a member of (omit to use default context)
-        #[clap(long = "context", short = 'c')]
+
+        #[clap(long, short, default_value = "default")]
         context: Option<Alias<ContextId>>,
```
miraclx:
```suggestion
        #[clap(long, short, default_value = "default")]
        context: Alias<ContextId>,
```

with `default_value` we don't need `Option`

crates/node/src/interactive_cli/identity.rs
```diff
@@ -54,24 +54,21 @@ enum AliasSubcommands {
...
     )]
     Remove {
-        /// Name of the alias to remove
         identity: Alias<PublicKey>,
-        /// The context that the identity is a member of (omit to use default context)
-        #[clap(long = "context", short = 'c')]
-        context: Option<Alias<ContextId>>,
+        /// The context that the identity is a member of (must be specified, use "default" for the default context)
+        #[clap(long, short, required = true)]
+        context: Alias<ContextId>,
```
miraclx: is this intentional? it makes sense, to prevent unintentional deletion, but just checking

If it is, you don't need `required` or the rest of the comments, since that is already the behaviour

```suggestion
        /// The context that the identity is a member of
        #[clap(long, short)]
        context: Alias<ContextId>,
```

crates/meroctl/src/cli/call.rs
```diff
@@ -29,7 +30,7 @@ pub const EXAMPLES: &str = r"
 ))]
 pub struct CallCommand {
     #[arg(value_name = "CONTEXT", help = "The context to call the method on")]
-    pub context: Alias<ContextId>,
+    pub context: Option<Alias<ContextId>>,
```
miraclx: why not use `default_value` here as well if you do for the interactive cli?

crates/node/src/interactive_cli/identity.rs
miraclx: Did you compare these approaches?

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -62,17 +64,19 @@ pub enum ContextIdentityAliasSubcommand {
...
 
     #[command(about = "Resolve the alias to a context identity")]
     Get {
         #[arg(help = "Name of the alias to look up")]
         identity: Alias<PublicKey>,
 
-        #[arg(long, short, help = "The context that the identity is a member of")]
-        context: Alias<ContextId>,
+        #[arg(help = "The context that the identity is a member of (omit to use default context)")]
+        #[clap(long = "context", short = 'c')]
```
miraclx: `arg` / `clap`, I'm not particularly opinonated about this, but it's ideal if we follow convention and use the same thing for now

in a separate PR we can determine if we want this changed

miraclx (commented):
should you choose to go with the `default_value` approach, this PR should shrink significantly and should mostly only contain

1. `default_value` on all `Alias<T>` command fields
2. `context use` & `context identity use`

that's about it

crates/meroctl/src/cli/context/identity.rs
rtb-12: so for now I shall not use `arg/clap` and just keep the `default_value`?

crates/meroctl/src/cli/call.rs
rtb-12: ok will do it , just missed that part

crates/node/src/interactive_cli/identity.rs
rtb-12: Yes it was intentional for unintentional deletion of `default` alias , but if it s. behaviour like that then i ll remove the required part

crates/meroctl/src/common.rs
rtb-12: But I guess keeping it is no harm , and would be easy to refer , but if you say I ll remove it

crates/meroctl/src/cli/context/alias.rs
rtb-12: Ok will change it

crates/meroctl/src/cli/context/identity.rs
miraclx: Keep `#[arg(long, short)]`, as opposed to `#[clap(long, short)]` which matches the rest of the codebase.

Plus, you shouldn't need the explicit specification (`#[clap(long = "context", short = "c")]`) UNLESS the variable name doesn't match what you're associating with.

rtb-12: ok

crates/meroctl/src/cli/context/alias.rs
rtb-12: For this then I have to rename the alias `chat-app` to `default` also n ? because in my implementation it ls like I create alias named as default for an context id ?

miraclx: I'm not sure I follow, those two were examples of what we should support by `UseCommand`

Either an alias is provided, or a `ContextId`

rtb-12: Can we do like make an alias rename command as I see there is no rename command for it and then using that only I can then implement the above use command for an alias name ?? Or shall I implement it like get the context ID from the alias name and then create a new alias named default ?

miraclx: > Or shall I implement it like get the context ID from the alias name and then create a new alias named default ?

Yeah.. just resolve the alias, then create one

As for rename, it shouldn't be needed, we can instead do the same thing we're doing here, let you alias new names to old names, and then remove an old name. But that can come in a later PR.

rtb-12:
Hi , @miraclx I have updated the PR , added `context identity use` and also add support of using alias in `contes use alias-name` , kindly review it

crates/meroctl/src/cli/context.rs
```diff
@@ -64,6 +65,8 @@ pub enum ContextSubCommands {
     Update(UpdateCommand),
     Identity(ContextIdentityCommand),
     Alias(ContextAliasCommand),
+    #[command(name = "use")]
+    Use(UseCommand),
```
miraclx: this attribute line is not necessary since the command name is already the variant name

```suggestion
    Use(UseCommand),
```

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +64,58 @@ impl ContextAliasCommand {
         Ok(())
     }
 }
+
+#[derive(Debug, Parser)]
+#[command(name = "use", about = "Set the default context")]
```
miraclx:
```suggestion
#[derive(Debug, Parser)]
#[command(about = "Set the default context")]
```

crates/meroctl/src/cli/call.rs
miraclx: this is still unresolved

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +64,58 @@ impl ContextAliasCommand {
...
+        let config = load_config(&environment.args.home, &environment.args.node_name)?;
+        let multiaddr = fetch_multiaddr(&config)?;
+
+        // Create "default" alias for the specified context ID
+        let default_alias: Alias<ContextId> =
+            "default".parse().expect("'default' is a valid alias name");
+
+        // Resolve the input (whether it's an alias or a context ID)
+        let lookup_response =
+            lookup_alias(multiaddr, &config.identity, self.context_id.clone(), None)
```
miraclx:
```suggestion
            lookup_alias(multiaddr, &config.identity, self.context_id, None)
```

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +64,58 @@ impl ContextAliasCommand {
         Ok(())
     }
 }
+
+#[derive(Debug, Parser)]
+#[command(name = "use", about = "Set the default context")]
+pub struct UseCommand {
+    pub context_id: Alias<ContextId>,
```
miraclx:
```suggestion
    pub context: Alias<ContextId>,
```

since it can EITHER be a `ContextId` or an `Alias<ContextId>`, this should be named `context` instead

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +64,58 @@ impl ContextAliasCommand {
...
+        let lookup_response =
+            lookup_alias(multiaddr, &config.identity, self.context_id.clone(), None)
+                .await
+                .wrap_err("Failed to resolve context alias")?;
+
+        // Extract the context ID from the lookup response
+        let resolved_context_id = lookup_response
+            .data
+            .value
+            .ok_or_else(|| eyre::eyre!("Failed to resolve context: no value found"))?;
```
miraclx: we cannot GUARANTEE the provided `context` WILL be an already existing alias.

So we should be calling `resolve_alias()` instead, extracting the `ContextId` with `ResolveResponse::value`

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +64,58 @@ impl ContextAliasCommand {
...
+
+        // Set the "default" alias to point to the resolved context ID
+        let res = create_alias(
+            multiaddr,
+            &config.identity,
+            default_alias,
+            None,
+            resolved_context_id.clone(),
```
miraclx: explicit `.clone` on `Alias<T>` is not necessary, since `Alias<T>: Copy`

https://github.com/calimero-network/core/blob/3d8a2774c4816e5598571d4f22d12fd5dcc47d34/crates/primitives/src/alias.rs#L44

crates/meroctl/src/cli/context/delete.rs
miraclx, line 1: I don't expect any change to this file.

crates/meroctl/src/cli/context/get.rs
```diff
@@ -23,7 +23,7 @@ pub struct GetCommand {
     pub command: GetSubcommand,
 
     #[arg(value_name = "CONTEXT", help = "Context we're operating on")]
-    pub context: Alias<ContextId>,
+    pub context: Option<Alias<ContextId>>,
```
miraclx: `default_value` over `Option`

crates/meroctl/src/common.rs
miraclx, line 1: I don't expect any changes to this file.

crates/node/src/interactive_cli/identity.rs
```diff
@@ -220,3 +238,19 @@ fn handle_alias_command(node: &Node, command: AliasSubcommands, ind: &str) -> Ey
...
+            .ok_or_eyre("Unable to resolve context alias")
+    } else {
+        let default_alias: Alias<ContextId> =
+            "default".parse().expect("'default' is a valid alias name");
+
+        node.ctx_manager
+            .lookup_alias(default_alias, None)?
+            .ok_or_eyre("No default context set. Please set one with 'context use <context-id>' or specify a context explicitly")
+    }
+}
```
miraclx: and we should no longer need this

crates/node/src/interactive_cli/context.rs
```diff
@@ -327,3 +338,20 @@ fn handle_alias_command(node: &Node, command: AliasCommands, ind: &str) -> EyreR
...
+    } else {
+        // Otherwise, use the default alias
+        let default_alias: Alias<ContextId> =
+            "default".parse().expect("'default' is a valid alias name");
+
+        node.ctx_manager
+            .lookup_alias(default_alias, None)?
+            .ok_or_eyre("No default context set. Please set one with 'context use <context-id>' or specify a context explicitly")
+    }
+}
```
miraclx: or this, plus this is a duplicate of the other one

rtb-12:
Hi , @miraclx ,i have updated the PR with the above  suggestions and review them

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +66,52 @@ impl ContextAliasCommand {
...
+            "default".parse().expect("'default' is a valid alias name");
+
+        let resolve_response = resolve_alias(multiaddr, &config.identity, self.context, None)
+            .await
+            .wrap_err("Failed to resolve context")?;
+
+        let resolved_context_id = resolve_response
+            .value()
+            .cloned()
+            .ok_or_else(|| eyre::eyre!("Failed to resolve context: no value found"))?;
```
miraclx:
```suggestion
            .ok_or_eyre("Failed to resolve context: no value found")?;
```

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +66,52 @@ impl ContextAliasCommand {
...
+            None,
+            resolved_context_id,
+        )
+        .await
+        .wrap_err("Failed to set default context")?;
+
+        environment.output.write(&res);
+
+        let input = self.context.to_string();
+        if input == resolved_context_id.to_string() {
```
miraclx:
```suggestion
        if self.context.as_str() == resolved_context_id.as_str() {
```

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -64,3 +66,52 @@ impl ContextAliasCommand {
...
+            println!("Default context set to: {}", resolved_context_id);
+        } else {
+            println!(
+                "Default context set to: {} (from alias '{}')",
+                resolved_context_id, input
```
miraclx:
```suggestion
                resolved_context_id, self.context
```

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -25,13 +25,22 @@ pub struct ContextIdentityCommand {
 pub enum ContextIdentitySubcommand {
     #[command(about = "List identities in a context", alias = "ls")]
     List {
-        #[arg(help = "The context whose identities we're listin")]
-        context: Alias<ContextId>,
+        #[arg(help = "The context whose identities we're listing (omit to use default context)")]
+        #[clap(long = "context", short = 'c')]
+        context: Option<Alias<ContextId>>,
```
miraclx: 1. follow convention: use `arg` instead of `clap`
2. we don't need to explicitly specify `long` and `short` when the field name isn't different
3. `default_value` instead of `Option`

```suggestion
        #[arg(help = "The context whose identities we're listing")]
        #[clap(long, short, default_value = "default")]
        context: Alias<ContextId>,
```

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -25,13 +25,22 @@ pub struct ContextIdentityCommand {
...
         #[arg(long, help = "Show only owned identities")]
         owned: bool,
     },
     #[command(about = "Manage identity aliases")]
     Alias(ContextIdentityAliasCommand),
+    #[command(about = "Set default identity for a context", name = "use")]
+    Use {
```
miraclx: the variant name already is use, you don't need `name = "use"`

```suggestion
    #[command(about = "Set default identity for a context")]
    Use {
```

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -25,13 +25,22 @@ pub struct ContextIdentityCommand {
...
+        #[arg(help = "The identity to set as default")]
+        identity: PublicKey,
+        #[arg(help = "The context to set the identity for (omit to use default context)")]
+        #[arg(long, short)]
+        context: Option<Alias<ContextId>>,
```
miraclx:
```suggestion
        #[arg(help = "The context to set the identity for")]
        #[arg(long, short, default_alias = "default")]
        context: Alias<ContextId>,
```

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -50,8 +59,9 @@ pub enum ContextIdentityAliasSubcommand {
         #[arg(help = "The identity to create an alias for")]
         identity: PublicKey,
 
-        #[arg(long, short, help = "The context that the identity is a member of")]
-        context: Alias<ContextId>,
+        #[arg(help = "The context that the identity is a member of (omit to use default context)")]
+        #[arg(long, short)]
+        context: Option<Alias<ContextId>>,
```
miraclx: same

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -62,17 +72,19 @@ pub enum ContextIdentityAliasSubcommand {
         #[arg(help = "Name of the alias to remove")]
         identity: Alias<PublicKey>,
 
-        #[arg(long, short, help = "The context that the identity is a member of")]
-        context: Alias<ContextId>,
+        #[arg(help = "The context that the identity is a member of (omit to use default context)")]
+        #[arg(long, short)]
+        context: Option<Alias<ContextId>>,
```
miraclx: I thought we agreed destructive action should not be implicit?

https://github.com/calimero-network/core/pull/1171#discussion_r2027017264

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -62,17 +72,19 @@ pub enum ContextIdentityAliasSubcommand {
...
     #[command(about = "Resolve the alias to a context identity")]
     Get {
         #[arg(help = "Name of the alias to look up")]
         identity: Alias<PublicKey>,
 
-        #[arg(long, short, help = "The context that the identity is a member of")]
-        context: Alias<ContextId>,
+        #[arg(help = "The context that the identity is a member of (omit to use default context)")]
+        #[arg(long, short)]
+        context: Option<Alias<ContextId>>,
```
miraclx: same

crates/node/src/interactive_cli/context.rs
```diff
@@ -77,18 +78,20 @@ enum Commands {
...
+        /// The context to delete (omit to use default context)
+        #[clap(long = "context", short = 'c')]
+        context: Option<Alias<ContextId>>,
     },
     /// Update the proxy for a context
     UpdateProxy {
-        /// The context to update the proxy for
-        context: Alias<ContextId>,
+        #[clap(long = "context", short = 'c')]
+        context: Option<Alias<ContextId>>,
```
miraclx: please change all patterns like this

```suggestion
        #[clap(long, short, default_value = "default")]
        context: Alias<ContextId>,
```

crates/node/src/interactive_cli/context.rs
```diff
@@ -285,6 +302,27 @@ impl ContextCommand {
...
+
+                let context_id = node
+                    .ctx_manager
+                    .resolve_alias(context, None)?
+                    .ok_or_eyre("unable to resolve context")?;
+
+                node.ctx_manager
+                    .create_alias(default_alias, None, context_id)?;
+
+                if context.as_str() != context_id.to_string() {
```
miraclx:
```suggestion
                if context.as_str() != context_id.as_str() {
```

crates/node/src/interactive_cli/identity.rs
```diff
@@ -30,6 +31,14 @@ enum IdentitySubcommands {
         #[command(subcommand)]
         command: AliasSubcommands,
     },
+    /// Set default identity for a context
+    Use {
+        /// The identity to set as default
+        identity: PublicKey,
```
miraclx:
```suggestion
        identity: Alias<PublicKey>,
```

crates/node/src/interactive_cli/identity.rs
```diff
@@ -30,6 +31,14 @@ enum IdentitySubcommands {
...
+        /// The context to set the default identity for (omit to use default context)
```
miraclx:
```suggestion
        /// The context to set the default identity for
```

crates/node/src/interactive_cli/identity.rs
```diff
@@ -53,23 +61,21 @@ enum AliasSubcommands {
         aliases = ["rm", "del", "delete"],
     )]
     Remove {
-        /// Name of the alias to remove
         identity: Alias<PublicKey>,
-        /// The context that the identity is a member of
+        /// The context whose alias we're removing
         #[arg(long, short)]
         context: Alias<ContextId>,
```
miraclx: this removes the description for the only required field - `IDENTITY`, and the description of `-c, --context` is incorrect, we're not removing the alias for the context, but for AN identity in the context

please revert these

crates/node/src/interactive_cli/identity.rs
```diff
@@ -53,23 +61,21 @@ enum AliasSubcommands {
...
     },
     #[command(about = "Resolve the alias to a context identity")]
     Get {
         /// Name of the alias to look up
         identity: Alias<PublicKey>,
-        /// The context that the identity is a member of
-        #[arg(long, short)]
+        #[arg(long, short, default_value = "default")]
         context: Alias<ContextId>,
```
miraclx: we've lost description again

crates/node/src/interactive_cli/identity.rs
```diff
@@ -53,23 +61,21 @@ enum AliasSubcommands {
...
     },
     #[command(about = "List context identity aliases", alias = "ls")]
     List {
-        /// The context whose aliases we're listing
+        #[arg(long, short, default_value = "default")]
         context: Option<Alias<ContextId>>,
     },
```
miraclx: same here

crates/node/src/interactive_cli/identity.rs
```diff
@@ -88,17 +94,44 @@ impl IdentityCommand {
             IdentitySubcommands::Alias { command } => {
                 handle_alias_command(node, command, &ind.to_string())?;
             }
+            IdentitySubcommands::Use { identity, context } => {
+                let context_id = node
+        .ctx_manager
+        .resolve_alias(context, None)?
+        .ok_or_eyre("unable to resolve context")?;
```
miraclx: `cargo +nightly fmt`

crates/node/src/interactive_cli/identity.rs
```diff
@@ -195,14 +228,21 @@ fn handle_alias_command(node: &Node, command: AliasSubcommands, ind: &str) -> Ey
...
+                    "default".parse().expect("'default' is a valid alias name");
 
-            let context_id = context
-                .map(|context| node.ctx_manager.resolve_alias(context, None))
-                .transpose()?
-                .flatten();
+                node.ctx_manager
+                    .lookup_alias(default_alias, None)?
+                    .ok_or_eyre("unable to resolve default context")?
+            };
```
miraclx, line 261: previously, `context identity list` would return all the identities for all the contexts when a context is not specified

and when one is specified, it only lists the identities within that context

to maintain that behaviour, you can introduce an `--all` flag

so we have one of the following

```console
context identity list
    (lists identities in `default` context)
context identity list --all
    (lists identities in all contexts, but this has to conflict with "context" being specified)
context identity list 2XGCNdL88mGVRVnhJpxAM32J9PBiQ1pw4s3BgeFf2juw
    (lists identities in one specific context)
```

rtb-12: HI , @miraclx can this be done in an another PR ?

rtb-12:
Hi , @miraclx updated the PR with the above suggestions

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -25,21 +25,23 @@ pub struct ContextIdentityCommand {
 pub enum ContextIdentitySubcommand {
     #[command(about = "List identities in a context", alias = "ls")]
     List {
-        #[arg(help = "The context whose identities we're listing (omit to use default context)")]
-        #[clap(long = "context", short = 'c')]
-        context: Option<Alias<ContextId>>,
+        #[arg(
+            help = "The context whose identities we're listing (omit to use default context)",
```
miraclx: can you remove points of "(omit to use default context)" entirely?

it should already be in the help report, clap should already say what the default is

crates/node/src/interactive_cli/context.rs
```diff
@@ -79,19 +79,19 @@ enum Commands {
     /// Leave a context
     Leave {
         /// The context to leave (omit to use default context)
-        #[clap(long = "context", short = 'c')]
-        context: Option<Alias<ContextId>>,
+        #[clap(long, short, default_value = "default")]
+        context: Alias<ContextId>,
```
miraclx: `leave` is a destructive action, should not be implicit

 ```suggestion
     Leave {
        /// The context to leave (omit to use default context)
        #[clap(long, short)]
        context: Alias<ContextId>,
```

File: crates/meroctl/src/cli/context/identity.rs [hunk 41]
```diff
@@ -156,14 +202,21 @@ async fn list_identities(
     multiaddr: &Multiaddr,
     client: &Client,
     keypair: &Keypair,
-    context: Alias<ContextId>,
+    context: Option<Alias<ContextId>>,
```
miraclx, line 222: is there any use of `list_identities` that specifies `None`?

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -16,7 +16,7 @@ use crate::output::{InfoLine, Report};
 pub struct WatchCommand {
     /// ContextId to stream events from
     #[arg(value_name = "CONTEXT", help = "Context to stream events from")]
-    pub context: Alias<ContextId>,
+    pub context: Option<Alias<ContextId>>,
```
miraclx: Surely this can use `default_value = "default"`

crates/node/src/interactive_cli/identity.rs
```diff
@@ -80,25 +87,57 @@ impl IdentityCommand {
...
         Ok(())
     }
 }
 
-fn list_identities(node: &Node, context: Alias<ContextId>, ind: &str) -> EyreResult<()> {
-    let context_id = node
-        .ctx_manager
-        .resolve_alias(context, None)?
-        .ok_or_eyre("unable to resolve")?;
+fn list_identities(node: &Node, context: Option<Alias<ContextId>>, ind: &str) -> EyreResult<()> {
```
miraclx, line 127: same here

```suggestion
fn list_identities(node: &Node, context: Alias<ContextId>, ind: &str) -> EyreResult<()> {
```

crates/node/src/interactive_cli/identity.rs
miraclx: Sure, why not!

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -25,13 +25,24 @@ pub struct ContextIdentityCommand {
...
     List {
-        #[arg(help = "The context whose identities we're listin")]
+        #[arg(
+            help = "The context whose identities we're listing (omit to use default context)",
+            default_value = "default"
+        )]
         context: Alias<ContextId>,
         #[arg(long, help = "Show only owned identities")]
         owned: bool,
     },
```
miraclx: also, this should have `long, short` for `context`, which matches how you've defined it for the interactive CLI

crates/node/src/interactive_cli/identity.rs
miraclx: Updated https://github.com/calimero-network/core/issues/1117.

rtb-12: Will pick this issue after the merger of this PR

crates/meroctl/src/cli/context/identity.rs
rtb-12: like  in the future when we implement the `--all` tag then specifying `None` will be useful

miraclx: fair, okay..

crates/node/src/interactive_cli/identity.rs
miraclx: factoring https://github.com/calimero-network/core/pull/1171#discussion_r2039935490, we can leave this too

crates/node/src/interactive_cli/context.rs
```diff
@@ -78,8 +78,8 @@ enum Commands {
     },
     /// Leave a context
     Leave {
-        /// The context to leave (omit to use default context)
-        #[clap(long, short, default_value = "default")]
+        /// The context to leave
```
miraclx:
```suggestion
        /// The context to leave
```

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -61,7 +61,7 @@ pub enum ContextIdentityAliasSubcommand {
         #[arg(help = "The identity to create an alias for")]
         identity: PublicKey,
 
-        #[arg(help = "The context that the identity is a member of (omit to use default context)")]
+        #[arg(help = "The context that the identity is a member of ")]
```
miraclx:
```suggestion
        #[arg(help = "The context that the identity is a member of")]
```

crates/node/src/interactive_cli/context.rs
```diff
@@ -60,7 +60,7 @@ enum Commands {
     },
     /// Invite a user to a context
     Invite {
-        /// The context to invite the user to (omit to use default context)
+        /// The context to invite the user to
```
miraclx:
```suggestion
        /// The context to invite the user to
```

miraclx (commented):
nothing `cargo +nightly fmt` can't fix

crates/node/src/interactive_cli/identity.rs
```diff
@@ -125,8 +125,8 @@ impl IdentityCommand {
     }
 }
 
-fn list_identities(node: &Node, context: Option<Alias<ContextId>>, ind: &str) -> EyreResult<()> {
-    let context_id = if let Some(ctx) = context {
+fn list_identities(node: &Node, context: Alias<ContextId>, ind: &str) -> EyreResult<()> {
+    let context_id = if let ctx = context {
```
miraclx: this is an infallible pattern match match, but anyway factoring your comment for `meroctl` we can revert to what you had before, since it'll be covered by the listing effort

https://github.com/calimero-network/core/pull/1171#discussion_r2039984637

rtb-12:
Any other changes or suggestions ?

crates/node/src/interactive_cli/context.rs
```diff
@@ -78,16 +79,18 @@ enum Commands {
...
     Delete {
         /// The context to delete
+        #[clap(long, short)]
         context: Alias<ContextId>,
     },
     /// Update the proxy for a context
     UpdateProxy {
-        /// The context to update the proxy for
+        #[clap(long, short, default_value = "default")]
         context: Alias<ContextId>,
```
miraclx: removed description

crates/node/src/interactive_cli/identity.rs
```diff
@@ -43,8 +52,7 @@ enum AliasSubcommands {
         name: Alias<PublicKey>,
         /// The identity to create an alias for
         identity: PublicKey,
-        /// The context that the identity is a member of
-        #[arg(long, short)]
+        #[arg(long, short, default_value = "default")]
         context: Alias<ContextId>,
     },
```
miraclx: removed description

crates/node/src/interactive_cli/identity.rs
miraclx: this is unresolved

to clarify, infallible / irrefutable here means the `else` block will never run because you replaced `Some(ctx)` for `ctx` which is unconditional

crates/meroctl/src/cli/context/watch.rs
miraclx: this is unresolved

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -25,13 +25,26 @@ pub struct ContextIdentityCommand {
 pub enum ContextIdentitySubcommand {
     #[command(about = "List identities in a context", alias = "ls")]
     List {
-        #[arg(help = "The context whose identities we're listin")]
+        #[arg(
+            help = "The context whose identities we're listing",
+            long,
+            short,
+            default_value = "default"
+        )]
```
miraclx: feel free to break this up

```suggestion
        #[arg(help = "The context whose identities we're listing")]
        #[arg(long, short, default_value = "default")]
```

crates/node/src/interactive_cli/identity.rs
```diff
@@ -53,7 +62,6 @@ enum AliasSubcommands {
         aliases = ["rm", "del", "delete"],
     )]
     Remove {
-        /// Name of the alias to remove
```
miraclx, line 53: ^ removed comment

miraclx (commented):
lgtm, just one last thing

miraclx (commented):
@rtb-12, for the files you've touched so far, I'd consider this ready.

But I also realize commands like `invite`, `update` haven't been updated for `meroctl`.

And for `merod` - `state`, `peers`, `call`.

rtb-12:
Hi , @miraclx for `invite` and `update`  I was not sure like whether it required to have `default` context by it self , I though it should be explicit , and for the `merod` commands , I will add I overlooked that part . Apart from this any other changes ?

miraclx:
> for `invite` and `update` I was not sure like whether it required to have default context by it self

I'd say this is fine, plus you already have a default context for `invite` in merod anyway..

- `invite` should have default for both `context` and `inviter`
- `update` should for `context` and `executor`

> Apart from this any other changes ?

None that I can readily see atm, we should be good

crates/node/src/interactive_cli/call.rs
miraclx, line 1: the executor should use the default too

crates/meroctl/src/cli/context/invite.rs
```diff
@@ -15,16 +15,18 @@ use crate::output::Report;
 #[derive(Debug, Parser)]
 #[command(about = "Create invitation to a context")]
 pub struct InviteCommand {
-    #[clap(
+    #[arg(
         value_name = "CONTEXT",
-        help = "The context for which invitation is created"
+        help = "The context for which invitation is created (omit to use default context)",
```
```diff
-(omit to use default context)
```

crates/node/src/interactive_cli/identity.rs
miraclx: ping

crates/node/src/interactive_cli/context.rs
```diff
@@ -114,6 +118,9 @@ enum Commands {
         #[command(subcommand)]
         command: AliasCommands,
     },
+    Use {
+        context: Alias<ContextId>,
+    },
```
miraclx: ^ please add description doc here

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -79,6 +79,54 @@ impl ContextAliasCommand {
     }
 }
 
+#[derive(Debug, Parser)]
+#[command(about = "Set the default context")]
+pub struct UseCommand {
+    pub context: Alias<ContextId>,
```
miraclx, line 86: ^ this too, the `context`

miraclx:
Also, [this](https://github.com/calimero-network/core/pull/1171#pullrequestreview-2739867787) is still pending <sub>([ref](https://github.com/calimero-network/core/pull/1171#discussion_r2025154044))</sub>:

> Can you include a preview of any of the --help reports that shows
> context use --help & context identity use --help
> any command that includes Alias<T> with default_value = "default", just so we know what that's like

Just show these (we just need examples in the PR):

```console
> context use --help

> context update --help

> context identity ls --help

> context identity use --help
```

rtb-12:
Screenshot 2025-04-16 at 6 04 53 PM
Screenshot 2025-04-16 at 6 05 09 PM

the update command is implemented in PR , so I had made another branch for the same

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -79,6 +79,55 @@ impl ContextAliasCommand {
...
+}
+
+impl UseCommand {
+    pub async fn run(self, environment: &Environment) -> EyreResult<()> {
+        let config = load_config(&environment.args.home, &environment.args.node_name)?;
+        let multiaddr = fetch_multiaddr(&config)?;
+
+        let default_alias: Alias<ContextId> =
+            "default".parse().expect("'default' is a valid alias name");
```
fbozic: Usage of `expect` is failing the [CI-checks](https://github.com/calimero-network/core/actions/runs/14492890644/job/40653283552?pr=1171). You can use [`.wrap_err(..)?`](https://docs.rs/eyre/latest/eyre/trait.WrapErr.html) instead of `expect`.

rtb-12: Done

miraclx:
You can run to catch them locally so you don't have to wait for CI to finish

```console
cargo clippy -- -A warnings
```

miraclx (approved):

chefsale (approved):

PR #1174, opened 2025-04-03
WIP No auth node

alenmestrov:
# [product] short description

are required for this change.

## Test plan

## Documentation update

### DIFF
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -251,7 +251,7 @@ pub(crate) fn setup(
         )
         .route("/dev/peers", get(get_peers_count_handler))
         .nest("/dev/alias", alias::service());
-        // .route_layer(from_fn(dev_mode_auth));
+    // .route_layer(from_fn(dev_mode_auth));
 
     let admin_router = Router::new()
         .merge(unprotected_router)
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -141,7 +141,7 @@ pub(crate) fn setup(
         .route("/generate-jwt-token", post(generate_jwt_token_handler))
         .route("/peers", get(get_peers_count_handler))
         .nest("/alias", alias::service())
-        .layer(AuthSignatureLayer::new(store))
+        // .layer(AuthSignatureLayer::new(store))
         .layer(Extension(Arc::clone(&shared_state)));
 
     let unprotected_router = Router::new()
@@ -250,8 +250,8 @@ pub(crate) fn setup(
             get(get_proposal_handler),
         )
         .route("/dev/peers", get(get_peers_count_handler))
-        .nest("/dev/alias", alias::service())
-        .route_layer(from_fn(dev_mode_auth));
+        .nest("/dev/alias", alias::service());
+        // .route_layer(from_fn(dev_mode_auth));
 
     let admin_router = Router::new()
         .merge(unprotected_router)
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -125,12 +125,10 @@ pub async fn start(
         if let Some((path, handler)) = jsonrpc::service(&config, server_sender.clone()) {
             app = app
                 .route(path, handler.clone())
-                .route_layer(JwtLayer::new(store.clone()))
                 .nest(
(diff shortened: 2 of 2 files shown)


crates/primitives/src/common.rs
```diff
@@ -5,6 +5,11 @@ pub const fn bool_true() -> bool {
     true
 }
 
+#[must_use]
+pub const fn bool_false() -> bool {
```
miraclx: `bool: Default` is already false

crates/server/src/admin/service.rs
```diff
@@ -55,12 +55,17 @@ use crate::AdminState;
 pub struct AdminConfig {
     #[serde(default = "calimero_primitives::common::bool_true")]
     pub enabled: bool,
+    #[serde(default = "calimero_primitives::common::bool_false")]
```
```suggestion
    #[serde(default)]
```

miraclx: alternatively, if we don't want this entry to be visible in the config file, we should `#[serde(skip)]`

crates/merod/src/cli/run.rs
```diff
@@ -12,7 +12,10 @@ use crate::cli::RootArgs;
 
 /// Run a node
 #[derive(Debug, Parser)]
-pub struct RunCommand;
+pub struct RunCommand {
+    #[arg(long)]
+    pub auth: bool,
```
miraclx:
```suggestion
    #[arg(long, default)]
    pub auth: bool,
```

so it's not a required flag

crates/server/src/lib.rs
```diff
@@ -123,16 +120,12 @@ pub async fn start(
...
-                        .route("/", handler)
-                        .route_layer(from_fn(dev_mode_auth))
-                        .layer(Extension(Arc::clone(&shared_state))),
-                );
+            app = app.route(path, handler.clone()).nest(
+                "/jsonrpc/dev",
+                Router::new()
+                    .route("/", handler)
+                    .layer(Extension(Arc::clone(&shared_state))),
+            );
```
miraclx: this doesn't factor in `auth_enabled`

crates/server/src/lib.rs
```diff
@@ -123,17 +123,31 @@ pub async fn start(
...
-                .nest(
-                    "/jsonrpc/dev",
-                    Router::new()
-                        .route("/", handler)
-                        .route_layer(from_fn(dev_mode_auth))
-                        .layer(Extension(Arc::clone(&shared_state))),
-                );
+            let mut jsonrpc_router = Router::new().route("/", handler.clone());
+
+            if config
```
fbozic: Shouldn't we have the same flag in the jsonrpc config and not use admin config for the jsonrpc server setup?

alenmestrov: I have updated the JSON-RPC config to have its own flag. The initial idea was that we have it turned on/off globally, and since we are going to remove it in future I placed it in only one place. But this makes more sense currently

crates/merod/src/cli/run.rs
```diff
@@ -13,7 +13,7 @@ use crate::cli::RootArgs;
 /// Run a node
 #[derive(Debug, Parser)]
 pub struct RunCommand {
-    #[arg(long)]
+    #[arg(long, default_value_t = false)]
```
miraclx:
```suggestion
    #[arg(long, default)]
```

crates/server/src/lib.rs
```diff
@@ -120,13 +123,31 @@ pub async fn start(
...
+                .route("/", handler)
+                .layer(Extension(Arc::clone(&shared_state)));
+
+            if config
+                .jsonrpc
+                .as_ref()
+                .map_or(false, |jsonrpc| jsonrpc.auth_enabled)
+            {
+                dev_router = dev_router.route_layer(from_fn(dev_mode_auth));
+            }
```
miraclx: why not move this into `jsonrpc::service` to match the convention in `admin::setup`

miraclx: I'm not sure why any of this was here to begin with, this should pretty much look like the `admin` block below

alenmestrov: moved the routing to service

miraclx (approved):
Thank you!

PR #1195, opened 2025-04-11
feat: introduce anyhow-style `bail!` macro for app error handling

miraclx:
## Description

This introduces the following:

1. `app::bail!` is equivalent to [`anyhow::bail!`](https://docs.rs/anyhow/latest/anyhow/macro.bail.html), returning an `Err` variant of `Result<_, calimero_sdk::types::Error>`
	This can be used in either of these cases:

	```rust
	app::bail!("unauthorized");
	
	app::bail!(Error::Unauthorized);
	```

2. It's recommended to return `app::Result<_>` in your application methods (if you want to take advantage of `app::bail!`)

4. It's recommended to define your own `Error` type that wraps the `StoreError` like the kv-store example:

	https://github.com/calimero-network/core/blob/d513acff5d67850ddf581b511bb6f3074153ec75/apps/kv-store/src/lib.rs#L26-L34

5. `calimero_sdk::types::Error::msg` previously accepted ONLY a `&str`, it can now accept anything `Display`-able.

## Test plan

Skipped for now

## Documentation update

We can do this after the hackathon


chefsale (approved):
lgtm - fix ci.

xilosada (approved):

PR #1197, opened 2025-04-12
feat: expand possible inputs to `app::bail!`

miraclx:
## Description

Iterations on https://github.com/calimero-network/core/pull/1195.

The initial impl would always `Display` serializable errors, which isn't the best if you intend to parse it at the caller.

This patch uses [some](https://www.reddit.com/r/rust/comments/rnn32g/rust_already_has_specialization/) [rust-trickery](https://lukaskalbertodt.github.io/2019/12/05/generalized-autoref-based-specialization.html), different from what we [previously](https://github.com/calimero-network/core/blob/0588c5e4cc0f9db2a83d76e4ce4c4a942d813e40/crates/sdk/macros/src/logic/method.rs#L136-L148) [had](https://github.com/calimero-network/core/blob/0588c5e4cc0f9db2a83d76e4ce4c4a942d813e40/crates/sdk/src/returns.rs) for return type serialization due to needing more than 2 overrides, but it essentially exploits related language features to enable this.

Now this expands possible inputs to `app::bail!` to the following:

```rust
// literal string (as before)
app::bail("unauthorized");

// Serialize if `MyError: Serialize` otherwise, `Display` if `MyError: Error`.
app::bail(MyError::Unauthorized);
```

## Test plan

tests are included, and they pass, plus apps are updated in #1199 and they compile

## Documentation update

after the hackathon

### DIFF
--- a/crates/sdk/macros/src/lib.rs
+++ b/crates/sdk/macros/src/lib.rs
@@ -5,6 +5,7 @@
 
 use macros::parse_macro_input;
 use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
 use quote::{quote, ToTokens};
 use syn::{Expr, ItemImpl};
 
@@ -86,4 +87,11 @@ pub fn emit(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as Expr);
 
     quote!(::calimero_sdk::event::emit(#input)).into()
+}
+
+#[proc_macro]
+pub fn bail(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as TokenStream2);
+
+    quote!(::calimero_sdk::__bail__!(#input)).into()
 }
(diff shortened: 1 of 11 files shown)


crates/sdk/macros/src/lib.rs
```diff
@@ -92,17 +92,13 @@ pub fn emit(input: TokenStream) -> TokenStream {
...
+        let wrapped = ::calimero_sdk::types::__private::Wrap(#input);
 
-    quote! {
-        return Err(::calimero_sdk::types::Error::from(#input));
-    }
+        {
+            use ::calimero_sdk::types::__private::*;
+            return ::core::result::Result::Err(wrapped.into_error());
+        }
+    }}
```
miraclx: this was written this way to sanitize the environment, something [return type](https://github.com/calimero-network/core/blob/0588c5e4cc0f9db2a83d76e4ce4c4a942d813e40/crates/sdk/macros/src/logic/method.rs#L136-L150) generalization can borrow from

chefsale (approved):

Issue #684, opened 2024-09-03
Open admin-dashboard from CLI


alenmestrov:
### Open Admin Dashboard from Node CLI

The goal is to implement a command to open the admin dashboard directly from the node CLI.

#### Example Usage

Typing the following command:

```bash
admin-dashboard <PORT>
```

Would open the default OS browser with the URL:

```text
http://localhost:<PORT>/admin-dashboard
```

#### Ideal Functionality

The ideal implementation would automatically find the server ports required to open the admin dashboard, removing the need to manually specify the `<PORT>`.

#### Initial Version

The initial version of this feature, which does not include the automatic port detection, is available here:

[Initial Version on GitHub](https://github.com/calimero-network/core/tree/feat/open_admin_cli)

iamgoeldhruv:
@chefsale 
I would like to work upon this issue.Please assign it to me

chefsale:
Assigned.

iamgoeldhruv:
Hi @miraclx, I'm currently working on this issue and have a general idea of the implementation. However, while exploring the codebase, I’m a bit unsure about the correct place to add the changes.

I was looking at core/crates/node/src/interactive_cli — is this the right location to implement the solution? If not, could you kindly point me in the right direction?

Your guidance would be really appreciated. Thank you!

PR #1205, opened 2025-04-15
Added command to open admin dashboard from node cli

iamgoeldhruv:
This pr is solving issue #684

##Description
This PR introduces a feature that allows opening the admin dashboard directly from the Node CLI (system terminal/command line).


iamgoeldhruv:
@miraclx 
Sir,I have tried implementing the above-mentioned issue.Please review the PR and approach to the issue according to your convenience and  and let me know if it requires changes or improvements.
Thanks

crates/node/Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde_json.workspace = true
 tokio = { workspace = true, features = ["io-std", "macros"] }
 tracing.workspace = true
 url.workspace = true
+webbrowser = "0.8"
```
miraclx: 1. please move this version association to the root `Cargo.toml`, and only keep `webbrowser.workspace = true` here, like everything else
2. `webbrowser v0.8` is over 2 years old, did you have any issues compiling v1.0.4?

crates/node/src/interactive_cli.rs
```diff
@@ -36,6 +37,8 @@ pub enum SubCommand {
     Peers(peers::PeersCommand),
     // Store(store::StoreCommand),
     State(state::StateCommand),
+    #[command(name = "admin-dashboard")]
+    AdminDashboard(admin_dashboard::AdminDashboardCommand),
```
miraclx: 1. when the variant name is the same as the name, you don't need to override the name with `#[command(name =..)]` - (by default, clap [will use](https://arc.net/l/quote/uioyazrw) `kebab-case`)
2. Use WebUI instead. So the command will be `webui`, we can then use `rename = "lower"` here

crates/node/src/interactive_cli/admin_dashboard.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[derive(Debug, Args)]
+pub struct AdminDashboardCommand {
+    #[clap(required = true)]
+    port: u16,
+}
+
+impl AdminDashboardCommand {
+    pub fn run(&self) -> Result<()> {
+        let url = format!("http://localhost:{}/admin-dashboard", self.port);
```
miraclx: take in the argument `&Node` here, on which you can add the server config, because we [knew](https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/node/src/lib.rs#L70) it when [constructing](https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/node/src/lib.rs#L160) `Node`

crates/node/src/interactive_cli/admin_dashboard.rs
```diff
@@ -0,0 +1,18 @@
+use clap::Args;
+use eyre::Result;
+use webbrowser;
+
+#[derive(Debug, Args)]
+pub struct AdminDashboardCommand {
+    #[clap(required = true)]
+    port: u16,
```
miraclx: then we should no longer need this

crates/node/src/interactive_cli.rs
iamgoeldhruv: @miraclx ,sir u mean here that i should use command webui instead of admin-dashboard

miraclx: Yes sir

Cargo.toml
```diff
@@ -144,6 +144,7 @@ wasmer-types = "4.2.5"
 web3 = "0.19.0"
 rcgen = "0.13.1"
 x509-parser = "0.16.0"
+webbrowser = "1.0.4"
```
miraclx: ~nit: [order](https://github.com/calimero-network/core/blob/master/STYLE.md#sort-cargotoml-dependencies-alphabetically)

crates/node/src/lib.rs
```diff
@@ -130,7 +131,7 @@ pub async fn start(config: NodeConfig) -> EyreResult<()> {
 
     #[expect(trivial_casts, reason = "Necessary here")]
     let mut server = Box::pin(calimero_server::start(
-        config.server,
+        server_config,
```
miraclx: this isn't defined anywhere?

did you mean?

```suggestion
        config.server.clone(),
```

crates/node/src/lib.rs
```diff
@@ -102,6 +102,7 @@ pub struct Node {
     ctx_manager: ContextManager,
     network_client: NetworkClient,
     node_events: broadcast::Sender<NodeEvent>,
+    pub server_config: ServerConfig,
```
miraclx: shouldn't need to be `pub`

crates/node/src/interactive_cli/webui.rs
```diff
@@ -0,0 +1,31 @@
...
+            .iter()
+            .find_map(|addr| {
+                addr.iter().find_map(|proto| match proto {
+                    Protocol::Tcp(port) => Some(port),
+                    _ => None,
+                })
+            })
+            .unwrap_or(DEFAULT_PORT);
+
+        let url = format!("http://localhost:{}/admin-dashboard", port);
```
miraclx: Consider reusing this from meroctl:

https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/meroctl/src/common.rs#L26-L48

We can move it to [`crates/primitives/src/common.rs`](https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/primitives/src/common.rs)

iamgoeldhruv:
@miraclx Thanks a lot for the suggestion! 🙏  
I've made the changes you recommended by replacing the inline port extraction logic with the `multiaddr_to_url` function from `primitives::common` and well as other minor changes.

However, I'm encountering a GitHub build failure and I'm not sure what’s causing it. 😞  
It would mean a lot if you could help me debug this or point me in the right direction.

Thanks again!

crates/node/src/interactive_cli/webui.rs
```diff
@@ -11,21 +12,23 @@ pub struct WebUICommand;
...
-                })
-            })
-            .unwrap_or(DEFAULT_PORT);
+            .get(0)
+            .ok_or_else(|| eyre!("No listen address found"))?;
+        let url: Url = multiaddr_to_url(addr, "/admin-dashboard").unwrap_or_else(|_| {
+            Url::parse(&format!(
+                "http://localhost:{}/admin-dashboard",
+                DEFAULT_PORT
+            ))
```
miraclx: if we can't determine the url from `multiaddr_to_url`, falling back to `DEFAULT_PORT` might be unideal as it might be that a totally different service would be using that port (if it's even bound at all)

so in the best case, we connect to totally random website that likely returns 404, in the worst case, the browser says unable to connect

imo, if `multiaddr_to_url` fails, we should report that error

crates/primitives/Cargo.toml
```diff
@@ -10,6 +10,8 @@ license.workspace = true
 bs58.workspace = true
 borsh = { workspace = true, features = ["derive"], optional = true }
 ed25519-dalek.workspace = true
+eyre.workspace = true
+libp2p.workspace = true
```
miraclx: we don't need the whole of `libp2p` here, try just importing `multiaddr`

crates/node/src/interactive_cli/webui.rs
```diff
@@ -11,21 +12,23 @@ pub struct WebUICommand;
...
+            .unwrap()
+        });
+
+        webbrowser::open(url.as_str()).map_err(|e| eyre!("Failed to open browser: {}", e))?;
```
miraclx: instead of

`.map_err(|_| eyre!("unable to .. {e}"))`

you can use

`.wrap_err("unable to ..")`

crates/node/src/interactive_cli.rs
```diff
@@ -36,6 +37,8 @@ pub enum SubCommand {
     Peers(peers::PeersCommand),
     // Store(store::StoreCommand),
     State(state::StateCommand),
+    #[command(name = "lower")]
```
miraclx: shouldn't this be `rename = _`?

Cargo.toml
```diff
@@ -141,10 +141,12 @@ url = "2.5.2"
 velcro = "0.5.4"
 wasmer = "4.2.5"
 wasmer-types = "4.2.5"
+webbrowser = "1.0.4" 
 web3 = "0.19.0"
 rcgen = "0.13.1"
 x509-parser = "0.16.0"
 
+
```
miraclx: ^ extraneous spacing

miraclx (commented):
Hi @iamgoeldhruv, as far as your compilation issue, it's due to primitives depending on `libp2p` which depends on `getrandom`

You can see this with

```console
$ cargo tree -p gen-ext -i getrandom
```

But I suspect even after you make [this](https://github.com/calimero-network/core/pull/1205#discussion_r2047433787) change, it'll still fail to compile due to

```console
multiaddr > libp2p-identity > ed25519-dalek > rand_core > getrandom
```

So, you simply need to make `multiaddr` an optional dependency, gated by a feature flag

Pretty much like we have this here:

https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/store/Cargo.toml#L28

crates/node/src/interactive_cli.rs
iamgoeldhruv: i tried rename after your suggestion but it gave compile error,so I moved to name

miraclx: Ah interesting, in that case it should be:

```suggestion
    #[command(name = "webui")]
```

iamgoeldhruv:
@miraclx ,all above mentioned changes are added.

crates/node/src/interactive_cli/webui.rs
```diff
@@ -15,20 +13,14 @@ impl WebUICommand {
         let addr = node
             .server_config
             .listen
-            .get(0)
+            .first()
             .ok_or_else(|| eyre!("No listen address found"))?;
```
miraclx: Since `server_config.listen` is a list of multiaddrs, you can loop through them all, converting each one until we find one that is valid

something like this

```rust
let mut urls = node.server_config.listen
    .iter()
    .map(|addr| multiaddr_to_url(addr, "/admin-dashboard"))
    .peekable();

let addr = 'done: {
    while let Some(url) = urls.next() {
        match addr {
            Ok(url) => break 'done url,
            Err(err) if urls.peek().is_none() => bail!(err),
            Err(_) => continue,
        };
    }

    bail!("no listen address found");
};

webbrowser::open(url.as_str()).wrap_err("Failed to open browser")?;

// ..
```

essentially making sure we check all multiaddrs before failing

crates/primitives/Cargo.toml
```diff
@@ -24,3 +24,4 @@ hex.workspace = true
 
 [lints]
 workspace = true
+
```
miraclx: ^ extraneous spacing

iamgoeldhruv:
@miraclx
Thank you so much for pointing out the iteration part — I completely missed that! I've now implemented the logic as you suggested.
Would appreciate a review whenever you get a chance. 🙏

miraclx (approved):

chefsale (approved):

Issue #1149, opened 2025-03-14
chore(config): remove protocol duplication in context config


miraclx:
As noted in https://github.com/calimero-network/core/pull/1148#discussion_r1995928083

We have an unnecessary duplication in the config specification

For example:

```toml
[context.config.near]
signer = "relayer"
protocol = "near"
network = "testnet"
contract_id = "calimero-context-config.testnet"
```

the `protocol` field is redundant (already named in the table)

brymut:
Hi @chefsale, I'd like to take a shot at this as I familiarise myself.

PR #1206, opened 2025-04-15
chore(config): Remove redundant protocol field from context config

Nathy-bajo:
Related to https://github.com/calimero-network/core/issues/1149

## Description

I identified and removed redundant protocol field declarations from protocol configuration files where the protocol was already specified in the table name (context.config.<protocol>).


crates/meroctl/src/cli/context/identity.rs
```diff
@@ -214,7 +214,7 @@ async fn identity_exists_in_context(
     let endpoint = format!("admin-api/dev/contexts/{}/identities", context_id);
     let url = multiaddr_to_url(multiaddr, &endpoint)?;
 
-    let response: GetContextIdentitiesResponse = reqwest::Client::new()
+    let response: GetContextIdentitiesResponse = Client::new()
         .get(url)
```
miraclx: thanks for spotting this, can you use the client?

```suggestion
    let response: GetContextIdentitiesResponse = client
        .get(url)
```

miraclx (commented):
Hi @Nathy-bajo, thanks for tackling this.

So while this is exposed to e2e-tests, it's not the source of the issue

When a node is initialized, it comes with these parameters by default, for example, here's NEAR

https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/merod/src/cli/init.rs#L200-L204

See how we define the table title as

> `"near".to_owned()`

and still define

> `protocol: "near".into()`

we shouldn't need [the second one](https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/context/config/src/client/config.rs#L22)

this will also mean in the context manager, you change the following lines:

https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/context/src/lib.rs#L203

https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/context/src/lib.rs#L235

crates/meroctl/src/cli/context/identity.rs
Nathy-bajo: Done!

Nathy-bajo:
> Hi @Nathy-bajo, thanks for tackling this.
> 
> So while this is exposed to e2e-tests, it's not the source of the issue
> 
> When a node is initialized, it comes with these parameters by default, for example, here's NEAR
> 
> https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/merod/src/cli/init.rs#L200-L204
> 
> See how we define the table title as
> 
> > `"near".to_owned()`
> 
> and still define
> 
> > `protocol: "near".into()`
> 
> we shouldn't need [the second one](https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/context/config/src/client/config.rs#L22)
> 
> this will also mean in the context manager, you change the following lines:
> 
> https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/context/src/lib.rs#L203
> 
> https://github.com/calimero-network/core/blob/50dbf0736a55ed0c48328ef524d3e472707b8677/crates/context/src/lib.rs#L235

Since we are removing the protocol field from ClientConfigParams, do we still need to get the protocol from the context?

crates/context/src/lib.rs
```diff
@@ -158,6 +158,7 @@ impl ContextManager {
         initialization_params: Vec<u8>,
         result_sender: oneshot::Sender<EyreResult<(ContextId, PublicKey)>>,
     ) -> EyreResult<()> {
+        let protocol_clone = protocol.to_owned();
```
miraclx: how about you take in a `String` instead of a `&str`

crates/context/src/lib.rs
```diff
@@ -200,7 +201,7 @@ impl ContextManager {
             &context,
             identity_secret,
             Some(ContextConfigParams {
-                protocol: config.protocol.as_str().into(),
+                protocol: protocol.into(),
```
```suggestion
                protocol: protocol.as_str().into(),
```

crates/context/src/lib.rs
```diff
@@ -232,7 +233,7 @@ impl ContextManager {
 
             this.config_client
                 .mutate::<ContextConfigEnv>(
-                    config.protocol.as_str().into(),
+                    std::borrow::Cow::Borrowed(&protocol_clone),
```
```suggestion
                    protocol.as_str().into(),
```

crates/context/src/lib.rs
```diff
@@ -482,7 +483,7 @@ impl ContextManager {
                 };
 
                 Ok(Some(ContextConfigParams {
-                    protocol: config.protocol.into_string().into(),
+                    protocol: config.protocol.into_string().into(),
```
miraclx: `cargo +nightly fmt`

miraclx (commented):
> Since we are removing the protocol field from ClientConfigParams, do we still need to get the protocol from the context?

Yes, the context should still have protocol information and we'll in-turn save that to be fetched later. But as you'd probably notice, we use diferent structs to represent those.

`ClientConfigParams` which we are removing `protocol` from, is for the config-file deserialization.
`ContextConfigParams` which still has `protocol` is for internal state of the context.

also, we still need the changes to e2e tests you had in your previous commit that now seem to have [been reintroduced](https://github.com/calimero-network/core/compare/23851a8abe7505dc2c75e31fd9ad9ff7b57d1b47..be7db2214239d3a70c6391c9b282d7c5e62a1b18#diff-3da8dc41ab9d4c59f3a23d6efbb6975364a5e9cf80982782d159da53a8443054)

```rust
format!("context.config.icp.protocol=\"{}\"", "icp"),
```

and others

Nathy-bajo:
> > Since we are removing the protocol field from ClientConfigParams, do we still need to get the protocol from the context?
> 
> Yes, the context should still have protocol information and we'll in-turn save that to be fetched later. But as you'd probably notice, we use diferent structs to represent those.
> 
> `ClientConfigParams` which we are removing `protocol` from, is for the config-file deserialization. `ContextConfigParams` which still has `protocol` is for internal state of the context.

Ohh I see now. 


> 
> also, we still need the changes to e2e tests you had in your previous commit that now seem to have [been reintroduced](https://github.com/calimero-network/core/compare/23851a8abe7505dc2c75e31fd9ad9ff7b57d1b47..be7db2214239d3a70c6391c9b282d7c5e62a1b18#diff-3da8dc41ab9d4c59f3a23d6efbb6975364a5e9cf80982782d159da53a8443054)
> 
> ```rust
> format!("context.config.icp.protocol=\"{}\"", "icp"),
> ```
> 
> and others

Done

crates/context/src/lib.rs
```diff
@@ -158,6 +158,7 @@ impl ContextManager {
         initialization_params: Vec<u8>,
         result_sender: oneshot::Sender<EyreResult<(ContextId, PublicKey)>>,
     ) -> EyreResult<()> {
+        let protocol_clone = protocol.to_string();
```
alenmestrov: The current implementation works correctly, but we can optimize the string handling. The `protocol.to_owned()` is created after the `finalizer` definition but before it's used, which is technically fine but could be more efficient and clearer.

Suggested improvement:
1. Move the `protocol_clone` creation before the `finalizer` definition:
```rust
let protocol_clone = protocol.to_owned();  // Create owned copy first
let finalizer = async move {
    // ...
}
```

2. Use `protocol_clone.clone().into()` in the async block:
```rust
this.config_client.mutate::<ContextConfigEnv>(
    protocol_clone.clone().into(),  // Use owned copy
    config.network.as_str().into(),
    config.contract_id.as_str().into(),
)
```

This change:
- Makes the code more explicit about string ownership
- Potentially reduces the number of string allocations
- Makes the code flow more logical (create owned copy before using it)

fbozic: `_clone` in variable name looks a bit odd, not sure if we need this here.

alenmestrov:
Also, make sure that e2e tests pass.

crates/context/src/lib.rs
miraclx: So, this is unresolved. What I meant here is instead of the parameter `protocol` being a `&str`, now that we know we need a `String`, you can simply change that there to a `String`.

With that, you'd no longer need a `protocol_clone`.

miraclx (approved):
Aside the pending comments, lgtm! Thank you for your contribution!

Approving to unblock.

crates/context/src/lib.rs
Nathy-bajo: Done!

Nathy-bajo:
> Aside the pending comments, lgtm! Thank you for your contribution!
> 
> Approving to unblock.

Thank you for your review!

I'll be exploring more issues then.

crates/context/src/lib.rs
miraclx: Line 154

>         protocol: &str,

let's make that `protocol: String`, at which point you can simplify the rest of the changes here and calling `query` / `mutate` should be `protocol.as_str().into()`

no `clone()`s or `to_{owned,string}()`s necessary

miraclx (commented):
nearly there

crates/node/src/interactive_cli/context.rs
```diff
@@ -292,7 +292,7 @@ impl ContextCommand {
                 let (tx, rx) = oneshot::channel();
 
                 node.ctx_manager.create_context(
-                    &protocol.as_str(),
+                    (&protocol.as_str()).to_string(),
```
miraclx:
```suggestion
                    protocol.as_str().to_owned(),
```

miraclx (commented):
So, `master` is currently broken due to e2e tests not having been updated following #1171.

We'll wait for #1207 which should fix those issues to be merged.

crates/node/src/interactive_cli/context.rs
Nathy-bajo: @miraclx Alright.

Nathy-bajo:
> So, `master` is currently broken due to e2e tests not having been updated following #1171.
> 
> We'll wait for #1207 which should fix those issues to be merged.

Alright.

frdomovic (approved):

PR #1207, opened 2025-04-15
Add update command to merod interactive cli

### LINKED ISSUE #1192: `merod`'s interactive CLI should support `context update`
`merod`'s interactive CLI is sorely missing `context update`, which is already present in `meroctl`.

We should support all features except, of course `--watch`

rtb-12:
Added the support of `update` command in merod's interactive cli
Fixes #1192


crates/node/src/interactive_cli/context.rs
```diff
@@ -400,6 +422,60 @@ impl ContextCommand {
...
+                            .await?;
+
+                        node.ctx_manager
+                            .update_application_id(context_id, application_id, public_key)
+                            .await?;
+
+                        println!(
+                            "{ind} Installed and updated application for context '{}'",
+                            pretty_alias(Some(context), &context_id)
+                        );
```
miraclx: after the application is installed, let's log a line that says we did just that

rtb-12:
Screenshot 2025-04-16 at 6 10 08 PM

miraclx (approved):
lgtm, thanks!

chefsale (approved):

rtb-12:
> just to note.. when #1171 merges, we'd have to update this PR to include `default_value` for the `context`, and `identity`.

Hi , updated PR with this change too

miraclx:
Amazing, thanks, looks good.

Looks like CI is failing on master due to the last introductions made to #1171 not having `#[arg(short, long)]`:

https://github.com/calimero-network/core/blob/8d9d5081927871b3647f354acd3a0e506974ddf1/crates/meroctl/src/cli/context/invite.rs#L19-L24

Can you double-check `meroctl context invite`? - this also applies to interactive CLI's `call` command.

Also, I just realized `meroctl call` doesn't have `default_value` for the executor.

rtb-12:
Screenshot 2025-04-17 at 3 48 30 PM
Any thing else to be done ?

crates/meroctl/src/cli/context/invite.rs
```diff
@@ -16,7 +16,9 @@ use crate::output::Report;
 #[derive(Debug, Parser)]
 #[command(about = "Create invitation to a context")]
 pub struct InviteCommand {
-    #[arg(
+    #[clap(
+        long = "context",
+        short = 'c',
```
```suggestion
        long,
        short,
```

PS: feel free to break it up into it's own `#[arg(..)]` annotation

miraclx (commented):
we're missing this

> this also applies to interactive CLI's call command.

miraclx:
Yeah, this is due to e2e tests not being updated in #1171.

@rtb-12, sorry for the back and forth, the e2e tests didn't run at all in #1171, so we didn't catch these there.

But now that meroctl arguments have changed, `e2e-tests` need to appropriately use the right format.

You can go through the list of e2e methods [here](https://github.com/calimero-network/core/blob/c2c35b6bf8657f7aad994233a89857982a41cddd/e2e-tests/src/meroctl.rs) and update each one to the right format.

I believe most of them should only require adding `"--context"`, before `context_id`

And it may only need to affect `json_rpc_execute`, `get_proposals` and `context_invite`

frdomovic (approved):

miraclx:
Hi @rtb-12, I want to confirm whether you are working on the patches or if we should fix them.

rtb-12:
Hi @miraclx actually I am on short trip this long weekend So was not able to do it by myself, I will work from Monday.
Apologies for the delay, I am fine if you want to fix it by yourself else I will do it by Monday

miraclx:
Ah okay, thanks for letting us know. Do enjoy your trip.

Issue #1074, opened 2025-01-29
CLI trunk versioning


miraclx:
When built from source, do not use the crate version for the CLI, instead use the git revision.

We can also include versions of build dependencies

Here's an example:

```console
$ near --version
near (release trunk) (build 1.1.0-4033-g6972c509c-modified) (rustc 1.72.0) (protocol 63) (db 38)
```

chefsale:
Assigned.

raveesh146:
@chefsale @miraclx  please review the PR

chefsale:
A bit more work then expected, bumping the reward to 250 USDC.

antonpaisov:
@raveesh146 I see you closed the first attempt PR, are you still planning to work on this, or should I remove you as an assignee?

antonpaisov:
hey @raveesh146 , I'm un-assigning you from the issue, if you're still working on it - pls ping me here or better on discord.

dotandev:
@antonpaisov 

I will like to take this up if it is up.

Thank you.

antonpaisov:
hey @dotandev !
sure thing, assigned, go for it!

Issue #1193, opened 2025-04-11
Unify CLIs


miraclx:
Arguments, usage syntax and documentation sometimes vary wildly between interactive CLI and `meroctl`.

We need to figure out a way to reduce the duplicated effort of maintaining both and keeping them in sync, and potentially evaluate if generics can help.

The key distinction between the two are;

1. `merod`'s interactive CLI talks to the node via function calls
2. `meroctl`'s interactive CLI talks to the node via server APIs
3. `meroctl` can support `watch` commands/args, while `merod` cannot - or perhaps it could but it'll have to look different from it's current implementation in `meroctl` so it doesn't block other operations from happening.

This will likely be a significant endeavour, requiring the champion to take initiative and be tasteful with their implementation.

onyedikachi-david:
Hello @miraclx, is this a paid bounty issue.

chefsale:
Hi, @onyedikachi-david. Yes, all the bounties are paid. This is a bigger piece of work and would be great maybe to start it off with a smaller bounty first to kick it off. Can you join our discord server and just say hi there that you are interested in the bounties and I will DM you.

onyedikachi-david:
Thanks @chefsale, Just joined the discord server, what channel? dev_general? I am abi.encode.onyedikachi on discord.

chefsale:
Yes, dev general is fine.

PR #1212, opened 2025-04-16
feat(cli): update version format to match near --version

raveesh146:
## Description
Updates the CLI version format to match the `near --version` format as specified in issue #1074. Changes include:
- Remove duplicate binary name from version output
- Format version string to match near --version format
- Include git revision for source builds
- Show rustc version as build dependency
- Add protocol version based on major version number

## Test plan
1. Build both CLIs:
   ```bash
   cargo build
   ```
2. Run version commands:
   ```bash
   ./target/debug/merod --version
   ./target/debug/meroctl --version
   ```
3. Verify output matches required format:
   ```
   merod (release trunk) (build 0.5.0-50dbf07-modified) (rustc 1.81.0) (protocol 0)
   meroctl (release trunk) (build 0.5.0-50dbf07-modified) (rustc 1.81.0) (protocol 0)
   ```

## Documentation update
No documentation updates required as this change only affects the CLI version output format.


raveesh146:
@chefsale please review the PR

crates/meroctl/src/cli/version.rs
```diff
@@ -0,0 +1,41 @@
...
+
+        format!(
+            "(release trunk) (build {}-{}{}) (rustc {}) (protocol {})",
+            env!("CARGO_PKG_VERSION"),
+            git_revision.trim(),
+            if is_modified { "-modified" } else { "" },
+            rustc_version.trim(),
+            env!("CARGO_PKG_VERSION").split('.').next().unwrap_or("0")
+        )
+    })
```
miraclx: 1. These shouldn't run at runtime, but be pre-computed at compile-time. See `neard` for reference:

	https://github.com/near/nearcore/blob/db3f73c1405562d4b4dd5a4980954dfb22bf78d8/neard/src/main.rs#L13-L31

2. Unlike NEAR, we do have a version attached, so we can keep it for now, we can revisit if we decide to pin all source versions to `0.0.0` at some later time.

	So I expect it to look like this:
	
	```console
	# on the latest tag:
	
	$ merod --version
	merod (release 0.5.0) (build 0.5.0) (commit e68ddb3e) (rustc 1.81.0)
	```
	
	```console
	# on a commit that is tagged (and dirty):
	
	$ merod --version
	merod (release 0.5.0) (build 0.5.0-modified) (commit e68ddb3e) (rustc 1.81.0)
	```
	
	```console
	# on a commit that isn't tagged (and isn't dirty):
	
	$ merod --version
	merod (release 0.5.0) (build 0.5.0-29-g55e4ef3e) (commit 55e4ef3e) (rustc 1.81.0)
	```
	
	```console
	# on a commit that isn't tagged (and is dirty):
	
	$ merod --version
	merod (release 0.5.0) (build 0.5.0-29-g55e4ef3e-modified) (commit 55e4ef3e) (rustc 1.81.0)
	```

crates/meroctl/src/cli/version.rs
```diff
@@ -0,0 +1,41 @@
...
+            .map(|s| s.split_whitespace().nth(1).unwrap_or("unknown").to_string())
+            .unwrap_or_else(|| "unknown".to_string());
+
+        format!(
+            "(release trunk) (build {}-{}{}) (rustc {}) (protocol {})",
+            env!("CARGO_PKG_VERSION"),
+            git_revision.trim(),
+            if is_modified { "-modified" } else { "" },
+            rustc_version.trim(),
+            env!("CARGO_PKG_VERSION").split('.').next().unwrap_or("0")
```
miraclx:
```console
$ git describe --always --dirty=-modified --tags --match '[0-9]*'
```

see https://github.com/near/nearcore/blob/db3f73c1405562d4b4dd5a4980954dfb22bf78d8/neard/build.rs#L52 for a decent reference

crates/meroctl/src/cli/version.rs
```diff
@@ -0,0 +1,41 @@
+use std::process::Command;
+use std::sync::OnceLock;
+
+static VERSION_INFO: OnceLock<String> = OnceLock::new();
+
+pub fn get_version_info() -> &'static str {
```
chefsale, line 1: The code seems duplicated, reuse the method across the two binaries. Maybe put it into utils folder.

miraclx: For now, there's no "ideal" place to keep deduplicated utils without introducing a new crate which I'd consider out of scope for this PR. But we can revisit this with https://github.com/calimero-network/core/issues/1193.

miraclx: On second thought, let's do just that. Since the final form of this PR will need `build.rs` scripts, it's best to keep it unified. Good catch!

@raveesh146, here's what to do.

Introduce a `crates/version` folder, you can do so by copying the `Cargo.toml` template from `crates/primitives` for example.

Define the package name as `calimero-version`, with it's version matching exactly what we have on `merod` and `meroctl`.

At this point, you can introduce the new crate to the root manifest [here](https://github.com/calimero-network/core/blob/c2c35b6bf8657f7aad994233a89857982a41cddd/Cargo.toml#L31) and [here](https://github.com/calimero-network/core/blob/c2c35b6bf8657f7aad994233a89857982a41cddd/Cargo.toml#L170).

This crate should have a build script which prepares environment variables based on what was described above.

Then the crate's lib should define a `CalimeroVersion` struct that contains all of these values.

This struct should then have a `Display` implementation which combines all the values into the pretty string as described above.

For good measure, feel free to introduce a `LazyLock<String>` since this only needs to be computed once per runtime.

Please reach out if you hit any snags.

speedxgul: yes, I'll get back soon after considering these inputs

raveesh146:
@miraclx please take a look at the PR

Cargo.toml
```diff
@@ -144,6 +145,7 @@ wasmer-types = "4.2.5"
 web3 = "0.19.0"
 rcgen = "0.13.1"
 x509-parser = "0.16.0"
+calimero-version = { path = "./crates/version" }
```
miraclx, line 148: move to line 172, in the section for calimero crates

crates/meroctl/src/cli/version.rs
miraclx, line 1: you no longer need this file

crates/merod/Cargo.toml
```diff
@@ -42,5 +42,7 @@ calimero-network.workspace = true
 calimero-server = { workspace = true, features = ["jsonrpc", "websocket", "admin"] }
 calimero-store.workspace = true
 
+calimero-version.workspace = true
```
miraclx, line 45: there shouldn't need to be any separatiion between these

crates/meroctl/build.rs
miraclx, line 1: you no longer need this file

crates/merod/build.rs
miraclx, line 1: you no longer need this file

crates/merod/src/cli/version.rs
miraclx, line 1: you no longer need this file

crates/version/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "calimero-version"
+version = "0.5.0"
+edition = "2021"
+
+[dependencies]
+once_cell = "1.19.0"
```
miraclx, line 7: [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) has been stable since rust 1.80, we don't need this dependency

crates/version/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "calimero-version"
+version = "0.5.0"
+edition = "2021"
+
```
miraclx, line 5: as mentioned, follow the template in other crates, there are some missing fields here

crates/version/build.rs
```diff
@@ -0,0 +1,40 @@
...
+        .unwrap_or_else(|| "unknown".to_string());
+
+    // Get rustc version
+    let rustc_version = Command::new("rustc")
+        .arg("--version")
+        .output()
+        .ok()
+        .and_then(|output| String::from_utf8(output.stdout).ok())
+        .map(|s| s.split_whitespace().nth(1).unwrap_or("unknown").to_string())
+        .unwrap_or_else(|| "unknown".to_string());
```
miraclx, line 34: the binary name isn't guaranteed to be `rustc`, use the [`rustc_version`](https://docs.rs/rustc_version/latest/rustc_version/) crate instead

crates/version/build.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    // Pass individual components to the binary
+    println!("cargo:rustc-env=GIT_DESCRIBE={}", git_describe.trim());
+    println!("cargo:rustc-env=GIT_COMMIT={}", git_commit.trim());
+    println!("cargo:rustc-env=RUSTC_VERSION={}", rustc_version.trim());
```
miraclx, line 39: `CALIMERO_BUILD`, `CALIMERO_COMMIT`, `CALIMERO_RUSTC_VERSION`

crates/version/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+        )
+    }
+}
+
+static VERSION: Lazy<CalimeroVersion> = Lazy::new(CalimeroVersion::new);
+
+pub fn get_version() -> &'static str {
+    // Convert the Display output to a static str
+    Box::leak(VERSION.to_string().into_boxed_str())
+}
```
miraclx, line 47: I don't see any need for this. Let's do this instead:

```rust
static VERSION: LazyLock<CalimeroVersion> = LazyLock::new(CalimeroVersion::new);
static VERSION_STR: LazyLock<String> = LazyLock::new(|| VERSION.to_string());

pub fn get_version() -> &'static CalimeroVersion {
    &*VERSION
}

pub fn get_version_str() -> &'static str {
    &*VERSION_STR
}
```

and.. move your statics to the top of the file (after imports)

crates/version/src/lib.rs
```diff
@@ -0,0 +1,47 @@
+use std::fmt;
+
+use once_cell::sync::Lazy;
+
+pub struct CalimeroVersion {
+    release_version: &'static str,
+    build_version: &'static str,
+    commit: &'static str,
+    rustc_version: &'static str,
```
miraclx, line 9: these should be `pub`

crates/version/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+impl CalimeroVersion {
+    fn new() -> Self {
+        let pkg_version = env!("CARGO_PKG_VERSION");
+        let git_describe = env!("GIT_DESCRIBE");
+
+        let build_version = if git_describe == "unknown" {
+            pkg_version.to_string()
+        } else {
+            git_describe.to_string()
+        };
```
miraclx, line 21: this is not necessary

crates/version/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        Self {
+            release_version: pkg_version,
+            build_version: Box::leak(build_version.into_boxed_str()),
```
miraclx, line 25: and therefore, this isn't either

crates/version/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub struct CalimeroVersion {
+    release_version: &'static str,
+    build_version: &'static str,
+    commit: &'static str,
+    rustc_version: &'static str,
+}
+
+impl CalimeroVersion {
+    fn new() -> Self {
```
miraclx, line 13: with those changes, this can be `const`

crates/meroctl/build.rs
miraclx, line 1: Also, we want to rebuild this crate when the following envs change (`cargo:rerun-if-env-changed=__`):

- `CARGO_PKG_VERSION`
- `PATH` incase the `git` / `rustc` binaries are overridden.

We also want to employ better error handling, use `eyre` - feel free to follow [this](https://github.com/near/nearcore/blob/43d048a649cca85336123bbec9f0c7d8e613729a/neard/build.rs#L119-L126).

The only exception is if there was some issue running `git`, in which case we're okay with `unknown`. We should still [raise a warning](https://github.com/near/nearcore/blob/43d048a649cca85336123bbec9f0c7d8e613729a/neard/build.rs#L60-L64) incase that was wrong. (plus it helps in our CI build, where we can treat warnings as hard errors)

miraclx:
@raveesh146, any updates on this?

#1226 is introducing some functionality I'd prefer to see hosted in the crate you're introducing to deduplicate it between the binaries.

PR #1214, opened 2025-04-17
feat: created Dockerfile for meroctl, optimized image generation for merod

alenmestrov:
## Description

This PR:

- Updates the `merod` Dockerfile to optimize the build process and streamline the resulting image for the merod binary.
- Implements new Dockerfile for meroctl, same as for `merod`

## Test plan

Build Docker image using the new Dockerfile.

For merod:
`docker build -t merod:latest -f merod.Dockerfile .`

For meroctl:
`docker build -t meroctl:latest -f meroctl.Dockerfile .`

## Documentation update

TBD


meroctl.Dockerfile
```diff
@@ -0,0 +1,42 @@
+# Use an official Rust image as the base
+FROM rust:latest as builder
```
fbozic: Let's pin Rust version to the same version as here: https://github.com/calimero-network/core/blob/master/rust-toolchain.toml

meroctl.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY e2e-tests ./e2e-tests
+
+# Build the meroctl binary
+RUN cargo build --release -p meroctl
+
+# Copy the binary to a location in PATH
+RUN cp /app/target/release/meroctl /usr/local/bin/meroctl
+
+# Default command (can be overridden in docker-compose)
+CMD ["--help"]
```
fbozic: ~nit: new line is missing

merod.Dockerfile
```diff
@@ -39,8 +40,11 @@ RUN pnpm install && pnpm run build
...
-ENTRYPOINT ["/app/target/release/merod"]
+# Copy the binary to a location in PATH
+RUN cp /app/target/release/merod /usr/local/bin/merod
+
+# Set the working directory
+WORKDIR /data
 
 # Default command (can be overridden in docker-compose)
-CMD ["--help"]
+CMD ["--help"]
```
fbozic: ~nit: new line missing

merod.Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN  apt-get update && apt-get install -y \
     curl \
     build-essential \
     pkg-config \
-    jq
+    jq && \
+    rm -rf /var/lib/apt/lists/*
 
 # Install Node.js (version 20) and pnpm
 RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
```
fbozic: I am not sure if we need nodejs stuff at all, but if we are keeping it, let's move it to a dedicated stage (call it `builder-nodejs`) and let's use nodejs as a base image in that stage. In the final stage (call it `runner`) just copy files from `builder-nodejs` stage.

meroctl.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY contracts ./contracts
+COPY apps ./apps
+COPY node-ui ./node-ui
+COPY e2e-tests ./e2e-tests
+
+# Build the meroctl binary
+RUN cargo build --release -p meroctl
+
+# Copy the binary to a location in PATH
+RUN cp /app/target/release/meroctl /usr/local/bin/meroctl
```
miraclx: Seconding https://github.com/calimero-network/core/pull/1214#discussion_r2048876990, even the rust build should be it's own stage.

So the final stage will be really small and contain only the binary and runtime dependencies (no build deps).

Which means you can check for the base image for the rust build image and likely use that (with very few dependency introductions).

miraclx (commented):
Would be nice to explore reducing duplication between these dockerfiles as they're mostly the same except for the last stage

We could likely do with 1 Dockerfile with 2 final targets.

Dockerfile
```diff
@@ -0,0 +1,138 @@
+# syntax=docker/dockerfile:1
+
+# Comments are provided throughout this file to help you get started.
+# If you need more help, visit the Dockerfile reference guide at
+# https://docs.docker.com/go/dockerfile-reference/
+
+# Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
```
miraclx, line 7: whose form is this? 😅

Dockerfile
```diff
@@ -0,0 +1,138 @@
...
+    --mount=type=cache,target=/usr/local/cargo/registry/ \
+    cargo build --locked --release -p merod && \
+    cp /app/target/release/merod /usr/local/bin/
+
+# Build meroctl with caching and copy to a persistent location in the image
+RUN --mount=type=cache,target=/app/target/ \
+    --mount=type=cache,target=/usr/local/cargo/git/db \
+    --mount=type=cache,target=/usr/local/cargo/registry/ \
+    cargo build --locked --release -p meroctl && \
+    cp /app/target/release/meroctl /usr/local/bin/
```
miraclx: Let's build them together

```console
cargo build --locked --release -p merod -p meroctl
```

then you can copy them together

```console
cp /app/target/release/{merod,meroctl} /usr/local/bin/
```

miraclx (commented):
Should we add a doc that shows how to get the 2 images built from the 1 dockerfile?

But of course, updating the PR description is fine for now until we can put these in the docs.

Dockerfile
fbozic: that is from `docker init` command, I would leave it as it is.

Dockerfile
fbozic: or alternative would be to interpolate `merod` or `meroctl` as cargo target from input arguments, in that case, you would have a single rust build statement and build (which should be a bit faster than building together)

merod.Dockerfile
alenmestrov: This is needed because build folder of node-ui isn't in the git, so if the user wants to build the merod image from fresh git clone it needs to be builded.

Dockerfile
miraclx: Ah oke

miraclx (approved):
Lgtm

frdomovic (approved):

PR #1217, opened 2025-04-18
feat: Introduced workflow for publishing merod and meroctl docker images

alenmestrov:
# [product] short description

## Description

Implemented workflow file for automatic release of `merod` and `meroctl` Docker images to GitHub Container Registry.

Key features:
- Follows the same rules for triggering as the binary release workflow
- Automatically builds when:
  - New releases are created
  - Changes to Cargo.toml, Cargo.lock, or crates/** on master branch
  - Pull requests with the release/* prefix
  - Manual trigger via GitHub Actions UI
- Creates multi-architecture images (amd64 and arm64)
- Images are tagged with proper version numbers from Cargo.toml
- Images include proper metadata and license information
- Container images are minimal, secure, and run as non-root users

## Test plan

- Manually trigger the workflow from GitHub Actions UI
- Check that correct Docker images were built and published to ghcr.io
- Pull images from ghcr.io/calimero-network/merod:latest
- Pull images from ghcr.io/calimero-network/meroctl:latest
- Run basic commands with both images

## Documentation update

N/A

### DIFF
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -1,0 +1,164 @@
+name: Build and Publish Docker Images
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - Cargo.toml
+      - Cargo.lock
+      - 'crates/**'
+      - 'core/Dockerfile'
+      - '.github/workflows/publish-docker-images.yml'
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - master
+    paths:
+      - Cargo.toml
+      - Cargo.lock
+      - 'crates/**'
+      - 'core/Dockerfile'
+      - '.github/workflows/publish-docker-images.yml'
+  workflow_run:
+    workflows: ["Release"]
+    types:
+      - completed
+
+jobs:
+  prepare:
+    name: Prepare Docker Build
+    runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'workflow_dispatch' || 
+      github.event_name == 'push' || 
+      github.event_name == 'pull_request' || 
+      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')
+    outputs:
+      version: ${{ steps.version_info.outputs.version }}
+      should_build: ${{ steps.version_info.outputs.should_build }}
+      prerelease: ${{ steps.version_info.outputs.prerelease }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Get version info
+        id: version_info
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Default values
+          echo "should_build=false" >> $GITHUB_OUTPUT
+          echo "prerelease=false" >> $GITHUB_OUTPUT
+          
+          if [ "${{ github.event_name }}" == "workflow_run" ]; then
+            # Extract version from the triggering workflow (Release)
+            latest_release=$(gh api repos/${{ github.repository }}/releases/latest --jq .tag_name)
+            if [ -n "$latest_release" ]; then
+              echo "Using latest release version: $latest_release"
+              echo "version=$latest_release" >> $GITHUB_OUTPUT
+              echo "should_build=true" >> $GITHUB_OUTPUT
+            else
+              echo "No release version found"
+              exit 0
+            fi
+          else
+            # Similar logic to the release workflow
+            if [ -f "Cargo.toml" ]; then
+              version_candidate=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "merod") | .version')
+              echo "Version from Cargo: $version_candidate"
+              
+              if [ "${{ github.ref }}" == "refs/heads/master" ]; then
+                echo "On master branch"
+                echo "version=$version_candidate" >> $GITHUB_OUTPUT
+                echo "should_build=true" >> $GITHUB_OUTPUT
+              elif [ "${{ github.event_name }}" == "pull_request" ] && [[ "${{ github.head_ref }}" == release/* ]]; then
+                echo "On release PR"
+                echo "version=prerelease-${{ github.event.number }}" >> $GITHUB_OUTPUT
+                echo "should_build=true" >> $GITHUB_OUTPUT
+                echo "prerelease=true" >> $GITHUB_OUTPUT
+              elif [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+                echo "Manual trigger"
+                echo "version=$version_candidate" >> $GITHUB_OUTPUT
+                echo "should_build=true" >> $GITHUB_OUTPUT
+              else
+                echo "Not building for this branch/PR"
+              fi
+            else
+              echo "No Cargo.toml found"
+            fi
+          fi
+
+  build-and-push:
+    name: Build and Push Docker Images
+    needs: prepare
+    if: needs.prepare.outputs.should_build == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Extract metadata for merod
+        id: meta-merod
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/merod
+          tags: |
+            type=raw,value=${{ needs.prepare.outputs.version }}
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
+
+      - name: Extract metadata for meroctl
+        id: meta-meroctl
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/meroctl
+          tags: |
+            type=raw,value=${{ needs.prepare.outputs.version }}
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
+      
+      - name: Build and push merod image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./core/Dockerfile
+          push: true
+          platforms: linux/amd64,linux/arm64
+          target: merod
+          tags: ${{ steps.meta-merod.outputs.tags }}
+          labels: ${{ steps.meta-merod.outputs.labels }}
+          annotations: |
+            org.opencontainers.image.description=Merod container image
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+      
+      - name: Build and push meroctl image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./core/Dockerfile
+          push: true
+          platforms: linux/amd64,linux/arm64
+          target: meroctl
+          tags: ${{ steps.meta-meroctl.outputs.tags }}
+          labels: ${{ steps.meta-meroctl.outputs.labels }}
+          annotations: |
+            org.opencontainers.image.description=Meroctl container image
+            org.opencontainers.image.source=https://github.com/${{ github.repository }} 
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,6 +64,10 @@ RUN --mount=type=cache,target=/app/target/ \
 # Create a minimal runner stage for merod
 FROM debian:bookworm-slim AS merod
 
+# Add labels for container metadata
+LABEL org.opencontainers.image.description="Merod daemon"
+LABEL org.opencontainers.image.licenses="MIT"
+
 # Install only the essential runtime dependencies
 RUN apt-get update && apt-get install -y \
     libssl3 \
@@ -98,6 +102,10 @@ CMD ["--help"]
 ################################################################################
 # Create a minimal runner stage for meroctl
 FROM debian:bookworm-slim AS meroctl
+
+# Add labels for container metadata
+LABEL org.opencontainers.image.description="Meroctl - Control tool for Merod daemon"
+LABEL org.opencontainers.image.licenses="MIT"
 
 # Install only the essential runtime dependencies
 RUN apt-get update && apt-get install -y \


.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,213 @@
...
+    paths:
+      - Cargo.toml
+      - Cargo.lock
+      - 'crates/**'
+      - 'core/Dockerfile'
+      - '.github/workflows/publish-docker-images.yml'
+  workflow_run:
+    workflows: ["Release"]
+    types:
+      - completed
```
fbozic: Container build and publish should be part of Release workflow, not after it.

.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,213 @@
...
+          build-args: |
+            BUILDKIT_INLINE_CACHE=1
+            CARGO_NET_GIT_FETCH_WITH_CLI=true
+          outputs: type=registry,name=ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}
+          provenance: mode=max
+          sbom: true
+          annotations: |
+            org.opencontainers.image.description=${{ matrix.image == 'merod' && 'Merod container image' || 'Meroctl container image' }}
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+            org.opencontainers.image.licenses=MIT
```
fbozic: ~nit: missing new line

.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,504 @@
+name: Build, Release and Publish
```
fbozic: Keep the name as it was, `Release`.

fbozic: Do not rename a file (at least in this PR), it makes reviewing much harder since GitHub can not show the diff properly.

.github/workflows/release.yml
```diff
@@ -5,11 +5,13 @@ on:
   push:
     branches:
       - master
+      - 'feat__*'
```
fbozic: leftover from testing? I don't think we need it, you can cut a prerelease if your branch starts with `release/`

.github/workflows/release.yml
```diff
@@ -18,20 +20,45 @@ on:
...
 
 jobs:
   prepare:
     name: Prepare
     runs-on: ubuntu-latest
+    if: |
+      github.event_name == 'workflow_dispatch' || 
+      github.event_name == 'push' || 
+      github.event_name == 'pull_request' || 
+      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')
```
fbozic: Why do we need this?

.github/workflows/release.yml
```diff
@@ -18,20 +20,45 @@ on:
       - Cargo.toml
       - Cargo.lock
       - 'crates/**'
-      - .github/workflows/release.yml
+      - 'Dockerfile'
+      - '.github/workflows/release.yml'
+  workflow_run:
+    workflows: ["Release"]
+    types:
+      - completed
```
fbozic: Why do we need this?

.github/workflows/release.yml
```diff
@@ -315,4 +501,4 @@ jobs:
           git commit -m "chore: bump formulas version"
           git push origin "${target_branch}"
 
-          gh pr create -f || true
+          gh pr create -f || true
```
fbozic: ~nit: new line

.github/workflows/release.yml
```diff
@@ -47,60 +74,137 @@ jobs:
...
+            echo "overwrite_release=true" >> $GITHUB_OUTPUT
+            
+          elif [[ "${{ github.ref }}" == refs/heads/feat__* ]]; then
+            branch_name=$(echo "${{ github.ref }}" | sed 's|refs/heads/||')
+            version="dev-${branch_name}"
+            
+            echo "build_binaries=false" >> $GITHUB_OUTPUT
+            echo "release_required=false" >> $GITHUB_OUTPUT
+            echo "build_docker=true" >> $GITHUB_OUTPUT
+            echo "prerelease=true" >> $GITHUB_OUTPUT
```
fbozic: Why do we need this?

.github/workflows/release.yml
```diff
@@ -112,8 +216,8 @@ jobs:
             target: x86_64-apple-darwin
           - os: macos-latest
             target: aarch64-apple-darwin
+      fail-fast: false
```
fbozic: Why not fail-fast? We shouldn't publish anything in case of failure.

.github/workflows/release.yml
```diff
@@ -47,60 +74,137 @@ jobs:
...
+            echo "overwrite_release=true" >> $GITHUB_OUTPUT
           else
-            echo "This is not a master branch or a release PR"
-            echo "build_required=false" >> $GITHUB_OUTPUT
+            version="dev-$(echo ${{ github.ref }} | sed 's|refs/heads/||')"
+            
+            echo "build_binaries=false" >> $GITHUB_OUTPUT
             echo "release_required=false" >> $GITHUB_OUTPUT
+            echo "build_docker=false" >> $GITHUB_OUTPUT
           fi
```
fbozic: Why do we need this? I don't see `workflow_dispatch` as a trigger. Also, I don't see any benefit from having `workflow_dispatch` as a trigger.

alenmestrov:
@fbozic, the feat_ and workflow dispatch were leftovers from testing, since we had a discussion about the approach we are going to take with this, I wanted to make sure that the implemented logic is aligned with everything before removing the test parts. 

I have now removed the test parts from the workflow file.

Dockerfile
miraclx, line 1: I don't think the `-prebuilt` stages need to be in the same file, you can have them in `.github/workflows/deps/prebuilt.Dockerfile` maybe?

Dockerfile
```diff
@@ -58,12 +58,17 @@ RUN --mount=type=cache,target=/app/target/ \
     --mount=type=cache,target=/usr/local/cargo/git/db \
     --mount=type=cache,target=/usr/local/cargo/registry/ \
     cargo build --locked --release -p merod -p meroctl && \
-    cp /app/target/release/{merod,meroctl} /usr/local/bin/
+    cp /app/target/release/merod /usr/local/bin/merod && \
+    cp /app/target/release/meroctl /usr/local/bin/meroctl
```
miraclx, line 62: this is fine, but did the old one not work? or is this for clarity?

Dockerfile
```diff
@@ -129,3 +139,101 @@ USER appuser
 # Set the entrypoint
 ENTRYPOINT ["meroctl"]
 CMD ["--help"]
+
+################################################################################
+# Create a minimal runner stage for merod using prebuilt binaries
+FROM ubuntu:24.04 AS merod-prebuilt
```
miraclx: also, looks to we can deduplicate these when we move them to their own dockerfile, since we have `${BINARY_NAME}`

As for the labels, you already have them set in the CI step, so you don't need them here.

.github/workflows/release.yml
```diff
@@ -1,15 +1,15 @@
...
-  workflow_dispatch:
   push:
     branches:
       - master
     paths:
       - Cargo.toml
       - Cargo.lock
       - 'crates/**'
-      - .github/workflows/release.yml
+      - 'Dockerfile'
```
miraclx: do we care about running this workflow when the docker file changes?

shouldn't it only run when there are changes to rust?

.github/workflows/release.yml
```diff
@@ -47,60 +63,104 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          echo "Validating binary versions"
-
+          echo "target_commit=${{ github.sha }}" >> $GITHUB_OUTPUT
+
```
miraclx, line 58:
```suggestion

```

.github/workflows/release.yml
```diff
@@ -47,60 +63,104 @@ jobs:
...
-          echo "Valid version candidate: $version_candidate"
-
-          echo "target_commit=${{ github.sha }}" >> $GITHUB_OUTPUT
-
+          
           if [ "${{ github.ref }}" == "refs/heads/master" ]; then
             version="$version_candidate"
-            echo "Master version: $version"
-
+
```
miraclx: remove all `/\s+$/` to get rid of these extraneous whitespace indentations

.github/workflows/release.yml
```diff
@@ -47,60 +63,104 @@ jobs:
...
+      - name: Extract metadata for meroctl
+        id: meroctl_metadata
+        uses: docker/metadata-action@v5
+        if: steps.version_info.outputs.build_docker == 'true'
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/meroctl
+          tags: |
+            type=raw,value=${{ steps.version_info.outputs.version }}
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
```
miraclx: I don't understand this, we're extracting tags and labels from the last image and reusing them for the new one?

Why would we want this? And isn't it the [chicken or the egg](https://en.wikipedia.org/wiki/Chicken_or_the_egg) problem.

.github/workflows/release.yml
```diff
@@ -47,60 +63,104 @@ jobs:
...
+        if: steps.version_info.outputs.build_docker == 'true'
+        run: |
+          # Define cache configuration for Buildx
+          if [ "${{ github.ref }}" == "refs/heads/master" ]; then
+            # Use a master-specific cache for stable releases
+            SCOPE="${{ github.repository_owner }}-docker-master"
+          else
+            # Use a feature branch cache for development
+            SCOPE="${{ github.repository_owner }}-docker-dev"
+          fi
```
miraclx: `cache-from` [takes a list](https://arc.net/l/quote/sajbuyfb), so for dev, it can be both master and dev

.github/workflows/release.yml
```diff
@@ -47,60 +63,104 @@ jobs:
...
+          
+          echo "Using cache scope: $SCOPE"
+          echo "cache_from=type=registry,ref=ghcr.io/${{ github.repository_owner }}/buildcache:$SCOPE type=gha,scope=$SCOPE" >> $GITHUB_OUTPUT
+          echo "cache_to=type=registry,ref=ghcr.io/${{ github.repository_owner }}/buildcache:$SCOPE,mode=max,compression=zstd" >> $GITHUB_OUTPUT
```
miraclx: also, why isn't this in the docker build job? it doesn't use any dependencies from the prepare step I can find

.github/workflows/release.yml
```diff
@@ -141,13 +200,13 @@ jobs:
       - name: Setup rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          key: rust-binaries-${{ runner.os }}-${{ matrix.target }}
+          key: rust-binaries-${{ runner.os }}-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
```
miraclx: This is already [factored in](https://arc.net/l/quote/rtkjnsmu) by the action

> ```yml
> # An additional cache key that is added alongside the automatic `job`-based
> # cache key and can be used to further differentiate jobs.
> # default: empty
> key: ""
> ```

imo, this should just be `${{ matrix.target }}`

as for the `rust-binaries` prefix, there's a `prefix-key` option

.github/workflows/release.yml
```diff
@@ -252,13 +314,91 @@ jobs:
...
+            BINARY_NAME=${{ matrix.image }}
+          tags: ${{ matrix.image == 'merod' && needs.prepare.outputs.merod_tags || needs.prepare.outputs.meroctl_tags }}
+          labels: ${{ matrix.image == 'merod' && needs.prepare.outputs.merod_labels || needs.prepare.outputs.meroctl_labels }}
+          cache-from: ${{ needs.prepare.outputs.docker_cache_from }}
+          cache-to: type=gha,scope=${{ github.repository_owner }}-${{ matrix.image }},mode=max
+          outputs: type=registry,name=ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}
+          provenance: mode=max
+          sbom: true
+          annotations: |
+            org.opencontainers.image.description=${{ matrix.image == 'merod' && 'Merod container image' || 'Meroctl container image' }}
```
miraclx: can we host this with the matrix?

```yml
matrix:
  image:
    - name: merod
      description: Merod daemon
    - name: meroctl
      description: Meroctl - Control tool for Merod daemon
```

then use `${{ matrix.image.name }}` wherever you have `${{ matrix.image }}`

.github/workflows/release.yml
```diff
@@ -252,13 +314,91 @@ jobs:
...
+          file: ./Dockerfile
+          push: true
+          platforms: linux/amd64,linux/arm64
+          target: ${{ matrix.image }}-prebuilt
+          build-args: |
+            BINARY_NAME=${{ matrix.image }}
+          tags: ${{ matrix.image == 'merod' && needs.prepare.outputs.merod_tags || needs.prepare.outputs.meroctl_tags }}
+          labels: ${{ matrix.image == 'merod' && needs.prepare.outputs.merod_labels || needs.prepare.outputs.meroctl_labels }}
+          cache-from: ${{ needs.prepare.outputs.docker_cache_from }}
+          cache-to: type=gha,scope=${{ github.repository_owner }}-${{ matrix.image }},mode=max
```
miraclx: this doesn't take into account what you have above, where the `SCOPE` is suffixed with `-dev` to avoid corrupting the cache for master

you already prepared `${{ needs.prepare.outputs.docker_cache_to }}`

miraclx (commented):
would've been nice to have explored the option of building the `arm64` docker image on `aarch64`, and `amd64` on `x86_64`, which should've solved the runtime issue as well without the overhead of virtualization

.github/workflows/release.yml
alenmestrov: The logic for the docker file can change, so I placed it here, otherwise, we would first need to update Dockerfile as a separate PR and then trigger this workflow for release.

Dockerfile
alenmestrov: the old one didn't work at the end, I can't remember the exact reason, if needed I can recreate it

miraclx: Nah, it's no problem, just checking so I understood what led to the change since I expected these to be the same thing

.github/workflows/release.yml
```diff
@@ -366,25 +317,51 @@ jobs:
...
+          tags: |
+            type=raw,value=${{ needs.prepare.outputs.version }}
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/master' }}
+      
+      - name: Build and push ${{ matrix.image.name }} image
         uses: docker/build-push-action@v5
         with:
           context: .
-          file: ./Dockerfile
+          file: ./prebuilt.Dockerfile
```
miraclx: this is only relevant for CI, so let's host it somewhere other than the root

.github/workflows/release.yml
miraclx: Ah, I misunderstood this, and that kids, is why you read docs before commenting.. haha

prebuilt.Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# Create a minimal runner stage for meroctl using prebuilt binaries
+FROM prebuilt-base AS meroctl-prebuilt
+
+# Set the working directory
+WORKDIR /app
+RUN chown appuser:appuser /app
+
+# Set the entrypoint
+ENTRYPOINT ["meroctl"]
+CMD ["--help"]
```
miraclx: this is fine, but we already have `$BINARY_NAME`, so we don't even need the stage distinction at the end if we can use `$BINARY_NAME`

can we use `$BINARY_NAME`?

I expect it to work, but idk

.github/workflows/release.yml
miraclx: yeah, but this is the `push` event, idk if this is relevant for the push event

cc: @fbozic

.github/workflows/deps/prebuilt.Dockerfile
```diff
@@ -37,25 +37,28 @@ RUN chmod +x /usr/local/bin/${BINARY_NAME}
...
-WORKDIR /app
-RUN chown appuser:appuser /app
-
-# Set the entrypoint
-ENTRYPOINT ["meroctl"]
-CMD ["--help"]
\ No newline at end of file
+# Create aliased targets for backward compatibility
+FROM generic-prebuilt AS merod-prebuilt
+FROM generic-prebuilt AS meroctl-prebuilt
```
miraclx: what backwards compatibility?

.github/workflows/deps/prebuilt.Dockerfile
```diff
@@ -37,25 +37,28 @@ RUN chmod +x /usr/local/bin/${BINARY_NAME}
...
+      mkdir -p /data && \
+      chown appuser:appuser /data && \
+      echo "/data" > /tmp/workdir; \
+    else \
+      mkdir -p /app && \
+      chown appuser:appuser /app && \
+      echo "/app" > /tmp/workdir; \
+    fi
+WORKDIR /placeholder
+RUN WORKDIR=$(cat /tmp/workdir) && rm /tmp/workdir && cd $WORKDIR
```
miraclx: I don't understand this, why not just pick one folder?

.github/workflows/deps/prebuilt.Dockerfile
miraclx, line 1: I don't think we need multi stage builds in this file since there is only one flow

.github/workflows/deps/prebuilt.Dockerfile
alenmestrov: updated

miraclx (approved):
lgtm, danke

frdomovic (approved):

Issue #1216, opened 2025-04-17
CI: `e2e-tests` doesn't exit when the node runs into an issue


miraclx:
Due to an issue with the server port binding when running e2e tests in CI ([ref](https://github.com/calimero-network/core/pull/1159)), `merod` may sometimes exit, and `e2e-tests`, as currently architected, won't detect that failure.

This leads to `e2e-tests` running forever "waiting" for `merod` to start and be reachable, unaware that it already died.

We need to devise a way to check if the child has exited while waiting for the child to come online.

Here is the relevant bit:

https://github.com/calimero-network/core/blob/c2c35b6bf8657f7aad994233a89857982a41cddd/e2e-tests/src/driver.rs#L304-L313

Everytime we detect an error, L312 runs, which waits a second.

We can first call [`Child::try_wait`](https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.try_wait) to check if the child process has exited.

And only when we know it hasn't, we proceed to sleep and then retry the connection.

PR #1218, opened 2025-04-18
fix(e2e-tests): detect and handle merod process failures during startup

Nathy-bajo:
## Description

Attempts to resolve https://github.com/calimero-network/core/issues/1216

- Added try_wait() method to Merod to check child process status.

- Modified connection wait loop to verify process health before retrying.

### DIFF
--- a/e2e-tests/src/driver.rs
+++ b/e2e-tests/src/driver.rs
@@ -303,16 +303,13 @@ impl Driver {
 
                 merod.run().await?;
 
-                let node_name = merod.name.clone();
-                e.insert(merod);
-
-                let merod_ref = &merods[&node_name];
+                let node_name = e.insert(merod);
 
                 while let Err(_) = try_join!(
                     TcpSocket::new_v4()?.connect(swarm_addr),
                     TcpSocket::new_v4()?.connect(server_addr)
                 ) {
-                    if let Some(exit_code) = merod_ref.try_wait().await? {
+                    if let Some(exit_code) = node_name.try_wait().await? {
                         bail!(
                             "merod process exited with code {} before becoming ready",
                             exit_code
--- a/e2e-tests/src/driver.rs
+++ b/e2e-tests/src/driver.rs
@@ -306,7 +306,7 @@ impl Driver {
                 let node_name = merod.name.clone();
                 e.insert(merod);
 
-                let merod_ref = merods.get(&node_name).unwrap();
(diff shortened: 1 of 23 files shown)


e2e-tests/src/driver.rs
```diff
@@ -303,12 +303,21 @@ impl Driver {
 
                 merod.run().await?;
 
-                let _ = e.insert(merod);
+                let node_name = merod.name.clone();
+                e.insert(merod);
+
+                let merod_ref = merods.get(&node_name).unwrap();
```
miraclx: `e.insert` already returns the reference to the inserted item, we don't need a separate get step

e2e-tests/src/merod.rs
```diff
@@ -138,4 +138,12 @@ impl Merod {
 
         Ok(child)
     }
+
+    pub async fn try_wait(&self) -> EyreResult<Option<i32>> {
+        if let Some(child) = self.process.borrow_mut().as_mut() {
+            Ok(child.try_wait()?.map(|status| status.code().unwrap_or(-1)))
```
miraclx, line 144: 👍🏽 it's subtle, but I see what you did here

e2e-tests/src/driver.rs
```diff
@@ -303,12 +303,21 @@ impl Driver {
 
                 merod.run().await?;
 
-                let _ = e.insert(merod);
+                let node_name = merod.name.clone();
+                e.insert(merod);
+
+                let merod_ref = &merods[&node_name];
 
                 while let Err(_) = try_join!(
```
miraclx:
```suggestion
                let merod = e.insert(merod);

                while let Err(_) = try_join!(
```

e2e-tests/src/driver.rs
```diff
@@ -303,16 +303,13 @@ impl Driver {
 
                 merod.run().await?;
 
-                let node_name = merod.name.clone();
-                e.insert(merod);
-
-                let merod_ref = &merods[&node_name];
+                let node_name = e.insert(merod);
```
miraclx: this variable name isn't descriptive of what is actually stored in the variable which is a `&mut Merod`

`merod` is more appropriate.

Nathy-bajo: Done.

miraclx (approved):

chefsale (approved):

Issue #1135, opened 2025-03-04
Clean up credentials parsing from config


miraclx:
Consider this from the config file:

```toml
[context.config.signer.relayer]
url = "http://3.125.79.112:63529/"

[context.config.near]
signer = "relayer"
protocol = "near"
network = "testnet"
contract_id = "calimero-context-config.testnet"

[context.config.signer.self.near.mainnet]
rpc_url = "https://rpc.mainnet.near.org/"
account_id = "60c257e5ab83bc594181d6b57b60af51c25e0e6774d99e01d3c5f58a4a66964c"
public_key = "ed25519:7Wi1zxHUqK5fz34UXHzVGb4nmbXiCPmGh4KdPx9ucHjH"
secret_key = "ed25519:<snip>"

[context.config.signer.self.near.testnet]
rpc_url = "https://rpc.testnet.near.org/"
account_id = "d2c930d3393621716f60222a1474c17ec586148edafeaafce2738c328390eb46"
public_key = "ed25519:FBpYwJayAt9e5JULa8EFSoyHoioGEnebq2E9EdsL9oMP"
secret_key = "ed25519:<snip>"
```

The following behaviour is expected:

1. For any protocol where the `account_id` is represented exactly the same as the `public_key`, we skip including the `public_key`.
2. When parsing the config, the [current logic](https://github.com/calimero-network/core/blob/2ee05c27a17fe80305a4479cee869af45c3cc07b/crates/context/config/src/client.rs#L60-L63) selects the `context.config.<protocol>` entry only when at least one `context.config.signer.self.<protocol>.<network>` entry exists, this is wrong! It means if `context.config.<protocol>.signer` == `"local"` and there are no credentials, it will still fallback to the relayer, which is unexpected behaviour. Instead it should make sure local credentials are available if and only if `context.config.<protocol>.signer` == `"local"`.
3. Remove the `Credentials` enum that attempts to parse each protocol's credentials, falling back to the next. As this requires we encode strict parsing to ensure a N+1 protocol's credentials don't mistakenly parse as N. Which may not always be possible as all the constraints may just be the same on two separate protocols (or they may share the same types. For example `ed25519_dalek::VerifyingKey`.

iamgoeldhruv:
@chefsale ,Can I work upon this issue.It looks familiar to me

chefsale:
Assigned @iamgoeldhruv.

iamgoeldhruv:
cool,thanks

[...]

PR #1219, opened 2025-04-18
Clean up credentials parsing from config

iamgoeldhruv:
This PR solves issue #1135

## Description

This PR introduces the following changes:

Skip public_key: For any protocol where the account_id is exactly the same as the public_key (e.g., NEAR, StarkNet), we no longer include the public_key in the wallet creation payload. This avoids redundancy and aligns with protocol expectations.

Fix signer resolution logic: Corrects the logic used when parsing protocol configuration. Previously, it selected context.config.<protocol> only if at least one context.config.signer.self.<protocol>.<network> entry existed. This allowed unexpected fallback to relayer mode, even when signer == "local" but no credentials were configured. Now, the logic ensures that:

If signer == "local", valid local credentials must be present.

If not, fallback to relayer does not occur unexpectedly.

Remove Credentials enum fallback: Eliminates the fallback-based parsing approach where the Credentials enum attempted to parse credentials for multiple protocols in sequence. This approach was error-prone:

Parsing ambiguity between protocols with similar credential constraints (e.g., same key types).

Risk of mistakenly parsing N+1 protocol credentials as a valid format for N.

Strict parsing per protocol is now enforced explicitly.

### DIFF
--- a/crates/context/config/src/client.rs
+++ b/crates/context/config/src/client.rs
@@ -6,6 +6,7 @@ use std::str::FromStr;
 use alloy::signers::local::PrivateKeySigner;
 use eyre::Context;
 use thiserror::Error;
+use std::collections::BTreeMap;
 
 pub mod config;
 pub mod env;
@@ -15,7 +16,7 @@ pub mod relayer;
 pub mod transport;
 pub mod utils;
 
-use config::{ClientConfig, ClientSelectedSigner, Credentials};
+use config::{ClientConfig, ClientSelectedSigner};
 use env::Method;
 use macros::transport;
 use protocol::{ethereum, icp, near, starknet, stellar, Protocol};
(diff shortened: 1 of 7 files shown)


crates/context/config/src/client/config.rs
```diff
@@ -61,7 +62,15 @@ pub struct ClientRelayerSigner {
...
-    pub credentials: Credentials,
+    pub near_credentials: Option<NearCredentials>,
+    #[serde(flatten)]
+    pub starknet_credentials: Option<StarknetCredentials>,
+    #[serde(flatten)]
+    pub ethereum_credentials: Option<EthereumCredentials>,
+    #[serde(flatten)]
+    pub icp_credentials: Option<IcpCredentials>,
+    #[serde(flatten)]
+    pub stellar_credentials: Option<StellarCredentials>,
```
miraclx: So what I expect here instead of this, is a protocol-agnostic approach.

`credentials` should essentially just be a `serde_json::Value`.

Then in `client.rs`, for every protocol simply call [`serde_json::from_value()`](https://docs.rs/serde_json/latest/serde_json/fn.from_value.html)

iamgoeldhruv:
@miraclx 
Sir, I’ve implemented the proposed changes for the issue discussed above.
Kindly review the implementation and let me know if there are any modifications or improvements needed.

Thanks!

crates/context/config/src/client.rs
```diff
@@ -63,199 +64,237 @@ impl Client<AnyTransport> {
...
-
         let transport = transport!(local.transport, relayer);
-
         Self::new(transport)
     }
 
     pub fn from_local_config(config: &ClientConfig) -> eyre::Result<Client<LocalTransports>> {
         let mut near_transport = None;
+        let mut starknet_transport = None;
+        let mut ethereum_transport = None;
```
chefsale, line 74: Use a map here instead of copy/paste transports and use the network type as key?

crates/context/config/src/client.rs
```diff
@@ -63,199 +64,237 @@ impl Client<AnyTransport> {
...
-                            account_id: credentials.account_id.clone(),
-                            access_key: credentials.secret_key.clone(),
-                        },
-                    );
-                }
-
-                near_transport = Some(near::NearTransport::new(&config));
+            let near_config = config.signer.local.protocols.get("near").ok_or_else(|| {
+                eyre::eyre!("Local signer selected for Near but no configuration found")
+            })?;
```
miraclx, line 90: try `ok_or_eyre`

```suggestion
            let near_config = config.signer.local.protocols.get("near").ok_or_eyre(
                "the protocol `near` is configured for local signing, but no credentials were found"
            )?;
```

crates/context/config/src/client.rs
```diff
@@ -63,199 +64,237 @@ impl Client<AnyTransport> {
...
+
+            let mut network_config = near::NearConfig {
+                networks: BTreeMap::new(),
+            };
+
+            for (network, signer) in &near_config.signers {
+                // Direct credential access
+                let credentials = signer.near_credentials.as_ref().ok_or_else(|| {
+                    eyre::eyre!("Near credentials missing for network {}", network)
+                })?;
```
miraclx, line 100: this should be `serde_json::from_value` where you deserialize to `near::Credentials`

crates/context/config/src/client.rs
```diff
@@ -63,199 +64,237 @@ impl Client<AnyTransport> {
...
+
+                // Handle public key redundancy
+                let public_key = credentials.public_key.to_string();
+
+                let _ignored = network_config.networks.insert(
+                    network.clone().into(),
+                    near::NetworkConfig {
+                        rpc_url: signer.rpc_url.clone(),
+                        account_id: credentials.account_id.clone(),
+                        public_key: Some(public_key),
```
miraclx, line 110: the `public_key` field is not needed for NEAR

crates/context/config/src/client/protocol/ethereum.rs
miraclx, line 1: not needed

crates/context/config/src/client/protocol/near.rs
```diff
@@ -112,6 +112,7 @@ pub struct NetworkConfig {
     pub rpc_url: Url,
     pub account_id: AccountId,
     pub access_key: SecretKey,
+    pub public_key: Option<String>,
```
miraclx, line 115: not needed

crates/context/config/src/client/protocol/starknet.rs
```diff
@@ -95,6 +95,7 @@ pub struct NetworkConfig {
     pub rpc_url: Url,
     pub account_id: Felt,
     pub access_key: Felt,
+    pub public_key: Option<String>,
```
miraclx, line 98: not needed

crates/context/config/src/client/config.rs
```diff
@@ -61,6 +62,16 @@ pub struct ClientLocalSigner {
...
+    #[serde(flatten)]
+    pub near_credentials: Option<NearCredentials>,
+    #[serde(flatten)]
+    pub starknet_credentials: Option<StarknetCredentials>,
+    #[serde(flatten)]
+    pub ethereum_credentials: Option<EthereumCredentials>,
+    #[serde(flatten)]
+    pub icp_credentials: Option<IcpCredentials>,
+    #[serde(flatten)]
+    pub stellar_credentials: Option<StellarCredentials>,
```
miraclx, line 74:
```suggestion
    #[serde(flatten)]
    pub credentials: serde_json::Value,
```

you can equivalently get rid of `Credentials` and `RawCredentials` below this

miraclx (commented):
when I said

> For any protocol where the account_id is represented exactly the same as the public_key, we skip including the public_key.

I meant the `<protocol>::Credentials` type can either have `public_key` or not, depending on if the public key is different from the account ID.

miraclx:
Hi @iamgoeldhruv, are you still working on this?

iamgoeldhruv:
hi @miraclx ,apologies for delay,actually I was busy with some college stuffs,will make the final changes in a week

antonpaisov:
hey @iamgoeldhruv 
pls let us know if you're still working on this, otherwise I'll have to un-assign the issue

antonpaisov:
closing as inactive

PR #1223, opened 2025-04-20
feat(node): Add alias substitution support for context calls

### LINKED ISSUE #1122: Consider alias replacement in execution args

Nathy-bajo:
## Description
Attempts to resolve #1122

This PR introduces alias substitution in context call payloads, allowing users to reference identities by their aliases (e.g., {jane}) instead of raw public keys.

## Test Plan
N/A

## Documentation Update

Yes


crates/meroctl/src/cli/call.rs
```diff
@@ -51,6 +51,12 @@ pub struct CallCommand {
 
     #[arg(long, default_value = "dontcare", help = "Id of the JsonRpc call")]
     pub id: Option<String>,
+    #[arg(
+        long = "substitute",
+        help = "Aliases to substitute in the payload (format: {alias})",
```
miraclx: this will be a bit confusing to read, is this the format for the substitution or the alias in the payload.. it should be easier to figure out

crates/meroctl/src/cli/call.rs
```diff
@@ -51,6 +51,12 @@ pub struct CallCommand {
...
+        value_name = "ALIAS"
+    )]
+    pub substitute: Vec<Alias<PublicKey>>,
```
miraclx: we can take advantage of clap supporting comma-separated lists to avoid

```
--substitute jake --substitute emily
```

crates/node/src/lib.rs
```diff
@@ -540,6 +550,31 @@ impl Node {
         Ok(outcome)
     }
 
+    async fn substitute_aliases_in_payload(
+        &self,
+        context_id: ContextId,
+        payload: Vec<u8>,
+        aliases: &[Alias<PublicKey>],
+    ) -> Result<Vec<u8>, CallError> {
+        let payload_str = String::from_utf8(payload).map_err(|_| CallError::InternalError)?;
```
miraclx: it's not guaranteed the whole of the payload will be a utf-8 encoded `String`, so trivial string replacement won't work here.

But if the user requested substitution, we trust they know what they're doing and can guarantee their binary data maintains its integrity even after the substitution

crates/server/src/jsonrpc.rs
```diff
@@ -186,6 +186,7 @@ pub(crate) async fn call(
             args,
             executor_public_key,
             outcome_sender,
+            vec![],
```
miraclx: this shouldn't default to being empty, otherwise the substitutions never work over the API (which `meroctl` uses)

crates/node/src/lib.rs
```diff
@@ -540,6 +550,31 @@ impl Node {
...
+
+        let mut modified_payload = payload_str;
+        for alias in aliases {
+            let placeholder = format!("{{{}}}", alias.as_ref());
+            if modified_payload.contains(&placeholder) {
+                let public_key = self
+                    .ctx_manager
+                    .resolve_alias(*alias, Some(context_id))
+                    .map_err(|_| CallError::InternalError)?
```
miraclx, line 576: feel free to introduce a new `CallError`

crates/node/src/interactive_cli/call.rs
```diff
@@ -74,7 +84,7 @@ impl CallCommand {
                                         .join("\n")
                                 )
                             } else {
-                                format!("(raw): {:?}", result.cyan())
+                                format!("(raw): {:?}", String::from_utf8_lossy(&result).cyan())
```
miraclx: please revert, it's also not guaranteed the result of an execution will be a `String`

crates/node/src/interactive_cli/call.rs
```diff
@@ -22,6 +22,8 @@ pub struct CallCommand {
     /// The identity of the executor
     #[clap(long = "as", default_value = "default")]
     executor: Alias<PublicKey>,
+    #[clap(long = "substitute")]
+    substitutes: Vec<Alias<PublicKey>>,
```
miraclx: after settling on the design from `meroctl`, you can apply them here too.. with doc-comments and everything

crates/node/src/lib.rs
miraclx: This is why I suggested those deps in my comment - https://github.com/calimero-network/core/issues/1122#issuecomment-2710975001

chefsale (approved):
Please fix miras comment for utf-8 and any outstanding comments.

crates/node/src/lib.rs
```diff
@@ -556,23 +557,35 @@ impl Node {
...
+
                 let public_key = self
                     .ctx_manager
                     .resolve_alias(*alias, Some(context_id))
-                    .map_err(|_| CallError::InternalError)?
+                    .map_err(|_| CallError::AliasResolutionFailed)?
                     .ok_or(CallError::InternalError)?;
 
-                modified_payload = modified_payload.replace(&placeholder, &public_key.to_string());
+                result.extend_from_slice(public_key.to_string().as_bytes());
```
miraclx: you can save an allocation by using [`PublicKey::as_str`](https://github.com/calimero-network/core/blob/24f0cc6bfea9476314bf57784a503cee05d1d0a5/crates/primitives/src/identity.rs#L123-L125)

```suggestion
                result.extend_from_slice(public_key.as_str().as_bytes());
```

crates/server-primitives/src/jsonrpc.rs
```diff
@@ -144,6 +145,7 @@ pub struct ExecuteRequest {
     pub method: String,
     pub args_json: Value,
     pub executor_public_key: PublicKey,
+    pub substitute: Vec<Alias<PublicKey>>,
```
miraclx, line 149: annotate with `#[serde(default)]` so it's not expected to be present

crates/node/Cargo.toml
```diff
@@ -15,6 +15,7 @@ futures-util = { workspace = true, features = ["io"] }
 libp2p.workspace = true
 owo-colors.workspace = true
 rand.workspace = true
+memchr.workspace = true
```
miraclx: we prefer to keep these alphabetically ordered

crates/node-primitives/src/lib.rs
```diff
@@ -58,4 +58,6 @@ pub enum CallError {
     ApplicationNotInstalled { application_id: ApplicationId },
     #[error("internal error")]
     InternalError,
+    #[error("Alias error")]
+    AliasResolutionFailed,
```
miraclx: it would be useful to mention which alias we had an issue with

```suggestion
    #[error("error resolving identity alias '{alias}'")]
    AliasResolutionFailed {
        alias: Alias<PublicKey>,
    },
```

crates/node/src/interactive_cli/call.rs
miraclx: we're still missing these

crates/node/src/lib.rs
```diff
@@ -476,6 +480,13 @@ impl Node {
             });
         }
 
+        let payload = if !aliases.is_empty() {
+            self.substitute_aliases_in_payload(context_id, payload, &aliases)
+                .await?
+        } else {
+            payload
+        };
```
miraclx: no need to redeclare, since the type didn't change

```suggestion
        if !aliases.is_empty() {
            payload = self.substitute_aliases_in_payload(context_id, payload, &aliases)
                .await?
        }
```

crates/node/src/lib.rs
```diff
@@ -540,6 +551,43 @@ impl Node {
...
+        if aliases.is_empty() {
+            return Ok(payload);
+        }
+
+        let mut result = Vec::with_capacity(payload.len());
+        let mut remaining = &payload[..];
+
+        for alias in aliases {
+            let placeholder = format!("{{{}}}", alias.as_ref());
+            let placeholder_bytes = placeholder.as_bytes();
```
miraclx: no need for the explicit `.as_ref()`, since `Alias<T>: Display`

```suggestion
            let needle = format!("{{{alias}}}").as_bytes();
```

crates/node/src/lib.rs
```diff
@@ -540,6 +551,43 @@ impl Node {
...
+
+            while let Some(pos) = memmem::find(remaining, placeholder_bytes) {
+                result.extend_from_slice(&remaining[..pos]);
+
+                let public_key = self
+                    .ctx_manager
+                    .resolve_alias(*alias, Some(context_id))
+                    .map_err(|_| CallError::AliasResolutionFailed)?
+                    .ok_or(CallError::InternalError)?;
```
miraclx: it's `Err` when there was a database error, `None` when the alias doesn't exist

```suggestion
                    .resolve_alias(*alias, Some(context_id))
                    .map_err(|_| CallError::InternalError)?
                    .ok_or_else(|| CallError::AliasResolutionFailed { alias })?;
```

miraclx (commented):
looking good, nearly there

crates/node/src/lib.rs
Nathy-bajo: I still have to implement it as 
```rust
let needle_str = format!("{{{alias}}}");
let needle = needle_str.as_bytes();
 ```

Comment (miraclx) [reply crates/node/src/lib.rs]: Lifetime issue I'm guessing, feel free to use `into_bytes`

Comment (Nathy-bajo) [reply crates/node/src/lib.rs]: Done

File: crates/node/src/lib.rs [hunk 15]
```diff
@@ -540,6 +550,45 @@ impl Node {
...
+            while let Some(pos) = memmem::find(remaining, &needle) {
+                result.extend_from_slice(&remaining[..pos]);
+
+                let public_key = self
+                    .ctx_manager
+                    .resolve_alias(alias.clone(), Some(context_id))
+                    .map_err(|_| CallError::InternalError)?
+                    .ok_or_else(|| CallError::AliasResolutionFailed {
+                        alias: alias.clone(),
+                    })?;
```
miraclx: you should never need to explicitly clone an alias, deref it to `Copy`

```suggestion
                    .ok_or_else(|| CallError::AliasResolutionFailed {
                        alias: *alias,
                    })?;
```

crates/meroctl/src/cli/call.rs
```diff
@@ -108,6 +115,7 @@ impl CallCommand {
             self.method,
             self.args.unwrap_or(json!({})),
             executor,
+            self.substitute.iter().cloned().collect(),
```
miraclx:
```suggestion
            self.substitute,
```

why are we cloning?

crates/node/src/lib.rs
Nathy-bajo: Alright!

miraclx:
and finally, the description: https://github.com/calimero-network/core/pull/1223#issuecomment-2817106757

miraclx (approved):

Issue #1072, opened 2025-01-29
Meroctl event listener


miraclx:
Add a `meroctl` command that listens for events happening in a context, and calls a shell command when it's detected.

We can add a count to only trigger N events.

But this can be used for, example, to trigger a system notification or perform some other action.

PR #1224, opened 2025-04-20
feat(meroctl): Add event-triggered command execution to context watch

Nathy-bajo:
## Description

Resolves https://github.com/calimero-network/core/issues/1072

I implemented a new meroctl command that listens for context events and triggers shell commands in response

## Test plan
Basic event watching (existing functionality)
`meroctl context watch`

Command execution
`meroctl context watch -x "echo 'Event received'"`

Event count limiting
`meroctl context watch -x "date" -c 3`

### DIFF
--- a/crates/meroctl/src/cli/context/watch.rs
+++ b/crates/meroctl/src/cli/context/watch.rs
@@ -1,10 +1,11 @@
+use std::process::Command;
+
 use calimero_primitives::alias::Alias;
 use calimero_primitives::context::ContextId;
 use calimero_server_primitives::ws::{Request, RequestPayload, Response, SubscribeRequest};
 use clap::Parser;
 use eyre::{OptionExt, Result as EyreResult};
 use futures_util::{SinkExt, StreamExt};
-use std::process::Command;
 use tokio_tungstenite::connect_async;
 use tokio_tungstenite::tungstenite::Message as WsMessage;
 
@@ -77,7 +78,7 @@ impl WatchCommand {
         environment
             .output
             .write(&InfoLine(&format!("Subscribed to context {}", context_id)));
-        
+
         if let Some(cmd) = &self.exec {
             environment
                 .output
@@ -101,14 +102,14 @@ impl WatchCommand {
                     if let WsMessage::Text(text) = msg {
                         let response = serde_json::from_str::<Response>(&text)?;
                         environment.output.write(&response);
-                        
+
                         if let Some(cmd) = &self.exec {
                             let output = Command::new("sh")
                                 .arg("-c")
                                 .arg(cmd)
                                 .output()
                                 .map_err(|e| eyre::eyre!("Failed to execute command: {}", e))?;
-                            
+
                             if !output.status.success() {
                                 environment.output.write(&ErrorLine(&format!(
                                     "Command failed: {}",
@@ -121,14 +122,14 @@ impl WatchCommand {
                                 )));
                             }
                         }
-                        
+
                         event_count += 1;
                     }
                 }
                 Err(err) => {
-                    environment.output.write(&ErrorLine(&format!(
-                        "Error receiving message: {err}"
-                    )));
+                    environment
+                        .output
+                        .write(&ErrorLine(&format!("Error receiving message: {err}")));
                 }
             }
         }
--- a/crates/meroctl/src/cli/context/watch.rs
+++ b/crates/meroctl/src/cli/context/watch.rs
@@ -1,11 +1,10 @@
-use std::process::Command;
-
 use calimero_primitives::alias::Alias;
 use calimero_primitives::context::ContextId;
 use calimero_server_primitives::ws::{Request, RequestPayload, Response, SubscribeRequest};
 use clap::Parser;
 use eyre::{OptionExt, Result as EyreResult};
 use futures_util::{SinkExt, StreamExt};
+use tokio::process::Command;
 use tokio_tungstenite::connect_async;
 use tokio_tungstenite::tungstenite::Message as WsMessage;
 
@@ -108,6 +107,7 @@ impl WatchCommand {
                                 .arg("-c")
                                 .arg(cmd)
                                 .output()
+                                .await
                                 .map_err(|e| eyre::eyre!("Failed to execute command: {}", e))?;
 
                             if !output.status.success() {
--- a/crates/meroctl/src/cli/context/watch.rs
+++ b/crates/meroctl/src/cli/context/watch.rs
@@ -4,15 +4,16 @@ use calimero_server_primitives::ws::{Request, RequestPayload, Response, Subscrib
 use clap::Parser;
 use eyre::{OptionExt, Result as EyreResult};
 use futures_util::{SinkExt, StreamExt};
+use std::process::Command;
 use tokio_tungstenite::connect_async;
 use tokio_tungstenite::tungstenite::Message as WsMessage;
 
 use crate::cli::Environment;
 use crate::common::{fetch_multiaddr, load_config, multiaddr_to_url, resolve_alias};
-use crate::output::{InfoLine, Report};
+use crate::output::{ErrorLine, InfoLine, Report};
 
 #[derive(Debug, Parser)]
-#[command(about = "Watch events from a context")]
+#[command(about = "Watch events from a context and optionally execute commands")]
 pub struct WatchCommand {
     /// ContextId to stream events from
     #[arg(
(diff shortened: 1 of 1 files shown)


crates/meroctl/src/cli/context/event.rs
```diff
@@ -0,0 +1,109 @@
...
+impl EventCommand {
+    pub async fn run(self, environment: &Environment) -> EyreResult<()> {
+        let config = load_config(&environment.args.home, &environment.args.node_name)?;
+        let multiaddr = fetch_multiaddr(&config)?;
+        let client = reqwest::Client::new();
+
+        let mut event_count = 0;
+        let url = multiaddr_to_url(
+            multiaddr,
+            &format!("admin-api/dev/contexts/{}/events", self.context_id),
```
miraclx: this is not a valid path ([ref](https://github.com/calimero-network/core/blob/7ad0fc11329ddfdc8342f9627b4806d42468b87e/crates/server/src/admin/service.rs#L106))

crates/meroctl/src/cli/context/event.rs
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            let response = do_request::<(), serde_json::Value>(
+                &client,
+                url.clone(),
+                None,
+                &config.identity,
+                RequestType::Get,
+            )
+            .await;
```
miraclx: and I'm not sure I understand this, `null` request body?

crates/meroctl/src/cli/context/event.rs
```diff
@@ -0,0 +1,109 @@
+use std::process::Command;
```
miraclx: use tokio's [`Command`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html) instead of `std`s

crates/meroctl/src/cli/context/event.rs
```diff
@@ -0,0 +1,109 @@
...
+    )]
+    interval: u64,
+
+    #[arg(
+        short,
+        long,
+        help = "Maximum number of events to trigger (0 for unlimited)",
+        default_value = "0"
+    )]
+    count: u32,
```
miraclx:
```suggestion
        help = "Maximum number of events to trigger (0 for unlimited)",
        default_value_t
    )]
    count: u32,
```

I'd have preferred an `Option<NonZeroU32>` but I can think of some arguments in favor of what you did here, so it's fine for now

crates/meroctl/src/cli/context/event.rs
```diff
@@ -0,0 +1,109 @@
...
+    #[arg(help = "The shell command to execute when event occurs")]
+    command: String,
+
+    #[arg(
+        short,
+        long,
+        help = "Interval in seconds to check for events",
+        default_value = "1"
+    )]
+    interval: u64,
```
miraclx: the event stream is from a websocket connection, we won't need this

miraclx (commented):
we already have the `watch` command, which is where we should incorporate this functionality

and command execution should be optional via the `-x, --exec` flag

Nathy-bajo:
> we already have the `watch` command, which is where we should incorporate this functionality
> 
> and command execution should be optional via the `-x, --exec` flag

I see, that makes more sense.

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -69,19 +77,60 @@ impl WatchCommand {
...
             .output
             .write(&InfoLine("Streaming events (press Ctrl+C to stop):"));
 
+        let mut event_count = 0;
         while let Some(message) = read.next().await {
+            if let Some(max_count) = self.count {
+                if event_count >= max_count {
+                    break;
+                }
+            }
```
miraclx: how about we put these on L104?

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -21,6 +22,14 @@ pub struct WatchCommand {
         default_value = "default"
     )]
     pub context: Alias<ContextId>,
+
+    /// Command to execute when an event is received
+    #[arg(short, long, value_name = "COMMAND")]
+    pub exec: Option<String>,
```
miraclx:
```suggestion
    /// Command to execute when an event is received
    #[arg(short = "x", long, value_name = "COMMAND")]
    pub exec: Option<String>,
```

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -69,19 +77,60 @@ impl WatchCommand {
...
             match message {
                 Ok(msg) => {
                     if let WsMessage::Text(text) = msg {
                         let response = serde_json::from_str::<Response>(&text)?;
                         environment.output.write(&response);
+
+                        if let Some(cmd) = &self.exec {
+                            let output = Command::new("sh")
+                                .arg("-c")
+                                .arg(cmd)
```
miraclx: I don't like that this isn't cross-platform, and we shouldn't be using the shell implicitly anyway *.._*

crates/meroctl/src/cli/context/watch.rs
miraclx: *_..* let's make this an `Option<Vec<String>>` instead with `#[arg(num_args = 1..)]` *.._*

crates/meroctl/src/cli/context/watch.rs
miraclx: *_..* then this can become

```suggestion
                        if let Some(cmd) = &self.exec {
                            let output = Command::new(cmd[0])
                                .args(&cmd[1..])
```

then if the user wanted to use a shell, they can

miraclx (commented):
on linux, for example:

```console
meroctl context watch <context-id> \
    -x notify-send -i "The App" "State Changed"
```

Nathy-bajo:
@miraclx Done.

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -23,12 +23,12 @@ pub struct WatchCommand {
     )]
     pub context: Alias<ContextId>,
 
-    /// Command to execute when an event is received
-    #[arg(short, long, value_name = "COMMAND")]
-    pub exec: Option<String>,
+    /// Command to execute when an event is received (can specify multiple args)
```
miraclx, line 41: it's not very clear what this means, is it

```rust
watch -x do some thing
```

or

```rust
watch -x do -x some -x thing
```

or

```rust
watch -x do,some,thing
```

you can add an example like [here](https://github.com/calimero-network/core/blob/a271088b8c2570eef4161343b220f5596e5764c2/crates/meroctl/src/cli.rs#L29-L35) and drop the suffix here

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -23,12 +23,12 @@ pub struct WatchCommand {
...
+    #[arg(short = 'x', long, value_name = "COMMAND", num_args = 1..)]
+    pub exec: Option<Vec<String>>,
 
     /// Maximum number of events to process before exiting
-    #[arg(short, long, value_name = "COUNT")]
+    #[arg(short = 'n', long, value_name = "COUNT")]
     pub count: Option<usize>,
```
miraclx: 👍🏽

crates/meroctl/src/cli/context/watch.rs
```diff
@@ -90,25 +91,25 @@ impl WatchCommand {
...
+                                if event_count >= max_count {
+                                    break;
+                                }
+                            }
+
+                            let output = if cmd.len() == 1 {
+                                Command::new(&cmd[0]).output().await
+                            } else {
+                                Command::new(&cmd[0]).args(&cmd[1..]).output().await
+                            }
```
miraclx: you shouldn't need distinction between 1 and N, clap should enforce the presence of at least one entry in the vec, but you can test this

crates/meroctl/src/cli/context/watch.rs
Nathy-bajo: > it's not very clear what this means, is it
> 
> ```rust
> watch -x do some thing
> ```

Yes

crates/meroctl/src/cli/context/watch.rs
Nathy-bajo: I've removed the redundant length check

chefsale (approved):

crates/meroctl/src/cli/context/watch.rs
miraclx: I meant the "can specify multiple args" in the description for the flag will be confusing to an external reader.

The examples should suffice, so we can leave that suffix out.

PR #1226, opened 2025-04-21
feat(merod,meroctl): add version checks 

### LINKED ISSUE #1062: meroctl and merod new version check
### Current problem
- devs are not aware when there is new version of merod and meroctl downloaded through homebrew and installation scriot.
- we are receiving a lot of issues caused by using old version

### Proposal
- add check before using merod and meroctl and inform devs if they are not using latest version and how to obtain it

### Goal
- assure that devs are aware if they are not using latest merod and meroctl release

cy4n1d3-p1x3l:
## Description
This PR fixes #1062. The users are now reminded every 24h to update their merod and meroctl binaries to avoid them having issues with older versions of these tools.

## Test Plan 
Tested locally.

## Documentation Update
Not needed.


crates/meroctl/src/version.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    if let Some(latest) = fetch_latest_version().await {
+        let current = match Version::parse(CURRENT_VERSION) {
+            Ok(v) => v,
+            Err(_) => return,
+        };
+
+        if latest > current {
+            println!("\n🔔 New version of meroctl available: v{latest} (current: v{current})");
+            println!("💡 To update: brew upgrade meroctl (or rerun your installer)\n");
```
miraclx: this assumes they're on mac? we can skip the platform specific instructions for now and just say there's an update

miraclx (commented):
1. we shouldn't need a special file just for this
2. maybe we pick a random number in a range of 1..10, and if it's 5, check for updates
3. to avoid slowing down the node startup, use `tokio::spawn` on the checker task
4. would be good to duplicate the version check logic ([#1212](https://github.com/calimero-network/core/pull/1212#issuecomment-2823472946) defines a `calimero-version` crate which could be used, but it's still WIP)

crates/meroctl/src/version.rs
cy4n1d3-p1x3l: Yess that can be updated.

cy4n1d3-p1x3l:
> 1. we shouldn't need a special file just for this
> 
>     2. maybe we pick a random number in a range of 1..10, and if it's 5, check for updates
> 
>     3. to avoid slowing down the node startup, use `tokio::spawn` on the checker task
> 
>     4. would be good to duplicate the version check logic ([#1212](https://github.com/calimero-network/core/pull/1212#issuecomment-2823472946) defines a `calimero-version` crate which could be used, but it's still WIP)

is the random number logic good for now or should i implement as #1212 was going to implement?

miraclx:
> is the random number logic good for now or should i ...

the random number logic is to determine when to check for updates, it shouldn't be blocked on #1212

> ...implement as https://github.com/calimero-network/core/pull/1212 was going to implement

as far as #1212, the incorporation proposition was to save us the hassle of duplicating the logic, for now you can maintain the duplication while we find some way to resume the effort in #1212, after which we can deduplicate it

cy4n1d3-p1x3l:
> > is the random number logic good for now or should i ...
> 
> the random number logic is to determine when to check for updates, it shouldn't be blocked on #1212
> 
> > ...implement as #1212 was going to implement
> 
> as far as #1212, the incorporation proposition was to save us the hassle of duplicating the logic, for now you can maintain the duplication while we find some way to resume the effort in #1212, after which we can deduplicate it

okay

crates/meroctl/src/version.rs
```diff
@@ -1,82 +1,50 @@
...
-    let cache: Option<ToolVersionInfo> = read_cache(&cache_path);
-
-    if let Some(info) = &cache {
-        let hours_since = (Utc::now() - info.last_checked).num_hours();
-        if hours_since < CHECK_INTERVAL_HOURS {
-            return;
-        }
+    let mut rng = rand::thread_rng();
+    let n: u8 = rng.gen();
+    if n != 11 {
```
miraclx: the odds of this is 1 in 255, is that intentional?

crates/meroctl/src/version.rs
```diff
@@ -1,82 +1,50 @@
...
 const CURRENT_VERSION: &str = env!("CARGO_PKG_VERSION");
-const CHECK_INTERVAL_HOURS: i64 = 24;
-
-#[derive(Serialize, Deserialize, Debug)]
-struct ToolVersionInfo {
-    version: String,
-    last_checked: DateTime<Utc>,
-}
 
 pub async fn check_for_update() {
```
miraclx: return `eyre::Result`

crates/meroctl/src/version.rs
```diff
@@ -1,82 +1,50 @@
...
         .get(url)
-        .header("User-Agent", format!("meroctl-version-check"))
+        .header("User-Agent", "meroctl-version-check")
         .send()
         .await
-        .ok()?;
+    {
+        Ok(resp) => resp,
+        Err(_) => return,
+    };
```
miraclx: so we can use `?` here and others

crates/merod/src/main.rs
```diff
@@ -7,16 +7,20 @@ use tracing_subscriber::prelude::*;
...
 mod defaults;
+mod version;
 
 #[tokio::main]
 async fn main() -> EyreResult<()> {
     setup()?;
 
     let command = RootCommand::parse();
 
+    check_for_update().await;
```
miraclx: `tokio::spawn` this task

```suggestion
    let _ignored = tokio::spawn(async {
        if let Err(err) = check_for_update().await {
            error!(.....);
        }
    });
```

crates/meroctl/src/version.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    let current = match Version::parse(CURRENT_VERSION) {
+        Ok(v) => v,
+        Err(_) => return,
+    };
+
+    if latest > current {
+        println!("A new version {latest} for meroctl is available.");
```
miraclx: the old version was pretty nice

CleanShot 2025-05-09 at 10 26 04@2x

crates/meroctl/src/version.rs
cy4n1d3-p1x3l: we wouldn't want the user to be repeatedly reported about this. or do we?

miraclx: yeah, not too frequently, as you have it is fine, thanks!

crates/meroctl/src/main.rs
```diff
@@ -18,10 +21,18 @@ async fn main() -> ExitCode {
...
+    let client = Client::new();
+    let mut rng = rand::thread_rng();
+    let n: u8 = rng.gen();
+    if n != 11 {
+        spawn(async move {
+            if let Err(err) = check_for_update(&client.clone()).await {
+                eprintln!("Version check failed: {}", err);
+            }
+        });
+    }
```
miraclx: `spawn` is only for `merod`, not `meroctl`

and use it as `tokio::spawn` directly, without the `use` import.

crates/merod/src/main.rs
```diff
@@ -2,21 +2,37 @@ use std::env::var;
...
+    let client = Client::new();
+    let mut rng = rand::thread_rng();
+    let n: u8 = rng.gen();
+    if n != 11 {
+        spawn(async move {
+            if let Err(err) = check_for_update(&client.clone()).await {
+                eprintln!("Version check failed: {}", err);
+            }
+        });
+    }
```
miraclx: 1. just realized the comparison logic is `!=`, so it will almost always do this check, we can make it `rand % 20` for a 12.75% chance of checking?
2. there's no reason to have the client constructed on the caller, if it's not reused elsewhere

```suggestion
    if rand::random::<u8>() % 20 {
        tokio::spawn(async move {
            if let Err(err) = check_for_update().await {
                eprintln!("Version check failed: {}", err);
            }
        });
    }
```

crates/merod/src/version.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    let response = client
+        .get(url)
+        .header("User-Agent", "merod-version-check")
+        .send()
+        .await?;
+    #[derive(Deserialize)]
+    struct Release {
+        tag_name: String,
+    }
```
miraclx: pull this out of the function

crates/merod/src/version.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    let release: Release = response.json().await?;
+    let latest = Version::parse(&release.tag_name)?;
+    let current = Version::parse(CURRENT_VERSION)?;
```
miraclx: why not use `Version` in the struct directly? (you'd need the `serde` feature flag for the `semver` crate)

```suggestion
    #[derive(Deserialize)]
    struct Release {
        tag_name: Version,
    }

    let release: Release = response.json().await?;
```

and as for `CURRENT_VERSION`, use a static variable instead of the const (see [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html))

Cargo.toml
```diff
@@ -101,8 +101,7 @@ paste = "1.0.15"
 prettyplease = "0.2.17"
 proc-macro2 = "1.0"
 quote = "1.0.37"
-rand = "0.8.5"
-rand_chacha = "0.3.1"
+rand = { version = "0.8", features = ["std"] }
```
miraclx: localize the features you need to the crate that needs them

but `std` is already a default feature, so this is not needed

crates/merod/src/main.rs
cy4n1d3-p1x3l: yeah that was left from testing,  %20 sounds good

Cargo.toml
```diff
@@ -97,7 +97,7 @@ paste = "1.0.15"
 prettyplease = "0.2.17"
 proc-macro2 = "1.0"
 quote = "1.0.37"
-rand = { version = "0.8", features = ["std"] }
+rand = { version = "0.8.5" }
```
miraclx:
```suggestion
rand = "0.8.5"
```

crates/meroctl/src/main.rs
```diff
@@ -16,8 +20,16 @@ async fn main() -> ExitCode {
         return ExitCode::FAILURE;
     }
 
-    let command = RootCommand::parse();
+    let client = Client::new();
+    let mut rng = rand::thread_rng();
+    let mut n: u8 = rng.gen();
+    if (n % 20) == 11 {
```
miraclx: why 11 here? it's probabilistically equivalent to 0, but it requires some proving, unlike 0

while proving it, I also realized the `%20 == 0` had fewer odds of selection than I initially stated

```text
div = 256 / 20 = 12.8
int = trunc(div) = 12
frac = div - int = 0.8
part = frac * 20 = 16
```

with this, the first `frac` numbers `[0, 15]` will be `ceil(div) = 13` / `256` = `5.08%`
and the rest, will be `int / 256` = `4.69%`

```python
import math

def odds(num, base, check):
    div = num / base
    int = math.trunc(div)
    needle = round(div - int, 5) * base
    count = math.ceil(div) if check < needle else int
    return count / num * 100

>>> odds(256, 20, 0)
5.078125

>>> odds(256, 20, 11)
5.078125

>>> odds(256, 20, 16)
4.6875
```

let's make it `%10 == 0`, for a ~10% chance

```python
>>> odds(256, 10, 0)
10.15625
```

crates/merod/src/version.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    let response = client
+        .get(url)
+        .header("User-Agent", "merod-version-check")
+        .send()
+        .await?;
+
+    let release: Release = response.json().await?;
+
+    if *&release.tag_name > *CURRENT_VERSION {
```
miraclx: these cancel out

```suggestion
    if release.tag_name > *CURRENT_VERSION {
```

crates/merod/src/version.rs
```diff
@@ -0,0 +1,35 @@
...
+        println!(
+            "\n🔔 New version of merod available: v{} (current: v{})",
+            *&release.tag_name, *CURRENT_VERSION
```
miraclx: you have it here too

crates/merod/src/main.rs
```diff
@@ -2,21 +2,36 @@ use std::env::var;
...
 #[tokio::main]
 async fn main() -> EyreResult<()> {
     setup()?;
 
     let command = RootCommand::parse();
 
+    let client = Client::new();
+    let mut rng = rand::thread_rng();
+    let mut n: u8 = rng.gen();
+    if (n % 20) == 11 {
```
miraclx: also, this is a lot of boilerplate for the small thing this does, so please compress these and the client should be constructed inside `check_for_update` not here

```suggestion
    if rand::random::<u8>() % 10 == 0 {
```

same applies to meroctl

crates/meroctl/src/main.rs
cy4n1d3-p1x3l: okayy

crates/merod/Cargo.toml
```diff
@@ -12,6 +12,7 @@ license.workspace = true
 alloy.workspace = true
 axum.workspace = true
 camino = { workspace = true, features = ["serde1"] }
+chrono.workspace = true
```
miraclx: This is unused?

miraclx (approved):
Rest lgtm

crates/merod/Cargo.toml
cy4n1d3-p1x3l: leftover from previous implementation

crates/meroctl/src/main.rs
```diff
@@ -1,13 +1,17 @@
 use std::process::ExitCode;
 
 use clap::Parser;
+use rand::Rng;
+use reqwest::Client;
 
 use crate::cli::RootCommand;
+use crate::version::check_for_update;
```
miraclx: these should actually be on line 15, since they're symbols from submodules

```suggestion
use cli::RootCommand;
use version::check_for_update;
```

crates/merod/src/main.rs
```diff
@@ -2,21 +2,33 @@ use std::env::var;
...
 use clap::Parser;
 use eyre::Result as EyreResult;
+use rand::Rng;
+use reqwest::Client;
 use tracing_subscriber::fmt::layer;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::{registry, EnvFilter};
 
 use crate::cli::RootCommand;
+use crate::version::check_for_update;
```
miraclx: same here

miraclx (commented):
also, would you be open to do the consolidation of the code here into the `calimero-version` crate when https://github.com/calimero-network/core/pull/1257#pullrequestreview-2862650300 lands?

this is not a blocker for this PR, which is why I've approved, just checking

cy4n1d3-p1x3l:
> also, would you be open to do the consolidation of the code here into the `calimero-version` crate when https://github.com/calimero-network/core/pull/1257#pullrequestreview-2862650300 lands?
> 
> this is not a blocker for this PR, which is why I've approved, just checking

Yeah can do. You mean to deduplicate the logic between merod and meroctl and have a uniform version check using the version crate?

crates/meroctl/src/main.rs
cy4n1d3-p1x3l: I thought ```cargo +nightly fmt``` fixes that

miraclx: the rules don't extend to this unfortunately, it can keep it in-place, but a human has to move them

miraclx:
Yup! That's exactly it.

crates/meroctl/src/main.rs
cy4n1d3-p1x3l: okayy

PR #1228, opened 2025-04-21
feat(meroctl): Implement pretty printing for command outputs

### LINKED ISSUE #902: Improve meroctl Report implementations
Ideally, reports should be pretty in meroctl. We need to define what "pretty" means and ideally implement some tooling to help pretty print objects.

Original comment: https://github.com/calimero-network/core/pull/888#discussion_r1824105584

Nathy-bajo:
## Description

Attempts to resolve #902

Update the Output struct to handle pretty printing:

- Add a new Format::PrettyText variant

- Modify the write() method to support the new format

- Implement pretty printing for all existing report types:

- Use colors, borders, and consistent formatting


frdomovic:
Could you also update the test plan please? Thank you!

Additionally some of the output prints
Screenshot 2025-04-22 at 11 40 18
Screenshot 2025-04-22 at 11 41 04

crates/meroctl/src/output.rs
```diff
@@ -31,7 +97,7 @@ impl Output {
                 Ok(json) => println!("{json}"),
                 Err(err) => eprintln!("Failed to serialize to JSON: {err}"),
             },
-            Format::PlainText => value.report(),
+            Format::PlainText => value.pretty_report(),
```
fbozic: The implementation of pretty print should be similar to json serialization. So instead of calling `value.pretty_report()`, we would call `table::to_string(&value)`. This change will also impact the `Report` trait, which might not be needed anymore. The benefit of this approach is that we don't have to implement `Report` for each struct, but we get that out of the box.

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+
+impl Report for PrettyTable {
+    fn report(&self) {
+        self.print();
+    }
+}
+
+pub struct PrettyTable {
+    headers: Vec<String>,
+    rows: Vec<Vec<String>>,
```
miraclx: 1. the struct should come before it's implementation
2. you can constrain the columns with const-generics

	```suggestion
	pub struct PrettyTable<'a, const N: usize> {
	    headers: [Cow<'a, str>, N],
	    rows: Vec<[Cow<'a, str>, N]>,
	```

the `Cow` let's you use a `&str` where you have it, and a `String` when you don't

crates/meroctl/src/output.rs
miraclx: This sounds interesting, but the issue with this, imo, is:

1. json serialization could be deeply nested
2. we don't need all items in the API response to be printed, just a select few
3. the types that impl `Report` aren't defined in this crate

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+
+impl PrettyTable {
+    pub fn new(headers: &[&str]) -> Self {
+        Self {
+            headers: headers.iter().map(|s| s.to_string()).collect(),
+            rows: Vec::new(),
+        }
+    }
+
+    pub fn add_row(&mut self, row: Vec<String>) {
```
miraclx:
```suggestion
    pub fn add_row(&mut self, row: [impl Into<Cow<'a, str>>; N]) {
```

you can use [`[T; N]::map`](https://doc.rust-lang.org/std/primitive.array.html#method.map) to make the `Into::into` conversion

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+        self.rows.push(row);
+    }
+
+    pub fn print(&self) {
+        if self.headers.is_empty() || self.rows.is_empty() {
+            return;
+        }
```
miraclx: with the changes, we now we know `self.headers` will never be empty

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+
+        let mut widths: Vec<usize> = self.headers.iter().map(|h| h.len()).collect();
+
+        for row in &self.rows {
+            for (i, cell) in row.iter().enumerate() {
+                if i < widths.len() {
+                    widths[i] = widths[i].max(cell.len());
+                }
```
miraclx: we also know `i` will never be larger than `widths.len()`

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+        let header = self
+            .headers
+            .iter()
+            .enumerate()
+            .map(|(i, h)| format!(" {:width$} ", h.bold().blue(), width = widths[i]))
+            .collect::<Vec<_>>()
+            .join("│");
+
+        println!("{}", header);
+        println!("{}", "─".repeat(header.len()));
```
miraclx: this didn't account for the ANSI colors that made up the colors of the headers

widths + ( separators + 2 spaces per separator )

```suggestion
        let header_width = widths.iter().sum() + (N - 1) * 3;
        println!("{}", "─".repeat(header_width));
```

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+    pub fn add_row(&mut self, row: Vec<String>) {
+        self.rows.push(row);
+    }
+
+    pub fn print(&self) {
+        if self.headers.is_empty() || self.rows.is_empty() {
+            return;
+        }
+
+        let mut widths: Vec<usize> = self.headers.iter().map(|h| h.len()).collect();
```
miraclx:
```suggestion
        let mut widths: [usize; N] = self.headers.map(|h| h.len());
```

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+            }
+        }
+
+        let header = self
+            .headers
+            .iter()
+            .enumerate()
+            .map(|(i, h)| format!(" {:width$} ", h.bold().blue(), width = widths[i]))
+            .collect::<Vec<_>>()
+            .join("│");
```
miraclx:
```suggestion
        for (i, (w, c)) in widths.iter().zip(self.headers.iter()).enumerate() {
            print!(
                "{} {:width$} ",
                c.bold().blue(),
                width = w,
                (i > 0).then_some("|").unwrap_or(""));
        }
        println!("");
```

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
...
+            println!("{}", row_str);
+        }
+    }
+}
+
+impl Report for PrettyTable {
+    fn report(&self) {
+        self.print();
+    }
+}
```
miraclx: I don't think this is used

crates/meroctl/src/output.rs
```diff
@@ -16,6 +16,72 @@ pub struct Output {
 
 pub trait Report {
     fn report(&self);
+
+    fn pretty_report(&self) {
+        self.report();
+    }
+}
```
miraclx: there's no other use of report than to do exactly what `pretty_report` does, we should only need one of these.

I'd say leave as-is.

crates/meroctl/src/cli/context/invite.rs
```diff
@@ -43,9 +44,12 @@ impl Report for InviteToContextResponse {
     fn report(&self) {
         match self.data {
             Some(ref payload) => {
-                println!("Invitation payload: {}", payload)
+                println!("{} {}", "✓".green(), "Invitation created".bold());
+                println!("  Payload: {}", payload.to_string().cyan());
```
```suggestion
                println!("  Payload: {}", payload.cyan());
```

crates/meroctl/src/cli/context/join.rs
```diff
@@ -35,10 +36,16 @@ impl Report for JoinContextResponse {
     fn report(&self) {
         match self.data {
             Some(ref payload) => {
-                println!("context_id {}", payload.context_id);
-                println!("member_public_key: {}", payload.member_public_key);
+                println!("{} {}", "✓".green(), "Successfully joined context".bold());
+                println!("  Context ID: {}", payload.context_id.to_string().cyan());
+                println!(
+                    "  Member Public Key: {}",
+                    payload.member_public_key.to_string().cyan()
```
miraclx:
```suggestion
                    payload.member_public_key.cyan()
```

crates/meroctl/src/cli/identity/generate.rs
```diff
@@ -12,8 +13,12 @@ pub struct GenerateCommand;
 
 impl Report for GenerateContextIdentityResponse {
     fn report(&self) {
-        println!("public_key: {}", self.data.public_key);
-        println!("private_key: {}", self.data.private_key);
+        println!("{} {}", "✓".green(), "Key pair generated".bold());
+        println!("  Public Key: {}", self.data.public_key.to_string().cyan());
```
miraclx:
```suggestion
        println!("  Public Key: {}", self.data.public_key.cyan());
```

crates/meroctl/src/cli/identity/generate.rs
```diff
@@ -12,8 +13,12 @@ pub struct GenerateCommand;
...
+        println!(
+            "  Private Key: {}",
+            self.data.private_key.to_string().cyan()
```
miraclx: `OwoColorize` is [implemented](https://docs.rs/owo-colors/latest/owo_colors/trait.OwoColorize.html#impl-OwoColorize-for-D) for any generic `D` unconditionally

and return types of its methods are `Display` if `D: Display` ([example](https://docs.rs/owo-colors/latest/owo_colors/struct.FgColorDisplay.html#impl-Display-for-FgColorDisplay%3C'a,+Color,+T%3E))

so if you can call `.to_string()` before calling any `OwoColorize` method like `.cyan()`, then you never need to call `.to_string()`

miraclx (commented):
to make sure you've exhaustively covered all `Report` impls, remove the `report` method on the trait, which should show all impls that still have it

you can then go through them, progressively updating them to `pretty_report`

after everything is covered, you can rename `pretty_report` back to `report`

Nathy-bajo:
@miraclx I tried a different approach. Please review!

crates/meroctl/Cargo.toml
```diff
@@ -16,6 +16,7 @@ clap = { workspace = true, features = ["env", "derive"] }
 color-eyre.workspace = true
 const_format.workspace = true
 dirs.workspace = true
+comfy-table = "7.0"
```
fbozic: Add `comfy-table` to root Cargo.toml and in the crate Cargo.toml use only `comfy-table = { workspace = true, features = [...]} `
And small nit, order it alphabetically, both in root and crate Cargo.toml

crates/meroctl/src/cli/app/list.rs
```diff
@@ -17,6 +18,40 @@ impl Report for ListApplicationsResponse {
...
+            if !app.metadata.is_empty() {
+                let mut meta_table = Table::new();
+                let _ = meta_table.set_header(vec![
+                    Cell::new(format!("Metadata for {}", app.id)).fg(Color::Green)
+                ]);
+
+                for data in &app.metadata {
+                    let _ = meta_table.add_row(vec![format!("{:?}", data)]);
+                }
+                println!("\n{meta_table}");
```
fbozic: Do we need extra `\n`?

crates/meroctl/Cargo.toml
Nathy-bajo: @fbozic Alright!

crates/meroctl/Cargo.toml
```diff
@@ -16,6 +16,7 @@ clap = { workspace = true, features = ["env", "derive"] }
 color-eyre.workspace = true
 const_format.workspace = true
 dirs.workspace = true
+comfy-table = { workspace = true }
```
fbozic: ~nit: alphabetical order

miraclx:
```suggestion
comfy-table.workspace = true
```

crates/meroctl/src/output.rs
miraclx: This isn't yet resolved, we still have both `report` and `pretty_report`, even though only one of them is being used.

Remove `pretty_report`, and keep the introduced functionality in `report`.

I'm not convinced we need to maintain both formats, unless there's something I'm missing cc: @fbozic

crates/meroctl/src/cli/app.rs
```diff
@@ -51,6 +52,21 @@ impl Report for Application {
...
+        let _ = table.add_row(vec![format!("ID: {}", self.id)]);
+        let _ = table.add_row(vec![format!("Size: {}", self.size)]);
+        let _ = table.add_row(vec![format!("Blob ID: {}", self.blob)]);
+        let _ = table.add_row(vec![format!("Source: {}", self.source)]);
+        let _ = table.add_row(vec!["Metadata:"]);
+
+        for item in &self.metadata {
+            let _ = table.add_row(vec![format!("  - {item:?}")]);
+        }
+        println!("{table}");
```
miraclx: Is there any value to printing a single column as a table? different from what we already had?

At least break up the "type" and "data" into adjacent columns instead of merging them in the same cell.

crates/meroctl/src/cli/app/get.rs
```diff
@@ -27,6 +28,19 @@ impl Report for GetApplicationResponse {
...
+        let mut table = Table::new();
+        let _ = table.set_header(vec![Cell::new("Application").fg(Color::Blue)]);
+
+        match &self.data.application {
+            Some(app) => app.pretty_report(),
+            None => {
+                let _ = table.add_row(vec!["No application found"]);
+                println!("{table}");
+            }
+        }
```
miraclx: I don't know about this, I don't think everything has to be a table

miraclx (approved):

miraclx:
For posterity, can you update the PR description with some screenshots, thanks! ([ref](https://github.com/calimero-network/core/pull/1228#pullrequestreview-2794140430))

crates/meroctl/src/output.rs
```diff
@@ -1,12 +1,12 @@
...
-use color_eyre::owo_colors::OwoColorize;
+use comfy_table::{Cell, Table};
 use serde::Serialize;
 
 #[derive(Clone, Copy, Debug, Default, ValueEnum)]
 pub enum Format {
     Json,
     #[default]
-    PlainText,
+    PrettyText,
```
miraclx: this will require a documentation update (`--output-format pretty-text`), since there's no other option to text than pretty text, I'd say we make this just `Text` or better yet, `Human`

cc: @fbozic

crates/meroctl/src/output.rs
```diff
@@ -41,7 +41,10 @@ pub struct InfoLine<'a>(pub &'a str);
 
 impl Report for InfoLine<'_> {
     fn report(&self) {
-        println!("{} {}", "[INFO]".green(), self.0);
+        let mut table = Table::new();
+        let _ = table.set_header(vec![Cell::new("INFO").fg(comfy_table::Color::Green)]);
+        let _ = table.add_row(vec![self.0]);
+        println!("{table}");
```
miraclx: I don't think we want a table here since this is a way to show unstructured data (that already doesn't have a JSON representation). The same applies to `ErrorLine`, but before you change it, let's see some screenshots.

cc: @fbozic

Nathy-bajo: I'll send them once #1239 gets merged. I was trying to set it up and kept getting that error but I never they were outdated.

Or maybe if you can test it on your end

crates/meroctl/src/output.rs
Nathy-bajo: Done.

crates/meroctl/src/output.rs
frdomovic: @Nathy-bajo #1239 will only cover 2 help commands which had additional "--" in help table fyi.

miraclx: @Nathy-bajo, unsure if you linked the right PR, but that's merged now

Nathy-bajo:
Screenshot 2025-04-30 at 22 02 20
Screenshot 2025-04-30 at 22 10 26
Screenshot 2025-04-30 at 22 11 26
Screenshot 2025-04-30 at 22 12 40
Screenshot 2025-04-30 at 22 29 06
Screenshot 2025-04-30 at 22 30 53



@miraclx

frdomovic:
> Screenshot 2025-04-30 at 22 02 20 Screenshot 2025-04-30 at 22 10 26 Screenshot 2025-04-30 at 22 11 26 Screenshot 2025-04-30 at 22 12 40 Screenshot 2025-04-30 at 22 29 06 Screenshot 2025-04-30 at 22 30 53
> @miraclx

Thanks for screenshots, looks good!

frdomovic (approved):

crates/meroctl/src/output.rs
miraclx: @Nathy-bajo, this is still unresolved, let's not use a table view for `InfoLine`, `ErrorLine` and `WarnLine`

they are specifically for unstructured outputs. everything else is for representing structured data.

Nathy-bajo: @miraclx I've reverted the changes for them.

crates/meroctl/src/output.rs
```diff
@@ -1,5 +1,5 @@
 use clap::ValueEnum;
-use comfy_table::{Cell, Table};
+use color_eyre::owo_colors::OwoColorize;
```
miraclx:
```suggestion
use owo_colors::OwoColorize;
```

miraclx (approved):
While I don't think ***everything*** needs to be tabulated, especially single-column reports for which I expected a untabulated, yet, pretty format, this is fine for now while we iteratively identify which pieces are unnecessarily tabulated for a later refactor.

cc: @fbozic for a final review

Nathy-bajo:
@frdomovic @fbozic @chefsale Ready for final review.

PR #1236, opened 2025-04-26
feat(e2e) : improve comment generation

### LINKED ISSUE #1166: Improve generation of comments from e2e tests
Currently, every e2e scenario report is written in separate comment, we would need to combine them in one comment for better overview.

cy4n1d3-p1x3l:
# [e2e tests] Improve generation of comments and support for multiple applications

## Description

The e2e tests now run for multiple applications. Fixes #1166 

## Test plan

Testing done locally.

## Documentation update

Not needed as of now.


cy4n1d3-p1x3l:
Workflows not updated yet. Waiting for a review @fbozic @chefsale @miraclx.

e2e-tests/src/protocol/near.rs
```diff
@@ -72,9 +73,11 @@ impl NearSandboxEnvironment {
     }
 
     pub async fn node_args(&self, node_name: &str) -> EyreResult<Vec<String>> {
+        let timestamp = Utc::now().timestamp();
+        let unique_node_name = format!("{}-{}", node_name, timestamp);
```
fbozic: I don't understand this change. Is it leftover from testing? One protocol sandbox should be enough for multiple apps that are being tested.

cy4n1d3-p1x3l: yeah it should be
initially testing was giving me errors

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
 
-        // Run scenarios directory by directory
-        let scenarios_dir = self.environment.input_dir.join("scenarios");
-        let mut entries = read_dir(scenarios_dir).await?;
-
-        while let Some(entry) = entries.next_entry().await? {
-            let path = entry.path();
-            if path.is_dir() {
-                let test_file_path = path.join("test.json");
+        let scenarios_dir = self.environment.input_dir.join("protocols");
```
fbozic: `protocols_dir` would probably be better name for variable

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+
+        for protocol_name in &self.environment.scenarios {
```
fbozic: I think you are not checking anymore if `--protocols` (`self.environment.protocols`) includes a protocol for which we are setting up the sandbox.   
Check this comment: https://github.com/calimero-network/core/pull/1236#pullrequestreview-2806469272, the simplest way should be to iterate over `self.environment.protocols`.

cy4n1d3-p1x3l: i did not yet rename scenario to protocol in many places. Is the confusion arising because of that?
Because i've implemented exactly like you commented.
The tests ran fine once which generated the report we can see in the comments but after i pushed another commit the tests failed.

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
-        let mut entries = read_dir(scenarios_dir).await?;
-
-        while let Some(entry) = entries.next_entry().await? {
-            let path = entry.path();
-            if path.is_dir() {
-                let test_file_path = path.join("test.json");
+        let protocols_dir = self.environment.input_dir.join("protocols");
+
+        for protocol_name in &self.environment.protocols {
+            let protocol_name = protocol_name.to_string().to_lowercase();
```
miraclx: impl `Eq, PartialEq` instead of string comparison

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+                    ProtocolSandboxConfig::Stellar(_) => "stellar",
+                    ProtocolSandboxConfig::Near(_) => "near",
+                    ProtocolSandboxConfig::Icp(_) => "icp",
+                    ProtocolSandboxConfig::Ethereum(_) => "ethereum",
+                };
+
+                if config_protocol_name == protocol_name {
+                    let sandbox_env = match sandbox_cfg {
+                        ProtocolSandboxConfig::Stellar(config) => {
+                            ProtocolSandboxEnvironment::Stellar(StellarSandboxEnvironment::init(
```
miraclx: here, there's no guarantee we don't call `init` twice for the same protocol

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+                        }
+                    };
+
+                    initialized_protocols.insert(protocol_name.clone(), sandbox_env);
+                    break;
+                }
+            }
+        }
+
+        for protocol_name in &self.environment.protocols {
```
miraclx, line 142: why iterate over the same list twice?

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+            let protocol_name = protocol_name.to_string().to_lowercase();
+            let protocol_path = protocols_dir.join(&protocol_name);
+
+            if !protocol_path.exists() || !protocol_path.is_dir() {
```
miraclx: [`Utf8Path::is_dir`](https://docs.rs/camino/latest/camino/struct.Utf8PathBuf.html#method.is_dir) will only be true if it exists and is a dir

```suggestion
            if !protocol_path.is_dir() {
```

e2e-tests/src/main.rs
```diff
@@ -92,26 +92,26 @@ pub struct RootArgs {
...
-            Scenario::Near => write!(f, "near"),
-            Scenario::Stellar => write!(f, "stellar"),
-            Scenario::Icp => write!(f, "icp"),
+            Protocol::Ethereum => write!(f, "ethereum"),
+            Protocol::Near => write!(f, "near"),
+            Protocol::Stellar => write!(f, "stellar"),
+            Protocol::Icp => write!(f, "icp"),
         }
     }
 }
```
miraclx: first, introduce a `Protocol::as_str` method which you can then call from here (if it's absolutely necessary, I'm not convinced it is)

e2e-tests/src/driver.rs
```diff
@@ -515,54 +497,46 @@ impl TestRunReport {
...
                         None => "-",
                         Some(None) => ":fast_forward:",
                         Some(Some(Ok(_))) => ":white_check_mark:",
                         Some(Some(Err(_))) => ":x:",
                     };
-
                     write!(&mut markdown, " {result} |")?;
                 }
-                writeln!(&mut markdown)?;
+                writeln!(&mut markdown, "\n")?;
```
miraclx, line 541: this is `writeln!`, so unless you want two newlines, it's not necessary

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+                    .output_writer
+                    .write_str(&format!("No directory for protocol: {}", protocol_name));
+                continue;
+            }
+
+            let Some(sandbox) = initialized_protocols.get(&protocol_name) else {
+                bail!("Sandbox not initialized for protocol: {}", protocol_name);
+            };
+
+            let mero = self.setup_mero(&vec![sandbox.clone()]).await?;
```
miraclx: 1. you can change `setup_mero` to accept a `&[_]` instead of a `&Vec<_>`

    ```suggestion
                let mero = self.setup_mero(&[sandbox.clone()]).await?;
    ```

    and finally you won't even need to clone if you took in a `&[impl AsRef<ProtocolSandboxEnvironment>]`

2. won't this mean we'll only ever have nodes with 1 protocol configuration? even though we can specify `--protocols near,ethereum`, it'll only setup `node1` with `near`, and `node2` with `ethereum`, forcing `node1` to use the relayer when it needs to do any ethereum operations and vice-versa, correct me if I'm wrong

cc: @fbozic

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+            let Some((inviter, invitees)) = self.pick_inviter_node(&mero.ds) else {
+                bail!("Not enough nodes to run test for protocol {protocol_name}")
+            };
+
+            self.environment
+                .output_writer
+                .write_str(&format!("Picked inviter: {inviter}"));
+            self.environment
+                .output_writer
+                .write_str(&format!("Picked invitees: {invitees:?}"));
```
miraclx, line 174: we'll also be printing these N times for every protocol

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+
+            let mut applications = read_dir(&protocol_path).await?;
+            while let Some(app) = applications.next_entry().await? {
```
miraclx, line 177: there's no checks to make sure the contents of the protocol path is a file

e2e-tests/src/driver.rs
```diff
@@ -97,131 +97,113 @@ impl Driver {
...
+                    self.environment.output_writer,
+                    &protocol_name,
+                    None,
+                    sandbox,
+                );
+                let test_file_path = app.path();
+                let app_name = test_file_path
+                    .file_stem()
+                    .and_then(|s| s.to_str())
+                    .unwrap_or_default();
```
miraclx: `default` (which is an empty string) is not a good default here

`None` is only returned when there wasn't a file name, we can either ignore such cases or throw an error, but defaulting isn't it

https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.file_stem:~:text=The%20stem%20is%3A-,None%2C%20if%20there%20is%20no%20file%20name%3B,-The%20entire%20file

e2e-tests/src/driver.rs
cy4n1d3-p1x3l: there is a initialized_protocol list for that purpose

miraclx: ah that's a bit hard to read, but I see most of this logic was pretty much here before with some improvements, and that's fair

can you use the bail-early approach you did here:

```rust
            if initialized_protocols.contains_key(&protocol_name) {
                continue;
            }
```

to this too:

```rust
                if config_protocol_name != protocol_name {
                    continue;
                }
                
                let sandbox_env = match __;
```

e2e-tests/src/driver.rs
cy4n1d3-p1x3l: yeah added purposefully
the md for different applications wont display correctly otherwise

e2e-tests/src/driver.rs
cy4n1d3-p1x3l: there can be two ways to implement
1. Initialize the required sandboxes all first and then run the required tests on the apps.
2. Iterate over the protocols only once and initialize the sanbox and run required tests.

the 1st one is implemented as suggested by @fbozic and fits with the earlier implementation.

What is the better way? cc @miraclx @fbozic

e2e-tests/src/driver.rs
cy4n1d3-p1x3l: if we tend to run all the apps on a single set of nodes then it would be printed only once per protocol.

e2e-tests/src/driver.rs
cy4n1d3-p1x3l: returning an error looks good.

e2e-tests/src/driver.rs
```diff
@@ -41,6 +41,14 @@ pub struct TestContext<'a> {
     pub proposal_id: Option<String>,
 }
 
+#[derive(Debug, Hash, Clone, clap::ValueEnum, Copy, PartialEq, Eq)]
+pub enum Protocol {
+    Ethereum,
+    Near,
+    Stellar,
+    Icp,
+}
```
miraclx: ~nit: I find it best to keep things that are exposed to the CLI all in one place, as it's easy to forget the `clap::ValueEnum` flag

so this should still be in `main.rs`, and if you have reasons to not have it there, it should at the very least be in `protocol.rs`

e2e-tests/src/driver.rs
miraclx: it's alright as you have it, sandboxes will be small anyway, so it shouldn't matter much

e2e-tests/src/main.rs
miraclx: this is unresolved. drop `Protocol: fmt::Display` which returns a `String` for `Protocol::as_str()` which returns `&'static str`

e2e-tests/src/driver.rs
```diff
@@ -95,56 +103,54 @@ impl Driver {
...
+
+                initialized_protocols.insert(*protocol, sandbox_env);
+                break;
             }
         }
 
         for protocol_name in &self.environment.protocols {
-            let protocol_name = protocol_name.to_string().to_lowercase();
-            let protocol_path = protocols_dir.join(&protocol_name);
+            let protocol_path = protocols_dir.join(&protocol_name.to_string().to_lowercase());
```
miraclx: then this can become:

```suggestion
            let protocol_path = protocols_dir.join(protocol_name.as_str());
```

e2e-tests/src/driver.rs
```diff
@@ -170,6 +176,10 @@ impl Driver {
 
             let mut applications = read_dir(&protocol_path).await?;
             while let Some(app) = applications.next_entry().await? {
+                let protocol_name = protocol_name.to_string().to_lowercase();
```
miraclx: you shouldn't need to make a string here, patch `TestContext` to take `Protocol` instead of `&str`

e2e-tests/src/driver.rs
miraclx: this is unresolved

e2e-tests/src/driver.rs
```diff
@@ -515,54 +511,46 @@ impl TestRunReport {
 
     fn to_markdown(&self) -> EyreResult<String> {
         let mut markdown = String::new();
-
         writeln!(&mut markdown, "## E2E tests report")?;
 
-        for (scenario, protocols) in &self.scenario_matrix {
-            writeln!(&mut markdown, "### Scenario: {scenario}")?;
+        for (protocol, applications) in &self.scenario_matrix {
+            writeln!(&mut markdown, "## Protocol: {protocol}")?;
```
miraclx:
```suggestion
            writeln!(&mut markdown, "### Protocol: {protocol}")?;
```

e2e-tests/src/driver.rs
```diff
@@ -95,133 +103,121 @@ impl Driver {
...
+                    self.environment.output_writer,
+                    &protocol_name,
+                    None,
+                    sandbox,
+                );
+                let test_file_path = app.path();
+                // let app_name = test_file_path
+                //     .file_stem()
+                //     .and_then(|s| s.to_str())
+                //     .unwrap_or_default();
```
miraclx: ^?

e2e-tests/src/driver.rs
miraclx: but why are we changing the inviter / invitee sets on every protocol, but not on every application run?

and why at all, why not keep it the same throughout the run? imo there's no reason to change it

e2e-tests/src/driver.rs
```diff
@@ -95,133 +103,121 @@ impl Driver {
...
 
-        // Run scenarios directory by directory
-        let scenarios_dir = self.environment.input_dir.join("scenarios");
-        let mut entries = read_dir(scenarios_dir).await?;
+            let mut applications = read_dir(&protocol_path).await?;
+            while let Some(app) = applications.next_entry().await? {
+                let protocol_name = protocol_name.to_string().to_lowercase();
+                if app.file_type().await?.is_dir() {
+                    continue;
+                }
```
miraclx: why not [`is_file`](https://doc.rust-lang.org/nightly/std/fs/struct.FileType.html#method.is_file)?

e2e-tests/src/driver.rs
```diff
@@ -95,133 +103,121 @@ impl Driver {
...
+                let Some(app_name) = test_file_path.file_stem().and_then(|s| s.to_str()) else {
+                    bail!("No application name found");
+                };
 
-        while let Some(entry) = entries.next_entry().await? {
-            let path = entry.path();
-            if path.is_dir() {
-                let test_file_path = path.join("test.json");
                 if test_file_path.exists() {
                     let test_content = read(&test_file_path).await?;
```
miraclx: `is_file`?

also, write it as

```rust
if !test_file_path.is_file() {
    continue;
}

let test_content = __;
```

e2e-tests/src/driver.rs
```diff
@@ -349,7 +337,7 @@ impl Driver {
         drop(
             report
                 .scenario_matrix
-                .entry(ctx.protocol_name.to_owned())
+                .entry((*ctx.protocol_name.as_str()).to_string())
```
miraclx: why the deref?

```suggestion
                .entry(ctx.protocol_name.as_str().to_string())
```

miraclx (commented):
the rest looks okay to me, I've pinged the other guys to pitch in about the comments in https://github.com/calimero-network/core/pull/1236#discussion_r2087487563 & https://github.com/calimero-network/core/pull/1236#discussion_r2103486645

cy4n1d3-p1x3l:
> > Though i'm curious to know about the solution to those comments.
> 
> In this form:
> 
>     1. Identify all protocols for all scenarios ahead of time
> 
>     2. Spin up `merod` with all protocols configured
> 
>     3. Pick inviter & invitees
> 
>     4. Run all scenarios using the preconfigured nodes

Ahh i see

PR #1237, opened 2025-04-27
feat(meroctl): Add support for remote node connections

### LINKED ISSUE #1068: `meroctl` should be able to connect remotely
## Is your feature request related to a problem? Please describe

`meroctl` only connects locally and uses the node's networking identity for authentication.

We need to be able to connect to a node without having access to the `config.toml` file.

More context: https://github.com/calimero-network/core/pull/405#discussion_r1658365880

> and for remote-control commands:
> 
> 1. Connect to the node running at `data/node1`, and list it's contexts
> 
> ```
> $ meroctl --home data/node1 context ls
> ```
> 
> 2. Connect to the node running at `http://localhost:2528`, and list it's contexts
> 
> ```
> $ meroctl --api http://localhost:2528 context ls
> ```
> 
> 3. Connect to the node aliased `mein-node`, and list it's contexts
> 
> ```
> $ meroctl connect local-node --path data/node1
> $ meroctl --node local-node context ls
> ```
> 
> ```
> $ meroctl connect remote-node --api https://my-remote-node:2528
> $ meroctl --node remote-node context ls
> ```

Nathy-bajo:
## Description

Attempts to resolve #1068

This PR implements remote connectivity for meroctl, allowing:

- Direct API connections via `--api` flag
- Pre-configured node aliases via `--node` flag
- Management of node aliases through new `connect` subcommand

## Test plan

N/A

## Documentation update

Yes


Nathy-bajo:
@miraclx It's incomplete but I need your review on this, please!

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum NodeConnection {
+    Local { path: Utf8PathBuf },
+    Remote { api: Url },
+}
+
+impl NodeConfig {
+    pub fn load() -> Result<Self, Box<dyn std::error::Error>> {
```
miraclx: 1. use `eyre::Result` over `Box<dyn Error>`

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,60 @@
...
+        Ok(())
+    }
+
+    fn config_path() -> Result<Utf8PathBuf, Box<dyn std::error::Error>> {
+        let mut path = dirs::config_dir().ok_or_else(|| {
+            std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                "Could not find config directory",
+            )
+        })?;
```
miraclx: `.ok_or_eyre("")?` should suffice

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,60 @@
...
+        path.push("meroctl");
+        path.push("nodes.toml");
+        Ok(Utf8PathBuf::from_path_buf(path).map_err(|_| {
+            std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "Failed to convert path to UTF-8",
+            )
+        })?)
```
miraclx: `.wrap_err`

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    fn config_path() -> Result<Utf8PathBuf, Box<dyn std::error::Error>> {
+        let mut path = dirs::config_dir().ok_or_else(|| {
+            std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                "Could not find config directory",
+            )
+        })?;
+        path.push("meroctl");
+        path.push("nodes.toml");
```
miraclx: `push` takes an `impl AsRef<Utf8Path>`, like join so these can be merged

```suggestion
        path.push("meroctl/nodes.toml");
```

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NodeConfig {
+    pub aliases: BTreeMap<String, NodeConnection>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub enum NodeConnection {
+    Local { path: Utf8PathBuf },
+    Remote { api: Url },
+}
```
miraclx: Can you show the toml file generated by this?

I suspect it's something like:

```toml
[aliases.node1.Local]
path = "<path>"
```

Ideally, we go for something like:

```toml
[nodes]
node1 = { path = "/path/to/node1" }
node2 = { url = "https://my-remote-node:2528" }
```

or, where a node has any other associated configuration overrides (out of scope for this PR), then it'll be something like:

```toml
[nodes.node1]
path = "/path/to/node1"
output_format = "json"

[nodes.node2]
url = "https://my-remote-node:2528"
api_key = ".."
```

crates/meroctl/src/main.rs
```diff
@@ -7,6 +7,7 @@ use crate::cli::RootCommand;
 mod cli;
 mod common;
 mod defaults;
+pub mod node_config;
```
miraclx: just `config` should suffice

crates/meroctl/src/cli.rs
```diff
@@ -56,6 +61,33 @@ pub enum SubCommands {
...
+}
+
+#[derive(Debug, Parser)]
+pub enum ConnectCommand {
+    #[command(name = "local")]
+    Local(LocalNodeCommand),
+
+    #[command(name = "remote")]
+    Remote(RemoteNodeCommand),
+}
```
miraclx: So, these.. in the issue description are the names of the node, not the type of the connection.

We can infer the type of the connection with the connection method (path / URL)

it should be:

```console
meroctl node add/connect <node-name> --url http://__
    connect to node at URL
meroctl node add/connect <node-name> --path ./path/to/node
    connect to node at PATH
meroctl node rm/disconnect <node-name>
    disconnect node named NAME
meroctl node ls/list
    show all nodes this meroctl instance already knows about
```

miraclx (commented):
had these drafts, forgot to submit

crates/meroctl/src/cli.rs
```diff
@@ -67,16 +67,42 @@ pub enum SubCommands {
...
+    #[command(name = "add")]
+    Add(AddNodeCommand),
+
+    /// Remove a node connection
+    #[command(name = "rm")]
+    Remove(RemoveNodeCommand),
+
+    /// List all configured nodes
+    #[command(name = "ls")]
+    List,
```
miraclx: move to it's own module, like the rest

miraclx: when the name is the same as your override, the override isn't needed

```suggestion
pub enum NodeCommand {
    /// Add or connect to a node
    #[command(alias = "connect")]
    Add(AddNodeCommand),

    /// Remove a node connection
    #[command(aliases = ["rm", "disconnect"])]
    Remove(RemoveNodeCommand),

    /// List all configured nodes
    #[command(alias = "ls")]
    List,
```

crates/meroctl/src/cli.rs
```diff
@@ -271,23 +297,42 @@ pub enum ConnectionInfo {
...
+                println!("{table}");
+                return Ok(());
             }
-        });
+        }
 
-        config
-            .save()
-            .map_err(|e| eyre!("Failed to save node config: {}", e))?;
+        config.save()?;
```
miraclx: we shouldn't save on `list`

crates/meroctl/src/node_config.rs
```diff
@@ -1,22 +1,25 @@
...
 use camino::Utf8PathBuf;
+use eyre::{eyre, Result};
 use serde::{Deserialize, Serialize};
 use url::Url;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct NodeConfig {
-    pub aliases: BTreeMap<String, NodeConnection>,
+    #[serde(rename = "nodes")]
+    pub nodes: BTreeMap<String, NodeConnection>,
```
miraclx: this attribute is not needed (the override name is the same as the field name)

crates/meroctl/src/main.rs
miraclx: this isn't yet resolved

crates/meroctl/src/cli.rs
```diff
@@ -61,6 +66,59 @@ pub enum SubCommands {
...
+    #[command(name = "ls")]
+    List,
+}
+
+#[derive(Debug, Parser)]
+pub struct LocalNodeCommand {
+    pub alias: String,
+
+    #[arg(long)]
+    pub path: Utf8PathBuf,
```
miraclx: missing docs

crates/meroctl/src/cli.rs
```diff
@@ -72,17 +130,33 @@ pub struct RootArgs {
 
     /// Name of node
     #[arg(short, long, value_name = "NAME")]
-    pub node_name: String,
+    pub node_name: Option<String>,
+
+    /// API endpoint URL
+    #[arg(long, value_name = "URL", conflicts_with = "node_name")]
+    pub api: Option<String>,
```
```suggestion
    pub api: Option<Url>,
```

crates/meroctl/src/cli.rs
```diff
@@ -72,17 +130,33 @@ pub struct RootArgs {
 
     /// Name of node
     #[arg(short, long, value_name = "NAME")]
-    pub node_name: String,
+    pub node_name: Option<String>,
```
miraclx: so imo, `--node-name & --home` should be folded into `--node`

when we receive `meroctl --node X`

first, we check if we have any nodes named X in the config, if not, we check the default home dir, and that's it

if a node exists in some other directory that is not the default home directory, you HAVE to use `meroctl node add <node> --path <path>`

miraclx: this will require some documentation update, cc: @frdomovic, @antonpaisov

crates/meroctl/src/cli/app/get.rs
```diff
@@ -31,21 +31,24 @@ impl Report for GetApplicationResponse {
...
+            Some(ConnectionInfo::Remote { api }) => {
+                let mut url = api.clone();
+                url.set_path(&format!("admin-api/dev/applications/{}", self.app_id));
+                (url, None)
+            }
+            None => return Err(eyre!("No connection configured")),
+        };
+
+        let response: GetApplicationResponse =
+            do_request(&Client::new(), url, None::<()>, keypair, RequestType::Get).await?;
```
miraclx: when the connection is `Remote` we still need the keypair, so for now, we have to keep the keypair inside the nodes config file.

which means on `meroctl nodes add` we can add a `--auth <auth>` backed by env var `MEROCTL_NODE_KEY`

crates/meroctl/src/cli.rs
```diff
@@ -173,3 +286,53 @@ where
...
+
+pub enum ConnectionInfo {
+    Local {
+        config: ConfigFile,
+        multiaddr: Multiaddr,
+    },
+    Remote {
+        api: Url,
+    },
+}
```
miraclx: tbh, I don't expect an enum for connection info, it should just be URL & auth key

Nathy-bajo: Alright, on it.

Nathy-bajo:
@miraclx All that's left for this is to implement the remote node connections on the `Commands,` which I am already doing. 

But I would appreciate your review on the remote node connection implementation.

crates/meroctl/src/cli.rs
```diff
@@ -1,18 +1,17 @@
...
 
 use bootstrap::BootstrapCommand;
-use calimero_config::ConfigFile;
 use camino::Utf8PathBuf;
 use clap::{Parser, Subcommand};
 use comfy_table::{Cell, Color, Table};
 use const_format::concatcp;
 use eyre::{eyre, Report as EyreReport};
-use libp2p::Multiaddr;
+use node::NodeCommand;
```
miraclx: move this to L31

crates/meroctl/src/cli.rs
```diff
@@ -61,6 +66,24 @@ pub enum SubCommands {
...
+    #[arg(long)]
+    pub path: Utf8PathBuf,
+}
+
+#[derive(Debug, Parser)]
+pub struct RemoteNodeCommand {
+    pub alias: String,
+
+    #[arg(long)]
+    pub api: Url,
```
miraclx: are these still used?

crates/meroctl/src/cli.rs
miraclx: this isn't yet resolved, `--node-node` shouldn't exist anymore

crates/meroctl/src/cli.rs
```diff
@@ -91,18 +130,69 @@ impl RootArgs {
...
+                            bs58::encode(config.identity.to_protobuf_encoding().unwrap())
+                                .into_string(),
+                        ),
+                    }
+                }
+            }
+            (None, Some(api_url)) => ConnectionInfo {
+                api_url: api_url.clone(),
+                auth_key: None,
+            },
```
miraclx: consider inheriting the env var that `meroctl nodes add` uses here as well

crates/meroctl/src/cli.rs
```diff
@@ -173,3 +264,8 @@ where
 {
     serializer.collect_str(&report)
 }
+
+pub struct ConnectionInfo {
+    pub api_url: Url,
+    pub auth_key: Option<String>,
```
miraclx: why not make this a `Keypair`?

```suggestion
    pub auth_key: Option<Keypair>,
```

crates/meroctl/src/cli.rs
```diff
@@ -91,18 +130,69 @@ impl RootArgs {
...
+                } else {
+                    // Fall back to checking default home directory
+                    let config = load_config(&defaults::default_node_dir(), node_name)?;
+                    let multiaddr = fetch_multiaddr(&config)?;
+                    let url = multiaddr_to_url(&multiaddr, "")?;
+                    ConnectionInfo {
+                        api_url: url,
+                        auth_key: Some(
+                            bs58::encode(config.identity.to_protobuf_encoding().unwrap())
+                                .into_string(),
```
miraclx: then you won't need this

crates/meroctl/src/cli/app/get.rs
```diff
@@ -31,24 +32,30 @@ impl Report for GetApplicationResponse {
...
-            &format!("admin-api/dev/applications/{}", self.app_id),
-        )?;
+        let mut url = connection.api_url.clone();
+        url.set_path(&format!("admin-api/dev/applications/{}", self.app_id));
+
+        let keypair = connection
+            .auth_key
+            .as_ref()
+            .and_then(|k| bs58::decode(k).into_vec().ok())
+            .and_then(|bytes| Keypair::from_protobuf_encoding(&bytes).ok());
```
miraclx: or this

crates/meroctl/src/cli/app/install.rs
```diff
@@ -59,15 +59,31 @@ impl InstallCommand {
...
+            .connection
+            .as_ref()
+            .ok_or_else(|| eyre!("No connection configured"))?;
+
+        let mut url = connection.api_url.clone();
+        url.set_path(if self.path.is_some() {
+            "admin-api/dev/install-dev-application"
+        } else {
+            "admin-api/dev/install-application"
+        });
```
miraclx: move `url.set_path` inside the if & else blocks, exposing the conditional

crates/meroctl/src/cli/node.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Path to local node
+    #[arg(long, conflicts_with = "url")]
+    pub path: Option<Utf8PathBuf>,
+
+    /// URL of remote node
+    #[arg(long, conflicts_with = "path")]
+    pub url: Option<Url>,
+
+    /// Authentication key for the node (can also be set via MEROCTL_NODE_KEY env var)
+    #[arg(long, env = "MEROCTL_NODE_KEY")]
```
miraclx: clap already says it

```suggestion
    /// Authentication key for the node
    #[arg(long, env = "MEROCTL_NODE_KEY")]
```

crates/meroctl/src/cli/node.rs
```diff
@@ -0,0 +1,89 @@
...
+            NodeCommand::Add(cmd) => {
+                let connection = match (cmd.path, cmd.url) {
+                    (Some(path), None) => NodeConnection::Local { path },
+                    (None, Some(url)) => NodeConnection::Remote {
+                        url,
+                        auth: cmd.auth,
+                    },
+                    _ => return Err(eyre!("Either --path or --url must be specified")),
+                };
+                drop(config.nodes.insert(cmd.name, connection));
```
miraclx: prefer `let _ignored = any()` over `drop(any())` and `let _ = any()` primarily due to inlay hints

crates/meroctl/src/main.rs
miraclx: this isn't yet resolved, it's not the node config, it's `meroctl`'s config

so the file name should be `config`

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,55 @@
+use std::collections::BTreeMap;
+
+use camino::Utf8PathBuf;
+use eyre::{eyre, Result};
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NodeConfig {
```
miraclx:
```suggestion
#[derive(Debug, Serialize, Deserialize)]
pub struct Config {
```

crates/meroctl/src/node_config.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+impl Default for NodeConfig {
+    fn default() -> Self {
+        Self {
+            nodes: BTreeMap::new(),
+        }
+    }
+}
```
miraclx: this is just a `derive(Default)`

crates/meroctl/src/cli.rs
```diff
@@ -91,18 +130,69 @@ impl RootArgs {
...
     }
 }
 
 impl RootCommand {
     pub async fn run(self) -> Result<(), CliError> {
         let output = Output::new(self.args.output_format);
-        let environment = Environment::new(self.args, output);
+
+        // Determine connection info
+        let connection = match (&self.args.node, &self.args.api) {
```
miraclx: conside breaking this out into a separate function, maybe an impl on `NodeConnection`

Nathy-bajo:
@miraclx Alright, thanks so much for the review!

Nathy-bajo:
@miraclx It probably still needs more nits but I poured my heart and soul into this.

crates/meroctl/src/cli/bootstrap/start.rs
```diff
@@ -313,11 +313,11 @@ impl StartBootstrapCommand {
         let (context_id, public_key) = create_context(
             environment,
             &client,
-            multiaddr,
+            connection,
             None,
             application_id,
             None,
-            &config.identity,
+            auth_key,
```
miraclx: since `auth_key` is now in connection, you shouldn't need to pass both of these, just use it from the connection

crates/meroctl/src/cli.rs
```diff
@@ -70,19 +78,29 @@ pub struct RootArgs {
...
-    /// Name of node
-    #[arg(short, long, value_name = "NAME")]
-    pub node_name: String,
+    /// API endpoint URL
+    #[arg(long, value_name = "URL", conflicts_with = "node_name")]
+    pub api: Option<Url>,
+
+    /// Use a pre-configured node alias
+    #[arg(long, value_name = "ALIAS", conflicts_with_all = &["node_name", "api"])]
+    pub node: Option<String>,
```
miraclx: there no longer is `node_name`:

```suggestion
    /// API endpoint URL
    #[arg(long, value_name = "URL")]
    pub api: Option<Url>,

    /// Use a pre-configured node alias
    #[arg(long, value_name = "ALIAS", conflicts_with = "api")]
    pub node: Option<String>,
```

crates/meroctl/src/cli/app/list.rs
```diff
@@ -50,13 +50,19 @@ impl Report for ListApplicationsResponse {
...
+        let mut url = connection.api_url.clone();
+        url.set_path("admin-api/dev/applications");
 
         let response: ListApplicationsResponse = do_request(
             &Client::new(),
-            multiaddr_to_url(fetch_multiaddr(&config)?, "admin-api/dev/applications")?,
+            url,
             None::<()>,
-            &config.identity,
+            connection.auth_key.as_ref(),
```
miraclx, line 65: not to hold back this PR any further so it certainly doesn't have to happen in this PR

but with the new design, it's very evident that there's still lots to improve upon with respect to the complexity of making requests

Now that we have a struct for connection info, we might as well keep the `Client` inside it

so at the end of the day, it becomes something like this:

```rust
environment
    .connection
    .get("admin-api/dev/applications", json!({}))
    .await?;
```

which significantly reduces the complexity and boilerplate needed to do really simple things

miraclx (approved):
lgtm, signing off.. there's some more work to do, but this is fine for the first iteration

crates/meroctl/src/cli/app/list.rs
Nathy-bajo: Yeah, that makes more sense, but maybe we should get this merged first. If I see an issue opened to refactor the code on this section I'll open another PR for that.

miraclx: Sounds good! I'll file one. #1275

miraclx:
Aah, forgot about e2e-tests, can you update here:

https://github.com/calimero-network/core/blob/5830b1204fb555fd90e21106fc005a0358194acc/e2e-tests/src/meroctl.rs#L241-L242

It should just work since `--home` is specified, or you could update the flow to first `meroctl node add` in an earlier step so this PR is also covered in those tests

Nathy-bajo:
@antonpaisov @miraclx @fbozic Ready to be merged.

e2e-tests/src/meroctl.rs
miraclx, line 1: can you add a diagnostic log line to

https://github.com/calimero-network/core/blob/bf58a82958d8cb2dedd2cd699e77589da61e5d72/e2e-tests/src/meroctl.rs#L255-L261

so we see what is received from `meroctl` that fails to parse?

Nathy-bajo: Done

e2e-tests/src/meroctl.rs
```diff
@@ -249,16 +250,43 @@ impl Meroctl {
...
 
         async move {
             let output = command
                 .stdout(Stdio::piped())
+                .stderr(Stdio::piped())
                 .spawn()?
                 .wait_with_output()
                 .await?;
 
-            Ok(serde_json::from_slice(&output.stdout)?)
```
miraclx: can you revert this, I just expected you to print the string representation of `output.stdout`

since it was failing on a deserialization error

as for `stderr`, we don't care about capturing it, by default it'll inherit from the parent which is acceptable

e2e-tests/src/meroctl.rs
```diff
@@ -262,6 +259,10 @@ impl Meroctl {
                 .wait_with_output()
                 .await?;
 
+            if output.stdout.is_empty() {
+                bail!("Command returned empty output");
+            }
```
miraclx:
```suggestion
            let stdout = String::from_utf8_lossy(&output.stdout);
            dbg!(stdout);
```

e2e-tests/src/meroctl.rs
```diff
@@ -228,18 +224,19 @@ impl Meroctl {
 
     fn run_cmd<'a>(
         &'a self,
-        node_name: &'a str,
+        node: &'a str,
         args: impl IntoIterator<Item = &'a str>,
     ) -> impl Future<Output = EyreResult<serde_json::Value>> + 'static {
         let mut command = Command::new(&self.binary);
+        let output_writer = self.output_writer.clone();
```
miraclx: revert

e2e-tests/src/meroctl.rs
```diff
@@ -253,7 +250,7 @@ impl Meroctl {
 
         command_line.push('\'');
 
-        self.output_writer.write_str(&command_line);
+        output_writer.write_str(&command_line);
```
miraclx: revert

crates/meroctl/src/common.rs
```diff
@@ -59,33 +60,53 @@ where
...
+                    });
+                }
+                return Ok(response.json::<O>().await?);
+            }
+            Err(e) => {
+                last_error = Some(e);
+                if attempt < MAX_RETRIES - 1 {
+                    tokio::time::sleep(RETRY_DELAY).await;
+                }
+            }
```
miraclx: not all errors are retryable though, I fear this cannot be introduced without special accounting for what is and isn't, which I'd consider out of scope for this PR

crates/meroctl/src/cli.rs
```diff
@@ -132,7 +134,20 @@ impl RootCommand {
...
                 if let Some(conn) = node_config.nodes.get(node_name) {
-                    conn.get_connection_info(Some(node_name)).await?
+                    match conn.get_connection_info(Some(node_name)).await {
+                        Ok(info) => {
+                            // Verify node is reachable
+                            if let Err(e) = Self::check_node_ready(&info).await {
+                                output.write(&ErrorLine(&format!("Node not ready: {}", e)));
+                                return Err(e);
+                            }
+                            info
```
miraclx: is this for diagnostic? otherwise shouldn't it apply to saved connections and `--api-url`?

Nathy-bajo: I was testing something out. Reverted.

crates/meroctl/src/config.rs
```diff
@@ -0,0 +1,88 @@
...
+#[serde(untagged)]
+pub enum NodeConnection {
+    Local { path: Utf8PathBuf },
+    Remote { url: Url, auth: Option<String> },
+}
+
+impl NodeConnection {
+    pub async fn get_connection_info(
+        &self,
+        node_name: Option<&str>,
```
miraclx: why is this an `Option`? it's only use is `Some` so this isn't necessary

crates/meroctl/src/cli.rs
```diff
@@ -91,18 +109,52 @@ impl RootArgs {
...
+                } else {
+                    // Fall back to checking default home directory
+                    let config = load_config(&defaults::default_node_dir(), node_name).await?;
+                    let multiaddr = fetch_multiaddr(&config)?;
+                    let url = multiaddr_to_url(&multiaddr, "")?;
+                    ConnectionInfo {
+                        api_url: url,
+                        auth_key: Some(config.identity),
+                    }
+                }
```
miraclx: how about you host this logic in `Config::get_connection()`

crates/meroctl/src/cli.rs
```diff
@@ -91,18 +109,52 @@ impl RootArgs {
...
+        // Determine connection info
+        let connection = match (&self.args.node, &self.args.api) {
+            (Some(node_name), None) => {
+                // Check if node exists in config
+                let node_config = Config::load()?;
+                if let Some(conn) = node_config.nodes.get(node_name) {
+                    conn.get_connection_info(Some(node_name)).await?
+                } else {
+                    // Fall back to checking default home directory
+                    let config = load_config(&defaults::default_node_dir(), node_name).await?;
```
miraclx: this shouldn't default to the "default node dir" but the value to `--home`

crates/meroctl/src/cli/context/invite.rs
```diff
@@ -74,31 +71,45 @@ impl InviteCommand {
...
+        let connection = environment
+            .connection
+            .as_ref()
+            .ok_or_eyre("No connection configured")?;
 
-        let multiaddr = fetch_multiaddr(&config)?;
+        let auth_key = connection
+            .auth_key
+            .as_ref()
+            .ok_or_eyre("No authentication key configured")?;
```
miraclx: an authentication key is not "required", can you change all cases where we expect it to be present?

Nathy-bajo: I see you marked the rest of the review as resolved. Is this the only change requested?

crates/meroctl/src/cli/context/alias.rs
```diff
@@ -71,8 +66,13 @@ impl ContextAliasCommand {
                     return Ok(());
                 }
 
-                let lookup_result =
-                    lookup_alias(&connection.api_url, auth_key, alias, None).await?;
+                let lookup_result = lookup_alias(
+                    &connection.api_url,
+                    connection.auth_key.as_ref().unwrap(),
```
miraclx: this is worse than what you had before, it's still expecting the auth key to be present only now it's panicking instead of returning an error

it should be the other way around.. `lookup_alias` should expect an `Option`

miraclx (commented):
please remove all `.unwrap`

crates/meroctl/src/cli/context/invite.rs
miraclx: Yep, looks to me like this is the last thing.

crates/meroctl/src/cli/context/alias.rs
Nathy-bajo: Done!
397401260-cfec0161-ecfd-4143-a92e-ca262aeb962d

Nathy-bajo:
> This pull request has been automatically marked as stale. **If this pull request is still relevant, please leave any comment** (for example, "bump"), and we'll keep it open. We are sorry that we haven't been able to prioritize reviewing it yet. Your contribution is very much appreciated.

bump

Nathy-bajo:
@miraclx I might need some help here, please. I've been trying to fix but It keeps leading me to another error.

Nathy-bajo:
@miraclx You're a lifesaver🙏❤️

crates/meroctl/src/cli/context/create.rs
```diff
@@ -330,16 +329,14 @@ async fn watch_app_and_update_context(
...
     context_id: ContextId,
     application_id: ApplicationId,
-    keypair: &Keypair,
+    keypair: Option<&Keypair>,
     member_public_key: PublicKey,
 ) -> EyreResult<()> {
-    let url = multiaddr_to_url(
-        base_multiaddr,
-        &format!("admin-api/dev/contexts/{context_id}/application"),
-    )?;
```
miraclx, line 342: the URL changed here

crates/meroctl/src/cli/context/invite.rs
```diff
@@ -110,15 +121,21 @@ impl InviteCommand {
...
+                    Some(context_id),
+                    self.invitee_id,
+                )
+                .await?;
+                environment.output.write(&res);
+            } else {
+                environment.output.write(&InfoLine(
+                    "Skipping alias creation - no authentication key provided",
+                ));
+            }
```
miraclx: this is wrong, we shouldn't care if the auth key is present or not

crates/meroctl/src/cli/proxy/get.rs
```diff
@@ -231,19 +236,18 @@ impl GetCommand {
...
-            multiaddr,
-            &format!(
-                "admin-api/dev/contexts/{}/proposals/{}/approvals/users",
-                context_id, proposal_id
-            ),
-        )?;
+        let mut url = connection.api_url.clone();
+        url.set_path(&format!(
+            "admin-api/dev/contexts/{}/proposals/{}/approvals/count",
+            context_id, proposal_id
```
miraclx: wrong url

crates/meroctl/src/cli/proxy/get.rs
```diff
@@ -284,19 +287,18 @@ impl GetCommand {
...
-            multiaddr,
-            &format!(
-                "admin-api/dev/contexts/{}/proposals/{}",
-                context_id, proposal_id
-            ),
-        )?;
+        let mut url = connection.api_url.clone();
+        url.set_path(&format!(
+            "admin-api/dev/contexts/{}/proposals/{}/approvals/count",
+            context_id, proposal_id
```
miraclx: wrong url

miraclx (commented):
some URLs changed in transition

crates/meroctl/src/cli/context/create.rs
Nathy-bajo: I see ✍️

PR #1238, opened 2025-04-28
feat(context): actor handlers for join/delete

fbozic:


crates/context/src/handlers/join_context.rs
```diff
@@ -0,0 +1,410 @@
...
+    node_client.has_blob(blob_id)
+}
+
+async fn subscribe(network_client: &NodeClient, context_id: &ContextId) -> eyre::Result<()> {
+    network_client.subscribe(context_id).await?;
+
+    info!(%context_id, "Subscribed to context");
+
+    Ok(())
+}
```
miraclx: We shouldn't need the method here, it's already accounted for

https://github.com/calimero-network/core/blob/fd6c5d2ae98659eedb40e342514aa889c62f26f3/crates/node/primitives/src/client.rs#L24-42

crates/context/src/handlers/join_context.rs
```diff
@@ -0,0 +1,410 @@
...
+    );
+
+    let application_id = ApplicationId::from(*Hash::hash_borsh(&application)?);
+
+    let mut handle = datastore.handle();
+
+    handle.put(&key::ApplicationMeta::new(application_id), &application)?;
+
+    Ok(application_id)
+}
```
miraclx: Same here

https://github.com/calimero-network/core/blob/fd6c5d2ae98659eedb40e342514aa889c62f26f3/crates/node/primitives/src/client/application.rs#L72-95

miraclx (commented):
Most of the utility methods tangential to the handler don't need functions, everything in there is already covered by NodeClient and ContextClient.

crates/context/src/handlers/join_context.rs
```diff
@@ -0,0 +1,348 @@
...
+    let (blob_id, size) = node_client
+        .add_blob(stream, expected_size, expected_hash.as_ref())
+        .await?;
+
+    if matches!(expected_size, Some(expected_size) if size != expected_size) {
+        eyre::bail!("fatal: blob size mismatch");
+    }
+
+    Ok((blob_id, size))
+}
```
miraclx: let's remove this piece, just call `node_client.add_blob` directly without this method

crates/context/src/handlers/join_context.rs
```diff
@@ -0,0 +1,348 @@
...
+    let handle = datastore.handle();
+
+    if let Some(application) = handle.get(&key::ApplicationMeta::new(*application_id))? {
+        if has_blob_available(node_client, &application.blob.blob_id())? {
+            return Ok(true);
+        }
+    }
+
+    Ok(false)
+}
```
miraclx: remove this function, use `node_client.has_application` instead

crates/context/src/handlers/join_context.rs
```diff
@@ -0,0 +1,348 @@
...
+
+pub fn has_blob_available(node_client: &NodeClient, blob_id: &BlobId) -> eyre::Result<bool> {
+    node_client.has_blob(blob_id)
+}
```
miraclx: there's no need for this function

miraclx (commented):
rest is okay

miraclx:
lgtm, thanks guys! 🎉

PR #1241, opened 2025-04-29
fix(cli): crashing identity listing

frdomovic:
# fix(cli+meroctl): crashing identity listing

## Description

When default context alias is not set it throwing error and not exactly saying what the issue is.  Instead of throwing error its now printing the message saying to setup default alias.

cc @miraclx @fbozic Also some other commands are also throwing errors instead of messages so let me know if they should also be updated. Related to image below.
Screenshot 2025-04-29 at 14 02 29

## Test plan

Screenshot 2025-04-29 at 13 44 53
Screenshot 2025-04-29 at 13 51 10

Screenshot 2025-05-05 at 16 29 40

## Documentation update
Identity part will be updated [here](https://calimero-network.github.io/tutorials/invitations-and-joinings) and docs [PR](https://github.com/calimero-network/calimero-network.github.io/pull/124)


miraclx:
I.. totally agree, but the node doesn't crash, just prints that error:

https://github.com/calimero-network/core/blob/a73f2ecd95fc2a510a885d891a3ed59090d6b1dc/crates/node/src/lib.rs#L183-L185

It only looks like a crash because we're printing the `Debug` representation of `eyre::Report`

frdomovic:
Updated meroctl also cc @miraclx 

Screenshot 2025-05-05 at 16 29 40

crates/meroctl/src/cli/context/identity.rs
```diff
@@ -111,10 +111,17 @@ async fn list_identities(
...
-        .ok_or_eyre("Failed to resolve context: no value found")?;
+    let context_id = match resolve_response.value().cloned() {
+        Some(id) => id,
+        None => {
+            let context_display = context
+                .as_ref()
+                .map(|alias| alias.to_string())
+                .unwrap_or_else(|| "default".to_string());
+            println!("Error: Unable to resolve context '{}'. Please verify the context ID exists or setup default context.", context_display);
+            return Ok(());
```
miraclx: for meroctl, it should be a hard error

i.e after the failure `echo $?` should be non-zero

so, let's `eyre::bail` instead

frdomovic: Updated 
Screenshot 2025-05-06 at 11 21 23

chefsale (approved):

PR #1246, opened 2025-05-06
feat: Actor model server methods implementation

alenmestrov:
## Description

Implemented ContextClient and NodeClient in server crate

## Test plan

N/A

## Documentation update

N/A

### DIFF
--- a/crates/server/src/admin/handlers/proposals.rs
+++ b/crates/server/src/admin/handlers/proposals.rs
@@ -35,7 +35,7 @@ pub async fn get_proposals_handler(
     Json(req): Json<GetProposalsRequest>,
 ) -> impl IntoResponse {
     let Some(external_config) = state.ctx_client.context_config(&context_id)? else {
-        return Err(parse_api_error(eyre::eyre!("Context not found"))).into_response();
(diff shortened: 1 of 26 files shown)


crates/server/src/admin/handlers/proposals.rs
```diff
@@ -34,9 +34,17 @@ pub async fn get_proposals_handler(
     Extension(state): Extension<Arc<AdminState>>,
     Json(req): Json<GetProposalsRequest>,
 ) -> impl IntoResponse {
-    match state
-        .ctx_manager
-        .get_proposals(context_id, req.offset, req.limit)
+    let Some(external_config) = state.ctx_client.context_config(&context_id)? else {
+        return Err(parse_api_error(eyre::eyre!("Context not found"))).into_response();
+    };
```
miraclx: without the `Err` should work just fine

```suggestion
    let Some(external_config) = state.ctx_client.context_config(&context_id)? else {
        return parse_api_error(eyre::eyre!("Context not found")).into_response();
    };
```

and the other cases

miraclx (approved):

Issue #234, opened 2024-04-29
Avoid all forms of blocking IO


miraclx:
Identify and replace all dependence on blocking IO, and migrate to the non-blocking alternatives provided by tokio:

For example: [`tokio::fs`](https://docs.rs/tokio/latest/tokio/fs/index.html) over `std::fs`

- [ ] implementation - 1h

 risk factor - 1h

Brijeshthummar02:
@miraclx  can you explain more about issue?

Nathy-bajo:
> [@miraclx](https://github.com/miraclx) can you explain more about issue?

Replace `[std::fs](use std::fs::{read_to_string, write};)` with `tokio::fs` or any other dependency on blocking IO

`tokio::fs` is better than `std::fs` because it performs non-blocking I/O, allowing concurrent task execution and efficient resource utilization without blocking threads.

dotandev:
@antonpaisov 

Can you assign me to this.

I want to take this up.

PR #1249, opened 2025-05-07
fix(tokio): Changed std::fs to  tokio:fs

dotandev:
## Description

Attempts to resolve https://github.com/calimero-network/core/issues/234

This PR implements the conversion of std:: to the asynchronous alternative tokio::

Change of std:: to tokio:: in the necessary positions.

## Test plan

N/A

## Documentation update

N/A


crates/config/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 use core::time::Duration;
-use std::fs::{read_to_string, write};
+use tokio::fs::{read_to_string, write};
```
miraclx: `cargo +nightly fmt`

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.81.0"
+channel = "stable"
```
miraclx: We like to keep this pinned for consistency, unless your PR uses features on stable that aren't available in 1.81

crates/config/Cargo.toml
```diff
@@ -16,6 +16,7 @@ libp2p-identity = { workspace = true, features = ["peerid", "serde"] }
 multiaddr.workspace = true
 serde = { workspace = true, features = ["derive"] }
 toml.workspace = true
+tokio = { workspace = true, features = ["full"] }
```
miraclx: ~nit: order

crates/runtime/examples/rps.rs
```diff
@@ -2,7 +2,6 @@
 #![allow(dead_code)]
 
 use std::env;
-use tokio::fs::File;
 use std::io::Read;
```
miraclx: should `io::Read` still be imported?

crates/config/Cargo.toml
miraclx: and also, we don't want the `"full"` feature flag, focus on what specifically you need, in this case `"fs"`

crates/runtime/Cargo.toml
```diff
@@ -14,6 +14,7 @@ owo-colors = { workspace = true, optional = true }
 rand.workspace = true
 serde = { workspace = true, features = ["derive"] }
 thiserror.workspace = true
+tokio = { workspace = true, features = ["fs", "io"] }
```
miraclx: this should be in `dev-dependencies`, optional by default and the `[[example]]` section should then have `required-features = ["tokio"]`

crates/runtime/examples/demo.rs
miraclx, line 1: on second thought, we don't need tokio at all for these examples, they're totally separate units from the node

essentially just tests, so the accommodations don't apply here

```console
git checkout master crates/runtime
```

crates/store/tests/rocksdb.rs
miraclx, line 1: same here, these are tests, file

```console
git checkout master crates/store
```

crates/merod/src/cli/init.rs
```diff
@@ -162,19 +162,23 @@ impl InitCommand {
...
-                } else {
-                    bail!("Failed to load existing configuration: {}", err);
+            match ConfigFile::load(&path).await {
+                Ok(_) => {
+                    if !self.force {
+                        warn!("Node is already initialized in {:?}", path);
+                        return Ok(());
+                    }
+                }
+                Err(err) => {
```
miraclx: `if let Err(err) = ..` is preferable here

crates/merod/src/cli/init.rs
```diff
@@ -171,11 +171,10 @@ impl InitCommand {
                 } else {
                     bail!("Failed to load existing configuration: {}", err);
                 }
-            }
-            if !self.force {
+            } else if !self.force {
                 warn!("Node is already initialized in {:?}", path);
                 return Ok(());
-            }
+            }
```
miraclx: `cargo +nightly fmt`

miraclx (approved):
lgtm, thanks!

alenmestrov (approved):

PR #1251, opened 2025-05-08
fix:  cleanup

alenmestrov:
## Description

Fixed various errors and warnings introduced with server and context changes.

## Test plan

N/A

## Documentation update

N/A

### DIFF
--- a/crates/context/Cargo.toml
+++ b/crates/context/Cargo.toml
@@ -8,21 +8,16 @@ license.workspace = true
 
 [dependencies]
 actix.workspace = true
-camino = { workspace = true, features = ["serde1"] }
-either.workspace = true
 eyre.workspace = true
-futures-util.workspace = true
 ouroboros.workspace = true
 rand.workspace = true
 reqwest = { workspace = true, features = ["stream"] }
 serde.workspace = true
 tokio = { workspace = true, features = ["sync", "macros"] }
-tokio-util.workspace = true
 tracing.workspace = true
 
 calimero-context-config = { workspace = true, features = ["client"] }
 calimero-context-primitives.workspace = true
-calimero-blobstore.workspace = true
 calimero-primitives = { workspace = true, features = ["borsh", "rand"] }
 calimero-runtime.workspace = true
 calimero-node-primitives.workspace = true
--- a/crates/context/primitives/src/client/external/proxy.rs
+++ b/crates/context/primitives/src/client/external/proxy.rs
@@ -1,6 +1,6 @@
 use calimero_context_config::client::env::proxy::ContextProxy;
 use calimero_context_config::repr::{ReprBytes, ReprTransmute};
-use calimero_context_config::types::ProposalId;
+use calimero_context_config::types::{ContextStorageEntry, ProposalId};
 use calimero_context_config::{Proposal, ProposalAction};
 use calimero_primitives::identity::PublicKey;
 use eyre::OptionExt;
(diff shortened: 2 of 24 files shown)


crates/server/src/admin/handlers/context/get_context_identities.rs
```diff
@@ -15,39 +17,55 @@ pub async fn handler(
...
         }
+        Err(err) => return parse_api_error(err).into_response(),
     };
 
-    let owned = req.uri().path().ends_with("identities-owned");
+    // Clone what we need
+    let id = context.id;
+    let ctx_client = state.ctx_client.clone();
+
+    // Process the stream in a blocking task to handle non-Send types
```
frdomovic: Looks okay to me, only small thing I noticed a lot of code comment and not sure if we need all of this commented cc @miraclx

crates/context/Cargo.toml
miraclx, line 1: tysm!

crates/context/primitives/src/client/external/proxy.rs
```diff
@@ -163,4 +163,20 @@ impl ExternalProxyClient<'_, '_> {
 
         Ok(approvals.num_approvals)
     }
+
+    pub async fn get_context_storage_entries(
```
miraclx: we can only access this through the `ContextClient`, so we don't need context in the name anymore 

Let's try `get_external_storage_entries`

crates/node/src/interactive_cli.rs
```diff
@@ -36,7 +36,7 @@ pub enum SubCommand {
     // Store(store::StoreCommand),
     State(state::StateCommand),
 }
-
+//instead of node take node client and context client
```
miraclx: 🙂 *<sub>leaving a pin here for visibility, so we don't forget</sub>*

crates/server/src/admin/handlers/context/join_context.rs
```diff
@@ -22,7 +22,7 @@ pub async fn handler(
 
     match result {
         Ok(result) => ApiResponse {
-            payload: JoinContextResponse::new(result),
+            payload: JoinContextResponse::new(Some((result.context_id, result.member_public_key))),
```
miraclx, line 25: 🙂 *<sub>leaving a note for myself during the integration</sub>*

crates/server/src/admin/handlers/proposals.rs
```diff
@@ -226,25 +267,43 @@ pub async fn get_proposal_approvers_handler(
...
+        Ok(proposal_approvers) => {
+            // Convert PublicKey to ContextIdentity before creating Repr
+            let context_identities = proposal_approvers
+                .into_iter()
+                .map(|pk| {
+                    // First get the bytes from the PublicKey, then create a ContextIdentity
+                    let id: ContextIdentity = pk.rt().expect("infallible conversion");
+                    Repr::new(id)
+                })
+                .collect();
```
miraclx: so same as before, with only a slight change

```rust
Ok(proposal_approvers) => ApiResponse {
    payload: GetProposalApproversResponse {
        data: proposal_approvers.into_iter().map(|pk| pk.rt()).collect::<Result<_, _>>()?,
    },
```

crates/server/src/admin/handlers/context/get_context_identities.rs
miraclx: for now it's okay while we work out how to remove the need for the special accommodations.

maybe use `fixme! __` comments instead? @alenmestrov

crates/context/Cargo.toml
alenmestrov: no problem :)

miraclx (approved):

Issue #1073, opened 2025-01-29
Remove node-ui, pull from repo instead


miraclx:
Remove the `node-ui`, to avoid maintaining its code across multiple fronts.

Since `merod` needs `node-ui` to be present when building, we can (as part of a build script), download the latest version of the `node-ui` repo (if the hash is different), and embed that into the node.

We shouldn't need to download the Node UI all of the time so long as it's cached.

iamgoeldhruv:
@chefsale ,can you please assign this issue to me?

chefsale:
Assigned.

PR #1252, opened 2025-05-08
feat: extract node ui code

frdomovic:
# feat: extract node ui code

## Description

As part of this [issue](https://github.com/calimero-network/core/issues/1073) removing all files and logic (workflows, scripts) connected to node-ui. Node-ui will instead be downloaded from admin-dashboard.

## Test plan
TBD

## Documentation update
TBD


scripts/download-node-ui.sh
miraclx, line 1: I'm guessing this is an earlier iteration? Can we update to use https://docs.rs/cached_path?

With which we no longer care about having the `node-ui` dir in the core.

crates/server/build.rs
miraclx, line 1: So here, we want the following behaviours:

1. use `cached_path` to download the archive
2. the URL should be overridable with an env variable `CALIMERO_WEB_UI_SRC` (since `cached_path` supports file paths and URLs)
3. an environment variable `CALIMERO_WEB_UI_FETCH` should let the user force redownload in case the cache is out of date
4. we want to rerun the build script if the contents of the directory changes ([`cargo::rerun-if-changed=PATH`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed))
5. use [`std::option_env!`](https://doc.rust-lang.org/std/macro.option_env.html) which will also rerun the build script when either of these envs change.
6. use [`cargo::rustc-env=VAR=VALUE`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env) to set the path environment variable which will then be used by the bundler.

see https://doc.rust-lang.org/cargo/reference/build-scripts.html for reference on writing build scripts

frdomovic: Updated with new build.rs script. Tested and has following behaviour:
1. Downloads repo into target (crates/server/target) - serves as cache
2. C/P build folder into node-ui (root/node-ui/build) - which service.rs takes in
3. Loads the UI properly

Rebuilds if the cache folder or node-ui folders are missing in the dir 

https://github.com/user-attachments/assets/9f9565c8-f3c0-4810-a8a1-af47320a324d

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+use cached_path::{Cache, Options};
+use zip::ZipArchive;
 
 fn main() {
-    let status = Command::new("sh")
-        .arg("../../scripts/download-node-ui.sh")
-        .status()
-        .expect("Failed to execute script");
+    let src = option_env!("CALIMERO_WEB_UI_SRC").unwrap_or(
+        "https://github.com/calimero-network/admin-dashboard/archive/refs/heads/master.zip",
```
miraclx: let's use a const value before `fn main`

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+            env::var("CARGO_TARGET_DIR").unwrap_or_else(|_| "target".into()),
+        ))
+        .build()
+        .expect("Failed to create cache");
+
+    let options = if force {
+        Options::default().extract()
+    } else {
+        Options::default()
+    };
```
miraclx: you already even use `extract` here, but I'm not sure I understand why we're only doing that when we're force-fetching

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+    let marker_file = project_root.join("build-deps-changed.txt");
+
+    let target_missing = !Path::new("target").exists();
+    let node_ui_missing = !static_files_target.exists();
+
+    if target_missing || node_ui_missing {
+        eprintln!("Triggering rebuild because `target/` or `node-ui/build` is missing.");
+        fs::write(&marker_file, format!("trigger: {:?}\n", std::time::SystemTime::now()))
+            .expect("Failed to write marker file");
+    }
```
miraclx: why do we have a marker file?

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
-use std::process::Command;
+use std::path::{Path, PathBuf};
+use std::{env, fs};
+
+use cached_path::{Cache, Options};
+use zip::ZipArchive;
```
miraclx: `cached_path` supports extracting archives ([ref](https://docs.rs/cached-path/latest/cached_path/struct.Options.html#method.extract)) which you already use, though only when force fetching?

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+        fs::remove_dir_all(&out_dir).expect("Failed to remove existing output directory");
+    }
+    fs::create_dir_all(&out_dir).expect("Failed to create output directory");
 
-    if !status.success() {
-        panic!("Script failed with status: {:?}", status);
+    let zip_file = fs::File::open(&archive_path).expect("Cannot open downloaded ZIP archive");
+    let mut zip = ZipArchive::new(zip_file).expect("Failed to read ZIP archive");
+    zip.extract(&out_dir)
+        .expect("Failed to extract ZIP archive");
```
miraclx: what we receive from the `cached_path_with_options` call should be the dir containing the admin dashboard, no further extraction necessary

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+            if file_type.is_dir() {
+                fs::create_dir_all(&dst_path).unwrap();
+                copy_dir_all(&entry.path(), &dst_path);
+            } else {
+                let _ = fs::copy(entry.path(), dst_path).expect("Failed to copy file");
+            }
+        }
     }
+
+    copy_dir_all(&extracted_build_path, &static_files_target);
```
miraclx: we shouldn't need these

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+
+    println!("cargo:rerun-if-env-changed=CALIMERO_WEB_UI_SRC");
+    println!("cargo:rerun-if-env-changed=CALIMERO_WEB_UI_FETCH");
```
miraclx: so long as you use `option_env!`, this is covered

> "As of 1.46, using [env!](https://doc.rust-lang.org/std/macro.env.html) and [option_env!](https://doc.rust-lang.org/std/macro.option_env.html) in source code will automatically detect changes and trigger rebuilds. rerun-if-env-changed is no longer needed for variables already referenced by these macros."

https://doc.rust-lang.org/cargo/reference/build-scripts.html#:~:text=As%20of%201.46%2C%20using%20env!%20and%20option_env!%20in%20source%20code%20will%20automatically%20detect%20changes%20and%20trigger%20rebuilds.%20rerun%2Dif%2Denv%2Dchanged%20is%20no%20longer%20needed%20for%20variables%20already%20referenced%20by%20these%20macros.

crates/server/build.rs
```diff
@@ -1,12 +1,95 @@
...
+
+    if Path::new(src).exists() {
+        println!("cargo:rerun-if-changed={}", src);
+    }
+
+    println!("cargo:rerun-if-changed={}", marker_file.display());
+
+    println!("cargo:rerun-if-changed=build.rs");
```
miraclx: > "Otherwise, specifying build.rs is redundant and unnecessary."

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed:~:text=Otherwise%2C%20specifying%20build.rs%20is%20redundant%20and%20unnecessary.

crates/server/build.rs
frdomovic: Removed

crates/server/build.rs
```diff
@@ -33,63 +16,23 @@ fn main() {
         .build()
         .expect("Failed to create cache");
 
-    let options = if force {
-        Options::default().extract()
-    } else {
-        Options::default()
-    };
+    let mut options = Options::default();
+    options.extract = true;
```
miraclx:
```suggestion
    let options = Options::default().extract();
```

crates/server/build.rs
```diff
@@ -33,63 +16,23 @@ fn main() {
...
-    println!("cargo:rerun-if-env-changed=CALIMERO_WEB_UI_SRC");
-    println!("cargo:rerun-if-env-changed=CALIMERO_WEB_UI_FETCH");
+    println!(
+        "cargo:rustc-env=CALIMERO_WEB_UI_PATH={}",
+        extracted_build_path.display()
+    );
 
     if Path::new(src).exists() {
         println!("cargo:rerun-if-changed={}", src);
     }
```
miraclx: I don't understand this line, checking if the URL exists as a file?

shouldn't you unconditionally issue a `rerun-if-changed` on `extracted_build_path`?

crates/server/src/admin/service.rs
```diff
@@ -314,69 +310,57 @@ pub(crate) fn site(config: &ServerConfig) -> Option<(&'static str, Router)> {
...
+        .join(base_path)
+        .join(rel_path);
+
+    match try_read_file(&full_path).await {
+        Ok(response) => Ok(response),
+        Err(_) if rel_path != "index.html" => {
+            let index_path = base_path.join("index.html");
+            try_read_file(&index_path)
+                .await
+                .map_err(|_| StatusCode::NOT_FOUND)
```
miraclx: this assumes the node is run on the same machine that built it

crates/server/src/admin/service.rs
```diff
@@ -69,11 +70,6 @@ impl AdminConfig {
     }
 }
 
-// Embed the contents of the admin-ui build directory into the binary
-#[derive(RustEmbed)]
-#[folder = "../../node-ui/build/"]
-struct NodeUiStaticFiles;
```
miraclx: we don't want to remove this, just that the path is now from the env variable, nothing else needs to change

miraclx (commented):
we're also missing `CALIMERO_WEB_UI_FETCH`, and while `cached_path` doesn't currently support forced download even with a cache-hit, I filed https://github.com/epwalsh/rust-cached-path/issues/80

and the maintainer has already implemented the patch, you can use that branch to test it out by adding this to the root manifest

```toml
[patch.crates-io]
cached-path = { git = "https://github.com/epwalsh/rust-cached-path.git", branch = "epwalsh/force-download" }
```

crates/server/src/admin/service.rs
frdomovic: Any idea on how to use it? using env in my case was throwing error that it needs to have a "static path"

crates/server/build.rs
frdomovic: When this line is removed using `cargo run -p merod -- --node-name node1 run` will only setup this once. If you manually delete the target folder it won't re-run it unless the build.rs file is updated again. And adding it will run the build.rs when you start the node and not have target folder already. 

Just covering this case..

crates/server/src/admin/service.rs
miraclx: according to https://docs.rs/rust-embed-impl/8.7.0/src/rust_embed_impl/lib.rs.html#365-368, it calls out to `shellexpand` when the `interpolate-folder-path` feature flag is activated (which it isn't by default)

so this should work (at least in theory):

```suggestion
// Embed the contents of the admin-ui build directory into the binary
#[derive(RustEmbed)]
#[folder = "$CALIMERO_WEB_UI_PATH"]
struct NodeUiStaticFiles;
```

frdomovic:
Ran tests 2 times and failing clippy is now passing, not sure about the build cc @miraclx @alenmestrov

crates/server/build.rs
```diff
@@ -0,0 +1,92 @@
...
+            eprintln!("Failed to send request: {e}");
+            std::process::exit(1);
+        })
+        .json()
+        .unwrap_or_else(|e| {
+            eprintln!("Failed to parse JSON: {e}");
+            std::process::exit(1);
+        });
+
+    let latest_release_src = format!("{}{}.zip", CALIMERO_WEB_UI_SRC, release.tag_name);
```
miraclx: instad of assuming a format here, there's a `.zipball_url` in response

crates/server/build.rs
```diff
@@ -0,0 +1,92 @@
...
+        .send()
+        .unwrap_or_else(|e| {
+            eprintln!("Failed to send request: {e}");
+            std::process::exit(1);
+        })
+        .json()
+        .unwrap_or_else(|e| {
+            eprintln!("Failed to parse JSON: {e}");
+            std::process::exit(1);
+        });
```
miraclx: these should only happen when the `CALIMERO_WEB_UI_SRC` env is not provided

crates/server/build.rs
```diff
@@ -0,0 +1,92 @@
...
+use reqwest::blocking::Client;
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+struct Release {
+    tag_name: String,
+}
+
+const CALIMERO_WEB_UI_RELEASE: &str =
+    "https://api.github.com/repos/calimero-network/admin-dashboard/releases/latest";
```
miraclx: with this, it makes sense to make the tag overridable with a env var

crates/server/build.rs
```diff
@@ -0,0 +1,92 @@
...
+            std::process::exit(1);
+        });
+
+    let latest_release_src = format!("{}{}.zip", CALIMERO_WEB_UI_SRC, release.tag_name);
+    let src = option_env!("CALIMERO_WEB_UI_SRC").unwrap_or(&latest_release_src);
+
+    let cache = Cache::builder()
+        .dir(PathBuf::from(
+            env::var("CARGO_TARGET_DIR").unwrap_or_else(|_| "target".into()),
+        ))
```
miraclx: 1. See https://redirect.github.com/rust-lang/cargo/issues/9661#issuecomment-1722358176
2. Put the files in a subdir of the `target` folder, so it doesn't get littered

crates/server/build.rs
miraclx: what confuses me is the `Path::new(src)`

when the path is a URL, won't that be false?

and also

1. at this point in the code, isn't it already guaranteed we have the web UI path existing?
2. we should unconditionally rerun the build script if the dir changes

miraclx (commented):
`cached-path v0.7.0` has been released with the forced download option.

crates/server/build.rs
frdomovic: zipball_url get 403

crates/server/build.rs
frdomovic: added a target subdir target/calimero-web-ui-cache.
Also this target is not in root folder but in root/crates/server/target/...

frdomovic:
`println!("cargo:rerun-if-changed={}", src);`
- removed this part cc @miraclx

crates/server/src/admin/service.rs
miraclx: Not sure what the issue was, but it works on my end 🤷🏽

miraclx (approved):

PR #1253, opened 2025-05-08
feat:  merod methods adjustment to context and node client

alenmestrov:
## Description

Reworked node crate to use Context and Node Clients instead of ctx_manager

## Test plan

N/A

## Documentation update

N/A

### DIFF
--- a/crates/node/src/interactive_cli/context.rs
+++ b/crates/node/src/interactive_cli/context.rs
@@ -5,9 +5,9 @@ use calimero_primitives::application::ApplicationId;
 use calimero_primitives::context::{ContextId, ContextInvitationPayload};
 use calimero_primitives::hash::Hash;
 use calimero_primitives::identity::{PrivateKey, PublicKey};
-use futures_util::{pin_mut, TryStreamExt};
 use clap::{Parser, Subcommand, ValueEnum};
 use eyre::{OptionExt, Result as EyreResult};
+use futures_util::{pin_mut, TryStreamExt};
 use owo_colors::OwoColorize;
 use serde_json::Value;
 
@@ -139,11 +139,11 @@ impl ContextCommand {
 
                 // Create a local handle to use within the async block that returns a stream
                 let ctx_client_ref = ctx_client;
-                
+
                 // Use a simpler approach by iterating manually through the stream
                 let mut context_ids = Vec::new();
                 pin_mut!(context_stream);
-                
+
                 // Manually iterate over the stream
                 while let Some(context_id_result) = context_stream.try_next().await? {
                     context_ids.push(context_id_result);
@@ -153,16 +153,16 @@ impl ContextCommand {
                 for context_id in context_ids {
                     if let Ok(Some(context)) = ctx_client_ref.get_context(&context_id) {
                         if let Ok(Some(config)) = ctx_client_ref.context_config(&context_id) {
-                    let entry = format!(
-                        "{c1:44} | {c2:44} | {c3:44} | {c4:8}",
+                            let entry = format!(
+                                "{c1:44} | {c2:44} | {c3:44} | {c4:8}",
                                 c1 = context.id,
                                 c2 = context.application_id,
                                 c3 = context.root_hash,
-                        c4 = config.protocol,
-                    );
+                                c4 = config.protocol,
+                            );
 
-                    for line in entry.lines() {
-                        println!("{ind} {}", line.cyan());
+                            for line in entry.lines() {
+                                println!("{ind} {}", line.cyan());
                             }
                         }
                     }
@@ -176,7 +176,7 @@ impl ContextCommand {
                     .join_context(private_key, invitation_payload)
                     .await?;
 
-                    println!(
+                println!(
                     "{ind} Joined context {} as {}, waiting for catchup to complete...",
                     response.context_id, response.member_public_key
                 );
@@ -203,19 +203,24 @@ impl ContextCommand {
                     .resolve_alias(application, None)?
                     .ok_or_eyre("unable to resolve")?;
 
-                let response = ctx_client.create_context(
-                    protocol.as_str().to_string(),
-                    &application_id,
-                    None,
-                    params
-                        .as_ref()
-                        .map(serde_json::to_vec)
-                        .transpose()?
-                        .unwrap_or_default(),
+                let response = ctx_client
+                    .create_context(
+                        protocol.as_str().to_string(),
+                        &application_id,
+                        None,
+                        params
+                            .as_ref()
+                            .map(serde_json::to_vec)
+                            .transpose()?
+                            .unwrap_or_default(),
                         context_seed.map(Into::into),
-                ).await?;
+                    )
+                    .await?;
 
-                println!("{ind} Created context {} for application {}", response.context_id, application_id);
+                println!(
+                    "{ind} Created context {} for application {}",
+                    response.context_id, application_id
+                );
             }
             Commands::Invite {
                 context,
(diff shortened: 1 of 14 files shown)


crates/node/src/interactive_cli/context.rs
```diff
@@ -136,55 +134,59 @@ impl ContextCommand {
...
-                        "{c1:44} | {c2:44} | {c3:44} | {c4:8}",
-                        c1 = context_id,
-                        c2 = v.application.application_id(),
-                        c3 = Hash::from(v.root_hash),
-                        c4 = config.protocol,
-                    );
+                // Manually iterate over the stream
+                while let Some(context_id_result) = context_stream.try_next().await? {
+                    context_ids.push(context_id_result);
+                }
```
miraclx: these comments are jarring

crates/node/src/interactive_cli/context.rs
```diff
@@ -136,55 +134,59 @@ impl ContextCommand {
...
 
-                    for line in entry.lines() {
-                        println!("{ind} {}", line.cyan());
+                // Process each context ID sequentially
+                for context_id in context_ids {
```
miraclx: why iterate twice?

crates/node/src/interactive_cli/context.rs
```diff
@@ -136,55 +134,59 @@ impl ContextCommand {
...
 
-                let handle = node.store.handle();
+                // First, get a list of contexts using get_contexts and try_collect
+                let context_stream = ctx_client.get_contexts(None).await;
 
-                for (k, v) in handle.iter::<ContextMetaKey>()?.entries() {
-                    let (k, v) = (k?, v?);
-                    let context_id = k.context_id();
+                // Create a local handle to use within the async block that returns a stream
+                let ctx_client_ref = ctx_client;
```
miraclx: I don't understand this line 😅

crates/node/src/interactive_cli/context.rs
```diff
@@ -197,56 +199,43 @@ impl ContextCommand {
...
+                        context_seed.map(Into::into),
+                    )
+                    .await?;
 
-                    println!("{ind} Unable to create context: {err:?}");
-                });
+                println!(
+                    "{ind} Created context {} for application {}",
+                    response.context_id, application_id
+                );
```
miraclx: let's preserve the success and error messages

```
println!("{ind} Created context {context_id} with identity {identity}");
```

crates/node/src/interactive_cli/context.rs
```diff
@@ -136,55 +134,59 @@ impl ContextCommand {
...
+                // Create a local handle to use within the async block that returns a stream
+                let ctx_client_ref = ctx_client;
 
-                    // Get the config for this context
-                    let config = handle
-                        .get(&ContextConfigKey::new(context_id))?
-                        .expect("Context config must exist with protocol");
+                // Use a simpler approach by iterating manually through the stream
+                let mut context_ids = Vec::new();
+                pin_mut!(context_stream);
```
miraclx: so this is fine, but since the `pin!` macro in stdlib got stabilized I've been using that instead

https://doc.rust-lang.org/std/pin/macro.pin.html

crates/node/src/interactive_cli/context.rs
```diff
@@ -136,55 +134,59 @@ impl ContextCommand {
                     c3 = "Root Hash"
                 );
 
-                let handle = node.store.handle();
+                // First, get a list of contexts using get_contexts and try_collect
+                let context_stream = ctx_client.get_contexts(None).await;
```
miraclx: why not name this `contexts`, it being a stream is an implementation detail

crates/node/src/interactive_cli/context.rs
```diff
@@ -259,47 +248,66 @@ impl ContextCommand {
...
+                        return Ok(());
+                    }
+                    Err(err) => return Err(err),
+                };
+
+                let external_client =
+                    match ctx_client.external_client(&context_id, &external_config) {
+                        Ok(client) => client,
+                        Err(err) => return Err(err),
+                    };
```
miraclx: this is essentially this:

```suggestion
                let external_client = ctx_client.external_client(&context_id, &external_config)?;
```

crates/node/src/interactive_cli/context.rs
```diff
@@ -259,47 +248,66 @@ impl ContextCommand {
...
-                node.ctx_manager
-                    .update_context_proxy(context_id, public_key)
+                let external_config = match ctx_client.context_config(&context_id) {
+                    Ok(Some(config)) => config,
+                    Ok(None) => {
+                        println!("{ind} Context configuration not found for {context_id}");
+                        return Ok(());
+                    }
+                    Err(err) => return Err(err),
+                };
```
miraclx: same here

```suggestion
                let Some(external_config) = ctx_client.context_config(&context_id)? else {
                    println!("{ind} Context configuration not found for {context_id}");
                    return Ok(());
                };
```

crates/node/src/interactive_cli/identity.rs
```diff
@@ -75,59 +77,58 @@ enum AliasSubcommands {
...
-    let first = 'first: {
-        let Some(k) = iter
-            .seek(ContextIdentityKey::new(context_id, [0; 32].into()))
-            .transpose()
-        else {
-            break 'first None;
-        };
+    // Pin the streams
+    futures_util::pin_mut!(stream);
+    futures_util::pin_mut!(stream_owned);
```
miraclx: whoops, that's an unintended effect of the API change..

how about `context_members` returns `(public_key, is_owned)`?

since we already know that information within the function

crates/node/src/interactive_cli/identity.rs
```diff
@@ -75,59 +77,58 @@ enum AliasSubcommands {
...
-    println!("{ind} {:44} | {}", "Identity", "Owned");
+    // Collect all owned identities to check which ones are owned
+    let owned_identities = stream_owned.try_collect::<Vec<_>>().await?;
 
-    for (k, v) in first.into_iter().chain(iter.entries()) {
-        let (k, v) = (k?, v?);
+    // Display each identity
+    for identity in identities {
+        // Check if this identity is in the owned identities list
+        let is_owned = owned_identities.iter().any(|owned| *owned == identity);
```
miraclx: also, `owned.contains(..)` would've worked just fine, but I prefer if we didn't use this approach at all

crates/node/src/interactive_cli/peers.rs
```diff
@@ -15,25 +14,26 @@ pub struct PeersCommand {
...
-            .map(|context| node.ctx_manager.resolve_alias(context, None))
+            .map(|context| node_client.resolve_alias(context, None))
             .transpose()?
             .flatten();
 
+        println!(
+            "{ind} Peers (General): {:#?}",
+            node_client.get_peers_count(context_id).await.cyan()
+        );
+
```
miraclx: this will print the number of peers connected for a context when the context is provided

`General` is for ALL nodes connected to the node, regardless of which context they're connected on

let's revert and just change the relevant bits in place without changing the structure

crates/node/src/interactive_cli/peers.rs
```diff
@@ -15,25 +14,26 @@ pub struct PeersCommand {
...
         if let Some(context_id) = context_id {
             let topic = TopicHash::from_raw(context_id);
             println!(
                 "{ind} Peers (Session) for Topic {}: {:#?}",
                 topic.clone(),
-                node.network_client.mesh_peer_count(topic).await.cyan()
+                node_client.get_peers_count(None).await.cyan()
```
miraclx: same here

crates/node/src/lib.rs
```diff
@@ -29,8 +29,8 @@ pub struct NodeManager {
     pub datastore: Store,
     pub blobstore: BlobManager,
 
-    pub context_manager: ContextClient,
-    pub network_manager: NetworkClient,
+    pub context_client: ContextClient,
+    pub node_client: NodeClient,
```
miraclx, line 33: restore `network_manager` as `network_client`

crates/node/src/interactive_cli/context.rs
```diff
@@ -134,25 +134,12 @@ impl ContextCommand {
...
-                    context_ids.push(context_id_result);
-                }
-
-                // Process each context ID sequentially
-                for context_id in context_ids {
-                    if let Ok(Some(context)) = ctx_client_ref.get_context(&context_id) {
-                        if let Ok(Some(config)) = ctx_client_ref.context_config(&context_id) {
+                while let Some(context_id) = contexts.try_next().await? {
+                    if let Ok(Some(context)) = ctx_client.get_context(&context_id) {
+                        if let Ok(Some(config)) = ctx_client.context_config(&context_id) {
```
miraclx: `let-else` can help reduce indentations here

```rust
let Some(..) = __ else {
    continue;
};
```

for both

miraclx (approved):

crates/node/src/interactive_cli/context.rs
miraclx: also, `?` is missing, we're potentially ignoring errors

Issue #919, opened 2024-11-01
Do not send private keys over the wire ever


fbozic:
Original comment: https://github.com/calimero-network/core/pull/918#discussion_r1826036086

Quote from Mira:
> The idea is, `new_identity` will store it ahead of time, and only return a public key.
So when you get back with an invitation payload that includes that public key, it'll take it from the local store and bind it to the context

onyedikachi-david:
please assign @antonpaisov

PR #1256, opened 2025-05-10
fix: remove private key exposure

onyedikachi-david:
# [Core] Server-Side Private Key Management for Context Join

## Description

Refactors context identity generation and joining. A new endpoint generates keys, temporarily stores the private key server-side, and returns only the public key. The `join_context` endpoint now takes this public key to retrieve the stored private key for joining, then deletes it. This enhances security by not requiring clients to handle private keys for joining.

## Documentation update

Fixes: #919

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -1610,4 +1610,41 @@ impl ContextManager {
 
         Ok(aliases)
     }
+
+    /// Stores a ContextIdentity value directly.
+    /// Used internally and potentially by admin handlers for pre-storing keys.
+    pub fn store_identity_value(
+        &self,
+        context_id: ContextId,
+        public_key: PublicKey,
+        value: ContextIdentityValue,
+    ) -> EyreResult<()> {
+        let key = ContextIdentityKey::new(context_id, public_key);
+        let mut handle = self.store.handle();
+        handle.put(&key, &value)?;
+        Ok(())
+    }
+
+    /// Retrieves a complete ContextIdentity value.
+    pub fn get_identity_value(
+        &self,
+        context_id: ContextId,
+        public_key: PublicKey,
+    ) -> EyreResult<Option<ContextIdentityValue>> {
+        let key = ContextIdentityKey::new(context_id, public_key);
+        let handle = self.store.handle();
+        handle.get(&key).map_err(eyre::Report::from)
+    }
+
+    /// Deletes a ContextIdentity value.
+    pub fn delete_identity_value(
+        &self,
+        context_id: ContextId,
+        public_key: PublicKey,
+    ) -> EyreResult<()> {
+        let key = ContextIdentityKey::new(context_id, public_key);
+        let mut handle = self.store.handle();
+        handle.delete(&key)?;
+        Ok(())
+    }
 }
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -1611,8 +1611,6 @@ impl ContextManager {
         Ok(aliases)
     }
 
-    /// Stores a ContextIdentity value directly.
-    /// Used internally and potentially by admin handlers for pre-storing keys.
     pub fn store_identity_value(
         &self,
         context_id: ContextId,
@@ -1625,7 +1623,6 @@ impl ContextManager {
         Ok(())
     }
 
-    /// Retrieves a complete ContextIdentity value.
     pub fn get_identity_value(
         &self,
         context_id: ContextId,
@@ -1636,7 +1633,6 @@ impl ContextManager {
         handle.get(&key).map_err(eyre::Report::from)
     }
 
-    /// Deletes a ContextIdentity value.
     pub fn delete_identity_value(
         &self,
         context_id: ContextId,
--- a/crates/meroctl/src/cli/context/identity/generate.rs
+++ b/crates/meroctl/src/cli/context/identity/generate.rs
@@ -13,14 +13,13 @@ pub struct GenerateCommand;
 impl Report for GenerateContextIdentityResponse {
     fn report(&self) {
         println!("public_key: {}", self.data.public_key);
-        println!("private_key: {}", self.data.private_key);
     }
 }
 
 impl GenerateCommand {
     pub async fn run(self, environment: &Environment) -> EyreResult<()> {
         let private_key = PrivateKey::random(&mut rand::thread_rng());
-        let response = GenerateContextIdentityResponse::new(private_key.public_key(), private_key);
+        let response = GenerateContextIdentityResponse::new(private_key.public_key());
         environment.output.write(&response);
 
         Ok(())
--- a/crates/server-primitives/src/admin.rs
+++ b/crates/server-primitives/src/admin.rs
@@ -5,7 +5,7 @@ use calimero_primitives::alias::Alias;
 use calimero_primitives::application::{Application, ApplicationId};
 use calimero_primitives::context::{Context, ContextId, ContextInvitationPayload};
 use calimero_primitives::hash::Hash;
-use calimero_primitives::identity::{ClientKey, ContextUser, PrivateKey, PublicKey, WalletType};
+use calimero_primitives::identity::{ClientKey, ContextUser, PublicKey, WalletType};
 use camino::Utf8PathBuf;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -346,17 +346,14 @@ impl InviteToContextResponse {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct JoinContextRequest {
-    pub private_key: PrivateKey,
+    pub public_key: PublicKey,
     pub invitation_payload: ContextInvitationPayload,
 }
 
 impl JoinContextRequest {
-    pub const fn new(
-        private_key: PrivateKey,
-        invitation_payload: ContextInvitationPayload,
-    ) -> Self {
+    pub const fn new(public_key: PublicKey, invitation_payload: ContextInvitationPayload) -> Self {
         Self {
-            private_key,
+            public_key,
             invitation_payload,
         }
     }
@@ -419,7 +416,6 @@ impl UpdateContextApplicationResponse {
 #[serde(rename_all = "camelCase")]
 pub struct GenerateContextIdentityResponseData {
     pub public_key: PublicKey,
-    pub private_key: PrivateKey,
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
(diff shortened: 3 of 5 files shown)


frdomovic:
@onyedikachi-david can you add test plan/ testing steps please?

crates/server-primitives/src/admin.rs
```diff
@@ -346,17 +346,14 @@ impl InviteToContextResponse {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct JoinContextRequest {
-    pub private_key: PrivateKey,
+    pub public_key: PublicKey,
     pub invitation_payload: ContextInvitationPayload,
 }
```
miraclx: the `public_key` is already in the invitation payload (as `invitee_payload`)

crates/server/src/admin/handlers/context/join_context.rs
```diff
@@ -1,19 +1,90 @@
...
+            } else {
+                ApiError {
+                    status_code: StatusCode::INTERNAL_SERVER_ERROR,
+                    message: "Failed to process join request".into(),
+                }
+            };
+            return api_error.into_response();
+        }
+    };
+
```
miraclx: none of this should be necessary at this level, fetching the private key for the identity should happen inside `join_context`

crates/server/src/admin/handlers/identity/generate_context_identity.rs
```diff
@@ -1,17 +1,46 @@
...
+        let placeholder_context_id = ContextId::from(PLACEHOLDER_CONTEXT_ID_BYTES);
+        let value = ContextIdentityValue {
+            private_key: Some(*private_key),
+            sender_key: None,
+        };
+        state
+            .ctx_manager
+            .store_identity_value(placeholder_context_id, public_key, value)?;
+        Ok::<(), eyre::Report>(())
+    })();
```
miraclx: this should be the responsibility of some function inside context manager (so it can be used by merod's interactive CLI too)

miraclx (commented):
these accomodations need to apply to merod's interactive CLI as well, so no special behaviour should live on the server, they should be in context manager instead

crates/context/src/lib.rs
```diff
@@ -345,20 +345,35 @@ impl ContextManager {
 
     pub async fn join_context(
         &self,
-        identity_secret: PrivateKey,
+        invitee_public_key_from_payload: PublicKey,
         invitation_payload: ContextInvitationPayload,
```
miraclx: the invitation payload already includes the public key, we don't need to take it in anymore

crates/context/src/lib.rs
```diff
@@ -346,20 +346,30 @@ impl ContextManager {
 
     pub async fn join_context(
         &self,
-        identity_secret: PrivateKey,
         invitation_payload: ContextInvitationPayload,
     ) -> EyreResult<Option<(ContextId, PublicKey)>> {
-        let (context_id, invitee_id, protocol, network_id, contract_id) =
+        let (context_id, invitee_id_from_payload, protocol, network_id, contract_id) =
```
miraclx: at this point, you no longer need to rename the variable

crates/context/src/lib.rs
```diff
@@ -346,20 +346,30 @@ impl ContextManager {
...
-            bail!("identity mismatch")
-        }
+        let placeholder_context_id = ContextId::from([0; 32]);
 
-        let handle = self.store.handle();
+        let stored_identity = self
+            .get_identity_value(placeholder_context_id, invitee_id_from_payload)?
+            .ok_or_else(|| {
+                eyre::eyre!(
+                    "Pre-stored private key not found for public key: {}",
```
miraclx: No need to indicate it was pre-stored, try something like

```rust
"Missing identity for {invitee_id}"
```

or even "Unknown identity", or permission denied works too

crates/context/src/lib.rs
```diff
@@ -346,20 +346,30 @@ impl ContextManager {
...
-        if identity_secret.public_key() != invitee_id {
-            bail!("identity mismatch")
-        }
+        let placeholder_context_id = ContextId::from([0; 32]);
 
-        let handle = self.store.handle();
+        let stored_identity = self
+            .get_identity_value(placeholder_context_id, invitee_id_from_payload)?
+            .ok_or_else(|| {
+                eyre::eyre!(
```
miraclx: [`.ok_or_eyre`](https://docs.rs/eyre/latest/eyre/trait.OptionExt.html#tymethod.ok_or_eyre)

crates/context/src/lib.rs
```diff
@@ -346,20 +346,30 @@ impl ContextManager {
...
+                    "Pre-stored private key not found for public key: {}",
+                    invitee_id_from_payload
+                )
+            })?;
+
+        let identity_secret = stored_identity
+            .private_key
+            .ok_or_else(|| eyre::eyre!("Stored identity value is missing private key"))?;
```
miraclx: we also want to fill in the sender_key here as well, also: `ok_or_eyre`

crates/context/src/lib.rs
```diff
@@ -1612,6 +1621,53 @@ impl ContextManager {
...
+        context_id: ContextId,
+        public_key: PublicKey,
+    ) -> EyreResult<()> {
+        let key = ContextIdentityKey::new(context_id, public_key);
+        let mut handle = self.store.handle();
+        handle.delete(&key)?;
+        Ok(())
+    }
+
+    pub fn pre_store_new_identity(&self) -> EyreResult<PublicKey> {
```
miraclx: just call it `new_identity`

crates/context/src/lib.rs
```diff
@@ -1612,6 +1621,53 @@ impl ContextManager {
...
+    pub fn delete_identity_value(
+        &self,
+        context_id: ContextId,
+        public_key: PublicKey,
+    ) -> EyreResult<()> {
+        let key = ContextIdentityKey::new(context_id, public_key);
+        let mut handle = self.store.handle();
+        handle.delete(&key)?;
+        Ok(())
+    }
```
miraclx: without calling these methods anywhere else, localize them

crates/meroctl/src/cli/context/join.rs
```diff
@@ -17,10 +17,10 @@ use crate::output::Report;
 #[command(about = "Join an application context")]
 pub struct JoinCommand {
     #[clap(
-        value_name = "PRIVATE_KEY",
-        help = "The private key for signing the join context request"
+        value_name = "PUBLIC_KEY",
+        help = "The public key of the identity joining the context (obtained from 'context identity generate')"
     )]
-    pub private_key: PrivateKey,
+    pub public_key: PublicKey,
```
miraclx: this shouldn't still be here

crates/context/src/lib.rs
```diff
@@ -298,7 +299,8 @@ impl ContextManager {
     fn add_context(
         &self,
         context: &Context,
-        identity_secret: PrivateKey,
+        main_identity_secret: PrivateKey,
+        sender_identity_secret: PrivateKey,
```
miraclx: what's with these names? - I don't expect any changes to `add_context`

crates/context/src/lib.rs
```diff
@@ -1654,14 +1663,15 @@ impl ContextManager {
...
+        let main_private_key = self.new_private_key();
+        let public_key = main_private_key.public_key();
+        let sender_private_key = self.new_private_key();
 
         let placeholder_context_id = ContextId::from([0u8; 32]);
         let value = ContextIdentityValue {
-            private_key: Some(*private_key),
-            sender_key: None,
+            private_key: Some(*main_private_key),
+            sender_key: Some(*sender_private_key),
```
miraclx: we don't need to pre-determine a sender key until join happens

crates/meroctl/src/cli/context/identity/generate.rs
```diff
@@ -13,20 +14,31 @@ pub struct GenerateCommand;
 
 impl Report for GenerateContextIdentityResponse {
     fn report(&self) {
+        println!("public_key: {}", self.data.public_key);
+
```
miraclx: this is already in the table

crates/node/src/interactive_cli/context.rs
```diff
@@ -225,15 +225,12 @@ impl ContextCommand {
                 }
             }
             Commands::Join {
-                private_key,
+                private_key: _,
```
miraclx: ? you should have removed the `private_key`

crates/server/src/admin/handlers/identity/generate_context_identity.rs
```diff
@@ -1,17 +1,27 @@
...
-
-    ApiResponse {
-        payload: GenerateContextIdentityResponse::new(private_key.public_key(), private_key),
+    match state.ctx_manager.new_identity() {
+        Ok(public_key) => ApiResponse {
+            payload: GenerateContextIdentityResponse::new(public_key),
+        }
+        .into_response(),
+        Err(e) => {
+            error!("Failed to generate and pre-store context identity: {}", e);
```
miraclx: "pre-storing" is an internal implementation detail

"Failed to generate context identity" is sufficient

miraclx (dismissed):

crates/context/src/lib.rs
```diff
@@ -346,20 +346,28 @@ impl ContextManager {
...
+            .get_identity_value(placeholder_context_id, invitee_id)?
+            .ok_or_eyre(eyre::eyre!(
+                "Missing identity for public key: {}",
+                invitee_id
+            ))?;
 
-        let identity_key = ContextIdentityKey::new(context_id, invitee_id);
+        let main_private_key_bytes = stored_identity
+            .private_key
+            .ok_or_eyre(eyre::eyre!("Stored identity value is missing private key"))?;
```
```suggestion
            .ok_or_eyre("Stored identity value is missing private key")?;
```

crates/context/src/lib.rs
```diff
@@ -346,20 +346,28 @@ impl ContextManager {
...
-        let handle = self.store.handle();
+        let stored_identity = self
+            .get_identity_value(placeholder_context_id, invitee_id)?
+            .ok_or_eyre(eyre::eyre!(
+                "Missing identity for public key: {}",
+                invitee_id
+            ))?;
 
-        let identity_key = ContextIdentityKey::new(context_id, invitee_id);
+        let main_private_key_bytes = stored_identity
```
```suggestion
        let private_key = stored_identity
```

or `identity_secret`

crates/context/src/lib.rs
miraclx: this is not resolved

onyedikachi-david:
bump

onyedikachi-david:
Done @miraclx

miraclx (approved):

miraclx:
> there are a few double dereferences like this: .send(**private_key, nonce). Not sure if they are required

yes they are, see [`756a6e6` (#1256)](https://github.com/calimero-network/core/pull/1256/commits/756a6e624996b52f54145f3652cc51e3de88fd08)

PR #1257, opened 2025-05-10
feat(cli): enhance --version outputs for merod cli

dotandev:
## Description

Attempts to resolve https://github.com/calimero-network/core/issues/1074.

This PR enhances the `--version` output of the `merod` CLI by embedding richer build-time metadata, including:

- Git commit hash  
- Git describe (tag + distance + dirty state)  
- Rust compiler version  
- Protocol version (from crate's major version)  

This improves traceability and simplifies debugging across environments with custom or unstable builds.

## Test Plan

Output of `./target/debug/merod --version` after build:

```bash
merod (release 0.6.0) (build 0.5.0-48-g9e3a7c9-dirty) (commit) (rustc 1.85.0) (protocol 0)

```

## Documentation Update
N/A


miraclx (commented):
Hi @dotandev, thanks for taking this up.

Can you apply the format established in the previous attempt: https://github.com/calimero-network/core/pull/1212

this patch applies to both `meroctl` and `merod`, so so you should host all version specific structures in a new `calimero-version` crate (or a new module in `calimero-primitives` works too, maybe even preferable)

crates/merod/build.rs
```diff
@@ -0,0 +1,24 @@
+use std::process::Command;
+
+fn main() {
+    println!("cargo:warning=Running build.rs");
+    let output = Command::new("git")
+        .args(["rev-parse", "--short=HEAD"])
```
miraclx: also, this is why you have [this issue](https://github.com/calimero-network/core/pull/1257#issuecomment-2868593491)

> it gives me the hash, but not with the cli, help me review if there is a reason for that.

```console
$ git rev-parse --short HEAD 2> >(sed 's/^/(stderr) /' >&1) > >(sed 's/^/(stdout) /')
(stdout) ff4ecb0e

$ echo $?
0

$ git rev-parse --short=HEAD 2> >(sed 's/^/(stderr) /' >&1) > >(sed 's/^/(stdout) /')
(stderr) fatal: Needed a single revision

$ echo $?
128
```

antonpaisov:
@miraclx could you please have a look at this one? :)

miraclx:
without looking too deeply, there's nothing immediately obvious to me

cc: @alenmestrov, maybe something is immediately obvious to you that won't be to me, but if this compiles on master, consider reverting your lockfile to the version on master, and rebuild

```console
```

crates/version/build.rs
```diff
@@ -0,0 +1,36 @@
...
+        .arg("--version")
+        .output()
+        .ok()
+        .and_then(|o| String::from_utf8(o.stdout).ok())
+        .map(|s| s.split_whitespace().nth(1).unwrap_or("unknown").to_string())
+        .unwrap_or_else(|| "unknown".to_string());
+
+    println!("cargo:rustc-env=GIT_DESCRIBE={}", git_describe.trim());
+    println!("cargo:rustc-env=GIT_COMMIT={}", git_commit.trim());
+    println!("cargo:rustc-env=RUSTC_VERSION={}", rustc_version.trim());
```
miraclx: Use these variable names https://github.com/calimero-network/core/pull/1212#discussion_r2049442905

> `CALIMERO_BUILD`, `CALIMERO_COMMIT`, `CALIMERO_RUSTC_VERSION`

crates/version/build.rs
```diff
@@ -0,0 +1,36 @@
...
+            "--always",
+            "--dirty=-modified",
+            "--tags",
+            "--match",
+            "[0-9]*",
+        ])
+        .output()
+        .ok()
+        .and_then(|o| String::from_utf8(o.stdout).ok())
+        .unwrap_or_else(|| "unknown".to_string());
```
miraclx: let's get better error handling here

consider breaking up the main function into

`fn main() {}` and `fn try_main() -> eyre::Result<()> {}`

within main, you can then match on the result of try_main, printing the error and exiting with code 1 if there is an error

for reference - https://github.com/near/nearcore/blob/221eb4ff26fcf3b964a7a9f875cfa14fa5277499/neard/build.rs#L119-L124

crates/version/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "calimero-version"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+build = "build.rs"
```
miraclx: this is the default, there won't be any need to explicitly specify it

crates/version/build.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    let rustc_version = Command::new("rustc")
+        .arg("--version")
+        .output()
+        .ok()
+        .and_then(|o| String::from_utf8(o.stdout).ok())
+        .map(|s| s.split_whitespace().nth(1).unwrap_or("unknown").to_string())
+        .unwrap_or_else(|| "unknown".to_string());
```
miraclx: you specify `rustc_version` as a dependency, but don't use it

crates/version/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[cfg(test)]
+mod tests;
+
+pub struct CalimeroVersion {
+    release: &'static str,
+    build: &'static str,
+    commit: &'static str,
+    rustc: &'static str,
+    protocol: &'static str,
```
miraclx: we don't need protocol, yet.

crates/version/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+        let describe = env!("GIT_DESCRIBE");
+        let commit = env!("GIT_COMMIT");
+        let rustc = env!("RUSTC_VERSION");
+        let protocol = env!("CARGO_PKG_VERSION_MAJOR");
+
+        let build = if describe == "unknown" {
+            release
+        } else {
+            Box::leak(describe.to_string().into_boxed_str())
+        };
```
miraclx: https://github.com/calimero-network/core/pull/1212#discussion_r2049457342

crates/version/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+use std::sync::LazyLock;
+
+#[cfg(test)]
+mod tests;
+
+pub struct CalimeroVersion {
+    release: &'static str,
+    build: &'static str,
+    commit: &'static str,
+    rustc: &'static str,
```
miraclx: https://github.com/calimero-network/core/pull/1212#discussion_r2049456880

crates/version/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+    protocol: &'static str,
+}
+
+impl CalimeroVersion {
+    fn new() -> Self {
```
miraclx: consider this API:

```rust
static CURRENT: LazyLock<CalimeroVersion> = LazyLock::new(|| CalimeroVersion {
    release: env!("CARGO_PKG_VERSION"),
    build: env!("CALIMERO_BUILD"),
    commit: env!("CALIMERO_COMMIT"),
    rustc: env!("CALIMERO_RUSTC_VERSION"),
});
static CURRENT_STR: LazyLock<String> = LazyLock::new(|| VERSION.to_string());

#[derive(Clone)]
pub struct CalimeroVersion<'a> {
    pub release: Cow<'a, str>,
    pub build: Cow<'a, str>,
    pub commit: Cow<'a, str>,
    pub rustc: Cow<'a, str>,
}

impl CalimeroVersion<'_> {
    pub fn current() -> CalimeroVersion<'static> {
        CURRENT.clone()
    }

    pub fn current_str() -> &'static str {
        &*CURRENT_STR
    }
}

impl fmt::Display for CalimeroVersion<'_> { ..
```

crates/version/src/tests.rs
```diff
@@ -0,0 +1,60 @@
+use crate::version_str;
+
+#[test]
+fn version_info_contains_crate_version() {
+    let info = version_str();
+    let expected = option_env!("CARGO_PKG_VERSION").unwrap_or("<missing>");
```
miraclx: I think this should be unconditional, `env!`, if it's not specified, the build script must not have succeeded, in which case, this doesn't run

crates/version/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "calimero-version"
+version = "0.1.0"
```
miraclx: now, this can relocate the versions specified in `merod` and `meroctl`, while we revert those to `.workspace = true`

crates/merod/src/cli.rs
```diff
@@ -31,7 +32,7 @@ pub const EXAMPLES: &str = r"
 ";
 
 #[derive(Debug, Parser)]
-#[command(author, version, about, long_about = None)]
+#[command(author, version = version_str(), about, long_about = None)]
```
miraclx: this'll then be

```suggestion
#[command(author, version = CalimeroVersion::current_str(), about, long_about = None)]
```

crates/merod/Cargo.toml
```diff
@@ -42,6 +43,7 @@ calimero-network.workspace = true
 calimero-server = { workspace = true, features = ["jsonrpc", "websocket", "admin"] }
 calimero-store.workspace = true
 calimero-store-rocksdb.workspace = true
+calimero-version.workspace = true
```
miraclx:
```suggestion
calimero-version.workspace = true
```

crates/version/src/lib.rs
```diff
@@ -1,58 +1,49 @@
...
-    release: &'static str,
-    build: &'static str,
-    commit: &'static str,
-    rustc: &'static str,
-    protocol: &'static str,
+static CURRENT: LazyLock<CalimeroVersion<'static>> = LazyLock::new(|| CalimeroVersion {
+    release: Cow::Borrowed(env!("CARGO_PKG_VERSION")),
+    build: Cow::Borrowed(env!("CALIMERO_BUILD")),
+    commit: Cow::Borrowed(env!("CALIMERO_COMMIT")),
+    rustc: Cow::Borrowed(env!("CALIMERO_RUSTC_VERSION")),
```
miraclx: this should work too, less clutter

```suggestion
    release: env!("CARGO_PKG_VERSION").into(),
    build: env!("CALIMERO_BUILD").into(),
    commit: env!("CALIMERO_COMMIT").into(),
    rustc: env!("CALIMERO_RUSTC_VERSION").into(),
```

crates/version/build.rs
```diff
@@ -1,36 +1,56 @@
...
 
-    let git_commit = Command::new("git")
-        .args(["rev-parse", "--short", "HEAD"])
-        .output()
-        .ok()
-        .and_then(|o| String::from_utf8(o.stdout).ok())
-        .unwrap_or_else(|| "unknown".to_string());
+    let git_commit = run_command("git", &["rev-parse", "--short", "HEAD"])?;
+
+    let rustc_version = version()
```
miraclx: version is ambiguous, this whole crate is for versions

```suggestion
    let rustc_version = rustc_version::version()
```

crates/version/src/lib.rs
```diff
@@ -1,58 +1,49 @@
...
-
-pub fn version() -> &'static CalimeroVersion {
-    &*VERSION
-}
+// pub use CalimeroVersion;
 
-pub fn version_str() -> &'static str {
-    &*VERSION_STRING
+pub fn current_str() -> &'static str {
+    CalimeroVersion::current_str()
```
miraclx: callers can call `CalimeroVersion::current_str()` directly though, you can remove both these

crates/merod/Cargo.toml
miraclx: this is unresolved

crates/merod/src/cli.rs
```diff
@@ -3,17 +3,18 @@ use clap::{Parser, Subcommand};
...
 
+use crate::cli::run::RunCommand;
 use crate::defaults;
 
 mod config;
 mod init;
 mod relay;
 mod run;
 
+use calimero_version;
```
miraclx: this shouldn't be here

crates/version/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+build = "build.rs"
+
+[build-dependencies]
+rustc_version = "0.4"
+eyre = "0.6"
```
miraclx: missing `\n` at EOF

crates/version/Cargo.toml
miraclx: this is unresolved

crates/version/Cargo.toml
miraclx: this is unresolved

crates/version/src/lib.rs
dotandev: I tried that, but it wasn't callable.

crates/merod/Cargo.toml
dotandev: I don't understand this comment.

crates/version/src/lib.rs
miraclx: what did you try?

crates/merod/Cargo.toml
miraclx: there are 3 extra spaces at the end

crates/version/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Clone)]
+pub struct CalimeroVersion<'a> {
+    pub release: Cow<'a, str>,
+    pub build: Cow<'a, str>,
+    pub commit: Cow<'a, str>,
+    pub rustc: Cow<'a, str>,
+}
+
+impl CalimeroVersion<'_> {
+    pub fn current() -> CalimeroVersion<'static> {
```
miraclx:
```suggestion
impl CalimeroVersion<'static> {
    pub fn current() -> Self {
```

crates/version/Cargo.toml
miraclx: this is still unresolved, I didn't mean remove `edition`, `repository` and `license`, those should be there

it's the `build =` that doesn't

crates/version/build.rs
```diff
@@ -1,7 +1,7 @@
 use std::process::Command;
 
 use eyre::{eyre, Result as EyreResult};
-use rustc_version::version;
+use rustc_version;
```
miraclx: this import is not needed

crates/version/build.rs
```diff
@@ -25,15 +25,15 @@ fn try_main() -> EyreResult<()> {
 
     let git_commit = run_command("git", &["rev-parse", "--short", "HEAD"])?;
 
-    let rustc_version = version()
+    let calimero_rustc_version = rustc_version::version()
```
miraclx: is there a reason this identifier is qualified with `calimero_` while the others aren't?

crates/meroctl/src/cli.rs
```diff
@@ -38,7 +38,7 @@ pub const EXAMPLES: &str = r"
 ";
 
 #[derive(Debug, Parser)]
-#[command(author, version, about, long_about = None)]
+#[command(author, version = calimero_version::current_str(), about, long_about = None)]
```
```suggestion
#[command(author, version = CalimeroVersion::current_str(), about, long_about = None)]
```

crates/merod/src/cli.rs
```diff
@@ -31,7 +31,7 @@ pub const EXAMPLES: &str = r"
 ";
 
 #[derive(Debug, Parser)]
-#[command(author, version, about, long_about = None)]
+#[command(author, version = calimero_version::current_str(), about, long_about = None)]
```
miraclx:
```suggestion
#[command(author, version = CalimeroVersion::current_str(), about, long_about = None)]
```

crates/version/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+            f,
+            "(release {}) (build {}) (commit {}) (rustc {})",
+            self.release, self.build, self.commit, self.rustc,
+        )
+    }
+}
+
+pub fn current_str() -> &'static str {
+    CalimeroVersion::current_str()
+}
```
miraclx: we really don't need this anymore

crates/version/src/tests.rs
```diff
@@ -1,4 +1,4 @@
-use crate::current_str;
+use crate::CalimeroVersion::current_str;
```
miraclx:
```suggestion
use crate::CalimeroVersion;
```

crates/merod/Cargo.toml
miraclx: still pending, please remove the whitespaces at the end of line 48

crates/merod/src/cli.rs
```diff
@@ -3,17 +3,18 @@ use clap::{Parser, Subcommand};
...
 
+use crate::cli::run::RunCommand;
 use crate::defaults;
 
 mod config;
 mod init;
 mod relay;
 mod run;
 
+use calimero_version::CalimeroVersion;
```
miraclx: move to line 5

crates/merod/src/cli.rs
```diff
@@ -3,17 +3,18 @@ use clap::{Parser, Subcommand};
 use const_format::concatcp;
 use eyre::Result as EyreResult;
 
+use crate::cli::run::RunCommand;
```
miraclx: remove

crates/merod/src/cli.rs
```diff
@@ -3,17 +3,18 @@ use clap::{Parser, Subcommand};
...
 mod config;
 mod init;
 mod relay;
 mod run;
 
+use calimero_version::CalimeroVersion;
 use config::ConfigCommand;
 use init::InitCommand;
 use relay::RelayCommand;
-use run::RunCommand;
```
miraclx: restore

crates/version/Cargo.toml
miraclx: still pending

crates/version/build.rs
miraclx: still pending

rust-toolchain.toml
```diff
@@ -1,5 +1,6 @@
 [toolchain]
-channel = "1.85.0"
+# channel = "1.85.0"
+channel = "stable"
```
miraclx: you can revert this

miraclx (commented):
nearly there

crates/merod/src/cli.rs
dotandev: it has implementations within the CLI.

crates/merod/src/cli.rs
dotandev: Oh, I get it now.

crates/merod/Cargo.toml
miraclx: I'm not sure what more needs to be said about this one that hasn't already been said.

there are 3 spaces at the end of this line, remove them, that's it

crates/version/Cargo.toml
miraclx: this version should not be 0.1.0, it should be whatever version was previously in `merod/Cargo.toml`

after which you can make the version in `merod` and `meroctl` `version.workspace = true` like every other crate

so only the calimero-version crate specifies a version

crates/version/build.rs
miraclx: this was a question but

```suggestion
    let rustc_version = rustc_version::version()
```

crates/meroctl/src/cli.rs
```diff
@@ -20,8 +20,8 @@ mod peers;
 mod proxy;
 
 use app::AppCommand;
-use call::CallCommand;
 use calimero_version::CalimeroVersion;
```
miraclx: move to line 10

https://github.com/calimero-network/core/blob/master/STYLE.md#organize-imports-according-to-the-stdexternalcrate-pattern

dotandev: https://github.com/Nathy-bajo/core/blob/213853b80714ef77dbccceb06d9cba480981e854/.github/workflows/release.yml#L61


this says "not found"

maybe private?

miraclx: https://github.com/calimero-network/core/blob/477610eca6ef4bc30939fe16ddf4c39653f70f4b/.github/workflows/release.yml#L61

dotandev: and it should be line 4 by that styling, not 10 right?

.github/workflows/release.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           
           version_candidate=""
           for binary in $(echo '${{ steps.setup_matrix.outputs.binary_matrix }}' | jq -r '.[]'); do
-            binary_version=$(cargo metadata --format-version 1 --no-deps | jq -r --arg binary "$binary" '.packages[] | select(.name == $binary) | .version')
+            binary_version=$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "calimero-version") | .version')
```
miraclx: we don't need to iterate anymore

crates/meroctl/src/cli.rs
miraclx: `calimero_version::` is an external crate (should be in line 10), `call::` is not (keep in line 23)

Issue #339, opened 2024-05-28
Pull out apps into separate repo(s)


miraclx:
Pull out `kv-store`, `only-peers` and `rock-paper-scissors` into separate repo(s), each with it's associated frontend client.

MatejVukosav:
bump

frdomovic:
1. Kv-Store is left in core for e2e purposes but has new repo containing logic and FE
2. Only-Peers removed
3. Rock-paper-scissors removed
4. Gen-Ext + Visited moved to new repo

This can be closed after [this](https://github.com/calimero-network/core/pull/1270) is merged to master

PR #1262, opened 2025-05-15
WIP feat(apps): remove KV-store files

frdomovic:
# feat(apps): remove KV-store files

## Description

Removing KV Store files.

TBD: removing kv-store from e2e tests or changing location of app

## Test plan

## Documentation update


frdomovic:
> > So, this is a tricky one, due to it's use in `e2e-tests`. Plus, it's like a minimal app example to use in development & testing quickly.
> > I'm inclined to leave it for now. wdyt?
> > cc: @fbozic
> > at the point we decide to fulfil the remaining tasks of the [issue](https://github.com/calimero-network/core/issues/339) (frontend, etc..) should we choose to, we can maintain a separate version of kv-store suited to an application that is meant to be used in that way
> 
> I think it makes sense to keep it here as it is. It is minimal example and I often use it for manual testing as well. And it is simpler if it is in the same repo

Agree with @miraclx also, as I already created a separate repo for app example with FE we can leave this as actually.

frdomovic:
Closing this PR as per agreement.

PR #1268, opened 2025-05-19
feat: pull prebuilt node-ui assets on server build

miraclx:
## Description

One-half of the resolution to https://github.com/calimero-network/core/issues/1073.

This patches `calimero-server` to pull `node-ui` from https://github.com/calimero-network/admin-dashboard at build time.

Making it available just in time to be embedded into the final binary.

Related: https://github.com/calimero-network/core/pull/1252, which removes `node-ui` from the repo.

## Test Plan

It compiles

## Documentation Update

None needed


frdomovic (approved):
LGTM, Merged previous PR to master

.github/workflows/ci-checks.yml
```diff
@@ -25,6 +25,8 @@ jobs:
           key: ci-checks
 
       - name: Build
+        env:
+          CALIMERO_WEBUI_FETCH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
fbozic: I am not sure if the default token has permission to interact with other repos, we can set up GitHub app for that.

miraclx: It's a public repo, so it's fine, even unauthenticated works too, but is ratelimited to hell. So the token alleviates that restriction.

frdomovic:
bump

miraclx:
There's some weird issues with embedding the node-ui after the download, linux-aarch64 fails to build as somehow it misinterpretes the sections of the binary containing the embedding as a diff architecture, leading to a link error.. Still investigating but there are other higher-priority tasks.

alenmestrov (approved):

Issue #1096, opened 2025-02-08
Error syncing up the nodes


iamaa0205:
## System Details (please complete the following information)

- OS Version: macOS 15.2
- Processor: M1
- Architecture: arm64

- Browser : chrome

## How are you running the node?
   Tried Both
- [.] merod binary _(obtained with homebrew or installation script)_
- [.] building source code _(cloned github repository)_

> **Note:** If you are building from the source make sure you are using latest
> master

## Which tools versions you have?

- merod: 0.3.1
- meroctl: 0.3.1

## Describe the issue

We created a context with a node, then invited another node into the context. The invitation was successful but the data in the nodes is not syncing up means the root hash of both the nodes for the same context is different.

## How can we reproduce the issue

1. merod -n host init --server-port 2500 --swarm-port 2600
2. merod -n host run
3. install application to the host node
4. Create a new context 
5. Create another node with similiar commands changing server-port and swarm-port
6. Send an invitation to the new node and accept the invite from that node

## Expected behavior

The new node's root hash should be similiar to that of the host node for the same context.

## Screenshots

<img width="884" alt="Image" src="https://github.com/user-attachments/assets/b7969013-2470-4af7-ad6b-5a4cea161c55" />

<img width="880" alt="Image" src="https://github.com/user-attachments/assets/4d45689b-5de3-4d3a-9baa-220fdae8521a" />

## Additional context

Also consider the error that pops up in the merod terminal 
```ERROR libp2p_mdns::behaviour::iface: error sending packet on iface address No route to host (os error 65) address=10.81.50.254```

cy4n1d3-p1x3l:
we deleted the rust cache and installed from scratch, then some nodes ending up syncing but still all of them could not be brought to the same state.

chefsale:
@miraclx @xilosada can we prioritize taking look into this?

iamaa0205:
I posted this issue while working on the project: https://dorahacks.io/buidl/22471 for Calimero x ICP hackathon.
So yeah.

xilosada:
@iamaa0205 @cy4n1d3-p1x3l Thank you for bringing this to our attention. The dev team will prioritize looking into this issue and we'll keep you posted on any developments. We appreciate your patience and contribution to the project.

chefsale:
@xabilosada whats the status on this one?

chefsale:
Bump. @xabilosada @miraclx

miraclx:
[...]

PR #1271, opened 2025-05-20
fix(node): init for apps

frdomovic:
# fix(node): init for apps

## Description

On installation of new apps which don't have any changes, throws error for all methods and unable to fix from UI / sdk.
Screenshot 2025-05-20 at 11 43 45

cc @miraclx propose other solutions if known please.

## Test plan
Can do set and get methods from apps

Screenshot 2025-05-20 at 11 43 59

## Documentation update
-

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -465,9 +465,10 @@ impl Node {
             return Err(CallError::ContextNotFound);
         };
 
-        if method != "init" && &*context.root_hash == &[0; 32] {
-            return Err(CallError::Uninitialized);
-        }
+        // TODO: remove this once we have a logic init method
+        // if method != "init" && &*context.root_hash == &[0; 32] {
+        //     return Err(CallError::Uninitialized);
+        // }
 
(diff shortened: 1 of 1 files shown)


crates/node/src/lib.rs
```diff
@@ -465,9 +465,10 @@ impl Node {
             return Err(CallError::ContextNotFound);
         };
 
-        if method != "init" && &*context.root_hash == &[0; 32] {
-            return Err(CallError::Uninitialized);
-        }
```
fbozic, line 470: I think this should stay as it is; we probably want to call the init method when we are installing the app. I thought that was implemented. Let me double-check the code.

miraclx: it only runs on context creation, which is the intended behaviour

This issue is due to the fact that the application in use doesn't write to a crdt state object (`UnorderedMap`, `UnorderedSet`, `Vector`)

See: https://github.com/calimero-network/core/issues/1096#issuecomment-2814125948

PR #1276, opened 2025-05-23
feat(meroctl): support alias listing

### LINKED ISSUE #1117: Add ability to list aliases

cy4n1d3-p1x3l:
# [meroctl] alias listing support

## Description

Meroctl now has alias listing support. fixes #1117 

## Test plan

Not done yet.

## Documentation update

N/A


crates/meroctl/src/cli/context/identity/alias.rs
```diff
@@ -86,6 +88,17 @@ pub enum ContextIdentityAliasSubcommand {
...
+
+    #[command(about = "List all the aliases under the context")]
+    List {
+        #[arg(help = "The context whose aliases need to be listed")]
+        #[arg(long, short, default_value = "default")]
+        context: Alias<ContextId>,
+
+        #[arg(help = "If all the aliases for all the contexts need to listed")]
+        #[arg(default_value = "false", conflicts_with = "context")]
+        all: bool,
```
miraclx: `long`?

crates/server/src/admin/service.rs
miraclx, line 1: see [`alias.rs`](https://github.com/calimero-network/core/blob/master/crates/server/src/admin/handlers/alias.rs), follow the convention established

miraclx (commented):
`meroctl context list` is missing

cy4n1d3-p1x3l:
> `meroctl context list` is missing

isn't that defined [here](https://github.com/calimero-network/core/blob/master/crates/meroctl/src/cli/context/list.rs)?

crates/meroctl/src/cli/context/identity/alias.rs
```diff
@@ -87,6 +89,17 @@ pub enum ContextIdentityAliasSubcommand {
...
+
+    #[command(about = "List all the aliases under the context")]
+    List {
+        #[arg(help = "The context whose aliases need to be listed")]
+        #[arg(long, short, default_value = "default")]
+        context: Alias<ContextId>,
+
+        #[arg(help = "If all the aliases for all the contexts need to listed")]
+        #[arg(long, default_value = "false", conflicts_with = "context")]
+        all: bool,
```
miraclx: `--all` was supposed to be meant for `context identity list`, we can remove it for `context identity alias list`, should they need all aliases across all contexts, a little shell scripting won't hurt

crates/meroctl/src/cli/context/identity/alias.rs
```diff
@@ -87,6 +89,17 @@ pub enum ContextIdentityAliasSubcommand {
         #[arg(long, short, default_value = "default")]
         context: Alias<ContextId>,
     },
+
+    #[command(about = "List all the aliases under the context")]
+    List {
```
miraclx: alias = "ls"

crates/meroctl/src/cli/context/identity/alias.rs
miraclx: after you do this, could you update `merod`'s `context identity alias list` to not be `Option`? it should use `default_value` instead, thanks!

crates/server/primitives/src/admin.rs
```diff
@@ -253,6 +253,33 @@ impl GetContextIdentitiesResponse {
...
+}
+
+impl GetContextAliasesResponse {
+    pub const fn new(aliases: Vec<ContextAlias>) -> Self {
+        Self {
+            data: ContextAliasesResponseData { aliases },
+        }
+    }
+}
+
```
miraclx: we won't only be listing identity aliases though, we need to list contexts aliases too

see how we make this accomodation for creating contexts:

https://github.com/calimero-network/core/blob/d7ddaf8562379538647a90e2271194c1d5bf336a/crates/server/primitives/src/admin.rs#L443

crates/server/src/admin/service.rs
```diff
@@ -136,6 +137,10 @@ pub(crate) fn setup(
             "/contexts/:context_id/identities-owned",
             get(get_context_identities::handler),
         )
+        .route(
+            "/contexts/:context_id/aliases",
+            get(get_context_aliases::handler),
+        )
```
miraclx: please follow convention, all alias related functionality should be in alias.rs as [mentioned](https://github.com/calimero-network/core/pull/1276#discussion_r2108136761)

miraclx (commented):
> > meroctl context list is missing
>
> isn't that defined [here](https://github.com/calimero-network/core/blob/master/crates/meroctl/src/cli/context/list.rs?rgh-link-date=2025-05-27T15%3A21%3A59Z)?

I meant `meroctl context alias list`

crates/meroctl/src/cli/context/identity/alias.rs
```diff
@@ -185,6 +194,48 @@ impl ContextIdentityAliasCommand {
...
+                    c2 = "Alias",
+                )));
+
+                for AliasRecord { alias, value } in ctx_response.data {
+                    environment.output.write(&ErrorLine(&format!(
+                        "{c1:44} | {c2}",
+                        c1 = value,
+                        c2 = alias,
+                    )));
+                }
```
miraclx: why not try impl `ListAliasesResponse<PublicKey>: Report`

crates/server/primitives/src/admin.rs
```diff
@@ -253,6 +253,25 @@ impl GetContextIdentitiesResponse {
...
+pub struct AliasRecord<T> {
+    pub alias: Alias<T>,
+    pub value: T,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ListAliasesResponse<T> {
+    pub data: Vec<AliasRecord<T>>,
+}
```
miraclx: isn't this a map?

crates/server/src/admin/handlers/alias/list_alias.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use crate::admin::service::{parse_api_error, ApiResponse};
+use crate::AdminState;
+
+pub async fn handler<T>(
+    Extension(state): Extension<Arc<AdminState>>,
+    maybe_scope: Option<Path<T::Scope>>,
+) -> impl IntoResponse
+where
+    T: Aliasable + Serialize + Clone + From<[u8; 32]>,
```
miraclx: can't find why `Clone` is required here, any hints?

crates/server/src/admin/handlers/alias.rs
```diff
@@ -1,10 +1,11 @@
-use axum::routing::{post, Router};
+use axum::routing::{get, post, Router};
 use calimero_primitives::application::ApplicationId;
 use calimero_primitives::context::ContextId;
 use calimero_primitives::identity::PublicKey;
 
 mod create_alias;
 mod delete_alias;
+mod list_alias;
```
miraclx:
```suggestion
mod list_aliases;
```

crates/server/primitives/src/admin.rs
cy4n1d3-p1x3l: using map we won't be able to store duplicate aliases as keys, also currently we can have multiple aliases for the value?

miraclx: > using map we won't be able to store duplicate aliases as keys

why would we have duplicate aliases as keys? you're either listing context aliases, application aliases or context identity aliases, all of which are unique

> also currently we can have multiple aliases for the value?

that.. won't change with a `BTreeMap<Alias<T>, T>`

crates/meroctl/src/cli/context/identity/alias.rs
```diff
@@ -222,19 +222,8 @@ impl ContextIdentityAliasCommand {
...
-                )));
 
-                for AliasRecord { alias, value } in ctx_response.data {
-                    environment.output.write(&ErrorLine(&format!(
-                        "{c1:44} | {c2}",
-                        c1 = value,
-                        c2 = alias,
-                    )));
-                }
+                environment.output.write(&ErrorLine(&ctx_response.report()));
```
miraclx: this should've been

```suggestion
                environment.output.write(&ctx_response);
```

but better yet, why not use `make_request` over `do_request` which does that for you?

crates/server/primitives/src/admin.rs
```diff
@@ -272,6 +272,25 @@ impl<T> ListAliasesResponse<T> {
...
+        writeln!(output, "{c1:44} | {c2}", c1 = "Identity", c2 = "Alias").unwrap();
+
+        for AliasRecord { alias, value } in &self.data {
+            writeln!(output, "{c1:44} | {c2}", c1 = value, c2 = alias).unwrap();
+        }
+
+        output
+    }
+}
+
```
miraclx: this should be in meroctl, and it should use the same table format we use everywhere else

crates/meroctl/src/cli/context/identity/alias.rs
```diff
@@ -185,6 +194,37 @@ impl ContextIdentityAliasCommand {
...
+                )
+                .await?;
+
+                environment.output.write(&ErrorLine(&format!(
+                    "{c1:44} : {c2:44}",
+                    c1 = "Context ID",
+                    c2 = context_id,
+                )));
+
+                environment.output.write(&ErrorLine(&ctx_response.report()));
```
miraclx:
```suggestion
                make_request(
                    &client,
                    multiaddr_to_url(
                        multiaddr,
                        &format!("/admin-api/dev/aliases/identity/{context_id}"),
                    )?,
                    None::<()>,
                    &config.identity,
                    RequestType::Get,
                )
                .await?;
```

miraclx (commented):
we're still missing `meroctl context alias list`

crates/meroctl/src/common.rs
```diff
@@ -429,3 +430,22 @@ where
 
     Ok(ResolveResponse { alias, value })
 }
+
+impl<T: fmt::Display + ScopedAlias + Eq> Report for ListAliasesResponse<T> {
```
miraclx:
```suggestion
impl<T: fmt::Display> Report for ListAliasesResponse<T> {
```

crates/server/primitives/src/admin.rs
```diff
@@ -255,39 +257,19 @@ impl GetContextIdentitiesResponse {
...
-
-#[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub struct ListAliasesResponse<T> {
-    pub data: Vec<AliasRecord<T>>,
+pub struct ListAliasesResponse<T>
+where
+    T: ScopedAlias + Eq,
+{
+    pub data: HashMap<Alias<T>, T>,
```
miraclx: let's make this a `BTreeMap` instead

crates/server/src/admin/handlers/alias/list_aliases.rs
```diff
@@ -1,37 +1,39 @@
...
 use crate::AdminState;
 
 pub async fn handler<T>(
     Extension(state): Extension<Arc<AdminState>>,
     maybe_scope: Option<Path<T::Scope>>,
 ) -> impl IntoResponse
 where
-    T: Aliasable + Serialize + Clone + From<[u8; 32]>,
-    T::Scope: Copy + PartialEq + StoreScopeCompat,
+    T: Aliasable<Scope: StoreScopeCompat + Copy + PartialEq> + Serialize + From<[u8; 32]> + Eq,
```
miraclx: when associated type bounds are more than 1 and the bounded type also has trait constraints, breaking them up is advisable

miraclx (commented):
you can relax your `PartialEq` requirement by manually implementing `Eq` & `PartialEq` on `Alias<T>` where `T: !{Partial,}Eq`, and since you're changing it to a `BTreeMap` we need a manual impl of `{Partial,}Ord` as well

crates/server/src/admin/handlers/alias/list_aliases.rs
```diff
@@ -1,37 +1,39 @@
...
 {
     let scope = maybe_scope.map(|Path(s)| s);
 
     let aliases_raw = match state.node_client.list_aliases::<T>(scope) {
         Ok(a) => a,
-        Err(err) => return parse_api_error(err).into_response(),
+        Err(err) => return (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response(),
```
miraclx: `parse_api_error` does exactly this, why unroll it?

alenmestrov (approved):

PR #1280, opened 2025-05-26
Feature/precompiled wasm binaries

### LINKED ISSUE #189: Pre-compiled WASM binaries
Store the pre-compiled WASM binaries instead of doing it at runtime every single time

- [ ] When an app is installed, compile it and store both the wasm and the precompiled version in the blob store.
  - [ ] The `ApplicationId` record in the store will need two Blob IDs. One for the WASM, and another for the precompiled blob.
- [ ] When you execute a transaction, load the precompiled version.
- [ ] When you help others with a catchup, share the wasm.

Wasmer has support for dumping / loading precompiled wasm blobs, so we'll use that.

We cannot share the precompiled version due to potential architecture differences between nodes.

onyedikachi-david:
# [performance] Implement mandatory pre-compiled WASM binaries

## Description

Implements pre-compiled WASM binaries to improve application execution performance by 40-50%. All applications are now precompiled during installation, eliminating runtime compilation overhead.

**Key Changes:**
- Made `precompiled_blob` required in `ApplicationMeta`
- Added `compile_and_serialize()` and `run_precompiled()` to runtime engine
- Updated installation to always precompile WASM modules
- Modified execution to use precompiled modules with fallback

Fixes #189.

## Test plan

- All runtime tests pass (11/11)
- Full workspace compiles successfully
- Tested installation, execution, and fallback scenarios

```bash
cargo test --package calimero-runtime
```


onyedikachi-david:
@miraclx  what do you mean by this: 

> "precompilation` is an implementation detail

miraclx:
> "precompilation` is an implementation detail

1. logs don't need to include that detail
2. we don't need specific APIs to make precompilation happen, it's just a regular compilation
3. it's "precompiled" because it was compiled at some previous point
4. the only cases that need specificity on precompilation are `Engine::from_precompiled` and the `Application` storage & fetching, anything else is unnecessary

onyedikachi-david:
Okay

crates/context/src/handlers/create_context.rs
```diff
@@ -240,17 +240,14 @@ async fn create_context(
...
-        &guard,
-        identity,
-        module,
-        "init".into(),
-        init_params.into(),
-        storage,
-    )
-    .await?;
+    let mut storage_mut = storage;
+    let outcome = module.run(context.id, identity, "init", &init_params, &mut storage_mut)?;
```
miraclx: since both execute and create_context need either the executable blob or the bytecode, you can share a function between them

crates/context/src/handlers/execute.rs
```diff
@@ -286,11 +286,61 @@ async fn internal_execute(
         });
     };
 
+    let precompiled_bytes = node_client
+        .get_precompiled_application_bytes(&context.application_id)
+        .await?;
```
miraclx: we don't want to pre-emptively read both the raw bytes and the precompiled version, get the precompiled version and fallback when it fails

crates/context/src/handlers/execute.rs
```diff
@@ -286,11 +286,61 @@ async fn internal_execute(
...
+            Err(_) => {
+                debug!("Cached compiled module is incompatible, compiling fresh");
+                let module = engine.compile(&blob)?;
+
+                if let Ok(serialized) = module.to_bytes() {
+                    if let Ok((blob_id, _)) = node_client
+                        .add_blob(serialized.as_ref(), Some(serialized.len() as u64), None)
+                        .await
+                    {
+                        drop(
```
miraclx: prefer `let _ignored` to `drop` since your editor will likely tell you the type you're dropping

in this case, `update_precompiled_blob` returns a `eyre::Result<()>`, you shouldn't be ignoring here at all

crates/node/primitives/src/client/application.rs
```diff
@@ -172,4 +199,24 @@ impl NodeClient {
 
         Ok(applications)
     }
+
+    pub async fn update_precompiled_blob(
+        &self,
+        application_id: &ApplicationId,
+        precompiled_blob_id: BlobId,
```
miraclx:
```suggestion
        blob_id: &BlobId,
```

it's already in the function name

crates/node/primitives/src/client/application.rs
```diff
@@ -172,4 +199,24 @@ impl NodeClient {
 
         Ok(applications)
     }
+
+    pub async fn update_precompiled_blob(
```
miraclx:
```suggestion
    pub fn update_precompiled_blob(
```

crates/node/primitives/src/client/application.rs
```diff
@@ -126,20 +152,21 @@ impl NodeClient {
...
-                    .into_async_read(),
-                expected_size,
+                wasm_bytes.as_ref(),
+                Some(wasm_bytes.len() as u64),
                 expected_hash,
             )
             .await?;
 
         self.install_application(&blob_id, size, &uri, metadata)
+            .await
```
miraclx: you can revert these now

crates/node/primitives/src/client/application.rs
```diff
@@ -69,7 +93,7 @@ impl NodeClient {
         Ok(false)
     }
 
-    fn install_application(
+    async fn install_application(
```
miraclx:
```suggestion
    fn install_application(
```

crates/context/src/handlers/execute.rs
miraclx: PS: prefer directly calling `get_blob_bytes` so we don't have to do a db lookup on every execution

see `ContextManager::get_or_fetch_context`

and you can make `ContextMeta` this:

crates/node/primitives/src/client/application.rs
```diff
@@ -57,6 +55,32 @@ impl NodeClient {
...
+
+        let Some(bytes) = self
+            .get_blob_bytes(&application.precompiled_blob.blob_id())
+            .await?
+        else {
+            bail!("fatal: application points to dangling precompiled blob");
+        };
+
+        Ok(Some(bytes))
+    }
```
miraclx: in retrospect, we should never need this function.. the same way we never use `get_application_bytes`

crates/runtime/Cargo.toml
miraclx, line 1:
```master
git checkout master Cargo.toml crates/runtime
```

crates/context/src/handlers/create_context.rs
```diff
@@ -240,17 +240,14 @@ async fn create_context(
...
+
+    let (outcome, storage) = (outcome, storage_mut);
```
miraclx: why this pattern? `storage` is defined 3 statements earlier, and can be moved closer if you wanted to localize the `mut` declaration, a new variable is not needed, and I don't understand `outcome = outcome`

crates/node/primitives/src/client/application.rs
```diff
@@ -130,10 +131,7 @@ impl NodeClient {
 
         let (blob_id, size) = self
             .add_blob(
-                response
-                    .bytes_stream()
-                    .map_err(io::Error::other)
-                    .into_async_read(),
+                StreamReader::new(response.bytes_stream().map_err(std::io::Error::other)).compat(),
```
miraclx: I don't expect any change here

crates/node/primitives/src/client/application.rs
```diff
@@ -81,6 +81,7 @@ impl NodeClient {
             size,
             source.to_string().into_boxed_str(),
             metadata.into_boxed_slice(),
+            key::BlobMeta::new(BlobId::from([0; 32])),
         );
 
         let application_id = ApplicationId::from(*Hash::hash_borsh(&application)?);
```
miraclx: see this line (#&NoBreak;87)... it's very important

now that you've added `precompiled_blob` to `ApplicationMeta`, it's being factored into the Application ID which should be consistent regardless of which platform you're on, but we cannot guarantee the same blob ID for the precompiled module on varying platforms.

so we're going to have to manually construct some other representative structure for the old behaviour of `ApplicationMeta`, being explicit.

you can try a tuple `(application.blob, application.size, &application.source, &application.metadata)`

so this becomes

```suggestion
        let application_id = {
            let components = (application.blob, application.size, &application.source, &application.metadata);
            ApplicationId::from(*Hash::hash_borsh(&components)?)
        };
```

crates/context/src/lib.rs
```diff
@@ -103,11 +104,20 @@ impl ContextManager {
                     return Ok(None);
                 };
 
+                let handle = self.datastore.handle();
+                let app_key = calimero_store::key::ApplicationMeta::new(context.application_id);
+                let compiled_blob = if let Some(app_meta) = handle.get(&app_key)? {
+                    app_meta.precompiled_blob.blob_id()
+                } else {
+                    BlobId::from([0; 32])
+                };
```
miraclx: `.bytecode` and `.compiled` should carry on to `calimero_primitives::application::Application`

you can either keep it on the struct itself, or change the existing `.blob` field to a new `ApplicationBlob { .bytecode, .compiled }`, which might be preferable

crates/context/src/handlers/update_application.rs
```diff
@@ -38,7 +38,8 @@ impl Handler<UpdateApplicationRequest> for ContextManager {
 
         ActorResponse::r#async(task.into_actor(self).map_ok(move |blob_id, act, _ctx| {
             if let Some(context) = act.contexts.get_mut(&context_id) {
-                context.blob = blob_id;
+                context.bytecode = blob_id;
+                context.compiled = BlobId::from([0; 32]);
```
miraclx: at this point, we should have the compiled blob ID if there is one.

the `update_application_id` should return `ApplicationBlob`

crates/context/src/handlers/execute.rs
```diff
@@ -273,24 +318,27 @@ async fn internal_execute(
...
 
     let storage = ContextStorage::from(datastore, context.id);
 
-    let module = engine.compile(&blob)?;
+    let mut storage_mut = storage;
+
+    let outcome = module.run(**guard, executor, &method, &input, &mut storage_mut)?;
 
-    let (outcome, storage) = execute(guard, executor, module, method, input, storage).await?;
+    let (outcome, storage) = (outcome, storage_mut);
```
miraclx: 1. `execute` should happen inside the global, multi-threaded tokio runtime, not in the single-threaded actix one (this is a blocking operation) this is what the `execute` method does
2. what is this pattern you keep doing?

crates/context/src/lib.rs
```diff
@@ -21,7 +21,8 @@ pub mod handlers;
 #[derive(Debug)]
 struct ContextMeta {
     meta: Context,
-    blob: BlobId,
+    bytecode: BlobId,
+    compiled: BlobId,
```
miraclx, line 26: We forgot to account for multiple contexts sharing the same bytecode, let's revert this line and add a new map to context manager to associate them

`modules: Map<BlobId, BlobId>`

where the key is the bytecode, and the value is the precompiled module

crates/context/src/handlers/execute.rs
```diff
@@ -32,6 +32,51 @@ pub mod storage;
...
+            {
+                if let Err(err) = node_client.update_precompiled_blob(application_id, &blob_id) {
+                    debug!("Failed to update precompiled blob: {}", err);
+                }
+            }
+        }
+    }
+
+    Ok(module)
+}
```
miraclx: this should be an associated method on `ContextManager`, much like `get_or_fetch_context`

the signature can be:

```rust
fn get_module(&self, blob_id: &BlobId) -> eyre::Result<calimero_runtime::Module>
````

File: crates/context/src/handlers/create_context.rs [hunk 17]
```diff
@@ -228,30 +230,29 @@ async fn create_context(
...
+        &node_client,
+        &engine,
+        &application.id,
+        &compiled_blob_id,
+        false,
     )
     .await?;
 
+    let outcome = module.run(context.id, identity, "init", &init_params, &mut storage)?;
+
```
miraclx: much like execute, you can revert everything here, and then use `self.get_module`

crates/context/src/handlers/execute.rs
```diff
@@ -137,14 +182,14 @@ impl Handler<ExecuteRequest> for ContextManager {
         let execute_task = guard_task.then(move |guard, act, _ctx| {
             let context = act
                 .get_or_fetch_context(&context_id)
-                .map(|c| c.map(|c| (c.meta, c.blob)));
+                .map(|c| c.map(|c| (c.meta, c.bytecode, c.compiled)));
```
miraclx: we need an earlier prepare step before the `execute` calls `get_or_fetch_context` and `get_module`, returning `(Context, Module)`

crates/context/src/lib.rs
miraclx: kinda sucks we lose semantic information on what's what, since they're both `BlobId`

another approach is to have `BTreeSet<ApplicationBlob>`, so long as `ApplicationBlob` impls:

- `Ord` and `PartialOrd` on just the `bytecode` blob ID (not derive as that would factor in the compiled blob ID too, which we don't need)

- `Borrow<BlobId>` so we can `self.modules.get(&blob_id)` to get the `&ApplicationBlob`

or-- just `Map<BlobId, ApplicationBlob>` which works too 🤷🏽‍♂️

crates/context/src/handlers/execute.rs
onyedikachi-david: How about implementing it into the he execute handler's async block? Because reading/loading the blob `get_blob_bytes()` is async

onyedikachi-david: @miraclx Please I need your input on this...

miraclx: Hm, I'm not sure I understand the blocker? You could just make the function `async`.

crates/context/src/handlers/execute.rs
```diff
@@ -179,35 +133,96 @@ impl Handler<ExecuteRequest> for ContextManager {
...
+                error!(?err, "failed to prepare execution");
+                err
+            })
+            .into_actor(act)
+        });
+
+        let execute_task = prepare_task.then(move |prepared_result, act, _ctx| {
+            let node_client = act.node_client.clone();
+            async move {
+                let (guard, mut context, module, datastore) = prepared_result?;
```
miraclx, line 165: why did you inline `internal_execute` here?

crates/context/src/lib.rs
```diff
@@ -127,6 +126,32 @@ impl ContextManager {
     }
 }
 
+async fn get_module(
+    node_client: &NodeClient,
+    runtime_engine: &calimero_runtime::Engine,
+    modules: &BTreeMap<BlobId, ApplicationBlob>,
+    blob_id: &BlobId,
+) -> eyre::Result<calimero_runtime::Module> {
+    if let Some(app_blob) = modules.values().find(|ab| ab.bytecode == *blob_id) {
```
miraclx, line 135:
```suggestion
    if let Some(app_blob) = modules.get(&blob_id) {
```

miraclx (commented):
a new round of review would've resulted in too many changes, I'll merge as-is into a secondary branch where I'll drive this to completion.

https://github.com/calimero-network/core/pull/1291

PR #1282, opened 2025-05-26
chore(e2e): verify external state should be a view call

miraclx:
## Description

Thing

## Test plan

e2e-tests check out

## Documentation update

None needed

### DIFF
--- a/e2e-tests/src/protocol/near.rs
+++ b/e2e-tests/src/protocol/near.rs
@@ -121,9 +121,8 @@ impl NearSandboxEnvironment {
 
         let result = self
             .mock_external_contract
-            .call(method_name)
+            .view(method_name)
             .args(arguments)
-            .transact()
             .await?;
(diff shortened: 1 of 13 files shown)


e2e-tests/src/protocol/near.rs
```diff
@@ -121,9 +121,8 @@ impl NearSandboxEnvironment {
 
         let result = self
             .mock_external_contract
-            .call(method_name)
+            .view(method_name)
             .args(arguments)
-            .transact()
             .await?;
```
miraclx, line 126: @alenmestrov, can you check my thinking on this?

it struck me as odd while investigating e2e issues for the actor PR

alenmestrov: true, it should be view since we are checking for the state change

frdomovic (approved):

PR #1283, opened 2025-05-27
chore(ci): patches to description check

miraclx:
## Description

1. CI should fail on bad description (missing from #1222)
2. Support quotes and backticks, mitigating code injection
	- Case Study (ref):
	  

3. add debug lines, just in case

## Test plan

It works

## Documentation update

WYSIWYG


frdomovic (approved):
all good

PR #1145, opened 2025-03-10
chore: remove contracts code, cleanup scripts and e2e-tests

fbozic:

PR #1287, opened 2025-05-28
chore(CI): restore disk space maximization

miraclx:
## Description

from time to time, our disk usage grows beyond the free disk space available on the root fs in the GHA runner, however there are other filesystems available to the runner with much more space

https://github.com/calimero-network/core/pull/909 looked into this and was able to combine all of them into a new volume

after a while, our build disk usage dropped, enough to fit within the root disk, after which https://github.com/calimero-network/core/pull/1145 removed this accomodation

Building https://github.com/calimero-network/core/pull/1268 just seemed to trigger that issue [once again](https://github.com/calimero-network/core/actions/runs/15299718513/job/43037046120?pr=1268#step:8:909), leading to this patch reintroducing it.

## Test plan

cherry-picked into https://github.com/calimero-network/core/pull/1268 at 0aac299, if the run passes there and here, we're good to go

## Documentation update

None needed

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -7,12 +7,37 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_HOME: ''
+  RUSTUP_HOME: ''
+  ALT_HOME: ''
+  STAGE_DIR: ''
 
 jobs:
   rust:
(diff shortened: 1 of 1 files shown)


frdomovic (approved):
Looks good.

alenmestrov (approved):

PR #1289, opened 2025-05-29
fix(config): Clean up credentials parsing from config.

dotandev:
## Description

Attempts to resolve https://github.com/calimero-network/core/issues/1135.

This PR cleans up credential parsing in the config crate and also generates a default keypair if absent:

-It ensures that the system doesn't fall back to the relayer when local credentials are specified but missing.

-It simplifies the credential parsing logic, reducing the risk of incorrect parsing due to overlapping credential formats.

-It cleans up the configuration file by omitting redundant information.

## Test Plan

N/A

## Documentation Update
N/A


frdomovic (dismissed):
Looks good

crates/config/src/lib.rs
```diff
@@ -48,12 +52,9 @@ pub struct SyncConfig {
 #[non_exhaustive]
 pub struct NetworkConfig {
     pub swarm: SwarmConfig,
-
```
frdomovic: This should be reverted as it is spaced-out on purpose through the entire file, also `cargo +nightly fmt` won't remove these empty lines. cc @miraclx

dotandev: Alright, I will fix that.

miraclx:
With no indications of progress in resolving this issue, I'm closing this as stale

PR #1291, opened 2025-06-05
feat: precompiled wasm blobs

miraclx:
## Description

The first time an application is executed, we compile the wasm bytecode, and export it's serialized form, to be saved as a blob. We can then later load this saved blob to avoid recompilation on subsequent execution against the same application.

~nit: if 2 applications share the same blob, we cannot safely assume they share the same precompiled artifact, since the application may later impact the compilation process, which makes the precompiled artifact unique for each application, not each blob.

Successor to https://github.com/calimero-network/core/pull/1280 (thanks, @onyedikachi-david), resolves https://github.com/calimero-network/core/issues/189.

## Test plan

e2t-tests & manual testing

## Documentation update

None needed

### DIFF
--- a/crates/server/src/jsonrpc.rs
+++ b/crates/server/src/jsonrpc.rs
@@ -10,7 +10,7 @@ use calimero_server_primitives::jsonrpc::{
 };
 use calimero_store::Store;
 use serde::{Deserialize, Serialize};
-use tracing::{debug, error, info};
+use tracing::{debug, info};
 
 use crate::config::ServerConfig;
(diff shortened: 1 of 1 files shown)


frdomovic (approved):

frdomovic:
Manual testing steps to recreate?

miraclx:
> Manual testing steps to recreate? ; if relevant

spin up a node, install an app, create a context with it, call a method, observe how long it takes to run, call the method again, observe how long it takes, restart the node, call the method again, see how long it takes

PR #1294, opened 2025-06-11
chore(ci): patch release pipeline

miraclx:
### tl;dr

in PR:

- never release binaries
- every PR gets a `:pr-N` docker tag (this one being `:pr-1294`)
- every new commit gets a `:<sha>` docker tag

on master:

- release binaries if the version changed (a version suffixed with `-pre.N` is a prerelease)
- release a `:<version>` docker tag if the version changed
- every new commit gets an `:edge` docker tag
- every new commit gets a `:<sha>` docker tag

### long form

1. The release process was unnecessarily complicated with some duplication and conditions that would never be triggered.
2. GitHub also now provides Arm64 & x86_64 runners for [macOS](https://github.com/actions/runner-images#available-images) and [ubuntu](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/), so we don't need virtualized cross-compilation, it can happen natively.
3. making a pre-release cannot be dependent on something non-obvious, implicit and obscure as the branch name being `release/*`, now, you must manually set the version in [crates/version/Cargo.toml](https://github.com/calimero-network/core/blob/3f425eb50508cae9eec7f6d620132e69b6be2477/crates/version/Cargo.toml#L3) to `<semver>-pre.N` where `N` is some incremental number, the release will then happen **after the PR merges to master**.
4. publish docker images as commitments to commits, every commit to a PR or master resuts in equivalent docker images tagged to the commit. example: `merod:f469a5e7`, `meroctl:f469a5e7`
5. publish docker images as commitments to PRs, every update to a PR subsequently updates the same tag. example: `pr-1294`

todo: impl cron job to clean docker images for;
- commit refs older than 1 day - `meroctl:f469a5e7`, `merod:f469a5e7`
- `pr-X` older than 1 week - `meroctl:pr-1294`, `merod:pr-1294`
- semver never - `meroctl:0.6.0`, `merod:0.6.0`, `meroctl:0.6.0-pre-2`, `merod:0.6.0-pre-2`
- `latest` never - `meroctl:latest`, `merod:latest`

## Test plan

Release run succeeds

## Documentation update

--

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -39,8 +39,7 @@ jobs:
       - name: Setup Rust Cache
         uses: Swatinem/rust-cache@v2
         with:
-          cache-on-failure: true
-          cache-all-crates: true
+          key: e2e-tests
 
       - name: Build Apps & Binaries
         run: |
--- a/.github/workflows/pr-description-check.yml
+++ b/.github/workflows/pr-description-check.yml
@@ -14,7 +14,7 @@ permissions:
   contents: read
 
 env:
-  REQUIRED_SECTIONS: |
+  REQUIRED_SECTIONS: |-
     ## Description
     ## Test Plan
     ## Documentation Update
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+defaults:
+  run:
+    shell: zsh
+
 env:
   BINARIES: |-
     merod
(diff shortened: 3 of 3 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -39,8 +39,7 @@ jobs:
       - name: Setup Rust Cache
         uses: Swatinem/rust-cache@v2
         with:
-          cache-on-failure: true
```
fbozic, line 42: This comes quite handy actually, since usually failures are in e2e tests, not in the build process. Or maybe now we don't need it anymore since the build is in a dedicated job?

miraclx: now the test is broken up into separate jobs, so it's guaranteed this build job will have already submitted it's cache, so it's not a concern

PR #1277, opened 2025-05-25
feat: Implement merod devnet command for local network management

Nathy-bajo:
## Description

Attempts to resolve  #1186

This PR introduces a new merod devnet command that allows developers to easily spin up and manage local networks for development and testing purposes. The implementation:

- Extracted common devnet functionality from e2e-tests into new calimero-sandbox crate

- Implemented merod devnet command with start/stop/status subcommands

- Updated e2e-tests to use new devnet crate instead of custom implementation

- Added protocol configuration and initialization support

- Improved devnet information printing with node addresses and protocol details

## Test plan

# Start a devnet with default config
merod devnet start --binary ./target/debug/merod

# Start with custom config
merod devnet start --config custom-config.json --binary ./merod --logs-dir ./logs

# Check status
merod devnet status

# Stop the network
merod devnet stop

PR #1302, opened 2025-06-14
fix: conditional connection to unshackle `meroctl node`

miraclx:
## Description

1. Remove `meroctl bootstrap`, it wasn't useful and it'll soon be replaced by `merod sandbox` https://github.com/calimero-network/core/pull/1277.
2. `meroctl node` does not need `--node` or `--api` but they were required anyway.
3. `meroctl node` was undocumented.

## Test plan

--

## Documentation update

--


alenmestrov (approved):

crates/meroctl/src/cli.rs
```diff
@@ -158,13 +139,13 @@ impl RootCommand {
         Ok(())
     }
 
-    async fn prepare_connection(&self) -> eyre::Result<ConnectionInfo> {
+    async fn prepare_connection(&self) -> eyre::Result<Option<ConnectionInfo>> {
```
fbozic, line 142: Why are we moving to `Result<Option<ConnectionInfo>>` from `Result<ConnectionInfo>`? Due to this?

```
// todo! if neither is selected, we should load the "default" config
            _ => return Ok(None),
```

Wouldn't it make more sense to return error here?

miraclx: `meroctl node` doesn't need a connection, if we returned an error it would, which was the previous behaviour.

PR #1307, opened 2025-06-16
feat: Permissions implementation

alenmestrov:
# [Auth] Implement permissions system for authentication gateway

## Description

This PR implements the core permissions system for the authentication gateway, providing a robust and flexible way to manage access control across different authentication providers and contexts.

Key components:
- Permission types and enums for granular access control
- Permission validation system for checking access rights
- Integration with existing storage and auth systems

Dependencies:
- Requires completed storage layer implementation
- Built on top of the basic server configuration

Changes include:
- `permissions/mod.rs`: Core permission system initialization and exports
- `permissions/types.rs`: Definition of permission types, roles, and access levels
- `permissions/validator.rs`: Permission validation and enforcement logic

## Test plan
N/A

## Documentation update
N/A


crates/auth/src/auth/permissions/types.rs
```diff
@@ -0,0 +1,599 @@
...
+pub enum KeyPermission {
+    All,
+    Create,
+    List,
+    Delete,
+}
+
+impl Permission {
+    /// Convert a permission string to a Permission enum
+    pub fn from_str(s: &str) -> Option<Self> {
```
miraclx: [`FromStr`](https://doc.rust-lang.org/nightly/std/str/trait.FromStr.html)

crates/auth/src/auth/permissions/types.rs
```diff
@@ -0,0 +1,599 @@
...
+                Some("delete") => Some(Permission::Keys(KeyPermission::Delete)),
+                _ => Some(Permission::Keys(KeyPermission::All)),
+            },
+            _ => None,
+        }
+    }
+
+    /// Convert a Permission enum to its string representation
+    pub fn to_string(&self) -> String {
+        match self {
```
miraclx: [`Display`](https://doc.rust-lang.org/nightly/std/fmt/trait.Display.html)

crates/auth/src/auth/permissions/types.rs
```diff
@@ -0,0 +1,599 @@
...
+                assert_eq!(context_id.unwrap(), "ctx-123");
+                assert!(user_id.is_none());
+            }
+            _ => panic!("Wrong permission type"),
+        }
+
+        // Test lookup permission with context and user
+        let perm = Permission::from_str("context:alias:lookup[ctx-123,user-456]").unwrap();
+        match perm {
+            Permission::Context(ContextPermission::Alias(AliasPermission::Lookup {
```
miraclx: this will be different from, for example, `context:leave` which from the logic here I'd expect to be represented as

```
context:leave[ctx-123,ctx-456]<user-123,user-456>
```

right?

it's worth noting that this means either of the identities can be used to operate in either of the contexts

crates/auth/src/auth/permissions/validator.rs
```diff
@@ -0,0 +1,598 @@
...
+            "DELETE" => HttpMethod::DELETE,
+            "PATCH" => HttpMethod::PATCH,
+            _ => HttpMethod::Any,
+        };
+
+        let mut required_permissions = Vec::new();
+
+        // First check path mappings
+        for mapping in PATH_MAPPINGS.iter() {
+            if let Ok(regex) = Regex::new(mapping.pattern) {
```
miraclx: since the regexes are static consider pre-compiling them on auth server startup, instead of on every request, you can call `Regex::new` in a `lazy_static`, and if you only want to compile on demand and reuse, use [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html)

crates/auth/src/auth/permissions/types.rs
alenmestrov: I have updated the code to match format:
permission_category:permission_action[ctx_id,user_id,method]

crates/auth/src/auth/permissions/validator.rs
```diff
@@ -1,189 +1,188 @@
...
 /// Permission validator for checking request permissions
 #[derive(Debug, Default)]
 pub struct PermissionValidator;
 
-/// Represents a path pattern and its required permissions
-struct PathMapping {
-    pattern: &'static str,
-    handler: fn(&[&str], HttpMethod) -> Vec<Permission>,
+macro_rules! route_permissions {
+    ($path:expr, $method:expr) => {{
```
miraclx: this macro is only used once? it should either be a method or inlined.

alenmestrov: switched to method

PR #1308, opened 2025-06-16
feat: Auth Providers

alenmestrov:
# [Auth] Implement authentication providers framework with NEAR wallet integration

## Description

This PR implements the complete authentication providers system, including both the core provider framework and the first provider implementation (NEAR wallet).

Key components:
1. Provider Framework:
   - Core provider traits and interfaces
   - Provider registry for managing multiple auth providers
   - Data registry for provider-specific storage
   - Macro system for simplified provider implementation

2. NEAR Wallet Implementation:
   - Complete NEAR wallet authentication flow
   - Account verification and signature validation
   - Integration with the provider framework

Files changed:
```crates/auth/src/providers/```
- `core/mod.rs`: Core provider system initialization
- `core/macros.rs`: Helper macros for provider implementation
- `core/provider.rs`: Base provider traits and interfaces
- `core/provider_registry.rs`: Provider registration and management
- `core/provider_data_registry.rs`: Provider data storage management
- `impls/mod.rs`: Provider implementations initialization
- `impls/near_wallet.rs`: NEAR wallet provider implementation
- `mod.rs`: Top-level provider system exports

Dependencies:
- Built on storage layer
- Requires permissions system
- Integrates with core auth components

## Test plan
N/A

## Documentation update
N/A

### CODE REVIEW

File: crates/auth/src/providers/impls/near_wallet.rs [hunk 1]
```diff
@@ -0,0 +1,773 @@
...
+            auth_data,
+        };
+
+        Ok(AuthRequestVerifier::new(verifier))
+    }
+
+    fn get_health_status(&self) -> eyre::Result<serde_json::Value> {
+        // Test the RPC connection to verify if the provider is healthy
+        // We're just checking if we can create a connection, but not actually making the request
+        let _client = JsonRpcClient::connect(&self.config.rpc_url);
```
miraclx: 1. `_client` isn't used at all here, we don't need the client?
2. if you had to reuse the client, why not store it in `NearWalletProvider`?

crates/auth/src/providers/impls/near_wallet.rs
```diff
@@ -0,0 +1,773 @@
...
+
+impl ProviderRegistration for NearWalletProviderRegistration {
+    fn provider_id(&self) -> &str {
+        "near_wallet"
+    }
+
+    fn create_provider(
+        &self,
+        context: ProviderContext,
+    ) -> Result<Box<dyn AuthProvider>, eyre::Error> {
```
miraclx, line 723:
```suggestion
    ) -> eyre::Result<Box<dyn AuthProvider>> {
```

crates/auth/src/providers/impls/near_wallet.rs
alenmestrov: I would leave it out for now, if it fails let it fail when checking the account, as for the provider implementation, I would leave it as is so that we don't bloat it since only provider that needs this is NEAR if I remember correctly

miraclx: yeah but what I meant was instead of calling

```rust
JsonRpcClient::connect(&self.config.rpc_url);
```

on every validation check, you can do that once on the near provider initialization and reuse the client, but this is a non-blocker so it's alright

miraclx (approved):
approving to unblock

PR #1309, opened 2025-06-16
feat: core auth implementation

alenmestrov:
# [Auth] Implement core authentication and security components

## Description

This PR implements the core authentication and security components for the authentication gateway, providing fundamental security features, middleware, and service layer functionality.

Key components:
- Security utilities and cryptographic functions
- Authentication middleware for request validation
- Core validation logic for auth flows
- Authentication service implementation
- Secrets management system

Files changed:
```crates/auth/src/```
- `auth/mod.rs`: Core authentication module initialization
- `auth/security.rs`: Security utilities and cryptographic functions
- `auth/middleware.rs`: Authentication middleware implementation
- `auth/validation.rs`: Request and token validation logic
- `auth/service.rs`: Core authentication service layer
- `secrets.rs`: Secrets management implementation

Dependencies:
- Built on server configuration layer
- Requires storage layer implementation
- Foundation for permissions and provider systems

## Test plan
N/A

## Documentation update
N/A

### CODE REVIEW

File: crates/auth/src/auth/middleware.rs [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+                    .iter()
+                    .any(|p| matches!(p, Permission::Context(ContextPermission::Execute(_, _, _))))
+                {
+                    required_permissions.push(Permission::Context(ContextPermission::Execute(
+                        ResourceScope::Global,
+                        UserScope::Any,
+                        None,
+                    )));
+                }
+            }
```
miraclx: why doesn't `determine_required_permissions` include the execute permission?

crates/auth/src/auth/middleware.rs
```diff
@@ -0,0 +1,157 @@
...
+///
+/// * `request` - The request to check
+/// * `next` - The next middleware
+///
+/// # Returns
+///
+/// * `Result<Response, (StatusCode, HeaderMap)>` - The response or error with headers
+pub async fn forward_auth_middleware(
+    state: Extension<Arc<AppState>>,
+    request: Request<Body>,
```
miraclx: why is this called `forward_auth_middleware` and documented as such? when we no longer have a reverse proxy implemented

unless I missed it

crates/auth/src/auth/security.rs
```diff
@@ -0,0 +1,194 @@
...
+use tower::{Layer, Service};
+
+use crate::config::{RateLimitConfig, SecurityHeadersConfig};
+
+/// Rate limiting state shared between middleware instances
+#[derive(Clone)]
+struct RateLimitState {
+    config: RateLimitConfig,
+    counters: Arc<DashMap<String, (u32, std::time::Instant)>>,
+}
```
miraclx, line 22: not sure I understand the rationale for implementing ratelimiting here

this can be a service in front of the gateway no?

crates/auth/src/auth/validation.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// HTML escape a string to prevent XSS
+pub fn escape_html(input: &str) -> String {
+    input
+        .replace('&', "&amp;")
+        .replace('<', "&lt;")
+        .replace('>', "&gt;")
+        .replace('"', "&quot;")
+        .replace('\'', "&#x27;")
+}
```
miraclx, line 88: are any of these used?

crates/auth/src/auth/middleware.rs
alenmestrov: resolved with permissions refactore

crates/auth/src/auth/validation.rs
alenmestrov: They should be used in api handlers, will check it out in the PR, in any case we will be using it.

crates/auth/src/auth/middleware.rs
alenmestrov: cleaned up the naming in next PR where its implemented

crates/auth/src/auth/security.rs
alenmestrov: Yes, but since we were going for completely standalone integration, I implemented this for cases that won't be using forward auth, and also we have endpoints that are accessible publicly

miraclx: But then that should either be the node or gateway's responsibility, all the auth server needs to do is authenticate

alenmestrov: We have on auth server also endpoints for key management, since they are stored on the server itself, that is why I have implemented this additional security level

miraclx (approved):

PR #1311, opened 2025-06-16
feat: API routes and their handlers

alenmestrov:
# [Auth] Implement API layer with handlers and routes

## Description

This PR implements the complete API layer for the authentication gateway, providing all necessary endpoints, handlers, and routing logic for authentication operations.
Connected missing pieces with rest of the implemementation

Key components:
- Route definitions and configuration
- Authentication handlers
- Permission management endpoints
- Root and client key management
- Request/response handling
- Error handling and responses

Files changed:
```crates/auth/src/api/```
- `mod.rs`: API module initialization
- `routes.rs`: Route definitions and configuration
- `handlers/mod.rs`: Handler initialization and common utilities
- `handlers/auth.rs`: Authentication flow handlers
- `handlers/permissions.rs`: Permission management endpoints
- `handlers/root_keys.rs`: Root key operations
- `handlers/client_keys.rs`: Client key management

Dependencies:
- Built on token management system
- Requires permissions system
- Uses provider implementations
- Integrates with storage layer
- Leverages core auth components

## Test plan
N/A

## Documentation update
N/A

### CODE REVIEW

File: crates/auth/src/api/handlers/mod.rs [hunk 1]
```diff
@@ -0,0 +1,220 @@
...
+use serde_json::json;
+use tracing::info;
+
+use crate::api::handlers::auth::success_response;
+use crate::server::AppState;
+
+/// Embed the contents of the auth frontend build directory into the binary
+#[derive(RustEmbed)]
+#[folder = "frontend/dist/"]
+struct AuthUiStaticFiles;
```
miraclx: would be nice if we end up fully embedding this in the webui? (admin dashboard) or at least an independent repo we can pull in at build time like we do for admin-dashboard

non-blocker, just curious if that is an eventuality or if there are arguments against it

crates/auth/src/bin/calimero-auth.rs
miraclx, line 1: not convinced we need any files/folders `calimero-` prefixed while we're already in the calimero repo

alenmestrov: this is resolved in the first PR, I moved everything to main.rs file

crates/auth/src/api/handlers/mod.rs
alenmestrov: I think we can do it, I can move auth frontend from core to separate repo and replicate the process we have for admin-dashboard

PR #1315, opened 2025-06-25
feat(Auth):  cli login

alenmestrov:
# [Auth] Implemented auth for Meroctl

## Description

Introduced login functionality for CLI 

## Test plan

## Documentation update

### DIFF
--- a/crates/meroctl/src/cli/auth.rs
+++ b/crates/meroctl/src/cli/auth.rs
@@ -1,0 +1,445 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+use std::io::{self, Write};
+
+use axum::extract::Query;
+use axum::response::Html;
+use axum::routing::get;
+use axum::Router;
+use clap::{Parser, Subcommand};
+use eyre::{bail, eyre, OptionExt, Result as EyreResult};
+use serde::Deserialize;
+use tokio::net::TcpListener;
+use tokio::sync::oneshot;
+use tokio::time::timeout;
+use url::Url;
+
+use crate::cli::storage::{create_storage, JwtToken, ProfileConfig};
+use crate::cli::Environment;
+use crate::connection::ConnectionInfo;
+
+#[derive(Debug, Parser)]
+pub struct AuthCommand {
+    #[command(subcommand)]
+    pub action: AuthAction,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum AuthAction {
+    /// Authenticate with the server using browser-based login
+    Login(LoginCommand),
+    /// Remove stored authentication tokens
+    Logout(LogoutCommand),
+    /// Show authentication info for a profile
+    Info(InfoCommand),
+    /// Set the active profile to use for all commands
+    SetActive(SetActiveCommand),
+    /// List all available authentication profiles
+    List(ListCommand),
+    /// Clear all authentication profiles and tokens
+    Clear(ClearCommand),
+}
+
+#[derive(Debug, Parser)]
+pub struct LoginCommand {
+    /// API endpoint URL to authenticate with
+    #[arg(long, value_name = "URL")]
+    pub api: Url,
+    
+    /// Profile name to store authentication tokens for
+    #[arg(long, default_value = "default")]
+    pub profile: String,
+    
+    /// Permissions to request during authentication (comma-separated)
+    #[arg(long)]
+    pub permissions: Option<String>,
+    
+    /// Port range to use for local callback server (e.g., 8080-8090)
+    #[arg(long, default_value = "9080-9090")]
+    pub port_range: String,
+
+    /// Timeout for the authentication flow in seconds
+    #[arg(long, default_value = "300")]
+    pub timeout: u64,
+}
+
+#[derive(Debug, Parser)]
+pub struct LogoutCommand {
+    /// Profile name to remove authentication tokens for
+    #[arg(long, default_value = "default")]
+    pub profile: String,
+}
+
+#[derive(Debug, Parser)]
+pub struct InfoCommand {
+    /// Profile name to check authentication info for
+    #[arg(long)]
+    pub profile: String,
+}
+
+#[derive(Debug, Parser)]
+pub struct SetActiveCommand {
+    /// Profile name to set as active
+    #[arg(long)]
+    pub profile: String,
+}
+
+#[derive(Debug, Parser)]
+pub struct ListCommand {
+    // No additional arguments needed for listing profiles
+}
+
+#[derive(Debug, Parser)]
+pub struct ClearCommand {
+    /// Skip confirmation prompt
+    #[arg(long, short)]
+    pub force: bool,
+}
+
+#[derive(Debug, Deserialize)]
+struct AuthCallback {
+    access_token: Option<String>,
+    refresh_token: Option<String>,
+}
+
+impl AuthCommand {
+    pub async fn run(&self, environment: &Environment) -> EyreResult<()> {
+        match &self.action {
+            AuthAction::Login(cmd) => cmd.run(environment).await,
+            AuthAction::Logout(cmd) => cmd.run(environment).await,
+            AuthAction::Info(cmd) => cmd.run(environment).await,
+            AuthAction::SetActive(cmd) => cmd.run(environment).await,
+            AuthAction::List(cmd) => cmd.run(environment).await,
+            AuthAction::Clear(cmd) => cmd.run(environment).await,
+        }
+    }
+}
+
+impl LoginCommand {
+    pub async fn run(&self, _environment: &Environment) -> EyreResult<()> {
+        // Use the API URL from the command arguments
+        let api_url = &self.api;
+
+        // Create a temporary connection for auth detection (no auth)
+        let temp_connection = ConnectionInfo::new(api_url.clone(), None, None);
+
+        // 1. Detect authentication mode
+        println!("🔍 Detecting authentication mode...");
+        let auth_mode = temp_connection.detect_auth_mode().await?;
+        
+        if auth_mode == "none" {
+            bail!("Server does not require authentication");
+        }
+
+        println!("✅ Server requires authentication (mode: {})", auth_mode);
+
+        // 2. Start local callback server
+        let (callback_port, callback_rx) = self.start_callback_server().await?;
+        println!("🌐 Started local callback server on port {}", callback_port);
+
+        // 3. Build auth URL and open browser
+        let auth_url = self.build_auth_url(api_url, callback_port)?;
+        println!("🔗 Opening browser to: {}", auth_url);
+        
+        if let Err(e) = webbrowser::open(&auth_url.to_string()) {
+            println!("⚠️  Failed to open browser automatically: {}", e);
+            println!("📋 Please manually open this URL in your browser:");
+            println!("   {}", auth_url);
+        }
+
+        // 4. Wait for callback with timeout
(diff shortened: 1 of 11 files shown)


Dockerfile.auth
```diff
@@ -0,0 +1,83 @@
...
+ARG APP_NAME
+
+# For production build, copy the source
+COPY . .
+
+# Build the release binary with forced frontend fetch
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/usr/local/cargo/git \
+    --mount=type=cache,target=/usr/src/calimero/target \
+    CALIMERO_AUTH_FRONTEND_FETCH=1 cargo build --release --bin ${APP_NAME} && \
```
miraclx: surely this isn't needed yeah? how about we leave the user to use `--build-arg CALIMERO_AUTH_FRONTEND_FETCH=1` should they need to

```suggestion
    cargo build --release --bin ${APP_NAME} && \
```

Dockerfile.auth
miraclx, line 1: can you update this file to match the changes made to https://github.com/calimero-network/core/blob/35c21784e3e785eee482c9f1cf5a59b66d1e2314/Dockerfile?

this is the diff - https://github.com/calimero-network/core/commit/29d61f1e7cd1d3abae5d2177aa8a3c2b29645939#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

docker-compose.prod.yml
```diff
@@ -0,0 +1,175 @@
...
+services:
+  # Volume initialization service
+  init_volume:
+    image: debian:bookworm-slim
+    user: root
+    volumes:
+      - calimero_auth_node:/calimero
+    command: >
+      /bin/sh -c "mkdir -p /calimero/data /calimero/logic /calimero/credentials &&
+                  chmod -R 777 /calimero"
```
miraclx: surely we don't need a whole container just to make some folders

docker-compose.prod.yml
```diff
@@ -0,0 +1,175 @@
...
+    ports:
+      - "2428:2428"
+      - "2528:2528"
+    entrypoint: ["/bin/sh", "-c"]
+    command: >
+      "mkdir -p /calimero/data &&
+      if [ ! -d /calimero/data/config ]; then
+        merod --node-name node1 --home /calimero/data init --server-host 0.0.0.0 --server-port 2428 --swarm-port 2528
+      fi &&
+      merod --node-name node1 --home /calimero/data run"
```
miraclx: you can use the `CALIMERO_HOME` env for the home dir

docker-compose.prod.yml
```diff
@@ -0,0 +1,175 @@
...
+    volumes:
+      - .:/usr/src/calimero:cached  # Mount the entire source code
+      - cargo-registry:/usr/local/cargo/registry  # Cache cargo registry
+      - cargo-git:/usr/local/cargo/git  # Cache git repositories
+      - target:/usr/src/calimero/target  # Cache build artifacts
+      - calimero_auth_data:/data/auth_db  # Persist RocksDB data
+    environment:
+      - RUST_LOG=debug
+      - CARGO_TARGET_DIR=/usr/src/calimero/target
+      - CALIMERO_AUTH_FRONTEND_FETCH=1
```
miraclx: these are build envs in runtime env position, and do we need either of them? you already set the build WORKDIR in the dockerfile

docker-compose.prod.yml
```diff
@@ -0,0 +1,175 @@
...
+      context: .
+      dockerfile: Dockerfile.auth
+      target: dev  # Use the development stage
+      args:
+        - CALIMERO_AUTH_FRONTEND_FETCH=1
+    volumes:
+      - .:/usr/src/calimero:cached  # Mount the entire source code
+      - cargo-registry:/usr/local/cargo/registry  # Cache cargo registry
+      - cargo-git:/usr/local/cargo/git  # Cache git repositories
+      - target:/usr/src/calimero/target  # Cache build artifacts
```
miraclx: aren't these already accounted for by buildx?

docker-compose.prod.yml
```diff
@@ -0,0 +1,175 @@
...
+
+  # Auth service
+  auth:
+    image: calimero/auth:dev
+    build:
+      context: .
+      dockerfile: Dockerfile.auth
+      target: dev  # Use the development stage
+      args:
+        - CALIMERO_AUTH_FRONTEND_FETCH=1
```
miraclx: this shouldn't be needed

docker-compose.prod.yml
```diff
@@ -0,0 +1,175 @@
...
+      # Define the service
+      - "traefik.http.services.node-core.loadbalancer.server.port=2428"
+
+  # Auth service
+  auth:
+    image: calimero/auth:dev
+    build:
+      context: .
+      dockerfile: Dockerfile.auth
+      target: dev  # Use the development stage
```
miraclx: why not the final stage?

crates/meroctl/src/connection.rs
```diff
@@ -1,26 +1,36 @@
...
 
+#[derive(Debug, thiserror::Error)]
+pub enum TokenError {
+    #[error("Token refresh failed")]
+    RefreshFailed,
+    #[error("Token expired and refresh token is not available")]
+    NoRefreshToken,
+    #[error("Failed to store refreshed token: {0}")]
+    StorageError(#[from] eyre::Error),
+}
```
miraclx: you should use `eyre::Report`, btw if you didn't know, errors like these don't embed backtrace information, that's somewhat tolerable in libraries but CLIs are end apps, `eyre`'s facilities are preferable

crates/meroctl/src/cli/storage/keychain.rs
```diff
@@ -0,0 +1,139 @@
+use std::sync::RwLock;
+
+use async_trait::async_trait;
+use eyre::{Context, Result as EyreResult};
```
miraclx:
```suggestion
use eyre::Context;
```

use `eyre::Result`

crates/meroctl/src/connection.rs
```diff
@@ -53,33 +63,209 @@ impl ConnectionInfo {
...
+                                        println!("❌ Failed to store refreshed token: {}", e);
+                                        return Err(eyre!("Failed to store refreshed token. Please check your keychain access and try again"));
+                                    }
+                                },
+                            }
+                        }
+                    }
+                }
+
+                return Err(eyre!(
```
miraclx: `bail!`

crates/meroctl/src/cli/auth.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Debug, Subcommand)]
+pub enum AuthAction {
+    /// Authenticate with the server using browser-based login
+    Login(LoginCommand),
+    /// Remove stored authentication tokens
+    Logout(LogoutCommand),
+    /// Show authentication info for a profile
+    Info(InfoCommand),
+    /// Set the active profile to use for all commands
+    SetActive(SetActiveCommand),
```
miraclx: `use`?

crates/meroctl/src/cli/context/list.rs
```diff
@@ -36,7 +36,7 @@ impl ListCommand {
             .as_ref()
             .ok_or_eyre("No connection configured")?;
 
-        let response: GetContextsResponse = connection.get("admin-api/dev/contexts").await?;
+        let response: GetContextsResponse = connection.get("admin-api/contexts").await?;
```
miraclx, line 35: how about the others?

crates/meroctl/src/cli/auth.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Debug, Parser)]
+pub struct InfoCommand {
+    #[arg(long)]
+    pub profile: String,
+}
+
+#[derive(Debug, Parser)]
+pub struct SetActiveCommand {
+    #[arg(long)]
+    pub profile: String,
```
miraclx: this can be positional

crates/meroctl/src/cli/context/list.rs
alenmestrov: This was just a test, so its a leftover, I unified them all here [1316](https://github.com/calimero-network/core/pull/1316)

crates/meroctl/src/cli/auth.rs
alenmestrov: This was implemented for better UX, so that users don't need to add `--profile` flag with every request

e2e-tests/src/meroctl.rs
```diff
@@ -212,6 +212,8 @@ impl Meroctl {
         let task = async move {
             let json = task.await?;
 
+            println!("JSON RPC response: {:?}", json);
+
```
miraclx: ^

crates/meroctl/src/connection.rs
```diff
@@ -1,86 +1,320 @@
...
+                builder =
+                    builder.header("Authorization", format!("Bearer {}", tokens.access_token));
+            }
+
+            let response = builder.send().await?;
+
+            if response.status() == 401 && self.jwt_tokens.lock().unwrap().is_some() {
+                if let Some(auth_error) = response.headers().get("x-auth-error") {
+                    if auth_error.to_str().unwrap_or("") == "token_expired" {
+                        println!("🔄 Token expired, attempting refresh...");
```
miraclx: err.. `print` is problematic here, especially when `--output-format json` is specified. for example in e2e-tests

alenmestrov: removed all status messages, bailing on error

crates/meroctl/src/cli/auth.rs
```diff
@@ -32,19 +32,17 @@ pub async fn authenticate(api_url: &Url) -> Result<JwtToken> {
...
 
     if let Err(e) = webbrowser::open(&auth_url.to_string()) {
-        println!("⚠️  Failed to open browser automatically: {}", e);
-        println!("📋 Please manually open this URL in your browser:");
-        println!("   {}", auth_url);
+        bail!(
+            "Failed to open browser: {}. Please manually open this URL: {}",
+            e,
+            auth_url
+        );
```
miraclx: this will cause meroctl to exit right? but that's not what we want `{Info,Warn,Error}Line` works here

crates/meroctl/src/connection.rs
```diff
@@ -75,9 +75,9 @@ impl ConnectionInfo {
 
             if response.status() == 401 && self.jwt_tokens.lock().unwrap().is_some() {
                 if let Some(auth_error) = response.headers().get("x-auth-error") {
+                    println!("Auth error: {}", auth_error.to_str().unwrap_or(""));
```
miraclx: ^

crates/meroctl/src/cli/auth.rs
```diff
@@ -35,12 +36,15 @@ pub async fn authenticate(api_url: &Url) -> Result<JwtToken> {
 
     let auth_url = build_auth_url(api_url, callback_port)?;
 
+    let info_msg = format!("Trying to open browser to start authentication");
+    InfoLine(&info_msg).report();
```
miraclx: `Report::report` should only be called when `--output-format human`, not unconditionally

that's what `environment.output.write` does

alenmestrov: Implemented logic for adding the Output to the ConnectionInfo, from which we then use `write()` for writing out messages

PR #1316, opened 2025-06-26
feat(Auth):  Node Server Cleanup and Meroctl Adjustment

alenmestrov:
# [Auth] Node Server Cleanup

## Description

- Removed authentication and authorization from the Calimero Node Server. 
- Unified dev endpoints with normal (admin-api) endpoints.
- Updated Meroctl to use normal endpoints instead of having a separate router.
- Removed `--auth` argument from merod startup command

 
## Test plan

TBD

## Documentation update

TBD


crates/meroctl/src/cli/app/get.rs
```diff
@@ -35,7 +35,7 @@ impl GetCommand {
             .ok_or_eyre("No connection configured")?;
 
         let response: GetApplicationResponse = connection
-            .get(&format!("admin-api/dev/applications/{}", self.app_id))
+            .get(&format!("admin-api/applications/{}", self.app_id))
```
miraclx, line 35: aah here they are

alenmestrov: :)

frdomovic (approved):
gj!

PR #1317, opened 2025-06-27
feat: cargo mero cli

iTranscend:
Introduces  a new `cargo mero` cli to facilitate creating, building, publishing & ABI generation of Calimero apps.

#### Create a new application
```sh
 $ cargo mero new --name gitter
```
#### Build app
```sh 
$ cargo mero build
```

#### Generate ABI
```sh
$ cargo mero abi
```

#### Publish app
```sh
$ cargo mero publish
```

### Todo

- [x] - Create
- [ ] - Build
- [ ] - ABI
- [ ] - Publish

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ members = [
     "./crates/version",
 
     "./apps/kv-store",
+    "./apps/app-template",
 
     "./e2e-tests",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ members = [
     "./crates/crypto",
     "./crates/meroctl",
     "./crates/merod",
+    "./crates/cargo-mero",
     "./crates/network",
     "./crates/network/primitives",
     "./crates/node",
--- a/crates/cargo-mero/Cargo.toml
+++ b/crates/cargo-mero/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "cargo-mero"
+version = "0.1.0"
+description = "CLI tool for building applications on the Calimero network"
+categories = ["command-line-utilities"]
+keywords = ["cli", "cargo", "subcommand"]
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+license-file.workspace = true
+
+[dependencies]
+tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
+eyre.workspace = true
+clap = { workspace = true, features = ["derive"] }
+const_format.workspace = true
+rust-embed.workspace = true
+
+[lints]
+workspace = true
--- a/crates/cargo-mero/src/build.rs
+++ b/crates/cargo-mero/src/build.rs
@@ -1,0 +1,3 @@
+pub async fn run() -> eyre::Result<()> {
+    Ok(())
+}
--- a/crates/cargo-mero/src/cli.rs
+++ b/crates/cargo-mero/src/cli.rs
@@ -1,0 +1,52 @@
+use std::path::PathBuf;
+
+use clap::{Parser, Subcommand};
+use const_format::concatcp;
+
+use crate::{abi, build, new};
+
+pub const EXAMPLES: &str = r"
+  # Create a new application
+  $ cargo mero new --name gitter
+
+  # Build app
+  $ cargo mero build
+
+  # Generate ABI
+  $ cargo mero abi
+
+  # Publish app
+  $ cargo mero publish
+";
+
+#[derive(Debug, Parser)]
+#[command(name = "mero")]
+#[command(bin_name = "cargo mero")]
+#[command(after_help = concatcp!("Examples: ", EXAMPLES))]
+pub struct RootCommand {
+    #[command(subcommand)]
+    command: SubCommands,
+}
+
+#[derive(Debug, Subcommand)]
+enum SubCommands {
+    New(NewCommand),
+    Build,
+    Abi,
+}
+
+#[derive(Debug, Parser)]
+pub struct NewCommand {
+    #[arg(long)]
+    pub name: PathBuf,
+}
+
+impl RootCommand {
+    pub async fn run(self) -> eyre::Result<()> {
+        match self.command {
+            SubCommands::Abi => abi::run().await,
+            SubCommands::New(args) => new::run(args).await,
+            SubCommands::Build => build::run().await,
+        }
+    }
+}
--- a/crates/cargo-mero/src/main.rs
+++ b/crates/cargo-mero/src/main.rs
@@ -1,0 +1,19 @@
+use clap::Parser;
+
+mod abi;
+mod build;
+mod cli;
+mod new;
+
+async fn init() -> eyre::Result<()> {
+    let command = cli::RootCommand::parse();
+
+    command.run().await
+}
+
+#[tokio::main]
+async fn main() -> eyre::Result<()> {
+    init().await?;
+
+    Ok(())
+}
--- a/crates/cargo-mero/src/new.rs
+++ b/crates/cargo-mero/src/new.rs
@@ -1,0 +1,36 @@
+use std::fs::File;
+use std::io::Write;
+use std::path::Path;
+use std::{env, fs};
+
+use rust_embed::RustEmbed;
+
+use crate::cli::NewCommand;
+
+#[derive(RustEmbed)]
+#[folder = "../../apps/app-template/"]
+struct AppTemplate;
+
+pub async fn run(args: NewCommand) -> eyre::Result<()> {
+    println!("Creating a new project {:?} from template...", args.name);
+    let path = args.name;
+
+    fs::create_dir_all(&path)?;
+    env::set_current_dir(path)?;
+
+    for file in AppTemplate::iter() {
+        let path = Path::new(&*file);
+
+        // Create directories if needed
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent)?;
+        }
+
+        let mut file_handle = File::create(path)?;
+        if let Some(file_content) = AppTemplate::get(&file) {
+            file_handle.write_all(&file_content.data)?;
+        }
+    }
+
+    Ok(())
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2321,6 +2321,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo-mero"
+version = "0.1.0"
+dependencies = [
+ "clap",
+ "const_format",
+ "eyre",
+ "rust-embed",
+ "tokio",
+]
+
+[[package]]
 name = "cc"
 version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -944,6 +944,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4c95c10ba0b00a02636238b814946408b1322d5ac4760326e6fb8ec956d85775"
 
 [[package]]
+name = "app-template"
+version = "0.1.0"
+dependencies = [
+ "calimero-sdk",
+ "calimero-storage",
+ "thiserror 1.0.69",
+]
+
+[[package]]
 name = "arbitrary"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/apps/app-template/Cargo.toml
+++ b/apps/app-template/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "app-template"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+license-file.workspace = true
+
+[lib]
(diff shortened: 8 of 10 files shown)


crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,97 @@
...
+        .arg("target")
+        .arg("add")
+        .arg("wasm32-unknown-unknown")
+        .output()
+        .await
+        .wrap_err("Adding wasm32-unknown-unknown target failed")?;
+
+    if !output.status.success() {
+        bail!("Adding wasm32-unknown-unknown target command failed");
+    }
```
miraclx: maybe we try `target list` instead?

https://github.com/near/cargo-near/blob/7866e8ee6ca992b12d3c9fc44c2724d08b90b1ad/cargo-near-build/src/cargo_native/target.rs#L35

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    let output = build_cmd.output().await.wrap_err("cargo build failed")?;
+
+    if !output.status.success() {
+        bail!("cargo build command failed");
+    }
+
+    // Copy wasm to res folder
+    fs::create_dir("res")?;
```
miraclx: we only want to do this if it doesn't exist

https://github.com/near/cargo-near/blob/2902b7e99d9d20344faf9e0f6a39f22782b35035/cargo-near-build/src/fs.rs#L54

crates/cargo-mero/src/utils.rs
```diff
@@ -0,0 +1,10 @@
+use std::ffi::OsStr;
+use std::process::Stdio;
+
+use tokio::process::Command;
+
+pub fn new_command<S: AsRef<OsStr>>(program: S) -> Command {
+    let mut command = Command::new(program);
+    let _ = command.stdout(Stdio::inherit()).stderr(Stdio::inherit());
```
miraclx: this is the default

https://doc.rust-lang.org/nightly/std/process/struct.Command.html#method.stdout

crates/cargo-mero/src/new.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::io::Write;
+use std::path::Path;
+use std::{env, fs};
+
+use rust_embed::RustEmbed;
+
+use crate::cli::NewCommand;
+
+#[derive(RustEmbed)]
+#[folder = "../../apps/app-template/"]
```
miraclx: let's colocate it with the consumer, since it's not an app itself

crates/cargo-mero/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+
+    command.run().await
+}
+
+#[tokio::main]
+async fn main() -> eyre::Result<()> {
+    init().await?;
+
+    Ok(())
+}
```
miraclx: since `init` doesn't do much more than parse and run, we can keep it with `main`

crates/cargo-mero/src/cli.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Debug, Parser)]
+pub struct NewCommand {
+    pub name: PathBuf,
+}
+
+#[derive(Debug, Parser)]
+pub struct BuildCommand {
+    pub args: Vec<String>,
+}
```
miraclx: https://github.com/near/cargo-near/blob/2902b7e99d9d20344faf9e0f6a39f22782b35035/cargo-near/src/commands/build/actions/non_reproducible_wasm/mod.rs

crates/cargo-mero/src/cli.rs
```diff
@@ -0,0 +1,64 @@
...
+enum MeroCmd {
+    #[command(subcommand)]
+    Mero(SubCommands),
+}
+
+#[derive(Debug, Subcommand)]
+enum SubCommands {
+    New(NewCommand),
+    Build(BuildCommand),
+    Abi,
```
miraclx: we can drop ABI for now, the issue was filed thinking that would've been introduced when this was to be implemented, but we're doing this first now, so it's out of scope

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,97 @@
...
+        .status()
+        .await
+        .is_ok()
+}
+
+/// Builds a path from two `&str`s
+fn build_path(a: &str, b: &str) -> PathBuf {
+    let mut path = PathBuf::from(a);
+    path.push(b);
+    path
```
miraclx: this is the same as

`Path::new(a).join(b)`, shouldn't need a function

Cargo.toml
```diff
@@ -53,6 +54,7 @@ bs58 = "0.5.0"
 bytes = "1.6.0"
 cached-path = "0.7.0"
 camino = "1.1.6"
+cargo_metadata = "0.20.0"
```
miraclx: ~nit: order

crates/cargo-mero/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+license.workspace = true
+license-file.workspace = true
+
+[dependencies]
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "process"] }
+eyre.workspace = true
+clap = { workspace = true, features = ["derive"] }
+const_format.workspace = true
+rust-embed.workspace = true
+cargo_metadata.workspace = true
```
miraclx: ~nit: order

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,112 @@
...
+    if args.verbose {
+        let _ = build_cmd.arg("--verbose");
+    }
+    if args.quiet {
+        let _ = build_cmd.arg("--quiet");
+    }
+    if let Some(features) = args.features {
+        let _ = build_cmd.arg("--features");
+        let _ = build_cmd.arg(features);
+    }
```
miraclx: `--features` should support multiple invocations

both of these are legal with `cargo build`:

- `--features feat1,feat2`
- `--features feat1 --features feat2`

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,112 @@
...
+        .wrap_err("cargo build failed")?;
+
+    if !output.success() {
+        bail!("cargo build command failed");
+    }
+
+    // Copy wasm to res folder
+    if !Path::new("res/").exists() {
+        fs::create_dir("res")?;
+    }
```
miraclx: `res` where? current directory?

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let package_name = MetadataCommand::new().exec()?.packages[0]
+        .name
+        .clone()
+        .into_inner()
+        .replace("-", "_");
+    let wasm_file = format!("{}.wasm", package_name);
+    let wasm_path = Path::new("./target/wasm32-unknown-unknown/app-release/").join(&wasm_file);
```
miraclx: this shouldn't be hardcoded

if `--no-release` is specified, for example: this will break

same with if the `CARGO_TARGET_DIR` env is specified

crates/cargo-mero/app-template/Cargo.lock
miraclx, line 1: preferably, let's not commit to a lockfile

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,112 @@
...
+    let mut build_cmd = Command::new("cargo");
+    let _ = build_cmd
+        .arg("build")
+        .arg("--target")
+        .arg("wasm32-unknown-unknown");
+
+    // Add additional pass-through cargo arguments
+    if !args.no_locked {
+        let _ = build_cmd.arg("--locked");
+    }
```
miraclx: it should be the other way around

```suggestion
    // Add additional pass-through cargo arguments
    if args.locked {
        let _ = build_cmd.arg("--locked");
    }
```

crates/cargo-mero/src/build.rs
iTranscend: yeah, res is created in the current dir

crates/cargo-mero/src/build.rs
```diff
@@ -42,19 +41,38 @@ pub async fn run(args: BuildOpts) -> eyre::Result<()> {
...
+    for message in Message::parse_stream(stdout_reader) {
+        match message? {
+            Message::CompilerArtifact(artifact) => {
+                artifacts.push(artifact);
+            }
+            _ => {}
+        }
+    }
+
+    let wasm_path = artifacts.last().unwrap().filenames[0].clone();
```
miraclx: since we don't need any more artifacts, we shouldn't need to allocate for all of them

instead of a `Vec`, you can use `None` instead, and call [`Option::<T>::replace`](https://doc.rust-lang.org/nightly/std/option/enum.Option.html#method.replace)

at which point you shouldn't need the `.clone` either

crates/cargo-mero/src/build.rs
```diff
@@ -42,19 +41,38 @@ pub async fn run(args: BuildOpts) -> eyre::Result<()> {
...
+    let wasm_path_str = wasm_path.clone().into_string();
+    let wasm_file = wasm_path_str.split("/").last().unwrap();
```
miraclx: [`Utf8Path::file_name`](https://docs.rs/camino/1.1.10/camino/struct.Utf8PathBuf.html#method.file_name)?

crates/cargo-mero/src/build.rs
```diff
@@ -64,15 +82,6 @@ pub async fn run(args: BuildOpts) -> eyre::Result<()> {
     if !Path::new("res/").exists() {
         fs::create_dir("res")?;
     }
```
miraclx: let's not assume they're running `mero build` in the same folder as the project being built

for example, the app may be in a subfolder like we have it, but the user builds in the root directory

yet, the `res` folder should be in the same folder as the project

crates/cargo-mero/src/build.rs
```diff
@@ -0,0 +1,119 @@
...
+    if !output.status.success() {
+        bail!("Adding wasm32-unknown-unknown target command failed");
+    }
+
+    let mut build_cmd = Command::new("cargo");
+    let _ = build_cmd
+        .arg("build")
+        .arg("--target")
+        .arg("wasm32-unknown-unknown")
+        .arg("--message-format=json-render-diagnostics");
```
miraclx, line 28: we need a `-p, --package <pkg_name>` option like `cargo build`

the value received here can then be resolved to the directory of the workspace member via

- `MetadataCommand::exec() -> Metadata` ([ref](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.MetadataCommand.html#method.exec))
- then find an item in `{Metadata}.packages` ([ref](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#structfield.packages)) that is a workspace member `{Metadata}.workspace_member` ([ref](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#structfield.workspace_members)) and `{Package}.name` ([ref](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Package.html#structfield.name)) is what the user specified ([here's](https://docs.rs/cargo_metadata/latest/src/cargo_metadata/lib.rs.html#212-217) a helper function to guide, we can use this but it'll allocate and require us to still do another round of iteration which we don't need)
- Once the package is identified, we can extract the path to the package via `{Package}.manifest_path` ([ref](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Package.html#structfield.manifest_path))

and for cases where `-p, --package <pkg>` is not specified, we can take either of these approaches:

1. just `cargo build`, which will build items specified in [`default-members`](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-default-members-field) (in the case of a workspace) or build the project when it's not a workspace
2. build any workspace member that depends on `calimero-sdk` 🤷🏽‍♂️ 

but the first option seems okay to me, when it fails it'll be a signal for the user to be more specific

crates/cargo-mero/src/cli.rs
```diff
@@ -63,6 +63,9 @@ pub struct BuildOpts {
     /// No default features
     #[clap(long)]
     pub no_default_features: bool,
+    /// Package to build
+    #[clap(long, short)]
+    pub package: String,
```
miraclx: this should still be optional

crates/cargo-mero/src/build.rs
```diff
@@ -83,20 +85,21 @@ pub async fn run(args: BuildOpts) -> eyre::Result<()> {
     }
 
     // Copy wasm to res folder
-    if !Path::new("res/").exists() {
-        fs::create_dir("res")?;
+    let res_path = Path::new(manifest_dir_path).join("res/");
```
miraclx:
```suggestion
    let res_path = Path::new(manifest_dir_path).join("res");
```

Cargo.toml
```diff
@@ -15,6 +15,7 @@ members = [
     "./crates/crypto",
     "./crates/meroctl",
     "./crates/merod",
+    "./crates/cargo-mero",
```
miraclx: ~nit: order

miraclx (approved):

Issue #1286, opened 2025-05-27
Validate PR titles as conventional commits


miraclx:
Validate that PR titles follow the conventional commits specification

https://www.conventionalcommits.org/en/v1.0.0/

Fail otherwise, blocking merge.

metaclips:
Can I take this issue?

PR #1324, opened 2025-07-06
ci: add commit lint workflow

metaclips:
# [CI] Validate PR follows calimero-network commit message style fixes https://github.com/calimero-network/core/issues/1286

## Description

[Ensures PRs follows conventional commit messaging style](https://www.conventionalcommits.org/en/v1.0.0/)

## Test plan

Tested in my personal repository to ensure PRs title follow conventional commit

## Documentation update

We comment on users PR if the workflow fails on how to follow our commit message rule.

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -2,7 +2,7 @@ name: PR Comment
 
 on:
   workflow_run:
-    workflows: ['e2e-tests', 'PR Description']
+    workflows: ['e2e-tests', 'PR Description', 'PR Title']
     types:
       - completed
 
--- a/.github/workflows/pr-title-lint.yml
+++ b/.github/workflows/pr-title-lint.yml
@@ -1,0 +1,55 @@
+name: PR Title
+# ^~~ if changed, update comment.yml as well
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      ERR_MSG: |-
+        The PR title `${{ github.event.pull_request.title }}` does not follow the Calimero-Network Commit Message Convention
+        https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
+
+        Our PR title should have the following structure:
+
+          <type>(<scope>): <subject>
+
+        Common errors to avoid:
+        1. The commit header <type>(<scope>): <subject> must be in lower case.
+        2. Allowed type values are: build, ci, docs, feat, fix, perf, refactor, test.
+
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        id: lint_pr_title
+
+      - name: Create PR Comment Payload
+        if: ${{ !cancelled() }}
+        run: >-
+          jq -n
+          --arg pr '${{ github.event.number }}'
+          --arg tag pr-lint-check
+          --arg mode "${{ steps.lint_pr_title.outputs.error_message != null && 'recreate' || 'delete' }}"
+          --arg message "${{ env.ERR_MSG }}"
+          '{pr: $pr, tag: $tag, mode: $mode, message: $message}' | tee payload.json
+
+      - name: Upload Comment
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: pr-comment-payload
+          path: payload.json


miraclx (commented):
Hi @metaclips, thanks for taking this up

Just a few notes;

1. we shouldn't care much what format the commits in a PR take, since fixing them will require force-pushes which aren't ideal per say, what we care about instead is the PR title (which eventually becomes the squash commit message), so all we have to do is check the PR title for compliance
2. when there's a violation, you can comment on the PR for visibility and fail CI like we do for the description check see it [here](https://github.com/calimero-network/core/blob/35c21784e3e785eee482c9f1cf5a59b66d1e2314/.github/workflows/pr-description-check.yml), and the other half of the comment handler where you have to append your workflow name to line 5
https://github.com/calimero-network/core/blob/35c21784e3e785eee482c9f1cf5a59b66d1e2314/.github/workflows/comment.yml#L5

.github/workflows/commit_lint.yml
miraclx, line 1: we can change the filename now

.github/workflows/commit_lint.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Install Commitlint
+        shell: bash
+        run: npm install --location=global @commitlint/cli@17.1.1
+
+      - name: Download Commitlint Config
+        shell: bash
+        run: |
+          if ! ls | grep commitlint.config.js; then
+            mv $COMMITLINT_CONFIG_PATH commitlint.config.js
+          fi
```
miraclx: this is post-checkout, dropping us to the root of the workspace, isn't it guaranteed the file will exist at this point?

.github/workflows/commit_lint.yml
```diff
@@ -0,0 +1,75 @@
...
+      pull-requests: write
+
+    env:
+      COMMITLINT_CONFIG_PATH: "commitlint.config.js"
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
miraclx: does this work?

miraclx: the CI run suggests maybe not - https://github.com/calimero-network/core/actions/runs/16128280299/job/45528947706?pr=1324#step:2:56

.github/workflows/commit_lint.yml
```diff
@@ -0,0 +1,75 @@
...
+            https://github.com/calimero-network/core/blob/master/commitlint.config.js
+
+            More about the Commit Message Convention
+            https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
+          '
+
+          echo "$PR_TITLE" | npx commitlint \
+            --config commitlint.config.js \
+            --help-url https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style || \
+            ( gh pr comment ${{ github.event.pull_request.number }} --body "$ERR" && exit 1)
```
miraclx: > To comment on PRs, I used GitHub's own CLI [PR comment command ](https://cli.github.com/manual/gh_pr_comment)as I felt it was more straightforward (please indicate if I should instead use the workflow). I have also tested in my personal repo to ensure things work

the issue is for security reasons, the github token for forks does not have permissions to write to the repo, which includes making comments. so we have the secondary workflow that is triggered for the PR, but on the master branch so it's not susceptible to any malicious code introduced in the PR

so as you have it defined, it will only work for members of the organization

.github/workflows/commit_lint.yml
metaclips: Removed.

.github/workflows/commit_lint.yml
metaclips: Noted, I used the comment workflow instead. We should use the `gh cli` command to comment to PRs instead of the third party action.

.github/workflows/commit_lint.yml
metaclips: I changed it to `pr-title-lint.yml` do indicate if there's a better rename

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+            Common errors to avoid:
+            1. The commit header <type>(<scope>): <subject> must be in lower case.
+            2. Allowed type values are: build, ci, docs, feat, fix, perf, refactor, test.
+
+            The linting rules are defined in:
+            https://github.com/calimero-network/core/blob/master/commitlint.config.js
+
+            More about the Commit Message Convention
+            https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
+          '
```
miraclx: can we make this an env?

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+            The PR title `${{ github.event.pull_request.title }}` does not follow the Calimero-Network Commit Message Convention
+            https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
+
+            Our PR title should have the following structure.
+
+              <type>(<scope>): <subject>
+              <BLANK LINE>
+              <body>
+              <BLANK LINE>
+              <footer>
```
miraclx: this is linting the PR title, we can't have a "body" in the title

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -ex
+          PR_TITLE="${{ github.event.pull_request.title }}"
+
+          ERR='
+            The PR title `${{ github.event.pull_request.title }}` does not follow the Calimero-Network Commit Message Convention
+            https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
+
+            Our PR title should have the following structure.
```
miraclx:
```suggestion
            Our PR title should have the following structure:
```

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Install Commitlint
+        shell: bash
+        run: npm install --location=global @commitlint/cli@17.1.1
+
+      - name: Lint Commit
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -ex
```
miraclx: can we do without `-x`? so the CI output is easier to parse

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+              '{pr: $pr, tag: $tag, mode: $mode, message: $message}' | tee payload.json
+            exit 1
+          else
+            jq -n \
+              --arg pr '${{ github.event.number }}' \
+              --arg tag commit-lint-check \
+              --arg mode "delete" \
+              --arg message "" \
+              '{pr: $pr, tag: $tag, mode: $mode, message: $message}' | tee payload.json
+          fi
```
miraclx: can we deduplicate the jq call? like we do in [description check](https://github.com/calimero-network/core/blob/35c21784e3e785eee482c9f1cf5a59b66d1e2314/.github/workflows/pr-description-check.yml#L66-L71)

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Upload Comment
+        if: ${{ failure() }}
```
miraclx: it should upload regardless of the status, so it should be `!cancelled()` because in the case of a success, we may need to "delete" the old comment if it existed

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
...
+      contents: read
+      pull-requests: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install Commitlint
+        shell: bash
+        run: npm install --location=global @commitlint/cli@17.1.1
```
miraclx: can we maybe use https://github.com/amannn/action-semantic-pull-request should help avoid the extra configuration file?

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,87 @@
+name: Lint Pull Request Title
```
miraclx: let's make this prettier, for example:

.github/workflows/pr-title-lint.yml
metaclips: Done

.github/workflows/pr-title-lint.yml
metaclips: Updated

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,61 @@
+name: Pull Request Title
```
miraclx: PR Title?

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,60 @@
...
+        Our PR title should have the following structure:
+
+          <type>(<scope>): <subject>
+
+        Common errors to avoid:
+        1. The commit header <type>(<scope>): <subject> must be in lower case.
+        2. Allowed type values are: build, ci, docs, feat, fix, perf, refactor, test.
+
+        The linting rules are defined in:
+        https://github.com/calimero-network/core/blob/master/commitlint.config.js
```
miraclx: we no longer have this

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,60 @@
...
+
+        More about the Commit Message Convention
+        https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
```
miraclx: this link appears twice in the message

.github/workflows/comment.yml
```diff
@@ -2,7 +2,7 @@ name: PR Comment
 
 on:
   workflow_run:
-    workflows: ['e2e-tests', 'PR Description']
+    workflows: ['e2e-tests', 'PR Description', 'Lint Pull Request Title']
```
miraclx: the workflow name has been updated

.github/workflows/pr-title-lint.yml
miraclx: also please indicate for others what to do when this changes

https://github.com/calimero-network/core/blob/35c21784e3e785eee482c9f1cf5a59b66d1e2314/.github/workflows/pr-description-check.yml#L2

.github/workflows/pr-title-lint.yml
```diff
@@ -0,0 +1,55 @@
...
+      ERR_MSG: |-
+        The PR title `${{ github.event.pull_request.title }}` does not follow the Calimero-Network Commit Message Convention
+        https://github.com/calimero-network/core/blob/master/CONTRIBUTING.md#commit-message-style
+
+        Our PR title should have the following structure:
+
+          <type>(<scope>): <subject>
+
+        Common errors to avoid:
+        1. The commit header <type>(<scope>): <subject> must be in lower case.
```
miraclx, line 33: Can you verify if this is still the case with the action were using?

metaclips: Can you approve workflow run so I can verify

metaclips: Yes, case sensitivity is still valid https://github.com/calimero-network/core/actions/runs/16146139654/job/45566080680?pr=1324

PR #1326, opened 2025-07-10
ci: add inplace workflow to run during required status check

metaclips:
# [product] Add inplace workflows to run during status checks

## Description

We have workflows that MUST pass for a PR to be merged, some of these workflow are ONLY run during Rust code changes blocking PRs that only don't make Rust change. We now have workflows that runs in-place to ensure PRs can be merged

## Test plan
All required status checks now pass. I also plan on creating another PR that makes CI changes only so as to reconfirm

Screenshot 2025-07-10 at 3 08 30 PM

## Documentation update

Documentation added to workflow file


miraclx (commented):
thanks for this, we can shrink it further though. here's a list of the required status checks

I think, a single ignored.yml file with these should work yeah?

Image

metaclips:
I think it's best we have each workflows have their ignored equivalent so that updating paths becoming dynamic.

.github/workflows/e2e-tests-ignored.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - run: echo "Ignored"
+
+  comment:
+    name: Submit Report
+    needs: test
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - run: echo "Ignored"
```
miraclx:
```suggestion
  test:
    name: ${{ matrix.protocol }}
    runs-on: ubuntu-latest
    strategy:
      matrix:
        protocol: [near, icp, stellar, ethereum]
      fail-fast: false
    steps:
      - run: echo "Ignored"
```

.github/workflows/release-ignored.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  prepare:
+    name: Prepare
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Ignored"
+
+  build-binaries:
+    name: Build Binaries
+    needs: prepare
```
miraclx:
```suggestion
  build-binaries:
    name: Build Binaries
```

.github/workflows/release-ignored.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  pull_request:
+    paths-ignore:
+      - Cargo.toml
+      - Cargo.lock
+      - "crates/**"
+
+permissions:
+  contents: read
```
miraclx: we can drop these as well

metaclips:
> > I think it's best we have each workflows have their ignored equivalent so that updating paths becoming dynamic.
> 
> that's alright, but we can still shrink these, and only focus on the actual required jobs

Done

frdomovic (approved):
lgtm

PR #1337, opened 2025-07-16
feat(Blob): Refactored Blob Handlers

alenmestrov:
# [Blob] Refactored blob handles

## Description

Moved Blob handling logic to `node_client` and updated HTTP Blob handlers to use methods from `node_client`

## Test plan

Upload, download and get blob info from [demo-blob-app](https://github.com/calimero-network/demo-blob-app)

## Documentation update

N/A. Internal optimization, everything front facing stayed the same


crates/server/src/admin/handlers/blob.rs
```diff
@@ -6,7 +6,10 @@ use axum::http::{header, StatusCode};
 use axum::response::{IntoResponse, Response};
 use axum::Extension;
 use calimero_primitives::blobs::BlobId;
+use calimero_store::key::BlobMeta as BlobMetaKey;
+use calimero_store::layer::LayerExt;
 use futures_util::{AsyncRead, StreamExt};
+use hex;
```
miraclx: ^

crates/server/src/admin/handlers/blob.rs
```diff
@@ -6,7 +6,10 @@ use axum::http::{header, StatusCode};
 use axum::response::{IntoResponse, Response};
 use axum::Extension;
 use calimero_primitives::blobs::BlobId;
+use calimero_store::key::BlobMeta as BlobMetaKey;
```
miraclx:
```suggestion
use calimero_store::key;
```

use `key::BlobMeta` directly where you need

crates/server/src/admin/handlers/blob.rs
```diff
@@ -161,28 +285,252 @@ pub async fn info_handler(
...
+                err
+            );
+        }
+    }
+
+    None
+}
+
+/// Detect MIME type from file magic bytes
+fn detect_mime_from_bytes(bytes: &[u8]) -> String {
```
miraclx: https://docs.rs/infer would've been nice, but for a first impl this is tolerable

crates/server/src/admin/handlers/blob.rs
```diff
@@ -161,28 +285,252 @@ pub async fn info_handler(
...
+                .header("Content-Type", &content_type)
+                .header("ETag", &etag)
+                .header("Cache-Control", "public, max-age=3600") // 1 hour cache
+                .header("X-Blob-ID", blob_id.to_string())
+                .header("X-Blob-Hash", hex::encode(blob_meta.hash))
+                .header("X-Blob-MIME-Type", &content_type)
+                .header(
+                    "Access-Control-Expose-Headers",
+                    "X-Blob-ID, X-Blob-Hash, X-Blob-MIME-Type, ETag",
+                )
```
miraclx: the only observable difference between `GET /blobs/{id}` and `HEAD /blobs/{id}` is the body is not available in a `HEAD` request

which means all these headers should be present in the `GET` response as well

crates/server/src/admin/handlers/blob.rs
```diff
@@ -95,6 +98,125 @@ pub async fn upload_handler(
...
+///
+/// Returns a list of all root blob IDs and their metadata. Root blobs are either:
+/// - Blobs that contain links to chunks (segmented large files)
+/// - Standalone blobs that aren't referenced as chunks by other blobs
+/// This excludes individual chunk blobs to provide a cleaner user experience.
+pub async fn list_handler(Extension(state): Extension<Arc<AdminState>>) -> impl IntoResponse {
+    let handle = state.store.clone().handle();
+
+    let iter_result = handle.iter::<BlobMetaKey>();
+    let mut iter = match iter_result {
```
miraclx: as discussed, this implementation is the responsibility of the `NodeClient`

crates/server/src/admin/handlers/blob.rs
```diff
@@ -143,10 +265,12 @@ pub async fn download_handler(
...
 
-/// Get blob info
+/// Delete a blob metadata by its ID
 ///
-/// Returns information about the blob including its size.
-pub async fn info_handler(
+/// Removes blob metadata from database only (files remain on disk for safety).
+/// This includes metadata for all associated chunk files for large blobs.
+/// Actual blob files are preserved to prevent breaking applications.
+pub async fn delete_handler(
```
miraclx, line 230: same here, so we can have a `merod` interactive CLI implementation

crates/node/primitives/src/client/blob.rs
```diff
@@ -1,10 +1,13 @@
...
 use calimero_blobstore::{Blob, Size};
-use calimero_primitives::blobs::BlobId;
+use calimero_primitives::blobs::{BlobId, BlobInfo, BlobMetadata};
 use calimero_primitives::hash::Hash;
+use calimero_store::key::BlobMeta;
+use calimero_store::layer::LayerExt;
 use eyre::bail;
-use futures_util::AsyncRead;
+use futures_util::{AsyncRead, StreamExt};
+use infer;
```
miraclx:
```suggestion
```

crates/node/primitives/src/client/blob.rs
```diff
@@ -1,10 +1,13 @@
 use std::sync::Arc;
 
 use calimero_blobstore::{Blob, Size};
-use calimero_primitives::blobs::BlobId;
+use calimero_primitives::blobs::{BlobId, BlobInfo, BlobMetadata};
 use calimero_primitives::hash::Hash;
+use calimero_store::key::BlobMeta;
```
miraclx: use `key::BlobMeta` where needed, prefixing `key` helps solve the ambiguity with possible uses of `types::BlobMeta`

crates/node/primitives/src/client/blob.rs
```diff
@@ -70,7 +73,233 @@ impl NodeClient {
...
+    }
+}
+
+/// Detect MIME type from file bytes using the infer crate
+fn detect_mime_from_bytes(bytes: &[u8]) -> String {
+    if let Some(kind) = infer::get(bytes) {
+        return kind.mime_type().to_owned();
+    }
+
+    "application/octet-stream".to_owned()
```
miraclx: `mime_type` already returns a `&'static str`

```suggestion
/// Detect MIME type from file bytes using the infer crate
fn detect_mime_from_bytes(bytes: &[u8]) -> &'static str {
    if let Some(kind) = infer::get(bytes) {
        return kind.mime_type();
    }

    "application/octet-stream"
```

miraclx (approved):

frdomovic (approved):
Please update proper PR description. ty

PR #1340, opened 2025-07-17
feat(Blob): Merod & Meroctl implementation

alenmestrov:
# [Blob] Add blob management commands for listing, info, and deletion in merod and meroctl

## Description

This PR implements blob management functionality for both `meroctl` and `merod` interactive CLI, providing users with the ability to list, inspect, and delete blobs.

**Changes made:**

### meroctl (HTTP-based, works with remote/local nodes)
- **`meroctl blob list`**: Lists all root blobs via admin-api with size information in table format
- **`meroctl blob info <blob_id>`**: Retrieves blob metadata (size, MIME type, hash) via HEAD request to admin-api  
- **`meroctl blob delete <blob_id>`**: Deletes blobs via admin-api with success/failure status

### merod interactive CLI (NodeClient-based, local only)
- **`blob list`**: Lists blobs using NodeClient directly with consistent table formatting
- **`blob info <blob_id>`**: Gets blob metadata using NodeClient with table output
- **`blob delete <blob_id>`**: Deletes blobs using NodeClient with status messages

### Technical improvements
- **Introduced HEAD request**: We now have the option to do HEAD requests in meroctl
- **JWT token refresh for HEAD requests**: Extracted auth retry logic into reusable helper function `execute_request_with_auth_retry` 

## Test plan

**Manual testing performed:**

1. **meroctl blob commands** (tested with running node):
   ```bash
   # List blobs
   meroctl blob ls|list --node <node_name>
   
   # Get blob info  
   meroctl blob info <blob_id> --node <node_name>
   
   # Delete blob
   meroctl blob delete|rm <blob_id> --node <node_name>
   ```
Screenshot 2025-07-17 at 13 40 58

2. **merod interactive CLI** (tested in interactive mode):
   ```bash
   merod --home <node_home>
   > blob ls|list
   > blob info <blob_id>  
   > blob delete|rm <blob_id>
   ```
Screenshot 2025-07-17 at 13 35 30

## Documentation Update

TBD


frdomovic (approved):
Can you please also add issue to docs that we add these new commands

crates/meroctl/src/cli/blob.rs
```diff
@@ -0,0 +1,49 @@
...
+    EXAMPLES
+))]
+pub struct BlobCommand {
+    #[command(subcommand)]
+    pub subcommand: BlobSubCommands,
+}
+
+#[derive(Debug, Subcommand, Copy, Clone)]
+pub enum BlobSubCommands {
+    Delete(delete::DeleteCommand),
```
miraclx: `alias = "rm"`

crates/meroctl/src/cli/blob.rs
```diff
@@ -0,0 +1,49 @@
...
+    Info(info::InfoCommand),
+    #[command(alias = "ls")]
+    List(list::ListCommand),
```
miraclx: can you verify if the order of these variants informs how they show up in the CLI?

if so, `delete` should not be first

crates/meroctl/src/cli/blob/info.rs
```diff
@@ -0,0 +1,88 @@
...
+        let _ = table.set_header(vec![
+            Cell::new("Blob ID").fg(Color::Blue),
+            Cell::new("Size (bytes)").fg(Color::Blue),
+            Cell::new("MIME Type").fg(Color::Blue),
+            Cell::new("Hash").fg(Color::Blue),
+        ]);
+
+        let _ = table.add_row(vec![
+            &self.data.blob_id.to_string(),
+            &format!("{}", self.data.size),
```
miraclx:
```suggestion
            &self.data.size.to_string(),
```

crates/meroctl/src/cli/blob/list.rs
```diff
@@ -0,0 +1,58 @@
...
+                Cell::new("Size").fg(Color::Blue),
+                Cell::new("Size (MB)").fg(Color::Blue),
+            ]);
+            for blob in &self.data.blobs {
+                let size_mb = blob.size as f64 / (1024.0 * 1024.0);
+                let _ = table.add_row(vec![
+                    blob.blob_id.to_string(),
+                    format!("{} bytes", blob.size),
+                    format!("{:.2} MB", size_mb),
+                ]);
```
miraclx: ideally we print one or the other, conventionally you can have a `-h` flag for human readable or `-b` for raw bytes, depending on what you pick for the default

crates/meroctl/src/cli/blob/list.rs
```diff
@@ -0,0 +1,58 @@
...
+                Cell::new("Blob ID").fg(Color::Blue),
+                Cell::new("Size").fg(Color::Blue),
+                Cell::new("Size (MB)").fg(Color::Blue),
+            ]);
+            for blob in &self.data.blobs {
+                let size_mb = blob.size as f64 / (1024.0 * 1024.0);
+                let _ = table.add_row(vec![
+                    blob.blob_id.to_string(),
+                    format!("{} bytes", blob.size),
+                    format!("{:.2} MB", size_mb),
```
miraclx: if you're using 1024, then it should be `MiB`, for IEC units, 1000 would be MB

crates/node/src/interactive_cli/blob.rs
```diff
@@ -0,0 +1,97 @@
...
+                    }
+                }
+                Err(err) => {
+                    eprintln!("{ind} Failed to list blobs: {}", err);
+                }
+            },
+            BlobSubCommand::Info { blob_id } => match node_client.get_blob_info(*blob_id).await {
+                Ok(Some(metadata)) => {
+                    println!(
+                        "  {:<44} | {:<22} | {:<20} | {}",
```
miraclx, line 57: `ind` = `>>`, for consistency anyway, even though this `  ` is tempting

crates/primitives/src/blobs.rs
```diff
@@ -72,7 +72,7 @@ impl FromStr for BlobId {
 }
 
 /// Core blob information
-#[derive(Debug, Serialize, Clone, Copy)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
 pub struct BlobInfo {
     /// The unique blob ID
```
miraclx, line 77: odd, these shouldn't be in `calimero-primitives`

crates/node/src/interactive_cli/blob.rs
alenmestrov: :) will change

crates/meroctl/src/cli/blob/list.rs
alenmestrov: this is a leftover, I abonded the MiB and left bytes for now, will remove this

crates/meroctl/src/cli/blob/list.rs
alenmestrov: yes, I left for now only raw bytes

Issue #1331, opened 2025-07-14
Blob discovery & selective announcement


miraclx:
We need to implement blob discovery via libp2p's `kad` implementation ([ref](https://docs.rs/libp2p/latest/libp2p/kad/index.html))

And consider if we intend to announce every blob or just ones that are deemed okay.

For example: documents shared in an app may be encrypted, we want to announce we have the encrypted one, so other peers in the context can retrieve them, but not the raw files.

We may perhaps consider introducing a syscall to the wasm environment to indicate to the node certain blob_id's are safe to announce.

Q: what if one context says it's safe and another, not.. what to do then?

PR #1342, opened 2025-07-21
feat(Blob): Discovery of Blobs on another nodes WIP

alenmestrov:
# [Blob] WIP MVP for discovery of blobs on another node

## Description

This is heavy WIP of Blob discovery implementation. This is a "hacked" solution for the curbChat and MeroDocs applications so that they can transfer file blobs between the nodes.
Resolves this [issue](https://github.com/calimero-network/core/issues/1331)

## Test plan

https://github.com/user-attachments/assets/a5e9a5f0-b9d9-43a5-88f0-7af2430057bf

## Documentation update

N/A. After production ready implementation we will update the documentation


crates/network/Cargo.toml
```diff
@@ -10,6 +10,9 @@ license.workspace = true
 actix.workspace = true
 eyre.workspace = true
 futures-util.workspace = true
+hex.workspace = true
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
 libp2p = { workspace = true, features = [
```
miraclx: ~nit: order

crates/network/primitives/src/client.rs
```diff
@@ -1,25 +1,27 @@
...
 }
 
 impl NetworkClient {
-    pub const fn new(network_manager: LazyRecipient<NetworkMessage>) -> Self {
+    pub const fn new(network_manager: calimero_utils_actix::LazyRecipient<NetworkMessage>) -> Self {
         Self { network_manager }
     }
 
-    pub async fn dial(&self, peer_addr: Multiaddr) -> eyre::Result<()> {
+    pub async fn dial(&self, peer_addr: libp2p::Multiaddr) -> eyre::Result<()> {
```
miraclx: why inline these?

crates/network/src/lib.rs
```diff
@@ -43,6 +44,7 @@ pub struct NetworkManager {
     discovery: Discovery,
     pending_dial: HashMap<PeerId, oneshot::Sender<EyreResult<()>>>,
     pending_bootstrap: HashMap<QueryId, oneshot::Sender<EyreResult<()>>>,
+    pending_blob_queries: Option<HashMap<QueryId, oneshot::Sender<eyre::Result<Vec<PeerId>>>>>,
```
miraclx: why is this an option?

crates/server/src/admin/handlers/blob.rs
```diff
@@ -74,22 +80,53 @@ pub async fn upload_handler(
...
-                blob_id,
-                size,
-                size as f64 / (1024.0 * 1024.0)
-            );
+            if context_id.is_some() {
+                tracing::info!(
+                    "Successfully uploaded streaming blob {} with size {} bytes ({:.1} MB) and announced to network",
+                    blob_id,
+                    size,
+                    size as f64 / (1024.0 * 1024.0)
```
miraclx: when using 1024, use MiB

crates/node/src/handlers/network_event.rs
```diff
@@ -2,76 +2,189 @@ use actix::{AsyncContext, Handler, Message, WrapFuture};
...
+use calimero_node_primitives::client::NodeClient;
 use calimero_node_primitives::sync::BroadcastMessage;
+use calimero_primitives::blobs::BlobId;
 use calimero_primitives::context::ContextId;
 use calimero_primitives::hash::Hash;
 use calimero_primitives::identity::PublicKey;
 use eyre::bail;
-use libp2p::PeerId;
+use futures_util::{SinkExt, StreamExt};
+use hex;
```
miraclx:
```suggestion
```

crates/node/primitives/src/client/blob.rs
```diff
@@ -69,6 +218,110 @@ impl NodeClient {
...
+            return Ok(Some(bytes));
+        }
+
+        // If not found locally, query the network
+        tracing::debug!(
+            blob_id = %hex::encode(&**blob_id),
+            context_id = %hex::encode(&**context_id),
+            "Blob not found locally, querying network"
+        );
+
```
miraclx: why hex-encode blob ID and context ID? they're already `Display`-able 🤷🏽‍♂️

miraclx (commented):
I don't think blobs need to be scoped out to contexts, at this level, it should be observably equivalent to IPFS

should simplify a lot of the complexity

and hex-encoding should only be used for showing (visually) hashes of the blob in meroctl or merod's interactive CLI which is different from BlobId (which should be represented as bs58)

crates/network/Cargo.toml
alenmestrov: this should be correct? in every other Cargo.toml is in this order?

miraclx: **L**-`ibp2p` should come before **S**-`erde`

crates/network/src/handlers/commands/announce_blob.rs
```diff
@@ -0,0 +1,63 @@
...
+            blob_id = %request.blob_id,
+            context_id = %request.context_id,
+            size = request.size,
+            "Announcing blob to DHT"
+        );
+
+        // Create a unique key for this blob in this context
+        let mut key_bytes = Vec::with_capacity(64);
+        key_bytes.extend_from_slice(&*request.context_id);
+        key_bytes.extend_from_slice(&*request.blob_id);
```
miraclx: you don't need a vec here


```suggestion
        let mut key_bytes = [0; 64];
        key_bytes.extend_from_slice(&*request.context_id);
        key_bytes.extend_from_slice(&*request.blob_id);
```

crates/network/src/handlers/commands/announce_blob.rs
```diff
@@ -0,0 +1,63 @@
...
+            request.blob_id,
+            request.context_id,
+            key_bytes.len(),
+        );
+
+        // Create a record with blob metadata (size and peer ID)
+        let peer_id = *self.swarm.local_peer_id();
+        let mut value = Vec::with_capacity(40); // 32 bytes peer_id + 8 bytes size
+        value.extend_from_slice(&peer_id.to_bytes());
+        value.extend_from_slice(&request.size.to_le_bytes());
```
miraclx: same here

crates/network/src/handlers/commands/query_blob.rs
```diff
@@ -0,0 +1,52 @@
...
+        );
+
+        let (sender, receiver) = oneshot::channel();
+
+        // Create search key based on context
+        let key = if let Some(context_id) = request.context_id {
+            // Search in specific context
+            let mut key_bytes = Vec::with_capacity(64);
+            key_bytes.extend_from_slice(&*context_id);
+            key_bytes.extend_from_slice(&*request.blob_id);
```
miraclx: ^

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,430 @@
...
+    let bytes = test_chunk.to_bytes();
+    let parsed = BlobChunk::from_bytes(&bytes)?;
+
+    if test_chunk.data != parsed.data || test_chunk.is_final != parsed.is_final {
+        return Err(eyre::eyre!("Binary protocol self-test failed"));
+    }
+
+    tracing::debug!("Binary protocol self-test passed");
+    Ok(())
+}
```
miraclx: sorry I don't understand this, why isn't this a test instead of running on runtime everytime we handle `RequestBlob`?

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,430 @@
...
+
+impl Handler<RequestBlob> for NetworkManager {
+    type Result = ResponseFuture<<RequestBlob as Message>::Result>;
+
+    fn handle(&mut self, request: RequestBlob, _ctx: &mut Context<Self>) -> Self::Result {
+        // Test binary protocol on first use
+        if let Err(e) = test_binary_protocol() {
+            warn!("Binary protocol test failed: {}", e);
+        }
```
miraclx: ^

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,430 @@
...
+
+        if bytes.len() != 9 + data_len {
+            return Err(eyre::eyre!("Invalid chunk data: length mismatch"));
+        }
+
+        let data = bytes[9..].to_vec();
+
+        Ok(Self { data, is_final })
+    }
+}
```
miraclx: why not borsh?

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,430 @@
...
+                            }
+                            Ok(None) => {
+                                // Emit failure event
+                                event_recipient.do_send(NetworkEvent::BlobDownloadFailed {
+                                    blob_id: request.blob_id,
+                                    context_id: request.context_id,
+                                    from_peer: request.peer_id,
+                                    error: "Stream closed during chunk transfer".to_owned(),
+                                });
+                                return Err(eyre!("Stream closed during chunk transfer"));
```
miraclx: typically stream closure is interpreted as EOF, you don't need an extra `is_final` flag in the packet

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,430 @@
...
+                        blob_id = %request.blob_id,
+                        context_id = %request.context_id,
+                        peer_id = %request.peer_id,
+                        size = ?blob_response.size,
+                        "Blob found, streaming chunks"
+                    );
+
+                    // Prepare to collect chunks
+                    let expected_size = blob_response.size.unwrap_or(0);
+                    let mut collected_data = Vec::with_capacity(expected_size as usize);
```
miraclx: why collect all the chunks in memory?

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,430 @@
...
+                                        return Err(eyre::eyre!("Failed to parse blob chunk as binary or JSON"));
+                                    }
+                                }
+                            }
+                        };
+
+                        // Add chunk data to collection
+                        collected_data.extend(blob_chunk.data.clone());
+                        chunk_count += 1;
+
```
miraclx: let's not, and also we shouldn't need to clone the chunk (which may potentially be 1MB)

crates/network/src/behaviour.rs
```diff
@@ -88,7 +88,7 @@ impl Behaviour {
                             kad_config,
                         );
 
-                        kad.set_mode(Some(kad::Mode::Client));
+                        kad.set_mode(Some(kad::Mode::Server));
```
miraclx, line 91: maybe we take out this line now? and let it automatically determine that

https://docs.rs/libp2p/latest/libp2p/kad/struct.Behaviour.html#method.set_mode

crates/node/primitives/src/client/blob.rs
```diff
@@ -40,32 +39,216 @@ impl NodeClient {
...
-            .await
-            .expect("Mailbox not to be dropped");
+        // Read all bytes from the blob stream
+        let mut data = Vec::new();
+        while let Some(chunk) = blob.next().await {
+            data.extend_from_slice(&chunk?);
+        }
+
+        Ok(Some(data.into()))
+    }
```
miraclx: why change this? the point of using the actor was so we can cache blobs across reads.

also, since this reads the entire content of the blob into memory, it's not to be used trivially.

It's tolerable for stuff like application blobs that are 1mb in the worst case, but often 300kb, not images or other media.

crates/sdk/src/env.rs
```diff
@@ -257,3 +257,36 @@ pub fn blob_close(fd: u64) -> [u8; 32] {
...
+/// let current_context = env::context_id();
+/// let announced = env::blob_announce_to_context(&blob_id, &current_context);
+/// ```
+pub fn blob_announce_to_context(blob_id: &[u8; 32], context_id: &[u8; 32]) -> bool {
+    unsafe {
+        sys::blob_announce_to_context(Buffer::from(&blob_id[..]), Buffer::from(&context_id[..]))
+            .try_into()
+    }
+    .unwrap_or_else(expected_boolean)
+}
```
miraclx: surely a context A shouldn't be able to announce a blob as a context B

crates/network/src/behaviour.rs
alenmestrov: I think that it defaulted to client mode, not server mode, thats why I explicitly set it to server mode. Will check if I can mitigate it so to be automatic

crates/network/src/handlers/commands/announce_blob.rs
```diff
@@ -0,0 +1,59 @@
...
+        info!(
+            "ANNOUNCE: blob_id={}, context_id={}, key_len={}",
+            request.blob_id,
+            request.context_id,
+            key.as_ref().len(),
+        );
+
+        // Create a record with blob metadata (size and peer ID)
+        let peer_id = *self.swarm.local_peer_id();
+        let value = [peer_id.to_bytes().as_slice(), &request.size.to_le_bytes()].concat();
```
miraclx, line 33: any reason we are prefixing the `value` component of the `Record` with the peer ID?

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,389 @@
...
+                                    error: format!("Failed to parse chunk: {}", e),
+                                });
+                                return Err(eyre::eyre!("Failed to parse blob chunk: {}", e));
+                            }
+                        };
+
+                        // Get chunk size before moving the data
+                        let chunk_size = blob_chunk.data.len();
+                        // Add chunk data to collection (move semantics, no clone)
+                        collected_data.extend(blob_chunk.data);
```
miraclx, line 273: huh?

crates/network/src/handlers/commands/request_blob.rs
```diff
@@ -0,0 +1,389 @@
...
+                    debug!(
+                        blob_id = %request.blob_id,
+                        context_id = %request.context_id,
+                        peer_id = %request.peer_id,
+                        size = ?blob_response.size,
+                        "Blob found, starting streaming download"
+                    );
+
+                    // Collect chunks efficiently (no pre-allocation based on expected size)
+                    let mut collected_data = Vec::new();
```
miraclx, line 176: there is no change here, it still collects the whole blob in memory

crates/sdk/src/env.rs
```diff
@@ -257,3 +257,50 @@ pub fn blob_close(fd: u64) -> [u8; 32] {
...
+        return false;
+    }
+
+    unsafe {
+        sys::blob_announce_to_context(
+            Buffer::from(&blob_id[..]),
+            Buffer::from(&target_context_id[..]),
+        )
+        .try_into()
+    }
```
miraclx, line 304: this is still incorrect, nothing stops a guest from directly calling `sys::blob_announce_to_context` themselves

the context ID should not be variable in or specified by the guest, at all

crates/sdk/src/sys.rs
```diff
@@ -46,6 +46,8 @@ wasm_imports! {
         fn blob_read(fd: PtrSizedInt, buf: BufferMut<'_>) -> PtrSizedInt;
         fn blob_write(fd: PtrSizedInt, data: Buffer<'_>) -> PtrSizedInt;
         fn blob_close(fd: PtrSizedInt, blob_id_buf: BufferMut<'_>) -> Bool;
+        // Network blob functions
+        fn blob_announce_to_context(blob_id: Buffer<'_>, context_id: Buffer<'_>) -> Bool;
```
miraclx, line 50:
```suggestion
        fn blob_announce(blob_id: Buffer<'_>) -> Bool;
```

PR #1344, opened 2025-07-23
fix(Context): Don't install application from path on initial context sync

alenmestrov:
# [Context] short description

## Description

Don't try to install the application in initial context sync if the path is a local path. 
Instead, create a reference in the storage and download the application on periodical context sync.

## Test plan

Run 2 nodes, install the application using local path on 1st node, create a context and invite the 2nd node to it. 
Run command `application ls` on the 2nd node and notice installed flag in the output. 
After the sync run the command again and you will see that the application is installed.

## Documentation update

N/A

### DIFF
--- a/crates/context/primitives/src/client/sync.rs
+++ b/crates/context/primitives/src/client/sync.rs
@@ -107,12 +107,12 @@ impl ContextClient {
                             .install_application_from_url(source, metadata, None)
                             .await?
                     }
-                    _ => {
-                        // fixme! we shouldn't assume both nodes run on the same machine
-                        self.node_client
-                            .install_application_from_path(source.path().into(), metadata)
-                            .await?
-                    }
+                    _ => self.node_client.install_application(
+                        &application.blob.bytecode,
+                        application.size,
+                        &source.into(),
+                        metadata,
+                    )?,
                 };
 
                 if application.id != derived_application_id {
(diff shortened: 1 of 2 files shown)


crates/context/primitives/src/client/sync.rs
```diff
@@ -107,12 +107,12 @@ impl ContextClient {
...
-                        self.node_client
-                            .install_application_from_path(source.path().into(), metadata)
-                            .await?
-                    }
+                    _ => self.node_client.install_application(
+                        &application.blob.bytecode,
+                        application.size,
+                        &source.into(),
+                        metadata,
+                    )?,
```
miraclx: as mentioned privately, if `from_url` fails, we should still fallback to normal install since "context config sync" shouldn't have a hard requirement of application blob availability

crates/context/primitives/src/client/sync.rs
```diff
@@ -103,16 +103,26 @@ impl ContextClient {
...
+                                metadata,
+                            )?,
+                        }
                     }
+                    _ => self.node_client.install_application(
+                        &application.blob.bytecode,
+                        application.size,
+                        &source.into(),
+                        metadata,
+                    )?,
```
miraclx: while this is good, it's worth mentioning you can deduplicate the install_application calls

consider a pattern like this:

```rust
let value = match item {
    "thing" => fallible().ok(),
    _ => None,
};

let value = match value {
    Some(value) => value,
    None => fallthrough_case(),
};
```

non-blocking of course, and up to you

miraclx (approved):

PR #1345, opened 2025-07-25
feat: impl append-only log for state deltas

miraclx:
## Description

Implement an append-only log for tracking state deltas produced from individual member mutations, and build a state sync mechanism backed by it.

## Test plan

Run 2 nodes in sync, then disconnect them and advance them independently, then reconnect them, You'll observe they eventually reconcile to the same state.

Or run `stage/test/prep.sh`

## Documentation update

None needed

### DIFF
--- a/crates/node/src/sync/state.rs
+++ b/crates/node/src/sync/state.rs
@@ -15,7 +15,7 @@ use tracing::debug;
 use super::{Sequencer, SyncManager};
 
 impl SyncManager {
-    #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


frdomovic (approved):
looks good

crates/node/src/sync/state.rs
```diff
@@ -15,6 +15,7 @@ use tracing::debug;
 use super::{Sequencer, SyncManager};
 
 impl SyncManager {
+    #[allow(dead_code)]
```
```suggestion
    #[expect(dead_code, reason = "superceded by delta sync")]
```

PR #1346, opened 2025-07-25
fix: revert docker changes

xilosada:
## Description
A few things required in the services were unitentionally removed during the optimization. This PR brings them back:

## Test Plan
`docker-compose -f docker-compose.prod.yml up`

## Documentation Update
Complete flow will be added to the next docs version

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -5,8 +5,6 @@ x-node-defaults: &node-defaults
     target: runtime
     secrets:
       - gh-token
-    args:
-      - CALIMERO_WEBUI_FETCH=1
   user: root
   volumes:
     - calimero_auth_node:/calimero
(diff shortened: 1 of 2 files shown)


alenmestrov (approved):

docker-compose.prod.yml
miraclx, line 1: seems none of the changes here will be needed if we changed the dockerfile to have this instead, dropping `CMD`:

```Dockerfile
ENV APP_NAME=${APP_NAME}
ENTRYPOINT ${APP_NAME} --config /etc/calimero/auth.toml --verbose
```

docker-compose.prod.yml
```diff
@@ -5,6 +5,8 @@ x-node-defaults: &node-defaults
     target: runtime
     secrets:
       - gh-token
+    args:
+      - CALIMERO_WEBUI_FETCH=1
```
miraclx: build args will be passed through from the host by default

xilosada:
```suggestion
```

docker-compose.prod.yml
```diff
@@ -79,14 +81,18 @@ services:
       target: runtime
       secrets:
         - gh-token
+      args:
+        - CALIMERO_AUTH_FRONTEND_FETCH=1
```
xilosada:
```suggestion
```

miraclx (approved):

PR #1348, opened 2025-07-27
(fix): added validation logic for local credentials.

dotandev:
## Description

Attempts to resolve https://github.com/calimero-network/core/issues/1135.

This PR cleans up credential parsing in the config crate and also generates a default keypair if absent:

-It ensures that the system doesn't fall back to the relayer when local credentials are specified but missing.

-It simplifies the credential parsing logic, reducing the risk of incorrect parsing due to overlapping credential formats.

-It cleans up the configuration file by omitting redundant information.

## Test Plan

N/A

## Documentation Update
N/A

### DIFF
--- a/.github/workflows/abi-guard.yml
+++ b/.github/workflows/abi-guard.yml
@@ -1,0 +1,63 @@
+name: ABI Guard
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches:
+      - feat/abi_v1
+      - main
+  push:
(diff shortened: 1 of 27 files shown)


miraclx (changes requested):
Hi, I'm not sure I understand this patch.

It's in no way related to the referenced issue? 🤷🏽‍♂️

antonpaisov:
@dotandev done, have a go :)

PR #1349, opened 2025-07-28
fix: added delete route and adjusted handler for uninstall app endpoint

alenmestrov:
# [Admin] short description

## Description

Added delete method type for uninstall application endpoint.
Adjusted handler to use query instead of body in request.

## Test plan

Run node in this branch, open admin-dashboard and try to delete installed application.

## Documentation update

N/A

### DIFF
--- a/crates/context/primitives/src/client/sync.rs
+++ b/crates/context/primitives/src/client/sync.rs
@@ -101,17 +101,24 @@ impl ContextClient {
 
                 let metadata = application.metadata;
 
-                let derived_application_id = match source.scheme() {
-                    "http" | "https" => {
-                        self.node_client
-                            .install_application_from_url(source, metadata, None)
-                            .await?
-                    }
-                    _ => {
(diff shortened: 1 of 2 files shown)


crates/server/src/admin/handlers/applications/uninstall_application.rs
```diff
@@ -1,22 +1,22 @@
...
+use calimero_primitives::application::ApplicationId;
+use calimero_server_primitives::admin::UninstallApplicationResponse;
 
 use crate::admin::service::ApiResponse;
 use crate::AdminState;
 
 pub async fn handler(
     Extension(state): Extension<Arc<AdminState>>,
-    Json(req): Json<UninstallApplicationRequest>,
+    Path(application_id): Path<ApplicationId>,
```
miraclx, line 15: if `UninstallApplicationRequest` is no longer needed, we should remove it

PR #1355, opened 2025-07-31
chore: release 0.8.0-rc.1

fbozic:
## Description
Prerelease

## Test Plan
n/a

## Documentation Update
n/a

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,11 +8,13 @@ on:
       - Cargo.toml
       - Cargo.lock
       - "crates/**"
+      - ".github/workflows/release.yml"
   pull_request:
     paths:
       - Cargo.toml
       - Cargo.lock
       - "crates/**"
+      - ".github/workflows/release.yml"
 
 permissions:
   contents: read
@@ -21,6 +23,8 @@ env:
   BINARIES: |-
     merod
     meroctl
+  AUTH_BINARIES: |-
+    mero-auth
 
 jobs:
   prepare:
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -21,6 +23,8 @@ env:
   BINARIES: |-
     merod
     meroctl
+  AUTH_BINARIES: |-
+    mero-auth
```
miraclx, line 27: I'm guessing this is to prevent the docker image build?

fbozic: yes, in `0.8.0` release we need to expand release workflow to release `auth` container as well

.github/workflows/release.yml
```diff
@@ -8,11 +8,13 @@ on:
       - Cargo.toml
       - Cargo.lock
       - "crates/**"
+      - ".github/workflows/release.yml"
```
miraclx, line 11: we should leave these out for the final merge

miraclx (approved):

.github/workflows/release.yml
fbozic: not sure if I agree 100%, so let's take a look at the situation for 0.8.0 release. in that PR I will have to modify `release.yaml` workflow in order to support new container release. if I would remove this now, in 0.8.0 release PR I would again have duplicate workflows running. I've added that path to both `release` and `release-ignored` so it should be in sync. also when you edit something in `release.yaml` it will run required binary builds.

frdomovic (approved):

.github/workflows/release.yml
fbozic: I'll merge as it is, we can adjust as we go along

PR #1359, opened 2025-08-01
fix: auth docker setup

alenmestrov:
# [Docker] short description

## Description

Fixed naming of auth crate and set user to root for auth

## Test plan

Run docker-compose -f docker-compose.prod.yml up

## Documentation update

N/a

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -79,6 +79,7 @@ services:
       target: runtime
       secrets:
         - gh-token
+    user: root
     volumes:
       - calimero_auth_data:/data
     environment:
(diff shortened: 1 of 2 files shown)


docker-compose.prod.yml
```diff
@@ -79,6 +79,7 @@ services:
       target: runtime
       secrets:
         - gh-token
+    user: root
     volumes:
       - calimero_auth_data:/data
```
miraclx, line 84: it's worth mentioning this is due to the volume being owned by the host, but used by the guest, so we have a priviledge mismatch

this is okay for now, but a better solution should eventually be devised

miraclx (approved):

PR #1360, opened 2025-08-01
fix(auth): add WebSocket authentication support

xilosada:
## Description

This pull request implements WebSocket authentication for the `node` service, designed to work behind a Traefik reverse proxy using its `forwardauth` middleware. The required method for clients is to pass their JWT as a `token` query parameter in the connection URL.

**Example URL:**
`ws://<host>/ws?token=<YOUR_JWT>`

### Implementation and Challenges

The primary challenge was integrating WebSocket's stateful handshake with Traefik's `forwardauth`, which is inherently stateless.

1.  **Query Parameter Method (Chosen Solution):**
    *   This approach maintains a clean separation of concerns. The `auth` service acts as a simple gatekeeper, validating the token passed in the `X-Forwarded-Uri` header.
    *   Upon successful validation (`200 OK`), Traefik completes the request by connecting the client to the `node` service, which handles the WebSocket connection.
    *   This is the standard and most robust solution for this architecture.

2.  **`Sec-WebSocket-Protocol` Header (Abandoned):**
    *   An experiment was conducted to "smuggle" the token in this header.
    *   This was proven to be fundamentally incompatible with the `forwardauth` model. `forwardauth` breaks the direct, point-to-point handshake required by the WebSocket protocol, as it cannot correctly proxy the necessary handshake headers between the `auth` service and the final `node` service.

## Test Plan

A shell script has been added to test the functionality.

**Prerequisite:**
* Ensure the Docker Compose environment is running.
* Generate a valid JWT from the auth service.
* Execute the test script with the token:
Run
```
>    ./scripts/test-websocket-auth.sh <YOUR_JWT_TOKEN>
```
Confirm that the script reports a successful connection with the token and a correctly failed connection without it.

## Documentation Update
N/A


crates/auth/src/api/handlers/auth.rs
```diff
@@ -377,6 +399,30 @@ pub async fn validate_handler(
...
+        .get("X-Forwarded-Uri")
+        .and_then(|value| value.to_str().ok())
+        .and_then(|uri_str| {
+            uri_str.split('?').nth(1).and_then(|query| {
+                query
+                    .split('&')
+                    .find(|param| param.starts_with("token="))
+                    .map(|param| &param[6..])
+            })
+        })
```
miraclx, line 423: is there a reason I'm missing that warrants supporting the `?token=` query param? why not just the header?

xilosada: browser's WebSocket API does not allow setting custom headers, there are other implementations but I don't want to add more dependencies to the client atm

xilosada: https://stackoverflow.com/a/77060459

It seems that we can smuggle the token in the Sec-WebSocket-Protocol header like Kubernetes. I'm ok with both options. Remote nodes will use ssl so in url is also fine. Thoughts?

alenmestrov: I'm up for the headers if they work

xilosada: I implemented Sec-WebSocket-Protocol in the last commit same [here](https://github.com/calimero-network/calimero-client-js/pull/59)

xilosada: Quick update, we are moving back to query parameters...
The problem is that Traefik drops headers like Sec-WebSocket-Protocol during forwardAuth, so the backend doesn't receive them. The client expects the same headers to be echoed back during the handshake, and when that doesn't happen, it disconnects. Query parameters work because it doesn’t depend on headers.

frdomovic (approved):
lgtm

crates/auth/src/api/handlers/auth.rs
miraclx: Thanks for checking, and the clarification.

PR #1372, opened 2025-08-05
feat(build): fetch pre-built admin dashboard from release assets

xilosada:
## Description

This PR updates the build process to fetch a pre-built admin dashboard artifact from the GitHub release assets, rather than downloading the entire source code. This change improves build efficiency and aligns with the new semantic-release workflow.

## Test plan

* cargo run -p merod -- --node-name node1 run
* Navigate to http://localhost:2428/admin-dashboard

## Documentation update
N/A

### DIFF
--- a/crates/server/build.rs
+++ b/crates/server/build.rs
@@ -8,8 +8,14 @@ use eyre::{bail, Context, OptionExt};
 use serde::Deserialize;
 
 #[derive(Deserialize)]
+struct Asset {
+    name: String,
+    browser_download_url: String,
+}
+
+#[derive(Deserialize)]
 struct Release {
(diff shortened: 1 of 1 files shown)


miraclx (commented):
Hm, tbh I think this changes way too much, and I'm not yet sure why.

Switching from downloading source.zip to build.zip should only be a few lines

1. `cached_path` helps us prevent redownloading the archive on every build, all the plumbing for managing the cache seems to have been removed
2. `CALIMERO_WEBUI_SRC` is named such to allow for URLs and local paths
3. did something break that warranted the patch to `target_dir`? an edge-case or something?

PR #1373, opened 2025-08-05
feat(build): fetch pre-built admin dashboard from release assets

xilosada:
## Description

This PR updates the build process to fetch a pre-built admin dashboard artifact from the GitHub release assets, rather than downloading the entire source code. This change improves build efficiency and aligns with the new semantic-release workflow.

## Test plan

 CALIMERO_WEBUI_FETCH=1 cargo run -p merod -- --node-name node1 run
 Navigate to http://localhost:2428/admin-dashboard/

## Documentation update
N/A transparent for the user

### DIFF
--- a/crates/server/build.rs
+++ b/crates/server/build.rs
@@ -83,23 +83,10 @@ fn try_main() -> eyre::Result<()> {
         }
     };
 
-    let mut builder = reqwest_compat::blocking::Client::builder().user_agent(USER_AGENT);
-
-    if let Some(token) = token {
-        let headers = [(
-            reqwest_compat::header::AUTHORIZATION,
-            format!("Bearer {token}").try_into()?,
-        )]
-        .into_iter();
-
-        builder = builder.default_headers(headers.collect());
-    }
-
     let webui_dir = if is_local_dir {
         Cow::from(Path::new(&*src))
     } else {
         let cache = Cache::builder()
-            .client_builder(builder)
             .freshness_lifetime(FRESHNESS_LIFETIME)
             .dir(target_dir()?.join("cache"))
             .build()?;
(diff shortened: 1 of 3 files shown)


crates/server/build.rs
```diff
@@ -59,37 +67,26 @@ fn try_main() -> eyre::Result<()> {
...
-            // atm, cached-path infers the archive type from the URL
-            // https://github.com/epwalsh/rust-cached-path/issues/68
-            // this is a temporary workaround for extraction support
-            zipball_url.push_str("?.zip");
+            let build_url = release
+                .assets
+                .into_iter()
+                .find(|asset| asset.name == "admin-dashboard-build.zip")
+                .map(|asset| asset.browser_download_url)
+                .ok_or_eyre("missing `admin-dashboard-build.zip` asset")?;
```
miraclx, line 80: instead of assuming a file name, can we maybe check that there is one file, and use that?

xilosada: You're right — the API response at
https://api.github.com/repos/calimero-network/admin-dashboard/releases/tags/v1.1.1
only includes the explicitly uploaded admin-dashboard-build.zip in the assets array.

The GitHub UI, however, shows additional auto-generated assets (Source code (zip) and Source code (tar.gz)), but these are not included in the assets list returned by the API. That's why I took this approach.

Taking the first asset blindly may work now, but it could break builds in the future if new assets are added.
I’d vote for checking explicitly for the ZIP name, it acts like a double check that we’re doing exactly what we intend, but both approaches are fine for me.

miraclx: Yeah I understand this, but;

1. the repo name is configurable
2. the version number is configurable

All these conspire to make this line too static.

We could either assume the repo has one asset, which is likely the case.

Or in the very least, make this configurable as well like the others.

It should be;

1. If there's one asset, use it.
2. If there are more than one, fail and ask the user to choose one via the env.

That's the better approach.

PR #1383, opened 2025-08-07
feat: setup is-authed endpoint

fbozic:
## Description
Add a dedicated dummy endpoint for checking if we are running behind auth or not. The health endpoint should be public most of the time. 

## Test Plan
n/a

## Documentation Update
n/a

### DIFF
--- a/crates/server/src/admin/service.rs
+++ b/crates/server/src/admin/service.rs
@@ -190,8 +190,9 @@ pub(crate) fn setup(
         )
         // Alias management
         .nest("/alias", alias::service())
-        // Health endpoints (previously unprotected)
         .route("/health", get(health_check_handler))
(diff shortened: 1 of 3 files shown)


crates/server/src/admin/service.rs
```diff
@@ -190,8 +190,10 @@ pub(crate) fn setup(
         )
         // Alias management
         .nest("/alias", alias::service())
-        // Health endpoints (previously unprotected)
+        // Health endpoints (unprotected)
```
miraclx: what does protection mean anymore 😅

fbozic: yeah good point, I'll remove it

frdomovic (approved):

PR #1388, opened 2025-08-08
feat(meroctl): make localhost 2528 default server if api or node not provided

chefsale:
Currently you have to specify either --node or --api to use meroctl, this adds localhost:2528 as default endpoint.

## Description
Before:
```
(venv) chefsale@chefsale-X670-GAMING-X-AX-V2 ~/w/merobox [0|1]> meroctl app list
+----------------------------------------------------------------------------------+
| ERROR                                                                            |
+==================================================================================+
| Error:                                                                           |
|    0: expected one of `--node` or `--api` to be set                     |
|                                                                                  |
| Location:                                                                        |
|    crates/meroctl/src/cli.rs:192                               |
|                                                                                  |
| Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it. |
| Run with RUST_BACKTRACE=full to include source snippets.                         |
+----------------------------------------------------------------------------------+

```

## Test plan

Now:
```
chefsale@chefsale-X670-GAMING-X-AX-V2 ~/w/core (add-default-endpoint)> ./target/debug/meroctl app ls
+----------------+--------+------+---------+
| Application ID | Source | Size | Blob ID |
+==========================================+
+----------------+--------+------+---------+
chefsale@chefsale-X670-GAMING-X-AX-V2 ~/w/core (add-default-endpoint)> ./target/debug/meroctl context ls
+------------+----------------+
| Context ID | Application ID |
+=============================+
+------------+----------------+
chefsale@chefsale-X670-GAMING-X-AX-V2 ~/w/core (add-default-endpoint)> ./target/debug/meroctl --api http://127.0.0.1:2528 app ls
+----------------+--------+------+---------+
| Application ID | Source | Size | Blob ID |
+==========================================+
+----------------+--------+------+---------+
```

## Documentation update
TBD

### DIFF
--- a/crates/meroctl/src/cli.rs
+++ b/crates/meroctl/src/cli.rs
@@ -5,7 +5,7 @@ use camino::Utf8PathBuf;
 use clap::{Parser, Subcommand};
 use comfy_table::{Cell, Color, Table};
 use const_format::concatcp;
-use eyre::{bail, OptionExt, Report as EyreReport, Result};
+use eyre::{bail, Report as EyreReport, Result};
 use serde::{Serialize, Serializer};
 use thiserror::Error as ThisError;
 use url::Url;
@@ -101,10 +101,10 @@ impl Environment {
         Self { output, connection }
     }
 
-    pub fn connection(&self) -> Result<&ConnectionInfo> {
+    pub fn connection(&self) -> &ConnectionInfo {
         self.connection
             .as_ref()
-            .ok_or_eyre("No connection available: this should not happen as we fall back to localhost:2528 by default")
+            .expect("Unable to create a connection.")
     }
 }
 
@@ -119,14 +119,7 @@ impl RootCommand {
         };
 
         let connection = if needs_connection {
-            match self.prepare_connection(output).await {
-                Ok(conn) => conn,
-                Err(err) => {
-                    let err = CliError::Other(err);
-                    output.write(&err);
-                    return Err(err);
-                }
-            }
+            Some(self.prepare_connection(output).await?)
         } else {
             None
         };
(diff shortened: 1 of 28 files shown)


miraclx (commented):
So long as the printed message when we're using default doesn't assume the user did any manual configuration and instead suggests to either spin up a node on the default port or configure one, this looks good to me.

Can we get an example of the error message?

crates/meroctl/src/cli.rs
```diff
@@ -197,8 +197,15 @@ impl RootCommand {
...
+                let default_url = "http://127.0.0.1:2528".parse()?;
+                let connection = authenticate_with_session_cache(
+                    &default_url,
+                    "default localhost server",
+                    output,
+                )
+                .await?;
+                Ok(Some(connection))
             }
             _ => Ok(None),
```
miraclx: I think we should be able to remove this case, and now this function should never return an option

crates/meroctl/src/cli.rs
```diff
@@ -104,7 +104,7 @@ impl Environment {
     pub fn connection(&self) -> Result<&ConnectionInfo> {
         self.connection
             .as_ref()
-            .ok_or_eyre("No node selected: set default node by running `meroctl node use <node_name>` or use `--node` or `--api` to manually select a node")
+            .ok_or_eyre("No connection available: this should not happen as we fall back to localhost:2528 by default")
```
miraclx: So we won't need this accommodation anymore

crates/meroctl/src/cli.rs
```diff
@@ -101,10 +101,10 @@ impl Environment {
         Self { output, connection }
     }
 
-    pub fn connection(&self) -> Result<&ConnectionInfo> {
+    pub fn connection(&self) -> &ConnectionInfo {
         self.connection
             .as_ref()
-            .ok_or_eyre("No node selected: set default node by running `meroctl node use <node_name>` or use `--node` or `--api` to manually select a node")
+            .expect("Unable to create a connection.")
```
miraclx: if it can fail, an error is preferrable to panicking

crates/meroctl/src/cli.rs
```diff
@@ -155,52 +148,51 @@ impl RootCommand {
...
+                    );
+                }
             }
-            _ => Ok(None),
+
+            // No active node set - fall back to default localhost server
+            let default_url = "http://127.0.0.1:2528".parse()?;
+            let connection =
+                authenticate_with_session_cache(&default_url, "default localhost server", output)
+                    .await?;
```
miraclx, line 195: I believe this is where authentication happens, so we need the custom error here if the currently reported error doesn't describe the issue appropriately.

PR #1406, opened 2025-08-18
feat: added NEAR_API_KEY env var to the client

chefsale:
Adding FastNEAR as default endpoint but needs API KEY.

## Description
^^^

## Test plan
It compiles.

## Documentation update
N/A for now.

### DIFF
--- a/crates/context/config/src/client/protocol/near.rs
+++ b/crates/context/config/src/client/protocol/near.rs
@@ -10,7 +10,7 @@ use near_jsonrpc_client::errors::{
 use near_jsonrpc_client::methods::query::{RpcQueryRequest, RpcQueryResponse};
 use near_jsonrpc_client::methods::send_tx::RpcSendTransactionRequest;
 use near_jsonrpc_client::methods::tx::RpcTransactionStatusRequest;
-use near_jsonrpc_client::JsonRpcClient;
+use near_jsonrpc_client::{auth, JsonRpcClient};
 use near_jsonrpc_primitives::types::query::QueryResponseKind;
 use near_jsonrpc_primitives::types::transactions::{RpcTransactionError, TransactionInfo};
 use near_primitives::account::id::ParseAccountError;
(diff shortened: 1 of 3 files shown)


crates/context/config/src/client/protocol/near.rs
```diff
@@ -147,6 +147,11 @@ impl<'a> NearTransport<'a> {
                     .expect("this is a valid header value");
             }
 
+            // Apply NEAR API key authentication if available
+            if let Ok(api_key) = env::var("NEAR_API_KEY") {
+                client = client.header(auth::ApiKey::new(&api_key).expect("valid API key"));
+            }
```
miraclx: Can we guarantee the RPC server doesn't use the Authentication header instead?

If not, let's maybe use both?

There's `auth::Authorization::bearer` method

crates/meroctl/src/cli.rs
```diff
@@ -48,7 +48,8 @@ pub const EXAMPLES: &str = r"
 #[command(author, version = CalimeroVersion::current_str(), about, long_about = None)]
 #[command(after_help = concatcp!(
     "Environment variables:\n",
-    "  CALIMERO_HOME    Directory for config and data\n\n",
+    "  CALIMERO_HOME    Directory for config and data\n",
+    "  NEAR_API_KEY    NEAR API key for blockchain operations\n\n",
```
miraclx: ~nit: indentation consistency

crates/meroctl/src/cli.rs
```diff
@@ -88,6 +89,11 @@ pub struct RootArgs {
 
     #[arg(long, value_name = "FORMAT", default_value_t, value_enum)]
     pub output_format: Format,
+
+    /// NEAR API key for blockchain operations
+    #[arg(long, value_name = "API_KEY")]
+    #[arg(env = "NEAR_API_KEY", hide_env_values = true)]
+    pub near_api_key: Option<String>,
```
miraclx: This isn't in use we can remove it

xilosada (approved):

PR #1419, opened 2025-08-21
chore: restore cargo deny rules, with alternative patch

miraclx:
## Description

Adding `LGPL-3.0-or-later` works fine, and in CI too it [appears](https://github.com/calimero-network/core/actions/runs/17117285922/job/48550750021)

## Test plan

--

## Documentation update

--

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: ["master"]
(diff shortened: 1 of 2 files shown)


chefsale:
Just for consistency sake, it would be nice to have the ci tools for checks have the same versions and behaviour like in the repo: https://github.com/calimero-network/core/actions/runs/17121025600/job/48561706870?pr=1420.

PR #1422, opened 2025-08-21
feat: support delete blob

### LINKED ISSUE #1421: support blob delete handler
## feature request related to a problem

Currently，admin/handlers/blob.rs:271  the method  `delete_handler` can not success  delete blob,
i want to fix it and  support blob delete handler success delete.

## Describe the solution you'd like

support blob delete handler success delete.

I want to try adding this feature.

## Additional context

Image

current the delete logic is get_blob_info

Image

847850277:
# support delete blob

close #1421 

## Description

support server can success delete blob

## Test plan

1、 run server

image

2、test delete blob

image

### DIFF
--- a/crates/server/src/admin/handlers/blob.rs
+++ b/crates/server/src/admin/handlers/blob.rs
@@ -283,54 +283,35 @@ pub async fn delete_handler(
         }
     };
 
-    // Get blob metadata first to include in headers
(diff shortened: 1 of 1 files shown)


miraclx (approved):
seeing as this fixes existing behaviour rather than introduces the concept of deleting blobs, this is okay with me

it's quite odd we had that previous behaviour for handling blob deletion

PR #1429, opened 2025-08-26
feat: implement prometheus metrics for network and context execution

fbozic:
## Description
Set up Prometheus metrics for network and context execution. Expose metrics on the existing server under `/metrics` path as a new server gated behind a feature flag.

## Test Plan
Run merod, curl `localhost:2428/metrics` and observe metrics.

## Documentation Update
n/a

### DIFF
--- a/scripts/stellar/deploy-devnet.sh
+++ b/scripts/stellar/deploy-devnet.sh
@@ -33,7 +33,7 @@ cd "${CONTRACTS_DIR}"
 # Start Stellar Quickstart container
 docker run --rm -d -p 8000:8000 \
     --name stellar \
-    stellar/quickstart \
+    stellar/quickstart:v508-latest \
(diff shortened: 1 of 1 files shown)


crates/context/src/handlers/execute.rs
```diff
@@ -169,16 +175,38 @@ impl Handler<ExecuteRequest> for ContextManager {
                     &guard,
                     &mut context,
                     executor,
-                    method.into(),
+                    method.clone().into(),
```
alenmestrov, line 178: do we need to clone here?

fbozic: image

yeah few lines below I'm using that method, only last usage doesn't have clone

alenmestrov: ok, ty

frdomovic (approved):

PR #1430, opened 2025-08-27
chore: rename apierror to clienterror and move to seperate crate to p…

chefsale:
…repare for extraction

## Description
Moving errors into seperate file to fix dependency hell of moving the client.

## Test plan
E2E

## Documentation update
N/A


xilosada:
🧹

Issue #1424, opened 2025-08-21
Replace WS with SSE


miraclx:
In the server, we have a websocket handler specifically for clients receiving node events, but that's counterintuitive because websockets are meant to be bidirectional, but we never actually send any messages to the node.

It would be ideal to change this to SSE instead.

After that, we need to update meroctl, make sure it works. And then follow with the JS client.

BipulLamsal:
hello i am interested, can i look into this issue?

CodeMan62:
as there is no activity since 5 days and on the website this issue show "up-to-grab" I will propose a PR asap no need to assign

antonpaisov:
@CodeMan62 can you please reach out to me on discord at @antonpaisov? 
@BipulLamsal please continue, but if there's no decent PR until the end of this week, I'll have to make it available for grabs again. Good luck!

PR #1434, opened 2025-08-29
feat: Replace WS with SSE server and SSE client for context event in meroctl

BipulLamsal:
Related to : Replace WS with SSE #1424 

## Description
Currently, context subscription is implemented using WebSockets. This PR introduces a Server-Sent Events (SSE) implementation to broadcast events based on node subscriptions. The API is now simplified to:
`GET /sse?contextId=<vec>`
The WebSocket implementation remains unchanged for backward compatibility, In later commit, we can remove the ws completely. Integration with actual node events in a live environment has not yet been verified.

## Test plan
- Verified that SEE emits hard coded events without auth 
- Tested SSE client with meroctl receiving events

## Documentation update
crates/server/Readme.md has to be updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1804,6 +1804,7 @@ dependencies = [
  "prometheus-client",
  "rand 0.8.5",
  "serde",
+ "serde_json",
  "tokio",
  "tracing",
 ]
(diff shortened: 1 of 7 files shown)


frdomovic:
@xilosada should we also connect this with client js PR ?

antonpaisov:
Yes, I already pointed @BipulLamsal to the client-js repo, waiting for the changes, also, please resolve the conflicts with master 🙌

PR #1460, opened 2025-09-03
ci: use stable 2024 edition version of `rustfmt`. Expand fmt options

sept-en:
# CI: use stable Rust formatting for code consistency

## Description

This PR updates our rustfmt setup to use the stable channel and the 2024 edition. This is essential to ensure code consistency.

The nightly `rustfmt`'s rules can change with the version upgrade, leading to unwanted formatting changes in PRs. This clutters diffs with stylistic fixes, making code reviews more difficult and hiding the true purpose of a PR.

By using the stable `rustfmt`, we guarantee fixed formatting behavior for Rust code. This means having less of stylystic changes in untouched code due to nightly version upgrades.

Additionally, we set to use the 2024 edition for formatting, we support an official style guide, making the code compaitble with [RFC 3338](https://rust-lang.github.io/rfcs/3338-style-evolution.html).

## Test plan

--

## Documentation update

The stylystic guide [STYLE.md](../blob/master/STYLE.md) has been updated.


chefsale:
Also, rerun it on the whole repo as it will fail. Not sure why we were using nightly in the first place.

xilosada:
I’m not strongly opinionated whether this should live in pre-commit or pre-push, in any case I'd also add the clippy check.

My main concern here is that, unlike JS repos, we don’t have a prepare script here, so I haven’t seen these hooks being triggered. Maybe worth clarifying how we expect them to be installed.

sept-en:
@chefsale the nightly version was likely to introduce more features to `rustfmt` that might cause new reformats, but usually unless formatting rules are explicitly changed in `rustfmt.toml`, the stable version won't reformat the previous nightly one.

However, as per your suggestion, I saw that stylystic change CI was not triggered indeed as there were no changes to Rust files. To trigger it, I added a commit with ghost changes (and reverted it later), so you can check the "CI / Rust" workflow successfully running in [493269a](https://github.com/calimero-network/core/pull/1460/commits/493269a3722679082cd89b712afa6fd79b8cff3b).

chefsale (approved):
Sounds good, ship it.

PR #1471, opened 2025-09-04
fix(Auth): Improve auth headers checking during JWT token verification

alenmestrov:
# [Auth] Fix for checking auth headers during verification

## Description

Changed to use host header for checking the request 

## Test plan

## Documentation update

### DIFF
--- a/crates/auth/src/api/handlers/auth.rs
+++ b/crates/auth/src/api/handlers/auth.rs
@@ -381,9 +381,16 @@ pub async fn validate_handler(
         Ok(claims) => {
             // Check node URL if token has node information
             if let Some(token_node_url) = &claims.node_url {
-                // Get the host from the request headers (use Host header since X-Forwarded-Host may not be set)
-                if let Some(host_header) = headers.get("host") {
-                    if let Ok(request_host) = host_header.to_str() {
+                // Get the original host from X-Forwarded-Host (set by Traefik) or fallback to Host header
+                let request_host = headers.get("X-Forwarded-Host")
+                    .or_else(|| headers.get("host"))
+                    .and_then(|h| h.to_str().ok());
+                    
+                if let Some(request_host) = request_host {
+                    // Skip validation if request is coming from internal auth service
+                    if request_host.starts_with("auth:") {
+                        debug!("Skipping host validation for internal auth service request");
+                    } else {
                         // Extract the host from the token's node URL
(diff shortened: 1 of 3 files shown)


crates/auth/src/api/handlers/auth.rs
```diff
@@ -295,18 +296,25 @@ pub async fn refresh_token_handler(
 
     // Check node URL if token has node information
     if let Some(token_node_url) = &refresh_claims.node_url {
```
chefsale, line 299: DRY: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

Let's make this a helper method instead.

Also we can use: https://doc.rust-lang.org/rust-by-example/error/option_unwrap/and_then.html

```
 if let Some(host_header) = headers.get("host")
        .and_then(|h| h.to_str().ok())
        .and_then(|host| {
            Url::parse(token_node_url).ok()
                .and_then(|token_url| token_url.host_str().map(|token_host| (host, token_host)))
        }) {
```

to simplify the code a bit.

crates/auth/src/api/handlers/auth.rs
```diff
@@ -374,18 +382,36 @@ pub async fn validate_handler(
...
-                            );
+                // Get the original host from X-Forwarded-Host (set by Traefik) or fallback to Host header
+                let request_host = headers
+                    .get("X-Forwarded-Host")
+                    .or_else(|| headers.get("host"))
+                    .and_then(|h| h.to_str().ok());
+
+                if let Some(request_host) = request_host {
+                    // Skip validation if request is coming from internal auth service
+                    if request_host.starts_with("auth:") {
```
chefsale: Use method from above.

crates/auth/src/auth/token/jwt.rs
```diff
@@ -270,14 +271,22 @@ impl TokenManager {
...
-            // Get referrer URL from headers
-            if let Some(referrer) = headers.get("referer") {
-                if let Ok(referrer_str) = referrer.to_str() {
-                    // Compare if referrer starts with the token's node URL
-                    if !referrer_str.starts_with(token_node_url) {
-                        return Err(AuthError::InvalidToken(
-                            "Token is not valid for this node".to_string(),
-                        ));
+            // Get the host from the request headers (fallback to Host header since X-Forwarded-Host may not be set)
+            if let Some(host_header) = headers.get("host") {
```
chefsale: ditto.

PR #1474, opened 2025-09-07
Runtime: implement first unit tests for logic/host functions.

sept-en:
# [product] short description

## Description

are required for this change.

## Test plan

## Documentation update

### DIFF
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -1081,3 +1081,353 @@ impl VMHostFunctions<'_> {
         Ok(bytes_read)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::store::{Key, Value};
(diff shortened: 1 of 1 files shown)


sept-en:
@chefsale I understand the concern wrt the file size, however, the idiomatic Rust convention is about placing unit tests in the same file as the code they're testing, inside a `mod test`. The main advantages for that are: 1) allowing to test the private methods/APIs; and 2) Maintainability and context awareness - both implementation and unit tests are part of the same context and usually should be updated side by side. The integration tests, on the other hand, are usually should be located in a separate `tests/` dir.

I agree that it's not that convenient to have very large files, but usually it's an indication that the specific module is quite big and has too many responsibilities. This specific module is unnecessary large, too, but the refactoring will be done later, in separate PRs. I think it's always better first having unit/regression tests before doing any refactoring.

PR #1476, opened 2025-09-08
feat(Meroctl): New application --watch command

alenmestrov:
# [Meroctl] Implemented application `watch` command

## Description

Added a new `meroctl app watch` command that monitors WASM file changes and automatically updates all contexts using a specific application. This addresses the need for bulk application updates across multiple contexts during development workflows.

**Key improvements over the existing `context update --watch`:**
- **Application-centric**: Targets all contexts using an application rather than a single context
- **No pre-installation required**: Automatically installs the application from the WASM file path
- **Automatic executor resolution**: Uses each context's default executor, eliminating the need for `--as` parameter
- **Dynamic context discovery**: Refreshes context list on each update to catch newly created contexts

**Motivation:** 
During development, developers often need to update the same application across multiple contexts. The existing `context update --watch` command only works for a single context, requiring manual updates for each context individually. This new command streamlines the development workflow by automatically propagating WASM changes to all relevant contexts.

**Dependencies:**
- Reuses existing `InstallDevApplicationRequest` and `UpdateContextApplicationRequest` APIs
- Uses the `notify` crate for file system watching (already in dependencies)
- Leverages existing alias resolution system for automatic executor detection

## Test plan

**Manual Testing:**

1. **Basic functionality test:**
   ```bash
   # Start watching a WASM file
   cargo run -p meroctl -- --node-name <node> --home <data_dir> app watch --path ./my-app.wasm
   
   # Modify the WASM file and save
   # Verify that all contexts using the application are updated
   ```

2. **Multi-context scenario:**
   ```bash
   # Create multiple contexts using the same application
   meroctl context create --application-id <app_id> --protocol near
   meroctl context create --application-id <app_id> --protocol near
   
   # Start watching
   meroctl app watch --path ./app.wasm --current-app-id <app_id>
   
   # Modify WASM file and verify both contexts are updated
   ```

3. **Error handling test:**
   ```bash
   # Test with contexts that have no default executor
   # Verify graceful error handling and continuation of other updates
   ```

4. **Metadata support test:**
   ```bash
   meroctl app watch --path ./app.wasm --metadata '{"version": "1.0.0"}'
   ```

**Comparison with existing command:**
- **Old:** `meroctl context update --context <specific_context> --path ./app.wasm --watch --as <executor>`
- **New:** `meroctl app watch --path ./app.wasm` (updates ALL contexts automatically)

**Expected output:**
```bash
Installing initial application from ./app.wasm to establish baseline...
Found 3 context(s) using application ABC123: [ctx1, ctx2, ctx3]
Watching for changes to ./app.wasm
📦 Installed new application version: DEF456
🔄 Updating 3 context(s) from ABC123 to DEF456...
✓ Updated context ctx1 with application DEF456
✓ Updated context ctx2 with application DEF456
✓ Updated context ctx3 with application DEF456
📊 Update complete: 3 successful, 0 failed
```

**End-to-end test case potential:**
Yes, this could be added to e2e tests by:
1. Creating multiple contexts with the same application
2. Starting the watch command in background
3. Modifying a test WASM file
4. Verifying all contexts are updated to the new application version
5. Checking that the watch process continues monitoring for additional changes

## Documentation update

**Required documentation updates:**

1. **CLI Reference Documentation:**
   - Add `meroctl app watch` command to the application management section
   - Include all command-line options: `--path`, `--metadata`, `--current-app-id`
   - Add usage examples for different scenarios

2. **Development Workflow Guide:**
   - Update development best practices to mention the new bulk update capability
   - Add section comparing `context update --watch` vs `app watch` use cases
   - Include troubleshooting section for common executor resolution issues

3. **API Documentation:**
   - No changes required (reuses existing APIs)

4. **Release Notes:**
   - Mention the new command in the next release notes
   - Highlight the improved developer experience for multi-context applications

---

> [!NOTE]
> Adds `meroctl app watch` to monitor a WASM file and bulk-update all contexts using the application on changes.
> 
> - **CLI – Applications**:
>   - Add `watch` subcommand wiring in `crates/meroctl/src/cli/app.rs` and module `watch`.
> - **New Command `app watch` (`crates/meroctl/src/cli/app/watch.rs`)**:
>   - Watches a WASM file via `notify`; on modify, installs new app blob (`InstallDevApplicationRequest`) and updates all contexts using the baseline app (`UpdateContextApplicationRequest`).
>   - Discovers target contexts dynamically by `application_id`; resolves executor via owned context identities.
>   - Options: `--path` (required), `--metadata`, `--current-app-id`.
>   - Handles no-contexts case with guidance; updates baseline `application_id` after successful rollouts.
> - **Tests**:
>   - Parsing and option handling for `WatchCommand`; utility behavior sanity checks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5ddada6ef672cfdf7eeee2d05b6ff604d49b6528. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


crates/meroctl/src/cli/app/watch.rs
```diff
@@ -252,6 +255,15 @@ async fn watch_app_and_update_contexts(
             "📊 Update complete: {} successful, {} failed",
             success_count, error_count
         )));
+
+        // If we had successful updates, update the baseline to track the new app ID
```
chefsale, line 228: only not sure why success_count > 0?

xilosada: @alenmestrov this doesn't match the PR summary:
```
Installing initial application from ./app.wasm to establish baseline...
Found 3 context(s) using application ABC123: [ctx1, ctx2, ctx3]
Watching for changes to ./app.wasm
📦 Installed new application version: DEF456
🔄 Updating 3 context(s) from ABC123 to DEF456...
✓ Updated context ctx1 with application DEF456
✓ Updated context ctx2 with application DEF456
✓ Updated context ctx3 with application DEF456
📊 Update complete: 3 successful, 0 failed
```

I'm not sure if the message provides much value but it should be if error_count is 0

alenmestrov: @chefsale The original logic was: only update the baseline application ID if we actually succeeded in updating contexts with the new application. Otherwise, keep tracking the old baseline.

However, we have two issues with this approach:
Initialization deadlock: If no contexts use the application at startup, we get stuck tracking an app ID that will never have contexts
Runtime deadlock: If all contexts disappear during development, we continue watching but never find anything to update

What I've implemented now is that in either of these cases, the process will stop gracefully and notify the user that no contexts were found to update.
This provides clear feedback to the user about what needs to happen next:
Create contexts using the application: meroctl context create --application-id <app_id> --protocol <protocol>
Then restart the watch command
This approach is much better than the original because:
No infinite loops looking for non-existent contexts
Clear user guidance on what to do next
Resource efficient - doesn't waste time watching when nothing can be updated
Predictable behavior - same logic at startup and during execution

sept-en (approved):

crates/meroctl/src/cli/app/watch.rs
```diff
@@ -0,0 +1,416 @@
...
+        // Refresh context list to catch any new contexts using the baseline app
+        let target_contexts = get_contexts_using_application(&client, &baseline_app_id).await?;
+
+        if target_contexts.is_empty() {
+            environment.output.write(&InfoLine(&format!(
+                "No contexts currently use application {}.",
+                baseline_app_id
+            )));
+            return handle_no_contexts_found(environment, &baseline_app_id);
+        }
```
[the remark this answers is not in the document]
chefsale: cc: @alenmestrov

chefsale (approved):
LGTM, but before merging lets fix:

The meroctl app watch command exits completely when no contexts are found during file monitoring. This terminates the continuous watch loop, preventing it from detecting and updating any new contexts created while the command is running.

crates/meroctl/src/cli/app/watch.rs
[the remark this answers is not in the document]
alenmestrov: well thats the idea right? why would you have running application watch if there are no contexts to be updated?

@chefsale

sept-en:
As discussed, we can have two different ways to handle the situation when there are no contexts using the application during the watch cycle:

1. We immediately exit with a message "No contexts found using application {app_id}". (Current approach).
2. We continuosly watch and check each N seconds if there are any contexts, notifying on each cycle that there are no contexts available (the number of attempts could also be limited).

As agreed, let's keep the first handling scenario for now and change to the second one later (if required).

PR #1489, opened 2025-09-10
chore(Merod): Decouple relayer

alenmestrov:
# [merod] Decouple relayer functionality into standalone calimero-relayer crate

## Description
This PR completely decouples the relayer functionality from the merod crate into a separate calimero-relayer crate. The relayer is now only available as a standalone service and has been removed entirely from merod.

## Key Changes:
- Created new calimero-relayer crate as a standalone binary
- Removed relay command and all relay-related code from merod
- Implemented standalone configuration system with:
  - Zero-setup defaults (Near testnet enabled out-of-the-box)
  - Environment variable support for all settings
  - TOML/JSON config file support
- Maintained all existing relay functionality and HTTP API compatibility

## Breaking Change:
merod relay command is no longer available. Users must now use the standalone calimero-relayer binary.

## Migration:
```bash
# OLD (no longer works):
merod --node-name my-node relay --listen 127.0.0.1:63529

# NEW (standalone):
# Works out-of-the-box with Near testnet enabled
calimero-relayer

# Optional: Enable additional protocols
export ENABLE_STARKNET=true
export ENABLE_ICP=true
export ENABLE_ETHEREUM=true
calimero-relayer
```

## Default Configuration
- **Near Protocol**: ✅ Enabled with working testnet credentials
- **Starknet**: ❌ Disabled (but has testnet credentials when enabled)
- **ICP**: ❌ Disabled (but has testnet credentials when enabled)
- **Ethereum**: ❌ Disabled (but has testnet credentials when enabled)

### Compilation Tests:
✅ cargo check -p merod - confirms merod compiles without relay dependencies
✅ cargo check -p calimero-relayer - confirms new crate compiles successfully
✅ cargo build --bin calimero-relayer - confirms standalone binary builds

### Functionality Tests:
✅ merod --help - confirms relay command is removed from CLI
✅ calimero-relayer --help - confirms standalone binary has correct interface
✅ All existing merod commands (init, run, config) work unchanged

### Integration Tests:
Run the following to verify the decoupling:

1. Test merod without relay:
```bash
cargo run -p merod -- --help
# Should show: config, init, run (no relay command)
```

2. Test standalone relayer:
```bash
# Should work out-of-the-box with Near testnet
cargo run --bin calimero-relayer

# Test enabling additional protocols
ENABLE_STARKNET=true cargo run --bin calimero-relayer
```

## Documentation update
Update documentation to reflect new realyer implementation
Remove reference to --relay command


crates/relayer/src/config.rs
```diff
@@ -0,0 +1,245 @@
...
+        secret_key: String,
+    },
+}
+
+impl Default for RelayerConfig {
+    fn default() -> Self {
+        let mut protocols = BTreeMap::new();
+
+        // Default NEAR configuration with testnet credentials
+        drop(protocols.insert("near".to_owned(), ProtocolConfig {
```
chefsale: lets move all this stuff into some constants.rs file

crates/relayer/src/config.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub contract_id: String,
+    /// Optional signing credentials for local signing
+    pub credentials: Option<ProtocolCredentials>,
+}
+
+/// Protocol-specific signing credentials
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "lowercase")]
+pub enum ProtocolCredentials {
+    Near {
```
chefsale: can this be some stuct and have a list of credentials which we add through some method?

crates/relayer/src/config.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+
+            // Override credentials from environment variables
+            let account_id_var = format!("{}_ACCOUNT_ID", prefix);
+            let public_key_var = format!("{}_PUBLIC_KEY", prefix);
+            let secret_key_var = format!("{}_SECRET_KEY", prefix);
+
+            if let Ok(account_id) = std::env::var(&account_id_var) {
+                let credentials = match protocol_name.as_str() {
+                    "near" => {
```
chefsale: again, lets use DRY here not to repeat the same implementations over and over again.

crates/relayer/src/main.rs
```diff
@@ -0,0 +1,363 @@
...
+use tokio::sync::{mpsc, oneshot};
+use tracing::{debug, info};
+use tracing_subscriber::fmt::layer;
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::{registry, EnvFilter};
+
+mod config;
+use config::{ProtocolCredentials, RelayerConfig};
+
+const DEFAULT_PORT: u16 = 63529; // Mero-rELAY = MELAY
```
chefsale: maybe move this into the constants.rs file?

crates/relayer/src/main.rs
```diff
@@ -0,0 +1,363 @@
...
+            ));
+
+            drop(protocols.insert(protocol_name.clone(), ClientLocalConfig { signers }));
+        }
+
+        let client_config = ClientConfig {
+            params,
+            signer: ClientSigner {
+                relayer: ClientRelayerSigner {
+                    url: "http://localhost:63529"
```
chefsale: constants.rs

crates/relayer/src/main.rs
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /// Convert relayer credentials to client credentials
+    fn convert_credentials(&self, creds: &ProtocolCredentials) -> EyreResult<Credentials> {
+        match creds {
+            ProtocolCredentials::Near {
+                account_id,
+                public_key,
+                secret_key,
+            } => Ok(Credentials::Near(NearCredentials {
```
chefsale: lets use DRY here as we repeat same code over and over.

crates/relayer/src/config.rs
```diff
@@ -0,0 +1,245 @@
...
+        // Default NEAR configuration with testnet credentials
+        drop(protocols.insert("near".to_owned(), ProtocolConfig {
+            enabled: true,
+            network: "testnet".to_owned(),
+            rpc_url: "https://rpc.testnet.near.org".parse().unwrap(),
+            contract_id: "calimero-context-config.testnet".to_owned(),
+            credentials: Some(ProtocolCredentials::Near {
+                account_id: "dev-1642425627065-33437663923179".to_owned(),
+                public_key: "ed25519:98GtfF5gBPUvBWNgz8N8WNEjXRgBhFLuSQ5MnFDEjJ8x".to_owned(),
+                secret_key: "ed25519:4YdVWc7hgBUWwE9kXd4SPKmCztbGkMdHfZL2fDWw8L7gCJmrYcWAjcvK5Wek94aKSGBdLKHb7DaKoXudp6BnTqCb".to_owned(),
```
chefsale: Also, let remove all secret keys and make them env vars. Except if tests. cc; @sept-en and @fbozic

crates/relayer/README.md
```diff
@@ -0,0 +1,196 @@
...
+**Default Secret Key Environment Variables:**
+```bash
+# For Near testnet to work out-of-the-box
+export NEAR_DEFAULT_SECRET_KEY="ed25519:3D4YudUQRE39Lc4JHghuB5WM8kbgDDa34mnrEP5DdTApVH81af7e2dWgNPEaiQfdJnZq1CNPp5im4Rg5b2rKtXFv"
+
+# Optional: For other protocols when enabled  
+export STARKNET_DEFAULT_SECRET_KEY="0x0178eb2a625c0a8d85b0a5fd69fc879f9884f5205ad9d1ba41db0d7d1a77950a"
+export ICP_DEFAULT_SECRET_KEY="MFECAQEwBQYDK2VwBCIEIJKDIfd1Ybt7xliQlRmXZGRWG8dJ1Dl9qKGT0pOhMwPjaE30"
+export ETHEREUM_DEFAULT_SECRET_KEY="0ac1e735c1ca39db4a9c54d4edf2c6a50a75a3b3dce1cd2a64e8f5a44d1e2d2c"
+```
```
sept-en: Pre-review comment:

Please remove all the secret keys from anywhere, even if those are randomly generated. The better approach would be to create `.env.example` file, where it will have its variables written like this:
```NEAR_DEFAULT_SECRET_KEY=<PUT_YOUR_SECRET_KEY_HERE>`

After that, guide in the README to create `.env` file from `.env.example`, e.g. "Create `.env` file template by `cp .env.example .env`. Fill the credentials...".

File: crates/relayer/README.md [hunk 9]
```diff
@@ -0,0 +1,196 @@
+# Calimero Relayer
+
+The `calimero-relayer` is a standalone relay server for external client interactions with the Calimero network. It forwards requests to the appropriate blockchain protocols based on its own configuration and operates independently of the main `merod` node.
```
sept-en, line 3:
```suggestion
The `calimero-relayer` is a standalone relay service for external client interactions with the Calimero network. It forwards requests to the appropriate blockchain protocols based on its own configuration and operates independently of the main `merod` node.
```

crates/relayer/README.md
```diff
@@ -0,0 +1,196 @@
...
+export ETHEREUM_DEFAULT_SECRET_KEY="0ac1e735c1ca39db4a9c54d4edf2c6a50a75a3b3dce1cd2a64e8f5a44d1e2d2c"
+```
+
+#### 2. Environment Variables (Custom Configuration)
+
+To customize protocols or credentials, use environment variables:
+
+```bash
+# Basic relayer settings
+export RELAYER_LISTEN="0.0.0.0:63529"
```
sept-en: IMO, better to rename the used variable to `RELAYER_LISTEN_URL`.

crates/relayer/README.md
```diff
@@ -0,0 +1,196 @@
...
+export ETHEREUM_SECRET_KEY="0ac1e735c1ca39db4a9c54d4edf2c6a50a75a3b3dce1cd2a64e8f5a44d1e2d2c"
+```
+
+#### 3. Configuration File
+
+Create a TOML or JSON configuration file:
+
+```toml
+# relayer-config.toml
+listen = "0.0.0.0:63529"
```
sept-en, line 110: as above, rather should be `listen_url`. Please rename in other places, too.

crates/relayer/README.md
```diff
@@ -0,0 +1,196 @@
...
+[protocols.near.credentials]
+type = "near"
+account_id = "your-account.testnet"
+public_key = "ed25519:..."
+secret_key = "ed25519:..."
+
+[protocols.starknet]
+network = "sepolia"
+rpc_url = "https://free-rpc.nethermind.io/sepolia-juno/"
+contract_id = "0x1b991ee006e2d1e372ab96d0a957401fa200358f317b681df2948f30e17c29c"
```
sept-en: Please no hardcoded contract addresses. If it's a random one, just fill with the obvious garbage data, e.g. `0xabcdcafecafecafe...abcd`

crates/relayer/README.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Ethereum  
+ENABLE_ETHEREUM=true
+ETHEREUM_ACCOUNT_ID=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
+ETHEREUM_SECRET_KEY=ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
+```
+
+## API
+
+The relayer exposes a single HTTP POST endpoint at `/` that accepts JSON requests in the format expected by the Calimero context configuration client.
```
sept-en: Please add the link/info where the expected format is defined.

crates/relayer/src/constants.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// Default port for the relayer service (Mero-rELAY = MELAY)
+pub const DEFAULT_PORT: u16 = 63529;
+
+/// Default listen address
+pub const DEFAULT_ADDR: SocketAddr =
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), DEFAULT_PORT);
+
+/// Default relayer URL for client configuration
+pub const DEFAULT_RELAYER_URL: &str = "http://localhost:63529";
```
sept-en, line 13: I think, this should use the `DEFAULT_PORT` const inside of it. Otherwise, could be accidentally changed.

crates/relayer/src/constants.rs
```diff
@@ -0,0 +1,103 @@
...
+        pub const PUBLIC_KEY: &str = "dummy";
+        pub const SECRET_KEY: &str = "dummy";
+    }
+
+    pub mod ethereum {
+        pub const ACCOUNT_ID: &str = "0x0000000000000000000000000000000000000000";
+        pub const SECRET_KEY: &str =
+            "0000000000000000000000000000000000000000000000000000000000000001";
+    }
+}
```
sept-en: I don't really like the approach that the credentials might be replaced with the dummy ones, if the protocol is enabled in the config, but no real credentials were provided. This way it won't perform any useful function for that protocol, but also would be quite misleading. IMO it's better to fail early if the specific protocol is enabled, but no credentials are provided.

crates/relayer/src/constants.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        pub const NAME: &str = "ethereum";
+        pub const DEFAULT_NETWORK: &str = "sepolia";
+        pub const DEFAULT_RPC_URL: &str = "https://sepolia.drpc.org";
+        pub const DEFAULT_CONTRACT_ID: &str = "0x83365DE41E1247511F4C5D10Fb1AFe59b96aD4dB";
+
+        /// Default testnet credentials (for testing only - production should use env vars)
+        pub const DEFAULT_ACCOUNT_ID: &str = "0x8ba1f109551bD432803012645Hac136c22C177ec";
+        // Note: Secret key removed - must be provided via environment variables
+    }
```
sept-en: as mentioned before, wrt the contract IDs, account IDs, public/secret keys. Also, I don't think there's any value of having default public/account IDs. Am I missing something?

crates/relayer/src/credentials.rs
```diff
@@ -0,0 +1,209 @@
...
+            protocols::ethereum::NAME => {
+                if let (Some(account_id), Some(secret_key)) = (&account_id, &secret_key) {
+                    if !account_id.is_empty() && !secret_key.is_empty() {
+                        return Some(ProtocolCredentials::Ethereum {
+                            account_id: account_id.clone(),
+                            secret_key: secret_key.clone(),
+                        });
+                    }
+                }
+            }
```
sept-en: I don't really like the duplication. For example, I think there could be another method in the trait - something like `get_credentials(protocol_name, Option<account_id>, public_key, secret_key)`. Then all this code will be a couple of lines instead and reduce the chance of making errors/typos. 

Feel free to defer to the follow-up PR.

crates/relayer/src/credentials.rs
```diff
@@ -0,0 +1,209 @@
...
+                            secret_key,
+                        });
+                    }
+                }
+            }
+            _ => {}
+        }
+
+        None
+    }
```
sept-en: In general, I think, maybe it's better to have `ProtocolCredentials` as the trait, then each Protocol implement this trait, implement `Default` trait for default credentials, etc. This way it would be more Rust-idiomatic way. Could be deferred to the follow-up PR.

crates/relayer/src/credentials.rs
```diff
@@ -0,0 +1,209 @@
...
+        })),
+        ProtocolCredentials::Ethereum {
+            account_id,
+            secret_key,
+        } => Ok(Credentials::Ethereum(EthereumCredentials {
+            account_id: account_id.clone(),
+            secret_key: secret_key.clone(),
+        })),
+    }
+}
```
sept-en: as mentioned above, if that to be refactored, `ClientCredentials` could rather implement `From<ProtocolCredentials>` trait.

crates/relayer/src/main.rs
```diff
@@ -0,0 +1,304 @@
...
+            let mut signers = BTreeMap::new();
+
+            // Create credentials based on protocol and what's available
+            let credentials = match protocol_config.credentials.as_ref() {
+                Some(creds) => self.convert_credentials(creds)?,
+                None => {
+                    // Generate dummy credentials for relayer-only mode
+                    // The relayer typically doesn't need real credentials for many operations
+                    self.generate_dummy_credentials(protocol_name)?
+                }
```
sept-en: if it doesn't need - let's not provide and let's not use these credentials (mentioned that above, too).

crates/relayer/src/main.rs
```diff
@@ -0,0 +1,304 @@
...
+            if let Ok(env_port) = env_port.parse() {
+                break 'port Some(env_port);
+            }
+            eprintln!(
+                "warning: invalid 'PORT' environment variable: '{}', ignoring..",
+                env_port
+            );
+        }
+        None
+    };
```
sept-en: Please no break labels, especially for that small piece of code. It's not justified here - makes it harder to read. Could be rather something like:

```rust
    // Check for PORT environment variable first
    let env_port = var("PORT")
        .ok()
        .and_then(|p| match p.parse() {
            Ok(port) => Some(port),
            Err(_) => {
                eprintln!("warning: invalid 'PORT' environment variable: '{}', ignoring..", p);
                None
            }
        });
```

crates/relayer/README.md
sept-en: Not resolved.
```suggestion
contract_id = "<PUT_CONTRACT_ID_HERE>"
```

crates/relayer/README.md
sept-en: not fixed. defer to another PR.

crates/relayer/src/constants.rs
sept-en: not resolved. defer to another PR.

crates/relayer/src/constants.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Ethereum Protocol configuration
+    pub mod ethereum {
+        pub const NAME: &str = "ethereum";
+        pub const DEFAULT_NETWORK: &str = "sepolia";
+        pub const DEFAULT_RPC_URL: &str = "https://sepolia.drpc.org";
+        pub const DEFAULT_CONTRACT_ID: &str = "0x83365DE41E1247511F4C5D10Fb1AFe59b96aD4dB";
+        // Note: All credentials must be provided via environment variables
+    }
```
sept-en, line 52: better to load these from some default config that is staged - e.g. `default_config.toml`. Defer to another PR.

sept-en (approved):
There are still some open comments, but to speed it up, let's defer to another PR.

LGTM!

crates/relayer/src/constants.rs
sept-en: Created #1506 to track.

crates/relayer/src/credentials.rs
sept-en: Add #1507 to track it.

PR #1494, opened 2025-09-16
feat: Implement global identity management with keypair store

### LINKED ISSUE #1350: Maintain keypair maps independently from contexts
At the moment, identities are exclusively scoped to the contexts they're a part of, which makes reusing identities across contexts somewhat painful, and many applications benefit from having the ability to maintain distinct groupings of identities while maintaining individual identities across them.

The idea then, is instead of keeping the private component of the keypair scoped to the context:

https://github.com/calimero-network/core/blob/ebae9ec0c626c6546e79c378686436fb865e952f/crates/store/src/types/context.rs#L102-L103

we create a new DB column ([ref](https://github.com/calimero-network/core/blob/ebae9ec0c626c6546e79c378686436fb865e952f/crates/store/src/db.rs#L17-L27)) for arbitrary keypairs.

This way, you can always receive and apply an arbitrary number of invitation payloads, which resolves the UX nightmare that is "identity creation before context joining on every subcontext"

As a result of this, we should have more robust identity management:

```
identity new
    (generates a new one, and displays the public key, same as currently)
identity rm/del/remove
    (removes the keypair from the local node, this will require some more thought as to what should happen with the key once removed. is it printed? backed up? worth considering)
identity ls/list
    (list the public keys)
identity export [id?]
    (should we print it? dump it to json?)
identity add/import
    (import from json? or CLI params?)
```

Some other questions:

Should we migrate context identity aliases to this global identity map? I'd personally vote no, but then do we create a new alias map for this one? is there any benefit to it?

WilliamNwoke:
# [product] short description

## Description
Resolves #1350
This PR implements a global identity management system that maintains keypair maps independently from contexts, resolving the UX nightmare of "identity creation before context joining on every subcontext".

Key Changes:

- New DB column: Added Keypairs column for storing global identities
- Keypair types: Created Keypair and related types for global identity storage
- ContextIdentity update: Changed from storing private keys to referencing global keypairs via keypair_ref
- CLI commands: Added full identity management CLI (identity generate/list/export/import/remove)
- API endpoints: Added REST API endpoints for identity operations
- Backward compatibility: Maintained existing context identity functionality

## Test plan

## Documentation update

yes

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -47,6 +47,7 @@ jobs:
           CALIMERO_WEBUI_FETCH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           ./apps/kv-store/build.sh
+          ./apps/kv-store-with-handlers/build.sh
           cargo build -p meroctl -p merod -p e2e-tests
 
       - name: Upload Build Artifacts (Binaries)
@@ -58,6 +59,7 @@ jobs:
             target/debug/merod
             target/debug/e2e-tests
             apps/kv-store/res/kv_store.wasm
+            apps/kv-store-with-handlers/res/kv_store_with_handlers.wasm
           retention-days: 2
 
   test:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2208,7 +2208,7 @@ dependencies = [
 
 [[package]]
 name = "calimero-version"
-version = "0.9.0-rc.3"
+version = "0.9.0"
 dependencies = [
  "eyre",
  "rustc_version 0.2.3",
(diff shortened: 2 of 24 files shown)


WilliamNwoke:
Please review @chefsale @sept-en

sept-en:
@WilliamNwoke thanks for the contribution!

We will review it soon.

WilliamNwoke:
> @WilliamNwoke thanks for the contribution!
> 
> We will review it soon.

Alright.

sept-en:
@WilliamNwoke we are currently refactoring and testing related modules. We are going to finish it soon and start reviewing the PR. Anton will also contact you on that.

PR #1495, opened 2025-09-16
chore(runtime): fixes, minor improvements. Add doc comments

sept-en:
# Runtime: fixes, minor improvements. Improve readability and add doc comments

## Description

Before, all the runtime functions, host functions, and other helpers used for host-guest interaction were not properly documented, and the readability was misleading in some places. This resulted into adding some bugs and regression during the implementation of other features recently (e.g. "storage collections bug"). Moreover, during the fixes for the aforementioned bugs, some new dangerous and non-idiomatic changes were introduced (e.g. implicitly allowing mutable access to guest memory even if it's not intended).

This PR addresses many of these concerns, targeting to reduce the error-prone approach, improve the documentation coverage, improve the code readability, and increase the maintainability of the code in general. The rest of the fixes will be addressed in the follow up PRs.

The more detailed changes are described below. These could be also used for a detailed description in the merge commit.

* Change `read_slice()` mutability to immutable.
* Add `read_guest_memory_slice_mut()` function.
* Change host functions to properly handle the input/output
  (immutable/mutable).
* Rename `read_slice()` -> `read_guest_memory_slice()`.
* Rename `read_str()` -> `read_guest_memory_str()`.
* Use `DIGEST_SIZE` const instead of hardcoded `32` values.
* Use KiB/MiB/GiB constants for setting VM Limits.
* Add doc comments.
* Add more tests.
* Other minor improvements.
* Fix several clippy warnings.

## Test plan

The existing unit tests should pass, also some new tests were added.

## Documentation update

--

### DIFF
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -1,19 +1,18 @@
 #![allow(single_use_lifetimes, unused_lifetimes, reason = "False positive")]
 #![allow(clippy::mem_forget, reason = "Safe for now")]
 
+use core::mem::MaybeUninit;
 use core::num::NonZeroU64;
 use core::{fmt, slice};
 use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap};
-use std::io::{Cursor, Read};
-use std::mem::MaybeUninit;
+use std::io::{Cursor, Error, Read};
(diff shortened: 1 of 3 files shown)


chefsale:
merged.

PR #1496, opened 2025-09-16
Clean up credentials parsing from config 

### LINKED ISSUE #1135: Clean up credentials parsing from config

CodeMan62:
## Description
- **Simplified Credential Parsing**: Replaced the `Credentials` enum with direct parsing using `serde_json::from_value` for each protocol (NEAR, Starknet, ICP, Ethereum). This allows for more dynamic credential handling without predefined variants.
- **Removed Legacy Structures**: Eliminated the `Credentials` enum and `RawCredentials` struct from `config.rs` to reduce complexity.
- **Enhanced Error Handling**: Added specific error messages for missing or invalid credentials, improving debugging and user feedback.
- **Added Tests**: Introduced unit tests in `client.rs` to verify credentials parsing and protocol selection logic, ensuring robustness.
## Test plan

I have added tests you can see

## Documentation update
N/A
closes #1135


CodeMan62:
Thanks for the quick reivew
1. I have mentioned a issue with the PR so the reason for these changes was the issue and the changes was explained in the issue 
2. I have tried my best at testing incase if you have any other format to test these changes let me know i would like to use that format.

CodeMan62:
bump

PR #1497, opened 2025-09-17
chore(runtime): logic mod refactoring

sept-en:
# Runtime: logic mod refactoring

## Description

The `logic` mod in runtime became quite enormous, so it was refactored. The existing host functions along with their related unit tests were moved into separate mods:
* blobs;
* governance;
* storage;
* system;
* utility.

Other than that, the PR contains other minor improvements: renaming some functions for better readability, split constants into a separate mod, etc.

## Test plan

The unit tests remain the same, now they are moved to the corresponding module of each host functions' group. E.g. previously all tests were running under the `logic::tests` umbrella, now running under the `logic::host_functions::tests` one. E.g. `logic::tests::test_commit` became `logic::host_functions::system::test_commit`.

## Documentation update

--


chefsale (approved):
🚢

PR #1500, opened 2025-09-18
fix(Server): dont log static files

alenmestrov:
# [Server] fix serving static files

## Description

Fixed issue with loading static files

## Test plan

Start node and navigate to admin-dashboard

## Documentation update

### DIFF
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -165,7 +165,7 @@ pub async fn start(
             .allow_private_network(true),
     );
 
-    // Middleware to log small JSON request bodies (without impacting handlers)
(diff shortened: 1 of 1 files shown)


crates/server/src/lib.rs
```diff
@@ -165,7 +165,7 @@ pub async fn start(
             .allow_private_network(true),
     );
 
-    // Middleware to log small JSON request bodies (without impacting handlers)
+    // Middleware to log small JSON request bodies (without impacting handlers) - COMMENTED OUT
```
sept-en, line 168: What does this updated comment means?

sept-en (commented):
LGTM!

I just had one question wrt the updated comment - please check.

PR #1502, opened 2025-09-24
feat(Relayer): Add endpoint for verifying NEAR wallets

alenmestrov:
# [product] short description

## Description

are required for this change.

## Test plan

## Documentation update


crates/auth/config/config.toml
```diff
@@ -72,4 +72,19 @@ enable_mock_auth = true
...
+
+# Relayer integration configuration
+[relayer]
+# Enable relayer for NEAR wallet verification (helps avoid rate limiting)
+enabled = true
+# URL of the relayer service (defaults to Calimero's managed relayer)
+url = "http://3.125.79.112:63529"
+# For self-hosted relayer, uncomment and modify:
+# url = "http://localhost:63529"        # Local development
+# url = "http://relayer:63529"          # Docker Compose
```
sept-en, line 85: I think it might be better to leave it empty by default, so the url is required to be specified. No need to specify the address of the manager relayer as this might change, too.

crates/auth/src/config.rs
```diff
@@ -385,6 +389,49 @@ impl Default for DevelopmentConfig {
...
+            enabled: false,
+            url: default_relayer_url(),
+            timeout_seconds: default_relayer_timeout(),
+            max_retries: default_relayer_max_retries(),
+        }
+    }
+}
+
+fn default_relayer_url() -> String {
+    "http://3.125.79.112:63529".to_string()
```
sept-en, line 424: Same, let's not expose this. It's better to either leave it empty and exit early if not provided, or point it to the local instance.

crates/auth/src/relayer.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// Request timeout duration
+    pub timeout: Duration,
+    /// Maximum number of retries
+    pub max_retries: u32,
+}
+
+impl Default for RelayerClientConfig {
+    fn default() -> Self {
+        Self {
+            relayer_url: Url::parse("http://localhost:63529").expect("valid URL"),
```
sept-en, line 59: better to use the config's default (e.g. `RelayerIntegrationConfig::Default` so you don't repeat default URLs multiple times. Also, the `expect` message should rather be "invalid relayer URL".

sept-en (commented):
Just a couple of minor comments on my side (didn't test it), while reviewing the PR with the conflicts.

PR #1512, opened 2025-09-30
chore: move cargo-mero from crates/ to tools/ and update workspace members and publish calimero-abi

chefsale:
## Description
moved the cargo mero into the tools dir, as preparation to moving the kv-store example here.

## Test plan
it builds

## Documentation update
needs to be published and then install will be simpler, nothing for now, all the same.


Cargo.toml
```diff
@@ -26,7 +27,7 @@ exclude = [
 resolver = "2"
 members = [
     "./crates/auth",
-    "./crates/cargo-mero",
+    # moved to tools/cargo-mero
```
antonpaisov:
```suggestion
```

antonpaisov (approved):

Cargo.toml
chefsale: fixed.

PR #1514, opened 2025-09-30
chore: Cleanup build warnings

antonpaisov:
Nothing much, updated the crate name + cleaned up some `cargo build` warnings.

## Description

na

## Test plan

na

## Documentation update

na


crates/server/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 use core::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
 
+use tower as _;
```
chefsale, line 4: do we even need this?

PR #1524, opened 2025-10-07
feat: add emitted events to state delta propagation and publish to the socket all changes received

chefsale:
Add emitted events to state delta propagation and publish to the socket all changes received 

Few things fixed:
1. events are now included and broadcasted via state mutations between nodes
2. statemutations event include the events on websocket publishing
3. statemutations are only published, not execution events which are deleted as it doesnt make sense to have it on view calls.

## Description
^^^

## Test plan
Tested locally with battleships.

## Documentation update
N/A at this moment

### DIFF
--- a/crates/primitives/src/events.rs
+++ b/crates/primitives/src/events.rs
@@ -28,13 +28,16 @@ pub enum ContextEventPayload {
 #[serde(rename_all = "camelCase")]
 pub struct StateMutationPayload {
     pub new_root: Hash,
-    pub events: Vec<ExecutionEvent>,
+    pub events: Option<Vec<ExecutionEvent>>,
 }
 
 impl StateMutationPayload {
     #[must_use]
     pub const fn with_root_and_events(new_root: Hash, events: Vec<ExecutionEvent>) -> Self {
-        Self { new_root, events }
+        Self {
+            new_root,
+            events: Some(events),
+        }
     }
 }
 


crates/node/src/handlers/network_event.rs
```diff
@@ -534,16 +540,32 @@ async fn handle_state_delta(
...
+            debug!(%context_id, events_count = events_payload.len(), "Re-emitting events to WS as StateMutation");
+            node_client.send_event(NodeEvent::Context(ContextEvent {
+                context_id,
+                payload: ContextEventPayload::StateMutation(
+                    StateMutationPayload::with_root_and_events(root_hash, events_payload),
+                ),
+            }))?;
+        } else {
+            debug!(%context_id, "No events after deserialization; skipping WS emit");
+        }
```
sept-en, line 568: why do we need to re-emit events?

crates/primitives/src/events.rs
```diff
@@ -22,19 +22,19 @@ pub struct ContextEvent {
...
     StateMutation(StateMutationPayload),
-    ExecutionEvent(ExecutionEventPayload),
 }
 
-#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct StateMutationPayload {
     pub new_root: Hash,
+    pub events: Vec<ExecutionEvent>,
```
sept-en: ideally, to make it optional to save some bytes, but could be done later.

sept-en (approved):
LGTM!

I had just a couple of comments/questions.

crates/node/src/handlers/network_event.rs
chefsale: We should re-emit to websocket as it wasn’t emitting the state delta changes we get from other peers, but good catch no need to broadcast to other node just to the socket. Have to check if I made a mistake here, but it seems to me like send_event is sent only to local bus aka websockets.

crates/primitives/src/events.rs
chefsale: Good catch. Will change it.

crates/node/src/handlers/network_event.rs
chefsale: Yeah, send_event only sends to local websocket.

PR #1528, opened 2025-10-09
chore: cleanup logging

fbozic:
## Description
Cleanup logging in node/context/network/server crates.

## Test Plan
n/a

## Documentation Update
n/a


crates/runtime/src/lib.rs
```diff
@@ -113,23 +118,33 @@ impl Module {
 
         let instance = match Instance::new(&mut store, &self.module, &imports) {
             Ok(instance) => instance,
-            Err(err) => return Ok(logic.finish(Some(err.into()))),
+            Err(err) => {
+                error!(%context_id, method, error=?err, "Failed to instantiate WASM module");
```
sh3ll3x3c, line 122: Do we want to emit errors here or in the caller?

fbozic: Let's log it here as well for starters. I want to see how it will look in Node Console for a node that has some load. I think later we should add metrics as well and reduce logging a bit.

PR #1545, opened 2025-10-21
feat(client/server): Add Server-Sent Events (SSE) support for context subscriptions

chefsale:
# [client-streaming] Add Server-Sent Events (SSE) support for context subscriptions

## Description

- Add SSE primitives module with request/response types
- Implement SSE service with subscription/unsubscription handlers
- Add SSE configuration to ServerConfig in both server and config crates
- Integrate SSE service into main server
- Update dependencies (tokio-stream, axum tokio feature)
- Update README with comprehensive SSE documentation and workflow diagrams
- Enable SSE by default in new node initializations

SSE provides an alternative to WebSockets for real-time context event subscriptions, with better firewall/proxy compatibility and automatic reconnection support. Both SSE and WebSocket support multiple context subscriptions per connection.

## Test plan
CI should pass, tested with battleships.

## Documentation update
@antonpaisov to add once battleships migrated to SSE.

### DIFF
--- a/crates/server/src/sse/handlers.rs
+++ b/crates/server/src/sse/handlers.rs
@@ -229,7 +229,8 @@ pub async fn sse_handler(
         .and_then(|s| s.split('-').next())
         .and_then(|id| id.parse::<ConnectionId>().ok());
 
-    let (commands_sender, commands_receiver) = mpsc::channel::<Command>(COMMAND_CHANNEL_BUFFER_SIZE);
+    let (commands_sender, commands_receiver) =
+        mpsc::channel::<Command>(COMMAND_CHANNEL_BUFFER_SIZE);
 
     let (session_id, session_state, is_reconnect) = if let Some(existing_session_id) = last_event_id
     {
--- a/crates/server/src/sse/storage.rs
+++ b/crates/server/src/sse/storage.rs
@@ -13,8 +13,8 @@ use super::session::PersistedSessionData;
 ///
 /// This function creates a unique key by taking the `session_id`, converting it
 /// to its big-endian byte representation, and using it as the first 8 bytes
-/// of the key's fragment (which is [`DIGEST_SIZE`] bytes). The remaining bytes 
-/// of the fragment are zero-padded. The entire key is scoped under 
+/// of the key's fragment (which is [`DIGEST_SIZE`] bytes). The remaining bytes
+/// of the fragment are zero-padded. The entire key is scoped under
 /// `SSE_SESSION_SCOPE` to prevent collisions.
 ///
 /// # Arguments


crates/server/src/sse/events.rs
```diff
@@ -0,0 +1,98 @@
...
+        let response = Response { body };
+
+        if let Err(err) = active_conn.commands.send(Command::Send(response)).await {
+            debug!(
+                %session_id,
+                %err,
+                "Failed to send event (connection likely closed, will retry on reconnect)",
+            );
+            // Don't break - session persists, connection might reconnect
+        };
```
sept-en, line 85: In this case, the event counter is already incremented. Should we increment the event counter after the send instead?

crates/server/src/sse/events.rs
```diff
@@ -0,0 +1,98 @@
...
+    while let Some(event) = events.next().await {
+        // Check if there's an active connection for this session
+        let active_connections = state.active_connections.read().await;
+        let Some(active_conn) = active_connections.get(&session_id).cloned() else {
+            debug!(%session_id, "No active connection, waiting for reconnection");
+            drop(active_connections);
+
+            // Wait a bit before checking again (connection might be reconnecting)
+            tokio::time::sleep(Duration::from_millis(100)).await;
+            continue;
```
sept-en, line 47: Won't we skip the current event in this case then?

crates/server/src/sse/handlers.rs
```diff
@@ -0,0 +1,342 @@
...
+use super::session::{now_secs, ActiveConnection, SessionState, SessionStateInner};
+use super::state::ServiceState;
+use super::storage::{delete_session, load_session, save_session};
+
+/// Handle subscription/unsubscription requests
+pub async fn handle_subscription(
+    Extension(state): Extension<Arc<ServiceState>>,
+    Json(request): Json<Request<serde_json::Value>>,
+) -> impl IntoResponse {
+    let session_id = match request.id.parse::<u64>() {
```
sept-en:
```suggestion
    let session_id = match request.id.parse::<ConnectionId>() {
```

crates/server/src/sse/handlers.rs
```diff
@@ -0,0 +1,342 @@
...
+                    (
+                        StatusCode::OK,
+                        Json(Response {
+                            body: ResponseBody::Result(serde_json::json!({
+                                "status": "unsubscribed",
+                                "contexts": ctxs.context_ids,
+                            })),
+                        }),
+                    )
+                }
```
sept-en: Maybe do the opposite - ignore the contexts that were not subscribed, and if there were at least one unsubscribed context - put it in the ok? Otherwise, I see it would be hard to figure out what was unsubscribed and what's not.

crates/server/src/sse/handlers.rs
```diff
@@ -0,0 +1,342 @@
...
+    let headers = request.headers();
+
+    // Check for Last-Event-ID header for reconnection
+    let last_event_id = headers
+        .get("Last-Event-ID")
+        .and_then(|v| v.to_str().ok())
+        .and_then(|s| s.split('-').next())
+        .and_then(|id| id.parse::<u64>().ok());
+
+    let (commands_sender, commands_receiver) = mpsc::channel::<Command>(32);
```
sept-en: Why this buffer size exactly? nevertheless, I think it's better to use a const or put that value into the `SseConfig`.

crates/server/src/sse/storage.rs
```diff
@@ -0,0 +1,58 @@
...
+use calimero_store::Store;
+use eyre::Result as EyreResult;
+
+use super::config::SSE_SESSION_SCOPE;
+use super::session::PersistedSessionData;
+
+/// Generate storage key for a session
+#[must_use]
+pub fn session_key(session_id: ConnectionId) -> GenericKey {
+    let mut fragment = [0u8; 32];
```
sept-en: Please use `calimero_primitives::common::DIGEST_SIZE`.
```suggestion
    let mut fragment = [0u8; DIGEST_SIZE];
```

crates/server/src/sse/storage.rs
```diff
@@ -0,0 +1,58 @@
...
+use calimero_store::key::Generic as GenericKey;
+use calimero_store::layer::{ReadLayer, WriteLayer};
+use calimero_store::slice::Slice;
+use calimero_store::Store;
+use eyre::Result as EyreResult;
+
+use super::config::SSE_SESSION_SCOPE;
+use super::session::PersistedSessionData;
+
+/// Generate storage key for a session
```
sept-en: I think it's important to add that info here, so it's clear the intention and the content of the function.
I'd also suggest to rename it to `session_storage_key` that reflects its purpose better.

```suggestion
/// Generates a generic storage key for a given session ID.
///
/// This function creates a unique key by taking the `session_id`, converting it
/// to its big-endian byte representation, and using it as the first 8 bytes
/// of the key's fragment. The remaining bytes of the fragment are zero-padded.
/// The entire key is scoped under `SSE_SESSION_SCOPE` to prevent collisions.
///
/// # Arguments
///
/// * `session_id` - The `ConnectionId` for which to generate a storage key.
///
/// # Returns
///
/// A `GenericKey` that can be used to uniquely identify session data in a storage backend.
```

crates/server/src/sse/events.rs
chefsale: Yes, events while disconnectted are ignored.

crates/server/src/sse/events.rs
chefsale: There was another issue related to this, incrementing the counter twice fixed now.

chefsale: Documented the behaviour a bit more in the code now.

crates/server/src/sse/handlers.rs
chefsale: Added comment and changed to const.

crates/server/src/sse/handlers.rs
chefsale: done

PR #1547, opened 2025-10-21
feat: implement blob storage application with file sharing capabilities

rtb-12:
# [Core] Add minimal blob API example application

## Description

Added a new `apps/blobs` application to demonstrate the Calimero blob storage API with a simple file sharing use case.

**Changes:**
- Implemented minimal file sharing contract showcasing blob API integration
- Core blob API demonstration: `env::blob_announce_to_context()` for network-wide blob discovery
- File operations: upload, delete, list, search, and blob ID retrieval
- Base58 encoding/decoding for safe blob ID serialization
- Comprehensive README with client-side integration examples and complete upload/download flows
- E2E workflow demonstrating multi-node blob synchronization

**Dependencies:**
- Added `apps/blobs` to workspace members in root `Cargo.toml`
- Uses existing workspace dependencies: `calimero-sdk`, `calimero-storage`, `bs58`

## Test plan

**Build verification:**
```bash
cd apps/blobs
./build.sh
```

**Expected output:** `res/blobs.wasm` and `res/abi.json` generated successfully

**Workflow testing:**
```bash
# Run the e2e workflow (requires Docker)
# Workflow verifies:
# - File upload with blob announcement
# - Multi-node blob discovery (Node 1 → Node 2)
# - Search, list, delete operations
# - Error handling
```

## Documentation update

- README.md included in `apps/blobs/` with complete API documentation and integration examples
- No external documentation updates required (example application for developers)

### DIFF
--- a/apps/blobs/src/lib.rs
+++ b/apps/blobs/src/lib.rs
@@ -389,7 +389,7 @@ impl FileShareState {
             .len()
             .map_err(|e| format!("Failed to get file count: {:?}", e))?;
 
-        let total_size = self.get_total_storage()?;
+        let total_size = self.get_total_files_size()?;
 
         let total_mb = (total_size as f64) / BYTES_PER_MB;
 


apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+use calimero_sdk::borsh::{BorshDeserialize, BorshSerialize};
+use calimero_sdk::serde::{Deserialize, Serialize};
+use calimero_sdk::{app, env};
+use calimero_storage::collections::UnorderedMap;
+
+// === BLOB ID ENCODING HELPERS ===
+
+/// Convert blob ID bytes to base58 string
+fn encode_blob_id_base58(blob_id_bytes: &[u8; 32]) -> String {
+    let mut buf = [0u8; 44];
```
sept-en: Please use consts instead of hardcoded numbers, everywhere in the code.

apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+pub struct FileRecord {
+    pub id: String,
+    pub name: String,
+    #[serde(serialize_with = "serialize_blob_id_bytes")]
+    pub blob_id: [u8; 32],
+    pub size: u64,
+    pub mime_type: String,
+    pub uploaded_by: String,
+    pub uploaded_at: u64,
+}
```
sept-en, line 102: Please add doc comments to each of the fields. Especially, so it's clear what's the format of `mime_type` and `uploaded_by`. Also I suggest that `uploaded_at` should have a notation of what's the timestamp - s/ms/unix/etc.

apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+
+/// Application state for the file sharing system
+#[app::state(emits = FileShareEvent)]
+#[derive(BorshDeserialize, BorshSerialize)]
+#[borsh(crate = "calimero_sdk::borsh")]
+pub struct FileShareState {
+    pub owner: String,
+    pub files: UnorderedMap<String, FileRecord>,
+    pub file_counter: u64,
+}
```
sept-en, line 120: Dito. Also, it's essential to specify the format of the `owner` string.

apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+            .map_err(|e| format!("Failed to access file: {:?}", e))?
+            .ok_or_else(|| format!("File not found: {}", file_id))?;
+
+        let file_name = file_record.name.clone();
+
+        // Remove the file from storage
+        self.files
+            .remove(&file_id)
+            .map_err(|e| format!("Failed to delete file: {:?}", e))?;
+
```
sept-en: You don't delete the blob though. But that'd require exposing these methods to the SDK.

apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn get_file(&self, file_id: String) -> Result<FileRecord, String> {
+        match self.files.get(&file_id) {
+            Ok(Some(file_record)) => Ok(file_record.clone()),
+            Ok(None) => Err(format!("File not found: {}", file_id)),
+            Err(e) => Err(format!("Failed to retrieve file: {:?}", e)),
+        }
+    }
+
+    /// Get blob ID for download (base58-encoded)
+    pub fn get_blob_id(&self, file_id: String) -> Result<String, String> {
```
sept-en: The name should be self-descriptive in this case. Otherwise, by default I'd expect that it returns a 32-byte array instead.

```suggestion
    /// Get blob ID for download (base58-encoded)
    pub fn get_blob_id_b58(&self, file_id: String) -> Result<String, String> {
```

apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+        let file_record = self.get_file(file_id)?;
+        Ok(encode_blob_id_base58(&file_record.blob_id))
+    }
+
+    /// Search files by name
```
sept-en: Please add more comments, specifying the returning Result (e.g. it's a vec of `FileRecord`s, not file names).

apps/blobs/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+                }
+            }
+        }
+
+        app::log!("Search for '{}' found {} results", query, results.len());
+
+        Ok(results)
+    }
+
+    /// Get total storage usage in bytes
```
sept-en: It's a little bit confusing. I'd say it's not the total storage usage - it's rather the total size of all files. The storage usage would be different and in the case of this app, would (also) include the size of the structs themselves.

apps/blobs/workflows/blobs-example.yml
```diff
@@ -0,0 +1,369 @@
...
+#   Blob Storage → returns blob_id → Client
+#   Client → contract.upload_file(blob_id, metadata) → Contract
+#   Contract → env::blob_announce_to_context(blob_id) → Network
+#   Network → All nodes discover blob → Distributed Storage
+#
+# DOWNLOAD FLOW:
+#   Client → contract.get_blob_id(file_id) → Contract
+#   Contract → returns blob_id → Client
+#   Client → blobClient.downloadBlob(blob_id, context_id) → Network
+#   Network → Finds peers with blob → Client receives data
```
sept-en: Please move it either to the top of the workflow, or to the README. the API workflow, IMO, better suits the README.

Alternatively, for each of the use cases, move these comments to be in-place with workflow tests (in blocks).

rtb-12: Updated this now , moved it too readme

apps/blobs/src/lib.rs
rtb-12: yes it was simply sum of total file sizes and added comment that clarifies the same

apps/blobs/src/lib.rs
rtb-12: Updated this and now added comments for the arguments and return types in all the functions

apps/blobs/src/lib.rs
rtb-12: Updated this

apps/blobs/src/lib.rs
rtb-12: Mentioned it in the comments and yes it needs to be done on sdk side as far as I know

apps/blobs/src/lib.rs
rtb-12: Done

apps/blobs/src/lib.rs
rtb-12: Done

apps/blobs/src/lib.rs
sept-en: @rtb-12 please rename the method accordingly, too. Could be maybe `get_total_files_size`

apps/blobs/src/lib.rs
sept-en: @rtb-12 please don't forget to update the workflow with this new name, and also a README. (the same for the `get_total_files_size()`.

sept-en (approved):
LGTM after the rest of the comments are resolved!

apps/blobs/src/lib.rs
rtb-12: Updated this and in workflow too.

apps/blobs/src/lib.rs
rtb-12: Renamed it now

PR #1549, opened 2025-10-22
fix(ws): handle non-WebSocket GET requests gracefully

chefsale:
Prevent node crash when regular GET requests are sent to /ws endpoint:

## Description

- Check if WebSocketUpgrade extraction succeeded before proceeding
- Validate presence of 'upgrade: websocket' header
- Return BAD_REQUEST (400) for invalid WebSocket upgrade attempts
- Return UPGRADE_REQUIRED (426) for regular GET requests
- Add debug logging for rejected requests
- Replace magic number with WS_COMMAND_CHANNEL_BUFFER_SIZE constant
- Remove redundant connection state check
- Improve error handling in socket close operation

This fixes a critical issue where sending a regular GET to /ws could crash the node.

## Test plan
- Started merod node on http://localhost:2528
- Sent GET request to http://localhost:2528/ws
- Node returned: HTTP 400 with message 'Invalid WebSocket upgrade request: Connection header did not include upgrade'
- Node continued running without crashing ✓

## Documentation update
N/A

### DIFF
--- a/crates/server/src/ws.rs
+++ b/crates/server/src/ws.rs
@@ -4,7 +4,9 @@ use std::collections::{HashMap, HashSet};
 use std::pin::pin;
 use std::sync::Arc;
 
+use axum::extract::ws::rejection::WebSocketUpgradeRejection;
 use axum::extract::ws::{CloseFrame, Message, WebSocket, WebSocketUpgrade};
+use axum::http::{HeaderMap, StatusCode};
 use axum::response::IntoResponse;
 use axum::routing::{get, MethodRouter};
 use axum::Extension;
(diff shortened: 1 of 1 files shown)


crates/server/src/ws.rs
```diff
@@ -413,3 +446,9 @@ macro_rules! mount_method {
 pub(crate) use mount_method;
 
 use crate::config::ServerConfig;
+
+/// WebSocket command channel buffer size
+///
+/// This controls how many WebSocket commands can be queued in the channel before
+/// the sender blocks. Should match SSE's COMMAND_CHANNEL_BUFFER_SIZE for consistency.
+const WS_COMMAND_CHANNEL_BUFFER_SIZE: usize = 32;
```
sh3ll3x3c, line 454: Should we then use the same parameter for both SSE and WS?

sh3ll3x3c (approved):

crates/server/src/ws.rs
chefsale: I put it in as seperate for now, but can merge it into one eventually. I think that at some point WS and SSE should have different functionalities, so makes sense to keep them seperated.

PR #1553, opened 2025-10-23
fix: enable state sync if out of sync for N deltas

chefsale:
^^^^

## Description
^^^

## Test plan
pray to god.

## Documentation update
N/A


crates/context/primitives/src/client.rs
```diff
@@ -731,13 +734,25 @@ impl ContextClient {
...
 
+        // Prune old deltas to prevent unbounded storage growth
+        // Keep only the last MAX_DELTAS_PER_MEMBER deltas
+        if height.get() > MAX_DELTAS_PER_MEMBER {
+            let old_height = height.get() - MAX_DELTAS_PER_MEMBER;
+            let old_key = key::ContextDelta::new(*context_id, *public_key, old_height);
+
+            // Ignore errors on delete - the delta may have already been pruned
+            let _ignored = handle.delete(&old_key);
+        }
```
sept-en, line 776: let's instead extract it to a separate function `prune_state_delta()`?

I'd also suggest to trace errors on delete - might be useful for the future debugging.

crates/node/src/handlers/network_event.rs
```diff
@@ -618,10 +618,37 @@ async fn handle_state_delta(
...
+    const MAX_DELTA_GAP_FOR_DELTA_SYNC: usize = 128;
+
     if let Some(known_height) = context_client.get_delta_height(&context_id, &author_id)? {
-        if known_height >= height || height.get() - known_height.get() > 1 {
-            debug!(%author_id, %context_id, "Received state delta much further ahead than known height, syncing..");
+        let gap = if height.get() > known_height.get() {
+            height.get() - known_height.get()
+        } else {
+            0
+        };
```
sept-en:
```suggestion
    let gap = height.get().saturating_sub(known_height.get());
```

crates/node/src/handlers/network_event.rs
```diff
@@ -618,10 +618,37 @@ async fn handle_state_delta(
...
+                %author_id,
+                %context_id,
+                gap=%gap,
+                "Received state delta with large gap (> {}), initiating sync",
+                MAX_DELTA_GAP_FOR_DELTA_SYNC
+            );
+
+            node_client.sync(Some(&context_id), Some(&source)).await?;
+            return Ok(());
+        }
```
sept-en: I guess, a typo?
```suggestion
        if known_height >= height && gap > MAX_DELTA_GAP_FOR_DELTA_SYNC {
            debug!(
                %author_id,
                %context_id,
                gap=%gap,
                "Received state delta with large gap (> {}), initiating sync",
                MAX_DELTA_GAP_FOR_DELTA_SYNC
            );

            node_client.sync(Some(&context_id), Some(&source)).await?;
            return Ok(());
        }
```

sept-en: Also, I think should be `if height > known_height`?

crates/context/primitives/src/client.rs
```diff
@@ -738,9 +742,61 @@ impl ContextClient {
...
+    /// * `context_id` - The context ID for the delta.
+    /// * `public_key` - The member whose deltas are being pruned.
+    /// * `current_height` - The current height to prune from.
+    fn prune_state_delta(
+        &self,
+        context_id: &ContextId,
+        public_key: &PublicKey,
+        current_height: &NonZeroUsize,
+    ) {
+        const MAX_DELTAS_PER_MEMBER: usize = 256;
```
sept-en: it's better to pass the `max_deltas_per_member` as a parameter, and then you can pass that specific constant from `handle_state_delta()`. This way it's easier for testing - e.g. for tests you would be able to issue 3 deltas and set the limit to 3 (instead of 256).

crates/node/src/sync/delta.rs
```diff
@@ -380,15 +380,30 @@ impl SyncManager {
...
+                            // This means we're missing too many deltas (likely pruned)
+                            const MAX_DELTA_GAP: usize = 128;
+                            if gap > MAX_DELTA_GAP {
+                                bail!(
+                                    "Delta gap too large ({} > {}), missing deltas for member {}",
+                                    gap,
+                                    MAX_DELTA_GAP,
+                                    member
+                                );
+                            }
```
sept-en, line 410: you re-define and re-use this constant `MAX_DELTA_GAP` twice (here and below), could you please move it to the module level - to the top of the file?

crates/context/primitives/src/client.rs
```diff
@@ -738,9 +746,61 @@ impl ContextClient {
...
 
+        // Prune old deltas to prevent unbounded storage growth
+        self.prune_state_delta(context_id, public_key, height, DELTA_RETENTION_LIMIT);
+
         Ok(())
     }
 
+    /// Prunes old state deltas to prevent unbounded storage growth.
+    ///
+    /// Keeps only the last 256 deltas per member. Older deltas are automatically deleted.
```
sept-en:
```suggestion
    /// Keeps only the last `max_deltas` deltas per member. Older deltas are automatically deleted.
```

sept-en (approved):

PR #1595, opened 2025-10-29
feat(wasm-abi): surface app state root metadata in emitted manifest

sh3ll3x3c:
Add app state metadata to the emitted ABI

## Description
Add app state metadata to the emitted ABI

## Test plan
Tested by building an app with the new ABI emitter and parsing the store data using the ABI data.

## Documentation update
N/A


chefsale:
@sh3ll3x3c can you just give an example run from the terminal input/output to understand a bit more whats going on before review. Just simple run and copy of logs is fine.

PR #1662, opened 2025-11-21
fix(context): add delay after join_context sync to ensure app state a…

sh3ll3x3c:
…vailability

# [product] short description

## Description

are required for this change.

## Test plan

## Documentation update

### DIFF
--- a/crates/context/src/handlers/join_context.rs
+++ b/crates/context/src/handlers/join_context.rs
@@ -147,5 +147,20 @@ async fn join_context(
     node_client.sync(Some(&context_id), None).await?;
     tracing::info!(%context_id, %invitee_id, "join_context: sync request sent successfully");
 
+    // CRITICAL FIX: Wait for app state to become available after initial sync
+    // When a node joins via open invitation, it needs time for:
+    // 1. Mesh formation (gossipsub)
+    // 2. DAG delta sync from peers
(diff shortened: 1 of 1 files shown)


crates/context/src/handlers/join_context.rs
```diff
@@ -147,5 +147,20 @@ async fn join_context(
...
+    // 3. App state materialization from deltas
+    //
+    // Without this delay, immediate function calls will fail with
+    // "Failed to find or read app state" because Root::fetch() returns None.
+    //
+    // This is especially critical for open invitation joins where the
+    // app was initialized (init() called) before the node joined.
+    tracing::info!(%context_id, %invitee_id, "join_context: waiting for app state to sync...");
+    tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
+    tracing::info!(%context_id, %invitee_id, "join_context: app state wait completed");
```
sept-en, line 163: I think the right way would be rather to add a wait timeout in the merobox workflow, as we do there with all the operations that require message propagation/sync. WDYT?

sh3ll3x3c: This issue persisted with the added timeouts in the workflow, which is why I started looking deeper into what's going on with the sync.

PR #1667, opened 2025-11-28
feat: implement relayer prometheus metrics

fbozic:
Add prometheus metrics to relayer

Spin up local relayer, point merod to local relayer, create context and fetch relayer metrics
```
❯ curl http://127.0.0.1:63529/metrics
# HELP mero_relayer_queue_depth Current depth of the request queue.
# TYPE mero_relayer_queue_depth gauge
mero_relayer_queue_depth 0
# HELP mero_relayer_http_requests_total Total number of HTTP requests.
# TYPE mero_relayer_http_requests_total counter
mero_relayer_http_requests_total_total{method="POST",path="/",status_code="200"} 2
# HELP mero_relayer_http_request_duration_seconds HTTP request duration in seconds.
# TYPE mero_relayer_http_request_duration_seconds histogram
mero_relayer_http_request_duration_seconds_sum{method="POST",path="/",status_code="200"} 5.921214
mero_relayer_http_request_duration_seconds_count{method="POST",path="/",status_code="200"} 2
mero_relayer_http_request_duration_seconds_bucket{le="0.001",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.002",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.004",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.008",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.016",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.032",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.064",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.128",method="POST",path="/",status_code="200"} 0
mero_relayer_http_request_duration_seconds_bucket{le="0.256",method="POST",path="/",status_code="200"} 1
mero_relayer_http_request_duration_seconds_bucket{le="0.512",method="POST",path="/",status_code="200"} 1
mero_relayer_http_request_duration_seconds_bucket{le="1.024",method="POST",path="/",status_code="200"} 1
mero_relayer_http_request_duration_seconds_bucket{le="2.048",method="POST",path="/",status_code="200"} 1
mero_relayer_http_request_duration_seconds_bucket{le="4.096",method="POST",path="/",status_code="200"} 1
mero_relayer_http_request_duration_seconds_bucket{le="8.192",method="POST",path="/",status_code="200"} 2
mero_relayer_http_request_duration_seconds_bucket{le="16.384",method="POST",path="/",status_code="200"} 2
mero_relayer_http_request_duration_seconds_bucket{le="+Inf",method="POST",path="/",status_code="200"} 2
# HELP mero_relayer_http_request_size_bytes HTTP request body size in bytes.
# TYPE mero_relayer_http_request_size_bytes histogram
mero_relayer_http_request_size_bytes_sum{method="POST",path="/"} 3154.0
mero_relayer_http_request_size_bytes_count{method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="64.0",method="POST",path="/"} 0
mero_relayer_http_request_size_bytes_bucket{le="128.0",method="POST",path="/"} 0
mero_relayer_http_request_size_bytes_bucket{le="256.0",method="POST",path="/"} 0
mero_relayer_http_request_size_bytes_bucket{le="512.0",method="POST",path="/"} 1
mero_relayer_http_request_size_bytes_bucket{le="1024.0",method="POST",path="/"} 1
mero_relayer_http_request_size_bytes_bucket{le="2048.0",method="POST",path="/"} 1
mero_relayer_http_request_size_bytes_bucket{le="4096.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="8192.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="16384.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="32768.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="65536.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="131072.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="262144.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="524288.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="1048576.0",method="POST",path="/"} 2
mero_relayer_http_request_size_bytes_bucket{le="+Inf",method="POST",path="/"} 2
# HELP mero_relayer_http_response_size_bytes HTTP response body size in bytes.
# TYPE mero_relayer_http_response_size_bytes histogram
# HELP mero_relayer_http_requests_active Number of HTTP requests currently being processed.
# TYPE mero_relayer_http_requests_active gauge
mero_relayer_http_requests_active 1
# HELP mero_relayer_protocol_requests_total Total number of protocol requests.
# TYPE mero_relayer_protocol_requests_total counter
mero_relayer_protocol_requests_total_total{protocol="near"} 2
# HELP mero_relayer_protocol_request_duration_seconds Protocol request duration in seconds.
# TYPE mero_relayer_protocol_request_duration_seconds histogram
mero_relayer_protocol_request_duration_seconds_sum{protocol="near",status="success"} 5.9200648750000008
mero_relayer_protocol_request_duration_seconds_count{protocol="near",status="success"} 2
mero_relayer_protocol_request_duration_seconds_bucket{le="0.001",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.002",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.004",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.008",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.016",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.032",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.064",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.128",protocol="near",status="success"} 0
mero_relayer_protocol_request_duration_seconds_bucket{le="0.256",protocol="near",status="success"} 1
mero_relayer_protocol_request_duration_seconds_bucket{le="0.512",protocol="near",status="success"} 1
mero_relayer_protocol_request_duration_seconds_bucket{le="1.024",protocol="near",status="success"} 1
mero_relayer_protocol_request_duration_seconds_bucket{le="2.048",protocol="near",status="success"} 1
mero_relayer_protocol_request_duration_seconds_bucket{le="4.096",protocol="near",status="success"} 1
mero_relayer_protocol_request_duration_seconds_bucket{le="8.192",protocol="near",status="success"} 2
mero_relayer_protocol_request_duration_seconds_bucket{le="16.384",protocol="near",status="success"} 2
mero_relayer_protocol_request_duration_seconds_bucket{le="+Inf",protocol="near",status="success"} 2
# HELP mero_relayer_protocol_errors_total Total number of protocol errors.
# TYPE mero_relayer_protocol_errors_total counter
# HELP mero_relayer_mock_operations_total Total number of mock relayer operations.
# TYPE mero_relayer_mock_operations_total counter
# EOF
```

---

> [!NOTE]
> Add Prometheus-based metrics to the relayer, including HTTP/protocol/queue/mock instrumentation and a new /metrics endpoint.
> 
> - **Relayer metrics/instrumentation**:
>   - Add `RelayerMetrics` (counters/gauges/histograms) for HTTP requests, protocol requests/errors/duration, queue depth, and mock operations (`crates/relayer/src/metrics.rs`).
>   - Introduce middleware `middleware::track_metrics` to record per-request HTTP metrics with path/method normalization (`crates/relayer/src/middleware.rs`).
>   - Instrument `RelayerService` to track queue depth, per-protocol request counts, durations, and errors; track mock operation counts; normalize protocol/method labels; import `Operation` for method detection (`crates/relayer/src/main.rs`).
>   - Expose a conditional `/metrics` endpoint using `prometheus_client::registry::Registry` and `axum::Extension`; include metrics in app state (`crates/relayer/src/main.rs`).
> - **Dependencies**:
>   - Add `prometheus-client` and `tower` to `crates/relayer/Cargo.toml`.
> - **Version**:
>   - Bump `calimero-version` to `0.10.0-rc.22`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 27f44340c11e1d201a87704451f511f69f1a559d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/relayer/src/main.rs
+++ b/crates/relayer/src/main.rs
@@ -17,7 +17,9 @@ use calimero_context_config::client::config::{
     ClientSelectedSigner, ClientSigner, Credentials, LocalConfig, RawCredentials,
 };
 use calimero_context_config::client::relayer::{RelayRequest, ServerError};
-use calimero_context_config::client::transport::{Transport, TransportArguments, TransportRequest};
+use calimero_context_config::client::transport::{
+    Operation, Transport, TransportArguments, TransportRequest,
+};
 use calimero_context_config::client::Client;
 use clap::Parser;
 use color_eyre::install;
@@ -47,47 +49,41 @@ use prometheus_client::registry::Registry;
 /// Only known protocols are tracked individually, others are aggregated as "other"
 fn normalize_protocol_name(protocol: &str) -> &str {
     match protocol {
-        protocols::near::NAME => "near",
-        protocols::starknet::NAME => "starknet",
-        protocols::icp::NAME => "icp",
-        protocols::ethereum::NAME => "ethereum",
-        protocols::mock_relayer::NAME => "mock-relayer",
+        protocols::near::NAME
+        | protocols::starknet::NAME
+        | protocols::icp::NAME
+        | protocols::ethereum::NAME
+        | protocols::mock_relayer::NAME => protocol,
         _ => "other",
     }
 }
 
 /// Normalize method name to prevent cardinality explosion
 /// Only known methods from handlers.rs are tracked individually, others are aggregated
-fn normalize_method_name(
-    operation: &calimero_context_config::client::transport::Operation<'_>,
-) -> &'static str {
-    use calimero_context_config::client::transport::Operation;
-
+fn normalize_method_name<'a>(operation: &'a Operation<'a>) -> &'a str {
     match operation {
         Operation::Read { method } => match method.as_ref() {
-            // Context-config query methods (from handlers.rs)
-            "application" => "application",
-            "application_revision" => "application_revision",
-            "members" => "members",
-            "members_revision" => "members_revision",
-            "has_member" => "has_member",
-            "privileges" => "privileges",
-            "get_proxy_contract" => "get_proxy_contract",
-            "fetch_nonce" => "fetch_nonce",
-            // Proxy query methods (from handlers.rs)
-            "proposals" => "proposals",
-            "proposal" => "proposal",
-            "get_number_of_active_proposals" => "get_number_of_active_proposals",
-            "get_number_of_proposal_approvals" => "get_number_of_proposal_approvals",
-            "get_proposal_approvers" => "get_proposal_approvers",
-            "get_context_value" => "get_context_value",
-            "get_context_storage_entries" => "get_context_storage_entries",
+            // Known read methods (from handlers.rs)
+            "application"
+            | "application_revision"
+            | "members"
+            | "members_revision"
+            | "has_member"
+            | "privileges"
+            | "get_proxy_contract"
+            | "fetch_nonce"
+            | "proposals"
+            | "proposal"
+            | "get_number_of_active_proposals"
+            | "get_number_of_proposal_approvals"
+            | "get_proposal_approvers"
+            | "get_context_value"
+            | "get_context_storage_entries" => method.as_ref(),
             _ => "other_read",
         },
         Operation::Write { method } => match method.as_ref() {
-            // Context-config mutate methods (from handlers.rs)
-            "mutate" => "mutate",
-            "proxy_mutate" => "proxy_mutate",
+            // Known write methods (from handlers.rs)
+            "mutate" | "proxy_mutate" => method.as_ref(),
             _ => "other_write",
         },
     }
@@ -248,17 +244,17 @@ impl RelayerService {
                     metrics.set_queue_depth(rx.len() as i64);
                 }
 
+                // Clone and normalize protocol name to prevent cardinality explosion
+                let protocol_str = request.protocol.clone();
+                let protocol_name = normalize_protocol_name(&protocol_str);
+                let start = std::time::Instant::now();
+
+                if let Some(ref metrics) = metrics {
+                    metrics.inc_protocol_requests(protocol_name);
+                }
+
                 // Check if this is a mock-relayer request
                 if request.protocol == protocols::mock_relayer::NAME {
-                    // Clone and normalize protocol name to prevent cardinality explosion
-                    let protocol_str = request.protocol.clone();
-                    let protocol_name = normalize_protocol_name(&protocol_str);
-                    if let Some(ref metrics) = metrics {
-                        metrics.inc_protocol_requests(protocol_name);
-                    }
-
-                    let start = std::time::Instant::now();
-
                     if let Some(ref mock) = mock_relayer {
                         debug!(
                             "Handling mock-relayer request for operation: {:?}",
@@ -332,13 +328,6 @@ impl RelayerService {
                     }
                 }
 
-                // Handle regular blockchain protocols
-                // Clone and normalize protocol name to prevent cardinality explosion
-                let protocol_str = request.protocol.clone();
-                let protocol_name = normalize_protocol_name(&protocol_str);
-
-                let start = std::time::Instant::now();
-
                 let args = TransportArguments {
                     protocol: request.protocol,
                     request: TransportRequest {
@@ -364,9 +353,6 @@ impl RelayerService {
                 // - Ok(Err(...)) = transport error (network, contract, etc.)
                 // - Err(...) = unsupported protocol
                 if let Some(ref metrics) = metrics {
-                    // Track the protocol request after we know the normalized name
-                    metrics.inc_protocol_requests(protocol_name);
-
                     let duration = start.elapsed();
                     let (status, error_type) = match &res {
                         Ok(Ok(_)) => ("success", None),
--- a/crates/relayer/src/middleware.rs
+++ b/crates/relayer/src/middleware.rs
@@ -7,6 +7,24 @@ use axum::response::IntoResponse;
 
 use crate::metrics::RelayerMetrics;
 
+/// RAII guard to ensure http_requests_active is decremented even on panic
+struct ActiveRequestGuard {
+    metrics: Arc<RelayerMetrics>,
+}
+
+impl ActiveRequestGuard {
+    fn new(metrics: Arc<RelayerMetrics>) -> Self {
+        metrics.inc_http_active();
+        Self { metrics }
+    }
+}
+
+impl Drop for ActiveRequestGuard {
+    fn drop(&mut self) {
+        self.metrics.dec_http_active();
+    }
+}
+
 /// Normalize HTTP method to prevent cardinality explosion
 /// Only standard HTTP methods are tracked individually, others are aggregated
 fn normalize_http_method(method: &axum::http::Method) -> &'static str {
@@ -57,7 +75,8 @@ pub async fn track_metrics(
     // Normalize HTTP method to prevent cardinality explosion
     let method = normalize_http_method(req.method());
 
-    metrics.inc_http_active();
+    // Use RAII guard to ensure active count is decremented even on panic/cancellation
+    let _active_guard = ActiveRequestGuard::new(metrics.clone());
 
     // Track request size, use 0 if content-length header is missing
     let request_size = req
@@ -71,7 +90,7 @@ pub async fn track_metrics(
 
     let response = next.run(req).await;
 
-    metrics.dec_http_active();
+    // Guard will decrement active count when dropped (even on panic)
 
     let status_code = response.status().as_u16().to_string();
     let duration = start.elapsed();


crates/relayer/src/main.rs
```diff
@@ -226,14 +293,45 @@ impl RelayerService {
...
+                // Track errors
+                if res.is_err() {
+                    metrics
+                        .protocol_errors_total
+                        .get_or_create(&ProtocolErrorLabels {
+                            protocol: protocol_name.clone(),
+                            error_type: "unsupported_protocol".to_string(),
+                        })
+                        .inc();
+                }
```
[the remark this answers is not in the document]
fbozic: Fixed

chefsale (approved):

crates/relayer/src/main.rs
```diff
@@ -33,22 +33,32 @@ use tracing_subscriber::{registry, EnvFilter};
...
+use prometheus_client::registry::Registry;
 
 /// Relayer service that handles incoming requests
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 struct RelayerService {
     config: RelayerConfig,
     mock_relayer: Option<MockRelayer>,
+    metrics: std::sync::Arc<RelayerMetrics>,
+    registry: std::sync::Arc<Registry>,
```
sept-en: Make it optional, please. If it's not optional, it will require much more effort for integration/e2e-testing.

crates/relayer/src/main.rs
```diff
@@ -174,29 +186,82 @@ impl RelayerService {
...
+                // Update queue depth
+                metrics.queue_depth.set(rx.len() as i64);
+                // Track protocol request
+                let protocol_name = request.protocol.to_string();
+                metrics
+                    .protocol_requests_total
+                    .get_or_create(&ProtocolLabels {
+                        protocol: protocol_name.clone(),
+                    })
+                    .inc();
```
sept-en: As you already have a dedicated `RelayerMetrics` struct, please implement helper methods there, so the code here will look like `metrics.protocol_requests_total_inc()`.

Please modify for other cases like that, too.

crates/relayer/src/main.rs
```diff
@@ -384,7 +526,12 @@ async fn main() -> EyreResult<()> {
         }
     }
 
-    let service = RelayerService::new(config);
+    // Initialize metrics registry
+    let mut registry = Registry::default();
+    let metrics = std::sync::Arc::new(RelayerMetrics::new(&mut registry));
+    let registry = std::sync::Arc::new(registry);
+
+    let service = RelayerService::new(config, Some(metrics), Some(registry));
```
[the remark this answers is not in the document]
fbozic: What are you talking about? How are metrics not optional, they are wrapped in Option?

crates/relayer/src/main.rs
```diff
@@ -174,39 +186,105 @@ impl RelayerService {
...
                             }
-                        });
+                        };
+
+                        let duration = start.elapsed();
+                        if let Some(ref metrics) = metrics {
+                            let (status, error_type) = match &res {
+                                Ok(Ok(_)) => ("success", None),
+                                Ok(Err(_)) => ("error", Some("mock_relayer_error")),
+                                Err(_) => unreachable!("Mock relayer never returns outer Err"),
```
sept-en: While currently the relayer never returns an `Err`, that might change in the future and it would be strange that the whole service might crash because of the metrics. I suggest the following: log a critical error and return a generic error status metric instead of crashing with `unreachable!()`

crates/relayer/src/main.rs
```diff
@@ -174,39 +186,105 @@ impl RelayerService {
...
+                                metrics.inc_protocol_errors(&protocol_name, error_type);
+                            }
+                        }
+
                         let _ignored = res_tx.send(res);
+
+                        // Update queue depth after processing completes
+                        if let Some(ref metrics) = metrics {
+                            metrics.set_queue_depth(rx.len() as i64);
+                        }
```
sept-en: You update `set_queue_depth(rx.len())` three times per loop iteration:
1. Immediately after receiving.
2. After mock processing.
3. After real protocol processing (e.g. near relayer). 

However, `rx.len()` always represents the amount of "tasks remaining in queue". Updating it at the end of the loop is functionally identical to updating it at the start of the next iteration (probably few nanoseconds apart). It unnecessary pollutes the metric. I suggest to update it only once, immediately after `rx.recv().await`.

crates/relayer/src/middleware.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    metrics.inc_http_active();
+
+    if let Some(content_length) = req.headers().get("content-length") {
+        if let Ok(size_str) = content_length.to_str() {
+            if let Ok(size) = size_str.parse::<u64>() {
+                metrics.record_http_request_size(&method, &path, size as f64);
+            }
+        }
+    }
```
sept-en: We indeed might have headers without "content-length" if we get a chunked response. However, this means that this metric might not be updated for many requests, it will be just silently ignored. I suggest at least log a `0` to keep metric "alive".

crates/relayer/src/middleware.rs
```diff
@@ -0,0 +1,59 @@
...
+    metrics: Option<Arc<RelayerMetrics>>,
+    req: Request,
+    next: Next,
+) -> impl IntoResponse {
+    let Some(metrics) = metrics else {
+        // No metrics available, just pass through
+        return next.run(req).await;
+    };
+
+    let path = req.uri().path();
```
sept-en: This is a potential target for a cardinality explosion. `req.uri().path()` returns the exact string sent by the client. You can overload and crash the relayer by sending requests to random paths - e.g. `/other1, /other2/, etc.`
This is because Prometheus creates a new memory entry for every unique path.
Instead, please fix it to maintain instead the list of allowed paths, merging all the other stuff into another path. Could be done in something like that (please verify on your side):

```suggestion
    let raw_path = req.uri().path();
    
    // Normalize path to prevent cardinality explosion.
    let path = match raw_path {
        "/" => "/",
        "/health" => "/health",
        "/metrics" => "/metrics",
        _ => "other", 
    };
```

crates/relayer/src/main.rs
```diff
@@ -174,39 +186,105 @@ impl RelayerService {
...
+                // Record queue depth (items waiting to be processed)
+                // rx.recv() has already removed the current item, so rx.len() is correct
+                if let Some(ref metrics) = metrics {
+                    metrics.set_queue_depth(rx.len() as i64);
+                }
+
+                let protocol_name = request.protocol.to_string();
+                if let Some(ref metrics) = metrics {
+                    metrics.inc_protocol_requests(&protocol_name);
+                }
```
sept-en: Same, possible cardinality explosion as the metric is recorded before the code checks if the protocol is supported/enabled. You can crash the relayer by flooding it with requests specifying random protocol names.

crates/relayer/src/main.rs
```diff
@@ -174,39 +186,105 @@ impl RelayerService {
...
-                        let res = mock.handle_request(request).await.map(Ok).map_err(|e| {
-                            debug!("Mock relayer error: {:?}", e);
-                            ServerError::UnsupportedProtocol {
-                                found: protocols::mock_relayer::NAME.into(),
-                                expected: vec![protocols::mock_relayer::NAME.into()].into(),
+
+                        let operation_name = format!("{:?}", request.operation);
+                        if let Some(ref metrics) = metrics {
+                            metrics.inc_mock_operations(&operation_name);
+                        }
```
sept-en: Another possible attack - the operation contains a method name inside, but here the method is accepted from the request without any validation. Meaning the same: you can flood the metrics with random method names and crash the relayer.

Similarly to other cases, I suggest having a list of allowed methods for read/write. You can find the currently used methods, for example, in `crates/relayer/src/mock/handlers.rs`.

crates/relayer/src/main.rs
```diff
@@ -33,22 +33,82 @@ use tracing_subscriber::{registry, EnvFilter};
...
+            "get_proxy_contract" => "get_proxy_contract",
+            "fetch_nonce" => "fetch_nonce",
+            // Proxy query methods (from handlers.rs)
+            "proposals" => "proposals",
+            "proposal" => "proposal",
+            "get_number_of_active_proposals" => "get_number_of_active_proposals",
+            "get_number_of_proposal_approvals" => "get_number_of_proposal_approvals",
+            "get_proposal_approvers" => "get_proposal_approvers",
+            "get_context_value" => "get_context_value",
+            "get_context_storage_entries" => "get_context_storage_entries",
```
sept-en: instead of repeating the same strings (which is also a subject to typo/mistake that would be hard to find) please just pipe the match, like that: `"application" | "application_revision" | "members" ... => method`.

Then, we have two match branches - for allowed methods and for other.

crates/relayer/src/main.rs
```diff
@@ -33,22 +33,82 @@ use tracing_subscriber::{registry, EnvFilter};
...
+            _ => "other_read",
+        },
+        Operation::Write { method } => match method.as_ref() {
+            // Context-config mutate methods (from handlers.rs)
+            "mutate" => "mutate",
+            "proxy_mutate" => "proxy_mutate",
```
sept-en: same as in the previous comment.

crates/relayer/src/main.rs
```diff
@@ -33,22 +33,82 @@ use tracing_subscriber::{registry, EnvFilter};
...
+
+/// Normalize protocol name to prevent cardinality explosion
+/// Only known protocols are tracked individually, others are aggregated as "other"
+fn normalize_protocol_name(protocol: &str) -> &str {
+    match protocol {
+        protocols::near::NAME => "near",
+        protocols::starknet::NAME => "starknet",
+        protocols::icp::NAME => "icp",
+        protocols::ethereum::NAME => "ethereum",
+        protocols::mock_relayer::NAME => "mock-relayer",
```
sept-en: same, you are basically matching `"near" => "near"`, instead pipe allowed protocol names and return `protocol` for them.

crates/relayer/src/main.rs
```diff
@@ -207,6 +333,12 @@ impl RelayerService {
                 }
 
                 // Handle regular blockchain protocols
+                // Clone and normalize protocol name to prevent cardinality explosion
+                let protocol_str = request.protocol.clone();
+                let protocol_name = normalize_protocol_name(&protocol_str);
```
sept-en: Unnecessary cloning.

```suggestion
                let protocol_name = normalize_protocol_name(&request.protocol);
```

crates/relayer/src/main.rs
```diff
@@ -174,29 +236,93 @@ impl RelayerService {
...
+                // rx.recv() has already removed the current item, so rx.len() is correct
+                if let Some(ref metrics) = metrics {
+                    metrics.set_queue_depth(rx.len() as i64);
+                }
+
                 // Check if this is a mock-relayer request
                 if request.protocol == protocols::mock_relayer::NAME {
+                    // Clone and normalize protocol name to prevent cardinality explosion
+                    let protocol_str = request.protocol.clone();
+                    let protocol_name = normalize_protocol_name(&protocol_str);
```
sept-en: no need to normalize the protocol name here as two lines before you already match it at: `                if request.protocol == protocols::mock_relayer::NAME {`. If you want to have it for consistency reasons, please remove the unnecessary clone as mentioned above, too.

crates/relayer/src/main.rs
```diff
@@ -33,22 +33,82 @@ use tracing_subscriber::{registry, EnvFilter};
...
+        _ => "other",
+    }
+}
+
+/// Normalize method name to prevent cardinality explosion
+/// Only known methods from handlers.rs are tracked individually, others are aggregated
+fn normalize_method_name(
+    operation: &calimero_context_config::client::transport::Operation<'_>,
+) -> &'static str {
+    use calimero_context_config::client::transport::Operation;
```
sept-en: as you are importing it inside the function anyway, better to move the import to the top of the file so the function signature will be more readable, too.

crates/relayer/src/main.rs
fbozic: Suggested change doesn't compile

sept-en (approved):

PR #1683, opened 2025-12-04
fix: state delta sync race condition

sh3ll3x3c:
# [node] Fix state delta race condition and modularize handler

## Description

Fixes a race condition where event handlers could fail because the application
blob was not yet available when processing state deltas from peers.

When a node receives a state delta via gossipsub, it may attempt to execute
event handlers before the application bytecode has been synced. This PR adds
explicit waiting for the application blob before handler execution, both for:
- Direct delta application
- Cascaded delta application (when pending deltas become ready)

Additionally, this PR refactors the large monolithic `handle_state_delta`
function into smaller, focused helper functions for better maintainability
and testability:
- `ensure_author_sender_key` - handles author identity key retrieval
- `decrypt_delta_actions` - handles artifact decryption
- `ensure_application_available` - waits for app blob with backoff
- `execute_cascaded_events` - handles cascaded event processing
- `parse_events_payload` / `execute_event_handlers_parsed` - event handling

## Test plan

- Unit tests added for helper functions (`decrypt_delta_actions`,
  `parse_events_payload`)
- E2E testing using merobox `workflow-examples/workflow-propagation-monitoring.yml` [workflow](https://github.com/calimero-network/merobox/pull/103)

## Documentation update

No public documentation update required. This is an internal fix to the node's
state synchronization logic.

---

> [!NOTE]
> Prevents handler execution before app blob availability and modularizes state delta processing; hardens key-share flow with mutual auth and deterministic roles.
> 
> - **State delta processing**:
>   - Ensure application blob availability before applying/handling deltas via `ensure_application_available` (also for cascaded events); early-bail to retry on rebroadcast.
>   - Refactor `handle_state_delta` into focused helpers: `ensure_author_sender_key`, `decrypt_delta_actions`, `choose_owned_identity`, `init_delta_store`, `execute_cascaded_events`, `parse_events_payload`.
>   - Improve cascade handling: track whether current delta applied and if handlers already executed; execute cascaded events only when app is available.
>   - Defer author key share when context root is zero (bootstrap) to avoid blocking state sync.
> - **Key share protocol** (`request_key_share_with_peer`):
>   - Implement deterministic initiator/responder roles, mutual challenge/response authentication, and ordered message sequencing; validate peer identity; store `sender_key`.
> - **Tests**:
>   - Add unit tests for `parse_events_payload` and `decrypt_delta_actions` (roundtrip and error cases).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 609f4f94828f477ba7bafa19ef7bc33c311d2dd5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/node/src/handlers/state_delta.rs
+++ b/crates/node/src/handlers/state_delta.rs
@@ -394,33 +394,47 @@ async fn init_delta_store(
     };
 
     if is_new_store {
-        if let Err(e) = delta_store_ref.load_persisted_deltas().await {
-            warn!(
-                ?e,
-                %context_id,
-                "Failed to load persisted deltas, starting with empty DAG"
-            );
-        }
+        let init_result = async {
+            if let Err(e) = delta_store_ref.load_persisted_deltas().await {
+                warn!(
+                    ?e,
+                    %context_id,
+                    "Failed to load persisted deltas, starting with empty DAG"
+                );
+            }
 
-        let missing_result = delta_store_ref.get_missing_parents().await;
-        if !missing_result.missing_ids.is_empty() {
+            let missing_result = delta_store_ref.get_missing_parents().await;
+            if !missing_result.missing_ids.is_empty() {
+                warn!(
+                    %context_id,
+                    missing_count = missing_result.missing_ids.len(),
+                    "Missing parents after loading persisted deltas - will request from network"
+                );
+            }
+
+            execute_cascaded_events(
+                &missing_result.cascaded_events,
+                node_clients,
+                &context_id,
+                &our_identity,
+                sync_timeout,
+                "initial load",
+                None,
+            )
+            .await
+        }
+        .await;
+
+        if let Err(err) = init_result {
             warn!(
                 %context_id,
-                missing_count = missing_result.missing_ids.len(),
-                "Missing parents after loading persisted deltas - will request from network"
+                ?err,
+                "Initial delta store setup failed - removing store to retry on next delta"
             );
+            // Remove the store so the next delta triggers a fresh init with retry
+            node_state.delta_stores.remove(&context_id);
+            return Err(err);
         }
-
-        execute_cascaded_events(
-            &missing_result.cascaded_events,
-            node_clients,
-            &context_id,
-            &our_identity,
-            sync_timeout,
-            "initial load",
-            None,
-        )
-        .await?;
     }
 
     Ok(DeltaStoreSetup {


crates/node/src/handlers/state_delta.rs
```diff
@@ -832,3 +967,86 @@ async fn request_key_share_with_peer(
...
+            if node_client.has_blob(&app.blob.bytecode)? {
+                info!(
+                    %context_id,
+                    %application_id,
+                    elapsed_ms = start.elapsed().as_millis(),
+                    "Application blob now available"
+                );
+                return Ok(());
+            }
+        }
```
sept-en, line 1347: if the peer deliberately (or by mistake) doesn't send you the given application, but sends you state deltas, you will have many tokio tasks that will be held until the `timeout` is reached. This way, you can cause DoS for the node.

I suggest that we rather need:
1. If the blob is missing, store the Delta (pending execution) in some temporary queue.
2. When blob download is finished, have a callback that will trigger the processing of all Deltas waiting for that specific `BlobId`.

This way instead of polling and consuming memory with threads, we will be using event-driven approach instead.

crates/node/src/handlers/state_delta.rs
```diff
@@ -61,13 +61,225 @@ pub async fn handle_state_delta(
...
+                    sync_timeout,
+                )
+                .await
+                {
+                    warn!(
+                        %context_id,
+                        delta_id = ?delta_id,
+                        error = %e,
+                        "Application not available for handler execution - handlers will be skipped"
+                    );
```
sept-en: here we are skipping the handlers, but the delta is already accepted. Meaning if the application is available later, we won't process these deltas again. What if we also store a queue for these deltas as mentioned before, or at least reject and don't apply the delta so it can retry later?

crates/node/src/handlers/state_delta.rs
sh3ll3x3c: Agreed, this is a problem. However a solution to this would require a bit of a re-architecture of the handler flow to to remove the polling and make it event driven which would be outside of the current scope. I'd definitely add it to the todo list and address it asap.

crates/node/src/handlers/state_delta.rs
```diff
@@ -61,14 +61,244 @@ pub async fn handle_state_delta(
...
+                } else {
+                    execute_event_handlers_parsed(
+                        &node_clients.context,
+                        &context_id,
+                        &our_identity,
+                        payload,
+                    )
+                    .await?;
+                }
```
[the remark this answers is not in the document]
sh3ll3x3c: Highly unlikely race condition window.

crates/node/src/handlers/state_delta.rs
```diff
@@ -757,78 +919,422 @@ async fn request_key_share_with_peer(
...
+        %application_id,
+        elapsed_ms = start.elapsed().as_millis(),
+        "Timeout waiting for application blob"
+    );
+
+    Err(eyre::eyre!(
+        "Application blob not available after {:?}",
+        timeout
+    ))
+}
```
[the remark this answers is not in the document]
sh3ll3x3c: To be addressed in a later PR.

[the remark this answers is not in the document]
sh3ll3x3c: https://github.com/calimero-network/core/issues/1688

crates/node/src/handlers/state_delta.rs
sh3ll3x3c: https://github.com/calimero-network/core/issues/1688

PR #1695, opened 2025-12-16
feat(auth): enable provider configuration status with user check

xilosada:
Add is_configured_with_users() method to AuthProvider trait to distinguish between technical configuration and actual user presence. This allows the providers API endpoint to show more accurate configuration status.

Changes:
- Add async is_configured_with_users() method to AuthProvider trait
- Update providers_handler to use new method for accurate status reporting
- Implement method for NEAR wallet provider (checks for NEAR keys)
- Implement method for user/password provider (checks for any users)

The new method allows each provider to define what 'configured' means:
- NEAR wallet: configured = has RPC URL AND has users with NEAR auth
- User/password: configured = has users (always technically configured)

---

> [!NOTE]
> Providers API now reports configuration based on both technical setup and presence of users/keys, with provider-specific logic and an efficient key existence check.
> 
> - **Auth API**:
>   - Update `providers_handler` in `crates/auth/src/api/handlers/mod.rs` to use async `is_configured_with_users()` per provider, logging errors and falling back to `is_configured()`; include `tracing` import.
> - **Providers**:
>   - Extend `AuthProvider` trait with async `is_configured_with_users()` to distinguish technical config from user presence in `crates/auth/src/providers/core/provider.rs`.
>   - Implement for `NearWalletProvider` and `UserPasswordProvider` to check for root keys matching their `auth_method`s via `KeyManager::has_any_key`.
> - **Storage**:
>   - Add `KeyManager::has_any_key(key_type, auth_methods)` for early-exit existence checks filtered by `auth_method` in `crates/auth/src/storage/key_manager.rs`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit afc3eb206663f09e5def0f2027952c0a2938e508. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/auth/src/storage/key_manager.rs
+++ b/crates/auth/src/storage/key_manager.rs
@@ -222,29 +222,34 @@ impl KeyManager {
             return Ok(false);
         }
 
-        // If no auth_method filter, any key counts as a match
-        if auth_methods.is_none() {
-            return Ok(true);
-        }
-
-        let auth_methods = auth_methods.unwrap();
+        let auth_methods = auth_methods;
 
         // Iterate through keys, deserializing one at a time until we find a match
         for key_path in keys {
             if let Some(data) = self.storage.get(&key_path).await? {
                 let key_data: Key = deserialize(&data)?;
 
-                // Check if this key matches the auth_method filter
-                let matches = if auth_methods.is_empty() {
-                    // Empty slice means we want keys with no auth_method
-                    key_data.auth_method.is_none()
+                // Skip revoked/invalid keys (consistent with get_key() behavior)
+                if !key_data.is_valid() {
+                    continue;
+                }
+
+                // If no auth_method filter, any valid key counts as a match
+                let matches = if let Some(auth_methods) = auth_methods {
+                    if auth_methods.is_empty() {
+                        // Empty slice means we want keys with no auth_method
+                        key_data.auth_method.is_none()
+                    } else {
+                        // Check if key's auth_method matches any of the provided methods
+                        key_data
+                            .auth_method
+                            .as_ref()
+                            .map(|m| auth_methods.contains(&m.as_str()))
+                            .unwrap_or(false)
+                    }
                 } else {
-                    // Check if key's auth_method matches any of the provided methods
-                    key_data
-                        .auth_method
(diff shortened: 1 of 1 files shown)


crates/auth/src/providers/impls/near_wallet.rs
```diff
@@ -536,9 +536,29 @@ impl AuthProvider for NearWalletProvider {
...
+        // Check if any root keys exist for this provider (auth_method = "near_wallet")
+        use crate::storage::models::KeyType;
+        let keys = self.key_manager.list_keys(KeyType::Root).await?;
+        // Check if any keys have the near_wallet auth method
+        let has_near_keys = keys.iter().any(|(_, key)| {
+            key.auth_method
+                .as_ref()
+                .map(|m| m == "near_wallet" || m == "near")
+                .unwrap_or(false)
+        });
```
sept-en: Listing keys may be consuming a lot of resources if we have many keys, idealy to peek and return early if it has at least any key. But then `key_manager` needs to have something like `has_any_key(prefix: KeyType::Root)` method. Or, alternatively, each key manager can keep track of `user_count` and have a respective method in the trait.

crates/auth/src/api/handlers/mod.rs
```diff
@@ -103,21 +103,24 @@ pub async fn health_handler(state: Extension<Arc<AppState>>) -> impl IntoRespons
...
-        .collect::<Vec<_>>();
+    let mut providers = Vec::new();
+
+    // Check each provider's configuration status
+    // Each provider implements its own logic for what "configured" means
+    for provider in state.0.auth_service.providers() {
+        let is_configured = provider
+            .is_configured_with_users()
+            .await
+            .unwrap_or_else(|_| provider.is_configured());
```
sept-en: I think it makes sense to add a `tracing::error` log on `unwrap_or_else()`, so if provider/DB didn't respond, we have a way to know it instead of just having a silent substitution (that might be hard to debug).

sept-en (approved):
LGTM!

Added some comments that I think might need to be addressed in the follow-up PRs.

PR #1726, opened 2026-01-09
fix: update relayer URL logic and enhance ConfigProtocol enum

rtb-12:
# [merod] Fix relayer URL selection for NEAR testnet in Docker mode

## Description

Fixes an issue where `merod init` incorrectly uses Docker host detection (`host.docker.internal:63529`) for all protocols when running inside Docker, breaking NEAR testnet which requires the remote EC2 relayer.

**Root Cause:** The `default_relayer_url()` function always detected Docker and returned `host.docker.internal:63529`, which works for local mock-relayer scenarios but fails for remote relayers like the NEAR testnet EC2 relayer.

**Solution:** Made relayer URL selection protocol-aware:
- **MockRelayer protocol**: Uses Docker host detection (preserves existing behavior for local relayer)
- **All other protocols** (Near, Ethereum, Starknet, ICP): Always uses EC2 URL (`http://63.179.161.75:63529`) regardless of Docker environment

**Changes:**
- Updated `InitCommand::run()` in `crates/merod/src/cli/init.rs` to check protocol before selecting relayer URL
- Added `PartialEq` and `Eq` traits to `ConfigProtocol` enum to enable protocol comparison

## Documentation update

No public documentation changes required.

---

> [!NOTE]
> - Protocol-aware relayer selection: `default_relayer_url(protocol)` in `crates/merod/src/defaults.rs` returns Docker host only for `ConfigProtocol::MockRelayer`, EC2 URL for all others
> - `InitCommand::run()` (`crates/merod/src/cli/init.rs`) updated to use `default_relayer_url(self.protocol)` when `--relayer-url` is not provided
> - `ConfigProtocol` now derives `PartialEq, Eq` and is re-exported from `crates/merod/src/cli.rs` for comparison and external use
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f87f281078ca9bd1895f24c407b464772558975. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/merobox-proposals.yml
+++ b/.github/workflows/merobox-proposals.yml
@@ -160,6 +160,12 @@ jobs:
           retention-days: 7
           if-no-files-found: warn
 
+      - name: Fail if workflow was cancelled
+        if: cancelled()
+        run: |
+          echo "NEAR proposals workflow was cancelled"
+          exit 1
+
   icp:
     name: Test ICP Proposals
     needs: [determine-image-tag, download-contracts]
@@ -248,6 +254,12 @@ jobs:
           retention-days: 7
           if-no-files-found: warn
 
+      - name: Fail if workflow was cancelled
+        if: cancelled()
+        run: |
+          echo "ICP proposals workflow was cancelled"
+          exit 1
+
   ethereum:
     name: Test Ethereum Proposals
     needs: [determine-image-tag, download-contracts]
@@ -334,6 +346,12 @@ jobs:
           retention-days: 7
           if-no-files-found: warn
 
+      - name: Fail if workflow was cancelled
+        if: cancelled()
+        run: |
+          echo "Ethereum proposals workflow was cancelled"
+          exit 1
+
   comment:
     name: Report Failed Workflows
     needs: [near, icp, ethereum]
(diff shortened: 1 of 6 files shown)


crates/merod/src/cli/init.rs
```diff
@@ -367,9 +367,15 @@ impl InitCommand {
...
-            .unwrap_or_else(defaults::default_relayer_url);
+        let relayer = self.relayer_url.unwrap_or_else(|| {
+            if self.protocol == ConfigProtocol::MockRelayer {
+                defaults::default_relayer_url()
+            } else {
+                defaults::DEFAULT_RELAYER_URL
+                    .parse()
+                    .expect("invalid default relayer URL")
+            }
+        });
```
sept-en: it's strange that in both cases we use `default_relayer_url()`, even though one is a function, and another is a constant.

Anyway, I think this behavior should be handled on the merobox side. So the config in the `merod` is always predictable - you have what you set in the config. Merobox though can modify the config file, or provide a cli argument, if needed.

xilosada: @rtb-12 @sept-en please agree if this will be solved in this PR or we create a different issue but let's fix merobox asap

rtb-12: I have updated the way default url is handle making the code look more clean , for this I think that as this fixes the issue right now and will unblock others using merobox , so it can be merged and for future we can create relayer handling as a proper feature in merobox having to pass it both by cli or explicitly mentioning it in the workflow

sept-en: Created #1730 to track it.

PR #1735, opened 2026-01-14
feat(node): always produce genesis delta. Other improvements. Refactoring

sept-en:
# Node: always produce genesis delta. Other improvements. Refactoring.

## Description

Node: always produce genesis delta. Other improvements. Refactoring.

* Node sync: always produce genesis delta even if there are no actions.
* Node sync: don't let node propagate `root_hash` if it has uninitalized context (`root_hash` is zero).
* `Hash`: add `is_zero()` helper function.
* Node: add `constants` mod and let the main node module to use these constants.

## Test plan

--

## Documentation update
--

---

> [!NOTE]
> **Node sync and initialization**
> 
> - Always create and persist a genesis `CausalDelta` on context init (parents `[0;32]`), set `dag_heads` to the new delta ID, even when `init` artifact has no actions (`create_context.rs`)
> - Log and store the init delta; remove prior fallbacks that used `root_hash` as a DAG head
> 
> **Node runtime**
> 
> - Introduce `node::constants` (heartbeat, pending-delta cleanup, blob cache limits) and replace hardcoded values in `lib.rs`
> - Skip hash heartbeat broadcasts when `context.root_hash.is_zero()` to avoid propagating uninitialized state
> - Refactor blob cache eviction and pending delta cleanup to use shared constants
> 
> **Primitives**
> 
> - Add `Hash::is_zero()` helper
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit acfad288c3e3853fdb693de9f1706e28656481de. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2470,7 +2470,7 @@ dependencies = [
 
 [[package]]
 name = "calimero-version"
-version = "0.10.0-rc.33"
+version = "0.10.0-rc.34"
 dependencies = [
  "eyre",
  "rustc_version 0.2.3",
--- a/crates/version/Cargo.toml
+++ b/crates/version/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "calimero-version"
-version = "0.10.0-rc.33"
+version = "0.10.0-rc.34"
 authors.workspace = true
 edition.workspace = true
 repository.workspace = true


crates/context/src/handlers/create_context.rs
```diff
@@ -298,55 +297,42 @@ async fn create_context(
...
-            context.dag_heads = vec![root_hash];
-            warn!(
-                context_id = %context.id,
-                "Init had empty artifact - using root_hash as dag_head"
-            );
-            None
-        }
+        Some(delta)
     } else {
         None
```
xilosada, line 337: Shouldn't we throw if the delta is `None`?
https://github.com/calimero-network/core/pull/1735/changes#diff-0f9abe5a3269f6eee9526de8c5d9d4b3f48882592f85c2980ef55e08dfd5ab3cR378

crates/node/src/lib.rs
```diff
@@ -344,6 +347,13 @@ impl Actor for NodeManager {
                             continue;
                         };
 
+                        // Do not broadcast heartbeat if the node is not initialized.
+                        // If the root hash is `[0; 32]` (represented as 1111...1111 in Base58), the node is uninitialized.
+                        if context.root_hash.is_zero() {
```
xilosada, line 352: good, this is what we were seeing in the logs 👍

crates/context/src/handlers/create_context.rs
sept-en: No, we return `None` If `outcome.root_hash` is also `None`. This means WASM execution in the init function did not commit any changes to the storage. Hence, the state of the context remains in the default state.

xilosada (approved):

PR #1760, opened 2026-01-31
feat(core): implement AppKey identity and mandatory bundle signing

rtb-12:
# [core] CIP-0001 bundle signature verification, AppKey identity, and signing toolchain

## Description

Implements core signature verification, application identity, and developer tooling from CIP-0001 (Calimero Application Lifecycle v1).

**Motivation:** CIP-0001 defines cryptographic update authority for applications. Only bundles signed by the same `signerId` may update an application, preventing unauthorized updates and namespace hijacking.

### Core Changes

**Signature Verification (`crates/node/primitives/src/bundle/`)**
- Added `SignerId` and `AppKey` types for cryptographic application identity
- Implemented RFC 8785 JSON canonicalization (JCS) for deterministic manifest hashing
- Implemented Ed25519 signature verification and `did:key` signerId derivation
- Wired signature verification into all bundle install paths
- Bundle `ApplicationId` now derived from `hash(package, signer_id)` instead of blob content
- Added `signer_id` field to `ApplicationMeta` in the store schema

**Bundle Manifest Types (`crates/node/primitives/src/bundle/mod.rs`)**
- `BundleManifest` struct with full CIP-0001 field support
- `BundleSignature` struct with `algorithm`, `publicKey`, `signature`, `signedAt` fields
- `BundleArtifact`, `BundleMetadata`, `BundleInterfaces`, `BundleLinks` support types

### Developer Tooling

**mero-sign CLI (`tools/mero-sign/`)**

New Rust CLI tool for signing bundle manifests, enabling CIP-0001 compliant bundle creation:

```bash
# Sign a manifest in-place
mero-sign sign manifest.json --key path/to/key.json

# Generate a new Ed25519 keypair
mero-sign generate-key --output my-key.json

# Derive the did:key signerId from a key file
mero-sign derive-signer-id --key my-key.json
```

**Key file format:**
```json
{
  "private_key": "<base64url-encoded 32-byte seed>",
  "public_key": "<base64url-encoded 32-byte public key>",
  "signer_id": "did:key:z6Mk..."
}
```

**Test Signing Key (`scripts/test-signing-key/`)**
- Pre-generated Ed25519 keypair for CI/development use
- `test-key.json` with known signerId for reproducible test bundles
- README with usage instructions and security warnings

**Updated Build Scripts**
- `apps/kv-store/build-bundle.sh` - automatically signs manifests with test key
- `apps/access-control/build-bundle.sh` - automatically signs manifests with test key

### Signed Manifest Structure

After signing, `manifest.json` contains:

```json
{
  "version": "1.0",
  "package": "com.calimero.kv-store",
  "appVersion": "1.0.0",
  "minRuntimeVersion": "1.0.0",
  "signerId": "did:key:z6Mk...",
  "wasm": { "path": "app.wasm", "size": 132912, "hash": null },
  "abi": { "path": "abi.json", "size": 1234, "hash": null },
  "signature": {
    "algorithm": "ed25519",
    "publicKey": "<base64url-encoded>",
    "signature": "<base64url-encoded>"
  }
}
```

### Main Behavior

- Same package + different signers → different ApplicationIds
- Same package + same signer + different versions → same ApplicationId
- Unsigned or tampered bundles are rejected at install time
- Signature verification derives `signerId` from embedded public key and validates it matches manifest's declared `signerId`

### Signing Algorithm

1. Load manifest JSON
2. Add `signerId` field (derived from public key as `did:key`)
3. Add `minRuntimeVersion` if not present
4. Canonicalize using RFC 8785 JCS (excluding `signature` field)
5. Compute SHA-256 hash of canonical bytes (= `bundleHash`)
6. Sign hash with Ed25519 private key
7. Add `signature` object with `algorithm`, `publicKey`, `signature` fields
8. Write updated manifest back to file

### Deferred

DSD (Desired State Document) and desired state anchoring are not defined or implemented. CIP-0001 migration items that depend on them are out of scope:
- Rewriting desired state references from appId to AppKey
- Handling multiple signers for the same appId via user/policy resolution

### Breaking Changes

- Bundles without a `signature` field are no longer installable
- Existing installations may require fresh stores due to the new ApplicationId derivation

### Dependencies

- `serde_json_canonicalizer` (added to workspace)
- `ed25519-dalek` with `rand_core` feature (for mero-sign)

### New Files
| Path | Description |
|------|-------------|
| `tools/mero-sign/Cargo.toml` | mero-sign CLI package configuration |
| `tools/mero-sign/src/main.rs` | CLI implementation with sign/generate-key/derive-signer-id commands |
| `scripts/test-signing-key/test-key.json` | Pre-generated test signing key |
| `scripts/test-signing-key/README.md` | Test key documentation and usage |

### Modified Files
| Path | Description |
|------|-------------|
| `Cargo.toml` | Added `mero-sign` to workspace members |
| `crates/node/primitives/src/bundle/mod.rs` | Bundle manifest and signature types |
| `crates/node/primitives/src/bundle/signature.rs` | Signature verification helpers |
| `apps/kv-store/build-bundle.sh` | Added manifest signing step |
| `apps/access-control/build-bundle.sh` | Added manifest signing step |

### Unit Tests

**Signature verification** (`crates/node/primitives/src/bundle/signature.rs` – 19 tests):
- `test_derive_signer_id_did_key` - did:key derivation format
- `test_canonicalize_manifest_removes_signature` - JCS excludes signature field
- `test_canonicalize_manifest_removes_transient_fields` - JCS excludes `_binary`, `_overwrite`
- `test_canonicalize_manifest_key_ordering` - RFC 8785 lexicographic ordering
- `test_compute_bundle_hash` - SHA-256 determinism
- `test_decode_public_key_valid` / `test_decode_public_key_invalid_length`
- `test_decode_signature_valid` / `test_decode_signature_invalid_length`
- `test_verify_ed25519_valid_signature` / `test_verify_ed25519_invalid_signature`
- `test_verify_manifest_signature_valid`
- `test_verify_manifest_signature_missing_signature`
- `test_verify_manifest_signature_wrong_algorithm`
- `test_verify_manifest_signature_signer_id_mismatch`
- `test_verify_manifest_signature_invalid_signature` (tampered manifest)
- `test_signer_id_derivation_is_stable`
- `test_bundle_hash_is_deterministic`
- `test_format_bundle_hash`

**mero-sign CLI** (`tools/mero-sign/src/main.rs` – 2 tests):
- `test_derive_signer_id_did_key`
- `test_generate_and_sign` - full key generation and signing flow

```bash
cargo test -p calimero-node-primitives --lib -- bundle::signature
cargo test -p mero-sign
```

### Integration Tests

(`crates/node/primitives/tests/bundle_installation.rs` – 4 tests):
- `test_bundle_installation_fails_without_signature`
- `test_bundle_installation_fails_with_invalid_signature`
- `test_bundle_installation_fails_signer_id_mismatch`
- `test_bundle_application_id_derived_from_package_and_signer_id`

```bash
cargo test -p calimero-node-primitives --test bundle_installation
```

### E2E Tests

**Merobox workflows** (require signed bundles):
- `apps/kv-store/workflows/bundle-invitation-test.yml` - Tests bundle installation and invitation flow with signed bundles
- `apps/kv-store/workflows/simple-store.yml` - Basic KV store operations

```bash
cd apps/kv-store
./build-bundle.sh  # Creates signed bundle
merobox bootstrap run workflows/bundle-invitation-test.yml \
    --near-devnet \
    --contracts-dir $CONTRACTS_DIR
```

### Manual Testing

```bash
# Build and test mero-sign CLI
cargo build -p mero-sign
cargo run -p mero-sign -- --help

# Generate a new key
cargo run -p mero-sign -- generate-key --output /tmp/test-key.json

# Derive signerId
cargo run -p mero-sign -- derive-signer-id --key scripts/test-signing-key/test-key.json

# Sign a test manifest
echo '{"version":"1.0","package":"test","appVersion":"1.0.0","wasm":{"path":"app.wasm","size":1024,"hash":null}}' > /tmp/manifest.json
cargo run -p mero-sign -- sign /tmp/manifest.json --key scripts/test-signing-key/test-key.json
cat /tmp/manifest.json  # Verify signature added
```

## Documentation Update

- **Internal docs**: Update bundle creation guides to require signatures
- **Public docs**: Document `signerId` requirement and `did:key` format for bundle manifests
- **Developer guide**: Document `mero-sign` CLI usage for bundle signing
- **CI/CD guide**: Document test signing key usage and when to generate production keys

---

> [!NOTE]
> **High Risk**
> High risk because it changes bundle installation identity and persistence (`ApplicationMeta` schema + `ApplicationId` derivation) and makes bundle installs contingent on new Ed25519/JCS signature verification and runtime/version/path validation, which can break existing bundles/stores if mismatched.
> 
> **Overview**
> **Enforces mandatory signing for `.mpk` bundles** by adding RFC 8785 (JCS) canonicalization + SHA-256 payload hashing and Ed25519 signature verification, including `did:key`-derived `signerId` matching, across all bundle install paths.
> 
> **Changes bundle identity and storage**: bundle `ApplicationId` is now derived from `(package, signer_id)` (enabling version upgrades under the same signer), `ApplicationMeta` gains a persisted `signer_id`, and non-bundle installs explicitly pass `None` (stored as an empty sentinel).
> 
> **Hardens bundle processing** with manifest `minRuntimeVersion` checks against current runtime and path-traversal validation for `package`, `appVersion`, and artifact paths; adds extensive unit/integration tests for signature failures, signer mismatch, and path traversal.
> 
> **Adds signing toolchain**: new `tools/mero-sign` CLI (sign/generate-key/derive-signer-id), a repo-committed *test-only* signing key under `scripts/test-signing-key/`, and app `build-bundle.sh` scripts updated to sign manifests; kv-store workflow switches to `ghcr.io/calimero-network/merod:edge`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 261ead441a346babb9b93ed787415afc11b424fb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


crates/node/primitives/src/client/application.rs
```diff
@@ -239,8 +239,13 @@ impl NodeClient {
         metadata: Vec<u8>,
         package: &str,
         version: &str,
+        signer_id: Option<&str>,
         is_bundle: bool,
     ) -> eyre::Result<ApplicationId> {
+        // For bundles: signer_id is required
+        // For non-bundles: signer_id is optional (backwards compatibility)
+        let signer_id_str = signer_id.unwrap_or("");
```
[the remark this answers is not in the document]
xilosada: We should also deprecate installation without bundles. Is it plan for a following PR?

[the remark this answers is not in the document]
rtb-12: I was not sure whether we are deprecating installation without bundles, so I kept it this way for backward compatibility. If we are deprecating it, it can be handled in a follow-up PR.

xilosada (commented):
## 🤖 AI Code Reviewer

**Reviewed by 3 agents** | Quality score: 85% | Review time: 287.5s

### 🟡 Issues pending resolution

✅ **7 Fixed** | 🆕 **15 New**

---

### ✅ Fixed Issues

<details>
<summary>The following issues from previous reviews have been addressed:</summary>

1. ~~Medium Severity~~ (`crates/node/primitives/src/client/application.rs:247`)
2. ~~Unknown Issue~~ (`crates/node/primitives/src/client/application.rs:247`)
3. ~~Low Severity~~ (`crates/primitives/src/application.rs:311`)
4. ~~High Severity~~ (`crates/node/primitives/src/client/application.rs:598`)
5. ~~Medium Severity~~ (`tools/mero-sign/src/main.rs:173`)
6. ~~High Severity~~ (`tools/mero-sign/src/main.rs:105`)
7. ~~Low Severity~~ (`apps/kv-store/workflows/bundle-invitation-test.yml:10`)

</details>

### 🆕 New Issues

*These issues were found in the latest changes:*

### 🟡 Warning (5)

#### 1. Private signing key committed to repository
**File:** `scripts/test-signing-key/test-key.json` (line 1-1) | **Consensus:** 1/3 agents

A private Ed25519 signing key is committed to the repository. While documented as a test key, this creates risk of accidental production use. Any bundles signed with this key can be forged by anyone with repository access. The build scripts (`apps/kv-store/build-bundle.sh`, `apps/access-control/build-bundle.sh`) automatically use this key, which could lead to test-signed bundles being deployed to production environments.

**Suggested fix:**
```
Consider generating test keys dynamically during CI/test runs rather than committing them. Add CI checks to detect and reject bundles signed with the test key in production deployments. Add the key path to a pre-commit hook that warns developers.
```

> *Found by: security-agent*

#### 2. Unnecessary full clone of bundle data for spawn_blocking
**File:** `crates/node/primitives/src/client/application.rs` (line 396-402) | **Consensus:** 1/3 agents

The bundle_data is cloned multiple times throughout the codebase (in install_bundle_from_url, install_bundle_from_file, install_bundle_from_blob, etc.) before passing to spawn_blocking. For large bundles (potentially several MB), this creates a full memory copy each time. The same pattern appears at lines 396, 586, 941, and others.

**Suggested fix:**
```
Use `Arc<Vec<u8>>` or `Arc<[u8]>` for bundle_data to share ownership without copying. Change `bundle_data.clone()` to `Arc::clone(&bundle_data)` and update the type signatures accordingly. This avoids O(n) memory allocation where n is bundle size.
```

> *Found by: performance-agent*

#### 3. minRuntimeVersion field is not validated against actual runtime
**File:** `crates/node/primitives/src/client/application.rs` (line 843-863) | **Consensus:** 1/3 agents

The manifest field `minRuntimeVersion` is parsed and stored but never validated against the actual runtime version during bundle installation. This allows bundles requiring a newer runtime to be installed on older runtimes, potentially causing undefined behavior or security issues if newer security features are expected but unavailable.

**Suggested fix:**
```
Add validation that compares `manifest.min_runtime_version` against the node's actual runtime version before allowing installation. Reject bundles where the minimum required version exceeds the current runtime version.
```

> *Found by: security-agent*

#### 4. Empty string used as sentinel value for missing signer_id
**File:** `crates/node/primitives/src/client/application.rs` (line 242-244) | **Consensus:** 1/3 agents

When `signer_id` is `None`, the code uses an empty string as a sentinel value that gets stored in `ApplicationMeta`. This conflates 'no signer' with 'signer with empty id', which could lead to subtle bugs if the empty string is ever treated as a valid signer_id. The database schema should distinguish between 'no signer' (non-bundle) and 'has signer' (bundle) cases.

**Suggested fix:**
```
Consider making `signer_id` in `ApplicationMeta` an `Option<Box<str>>` to explicitly represent the absence of a signer, or add a comment documenting that empty string is the canonical representation for non-bundle applications.
```

> *Found by: quality-agent*

#### 5. BundleManifest signer_id optionality may cause confusion
**File:** `crates/node/primitives/src/bundle/mod.rs` (line 73-80) | **Consensus:** 1/3 agents

The `signer_id` field is `Option<String>` in the struct, but `verify_manifest_signature` requires it to be present for signed bundles. The documentation doesn't clarify when `signer_id` can be `None` vs when it must be `Some`. This could confuse developers about whether unsigned manifests are valid at the type level.

**Suggested fix:**
```
Add a doc comment explaining that `signer_id` is required when `signature` is present, and both are required for bundle installation. Consider whether the types should enforce this invariant (e.g., a separate `SignedBundleManifest` type).
```

> *Found by: quality-agent*

### 💡 Suggestion (7)

#### 1. Path validation functions lack dedicated unit tests
**File:** `crates/node/primitives/src/client/application.rs` (line 741-799) | **Consensus:** 1/3 agents

The `validate_path_component` and `validate_artifact_path` functions implement critical security checks for path traversal attacks, but they are private functions without dedicated unit tests. While the integration tests may exercise some paths, edge cases like Unicode path separators, URL-encoded sequences, or Windows UNC paths are not explicitly tested.

**Suggested fix:**
```
Add unit tests for `validate_path_component` and `validate_artifact_path` covering edge cases: Unicode path separators (e.g., '／'), URL-encoded sequences (e.g., '%2e%2e'), empty strings, very long paths, and Windows-specific patterns.
```

> *Found by: quality-agent*

#### 2. Double parsing of manifest JSON
**File:** `crates/node/primitives/src/client/application.rs` (line 827-831) | **Consensus:** 1/3 agents

The manifest is parsed twice: first as serde_json::Value (for signature verification), then from_value() clones and re-parses into BundleManifest. The clone at line 831 `serde_json::from_value(manifest_json.clone())` allocates again.

**Suggested fix:**
```
Consider deserializing directly into a struct that contains both the typed fields and preserves the raw JSON for signature verification, or use serde's RawValue to avoid the double parse. Alternatively, verify the signature first on the raw string, then parse once into the typed struct.
```

> *Found by: performance-agent*

#### 3. Full JSON value clone in canonicalize_manifest
**File:** `crates/node/primitives/src/bundle/signature.rs` (line 56-70) | **Consensus:** 1/3 agents

The canonicalize_manifest function clones the entire manifest JSON value just to remove the signature field. For manifests with large embedded data or many fields, this creates unnecessary allocations. This function is called during signature verification which happens on every bundle installation.

**Suggested fix:**
```
Consider using a custom serializer that skips the signature field during serialization instead of cloning and mutating. Alternatively, if the manifest structure is known, serialize fields individually. However, given the RFC 8785 canonicalization requirement, this may be acceptable complexity trade-off.
```

> *Found by: performance-agent*

#### 4. verify_manifest_signature couples JSON parsing with verification
**File:** `crates/node/primitives/src/bundle/signature.rs` (line 193-200) | **Consensus:** 1/3 agents

The `verify_manifest_signature` function takes a `serde_json::Value` and performs both field extraction and cryptographic verification. This makes it harder to test individual verification steps in isolation and requires callers to parse JSON before calling. Consider separating the JSON field extraction from the pure cryptographic verification logic.

**Suggested fix:**
```
Consider adding a lower-level `verify_signature(canonical_bytes: &[u8], public_key: &[u8; 32], signature: &[u8; 64], expected_signer_id: &str) -> Result<ManifestVerification>` function that the JSON-aware function delegates to.
```

> *Found by: quality-agent*

#### 5. Underscore-prefix field stripping may have unintended side effects
**File:** `crates/node/primitives/src/bundle/signature.rs` (line 76-81) | **Consensus:** 1/3 agents

The canonicalization strips all underscore-prefixed fields (`obj.retain(|k, _| !k.starts_with('_'))`). While this prevents signature confusion attacks, it could also strip legitimate fields from future manifest versions that start with underscore. Additionally, the comment mentions `_binary` and `_overwrite` but a comprehensive list of expected transient fields is not documented.

**Suggested fix:**
```
Consider using an explicit allowlist of known transient fields to remove rather than a pattern-based approach, or document the underscore-prefix convention as reserved for transient/unsigned fields in the manifest specification.
```

> *Found by: security-agent*

#### 6. Empty signer_id for non-bundle installations may cause identity confusion
**File:** `crates/node/primitives/src/client/application.rs` (line 239-244) | **Consensus:** 1/3 agents

Non-bundle installations use an empty string for `signer_id` (`signer_id.unwrap_or("")`). This means non-bundle applications share a common 'anonymous' signer identity namespace. If upgrade paths from non-bundle to bundle installations are ever supported, this could create identity/authority confusion where applications installed without signature verification could be associated with signed bundles.

**Suggested fix:**
```
Consider using a distinct sentinel value (e.g., a specific 'unsigned' marker) rather than empty string, and document that non-bundle installations cannot be upgraded to signed bundle installations without re-installation.
```

> *Found by: security-agent*

#### 7. Signature algorithm check could be case-sensitive bypass target
**File:** `crates/node/primitives/src/bundle/signature.rs` (line 200-210) | **Consensus:** 1/3 agents

The algorithm validation uses exact string comparison (`algorithm == "ed25519"`). While this is currently correct, if other code paths normalize or case-fold the algorithm string before reaching this check, variations like 'ED25519' or 'Ed25519' could potentially bypass validation. The check is correct as implemented, but the specification should explicitly state case-sensitivity requirements.

**Suggested fix:**
```
Add a comment documenting that algorithm comparison is intentionally case-sensitive per specification, or use case-insensitive comparison with `.eq_ignore_ascii_case()` if the spec allows case variation.
```

> *Found by: security-agent*

### 📝 Nitpick (3)

#### 1. Using chars().nth() instead of byte indexing for ASCII check
**File:** `crates/node/primitives/src/client/application.rs` (line 757-766) | **Consensus:** 1/3 agents

In validate_path_component, `value.chars().nth(1)` is O(n) for UTF-8 strings in general, though practically fast for short strings. Since we're checking for Windows drive letters (ASCII ':'), byte indexing would be O(1).

**Suggested fix:**
```
Replace `value.chars().nth(1) == Some(':')` with `value.as_bytes().get(1) == Some(&b':')` for guaranteed O(1) access, consistent with the similar check in validate_artifact_path at line 787.
```

> *Found by: performance-agent*

#### 2. Test helper sign_manifest duplicates production signing logic
**File:** `crates/node/primitives/tests/bundle_installation.rs` (line 47-56) | **Consensus:** 1/3 agents

The `sign_manifest` helper in the test file duplicates the signing logic that presumably exists in `mero-sign`. If the signing algorithm changes, tests and the CLI tool could diverge. Consider exposing a shared signing function from a common location that both the tests and the CLI can use.

**Suggested fix:**
```
Extract signing logic to a shared library function (perhaps in `calimero-node-primitives::bundle::signature`) that both `mero-sign` and tests can import, ensuring consistency.
```

> *Found by: quality-agent*

#### 3. Test code sets dummy signature in struct after actual signing
**File:** `crates/node/primitives/tests/bundle_installation.rs` (line 103-109) | **Consensus:** 1/3 agents

After signing the manifest JSON, the test updates manifest.signature with a dummy signature (signing [0u8; 32] instead of the actual payload). While this doesn't affect test correctness since the JSON is what gets serialized, it creates an inconsistency between the struct and the serialized form that could confuse future maintainers.

**Suggested fix:**
```
Either remove the struct signature assignment entirely (it's not used after this point since manifest_json is serialized), or compute and store the actual signature values in the struct for consistency.
```

> *Found by: performance-agent*


---

<sub>🤖 Generated by [AI Code Reviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-29ad7bf6`</sub>

crates/node/primitives/src/client/application.rs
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/primitives/src/application.rs
```diff
@@ -75,6 +84,283 @@ impl FromStr for ApplicationId {
...
+        let separator_pos = s
+            .find(':')
+            .ok_or_else(|| InvalidAppKey::InvalidFormat(s.to_owned()))?;
+
+        let app_id = &s[..separator_pos];
+        let signer_id_str = &s[separator_pos + 1..];
+
+        let signer_id = SignerId::new(signer_id_str)?;
+        Self::new(app_id, signer_id)
+    }
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
```diff
@@ -573,10 +594,19 @@ impl NodeClient {
...
 
+        // Verify manifest signature and derive signerId
+        let verification = verify_manifest_signature(&manifest_json)?;
+        let signer_id = verification.signer_id;
+        debug!(
+            signer_id = %signer_id,
+            bundle_hash = %hex::encode(verification.bundle_hash),
+            "bundle manifest signature verified"
+        );
+
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

tools/mero-sign/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+            serde_json::Value::String(signer_id.clone()),
+        );
+        // Also add minRuntimeVersion if not present
+        if !obj.contains_key("minRuntimeVersion") {
+            obj.insert(
+                "minRuntimeVersion".to_string(),
+                serde_json::Value::String("1.0.0".to_string()),
+            );
+        }
+    }
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

tools/mero-sign/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+/// Removes the signature field before canonicalization.
+fn canonicalize_manifest(manifest_json: &serde_json::Value) -> Result<Vec<u8>> {
+    // Clone and remove the signature field for canonicalization
+    let mut signing_view = manifest_json.clone();
+    if let Some(obj) = signing_view.as_object_mut() {
+        obj.remove("signature");
+        // Also remove transient fields
+        obj.remove("_binary");
+        obj.remove("_overwrite");
+    }
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

apps/kv-store/workflows/bundle-invitation-test.yml
```diff
@@ -7,7 +7,7 @@ force_pull_image: false
 nodes:
   chain_id: testnet-1
   count: 2
-  image: merod:latest
+  image: ghcr.io/calimero-network/merod:edge
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

scripts/test-signing-key/test-key.json
```diff
@@ -0,0 +1,5 @@
+{
```
xilosada, line 1: 🟡 **Private signing key committed to repository**

A private Ed25519 signing key is committed to the repository. While documented as a test key, this creates risk of accidental production use. Any bundles signed with this key can be forged by anyone with repository access. The build scripts (`apps/kv-store/build-bundle.sh`, `apps/access-control/build-bundle.sh`) automatically use this key, which could lead to test-signed bundles being deployed to production environments.

**Suggested fix:**
```
Consider generating test keys dynamically during CI/test runs rather than committing them. Add CI checks to detect and reject bundles signed with the test key in production deployments. Add the key path to a pre-commit hook that warns developers.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -388,14 +396,16 @@ impl NodeClient {
                 "bundle downloaded and stored as blob"
```
xilosada, line 402: 🟡 **Unnecessary full clone of bundle data for spawn_blocking**

The bundle_data is cloned multiple times throughout the codebase (in install_bundle_from_url, install_bundle_from_file, install_bundle_from_blob, etc.) before passing to spawn_blocking. For large bundles (potentially several MB), this creates a full memory copy each time. The same pattern appears at lines 396, 586, 941, and others.

**Suggested fix:**
```
Use `Arc<Vec<u8>>` or `Arc<[u8]>` for bundle_data to share ownership without copying. Change `bundle_data.clone()` to `Arc::clone(&bundle_data)` and update the type signatures accordingly. This avoids O(n) memory allocation where n is bundle size.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -746,7 +843,28 @@ impl NodeClient {
                     bail!("bundle manifest 'appVersion' field is empty");
```
xilosada, line 853: 🟡 **minRuntimeVersion field is not validated against actual runtime**

The manifest field `minRuntimeVersion` is parsed and stored but never validated against the actual runtime version during bundle installation. This allows bundles requiring a newer runtime to be installed on older runtimes, potentially causing undefined behavior or security issues if newer security features are expected but unavailable.

**Suggested fix:**
```
Add validation that compares `manifest.min_runtime_version` against the node's actual runtime version before allowing installation. Reject bundles where the minimum required version exceeds the current runtime version.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -239,8 +239,13 @@ impl NodeClient {
         metadata: Vec<u8>,
         package: &str,
         version: &str,
+        signer_id: Option<&str>,
```
xilosada, line 244: 🟡 **Empty string used as sentinel value for missing signer_id**

When `signer_id` is `None`, the code uses an empty string as a sentinel value that gets stored in `ApplicationMeta`. This conflates 'no signer' with 'signer with empty id', which could lead to subtle bugs if the empty string is ever treated as a valid signer_id. The database schema should distinguish between 'no signer' (non-bundle) and 'has signer' (bundle) cases.

**Suggested fix:**
```
Consider making `signer_id` in `ApplicationMeta` an `Option<Box<str>>` to explicitly represent the absence of a signer, or add a comment documenting that empty string is the canonical representation for non-bundle applications.
```

crates/node/primitives/src/bundle/mod.rs
```diff
@@ -58,6 +73,13 @@ pub struct BundleManifest {
     pub package: String,
```
xilosada, line 73: 🟡 **BundleManifest signer_id optionality may cause confusion**

The `signer_id` field is `Option<String>` in the struct, but `verify_manifest_signature` requires it to be present for signed bundles. The documentation doesn't clarify when `signer_id` can be `None` vs when it must be `Some`. This could confuse developers about whether unsigned manifests are valid at the type level.

**Suggested fix:**
```
Add a doc comment explaining that `signer_id` is required when `signature` is present, and both are required for bundle installation. Consider whether the types should enforce this invariant (e.g., a separate `SignedBundleManifest` type).
```

crates/node/primitives/src/client/application.rs
```diff
@@ -723,8 +741,81 @@ impl NodeClient {
         Ok(application_id)
```
xilosada, line 747: 💡 **Path validation functions lack dedicated unit tests**

The `validate_path_component` and `validate_artifact_path` functions implement critical security checks for path traversal attacks, but they are private functions without dedicated unit tests. While the integration tests may exercise some paths, edge cases like Unicode path separators, URL-encoded sequences, or Windows UNC paths are not explicitly tested.

**Suggested fix:**
```
Add unit tests for `validate_path_component` and `validate_artifact_path` covering edge cases: Unicode path separators (e.g., '／'), URL-encoded sequences (e.g., '%2e%2e'), empty strings, very long paths, and Windows-specific patterns.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -733,9 +824,15 @@ impl NodeClient {
             let path = entry.path()?;
 
             if path.file_name().and_then(|n| n.to_str()) == Some("manifest.json") {
-                let mut manifest_json = String::new();
-                entry.read_to_string(&mut manifest_json)?;
-                let manifest: BundleManifest = serde_json::from_str(&manifest_json)
+                let mut manifest_str = String::new();
```
xilosada, line 833: 💡 **Double parsing of manifest JSON**

The manifest is parsed twice: first as serde_json::Value (for signature verification), then from_value() clones and re-parses into BundleManifest. The clone at line 831 `serde_json::from_value(manifest_json.clone())` allocates again.

**Suggested fix:**
```
Consider deserializing directly into a struct that contains both the typed fields and preserves the raw JSON for signature verification, or use serde's RawValue to avoid the double parse. Alternatively, verify the signature first on the raw string, then parse once into the typed struct.
```

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,658 @@
...
+    let encoded = bs58::encode(&multicodec_key).into_string();
+
+    format!("did:key:z{}", encoded)
+}
+
+/// Canonicalizes a manifest JSON value using RFC 8785 (JCS).
+///
+/// The `signature` field MUST be excluded before canonicalization.
+/// This function clones the input and removes the signature field.
+///
```
xilosada, line 56: 💡 **Full JSON value clone in canonicalize_manifest**

The canonicalize_manifest function clones the entire manifest JSON value just to remove the signature field. For manifests with large embedded data or many fields, this creates unnecessary allocations. This function is called during signature verification which happens on every bundle installation.

**Suggested fix:**
```
Consider using a custom serializer that skips the signature field during serialization instead of cloning and mutating. Alternatively, if the manifest structure is known, serialize fields individually. However, given the RFC 8785 canonicalization requirement, this may be acceptable complexity trade-off.
```

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,658 @@
...
+/// This function performs the complete verification flow:
+/// 1. Extracts and validates the signature object
+/// 2. Decodes the base64url public key and signature
+/// 3. Canonicalizes the manifest (excluding signature field) using RFC 8785
+/// 4. Computes the signing payload (SHA-256 of canonical bytes)
+/// 5. Verifies the Ed25519 signature
+/// 6. Derives the signerId (did:key) from the public key
+/// 7. Validates that the derived signerId matches the manifest's signerId
+///
+/// # Arguments
```
xilosada, line 198: 💡 **verify_manifest_signature couples JSON parsing with verification**

The `verify_manifest_signature` function takes a `serde_json::Value` and performs both field extraction and cryptographic verification. This makes it harder to test individual verification steps in isolation and requires callers to parse JSON before calling. Consider separating the JSON field extraction from the pure cryptographic verification logic.

**Suggested fix:**
```
Consider adding a lower-level `verify_signature(canonical_bytes: &[u8], public_key: &[u8; 32], signature: &[u8; 64], expected_signer_id: &str) -> Result<ManifestVerification>` function that the JSON-aware function delegates to.
```

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,658 @@
...
+        // Remove all underscore-prefixed fields to prevent signature confusion
+        // Unknown underscore-prefixed fields could be canonicalized and signed but ignored
+        // during processing, potentially leading to signature confusion or replay attacks
+        obj.retain(|k, _| !k.starts_with('_'));
+    }
+
+    // Canonicalize using RFC 8785 JCS
+    let canonical_bytes = serde_json_canonicalizer::to_vec(&signing_view)
+        .context("failed to canonicalize manifest JSON")?;
+
```
xilosada, line 81: 💡 **Underscore-prefix field stripping may have unintended side effects**

The canonicalization strips all underscore-prefixed fields (`obj.retain(|k, _| !k.starts_with('_'))`). While this prevents signature confusion attacks, it could also strip legitimate fields from future manifest versions that start with underscore. Additionally, the comment mentions `_binary` and `_overwrite` but a comprehensive list of expected transient fields is not documented.

**Suggested fix:**
```
Consider using an explicit allowlist of known transient fields to remove rather than a pattern-based approach, or document the underscore-prefix convention as reserved for transient/unsigned fields in the manifest specification.
```

crates/node/primitives/src/client/application.rs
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

scripts/test-signing-key/test-key.json
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/mod.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/signature.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/signature.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/signature.rs
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/tests/bundle_installation.rs
```diff
@@ -1396,6 +1434,479 @@ async fn test_bundle_blob_sharing_integration() {
...
+        "minRuntimeVersion": "1.0.0",
+        "wasm": {
+            "path": "app.wasm",
+            "size": wasm_content.len()
+        },
+        "migrations": []
+    });
+
+    // But actually sign with key A (mismatch!)
+    sign_manifest(&mut manifest, &signing_key_a);
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,726 @@
...
+    if let Some(obj) = manifest_json.as_object_mut() {
+        obj.insert(
+            "signature".to_string(),
+            serde_json::json!({
+                "algorithm": "ed25519",
+                "publicKey": public_key_b64,
+                "signature": signature_b64
+            }),
+        );
+    }
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,724 @@
...
+    use ed25519_dalek::Signer;
+
+    let verifying_key = signing_key.verifying_key();
+    let signer_id = derive_signer_id_did_key(verifying_key.as_bytes());
+
+    // Add signerId to the manifest only if not already present
+    if let Some(obj) = manifest_json.as_object_mut() {
+        obj.entry("signerId".to_string())
+            .or_insert_with(|| serde_json::Value::String(signer_id.clone()));
+    }
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

apps/kv-store/build-bundle.sh
```diff
@@ -30,6 +30,7 @@ cat > res/bundle-temp/manifest.json <<EOF
   "version": "1.0",
   "package": "com.calimero.kv-store",
   "appVersion": "1.0.0",
+  "minRuntimeVersion": "1.0.0",
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
```diff
@@ -1563,3 +1716,166 @@ impl NodeClient {
...
+    fn test_validate_artifact_path_very_long() {
+        // Test with a very long path (potential DoS)
+        let long_path = "a".repeat(10000);
+        // Current implementation doesn't check length, but this documents the consideration
+        let result = NodeClient::validate_artifact_path(&long_path, "test");
+        // Should either pass (if length is not checked) or fail (if length limit is added)
+        assert!(
+            result.is_ok() || result.is_err(),
+            "Very long path should be handled"
+        );
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/client/application.rs
```diff
@@ -80,8 +81,8 @@ impl NodeClient {
         if is_bundle {
             // This is a bundle, extract WASM from extracted directory or bundle
             // Extract manifest (blocking I/O wrapped in spawn_blocking)
-            let blob_bytes_clone = blob_bytes.clone();
-            let manifest = tokio::task::spawn_blocking(move || {
+            let blob_bytes_clone = Arc::clone(&blob_bytes);
+            let (_, manifest) = tokio::task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/node/primitives/src/bundle/mod.rs
```diff
@@ -58,6 +73,20 @@ pub struct BundleManifest {
...
```
xilosada, line 73: 🟡 **Type-documentation mismatch for signer_id field**

The `signer_id` field is documented as 'required for all bundle manifests' but is typed as `Option<String>`. The doc comment explicitly states 'Both `signer_id` and `signature` must be present for bundle installation' yet the type allows `None`. This creates confusion about the contract - callers cannot know from the type alone that this is effectively required.

**Suggested fix:**
```
Either make the field non-optional (`pub signer_id: String`) with a separate pre-verification struct for parsing, or update the documentation to clearly state it's 'required at verification time' rather than 'required for all manifests'. Consider adding a `verify()` method that returns a validated manifest type.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -239,8 +241,17 @@ impl NodeClient {
         metadata: Vec<u8>,
         package: &str,
         version: &str,
```
xilosada, line 243: 🟡 **Sentinel value pattern for signer_id creates confusing API**

The code uses an empty string as a sentinel value for non-bundle applications (`signer_id.unwrap_or("")`). While documented in comments, this pattern creates ambiguity between 'no signer' and 'empty signer' states, and the explanation that 'Non-bundle installations cannot be upgraded to signed bundle installations' hints at a deeper design issue. The store schema should use `Option<Box<str>>` to properly represent the absence of a signer_id.

**Suggested fix:**
```
Consider using `Option<Box<str>>` in `ApplicationMeta` instead of `Box<str>` for `signer_id`. Update the schema with proper migration handling, or if that's infeasible now, add a dedicated constant like `const NO_SIGNER: &str = ""` with explicit documentation about the sentinel value semantics.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -723,8 +747,81 @@ impl NodeClient {
...
+    /// `applications` directory by using path traversal sequences in package or version fields.
+    fn validate_path_component(value: &str, field_name: &str) -> eyre::Result<()> {
+        // Check for parent directory traversal
+        if value.contains("..") {
+            bail!("{} contains path traversal sequence '..'", field_name);
+        }
+        // Check for directory separators (Unix and Windows)
+        if value.contains('/') || value.contains('\\') {
+            bail!("{} contains directory separator", field_name);
+        }
```
xilosada, line 763: 🟡 **Path validation may miss edge cases with encoded or normalized paths**

The `validate_path_component` and `validate_artifact_path` functions check for common path traversal patterns, but may not handle all edge cases such as URL-encoded sequences (%2e%2e), Unicode normalization attacks (e.g., using Unicode characters that normalize to '.'), or OS-specific path handling quirks. While the current implementation covers the most common attack vectors, a defense-in-depth approach would be beneficial.

**Suggested fix:**
```
Consider normalizing paths before validation, rejecting any non-ASCII characters in paths, or using a well-tested path sanitization library. Also verify the final resolved path after extraction stays within the expected directory.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -723,8 +747,81 @@ impl NodeClient {
...
-    fn extract_bundle_manifest(bundle_data: &[u8]) -> eyre::Result<BundleManifest> {
+    /// Validates that a string is safe for use as a filesystem path component.
+    /// Returns an error if the string contains path traversal or other unsafe characters.
+    ///
+    /// This prevents malicious bundle manifests from writing files outside the intended
+    /// `applications` directory by using path traversal sequences in package or version fields.
+    fn validate_path_component(value: &str, field_name: &str) -> eyre::Result<()> {
+        // Check for parent directory traversal
+        if value.contains("..") {
+            bail!("{} contains path traversal sequence '..'", field_name);
```
xilosada, line 758: 🟡 **Path component validation allows edge cases that artifact validation rejects**

The `validate_path_component` function uses `contains("..")` which would reject strings like `foo..bar` (potentially valid package name with consecutive dots), while `validate_artifact_path` correctly checks for `..` as a path *component* using `split('/').any(|c| c == "..")`. This inconsistency could cause valid package names with consecutive dots to be incorrectly rejected.

**Suggested fix:**
```
Consider whether package names like `com.example..test` should be allowed. If not, the current behavior is fine but should be documented. If they should be allowed, change to check for `.` + `/` combinations or use a more precise regex.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -416,7 +432,7 @@ impl NodeClient {
                 .join("extracted");
 
             // Wrap blocking I/O in spawn_blocking to avoid blocking async runtime
-            let bundle_data_clone = bundle_data.clone();
+            let bundle_data_clone = Arc::clone(&bundle_data);
             let manifest_clone = manifest.clone();
```
xilosada, line 436: 💡 **Duplicated bundle metadata extraction logic across three methods**

The code for extracting `bundle_metadata` from the manifest (lines building `metadata_obj` with package, version, name, description, icon, tags, license, and links) is duplicated in `install_application_from_url`, `install_bundle_from_path`, and `install_application_from_bundle_blob`. This violates DRY and increases maintenance burden when the metadata schema changes.

**Suggested fix:**
```
Extract the metadata building logic into a helper method like `fn extract_bundle_metadata(manifest: &BundleManifest) -> eyre::Result<Vec<u8>>` and call it from all three locations.
```

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,727 @@
...
+    if let Some(obj) = signing_view.as_object_mut() {
+        obj.remove("signature");
+        // Remove all underscore-prefixed fields to prevent signature confusion.
+        // The underscore-prefix convention is reserved for transient/unsigned fields that
+        // should not be included in signature verification. Known transient fields include:
+        // - _binary: Binary data references (not part of canonical JSON)
+        // - _overwrite: Overwrite flags for migration artifacts
+        // Any future fields starting with underscore are also stripped to prevent signature
+        // confusion attacks where fields could be canonicalized and signed but ignored during
+        // processing, potentially leading to replay attacks.
```
xilosada, line 74: 💡 **Underscore-prefix stripping rule should be documented in manifest spec**

The canonicalization removes all underscore-prefixed fields to prevent signature confusion attacks. This is a good security measure, but it creates an implicit contract that any field starting with underscore is transient and unsigned. If this convention is not documented in the bundle manifest specification, developers may unknowingly add underscore-prefixed fields expecting them to be signed, leading to unexpected behavior.

**Suggested fix:**
```
Document this convention clearly in the bundle manifest specification and consider adding a warning log when underscore-prefixed fields are stripped during verification.
```

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,727 @@
...
+    let signing_payload = compute_signing_payload(&canonical_bytes);
+
+    // Verify the Ed25519 signature
+    verify_ed25519(&signature_bytes, &public_key_bytes, &signing_payload)?;
+
+    // Derive the signerId from the public key
+    let derived_signer_id = derive_signer_id_did_key(&public_key_bytes);
+
+    // Validate that the manifest's signerId matches the derived signerId
+    // signerId is required for signed bundles to ensure identity verification
```
xilosada, line 253: 💡 **sign_manifest_json silently overwrites existing signerId without warning**

The `sign_manifest_json` function always overwrites any existing `signerId` field with the one derived from the signing key. While the comment explains this ensures consistency, it could surprise callers who pass a manifest with a pre-existing `signerId` expecting it to be preserved or validated. The silent overwrite behavior should either be documented in the function's doc comment or return information about whether an overwrite occurred.

**Suggested fix:**
```
Add to the function's doc comment: '# Note\nThis function will overwrite any existing `signerId` field to ensure consistency with the signing key.' Alternatively, add a `warn!` log when overwriting a different signerId, or return a struct that includes `overwritten_signer_id: Option<String>`.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -733,9 +830,19 @@ impl NodeClient {
             let path = entry.path()?;
```
xilosada, line 830: 💡 **Double JSON clone during manifest extraction and verification**

The manifest JSON value is cloned twice: once in `extract_bundle_manifest` (line ~841 `from_value(manifest_json.clone())`) and once in `canonicalize_manifest` (called via `verify_manifest_signature`). For typical small manifests this is negligible, but for manifests with many resources/migrations this adds unnecessary allocation overhead.

**Suggested fix:**
```
Consider having `verify_manifest_signature` take ownership of the JSON value since it's only needed for verification, or restructure to parse the BundleManifest from the canonical bytes after verification. Alternatively, use `serde_json::from_value` without clone by verifying the signature first on the raw JSON string before typed parsing.
```

crates/node/primitives/src/bundle/signature.rs
```diff
@@ -0,0 +1,727 @@
...
+        "unsupported signature algorithm: '{}', expected 'ed25519'",
+        algorithm
+    );
+
+    let public_key_b64 = signature_obj
+        .get("publicKey")
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| eyre::eyre!("signature missing 'publicKey' field"))?;
+
+    let signature_b64 = signature_obj
```
xilosada, line 231: 💡 **Error message leaks derived signerId in mismatch case**

When signerId verification fails, the error message includes both the declared and derived signerIds. While signerIds are derived from public keys and thus not strictly secret, this verbose error message could assist an attacker in understanding the expected signerId format or in debugging their tampering attempts.

**Suggested fix:**
```
Consider a less verbose error message like 'signerId does not match signature public key' without revealing the derived value.
```

crates/node/primitives/src/client/application.rs
```diff
@@ -723,8 +747,81 @@ impl NodeClient {
...
+        Ok((verification, manifest))
+    }
+
+    /// Extract and parse bundle manifest from bundle archive data.
+    /// Returns both the raw JSON value (for signature verification) and the typed manifest.
+    fn extract_bundle_manifest(
+        bundle_data: &[u8],
+    ) -> eyre::Result<(serde_json::Value, BundleManifest)> {
         let tar = GzDecoder::new(bundle_data);
         let mut archive = Archive::new(tar);
```
xilosada, line 827: 💡 **Manifest validation performed before signature verification**

In `extract_bundle_manifest`, extensive validation (field checks, path component validation, runtime version parsing) is performed before signature verification happens in the caller. If a bundle has an invalid signature, all this validation work is wasted. For rejected bundles, this adds unnecessary CPU overhead.

**Suggested fix:**
```
Consider restructuring to verify the signature first on the raw JSON before performing detailed validation. This would reject tampered bundles faster without doing unnecessary work.
```

xilosada (approved):

xilosada (approved):

PR #1828, opened 2026-02-03
fix(merod+meroctl): merod + meroctl command tests + help outputs

frdomovic:
# fix(merod+meroctl): merod + meroctl command tests + help outputs

## Description
Fixing some outdated --help instructions. In others adding "--node node_id" flag because the default is not automatically set.
Application helper displaying wrong flag examples ( application -> app) 
Also fixed some outputs that were not properly done -> e.g. view package for installed app would return unknown. fetch latest app version would return only application id instead of app id + version
Screenshot 2026-02-03 at 14 08 58
Screenshot 2026-02-03 at 14 25 50

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the `get-latest-version` API shape and related client/server plumbing, plus adjusts application installation metadata for non-bundle installs; could break consumers expecting the old response or relying on default package/version values.
> 
> **Overview**
> Updates `meroctl` help/examples to match current CLI usage (use `app` instead of `application`, add explicit `--node` in examples, and refresh `context create`/watch examples including `--protocol` and `--current-app-id`).
> 
> Fixes latest-version and install metadata reporting: the node client `get_latest_version` now returns both the version string and `ApplicationId`, the server `GetLatestVersionResponse` includes `version`, and `meroctl` renders a two-column version/app-id table. Dev installs now pass through optional `package`/`version` to `install_application_from_path` (falling back to `unknown`/`0.0.0`), and bundle installation tests are updated for the new function signatures/return types.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cdcc805451ca611fd1c673e8d50ba53446d9ad79. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/meroctl/src/cli/validation.rs
+++ b/crates/meroctl/src/cli/validation.rs
@@ -8,21 +8,6 @@ use std::path::Path;
 use eyre::bail;
 use eyre::Result;
 
-/// Validates that a string is non-empty.
-///
-/// # Arguments
-/// * `value` - The string to validate
-/// * `field_name` - Name of the field for error messages
-///
-/// # Errors
-/// Returns an error if the string is empty or contains only whitespace.
-pub fn validate_non_empty(value: &str, field_name: &str) -> Result<()> {
-    if value.trim().is_empty() {
-        bail!("{} cannot be empty", field_name);
-    }
-    Ok(())
-}
-
 /// Validates that a file exists and is readable.
 ///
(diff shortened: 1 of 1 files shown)


chefsale (approved):
🚢

PR #1840, opened 2026-02-03
fix: Add timeout configuration for WASM execution

chefsale:
N/A

---

> [!NOTE]
> **Medium Risk**
> Changes WASM execution semantics by introducing a default operation budget and mapping metering exhaustion to a new `ExecutionTimeout` trap, which may affect long-running contracts and precompiled/headless module behavior.
> 
> **Overview**
> **Adds an execution timeout for WASM calls** by introducing operation-limit metering (via `wasmer-middlewares`) and a new `VMLimits.max_operations` default (300M) that is applied at compile time.
> 
> `Engine` creation is refactored to always use Cranelift with metering middleware when enabled, provides `Engine::with_limits`, and explicitly disables operation limiting for `Engine::headless()` (documented as a security/behavior caveat for precompiled modules).
> 
> Runtime error handling now detects metering exhaustion and returns `FunctionCallError::WasmTrap(WasmTrap::ExecutionTimeout)`, with new integration tests covering custom limits, disabled limits, and infinite-loop termination.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3ac5ba369d9b9828248f66c861dee1b1aa6484e4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/node/primitives/src/sync.rs
+++ b/crates/node/primitives/src/sync.rs
@@ -42,6 +42,7 @@ pub mod protocol;
 pub mod snapshot;
 pub mod state_machine;
 pub mod subtree;
+pub mod wire;
 
 // =============================================================================
 // Re-exports
@@ -76,13 +77,16 @@ pub use bloom_filter::{
     BloomFilterRequest, BloomFilterResponse, DeltaIdBloomFilter, DEFAULT_BLOOM_FP_RATE,
 };
 
-// Snapshot and wire protocol types
+// Wire protocol types (used by all sync protocols)
+pub use wire::{InitPayload, MessagePayload, StreamMessage, MAX_TREE_REQUEST_DEPTH};
+
+// Snapshot types
 pub use snapshot::{
-    check_snapshot_safety, BroadcastMessage, InitPayload, MessagePayload, SnapshotBoundaryRequest,
-    SnapshotBoundaryResponse, SnapshotComplete, SnapshotCursor, SnapshotEntity, SnapshotEntityPage,
-    SnapshotError, SnapshotPage, SnapshotRequest, SnapshotStreamRequest, SnapshotVerifyResult,
-    StreamMessage, DEFAULT_SNAPSHOT_PAGE_SIZE, MAX_COMPRESSED_PAYLOAD_SIZE, MAX_DAG_HEADS,
-    MAX_ENTITIES_PER_PAGE, MAX_ENTITY_DATA_SIZE, MAX_SNAPSHOT_PAGES, MAX_SNAPSHOT_PAGE_SIZE,
+    check_snapshot_safety, BroadcastMessage, SnapshotBoundaryRequest, SnapshotBoundaryResponse,
+    SnapshotComplete, SnapshotCursor, SnapshotEntity, SnapshotEntityPage, SnapshotError,
+    SnapshotPage, SnapshotRequest, SnapshotStreamRequest, SnapshotVerifyResult,
+    DEFAULT_SNAPSHOT_PAGE_SIZE, MAX_COMPRESSED_PAYLOAD_SIZE, MAX_DAG_HEADS, MAX_ENTITIES_PER_PAGE,
+    MAX_ENTITY_DATA_SIZE, MAX_SNAPSHOT_PAGES, MAX_SNAPSHOT_PAGE_SIZE,
 };
 
 // Subtree prefetch types
--- a/crates/node/primitives/src/sync/snapshot.rs
+++ b/crates/node/primitives/src/sync/snapshot.rs
@@ -1,6 +1,8 @@
-//! Snapshot sync types and wire protocol messages (CIP §6 - Snapshot Sync Constraints).
+//! Snapshot sync types (CIP §6 - Snapshot Sync Constraints).
 //!
-//! Types for snapshot-based synchronization and general sync wire messages.
+//! Types for snapshot-based synchronization.
+//!
+//! Wire protocol types (StreamMessage, InitPayload, MessagePayload) are in [`super::wire`].
 //!
 //! # When to Use
 //!
@@ -23,10 +25,9 @@ use std::borrow::Cow;
 use borsh::{BorshDeserialize, BorshSerialize};
 use calimero_crypto::Nonce;
 use calimero_network_primitives::specialized_node_invite::SpecializedNodeType;
-use calimero_primitives::blobs::BlobId;
 use calimero_primitives::context::ContextId;
 use calimero_primitives::hash::Hash;
-use calimero_primitives::identity::{PrivateKey, PublicKey};
+use calimero_primitives::identity::PublicKey;
 
 use super::hash_comparison::LeafMetadata;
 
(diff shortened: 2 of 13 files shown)


crates/runtime/src/lib.rs
```diff
@@ -165,9 +173,17 @@ impl Module {
 
         let imports = logic.imports(&mut store);
 
+        let _timeout = self.limits.max_execution_time;
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/runtime/src/errors.rs
```diff
@@ -171,6 +171,8 @@ pub enum WasmTrap {
     Unreachable,
     #[error("unaligned atomic operation")]
     UnalignedAtomic,
+    #[error("execution exceeded timeout")]
+    ExecutionTimeout,
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/runtime/src/lib.rs
```diff
@@ -54,24 +52,58 @@ impl Engine {
...
+            // Cost function: each WASM operation costs 1 unit
+            let metering = Arc::new(Metering::new(fuel, |_op| 1));
+            compiler.push_middleware(metering);
+            debug!(
+                timeout_secs = limits.max_execution_time.as_secs(),
+                fuel, "Execution timeout metering enabled"
+            );
+        }
+
+        wasmer::Engine::from(compiler)
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/runtime/src/lib.rs
```diff
@@ -230,6 +266,7 @@ impl Module {
         logic: &mut VMLogic<'_>,
         method: &str,
         context_id: &ContextId,
+        fuel: u64,
```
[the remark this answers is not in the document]
xilosada: ✅ **Resolved** - This issue has been addressed in the latest changes.

crates/runtime/Cargo.toml
```diff
@@ -25,8 +25,9 @@ tokio-stream.workspace = true
 tracing.workspace = true
 ureq.workspace = true
 wasmer.workspace = true
+wasmer-compiler-cranelift.workspace = true
```
xilosada, line 28: 🟡 **Dependency version not using workspace**

The `wasmer-middlewares` dependency uses a hardcoded version `"6.1.0"` instead of workspace versioning like other wasmer dependencies (`wasmer.workspace = true`, `wasmer-types.workspace = true`). This could lead to version mismatches if the workspace updates wasmer versions.

**Suggested fix:**
```
Add `wasmer-middlewares` to the workspace `[dependencies]` section and use `wasmer-middlewares.workspace = true` here to maintain version consistency with other wasmer crates.
```

crates/runtime/src/lib.rs
```diff
@@ -54,29 +52,55 @@ impl Engine {
...
+            );
+        }
+
+        wasmer::Engine::from(compiler)
     }
 
+    /// Create a headless engine for running precompiled modules.
+    ///
+    /// **Note:** Headless engines cannot enforce operation limits because the metering
+    /// middleware must be applied at compile time. Modules loaded with a headless engine
```
xilosada, line 156: 🟡 **Headless engine silently disables operation limits**

The `headless()` function explicitly disables operation limits by setting `max_operations = 0`. While documented in the docstring, this creates a security risk: precompiled modules loaded via headless engines will run without any execution time protection, potentially allowing infinite loops or DoS if an attacker can cause a headless engine to be used. The `VMLimits` returned by `headless()` differs from `VMLimits::default()` in a security-critical way.

**Suggested fix:**
```
Consider either: (1) preventing headless engines from being used in production contexts where limits are required, (2) returning an error or logging a security warning when `headless()` is used with modules that would normally be limited, or (3) adding a compile-time flag or runtime check to ensure headless engines are only used when explicitly intended.
```

crates/runtime/src/lib.rs
```diff
@@ -54,29 +52,55 @@ impl Engine {
...
-        wasmer::Engine::default()
+        // Add metering middleware if operation limit is set
+        if limits.max_operations > 0 {
+            // Cost function: each WASM operation costs 1 unit
+            let metering = Arc::new(Metering::new(limits.max_operations, |_op| 1));
+            compiler.push_middleware(metering);
+            debug!(
+                max_operations = limits.max_operations,
+                "Execution limit metering enabled"
+            );
```
xilosada, line 145: 🟡 **Headless engine metering inconsistency with precompiled modules**

When a module is compiled with metering enabled (max_operations > 0) but later loaded via a headless engine, the metering middleware is still baked into the compiled code and will trigger traps when exhausted. However, since headless engines force `max_operations = 0`, the error handling in `create_instance_and_call` won't check `get_remaining_points` and won't convert the metering trap to `WasmTrap::ExecutionTimeout`. The error will surface as an opaque runtime error rather than the proper timeout error, making debugging harder and breaking error handling contracts.

**Suggested fix:**
```
Either detect metering traps in the error handler regardless of `max_ops` value, or document clearly that precompiled modules must be loaded with matching engine configurations. Consider adding a method to detect if a loaded module has metering enabled.
```

crates/runtime/src/lib.rs
```diff
@@ -328,12 +364,39 @@ impl Module {
                 "WASM method execution failed"
```
xilosada, line 502: 🟡 **Error context lost when metering exhausted**

When metering points are exhausted, the code returns `ExecutionTimeout` and discards the original runtime error. This shadows any additional error context that wasmer might provide, making debugging harder. The original error should at least be logged before returning the more specific timeout error.

**Suggested fix:**
```
Log the original error before returning ExecutionTimeout: `debug!(%context_id, method, original_error = ?err, "Metering exhausted, returning ExecutionTimeout");`
```

crates/runtime/src/lib.rs
```diff
@@ -328,12 +364,39 @@ impl Module {
...
             );
 
+            // Check if execution exceeded operation limit
```
xilosada: 🟡 **Metering exhaustion check may miss edge cases on error path**

The check for `MeteringPoints::Exhausted` only happens after a RuntimeError occurs. However, wasmer metering can return exhausted state even without a trap if the last instruction exactly depleted the budget. Additionally, if a host function error occurs simultaneously with metering exhaustion, the metering error will shadow the host error, potentially hiding the root cause.

**Suggested fix:**
```
Consider checking metering state before attempting to downcast the error, and potentially include both pieces of information in the error if metering was also exhausted during a host error.
```

crates/runtime/src/lib.rs
```diff
@@ -54,29 +52,55 @@ impl Engine {
         Self { limits, engine }
```
xilosada, line 105: 💡 **Insufficient documentation on Engine::with_limits**

The `with_limits` method lacks documentation about an important constraint: the metering middleware is baked into compiled modules at compile time. If users want different operation limits for the same WASM code, they must recompile with a new Engine. This limitation should be documented to prevent confusion.

**Suggested fix:**
```
Add documentation: `/// Note: The operation limit is embedded at compile time. Modules compiled with this engine
/// will use the specified `max_operations` limit. To use different limits, create a new Engine
/// and recompile the module.`
```

crates/runtime/src/lib.rs
```diff
@@ -613,4 +676,131 @@ mod wasm_panic_integration_tests {
             other => panic!("Expected WasmTrap::StackOverflow error, got: {other:?}"),
```
xilosada, line 1070: 💡 **Missing test for headless engine behavior with precompiled metered modules**

The documentation states that headless engines cannot enforce operation limits, but there's no test verifying this behavior. A test should confirm that a module compiled with metering runs correctly (or with documented behavior) when loaded via a headless engine, especially since the headless engine sets `max_operations = 0`.

**Suggested fix:**
```
Add a test that compiles a module with a full engine (with metering), serializes it, then deserializes with a headless engine and verifies the expected behavior.
```

crates/runtime/src/logic.rs
```diff
@@ -150,6 +154,13 @@ pub struct VMLimits {
     pub max_blob_handles: u64,
     /// The maximum size of a single chunk when writing to or reading from a blob.
     pub max_blob_chunk_size: u64,
+    /// The maximum number of WASM operations allowed per execution.
+    /// Each WASM instruction counts as one operation. When the limit is reached,
+    /// execution is terminated with an `ExecutionTimeout` error.
+    /// This prevents infinite loops and long-running computations from blocking
```
xilosada, line 189: 💡 **Operation limit of zero allows unbounded execution**

Setting `max_operations = 0` completely disables execution limits. If `VMLimits` can be configured by external input or less-trusted configuration sources, an attacker could set this to zero to allow unbounded WASM execution, enabling DoS attacks via infinite loops or computationally expensive operations.

**Suggested fix:**
```
Consider making the zero-disabling behavior opt-in with a separate flag (e.g., `disable_operation_limit: bool`) rather than overloading the meaning of zero. Alternatively, add a minimum allowed value for `max_operations` when limits are enabled, or validate that this value cannot be externally set to zero in production.
```

crates/runtime/src/lib.rs
```diff
@@ -248,6 +277,13 @@ impl Module {
             }
         };
 
+        // Set execution limit before any WASM function calls (including hooks)
+        // to ensure all execution is protected by the operation limit.
+        if max_ops > 0 {
```
xilosada, line 420: 💡 **Metering points reset on every method call including hooks**

The metering points are set to `max_ops` before execution, which is correct. However, after the registration hook (`__calimero_register_crdt_types`) runs, the remaining points aren't preserved. If the hook consumes significant operations, the actual method gets fewer operations than `max_ops`. This is likely intentional for security, but could cause surprising behavior where the same method succeeds or fails depending on whether the hook was called.

**Suggested fix:**
```
Document this behavior explicitly - that `max_operations` is a budget for the entire execution including hooks, not just the user method. Alternatively, consider tracking hook vs method consumption separately for debugging/metrics.
```

crates/runtime/src/lib.rs
```diff
@@ -54,29 +52,55 @@ impl Engine {
...
-                config.enable_perfmap();
-                return wasmer::Engine::from(config);
+                compiler.enable_perfmap();
             }
         }
 
-        // Default engine (no profiling)
-        wasmer::Engine::default()
+        // Add metering middleware if operation limit is set
+        if limits.max_operations > 0 {
```
xilosada, line 138: 💡 **Uniform cost function may allow resource amplification**

The metering cost function assigns each WASM operation a cost of 1 unit (`|_op| 1`), regardless of actual resource consumption. Operations like memory accesses, function calls, or certain instructions may have significantly different real costs. An attacker could craft WASM code that maximizes expensive operations while staying within the operation count limit, potentially causing resource exhaustion (CPU time, memory bandwidth) despite passing the metering check.

**Suggested fix:**
```
Consider implementing a weighted cost function that assigns higher costs to expensive operations (e.g., memory grow, indirect calls, division). This provides better DoS protection by more accurately reflecting actual resource consumption.
```

PR #2169, opened 2026-04-20
feat(auto-follow): DAG-driven group member auto-join

chefsale:
## Summary

End-to-end implementation of **auto-follow group membership** — opt-in, per-member flags that auto-join new contexts (and later subgroups) via governance-DAG events. Design doc: [`architecture/auto-follow.html`](https://github.com/calimero-network/core/blob/feat/auto-follow-phase-1-op-events/architecture/auto-follow.html) (included).

Closes the HA fleet-node convergence gap: a TEE member admitted via `fleet-join` now automatically replicates every subsequent context in the group, no sidecar polling or MDMA-driven resync required.

## Phases in this PR

| Phase | What |
| --- | --- |
| 1 | Process-wide op-apply event channel (`crates/context/src/op_events.rs`) |
| 2 | `AutoFollowFlags` on `GroupMemberValue` (back-compat Borsh migration) + `GroupOp::MemberSetAutoFollow` admin-or-self op |
| 3 | Auto-follow handler task with token-bucket rate limit + cancellable lifecycle |
| 4 | Fleet `fleet-join` self-publishes `MemberSetAutoFollow` (self-signed) after admission — flags default to `{contexts: true, subgroups: true}` for `ReadOnlyTee` |
| 5 | Integration test exercising the full in-process flow (member-add → flag set → context registered → events fire → store state verified) |

Subgroup auto-follow on the handler side (`auto_follow.subgroups`) is intentionally deferred to a follow-up; the flag is persisted and flips fire events, but the emit-self-admission path lives with its own design decision on regular-role semantics.

## Design highlights

- **DAG-native**: every trigger is a governance op applied through the DAG; offline catch-up on restart is free via replay, no reconcile loop.
- **No admission bypass**: every auto-join traverses the same `join_context` / admission paths as a human-driven call.
- **Per-member opt-in**: regular members default to `false/false`; the flag is toggled by an admin-or-self governance op.
- **Rate-limited**: 20 joins/min default. Bucket refill task aborts cleanly on limiter drop.
- **Backward-compatible storage**: `GroupMemberValue` grows a new field via a custom `BorshDeserialize` that distinguishes a clean legacy tail (0 bytes) from partial corruption (1 byte), so existing records keep deserializing and corruption surfaces as an error.

## Review rounds addressed

Two rounds of cursor-bugbot + meroreviewer findings resolved:

- **Round 1**: unasserted `matches!`, dead test code, Borsh EOF masking corruption, handler re-spawn on actor restart.
- **Round 2**: TEE admission auth bug (flag publish moved to fleet_join, signed by self), handler lifecycle (spawn/shutdown + AbortHandle), `RateLimiter` refill task cancellation, backfill truncation warn, `BACKFILL_LIMIT` named constant, `OpEvent: PartialEq`, `u32` bounds on burst.

Design doc lives under the existing `architecture/` site (not in a loose `docs/adr/` folder) and is linked from `nav.js`.

## Test plan

- [x] `cargo check --workspace` clean, `cargo fmt` clean.
- [x] `cargo test -p calimero-context --lib` — 108 pass (includes 6 auto-follow tests, 3 op_events, 2 rate-limiter, 1 end-to-end integration).
- [x] `cargo test -p calimero-store --features datatypes --lib` — 31 pass (includes new legacy/roundtrip/corruption Borsh tests).
- [ ] Reviewer confirms the decisions in `architecture/auto-follow.html` and the shape of the TEE fleet integration.

---

> [!NOTE]
> **Medium Risk**
> Touches governance mutation paths, adds a new background task, and changes persisted `GroupMemberValue` encoding/decoding (migration + corruption handling), which could impact membership behavior and data compatibility if incorrect.
> 
> **Overview**
> Adds an opt-in **auto-follow** mechanism so group members can automatically `join_context` when new contexts are registered, driven by governance-DAG apply events rather than polling.
> 
> This introduces a process-wide `op_events` broadcast and a background `auto_follow` task (spawned from `ContextManager::started`) that rate-limits and backfills joins when `MemberSetAutoFollow` enables context following.
> 
> It extends governance/storage to persist per-member `AutoFollowFlags` (backward-compatible Borsh decode) and adds `GroupOp::MemberSetAutoFollow` with admin-or-self authorization; `fleet-join` now self-publishes this op after TEE admission and surfaces `auto_follow_enabled` in the admin/`meroctl` output. Documentation and tests were added to cover the new op, event emission, migration behavior, and rate limiter.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2107c24804a67fccf75d5c525ba68e1dd250229. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Pushed Phase 2 (\`2ac31f18\`) onto this branch — this PR now covers both:

**Phase 1** (already described above) — governance-DAG op-apply event channel.

**Phase 2** — \`AutoFollowFlags\` + \`MemberSetAutoFollow\` op:
- \`store::key::AutoFollowFlags { contexts, subgroups }\` appended to \`GroupMemberValue\`. Custom \`BorshDeserialize\` reads legacy three-field records by defaulting on EOF; every write upgrades the on-disk layout.
- New \`GroupOp::MemberSetAutoFollow { target, auto_follow_contexts, auto_follow_subgroups }\`.
- \`apply_group_op_mutations\` handles it with admin-or-self authorization.
- \`add_group_member\` preserves existing \`auto_follow\` across upserts so \`MemberRoleSet\` does not clobber flags.
- 5 new unit tests: admin-sets, self-sets, non-admin rejection, unknown-target rejection, flags-survive-role-change.

No observable behavior change yet — the handler that reacts to \`OpEvent::AutoFollowSet\` / \`ContextRegistered\` / \`SubgroupNested\` is Phase 3 and will land as its own PR. REST API / meroctl surface for setting flags manually is deferred; Phase 4 sets them programmatically at TEE admission so CLI is a convenience, not a blocker.

Full suite: \`cargo test -p calimero-context --lib\` — 105 tests pass.

Happy to split into separate PRs if that reads better.

chefsale:
Phase 3 pushed (\`8f8c8b53\`). This PR now covers Phases 1 + 2 + 3.

**Phase 3 — auto-follow handler for context auto-join:**

- New module \`crates/context/src/auto_follow.rs\`.
- Spawned from \`ContextManager::started\` as a detached tokio task that subscribes to \`op_events\`.
- Reactions:
  - \`OpEvent::ContextRegistered { group, context }\` → if self is a member of \`group\` with \`auto_follow.contexts = true\`, emit \`JoinContext\`.
  - \`OpEvent::AutoFollowSet { group, member = self, contexts: true }\` → enumerate existing contexts and join any we haven't. Covers the "flag flipped on after contexts already exist" case without a separate reconcile loop.
- Rate limit: token-bucket, default 20/min per node. Refill task tops up on \`per/burst\` intervals.
- Structured \`info!\` log per auto-join with group_id + context_id; \`warn!\` on failure.

**Subgroup auto-follow (\`auto_follow.subgroups\`) deferred** — will land as a follow-up. For \`ReadOnlyTee\` it reuses the TDX attestation flow from \`fleet_join.rs\`; for regular roles it needs a new admission op since \`MemberAdded\` must be admin-signed. Out of scope here.

**Tests:** 2 unit tests for the rate limiter (burst-then-throttle, serial-acquires-take-expected-time). Full auto-join paths will be exercised by the Phase 5 e2e harness.

Suite: \`cargo test -p calimero-context --lib\` — 107 tests pass.

chefsale:
Phase 4 pushed (\`c56b7138\`). This PR now covers Phases 1 + 2 + 3 + 4.

**Phase 4 — auto-follow on TEE admission:**

\`admit_tee_node\` now publishes \`MemberSetAutoFollow { contexts: true, subgroups: true }\` immediately after \`MemberJoinedViaTeeAttestation\`. TEE fleet nodes are auto-enrolled at admission time — no manual meroctl or MDMA-driven refresh.

Failure mode: if the auto-follow publish fails, admission still succeeds; we \`warn!\`-log. Operators retry via the governance op.

**Effect on the original HA bug:**

Combined with Phase 3, a TEE fleet node that admits via \`fleet-join\` now automatically picks up every context created in the group afterwards. The sidecar no longer needs to poll for new contexts; MDMA no longer needs a client-side resync path.

Subgroup auto-follow (for both \`ReadOnlyTee\` via re-attestation and regular roles via a new admission op) is deferred — separate PR once we pick between the two admission semantics for the regular-role case.

Suite: 107 tests pass.

**Remaining:**
- Phase 5: extend \`e2e-tests/group-nesting\` to cover post-admission context create + offline-then-replay catch-up.
- Phase 6: simplify \`mero-tee\` sidecar (once rc.28+ ships with this PR) and \`tauri-app\` / \`mdma\` client paths.

chefsale:
Phase 5 pushed (\`760f02a7\`). This PR now covers Phases 1 + 2 + 3 + 4 + 5.

**Phase 5 — end-to-end test:**

Added \`end_to_end_event_fires_after_op_apply\` in \`group_store/tests.rs\`. Walks the full Phase 1–4 path in-process:

1. \`MemberSetAutoFollow\` on self → verify flags land in \`GroupMemberValue\`.
2. Admin applies \`ContextRegistered\` op.
3. Assert both \`OpEvent::AutoFollowSet\` and \`OpEvent::ContextRegistered\` fire with correct payloads.

This exercises everything the Phase 3 handler subscribes to short of the actor-driven \`JoinContext\` emission (which needs a full merod instance).

**Deferred e2e:** full two-node merobox scenario needs either

1. Admin API surface for \`MemberSetAutoFollow\` (non-TEE path via meroctl), or
2. \`mero-tee\` CI adding an auto-follow scenario on real TDX hardware (TEE path, which is Phase 4's driver).

Both are beyond this PR; tracked separately.

Suite: **108 tests pass**.

**One more phase left — Phase 6**, which deprecates the sidecar per-group polling and tauri/mdma client-resync paths. That work mostly lives in \`mero-tee\` and \`tauri-app\`, not here; happy to do it in a follow-up PR against those repos after rc.28 ships with this PR.

chefsale:
Review fixes pushed (\`47b66ba5\`). Addressing each finding:

1. **[Med] op_events.rs — \`matches!\` result not asserted** → wrapped in \`assert!()\` with descriptive message.
2. **[Med] lib.rs:281 — duplicate handler on actor restart** → added an \`AtomicBool\` idempotency guard inside \`auto_follow::spawn\`. Only the first call starts the task; later ones \`debug!\`-log and return.
3. **[Low] tests.rs:2857 — dead code + \`.ok()\`** → removed the \`test_store_seed\` indirection, use \`.unwrap()\`, and added a post-condition asserting the target's flags were NOT mutated (robust to future authorization-order refactors).
4. **[Info] Borsh EOF fallback mask corruption** → reworked the custom deserializer to distinguish cleanly: read the trailing bytes one-at-a-time so 0 bytes (legacy) and 1 byte (corruption) are no longer conflated. Added three tests — legacy roundtrip, new-format roundtrip, and corruption rejection. The last one proves a partial trailing byte now surfaces as \`UnexpectedEof\` instead of silently defaulting.
5. **[Info] OnceLock test pollution** → documented the tag-based filter pattern on \`NOTIFIER\` so future test authors follow suit.

Suites:
- \`cargo test -p calimero-context --lib\` — 108 pass
- \`cargo test -p calimero-store --features datatypes --lib\` — 31 pass

chefsale:
Round 3 fixes pushed (\`0e95fee5\`). Five new findings triaged — 3 fixed, 2 deferred with rationale.

**Fixed:**

- \`auto_follow.rs:108\` — Rate limiter silent degradation. \`acquire()\` now returns \`Result<(), RateLimiterClosed>\`. Callers bail the current event on shutdown instead of proceeding unlimited. Both the \`ContextRegistered\` fast path and the backfill loop honor the signal.
- \`fleet_join.rs:195\` — Silent auto-follow failure in response. The fleet-join response JSON now includes \`auto_follow_enabled: bool\`; mirrored on the typed \`FleetJoinResponse\` primitive so \`meroctl tee fleet-join\` surfaces it in both human and JSON output.

**Deferred (principled):**

- \`auto_follow.rs:104\` — \`permit.forget()\` loses tokens on fast-failing ops. Noted in the method doc as an intentional trade-off: the limiter is a coarse amplification guard, not a retry-aware budgeter. Fix (RAII permit guard) is a real API change and not warranted until the trade-off is observed to bite.
- \`auto_follow.rs:125\` — \`HANDLE\` stores only \`AbortHandle\`. The spawn-idempotency guard already blocks re-spawn while a prior handler is live; after \`shutdown()\`, the aborted task drops its \`Arc<RateLimiter>\` which cascades the refill-task abort via \`Drop\`. No observable leak in the intended lifecycle. Happy to switch to storing the \`Arc<RateLimiter>\` if the reviewer has a concrete scenario in mind.
- \`auto_follow.rs:258\` — Backfill races with \`ContextRegistered\` events. Reviewer acknowledges benign; \`join_context\` is idempotent. Duplicate costs one rate-limit token, no incorrect state. Not worth a dedupe set in the handler.

Suite: 108 context + 31 store pass; \`cargo fmt\` clean; workspace check clean.

chefsale:
Round 4 pushed (\`e2107c24\`). Real bug caught by the reviewer — my round-3 \`RateLimiterClosed\` error path was unreachable.

**Root cause**: tokio's \`Semaphore::acquire\` returns \`Err\` only when \`close()\` is explicitly called, not when the \`Arc<Semaphore>\` is dropped. \`shutdown()\` was calling \`abort()\` on the task's \`AbortHandle\`, which cancels the pending \`.await\` at the task layer — the \`Err\` arm never ran.

**Fix**:
- New \`RateLimiter::close()\` method that explicitly closes the semaphore.
- \`HANDLE\` now holds \`(AbortHandle, Arc<RateLimiter>)\` instead of just the handle, so \`shutdown()\` can close the limiter.
- Ordering in \`shutdown()\`: close → abort. Close surfaces the shutdown through the handler's normal control flow (bail the current event); abort is belt-and-suspenders so \`rx.recv()\` also terminates.

This also retroactively closes round-3 defer \"HANDLE only stores AbortHandle\" — the motivation for storing the \`Arc<RateLimiter>\` is exactly to make \`close()\` reachable from \`shutdown()\`.

**New test**: \`rate_limiter_close_makes_acquire_error\` parks a waiter on \`acquire()\`, calls \`close()\`, asserts the waiter resolves with \`Err(RateLimiterClosed)\` and subsequent acquires are also \`Err\`. Suite: 109 pass.

PR #2172, opened 2026-04-21
perf(node/sync): parallelize peer probing and random-peer sync fallback

chefsale:
## Summary

Parallelizes only the **read-only peer probe** in the sync manager. The random-peer sync fallback remains serial after PR review flagged concurrency hazards in `initiate_sync_inner`.

- **`find_peer_with_state`** (uninitialized-bootstrap path): probes race concurrently via `buffer_unordered(DEFAULT_PEER_STATE_PROBE_CONCURRENCY = 4)`. Each probe is a single `DagHeadsRequest` — read-only, no per-context state writes — so racing probes and cancelling the losers is safe.
- **Random-peer sync fallback**: unchanged from master (serial loop). Earlier commits on this branch had parallelised it; reverted in `0be293a4` after review.

## Motivation

Before this change, `find_peer_with_state` iterated peers serially with a `sync_config.timeout / 6` budget per peer (~5 s at defaults). A single slow or unreachable peer could add ~5 s of wall-clock before the next peer was even probed. Now tail latency is bounded by the fastest responder.

## Why the full-sync fallback stayed serial

`initiate_sync_inner` is not concurrency-safe for the same context:

- `snapshot.rs:581` writes a per-context `sync-in-progress` marker to RocksDB without synchronisation; concurrent snapshot syncs would race on it and on the snapshot-page cleanup logic (which documents "assumes no concurrent writes occur to the context's state during snapshot sync").
- `start_sync_session` (state.rs:235) inserts into a DashMap with no Drop guard. A future cancelled between `start_sync_session` and cleanup leaves the session pinned, so `should_buffer_delta` returns true forever and subsequent deltas buffer indefinitely.

Making the full-sync path concurrency-safe would require a per-context lock or Drop-guard refactor and is out of scope for this PR.

## Scope

- Does not change the probe/sync wire protocol, timeouts, or gossipsub config.
- A follow-up PR will likely tune gossipsub mesh parameters (`Config::default()` at `crates/network/src/behaviour.rs:111`) — the more probable root cause of the small-cluster sync lag in merobox workflows.
- Expected impact of this PR alone: tail-latency win on partition / dead-peer scenarios during uninitialized bootstrap. Modest impact on the 3-node merobox flake.

## Test plan

- [x] `cargo check -p calimero-node` passes
- [ ] CI green
- [ ] Run merobox workflows (e.g. `kv-store-with-handlers`, `three-node-sync`) and compare convergence times vs. master
- [ ] Manual: trigger a sync with one peer deliberately slow/unreachable during uninitialized bootstrap and confirm the probe no longer blocks on that peer's timeout

## Review feedback resolution

- Concurrent `initiate_sync` race on sync-in-progress marker / sync session leak / snapshot-page cleanup — **fixed by reverting** `buffer_unordered` in the fallback (kept in probe path, which is read-only).
- PR description mismatch with constants — **fixed** (now reflects the single remaining constant = 4).
- `peers.clone()` in fallback — **N/A after revert** (original loop uses `choose_multiple` which doesn't clone).
- `SyncConfig` field vs module-level const — **deferred**; the one remaining const is narrowly scoped and changing the config surface is out of scope here.
- Probe-path stream cleanup on cancellation — libp2p streams have their own idle timeouts; an explicit Drop guard could be added later if it proves needed in practice.

---

> [!NOTE]
> **Medium Risk**
> Changes sync bootstrap behavior by introducing concurrent peer probing and new config, which can affect network load and timing; core `initiate_sync` path remains serial to avoid per-context state races.
> 
> **Overview**
> Improves uninitialized-node bootstrap by probing multiple peers *concurrently* in `find_peer_with_state` (via `buffer_unordered`) so slow/unreachable peers don’t add per-peer timeout latency before a stateful peer is found.
> 
> Adds a new `SyncConfig.peer_state_probe_concurrency` (default `4`) to cap parallel probes, and clarifies in-code that the subsequent random-peer sync fallback remains **serial** due to per-context sync state concurrency hazards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e36c9dccc450ac62dbe7147bd65883f217331f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/config.rs
+++ b/crates/node/src/sync/config.rs
@@ -60,14 +60,10 @@ pub const DEFAULT_MESH_RETRY_DELAY_MS_UNINITIALIZED: u64 = 1_000;
 /// Max concurrent peer probes when looking for a peer with state.
 /// Typical meshes are 2-20 peers; a pool of 4 is enough parallelism
 /// that the tail is bounded by the fastest responder, without racing
-/// the whole mesh simultaneously on larger deployments.
+/// the whole mesh simultaneously on larger deployments. The probe
+/// itself is read-only (a single `DagHeadsRequest`), so parallelising
+/// it does not risk racing on per-context sync state.
 pub const DEFAULT_PEER_STATE_PROBE_CONCURRENCY: usize = 4;
-
-/// Max concurrent full sync attempts in the random-peer fallback.
-/// Each attempt is heavy (full sync protocol), so we keep this low:
-/// the goal is to avoid a single slow/dead peer stalling the whole
-/// retry chain, not to run parallel syncs against multiple peers.
-pub const DEFAULT_PEER_SYNC_CONCURRENCY: usize = 2;
 
 /// Synchronization configuration.
 ///
--- a/crates/node/src/sync/config.rs
+++ b/crates/node/src/sync/config.rs
@@ -87,6 +87,9 @@ pub struct SyncConfig {
 
     /// Maximum delta gap before falling back to full resync
     pub delta_sync_threshold: usize,
+
+    /// Max concurrent peer probes in `find_peer_with_state`.
+    pub peer_state_probe_concurrency: usize,
 }
 
 impl Default for SyncConfig {
@@ -98,6 +101,7 @@ impl Default for SyncConfig {
             max_concurrent: DEFAULT_MAX_CONCURRENT_SYNCS,
             snapshot_chunk_size: DEFAULT_SNAPSHOT_CHUNK_SIZE,
             delta_sync_threshold: DEFAULT_DELTA_SYNC_THRESHOLD,
+            peer_state_probe_concurrency: DEFAULT_PEER_STATE_PROBE_CONCURRENCY,
         }
     }
 }
--- a/crates/node/src/sync/config.rs
+++ b/crates/node/src/sync/config.rs
@@ -58,15 +58,16 @@ pub const DEFAULT_MESH_RETRIES_UNINITIALIZED: u32 = 10;
 pub const DEFAULT_MESH_RETRY_DELAY_MS_UNINITIALIZED: u64 = 1_000;
 
 /// Max concurrent peer probes when looking for a peer with state.
-/// Each probe is one small round-trip, so we can race many in parallel;
-/// the cap only matters when a context has a large mesh.
-pub const DEFAULT_PEER_STATE_PROBE_CONCURRENCY: usize = 16;
+/// Typical meshes are 2-20 peers; a pool of 4 is enough parallelism
+/// that the tail is bounded by the fastest responder, without racing
+/// the whole mesh simultaneously on larger deployments.
+pub const DEFAULT_PEER_STATE_PROBE_CONCURRENCY: usize = 4;
 
 /// Max concurrent full sync attempts in the random-peer fallback.
-/// Kept small because each attempt is a heavy sync; the goal is to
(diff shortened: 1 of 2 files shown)


chefsale:
Thanks for the careful review — the correctness critique was right. Summary of what I did:

**Reverted the unsafe half** (commit `0be293a4`): `buffer_unordered` in the random-peer sync fallback is gone; that loop is back to the original serial pattern. The concerns are real — I verified each one against the codebase:

- Sync-in-progress marker (`snapshot.rs:581` writes via RocksDB, no guard) would race between concurrent snapshot syncs for the same context.
- `start_sync_session` (`state.rs:235`) inserts into a DashMap with **no Drop guard**. A future cancelled between `start_sync_session` and `cancel_sync_session` leaves the session pinned, so `should_buffer_delta` returns `true` forever and deltas silently buffer indefinitely. Classic Rust async-cancellation pitfall.
- `request_and_apply_snapshot_pages` documents "assumes no concurrent writes occur to the context's state during snapshot sync"; that assumption is load-bearing for its cleanup step.

Any of these alone would justify reverting; together they're decisive.

**Kept the safe half**: parallel probe in `find_peer_with_state`. Each probe opens a stream, sends `DagHeadsRequest`, reads the response — read-only on the local node, no per-context sync state touched. Cancelling the losers just drops the stream; libp2p handles that via its own idle timeouts.

**Nits addressed:**
- PR description now matches the single remaining constant (`DEFAULT_PEER_STATE_PROBE_CONCURRENCY = 4`).
- `peers.clone()` is gone along with the revert (the original loop uses `choose_multiple` on the slice directly).
- `DEFAULT_PEER_SYNC_CONCURRENCY` removed since it's no longer referenced.

**Nits deferred (intentionally):**
- Moving the remaining const into `SyncConfig` fields — sensible code-org point, but out of scope for a narrow perf fix touching one knob.
- Drop guard on probe streams — libp2p's own idle handling is likely sufficient; can revisit if it shows up in profiles.

Net effect: tail-latency improvement in the uninitialized-bootstrap probe path. Does **not** claim to fix the 3-node merobox flake on its own — that's almost certainly the gossipsub default config (`mesh_n_low=4` on a 3-node mesh), which will be a separate PR.

chefsale:
Addressed all three follow-up comments in `1e36c9dc`:

- **SyncConfig field (`config.rs:63`)** — done. Added `peer_state_probe_concurrency: usize` to `SyncConfig` with the existing constant as its default. Usage switched to `self.sync_config.peer_state_probe_concurrency`, so the value is now overridable at runtime / in tests like the other tuning knobs.
- **Lost diagnostic fields on info-level success log (`manager/mod.rs:693`)** — done. Threaded `heads_count` and `root_hash` through the probe result tuple; the `"Found peer with state for bootstrapping"` info log once again emits them, matching the original serial loop.
- **Stream cleanup on cancellation (`manager/mod.rs:648`)** — documented as an accepted trade-off with an inline comment above the probe block. The practical problem is that Rust's `Drop` does not support async work, so a graceful-close Drop guard is not feasible without adding an actor-style cleanup channel. Given the probe is read-only on the local node (no partial state to unwind) and libp2p's idle timeout handles the peer-side stream, the cost is at most a write-error log entry on the responder, which is acceptable. Happy to revisit if this shows up as actual noise in practice.

PR #2176, opened 2026-04-21
ci: capture merobox runtime logs + instrument state-delta latency

chefsale:
… path

Throwaway branch for measuring actual state-delta latency in merobox workflows. Do not merge — this exists purely to collect numbers so we can attribute the "sometimes slow pubsub even when mesh is formed" symptom to a specific stage.

Adds five `info!` events on the `sync_latency` target, correlatable across nodes by `delta_id`:

- `event=publish` on the author node just before `node_client.broadcast`. Emits `publish_at_ms` (Unix epoch) for cross-node wire-time math, plus parent count / event flag / artifact size.
- `event=receive` at entry to `handle_state_delta` on receivers. Emits `receive_at_ms` — diff with `publish_at_ms` for wire latency.
- `event=apply` around `add_delta_with_events` — `apply_ms` includes WASM `__calimero_sync_next`.
- `event=missing_parents` when the post-apply `get_missing_parents` returns a non-empty set — logs `missing_count` so we can attribute lag to the serial backfill loop.
- `event=handlers` around `execute_event_handlers_parsed` — emits `event_count`, `handler_count`, `handlers_ms`.

Usage:
  RUST_LOG="sync_latency=info,warn" merod --node nodeX run ...

Grep logs by `delta_id=...` to reconstruct full per-delta timeline across all three nodes. Expected clean split: wire time (publish→receive), apply time, handler cascade time, missing-parent backfill time.

# [product] short description

## Description

are required for this change.

## Test plan

## Documentation update

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -210,16 +210,24 @@ jobs:
                   --image merod:local \
                   --e2e-mode; then
                   success=true
-                  break
               else
-                  echo "Collecting Docker logs for ${{ matrix.workflow }} attempt $attempt"
-                  for container in $(docker ps -a --format "{{.Names}}" 2>/dev/null || true); do
-                      if [ -n "$container" ]; then
-                          docker logs "$container" > "$GITHUB_WORKSPACE/docker-logs/${{ matrix.workflow }}-attempt${attempt}-${container}.log" 2>&1 || true
-                      fi
-                  done
-                  attempt=$((attempt + 1))
+                  attempt_failed=1
               fi
+
+              # debug branch: always capture runtime docker logs so we can see
+              # SYNC_LATENCY events emitted by the merod containers, not just
+              # the init output. Happens on both success and failure.
+              echo "Collecting Docker logs for ${{ matrix.workflow }} attempt $attempt"
+              for container in $(docker ps -a --format "{{.Names}}" 2>/dev/null || true); do
+                  if [ -n "$container" ]; then
+                      docker logs "$container" > "$GITHUB_WORKSPACE/docker-logs/${{ matrix.workflow }}-attempt${attempt}-${container}.log" 2>&1 || true
+                  fi
+              done
+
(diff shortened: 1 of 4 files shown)


chefsale:
Pushed two fixes to make this branch actually produce data:

1. **Removed `target: \"sync_latency\"`** from the `info!` macros. The merod default tracing filter is `merod=info,calimero_=info` (`merod/src/main.rs:47`) — any event whose target doesn't prefix-match is dropped. Since these handlers live in `calimero_context` and `calimero_node`, using the default module target means the existing filter picks them up. Events now carry `SYNC_LATENCY` as a message prefix plus a `sync_event=` field so `grep SYNC_LATENCY` gives the per-delta trace.

2. **Always capture runtime docker logs in the e2e workflow**, not only on failure. Previously the `for container in $(docker ps ...); do docker logs ... ; done` block was only reached when an attempt failed; success path skipped it. Moved the capture out of the `else` branch so every attempt (success or failure) writes container logs to the existing artifact upload path.

Also renamed the PR title from `debug:` to `ci:` so the Conventional Commits lint doesn't fail.

Check artifacts after the next CI run — the `logs-*` artifacts should contain the actual running-node output with `SYNC_LATENCY` events emitted per delta. Grep any artifact log with:
```
grep 'SYNC_LATENCY\|sync_event=' <log-file>
```
to reconstruct per-delta timing across publish → receive → apply → handlers.

chefsale:
Closing — served its purpose as a scratchpad to collect instrumented runtime logs and identify the bug. Two artifacts survive:

- The log-capture fix was extracted to #2177 (focused PR, off master).
- The pending-children cascade fix (`fix(node/delta): cascade pending children after restoring parent from DB`) will be extracted to its own PR off master once the in-progress CI run on this branch confirms it eliminates the `applied=false` / ~5.5s stall.

The instrumentation itself stays on the `debug/sync-latency-instrumentation` branch in git history; it shouldn't merge (info-level noise on a hot path). Reopening not expected.

PR #2178, opened 2026-04-21
fix(node/delta): cascade pending children after restoring parent from DB

chefsale:
## Summary

Closes a latency bug where a delta received via gossipsub whose parent was **authored locally** stayed stuck in the pending queue for up to 10 seconds, even though its parent was present in the database and the gossipsub mesh was fully formed. Observable symptom: a receiver's \`root_hash\` lagging the author's for ~5 s after a perfectly fast delivery.

## Root cause

The in-memory DAG and the database can disagree about what is "applied."

When a node authors a delta, the delta is persisted to RocksDB with \`applied=true\` but is not held in the in-memory DAG cache once it's flushed. When a child delta then arrives via gossipsub, \`add_delta_with_events\` returns \`applied=false\` because the parent is not in the in-memory DAG. \`get_missing_parents\` locates the parent in the DB and loads it via \`DAG::restore_applied_delta\` — which updates topology but does **not** call \`apply_pending\`. \`DAG::add_delta\` does. So:

- On the first-known-parent path (\`add_delta\`): children in pending get applied automatically via \`apply_pending\`.
- On the parent-restored-from-DB path (\`restore_applied_delta\`): children stay stuck in pending because \`apply_pending\` is never called.

Nothing else triggers reprocessing of pending deltas locally. The periodic sync loop (\`DEFAULT_SYNC_FREQUENCY_SECS = 10\`) eventually fires a sync session that runs \`try_process_pending\`, which is what actually unblocks the stalled delta — hence the observed multi-second lag.

## Fix

After the parent-restore loop in \`get_missing_parents\`, if any parent was actually added to the DAG (either via \`restore_applied_delta\` or \`add_delta\`), explicitly run \`DAG::try_process_pending\`. Diff the pending-id set before vs. after to identify children that got unblocked. For each such child: persist it as \`applied=true\` with \`events=None\` (matching what \`add_delta_with_events\` does on its own cascade path) and forward its events via \`MissingParentsResult.cascaded_events\` so \`execute_cascaded_events\` runs handlers and marks the current delta applied.

Also tightens the stale warning at the call site: the \`"Delta pending - parents exist but not yet applied"\` message used to fire even when the explicit cascade had already applied the delta. It now fires only when the delta is genuinely still unapplied after the cascade.

## Verification

Ran the instrumented debug branch (\`debug/sync-latency-instrumentation\`, run [24722604353](https://github.com/calimero-network/core/actions/runs/24722604353)) which has publish/receive/apply timestamps on every delta. On the 3-node test case that previously showed the stall, node-2's timeline for the affected delta:

\`\`\`
t=12:43:07.890813   apply returned applied=false   (pending)
t=12:43:07.890890   parent restored from DB
t=12:43:07.898611   "Cascaded pending deltas after restoring parent"  (this fix)
t=12:43:07.898692   marked applied in state_delta handler
\`\`\`

**Gap pending -> applied: 7.88 ms.** Was ~5,500 ms before this fix. ~700x tail-latency reduction on the specific path.

All 13 e2e matrix jobs on the same run pass with the fix applied.

## Scope

- Does not change the wire protocol, DAG data structures, or sync protocols.
- Purely adds a cascade step that \`add_delta_with_events\` already performs for its own code path, so the two entry points into the DAG now produce identical on-disk + event state.
- No new config knobs.

## Test plan

- [x] \`cargo check -p calimero-node\` passes
- [x] All 13 e2e matrix jobs green on the instrumented branch containing this fix
- [x] Timing verified via \`SYNC_LATENCY\` instrumentation (reduction from ~5.5 s to ~8 ms on the specific pending-parent-in-DB path)
- [ ] CI green on this PR

---

> [!NOTE]
> **Medium Risk**
> Touches core delta application/sync flow and persistence (`get_missing_parents`, handler cascade detection), which can affect convergence and handler execution ordering if incorrect. Changes are localized but involve locking and DB/DAG consistency logic.
> 
> **Overview**
> Fixes a sync-latency case where a delta could remain pending when its parent already existed in the DB but wasn’t in the in-memory DAG. `DeltaStore::get_missing_parents` now plans DB-found parents, applies/restores them under a single DAG write lock, explicitly runs `try_process_pending`, persists any newly-cascaded deltas, and updates persisted `dag_heads`; it also returns `cascaded_ids` so callers can detect cascade-applied deltas without extra DAG reads.
> 
> Updates the state-delta handler to treat the current delta as applied when it appears in `cascaded_ids`, and to fall back to a direct DAG check after peer-fetching missing parents to avoid incorrectly skipping handlers. Separately, workflow YAMLs drop the `wait_for_sync.check_interval` field across multiple e2e workflows (and tighten comments accordingly).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 441a1016c16840ac255118b7cfb0478bb1acd6ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Both findings addressed in \`ea5c9a5d\`:

**High: cascaded deltas without events never persisted** — real bug. Events-less deltas are never pre-persisted (per \`add_delta_internal\`), so \`handle.get(&db_key)\` returns \`Ok(None)\` for them. The old code treated that as "nothing to persist" and silently skipped, even though WASM had already mutated state — leaving the DB's \`ContextDagDelta\` set out of sync with the DAG on restart. Fixed by rebuilding the record from \`dag.get_delta(cid)\` (authoritative source for delta body) and only reading the DB to recover any stored events. Exactly matches the cascade in \`add_delta_with_events\`.

**Medium: missing \`update_dag_heads\` after cascade** — real bug. In-memory heads shift when cascaded children apply, but the DB's \`dag_heads\` never got updated, so sync handshakes + \`broadcast_heartbeat\` ran with stale state. Fixed by capturing the new heads off the DAG before dropping the write lock, then calling \`context_client.update_dag_heads\` after the persist loop, mirroring the "CRITICAL" write in \`add_delta_with_events\` lines 929-944.

Refactored the whole block so the code shape is an obvious mirror of the existing cascade path: DAG is authoritative for delta body, DB is authoritative for events, both get persisted in one pass, heads get updated at the end.

chefsale:
Addressed the new batch of review comments in \`e63047f2\`. Summary:

**cursor[bot] — "Missing \`dag_heads\` database update after cascade" (Medium)** and **"Cascaded deltas without events never persisted to DB" (High)**: these are the pre-existing cursor comments from commit \`726b00b0\`, re-surfaced on every subsequent commit by the bot without the issues being marked resolved. Both were already fixed in \`ea5c9a5d\` (the previous push) — explicit \`update_dag_heads\` call after the cascade, and \`dag.get_delta(cid)\` used as the authoritative source for delta body so events-less deltas that were never pre-persisted still get written. No new change needed.

**meroreviewer[bot] — "TOCTOU window between DAG write-lock drop and read-lock reacquire" (Medium)**: real and now fixed. The new code collects both the cascaded-delta bodies (cloned) and the current heads inside a single write-lock scope, then releases the lock before any DB I/O. Bodies and heads are now captured atomically, so no amount of concurrent DAG mutation between the lock drop and the DB writes can make them inconsistent.

**meroreviewer[bot] — "DAG read lock held during DB I/O for cascade persistence"**: fixed by the same refactor. The persist loop now runs with zero DAG locks held. The standard "collect-then-release" pattern the reviewer suggested.

**meroreviewer[bot] — "Fully-qualified \`std::collections::HashSet\`"**: fixed — uses the already-imported short form throughout.

**meroreviewer[bot] — "Pending snapshot captured before loop can observe deltas cascaded by \`add_delta\` inside the loop"**: after tracing the flow I think this is a false alarm, *not* a real double-forward risk. A delta can only be in \`pending\` at most once, so the diff counts each cascaded delta exactly once. The \`add_delta\` branch inside the loop calls \`apply_pending\` (via \`apply_delta\`), which removes cascaded children from \`pending\` but does not persist them to the DB or forward their events — that's the *caller's* job. Our post-loop block is the only place that handles persistence + event forwarding for these deltas, so it's not a duplicate of anything. If the bot has a concrete example where double-forwarding happens I'd want to see it, but from the code flow I traced it's handled correctly.

**cursor[bot] — "Cascade-persist logic duplicated across two code paths" (Medium)**: acknowledged. The duplication is real — our cascade-persist block is structurally very close to the one in \`add_delta_with_events:785-886\`. Deliberately leaving it for a follow-up PR rather than bundling with this fix, because extracting a shared helper pulls in the \`update_dag_heads\` site inside \`add_delta_with_events\` too and broadens scope beyond the bug being fixed here. Opened a TODO for a future refactor; happy to do it immediately in a separate PR if reviewers would rather land it together.

chefsale:
Pushed \`108de0e4\`. Handling each of the comments still open on the thread:

**cursor[bot] "Missing \`dag_heads\` database update after cascade" (Medium)** and **"Cascaded deltas without events never persisted to DB" (High)** — both reference commit \`726b00b0\` in the bot metadata. Those were fixed in \`ea5c9a5d\` three commits ago. The bot keeps re-posting them because the original threads were never marked resolved; ignore both.

**meroreviewer[bot] "Pending snapshot captured before loop can observe deltas cascaded by \`add_delta\` inside the loop"** — re-analysed and still a false alarm. \`add_delta\`'s internal \`apply_pending\` updates the DAG but does *not* persist cascaded children to the DB or forward their events — that's the caller's job. Our post-loop block is the only site that handles persistence + event forwarding for deltas cascaded during \`get_missing_parents\`, whether they came via \`add_delta\`'s internal cascade or the explicit \`try_process_pending\` after the loop. A delta can only be in \`pending\` once, so the pending-diff counts each cascaded delta exactly once. No double-forwarding path exists. Leaving unchanged.

**meroreviewer[bot] "Fully-qualified \`std::collections::HashSet\`"** — the sites in \`get_missing_parents\` were already converted in \`e63047f2\`. I also converted the two remaining sites in \`add_delta_internal\` (lines 709, 721 pre-edit) in \`108de0e4\` for file-wide consistency.

**cursor[bot] "Warning fires for successfully cascaded events-less deltas" (Low)** — real cosmetic issue, fixed in \`108de0e4\`. The warn site in \`handle_state_delta\` was checking a stale \`applied\` because \`cascade_outcome.applied_current\` only flips true for deltas that appear in \`cascaded_events\`, and \`cascaded_events\` only carries deltas with persisted events. Events-less deltas that successfully cascade in the DAG kept \`applied=false\` at the warn point and got an incorrect "child did not apply during cascade" message, then the \`was_cascaded\` check a few lines down silently corrected \`applied\`. Moved the \`dag_has_delta_applied\` check + \`applied = true\` flip ahead of the warn branches so the warn fires only when the delta is genuinely still unapplied. No functional change downstream.

chefsale:
Pushed \`a3a77641\`. Handling this round:

**cursor[bot] \"Missing \`dag_heads\` update\" (M)** and **\"Cascaded deltas without events not persisted\" (H)**: both reference commit \`726b00b0\` in the bot metadata. Fixed in \`ea5c9a5d\` five commits ago. The bot keeps reviving them on every push because the threads on the old commit never got marked resolved; not going to patch the same thing again.

**meroreviewer[bot] \"Pending snapshot captured before loop can observe deltas cascaded by \`add_delta\` inside the loop\"** and **meroreviewer[bot] \"TOCTOU gap between pending_before snapshot and parent-addition loop\"**: same root cause, now structurally fixed. \`get_missing_parents\` was restructured into three phases: *plan* (DB lookups, no lock), *apply* (single DAG write lock for restores + adds + \`try_process_pending\` + snapshotting bodies/heads), *persist* (DB writes, no lock). \`pending_before\` now lives inside the same write-lock scope as every DAG mutation and \`try_process_pending\`, so the diff only ever reflects cascades this call actually performed. No TOCTOU window remains.

**meroreviewer[bot] \"Fully-qualified \`std::collections::HashSet\`\"**: fixed in \`108de0e4\`. All remaining sites in \`delta_store.rs\` use the short form.

**cursor[bot] \"Warning fires for successfully cascaded events-less deltas\" (L)**: fixed in \`108de0e4\` by moving the \`dag_has_delta_applied\` check ahead of the warn branch.

**cursor[bot] \"Cascaded delta persistence overwrites stored events with None\" (M)**: not introduced by this PR — \`add_delta_with_events\` has had this exact pattern for a long time, and my change mirrors it intentionally. Keeping \`events: Some(...)\` in the DB until we've *confirmed* the handler ran successfully is a cross-cutting change to the cascade-events lifecycle that would need to touch both cascade paths and update the post-handler cleanup. Out of scope for this PR; happy to open a follow-up issue for it if you want the behaviour tightened.

chefsale:
Addressed both security-reviewer follow-ups in \`909dee3a\`:

**1. Verbose info-level logging** — \`DELTA_DEBUG: Actions in delta before WASM execution\` downgraded from \`info!\` to \`debug!\`, and the action-types \`Vec<&str>\` collection is now gated behind \`tracing::enabled!(Level::DEBUG)\` so it's skipped entirely when debug logging is off. The other two log sites the reviewer referenced (the merge-scenario info at the top and \`DELTA_APPLY_TIMING\` at the bottom) were intentionally left at their current levels — the merge log is rare (only fires when a concurrent-branch merge is detected, which is operationally interesting) and \`DELTA_APPLY_TIMING\` is an established name used by operators to extract per-delta wall-clock numbers for debugging sync lag. Happy to lower those too if preferred, but those felt like genuine info-level signals rather than debug noise.

**2. Double DAG query for cascade detection** — exactly the suggested fix. \`MissingParentsResult\` gains a \`cascaded_ids: Vec<[u8; 32]>\` field carrying *all* cascaded deltas from \`get_missing_parents\`, including events-less ones. \`handle_state_delta\` replaces \`dag_has_delta_applied(&delta_id).await\` (which took the DAG read lock) with \`missing_result.cascaded_ids.contains(&delta_id)\` (no lock, O(n) in the cascaded-set size, typically 1-3 entries). The list is captured under the same write-lock scope as \`try_process_pending\` in the recent refactor so it's race-free by construction.

Build + \`cargo fmt\` clean.

chefsale:
Pushed \`443a6833\`. Handling this round:

**cursor[bot] \"Missing \`dag_heads\` database update\" (M)**, **\"Cascaded deltas without events not persisted\" (H)**, **\"Warning fires for cascaded events-less deltas\" (L)**, **\"Cascaded delta persistence overwrites stored events with None\" (M)** — all stale, all reference old commits (\`726b00b0\`, \`e63047f2\`, \`108de0e4\`). The bot keeps reviving them on every push. Previously addressed (\`ea5c9a5d\`, \`108de0e4\`) or deferred as an existing cross-path pattern (events-overwrite — same concern applies to \`add_delta_with_events\` and would need a cross-cutting refactor of the cascade-events lifecycle).

**cursor[bot] \"Removed fallback misses cascades from peer-fetched parents\" (High)** — real regression I introduced in \`909dee3a\` when swapping \`dag_has_delta_applied\` for \`cascaded_ids.contains\`. The security reviewer's O(1) suggestion was correct for the *common* case (cascades inside \`get_missing_parents\`), but I dropped the *other* cascade path entirely — cascades triggered inside \`request_missing_deltas\` when \`DeltaStore::add_delta\` is fed a peer-fetched parent and internally calls \`apply_pending\`. That loop discards \`AddDeltaResult.cascaded_events\`, so the caller never sees the cascade happened. Old code caught it with \`dag_has_delta_applied\` post-\`request_missing_deltas\`; I removed that check wholesale. Added it back but only in the peer-fetch branch — the \`cascaded_ids.contains\` fast path still covers the get_missing_parents-only cascade with no lock.

**meroreviewer[bot] \"Duplicated cascade persistence logic between add_delta_internal and get_missing_parents\"** — acknowledged earlier in this thread as deliberately out of scope. Extracting a shared helper would pull in \`update_dag_heads\` / event-forwarding ordering in both paths and broaden this PR beyond the bug fix. Happy to do it as a separate refactor PR once this lands.

**meroreviewer[bot] \"dag_heads updated even when no cascade occurred\"** — intentional. \`restore_applied_delta\` mutates in-memory heads (adds the restored delta, removes its parents). Even without a cascade, those heads differ from the pre-call state, so the DB write isn't redundant. And \`update_dag_heads\` is idempotent + cheap (single RocksDB put); a rare spurious write in the edge case where in-memory heads happened to already match DB heads is preferable to carrying extra state to track which writes are truly necessary.

**meroreviewer[bot] \"Linear scan on Vec for cascaded_ids.contains()\"** — micro-perf. Typical \`cascaded_ids.len()\` is 1-3 in practice; linear scan at that size is strictly faster than a HashSet allocation + hash cost. If this ever becomes measurable under load we can switch the field to HashSet without changing the public API.

chefsale:
Handling this round. One actual change, four deferrals with rationale.

**meroreviewer \"Function-scoped enum \`ParentPlan\` should be extracted\"** — done in \`441a1016\`. Moved next to \`MissingParentsResult\` at module level. Behaviour unchanged.

**meroreviewer \"Missing handler execution for deltas applied via peer-fetch cascade\"** — real concern, but pre-existing. \`request_missing_deltas\` calls \`DeltaStore::add_delta\` per fetched parent and discards each \`AddDeltaResult.cascaded_events\`. This PR doesn't introduce that behaviour; the old \`dag_has_delta_applied\` check it replaced never forwarded peer-fetch-cascade events either — it only detected whether the *current* delta was applied. Fixing the broader \"cascades during peer-fetch don't forward events\" bug requires either threading cascaded events out of \`request_missing_deltas\` or pushing \`execute_cascaded_events\` down into it. Separate PR; noting as a TODO.

**meroreviewer \"Duplicated cascade persistence logic between get_missing_parents and add_delta_internal\"** — acknowledged multiple times in this thread. Extracting a shared \`persist_cascaded_deltas\` helper is real code-health work that pulls in \`update_dag_heads\` ordering, event forwarding, and the head-root-hashes write in both paths. Deliberately out of scope for a focused cascade-correctness PR; happy to do it next.

**meroreviewer \"Cascaded delta persistence in phase 3 may produce inconsistent DB state on partial failure\"** — also pre-existing: \`add_delta_with_events\`'s cascade persistence loop has the same non-atomic pattern. The concrete failure mode requires \`handle.put\` partial success across multiple keys, which on RocksDB is rare but not impossible on disk-full or similar. Would need a \`WriteBatch\` wrapper that both cascade paths use. Noted; out of scope for this PR.

**meroreviewer \"Extended DAG write-lock held during WASM execution\"** — intentional trade-off from the earlier TOCTOU fix. \`apply_pending\` inside \`try_process_pending\` has always required the write lock (it mutates \`self.pending\`, \`self.applied\`, \`self.heads\`), so the only behavioural change is that we now hold the write lock across the parent-restore loop *and* \`try_process_pending\` instead of per-iteration + separate. The DoS surface is bounded to one \`DeltaStore\` (per-context), and the flood scenario the bot describes requires an attacker to get many deltas pending on a common parent that's only in the DB, which already requires legitimate gossipsub participation. Capping \`try_process_pending\` with periodic lock release would be a real architectural change to \`calimero-dag\`; deferring to a dedicated \"cascade throttling\" PR if this ever becomes a real-world issue.

chefsale:
Good point — converted the deferred concerns from this PR into tracked issues so they don't get lost in this thread:

- **#2182** — node/delta: forward cascaded events produced by peer-fetched parents (from the peer-fetch-cascade finding)
- **#2183** — node/delta: extract shared \`persist_cascaded_deltas\` helper (from the duplicated-cascade-persistence finding)
- **#2184** — node/delta: use \`WriteBatch\` for atomic cascade-delta persistence (from the partial-failure finding)
- **#2185** — node/delta: preserve events in DB until handler execution is confirmed (from the events-cleared-before-handler-confirms finding)
- **#2186** — node/delta: throttle cascade processing to avoid long DAG write-lock holds (from the lock-hold-DoS finding)

Each issue names the PR review thread it came from, describes the fix shape, and lists acceptance criteria. #2183 is the natural sequencing anchor — #2184 and #2185 get easier once the shared persist helper exists.

This PR stays scoped to the original bug fix (pending-children cascade after parent restore). The structural follow-ups land separately against master.

chefsale:
**cursor[bot] \"Add-path parent not persisted as applied in DB\" (Low)** — real finding, but pre-existing. Verified master's \`get_missing_parents\` \`else\` branch (pre-#2178) has the exact same gap: when a parent is loaded from DB with \`applied: false\` and run through \`dag.add_delta\`, the DB record is never re-written with \`applied: true\`. My PR inherits the behaviour rather than introducing it. Bot is right that this is an on-disk inconsistency the other cascade path (\`add_delta_internal\`) deliberately avoids.

Tracked as **#2187** — sits alongside #2182-#2186 in the cascade-path follow-up bundle. Natural to land together with #2183 (shared \`persist_cascaded_deltas\` helper), since once that helper exists, making it also handle the Add-path parent is a one-line change rather than two duplicated call sites.

Not fixing in this PR for the same reason as the other deferrals: it's a pre-existing pattern, it doesn't regress behaviour, and the fix naturally belongs in the shared-helper PR rather than this narrow cascade-correctness bug fix.

PR #2180, opened 2026-04-21
perf(ci): install merobox from apt instead of building from git

chefsale:
## Summary

Replaces the \`setup-merobox\` composite action's \`uv pip install git+...\` path with a prebuilt merobox install from the project's APT repository. Cuts ~40-60s off every CI job that uses merobox (e2e-rust-apps, e2e-rust-apps-release, fuzzy-load-test).

Before:
\`\`\`yaml
- uses: actions/setup-python@v5
  with:
    python-version: '3.11'
- shell: bash
  env:
    RUST_LOG: warn   # workaround: maturin stdout parser trips on tracing output
  run: |
    python -m pip install --upgrade pip
    python -m pip install uv
    uv pip install --system \
      "calimero-client-py @ git+https://github.com/calimero-network/calimero-client-py.git@master" \
      "merobox @ git+https://github.com/calimero-network/merobox.git@master"
\`\`\`

After:
\`\`\`yaml
- shell: bash
  run: |
    curl -fsSL https://calimero-network.github.io/merobox/gpg.key \
      | sudo tee /usr/share/keyrings/merobox.gpg > /dev/null
    echo "deb [signed-by=/usr/share/keyrings/merobox.gpg] https://calimero-network.github.io/merobox stable main" \
      | sudo tee /etc/apt/sources.list.d/merobox.list
    sudo apt-get update
    sudo apt-get install -y merobox
    merobox --version
\`\`\`

## What this removes

- Python setup (~5-10s).
- \`uv pip install\` step, including the transitive \`calimero-client-py\` sdist build via \`maturin\` (~30-40s on cold caches).
- The \`RUST_LOG=warn\` workaround — that existed solely because maturin's stdout parser was getting confused by tracing output while building the sdist. No source build = no workaround needed.

## Also removed

\`.github/actions/setup-merobox-pypi\` — a stale copy of an earlier APT migration attempt whose content was almost identical to this PR's action, with a misleading \`-pypi\` name (it's APT, not PyPI) and no workflow references. Deleting it to avoid future confusion about which action to use.

## Scope

All three consumers of \`./.github/actions/setup-merobox\` pick this up transparently — same action path, same exposed \`merobox\` binary on \`PATH\`. No workflow YAMLs changed.

- \`.github/workflows/e2e-rust-apps.yml\`
- \`.github/workflows/e2e-rust-apps-release.yml\`
- \`.github/workflows/fuzzy-load-test.yml\`

## Test plan

- [x] Action YAML is valid GitHub Actions syntax (no changes to input/output contract)
- [ ] CI on this PR exercises \`e2e-rust-apps\` with the new action and all 13 matrix entries pass
- [ ] Observable time saving in the Setup Merobox step vs. pre-PR runs on master

## Notes

- APT package availability is the project's responsibility (hosted at \`calimero-network.github.io/merobox\`); if the repo goes down, CI fails closed on the \`apt-get install\` step. Reverting this PR is trivial if that ever becomes operational pain.
- Linux-only. All three consuming workflows run on \`ubuntu-*\` runners so this is a non-regression. The previous action worked on macOS via Python but nothing currently exercises that.

---

> [!NOTE]
> **Medium Risk**
> CI behavior changes and shifts `merobox` installation to an external APT repository; failures or version mismatches there would break workflows, but no production/runtime code is affected.
> 
> **Overview**
> Updates the `setup-merobox` composite action to install the prebuilt `merobox` package from the project’s APT repository, removing Python setup/`uv pip install` and the source-build workaround env config.
> 
> Deletes the unused `.github/actions/setup-merobox-pypi` action, and removes the `pull_request` trigger from `fuzzy-load-test` (leaving `push` to `master` and `workflow_dispatch`) to avoid failing PR runs that depend on unpublished profiling images.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 974ac229f445b15e7428f6b483f0af295f262382. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Failing \"Run KV Store Fuzzy Test with Profiling\" isn't caused by the apt-install change itself — it's a pre-existing CI trigger misconfiguration that my PR is the first to hit.

What happens:
- \`fuzzy-load-test.yml\` has a PR trigger on \`.github/actions/setup-merobox/**\` (line 19).
- The fuzzy test then asks \`scripts/determine-image-tag.sh\` for the profiling image \`ghcr.io/calimero-network/merod:pr-{N}-profiling\`.
- The release workflow's \`release-merod-profiling-container\` job is gated \`if: \${{ github.event_name != 'pull_request' }}\`, so that tag is never published for PRs.
- \`docker pull\` 500s with \`"manifest unknown"\`, the test aborts.

Pushed \`8308dcaf\` removing the stray trigger. Justification: fuzzy-load-test is a perf/soak test; its behaviour is orthogonal to how merobox is installed. The e2e-rust-apps workflow already exercises setup-merobox end-to-end on every PR, which is the right place to catch install-path regressions. Fuzzy-load-test still fires on master pushes, on \`workflows/fuzzy-tests/**\` or its own YAML changes, and via \`workflow_dispatch\`.

chefsale:
Handled in \`ce399d8e\`:

**cursor[bot] \"PR path trigger for setup-merobox accidentally removed\" (Medium)** — fair point, even though the original removal was intentional (see the earlier PR comment explaining why PRs that touch setup-merobox fail against \`release-merod-profiling-container\` being gated off for PRs). The asymmetry between \`push.paths\` and \`pull_request.paths\` was the actual ugliness the bot caught. Dropped setup-merobox from \`push.paths\` too so both are symmetric. Rationale stands: e2e-rust-apps is the workflow that validates the merobox install path end-to-end; fuzzy-load-test is a perf/soak test whose behaviour is orthogonal.

**meroreviewer[bot] \"GPG key fetched over HTTPS but stored without dearmoring\"** — wrong in this specific case. Verified the key that this action fetches:

\`\`\`sh
\$ curl -fsSL https://calimero-network.github.io/merobox/gpg.key | file -
/dev/stdin: OpenPGP Public Key Version 4, Created Tue Jan 27 12:50:25 2026, RSA (Encrypt or Sign, 4096 bits); User ID; Signature; OpenPGP Certificate
\`\`\`

It's already binary OpenPGP, not ASCII-armored. Piping through \`gpg --dearmor\` here would at best pass through unchanged and at worst error with \"No valid OpenPGP data found\" on older GnuPG. Leaving as-is.

The second half of that comment (\"no pinning or checksum verification of the key\") is a real concern, but pinning the key SHA means every key rotation needs a coordinated PR to update it. Noted as a potential follow-up rather than a blocker — the repo we're fetching from is \`calimero-network.github.io\`, owned by the same org, so it's not really a \"third-party\" supply-chain vector in the usual sense.

**meroreviewer[bot] \"External APT repository added without version pinning\"** — same rationale: the APT repo is owned by \`calimero-network\` (same org as this repo), and the pre-PR action installed merobox from \`git+https://github.com/calimero-network/merobox.git@master\` — also no pin, same trust boundary. Version pinning \`merobox=<VERSION>\` would be a real reproducibility win, but it requires someone to maintain a \"current supported version\" marker and bump it explicitly; out of scope for this PR's narrow switch from pip-install-from-git to apt-install-from-same-org. Happy to do it as a follow-up if you want.

Issue #2185, opened 2026-04-21
node/delta: preserve events in DB until handler execution is confirmed


chefsale:
## Context

Deferred from #2178 (review comment by cursor[bot]).

Both cascade-persist paths in \`crates/node/src/delta_store.rs\` unconditionally write \`events: None\` into the \`ContextDagDelta\` record after reading \`stored_events\` moments earlier and pushing them into an in-memory \`Vec\`:

\`\`\`rust
// crates/node/src/delta_store.rs (both cascade paths)
if let Some(ref events_data) = stored_events {
    all_cascaded_events.push((*cid, events_data.clone()));
}

let record = ContextDagDelta {
    /* … */
    events: None,   // cleared immediately
};
handle.put(&db_key, &record)?;
\`\`\`

The caller then runs \`execute_cascaded_events\`. If that fails — e.g., because the application bundle isn't yet available (\`ensure_application_available\` returns err), or WASM execution itself fails, or the process crashes between the DB write and the handler call — the events data is gone:

- DB now has \`events: None\` on the cascaded delta
- The in-memory \`Vec<([u8; 32], Vec<u8>)>\` is ephemeral and evaporates when the caller returns / the task is dropped
- On restart, \`handle.get(&db_key)\` returns \`events: None\` → handlers permanently skipped

## Fix shape

Preserve \`events\` in the DB record until the handler is confirmed to have executed. Two patterns that work:

1. **Delayed clear**: after \`execute_cascaded_events\` succeeds for a cascaded delta, re-write the DB record with \`events: None\`. More DB writes but clean failure semantics.

2. **Idempotent handler execution + crash-safe replay**: keep \`events: Some\` in the DB indefinitely but mark handler-executed separately; on startup, re-run handlers for any delta whose \`events\` is still present. Requires handlers to be idempotent (they probably aren't yet).

(1) is the smaller change. (2) is the more principled answer if handlers grow to non-idempotent side effects.

## Observable symptom

- Receiver cascades delta X with events; handler \`foo\` starts; process dies mid-handler
- On restart, \`load_persisted_deltas_into_dag\` replays X without events → handler never completes
- State converges (CRDT), but \`foo\`'s side effects (counters, external notifications, etc.) never happen

## Dependencies

- Logically groups with #2183 (shared cascade-persist helper). Easier to land once the helper exists — one place to add the delayed-clear step.
[...]

PR #2189, opened 2026-04-21
fix(node/delta): forward cascaded events from peer-fetched parents

### LINKED ISSUE #2182: node/delta: forward cascaded events produced by peer-fetched parents
## Context

Deferred from #2178 (review comment by meroreviewer).

\`handle_state_delta\` calls \`request_missing_deltas\` when a delta's parents are truly missing from both the DAG and the DB. That function fetches each missing parent from a peer and feeds it back through \`DeltaStore::add_delta\` at \`crates/node/src/handlers/state_delta/mod.rs:1002\`:

\`\`\`rust
for (dag_delta, delta_id) in fetched_deltas {
    if let Err(e) = delta_store.add_delta(dag_delta).await {
        warn!(?e, %context_id, delta_id = ?delta_id, "Failed to add fetched delta to DAG");
    }
}
\`\`\`

\`DeltaStore::add_delta\` internally calls \`apply_pending\`, which can cascade-apply children waiting on the just-fetched parent. Each of those cascades may carry events (stored in the \`ContextDagDelta.events\` column) that should be forwarded to \`execute_event_handlers_parsed\`. Right now \`AddDeltaResult.cascaded_events\` is discarded every iteration — the cascade apply happens in memory and on disk, but no handler ever runs for those cascaded deltas.

This is a pre-existing bug, not a regression from #2178. #2178 replaced an old \`dag_has_delta_applied\` check with \`cascaded_ids.contains(&delta_id)\`, but neither the old nor the new check ever forwarded cascaded-event handlers from the peer-fetch path; the old check only updated the *current* delta's \`applied\` flag.

## Fix shape

Two options, both reasonable:

1. Have \`request_missing_deltas\` collect and return \`cascaded_events\` from each \`add_delta\` call. Caller (\`handle_state_delta\`) runs \`execute_cascaded_events\` on the aggregated set after the loop.

2. Invoke \`execute_cascaded_events\` directly inside \`request_missing_deltas\` whenever \`add_delta\` returns a non-empty \`cascaded_events\`. Needs plumbing the \`node_clients\`/\`our_identity\`/\`sync_timeout\` arguments down.

(1) keeps the separation-of-concerns (fetching vs. handler execution); (2) is tighter locality but adds more args.

## Observable symptom

If a peer delivers child delta X before its parent P, and X carries an event:
- X goes pending on receive
- Receiver requests P from another peer
- \`add_delta(P)\` cascades X → X is applied in DAG + DB
- But X's event never reaches \`execute_event_handlers_parsed\` → handler silently never runs
- State converges across nodes (CRDT), but handler-driven side effects diverge

## Acceptance

- [ ] When a peer-fetched parent triggers a child cascade, the child's handler(s) run exactly once (matching the behaviour when the child arrives after its parent).
- [ ] Existing tests still pass; add a targeted test that exercises "child arrives before parent, events present" across 2 nodes.

## References

- PR #2178 review comment: https://github.com/calimero-network/core/pull/2178#discussion_r...1754fe2064c0
- Relevant code path: \`crates/node/src/handlers/state_delta/mod.rs:879-1005\` (\`request_missing_deltas\`), plus the caller at line 360-372.

chefsale:
Closes #2182.

## Summary

When a child delta arrives via gossipsub before its parents are in the DAG or DB, `handle_state_delta` fetches the missing parents from a peer via `request_missing_deltas`. Each peer-fetched parent is then fed back through `DeltaStore::add_delta`, whose internal `apply_pending` can cascade any pending children — carrying along events that were pre-persisted when those deltas went pending.

Before this change, `request_missing_deltas` called `add_delta`, which discards `AddDeltaResult.cascaded_events`. Cascaded children got applied in DAG + DB, but their events never reached `execute_cascaded_events`, so their handlers **never ran**. State still converged across nodes (CRDT), but handler-driven side effects silently diverged — metrics, counters, outbound notifications, whatever the handler does.

## Fix

- `request_missing_deltas` switched to `add_delta_with_events(delta, None)` so each call's `cascaded_events` is visible. Accumulates them into a `Vec<([u8; 32], Vec<u8>)>` and returns that. Signature changed from `Result<()>` to `Result<Vec<([u8; 32], Vec<u8>)>>`.
- The caller in `handle_state_delta` runs `execute_cascaded_events` on the returned list — same path already used for cascades inside `get_missing_parents` — and updates `applied` and `handlers_already_executed` from the outcome.
- Kept the existing `dag_has_delta_applied` check after `request_missing_deltas`: events-less deltas aren't in `cascaded_events`, but the DAG state still reflects the apply, so we mustn't warn "child did not apply" in that case.

The peer-fetched parent itself has no events on the wire (the `DeltaResponse` message doesn't carry them), so `None` is passed as the events arg. Only cascaded children contribute to the returned list.

## Scope

One file, one function signature change, one call site updated. No protocol changes, no storage changes, no new config.

Only one caller of this `request_missing_deltas` in the codebase (in `state_delta/mod.rs` itself). The sync manager has a differently-typed method of the same name (`crates/node/src/sync/delta_request.rs`) that this PR doesn't touch.

## Observable behaviour change

Scenario: receiver gets child delta X with events; X's parent P isn't yet on the receiver.

| Before | After |
|---|---|
| `X` pending, `P` fetched from peer, `add_delta(P)` cascades X in DAG + DB | Same up to here |
| `X`'s events stored in DB, never read back, no handler ever runs | `X`'s events returned up the call chain, handler runs via `execute_cascaded_events` |

State convergence is unchanged (already happened). Handler-execution correctness is restored.

## Test plan

- [x] `cargo check -p calimero-node` passes
- [x] `cargo fmt --check` clean
- [ ] CI green
- [ ] Ideally: integration test that delivers a child delta before its parent, with events on the child, and asserts the child's handler runs exactly once. (Not adding in this PR; targeted fix first, test coverage as follow-up.)

---

> [!NOTE]
> **Medium Risk**
> Changes state-delta sync/handler execution flow so newly surfaced cascades run application handlers; mistakes could lead to missed or duplicated handler side effects during peer catch-up.
> 
> **Overview**
> Fixes a gap in `handle_state_delta` where fetching missing parent deltas from a peer could *apply* pending children but silently drop their persisted `events`, causing handlers to never run.
> 
> `request_missing_deltas` now uses `DeltaStore::add_delta_with_events(..., None)` and returns aggregated `cascaded_events`; the caller runs `execute_cascaded_events` on that list and updates `applied`/`handlers_already_executed`, with an additional DAG applied check to avoid misleading "still pending" warnings for events-less cascades.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95e0a28aaeae834ecc80e54d93e0b6f3d554264d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/handlers/state_delta/mod.rs
+++ b/crates/node/src/handlers/state_delta/mod.rs
@@ -357,7 +357,7 @@ pub async fn handle_state_delta(
                 "Delta pending due to missing parents - requesting them from peer"
             );
 
-            if let Err(e) = request_missing_deltas(
+            match request_missing_deltas(
                 network_client,
                 sync_timeout,
                 context_id,
@@ -368,20 +368,39 @@ pub async fn handle_state_delta(
             )
             .await
             {
-                warn!(?e, %context_id, ?source, "Failed to request missing deltas");
+                Ok(peer_fetch_cascaded_events) => {
+                    // Peer-fetched parents can cascade pending children via
+                    // `apply_pending` inside `add_delta_with_events`. Those
+                    // cascaded children's events were discarded before this
+                    // fix — now they ride back on `peer_fetch_cascaded_events`
+                    // and go through `execute_cascaded_events` exactly like
+                    // the cascade path inside `get_missing_parents`.
+                    if !peer_fetch_cascaded_events.is_empty() {
+                        let cascade_outcome = execute_cascaded_events(
+                            &peer_fetch_cascaded_events,
+                            &node_clients,
+                            &context_id,
+                            &our_identity,
+                            sync_timeout,
+                            "peer-fetch cascade",
+                            Some(&delta_id),
+                        )
+                        .await?;
+                        applied |= cascade_outcome.applied_current;
+                        handlers_already_executed |= cascade_outcome.handlers_executed_for_current;
+                    }
+                }
+                Err(e) => {
+                    warn!(?e, %context_id, ?source, "Failed to request missing deltas");
+                }
             }
 
-            // `request_missing_deltas` fetches parents from peers and calls
-            // `DeltaStore::add_delta` internally. Those `add_delta` calls
-            // can cascade the current delta via `apply_pending`, but
-            // `add_delta`'s `AddDeltaResult.cascaded_events` is discarded
-            // inside the fetch loop — there's no way for us to see which
-            // deltas it cascaded without checking the DAG. `cascaded_ids`
-            // from the earlier `get_missing_parents` doesn't cover this
-            // path (cascades happened *after* that call returned), so
-            // consult the DAG directly here to avoid the misleading
-            // "child did not apply" downstream warn + permanent handler
-            // skip on peer-fetched parents.
+            // Some peer-fetched cascades may still apply the current delta
+            // without having its events in the DB (events-less deltas are
+            // never pre-persisted, so they won't show up in
+            // `peer_fetch_cascaded_events`). The DAG state reflects the
+            // apply regardless; check it before falling through to the
+            // "still pending" path so we don't warn misleadingly.
             if !applied && delta_store_ref.dag_has_delta_applied(&delta_id).await {
                 info!(
                     %context_id,
@@ -896,6 +915,14 @@ fn emit_state_mutation_event_parsed(
 ///
 /// Opens a stream to the peer and requests each missing delta sequentially.
 /// Adds successfully retrieved deltas back to the DAG for processing.
+/// Fetch missing ancestor deltas from a peer and add them to the DAG in
+/// topological order.
+///
+/// Returns the aggregated `cascaded_events` from every `add_delta_with_events`
+/// call. Each peer-fetched parent that resolves a pending child carries that
+/// child's stored events along in its `AddDeltaResult`; callers *must* run
+/// `execute_cascaded_events` on the returned list, otherwise handler execution
+/// for cascaded children silently never happens.
 async fn request_missing_deltas(
     network_client: calimero_network_primitives::client::NetworkClient,
     sync_timeout: std::time::Duration,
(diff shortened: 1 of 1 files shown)


chefsale:
Traced the concrete error paths and concluded this is a shape-worth-fixing but not a live bug today.

**Today**: every code path inside \`execute_cascaded_events\` that could produce an error is already converted to \`warn!\` before it reaches the \`?\` operator — \`ensure_application_available\` is called with \`.is_ok()\`, serde_json errors hit an outer \`warn!\`-and-continue arm, and handler WASM failures are swallowed inside \`execute_event_handlers_parsed\`'s inner match. The \`?\` you flagged is a trap door nothing currently walks through.

**However**, two things are still worth addressing separately from this PR:

1. The shape invites future regressions. A new error path added to \`execute_cascaded_events\` would silently acquire \"abort the main delta's handler + WebSocket emission + heartbeat broadcast\" semantics, which almost certainly isn't what the author intended.

2. The handler-error-is-\`warn!\`-and-drop behaviour inside \`execute_event_handlers_parsed\` is itself a policy choice worth re-examining — handlers with external side effects (counters, outbound notifications) silently disappearing on error isn't obviously right. That's the same question #2185 forces us to confront.

Filed **#2190** — \"audit error-propagation semantics around cascade handler execution\" — which ties the \`?\` concern to #2185's handler-confirmation work. Both are about the underlying design question: when a cascade handler can't run or fails, what happens to the event? Right answer requires picking one of log-and-continue / retry-via-sync / durable retry queue and applying it uniformly to both call sites.

Not changing this PR: the issue is pre-existing (same pattern at the \`get_missing_parents\` call site), my PR doesn't widen the attack surface in practice (no new propagating error path), and fixing properly needs a design decision that spans more than this fix.

PR #2212, opened 2026-04-22
feat(storage): add AuthoredMap and AuthoredVector primitives (#2201)

### LINKED ISSUE #2201: proposal: AuthoredMap<K, V> and AuthoredVector<V> — shared keyspace with per-entry ownership
## Problem

`StorageType::User` today is tied to `UserStorage<T>`, which is internally `UnorderedMap<PublicKey, T>` — one slot per user, disjoint keyspaces. This covers "each user has their own data" cleanly, but leaves a gap: **a shared keyspace where any member can insert new entries, but each entry is owned by its creator forever** (only the creator can modify/delete their own entries).

This gap cannot be closed by composing existing primitives. Concretely:

- **`UnorderedMap<K, UserStorage<V>>`** — inner `UserStorage` is still `PublicKey → V`, not "owned by one user." Any pubkey can claim its own slot at each outer key. Outer map is `Public`, so K can be overwritten wholesale. No per-entry owner.
- **`UnorderedMap<K, SharedStorage<V>>`** (after #2197) — similar: outer is `Public`, so a tampered peer can overwrite K with a different `SharedStorage<V>` carrying different writers. Per-entry ownership defeated at container level ("override the key" asymmetry documented in #2197).
- **`Vector<UserStorage<V>>`** — each Vector position is Public; any member can replace the whole `UserStorage<V>` at a position. Same hole.

The root cause is structural: auth stamps live on the innermost storage element via `StorageType`, but enclosing `Public` containers are not protected by those inner stamps. To get per-entry ownership in a shared keyspace, the **stamp has to live on the container entry itself**.

### Concrete motivating case

`calimero-studio/foundation-app` chat:

- Messages need a **shared ordered keyspace** (everyone sees the chronological feed).
- Each message needs **per-sender edit/delete authorization**.
- Tampered-WASM peers must not be able to forge edits to other senders' messages or selectively hide them from the feed.

Current viable patterns with existing primitives:

- `Vector<Message>` (Public) — O(limit) reads with ordering, but no auth: a tampered peer can forge edits or call `Vector::update(idx, tombstone)` on anyone's message.
- `UserStorage<UserMessages>` where `UserMessages = { messages: UnorderedMap<String, Message> }` — full auth, but disjoint per-user slots. Reads become `O(senders × msgs_per_sender)` because there's no shared ordering.
- **Hybrid** (`UserStorage` + public `Vector<MessageRef>`) — restores `O(limit)` reads but the public Vector is tamperable: any member can `update(idx, junk)` on an existing ref, and readers skip unresolved refs → selective message hiding. Not viable.

The primitive that cleanly solves this doesn't exist today.

## Describe the solution you'd like

Two primitives sharing the same mechanism (per-entry `StorageType::User { owner = creator_at_insert }`), differing only in collection shape:

### `AuthoredMap<K, V>`

```rust
pub struct AuthoredMap<K, V: Mergeable> {
    inner: UnorderedMap<K, V>,     // entries stamped individually
    storage: Element,
}

impl<K, V> AuthoredMap<K, V> {
    fn new() -> Self;
    fn new_with_field_name(field_name: &str) -> Self;

    /// Insert a new entry. Stamps with owner = executor_id.
    /// Fails if K already exists (use `update` or choose a new K).
    fn insert(&mut self, k: K, v: V) -> Result<(), StoreError>;

    /// Update existing entry. Runtime signs with executor's keypair;
    /// peers verify signer == stored owner on merge. Rejected otherwise.
    fn update(&mut self, k: &K, v: V) -> Result<(), StoreError>;

    /// Remove existing entry. Same auth rule as update.
    fn remove(&mut self, k: &K) -> Result<Option<V>, StoreError>;

    fn get(&self, k: &K) -> Result<Option<V>, StoreError>;
    fn owner_of(&self, k: &K) -> Result<Option<PublicKey>, StoreError>;
    fn entries(&self) -> Result<impl Iterator<Item = (K, V)>, StoreError>;
}
```

### `AuthoredVector<V>`

Same mechanism, ordered by insertion position:

```rust
pub struct AuthoredVector<V: Mergeable> {
    inner: Vector<V>,              // each position stamped individually
    storage: Element,
}

impl<V> AuthoredVector<V> {
    fn push(&mut self, v: V) -> Result<usize, StoreError>;      // anyone; position owned by pusher
    fn get(&self, idx: usize) -> Result<Option<V>, StoreError>;
    fn update(&mut self, idx: usize, v: V) -> Result<(), StoreError>;  // signer == owner
    fn tombstone(&mut self, idx: usize) -> Result<(), StoreError>;     // signer == owner
    fn owner_of(&self, idx: usize) -> Result<Option<PublicKey>, StoreError>;
    fn len(&self) -> Result<usize, StoreError>;
    fn iter(&self) -> Result<impl Iterator<Item = V>, StoreError>;
}
```

### Shared mechanics

Both use `StorageType::User { owner, signature_data }` per entry — the **same storage type that already exists for `UserStorage`**. What's new is applying it per-entry in a shared keyspace rather than keying the whole collection by `PublicKey`.

- **Insert / push**: runtime stamps the entry with `owner = executor_id`, signs the delta with the executor's keypair.
- **Update / remove / tombstone**: runtime signs; peers verify `signer == stored owner` on merge. Mismatched deltas are rejected.
- **Reads**: unrestricted for any context member.
- **Owner is immutable** per entry (set at insert, never rotated). For rotatable writer sets, use `SharedStorage` (#2197).

### Usage: chat with `AuthoredVector`

```rust
#[app::state]
pub struct RoomState {
    messages: AuthoredVector<Message>,
}

impl RoomState {
    pub fn send_message(&mut self, body: String) -> app::Result<usize> {
        self.messages.push(Message { body, timestamp_ms: now(), ... })
    }

    pub fn edit_message(&mut self, idx: usize, new_body: String) -> app::Result<()> {
        let mut msg = self.messages.get(idx)?.ok_or(NotFound)?;
        msg.body = new_body;
        self.messages.update(idx, msg)?;   // runtime rejects if not owner
        Ok(())
    }

    pub fn get_recent_messages(&self, count: usize) -> app::Result<Vec<Message>> {
        // Natural chronological order; O(count) reads.
        let len = self.messages.len()?;
        let start = len.saturating_sub(count);
        (start..len)
            .filter_map(|i| self.messages.get(i).ok().flatten())
            .filter(|m| !m.is_tombstone())
            .collect()
    }
}
```

No in-WASM `caller == sender` checks. No per-user k-way merge. Auth enforced by the storage layer.

## Design questions

1. **Tombstone vs physical `remove` for `AuthoredVector`.** Physical `remove(idx)` shifts subsequent indices, which makes concurrent pushes a merge nightmare (two peers push "at the same position," indices diverge, owner stamps have to follow). Recommendation: no physical remove — only `tombstone(idx)` which writes a placeholder value. Readers filter.

2. **Concurrent pushes from two peers.** Standard CRDT Vector gives each push a unique position-id; merge converges. Owner stamp is per-entry, so each push's stamp is its own pusher. Should work with existing Vector CRDT semantics.

3. **`owner_of` API.** Exposing the owner pubkey is useful for app-level UI ("edited by X"). Should it be part of the primary API or a separate read?

4. **Insert-vs-update distinction for `AuthoredMap`.** If `insert(k, v)` is called on an existing K, should it fail (explicit separation from `update`) or silently upsert under the caller's ownership (which would allow anyone to reassign ownership of existing keys — unacceptable)? Recommendation: `insert` fails on existing K; ownership transfer requires explicit `remove` + re-insert by the new owner.

5. **`owner` field serialization overhead.** Each entry carries a 32-byte owner pubkey. Unavoidable for per-entry auth; negligible at most scales, worth mentioning for very-many-small-entries workloads.

## Describe alternatives you've considered

**Composition of existing primitives.** Walked through `UnorderedMap<K, UserStorage<V>>`, `Vector<UserStorage<V>>`, `UnorderedMap<K, SharedStorage<V>>` above. All leave auth gaps at the enclosing container level. Rejected.

**In-WASM signer checks over `Public` storage.** Doesn't survive tampered WASM — peers apply `Public` deltas at the storage layer without re-executing WASM. Rejected.

**Extend `SharedStorage` with per-entry mode.** `SharedStorage` (#2197) is collection-level with mutable writer sets. `AuthoredMap`/`AuthoredVector` are per-entry with immutable single owner. Conceptually related but materially different shapes; unifying them would force single-owner entries to pay `BTreeSet<PublicKey>` overhead and rotation complexity they don't need. Keep them as distinct primitives.

**`AuthoredMap` with multi-author per entry.** Rejected because it reduces to `UnorderedMap<K, SharedStorage<V>>` composition with the "override the key" asymmetry, OR requires effectively building both primitives into one. Keep AuthoredMap single-author; use SharedStorage for co-owned cases.

## Relationship to #2197 (`SharedStorage`)

Complementary, not overlapping:

| | `SharedStorage<T>` | `AuthoredMap<K, V>` / `AuthoredVector<V>` |
|---|---|---|
| Keyspace | one slot with T | shared K→V or ordered V |
| Writer set | collection-level | per-entry, size 1 |
| Mutable writers | yes (rotate) | no (fixed at insert) |
| Use case | co-owned document, admin pool | chat, forum, UGC map |

Both use runtime signing + merge-time peer verification against a stored authorization field. `SharedStorage` stores one writer set for the whole collection; `AuthoredMap` stores one owner per entry.

## Out of scope

- Multi-author per entry in a shared keyspace (use `SharedStorage` composition or a future `CoAuthoredMap` primitive).
- Ownership transfer (explicit `remove` + re-insert covers it; in-place transfer is a different primitive).
- Threshold signing.
- Cross-context ownership (owner in another context authorizes writes here).

## Additional context

Followup to #2197. Conversation analyzing authorization primitives for `calimero-studio/foundation-app`:

- `UserStorage` solves per-user disjoint data (covered today).
- `SharedStorage` (#2197) solves collection-level co-ownership (proposed).
- `AuthoredMap`/`AuthoredVector` solve per-entry ownership in a shared keyspace (this issue).

For the foundation-app chat use case specifically, `AuthoredVector<Message>` is the ideal fit — gives `O(limit)` chronologically-ordered reads with per-sender edit/delete authorization enforced cryptographically at merge time. No hybrid indices, no k-way merge, no in-WASM honor-system checks.

xilosada:
## Summary

Adds two new storage primitives that close a concrete gap in the authorization model: **per-entry ownership in a shared keyspace**.

- `AuthoredMap<K, V>` — any context member can insert a new key; only the inserter can update or remove their own entry.
- `AuthoredVector<V>` — any member can push at the end; only the author can update or tombstone their slot.

Both stamp each entry with `StorageType::User { owner = executor_id }`. Non-owner writes are rejected locally (short-circuit) and at merge time via the existing signature-verification path in `Interface::apply_action` — no new verification code.

Closes #2201.

### Why this can't be composed from existing primitives

`UnorderedMap<K, UserStorage<V>>` and `Vector<UserStorage<V>>` leave the outer container `Public`, so a tampered peer can overwrite `K` or the whole slot wholesale. The authorization stamp has to live on the container entry itself — that's what `AuthoredMap` / `AuthoredVector` provide.

### API

```rust
// AuthoredMap
fn insert(&mut self, k: K, v: V) -> Result<(), StoreError>;   // errors if K exists
fn update(&mut self, k: &K, v: V) -> Result<(), StoreError>;  // owner-only
fn remove(&mut self, k: &K) -> Result<Option<V>, StoreError>; // owner-only
fn get / contains / owner_of / entries / len

// AuthoredVector
fn push(&mut self, v: V) -> Result<usize, StoreError>;        // returns index
fn update(&mut self, idx: usize, v: V) -> Result<(), StoreError>;    // owner-only
fn tombstone(&mut self, idx: usize) -> Result<(), StoreError>;       // owner-only, V: Default
fn get / owner_of / iter / len
```

No physical `remove` on `AuthoredVector` — `tombstone` writes `V::default()` into the slot so concurrent-push merge semantics aren't disturbed by shifting indices.

### Design decisions (per the issue's open questions)

- `insert` on existing `K` fails rather than silently upserting — prevents ownership reassignment via insert.
- Local non-owner `update`/`remove` return `ActionNotAllowed` in addition to merge-time rejection (defense in depth, surfaces bugs in-process).
- Owner is immutable per entry — ownership transfer requires explicit `remove` + re-insert.
- Reused `CrdtType::UserStorage` at the collection level; no new `CrdtType` variant needed since per-entry auth is dispatched by `StorageType`, not by the CRDT kind.

### Implementation notes

- Mirrored the `UserStorage` pattern: wrap the base collection, hold an `Element` for top-level metadata, delegate `Data`/`Mergeable`/`CrdtMeta`.
- `Vector` gains two `pub(crate)` helpers — `push_with_storage_type` and `entry_id_at` — mirroring the existing `UnorderedMap::insert_with_storage_type`.

## Test plan

Built via TDD — 24 new tests, all passing; 387 total storage lib tests (up from 363), no regressions.

- [x] AuthoredMap unit tests (11): insert stamps owner, insert-rejects-existing-K, update/remove by owner, update/remove by non-owner rejected, shared keyspace across two users, owner_of, entries.
- [x] AuthoredVector unit tests (9): push stamps owner, concurrent pushes preserve per-entry ownership, update/tombstone by owner, update/tombstone by non-owner rejected, iter order, out-of-bounds.
- [x] Merge-time tests (4, in `tests/authored_primitives.rs`): insert via the primitive, then feed hand-crafted `Action::Update` / `Action::DeleteRef` into `Interface::apply_action` — assert `InvalidSignature` for forged-owner updates and non-owner deletes on both primitives.
- [x] `cargo test -p calimero-storage --lib` — 387 passed, 0 failed.
- [x] `cargo build --workspace` — clean.
- [ ] End-to-end foundation-app chat use case (motivating example in the issue) — out of scope here; unblocked now.

---

> [!NOTE]
> **Medium Risk**
> Touches authorization-adjacent storage behavior by introducing new primitives that rely on `StorageType::User` metadata and merge-time signature enforcement; incorrect stamping or helper logic could allow unintended writes despite extensive tests.
> 
> **Overview**
> Adds new storage collections `AuthoredMap<K,V>` and `AuthoredVector<V>` that stamp each entry/slot with `StorageType::User { owner }` (from `env::executor_id()`), allowing any member to add new entries while restricting updates/deletes (or vector tombstoning) to the original author.
> 
> Updates `Vector` with internal helpers (`push_with_storage_type`, `entry_id_at`) to support per-entry metadata stamping and owner lookups, wires the new primitives into `collections` exports, and adds unit + merge-time tests that assert forged/non-owner `Update`/`DeleteRef` actions are rejected by `Interface::apply_action` signature verification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0bcf6543ea957d089ccf4b3b44f12893df206789. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xilosada:
Addressed the review comments in 0bcf6543:

**🔴 Critical**
- `Missing # Errors docs` on both primitives (`authored_map.rs`, `authored_vector.rs`) — fixed in the earlier commit 970498ee.
- `Mergeable::merge bypasses ownership` (both primitives, duplicated as #3/#4/#8/#9) — `AuthoredMap::merge` and `AuthoredVector::merge` are now documented no-ops. Delegating to `UnorderedMap::merge`/`Vector::merge` was unsafe because those paths re-insert remote-only entries with `StorageType::Public`, stripping ownership. Per-entry merge already runs through `Interface::apply_action` on the signed-delta path, and the `CrdtType::UserStorage` container dispatch in `merge.rs` takes the byte-level path — so the structural merge is not needed at runtime, and a no-op is strictly safer than the previous delegation.

**🟡 Medium**
- `update` doc mismatch (OOB variant was `ActionNotAllowed` not `NotFound`) — fixed: OOB now returns `InvalidData` with a clear message; `ActionNotAllowed` is reserved for auth failures on valid addresses. Doc summary and `# Errors` section are aligned. The existing `update_out_of_bounds_errors` test still passes (it only substring-matches `"out of bounds"`).
- `require_owner` using `ActionNotAllowed` for OOB (plus the self-acknowledged "Fabricate a zero-id" comment) — fixed alongside the above. Chose `InvalidData` over a new error variant to keep the blast radius small.
- `AuthoredVector::collections() returns empty` asymmetry — kept `BTreeMap::new()` but added a comment explaining that `Vector<V>` doesn't implement `Data` (its entries are values, not nested CRDTs), so there's no delegate hook — unlike `AuthoredMap` which forwards to `UnorderedMap::collections()`.
- `Nonce timing flake` in tests — nonces for forged actions are now `env::time_now() + FORGED_NONCE_OFFSET_NS` (1s). Eliminates the theoretical `NonceReplay` race on fast clocks.

**💡 Nits**
- `push index fragility` (`len.saturating_sub(1)`) — `Vector::push_with_storage_type` now returns the index directly, with a `debug_assert!(len >= 1)` inside (safe since the subtract happens after a successful push). `AuthoredVector::push` just propagates that return value.

All 24 new storage tests still pass; `cargo clippy -- -A warnings` is clean on the full workspace.

chefsale (approved):

Issue #2215, opened 2026-04-22
storage: merge-apply tail latency lives in WASM-side CRDT algorithms (not host callbacks)


chefsale:
## Context

Replaces #2208 with corrected framing after PR #2214's sub-timer measurement.

## What the data shows

PR #2214 instrumented per-callback cost in \`logic/imports.rs\`. Results from e2e artifacts (n=367 WASM runs):

| Metric | Value |
|--------|-------|
| Avg per-callback cost | ~3.8 µs |
| Of which ouroboros rebuild | **86 ns** |
| Of which host function body | ~3.7 µs |
| Callbacks on worst-outlier run | 11,524 |
| Total host-side cost on worst outlier | ~44 ms |
| \`wasm_ms\` on worst outlier | **908 ms** |
| Remaining (pure WASM-side) | **~864 ms** |

So **~95% of the tail is in WASM-side algorithmic code**, not in host callbacks or wasmer boundary overhead. The host callback hypothesis (tracked in the now-closed #2208) turned out to not be the dominant cost.

## Where the WASM-side time is spent

From trace analysis of \`crates/storage\`:

1. **\`UnorderedMap::merge\`** (\`crates/storage/src/collections/crdt_impls.rs:~272\`) iterates every entry of the incoming replica, calling \`get(key)\` per entry. Scales O(N) in app state size. For a 50-entry map merged 3 times (one per action), that's 150+ entry reads plus N-per-get storage operations.

2. **\`Compare\` action processing** (\`crates/storage/src/interface.rs:~715\`, \`generate_comparison_data\`) walks the full tree per entity: \`get_hashes_for\` + \`get_metadata\` + \`get_ancestors_of\` + \`get_collection_names_for\` + \`get_children_of\` per collection. For a root with 50 children that's ~250 reads per Compare.

3. **\`Interface::apply_action\`** + **\`recalculate_ancestor_hashes_for\`** (\`crates/storage/src/interface.rs:109\`, \`crates/storage/src/index.rs:303\`) — ancestor walks per action. PR #2211 partially addressed the double-read issue but the ancestor walk still runs per-action.

All three are in WASM (compiled into the app module), not host. They're algorithmic, not framework cost.

## Fix candidates (rough order of effort : impact)

1. **Lazy \`Compare\`** — skip the tree walk when upstream hashes already indicate no divergence. Likely the biggest single win if Compare dominates the 864ms.
[...]

Issue #2208, opened 2026-04-22
runtime/storage: merge-apply making thousands of WASM→host storage callbacks per action


chefsale:
## Context

Surfaced by PR #2206's sub-timer instrumentation and log analysis.

A single 3-action merge-apply in the e2e-kv-store test (\`__calimero_sync_next\` WASM invocation, \`is_merge=true\`) logged in 139ms:

| host callback | count |
|---|---|
| \`storage_read\` | **2,167** |
| \`storage_write\` | **727** |
| other host ops (memory, utility, etc.) | ~7,600 |
| **total WASM→host crossings** | **10,523** |

That's ~720 storage reads *per action* in a three-action delta for a simple KV app. 100x more than the action itself should logically need.

## Why it matters

Each WASM→host crossing costs on the order of microseconds regardless of what the callback does (wasmer boundary + ouroboros \`VMHostFunctions\` rebuild + the actual host work). With 10k+ crossings per apply, framework overhead alone can reach 20-50ms per WASM run — and that's before counting debug log dispatch on top.

This is also the structural reason PR #2196's original 918ms apply-latency outlier existed. PR #2203's criterion benches already ruled out the three originally-guessed suspects (merkle rehash, \`merge_root_state\` framework, RocksDB read latency). PR #2206's sub-timers pointed inside \`function.call\`. This is what's inside.

## Root-cause hypothesis

The merge path in \`crates/storage/src/collections/root.rs\` (\`Root::sync\`) iterates actions. For each action, \`apply_action\` traverses the entity tree — under a merge scenario it additionally:
- Reads existing root state.
- Walks children of root to recompute merkle hashes.
- Calls the app's registered \`Mergeable::merge\`, which for an app using nested CRDT collections (like \`UnorderedMap\`) internally iterates entries — each iteration being backed by more storage reads via \`Collection::entries\`.

For an \`UnorderedMap\`-heavy state, merging N entries is \`O(N)\` storage reads minimum. For our observed ratio (~720 reads/action), either the state is larger than expected or the merge walk is quadratic.

## Fix directions

In rough order of likely impact vs. effort:

[...]

Issue #2199, opened 2026-04-22
storage: WASM merge-apply cost scales with storage size (full-merkle recompute)


chefsale:
## Context

Surfaced by the instrumentation in PR #2196 during the #2186 measurement work. Not a regression — long-standing behaviour in the merge-apply path that's only now been quantified.

## The signal

PR #2196's e2e runs captured two lock-hold warnings (≥500ms), both on single-delta applies with zero cascade activity:

\`\`\`
WARN delta_store: DAG write lock held for long tail (#2186) site=\"add_delta_internal\" hold_ms=918.54 plans_count=0 cascaded_count=0
WARN delta_store: DAG write lock held for long tail (#2186) site=\"add_delta_internal\" hold_ms=971.98 plans_count=0 cascaded_count=0
\`\`\`

Log context shows hold time was entirely WASM execution:

\`\`\`
DEBUG WASM sync completed execution ... is_merge=true wasm_ms=\"918.38\"
INFO  DELTA_APPLY_TIMING ... action_count=3 ... was_merge=true wasm_ms=\"918.38\" total_ms=\"918.50\"
WARN  DAG write lock held for long tail (#2186) ... hold_ms=918.54
\`\`\`

A 3-action delta taking ~918ms of WASM is the surprise. The structural cause comes out cleanly when you trace what a merge-apply does that a sequential apply doesn't:

## Root cause chain

Three extras on the merge path (none present on the sequential apply path):

1. **Extra storage read** — \`crates/storage/src/interface.rs:925\` reads the existing root entity back from RocksDB for merge input.

2. **Nested WASM callback** — \`interface.rs:1053\` invokes \`merge_root_state\`, which is itself a WASM call *inside* the outer \`__calimero_sync_next\` invocation. The 918ms \`wasm_ms\` is the sum of all WASM crossings during this apply, not one.

3. **Full merkle recomputation over all root children** — \`crates/storage/src/index.rs:439-481\` → \`calculate_full_hash_for_children\`:
   \`\`\`rust
   for child in children_vec {
       hasher.update(child.merkle_hash());
   }
   \`\`\`
   O(children-of-root) SHA256 operations per apply. No incremental maintenance.

## Why it compounds under concurrent writers

- \`group-3node\` e2e test: three nodes writing concurrently.
- Every incoming delta from the other two authors hits \`is_merge=true\` on each node.
- Every merge apply then re-hashes all children of root.
[...]

PR #2217, opened 2026-04-22
fix(ci): harvest profiling data from host bind mount

chefsale:
## Problem

The profiling-data artifacts uploaded by the fuzzy-load-test workflow contain **only jemalloc heap dumps from `merod init`** — 1073 tiny (5–12 KB) files per run, all from the init phase. No CPU perf.data, no runtime heap, no perf.map, no data from the 45-minute `merod run` phase that we actually care about for #2199 / #2215.

Concretely, in run [24787476234](https://github.com/calimero-network/core/actions/runs/24787476234), the collection step logs:

```
Collecting from container: fuzzy-kv-node-4-init
  Could not stop perf (container may have stopped)
  WARNING: Container not running - collecting existing data only
... (same for nodes 3, 2, 1)
```

All four containers listed are `-init`.

## Root cause

Merobox creates each node as two containers (manager.py:327):
- `fuzzy-*-node-N-init` — runs `merod init`, exits, fails to auto-remove (so it lingers in Exited state)
- `fuzzy-*-node-N` — runs `merod run`, and is explicitly **removed during merobox's graceful shutdown** before our `collect-from-containers.sh` step executes

By the time `docker ps -a --filter "label=calimero.node=true"` runs, only the dead init containers are still present. Their /profiling/data rootfs is intact, so `docker cp` succeeds — but it contains only init-phase data.

## Fix

`entrypoint-profiling.sh` now mirrors `/profiling/data` (and reports) into `$CALIMERO_HOME/profiling-dump` in the shutdown path. Merobox bind-mounts `/app/data` to `workflows/fuzzy-tests/<test>/data/<node>/` on the host (manager.py:474), so the dump **survives container removal**.

A new workflow step after the existing collector harvests those dumps into the `profiling-data-*` artifact tree.

Both preserve paths fire:
- `cleanup()` trap on SIGTERM/SIGINT (merobox graceful shutdown)
- Mainline `wait $MAIN_PID; stop_profiling; preserve_to_host_mount` on normal exit

Init containers are untouched: they hit the `exec "$@"` branch (because `SHOULD_PROFILE_WITH_PERF=false`), which replaces the shell and drops the trap.

## What the artifact will look like after this merges

Today: `profiling-data-kv-store-*/fuzzy-kv-node-N-init/jemalloc.1.*.heap` × 1073 files.

After: `profiling-data-kv-store-*/fuzzy-kv-node-N/` containing `perf-fuzzy-kv-node-N.data` (100–500 MB each, 45min × 99Hz), `perf-fuzzy-kv-node-N-<pid>.map` (WASM symbolization), `jemalloc.*.heap` (runtime heap, not init), and `memory-stats-*.log`.

Artifact size will grow from ~1.7MB to probably 1–2GB per run. Still well under GitHub's per-artifact limit.

## Test plan

- [ ] Merge → next `release.yml` rebuilds `ghcr.io/calimero-network/merod:edge-profiling` with the new entrypoint
- [ ] Manually trigger `fuzzy-load-test` workflow (or wait for next master push)
- [ ] Download the new `profiling-data-kv-store-*` artifact, verify it contains `perf-fuzzy-kv-node-*.data` files with real sample counts (`perf report -i file.data --stdio | head`)
- [ ] Run `generate-flamegraph.sh` locally against one of the perf.data files, confirm a real CPU flamegraph with WASM frames visible (needs the matching perf.map)

## Related

- #2199 — umbrella merge-apply cost. Blocked on getting real CPU profiles.
- #2215 — successor to #2208 with corrected framing: WASM-side CRDT algorithm cost. Needs this data.

---

> [!NOTE]
> **Medium Risk**
> CI profiling collection logic changes and introduces large new artifacts; failures could hide profiling output or increase workflow time/storage usage.
> 
> **Overview**
> Fixes missing runtime profiling artifacts in `fuzzy-load-test` by **mirroring `/profiling/{data,reports}` into `$CALIMERO_HOME/profiling-dump` on shutdown** (`entrypoint-profiling.sh`) and then **harvesting those dumps from the host bind-mount tree** into `profiling-data/*` via the new `harvest-host-profiling.sh` step.
> 
> Also increases the `perf` graceful-stop wait from 5s to 15s to reduce truncation risk for long-running `perf.data` files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a01eb90d4829db0481722902b426a45d9cc78604. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/fuzzy-load-test.yml
+++ b/.github/workflows/fuzzy-load-test.yml
@@ -251,36 +251,10 @@ jobs:
       - name: Collect profiling data from host-side bind mounts
         if: always() && steps.image.outputs.profiling == 'true'
         run: |
-          # Runtime containers (fuzzy-kv-node-N) are removed during merobox's
-          # graceful shutdown before the Docker-based collector runs above —
-          # so their /profiling/data vanishes with them. entrypoint-profiling.sh
-          # mirrors that data into /app/data/profiling-dump (a host bind mount)
-          # on shutdown; harvest it here regardless of container state.
-          SRC_ROOT="workflows/fuzzy-tests/kv-store/data"
-          DEST_ROOT="profiling-data/kv-store"
-          if [ ! -d "$SRC_ROOT" ]; then
-            echo "No data dir at $SRC_ROOT — nothing to harvest"
-            exit 0
-          fi
-          found=0
-          for node_dir in "$SRC_ROOT"/*/; do
-            [ -d "$node_dir" ] || continue
-            node_name=$(basename "$node_dir")
-            dump="$node_dir/profiling-dump"
-            if [ -d "$dump" ]; then
-              dest="$DEST_ROOT/$node_name"
-              mkdir -p "$dest"
-              cp -r "$dump/." "$dest/" 2>/dev/null || true
-              size=$(du -sh "$dest" 2>/dev/null | awk '{print $1}')
-              perf_count=$(find "$dest" -maxdepth 2 -name 'perf-*.data' 2>/dev/null | wc -l | tr -d ' ')
-              heap_count=$(find "$dest" -maxdepth 2 -name 'jemalloc.*.heap' 2>/dev/null | wc -l | tr -d ' ')
-              echo "  $node_name: ${size:-?} (perf.data=$perf_count, heap=$heap_count)"
-              found=$((found + 1))
-            else
-              echo "  $node_name: no profiling-dump (runtime container may not have used profiling entrypoint)"
-            fi
-          done
-          echo "Harvested profiling dumps from $found nodes"
+          chmod +x scripts/profiling/harvest-host-profiling.sh
+          ./scripts/profiling/harvest-host-profiling.sh \
+            "workflows/fuzzy-tests/kv-store/data" \
+            "profiling-data/kv-store"
 
       - name: Set test result
         id: set-result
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed all three bot review comments in 2091ca7d:

**@meroreviewer (duplication + silent cp errors)**: Extracted the harvest block into `scripts/profiling/harvest-host-profiling.sh`. Both workflow jobs now shell out to it — single source of truth, ~58 lines of inline YAML gone. The script captures cp stderr to a tempfile and echoes a `WARNING` line on failure instead of discarding it.

**@cursor (`mkdir` under `set -e`)**: Good catch. `preserve_to_host_mount` now uses `if ! mkdir -p … ; then WARN; fi` everywhere, explicitly `return 0`, and captures stderr for diagnostics. A copy failure no longer masks merod's real exit code.

Local smoke test of the extracted script:
```
$ ./scripts/profiling/harvest-host-profiling.sh /tmp/test-harvest/src /tmp/test-harvest/dest
  node-1: 8.0K (perf.data=1, heap=1)
  node-2: no profiling-dump (runtime container may not have used profiling entrypoint)
Harvested profiling dumps from 1 nodes
```

Not addressing the doc-review suggestion: `AGENTS.md` / `CONTRIBUTING.md` don't cover CI artifact collection, and this PR doesn't change public-facing or contributor workflow — just plumbs an already-documented profiling pipeline end-to-end.

chefsale:
Second pass addresses the new bot comments in 42519000:

- **Race on perf flush (meroreviewer)** — Bumped `stop_profiling`'s SIGINT→SIGKILL timeout from 5s to 15s. A 45-min 99 Hz perf.data is big enough that the final flush on SIGINT can outrun a 5-second window; truncation here would defeat the whole point of this PR.
- **Symlink race on `/tmp/preserve.err` (meroreviewer)** — Switched to `mktemp -t preserve.err.XXXXXX`. Low-risk inside a single-root container (CWE-377 needs a co-tenant) but free to fix and now consistent with `harvest-host-profiling.sh`.
- **Unguarded `mkdir` in harvest script (meroreviewer)** — The `mkdir -p "$dest"` now has explicit error handling with stderr capture and a `continue` on failure, so a permissions/disk issue yields a clear per-node ERROR line instead of a confusing cp failure.

Skipping the "missing trailing newline" note — the file already ends with `0x0a` (verified with `xxd` and `wc -l` reports 51 lines). That one's a false positive.

chefsale:
Addressed the cursor review in a01eb90d.

You're right that the earlier `${CALIMERO_HOME:-/app/data}` + "Merobox mounts `/app/data`" framing was misleading. The real invariant:

- **Merobox (CI)**: sets `CALIMERO_HOME=/app/data` via env and bind-mounts the host dir to `/app/data` (manager.py). This is the path this PR cares about.
- **Standalone `docker run`**: `prebuilt.profiling.Dockerfile` sets `ENV CALIMERO_HOME=/data` and `VOLUME /data`, so the path lives on a Docker-managed volume even without `-v`.

Both deployments already set `CALIMERO_HOME`, so the fallback was never load-bearing — and masked the fact that an unset var would silently write into a non-persistent dir. Removed the fallback, warn-and-skip if unset, and rewrote the doc block in both files to describe both deployments explicitly.

Issue #1993, opened 2026-02-14
perf(storage): O(1) merge function lookup via type hints


xilosada:
## Summary

The merge registry currently uses O(n) brute-force lookup - it tries each registered merge function until one succeeds. This should be optimized to O(1) by storing type hints with root entity data.

## Current Behavior

```rust
// crates/storage/src/merge/registry.rs
for (_type_id, merge_fn) in registry.iter() {
    if let Ok(merged) = merge_fn(existing, incoming, existing_ts, incoming_ts) {
        return MergeRegistryResult::Success(merged);
    }
}
```

## Proposed Solution

1. Add `type_id: Option<TypeId>` to root entity metadata
2. Store the type ID when saving root state via `register_crdt_merge`
3. Use it for O(1) dispatch in `try_merge_registered`

## Impact

- **Current**: O(n) per merge where n = number of registered types
- **After**: O(1) per merge

For most apps with 1-2 types, current performance is acceptable. This becomes important for apps with many registered CRDT types.

## Related

- Fits under **Phase 5: Optimization** (see POC-IMPLEMENTATION-NOTES.md)
- Identified during PR #1988 review

## Files to Modify

- `crates/storage/src/merge/registry.rs` - Add type-based dispatch
- `crates/storage/src/collections/root.rs` - Store type hint on save
- Entity metadata schema (if type hint needs persistence)

chefsale:
Closing as obsolete after re-analysis against current `main`.

**The performance motivation no longer holds.** The merge registry is populated exclusively by the `#[app::state]` macro, which emits exactly one `register_crdt_merge::<AppState>()` per WASM module (`crates/sdk/macros/src/state.rs:458-459`). A WASM app has a single root state type, so `MERGE_REGISTRY` holds exactly one entry and the "O(n) brute-force" loop in `try_merge_registered` runs a single iteration in production. `n > 1` only happens in the test suite, where many types share one registry. There is no real-world perf win to capture.

**The architecture moved out from under the proposal.** Since this issue was filed (from PR #1988 review), the registry became WASM-side only — the host binary no longer links it (root-cause fix for #2469). Host-side root merges now route through the macro-generated `__calimero_merge_root_state` export, not the registry loop, so the registry is no longer the dispatch boundary the issue assumed.

**The proposed plan is also partly unworkable as written.** Persisting `type_id: Option<TypeId>` in root entity metadata won't work — `TypeId` is process-local and not stable across builds/modules, so it can't be stored and reloaded.

[...]

PR #2219, opened 2026-04-22
fix(storage): make merge registry thread-local under #[cfg(test)]

chefsale:
## Problem

Intermittent CI failures in `calimero-storage`'s test suite. Symptom: three tests panic on assertions that should always pass:

- `interface::tests::user_storage_signature_verification::user_update_with_valid_signature_succeeds`
- `interface::tests::user_storage_replay_protection::sequential_updates_with_increasing_nonces_succeed`
- `interface::tests::storage_type_edge_cases::user_delete_replay_protection`

Observed in runs [24799493814](https://github.com/calimero-network/core/actions/runs/24799493814/job/72578008668) and [24799790060](https://github.com/calimero-network/core/actions/runs/24799790060/job/72579053356). Not reproducible locally.

## Root cause

`MERGE_REGISTRY` is a process-global `RwLock<HashMap<TypeId, MergeFn>>`. Three tests in `merge::registry::tests` and `tests::merge_integration` call `clear_merge_registry()` to set up clean-slate assertions, and carry `#[serial]`. The attribute from `serial_test` serialises those three against *each other* — it does **not** prevent unrelated non-serial tests from running concurrently with them.

The racy interleaving:

```
t=0  non-serial test:  register_test_merge_functions()   // registry populated
t=1  serial test:      clear_merge_registry()            // registry empty
t=2  non-serial test:  apply_action(...)                 // lookup -> None
                         └─ assertion fails
t=3  serial test:      register_crdt_merge::<TestState>() // too late
```

Everything observed — failure signature, intermittence, no reproduction under low load — matches.

## Fix

Production keeps the global `RwLock`. Under `#[cfg(test)]` the backing swaps to `thread_local!`, so each test-runner worker thread gets its own private registry. One thread clearing can't affect dispatch on another — the race is eliminated by construction.

All call sites stay unchanged; access is routed through small `with_registry` / `with_registry_mut` helpers that are cfg-gated to pick the right backend.

## Why not tag more tests `#[serial]`

Would require finding and annotating every test that touches the registry directly or transitively (~20+, growing). Easy to forget. Thread-local makes the isolation structural.

## Test plan

- [x] `cargo test -p calimero-storage --lib` — 387/387 pass, five runs in a row
- [x] `cargo check --workspace` clean
- [ ] CI green across a few merges (to confirm the flake is gone, not just quieter)

---

> [!NOTE]
> **Medium Risk**
> Touches core merge-registry dispatch/registration paths (now routed through new helpers) and changes test-vs-prod storage semantics, which could impact merge behavior if the cfg-gated implementations diverge. Added integration/concurrency tests reduce risk by exercising the production `RwLock` backend.
> 
> **Overview**
> Fixes flaky `calimero-storage` tests by making the merge-function registry **thread-local under `#[cfg(test)]`**, preventing parallel test threads from clearing or overwriting each other’s registrations.
> 
> Refactors registry access behind `with_registry`/`with_registry_mut` (prod: global `RwLock`, tests: thread-local `RefCell` with clearer re-entrancy failures), and adds new integration tests (`merge_registry_integration.rs` and a concurrent stress test) to cover the production `RwLock` register+dispatch path that unit tests can’t reach.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ebc1dfd0808d27b552ec7a07e9d240b29421cc73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/storage/src/merge/registry.rs
+++ b/crates/storage/src/merge/registry.rs
@@ -54,10 +54,27 @@ static MERGE_REGISTRY: LazyLock<RwLock<HashMap<TypeId, MergeFn>>> =
 // against each other — unrelated non-serial tests still ran in
 // parallel with them. Thread-local storage makes each test's
 // registry state private to its own thread, so the race can't occur.
+//
+// Trade-off: unit tests can no longer observe cross-thread visibility
+// of registrations (a property the production RwLock does provide).
+// We consider this acceptable — the cross-thread-share story is
+// delegated to `std::sync::RwLock`, which we trust, and in practice
+// apps register their types once at startup before any async workers
+// are spawned. If cross-thread dispatch ever becomes part of the
+// behaviour-under-test (not just implementation detail), that belongs
+// in an integration test compiled without `#[cfg(test)]` rather than
+// here.
 #[cfg(test)]
 thread_local! {
     static MERGE_REGISTRY: RefCell<HashMap<TypeId, MergeFn>> = RefCell::new(HashMap::new());
 }
+
+// Both backends assume merge functions don't call back into the registry
+// while one is being dispatched. A reentrant call would deadlock under
+// RwLock (write-during-read) or panic under RefCell (already borrowed) —
+// either way a bug. The merge closure built in `register_crdt_merge` only
+// calls borsh and the type's own `Mergeable::merge`; adding registry
+// access to it would break this invariant.
 
 /// Run `f` with mutable access to the registry.
 #[cfg(not(test))]
@@ -72,9 +89,21 @@ fn with_registry_mut<R>(f: impl FnOnce(&mut HashMap<TypeId, MergeFn>) -> R) -> R
     f(&mut registry)
 }
 
+/// Test-only variant: uses thread-local RefCell instead of global RwLock.
+/// try_borrow_mut to surface re-entrant access as a clear panic message
+/// rather than the less-useful default `BorrowMutError` debug print.
 #[cfg(test)]
 fn with_registry_mut<R>(f: impl FnOnce(&mut HashMap<TypeId, MergeFn>) -> R) -> R {
-    MERGE_REGISTRY.with(|r| f(&mut r.borrow_mut()))
+    MERGE_REGISTRY.with(|r| {
+        let mut borrowed = r.try_borrow_mut().unwrap_or_else(|e| {
+            panic!(
+                "MERGE_REGISTRY RefCell already borrowed ({e}). Merge \
+                 functions must not call try_merge_registered / \
+                 register_crdt_merge / clear_merge_registry (see module docs)."
+            )
+        });
+        f(&mut borrowed)
+    })
 }
 
 /// Run `f` with read-only access to the registry.
@@ -90,9 +119,19 @@ fn with_registry<R>(f: impl FnOnce(&HashMap<TypeId, MergeFn>) -> R) -> R {
     f(&registry)
 }
 
+/// Test-only variant: uses thread-local RefCell instead of global RwLock.
 #[cfg(test)]
 fn with_registry<R>(f: impl FnOnce(&HashMap<TypeId, MergeFn>) -> R) -> R {
-    MERGE_REGISTRY.with(|r| f(&r.borrow()))
+    MERGE_REGISTRY.with(|r| {
+        let borrowed = r.try_borrow().unwrap_or_else(|e| {
+            panic!(
+                "MERGE_REGISTRY RefCell already mutably borrowed ({e}). \
+                 Merge functions must not call try_merge_registered / \
+                 register_crdt_merge / clear_merge_registry (see module docs)."
+            )
+        });
+        f(&borrowed)
+    })
 }
 
 /// Register a CRDT merge function for a type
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed all three in c3a197bb:

- **RefCell re-entrancy (MEDIUM)** — both test helpers now use `try_borrow` / `try_borrow_mut` and panic with a diagnostic message naming the invariant. Added a block comment above the helpers explaining both failure modes (deadlock under RwLock, panic under RefCell) and why a re-entrant merge function would be a bug either way. The merge closure built in `register_crdt_merge` only calls borsh and `Mergeable::merge` — the invariant is easy to uphold.
- **Cross-thread visibility divergence (LOW)** — added a trade-off note to the `thread_local!` block. Cross-thread sharing is delegated to `std::sync::RwLock` (which we trust), and in practice apps register once at startup before any async workers are spawned. If cross-thread dispatch ever becomes behaviour-under-test, it belongs in an integration test compiled *without* `#[cfg(test)]` — not a unit test that would just be exercising std's locking primitives.
- **Missing doc comments on test variants (STYLE)** — added brief comments on both.

Storage tests still 387/387, five clean runs.

---

Separately, re: the "profiling didnt run" observation — the fuzzy-load-test didn't auto-fire on the #2217 or #2218 merges because its path filter is `crates/** | apps/** | workflows/fuzzy-tests/** | .github/workflows/fuzzy-load-test.yml` — none of which matched. I manually dispatched [run 24801298877](https://github.com/calimero-network/core/actions/runs/24801298877) which is running now with the rebuilt `edge-profiling` image; artifacts at the end will tell us whether #2217's host-mount harvest actually produces real `perf.data` under the runtime container.

Longer-term fix for auto-trigger would be a `workflow_run` dependency or expanded path filter on fuzzy-load-test — happy to open a follow-up once we've confirmed the current run produces the data we expect.

chefsale:
Addressed in 019b3065.

Added `crates/storage/tests/merge_registry_integration.rs` — an integration test that compiles WITHOUT `#[cfg(test)]`, so it links against the production `LazyLock<RwLock<HashMap<TypeId, MergeFn>>>` backend that the thread-local `RefCell` unit-test backend bypasses.

It covers the happy path: `register_crdt_merge::<T>()` then `try_merge_registered(...)` returns `Success` with the expected merged bytes. That's the plumbing we care about; the per-type `Mergeable::merge` logic has its own coverage in the unit tests.

Verified locally:
```
$ cargo test -p calimero-storage --test merge_registry_integration
running 1 test
test register_and_dispatch_against_production_registry ... ok
```

Not covered by design: the `abort()`-on-poison branches in `with_registry` / `with_registry_mut`. Exercising them requires panicking inside the lock's critical section, and the response (`std::process::abort()`) would take down the test runner with it. Those branches are 3 lines each and reviewable by eye.

chefsale:
Rebased onto latest master (dropped the merge commit) and addressed the review:

**#1 — misleading cfg(test) claim in integration test doc.** Fixed. The binary itself *is* built with `cfg(test)` (that's how `#[test]` works); the important distinction is that the *library* it links against isn't. Doc comment now says that explicitly.

**#2 — non-deterministic dispatch in `try_merge_registered`.** Pre-existing, tracked by the `TODO` on line ~224 (issue #1993). Reviewer explicitly noted no action needed in this PR; skipped.

**#3 — test/prod paths fully diverge.** Added a pointer-comment in `src/merge/registry.rs` right above the cfg-gated helpers, telling future maintainers that `tests/merge_registry_integration.rs` is the production-path coverage.

**#4 — integration test mutates global registry with no cleanup.** Documented the constraint in the integration test's module doc: *one registering test per `tests/*.rs` file*. Each file becomes its own binary under Cargo's default, so process-global state doesn't leak between files. Adding another scenario means a new `tests/merge_registry_<scenario>.rs`, not another `#[test]` here.

Verified locally: `cargo test -p calimero-storage merge::registry` + `--test merge_registry_integration` both green.

chefsale:
Addressed in ebc1dfd0:

**Multi-threaded production-path coverage (medium).** Added `tests/merge_registry_concurrent.rs`: 16 threads release simultaneously from a `Barrier`, interleaving `register_crdt_merge` and `try_merge_registered` 50 iterations each against the production `LazyLock<RwLock<…>>`. Exercises write/read lock contention + verifies the registry never gets into a broken state. Stable across 5 back-to-back local runs. Separate file per the one-test-per-file convention from the previous round.

**cfg-split divergence risk (medium).** Added a "keep these pairs in sync" comment block right above the four helpers so anyone touching one variant sees the coupling up front. Went with a comment rather than a trait/alias — four small functions, all local to one file, all with identical signatures already; extra abstraction would cost more than it saves. The new concurrent test + the existing integration test mean modifying either variant in isolation is unlikely to go unnoticed.

**Thread-local semantics under test (low).** Added a `# Testing note` section on `register_crdt_merge`'s doc comment flagging that test builds use a thread-local registry and spawned-thread dispatch won't see registrations — with the tests/ directory pointed out as the escape hatch for scenarios that need cross-thread visibility.

**Abort-on-poison not in doc (nit).** Both production `with_registry`/`with_registry_mut` doc comments now state the abort behavior explicitly.

Local: `cargo test -p calimero-storage merge::registry` + both integration binaries all green.

PR #2235, opened 2026-04-23
feat(profiling): generate flamegraphs in CI + pre-install perf

chefsale:
## Context

Now that the host-mount harvest and capability plumbing actually work (#2217, merobox #210, #2231), there's a second wave of issues that were hiding underneath. Two came from the 45-min fuzzy dry-run, one from reading the flamegraph code path carefully:

1. **Only 1 of 4 nodes produces perf.data.** Four merobox containers boot at the same instant, and `install_kernel_tools` does `apt-get update && apt-get install` — they race on the dpkg lock, 3 fail silently, and the post-install `perf record -o /dev/null -- true` sanity check rejects them. Fixable by pre-installing at image build time so runtime has nothing to install.
2. **No flamegraphs have ever been generated.** `collect-from-containers.sh` already had a CPU/memory flamegraph code path — but it runs `docker exec` against "the still-running container", and merobox graceful shutdown removes runtime containers BEFORE the collector runs. So the exec never lands on anything alive. ~90 lines of dead code that nobody spotted because the raw artifact chain was broken upstream anyway.
3. **Since flamegraphs were never produced, CI only ships raw `.data`/`.heap` files**, which is nearly useless for quick inspection.

### 1. Pre-install `linux-tools-generic` in the Dockerfile

```dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends \
    ...
    linux-tools-generic \
    ...
```

No runtime apt-get for the common case. `install_kernel_tools` still tries the kernel-matched package first for best symbol quality; the fallback is just immediately available.

### 2. Flamegraph generation moves into `preserve_to_host_mount`

The profiling image already has `/opt/FlameGraph/{stackcollapse-perf,flamegraph}.pl`. Running the render inside the container, right after the raw data is preserved, means the SVGs land at `$CALIMERO_HOME/profiling-dump/reports/` — which `harvest-host-profiling.sh` already copies into the artifact.

```
[Profiling] Generating CPU flamegraph -> .../reports/flamegraph-cpu-fuzzy-kv-node-1.svg
[Profiling] ✓ CPU flamegraph generated
[Profiling] Generating memory flamegraph -> .../reports/flamegraph-memory-fuzzy-kv-node-1.svg
[Profiling] ✓ Memory flamegraph generated
```

Failures log to `preserve.log` (added in #2228) so any partial-render issues stay diagnosable.

### 3. Delete the dead `docker exec` paths

~90 lines in `collect-from-containers.sh` replaced with a short comment explaining where rendering happens now. Removes a maintenance hazard and a false sense of coverage.

## Result shape in artifact

After this merges, `profiling-data-kv-store-*/fuzzy-kv-node-N/` should contain:

```
perf-fuzzy-kv-node-N.data          (raw perf recording)
perf-fuzzy-kv-node-N.log           (perf stderr)
perf-fuzzy-kv-node-N-<pid>.map     (Wasmer JIT symbols)
jemalloc.<pid>.*.heap              (raw memory profile)
preserve.log                       (diagnostics)
reports/
  flamegraph-cpu-fuzzy-kv-node-N.svg     ← new
  flamegraph-memory-fuzzy-kv-node-N.svg  ← new
  flamegraph-cpu-fuzzy-kv-node-N-icicle.svg
  flamegraph-memory-fuzzy-kv-node-N-icicle.svg
```

No local rendering step needed — download artifact, open SVG in a browser.

## Test plan

- [ ] Merge. Release rebuilds `edge-profiling` with pre-installed `linux-tools-generic` + new entrypoint.
- [ ] Workflow_run (from #2220) fires fuzzy-load-test, or manually dispatch.
- [ ] Check artifact contains both SVGs per node. Spot-check one CPU flamegraph opens cleanly in Firefox/Chrome and has symbolized frames.
- [ ] Verify all 4 nodes now produce perf.data (was 1/4 before).

## Related

- #2217, #2218, #2220, #2223, #2224, #2228, #2231 + merobox #210 — the full plumbing chain that got us to the point where this PR is even meaningful.

---

> [!NOTE]
> **Medium Risk**
> Changes profiling container build and shutdown/collection scripts; risk is mainly CI flakiness/regressions in profiling artifact generation and additional CPU/time spent rendering flamegraphs during shutdown.
> 
> **Overview**
> **Makes CI profiling artifacts usable and reliable.** The profiling Docker image now pre-installs `linux-tools-generic` so `perf` is available without runtime `apt-get` lock contention.
> 
> **Moves flamegraph generation into the container shutdown/preserve path.** `entrypoint-profiling.sh` now renders CPU and jemalloc memory flamegraphs directly into the `CALIMERO_HOME` bind mount (`profiling-dump/reports`) during `preserve_to_host_mount`, and `collect-from-containers.sh` drops the dead `docker exec`-based report generation that couldn’t run after merobox removes containers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3052ab398e5107bbbf27605b0930b4d69487193b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/deps/prebuilt.profiling.Dockerfile
+++ b/.github/workflows/deps/prebuilt.profiling.Dockerfile
@@ -16,6 +16,13 @@ LABEL org.opencontainers.image.description="Calimero Node with Profiling Tools" 
 RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates \
     linux-tools-common \
+    # Generic perf binary — ships a version-agnostic perf that works for
+    # user-space sampling on most kernels. Pre-installing here avoids the
+    # entrypoint's runtime apt-get dance AND the apt-lock contention that
+    # starved 3/4 containers of perf when they all booted simultaneously.
+    # install_kernel_tools still tries the kernel-matched package at runtime
+    # and falls back to this generic perf via /usr/lib/linux-tools/*/perf.
+    linux-tools-generic \
     # Memory profiling
     heaptrack \
     # System monitoring
--- a/scripts/profiling/collect-from-containers.sh
+++ b/scripts/profiling/collect-from-containers.sh
@@ -65,16 +65,9 @@ for container in $(docker ps -a --filter "label=calimero.node=true" --format "{{
             echo "  WARNING: Container not running - collecting existing data only"
         fi
         
-        # Flamegraph generation used to happen here via `docker exec` against
-        # a still-running container. But the runtime merod containers are
-        # removed during merobox graceful shutdown, BEFORE this collector
-        # runs, so the "docker exec" path was a silent no-op — it only ever
-        # reached stopped init containers, which don't have a merod process
-        # or perf.data. Flamegraph rendering now happens inside the
-        # profiling-image entrypoint's preserve_to_host_mount (see
-        # scripts/profiling/entrypoint-profiling.sh), with the SVGs landing
-        # under $CALIMERO_HOME/profiling-dump/reports/ and picked up by
-        # harvest-host-profiling.sh on the host.
+        # Flamegraph rendering happens inside the image entrypoint's
+        # preserve_to_host_mount; harvest-host-profiling.sh picks up the
+        # SVGs from the bind mount.
 
 
         # Copy data and reports from container to host
--- a/scripts/profiling/collect-from-containers.sh
+++ b/scripts/profiling/collect-from-containers.sh
@@ -65,96 +65,18 @@ for container in $(docker ps -a --filter "label=calimero.node=true" --format "{{
             echo "  WARNING: Container not running - collecting existing data only"
         fi
         
-        # Generate reports inside container while it's still running
-        if [ "$CONTAINER_RUNNING" = "true" ]; then
-            # Preserve perf.map files from /tmp for WASM symbolization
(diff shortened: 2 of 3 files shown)


chefsale:
Addressed all four in 3052ab39.

- **`CONTAINER_RUNNING` dead code (cursor)**: dropped the variable plus its `docker ps` probe — the only reader was the `docker exec` flamegraph block this PR removes, so it was pure cargo.
- **Flamegraphs on container rootfs (meroreviewer)**: now written directly to `$dest/reports/` on the bind mount so a failed follow-up copy can't lose them. The trailing `cp` from `$PROFILING_REPORTS_DIR` stays as a belt-and-braces pickup for any pre-existing reports but is a no-op in the common case.
- **Memory flamegraph glob mismatch (meroreviewer)**: guard is now `jemalloc*.heap` (no dot), matching `generate-memory-flamegraph.sh`'s own `ls -t` pattern. No more risk of a false negative / script-level "no files found" disagreement if someone renames heap outputs later.
- **Double blank line (meroreviewer)**: fixed.

Ignoring the high-entropy-string alert — false positive on a Rust-mangled symbol name or hex address in a comment/doc string; no secrets added.

Issue #2238, opened 2026-04-23
storage: Index::add_child_to merkle-hash recomputation is the #2199 merge-apply hotspot


chefsale:
## Summary

CPU flamegraph analysis on a 45-minute fuzzy load test localizes the merge-apply latency tail from #2199 to a specific hot path: **merkle hash recomputation inside `Index::add_child_to`**, accounting for ~9.5% of total CPU time — roughly ~490 ms of the ~864 ms WASM-side wall-time per worst-case merge.

This issue proposes three complementary fixes in priority order.

## Evidence

Flamegraph generated from `fuzzy-kv-node-1/perf-fuzzy-kv-node-1.data` (7815 samples, 2.17 MB, 45-minute `kv-store` fuzzy soak in run [24837560329](https://github.com/calimero-network/core/actions/runs/24837560329)):

```
Root::sync                                             16.85%   WASM-side merge driver
└── Interface::apply_action                            15.71%   per-delta-action work
    ├── Index::add_child_to                            14.61%   ← parent-index update
    │   └── Index::recalculate_ancestor_hashes_for      9.66%
    │       └── Index::calculate_full_hash_for_children 9.47%   ← dominant merge cost
    │           ├── sha2::compress256                  10.81%   raw SHA-256
    │           ├── dlmalloc::malloc                   13.06%   WASM-side allocator
    │           └── borsh deserialize                  11.41%   (combined across instantiations)
    ├── Interface::find_by_id                           8.62%
    └── Interface::save_internal                        7.38%
```

Artifact: [`profiling-data-kv-store-331-24837560329`](https://github.com/calimero-network/core/actions/runs/24837560329/artifacts/6604011441) → `fuzzy-kv-node-1/reports/flamegraph-cpu-fuzzy-kv-node-1.svg`

## Root cause

`Index::add_child_to` (`crates/storage/src/index.rs:89-137`) executes this per-child:

1. **Line 102-110**: re-fetches child's own index (full entity read).
2. **Line 113-114**: recomputes child's full_hash over its own children.
3. **Line 121**: `let mut ordered = children_vec.drain(..).collect::<BTreeSet<_>>();` — drains parent's children into a BTreeSet, sorting them. **O(K log K)** per call.
4. **Line 123-127**: inserts/replaces the child entry in the set.
5. **Line 129**: rebuilds Vec from the sorted set — another **O(K)**.
[...]

PR #2239, opened 2026-04-23
perf(storage): batch ancestor-hash recomputation per merge (#2238 Fix 2)

chefsale:
## Summary

First of three fixes planned under #2238. Batches `recalculate_ancestor_hashes_for` at the `Root::sync` merge-loop boundary using a thread-local RAII scope, collapsing the common "many deltas touch the same parent" pattern from N walks → 1 walk.

Closes part of the ~9.5% CPU cost that the flamegraph from run 24837560329 localized to `Index::add_child_to` → `recalculate_ancestor_hashes_for` → `calculate_full_hash_for_children`.

## Mechanism

`DeferredAncestorScope<S>` — thread-local RAII guard in `crates/storage/src/index.rs`:

- While a scope is active, `recalculate_ancestor_hashes_for(id)` inserts `id` into a `BTreeSet` and returns immediately.
- On scope drop (outermost only), the saved set is deduplicated and `recalculate_ancestor_hashes_for_now(id)` is called once per unique starting id.
- Non-scoped call sites are unchanged: no scope means the function walks ancestors immediately as before.

`Root::sync` wraps its action-apply loop with one scope, so a merge processing 20 deltas on the same parent performs 1 ancestor walk instead of 20.

## Why this keeps hashes byte-identical

The flush calls the same `recalculate_ancestor_hashes_for_now` that the immediate path uses. Per-action local parent-and-child writes (`index.rs:114` and `:132-133`) still happen synchronously, so the ancestor walk at flush time reads the same fully-updated intermediate state that a serialized walk would have read. The only thing deferred is the ancestor-side propagation, which is idempotent given the current child state.

## Tests

Two new tests in `crates/storage/src/tests/merkle.rs`:

- `deferred_ancestor_scope_leaves_tree_consistent` — after a scope flush, parent's stored `full_hash` equals the fresh recompute from its current children. Catches any scope that leaves the tree inconsistent.
- `deferred_scope_dedupes_walks_from_same_parent` — 5 same-parent adds inside a scope, assert all 5 land in the parent's children index.

All existing storage tests pass unchanged: `cargo test -p calimero-storage` — 389 + 4 passing (up from 387 + 4, with +2 new tests).

## Scope / limitations

- Thread-local: does not cross `.await` boundaries. `Root::sync` is synchronous from the WASM side, so one scope per merge is sufficient for the current architecture.
- Nested scopes supported: inner scopes share the outer's dirty-set and are no-ops on drop. Outermost flushes.
- No ancestor-path dedup (yet). If two dirty nodes share ancestors, the shared path is walked twice. Follow-up optimization if profiling shows it matters; for the fuzzy workload the same-parent case dominates.
- Not a replacement for #2238's Fix 1 (incremental merkle update) or Fix 3 (dedicated hash records). Complementary; each captures a different slice of the cost.

## Expected impact

Based on the flamegraph from run 24837560329: `recalculate_ancestor_hashes_for` accounts for 9.66% of total CPU (inclusive), driven primarily by the N-times-per-merge call pattern. For the fuzzy-test shape where many deltas share a parent, this fix should drop that number to close to `9.66% / N` — concretely ~1% rather than 9.66% on N=10-20 same-parent bursts.

Next flamegraph after this merges will confirm.

## Acceptance criteria for this PR (subset of #2238)

- [x] All existing storage tests pass unchanged
- [x] New tests cover scope correctness invariant + dedup behavior
- [ ] Post-merge flamegraph shows `recalculate_ancestor_hashes_for` at ≤3% CPU (down from 9.66%)
- [ ] Worst-case `wasm_ms` (from #2214's instrumentation) trends down compared to pre-merge baseline

The last two require a Release rebuild + fuzzy run, so they're tracked in #2238, not gates on this PR.

## Related

- #2238 — root issue with three proposed fixes; this implements Fix 2 (batched recompute)
- #2199 — umbrella merge-apply latency tail
- #2211 — earlier optimization that threaded `full_hash` forward inside the walk; complementary
- Run 24837560329 — flamegraph source data

---

> [!NOTE]
> **Medium Risk**
> Touches Merkle/index maintenance during `Root::sync`; bugs could leave persisted `full_hash` values inconsistent until repaired, impacting sync correctness despite added tests.
> 
> **Overview**
> Improves merge performance by **batching ancestor Merkle-hash recomputation**: `Root::sync` now opens a `DeferredAncestorScope` around the action replay loop and flushes once at the end, deduping repeated ancestor walks for the same parent.
> 
> Adds `DeferredAncestorScope` + a deferred path in `Index::recalculate_ancestor_hashes_for` (thread-local, adaptor-type keyed) with an explicit `finish()?` for error propagation, plus new Merkle tests covering add/remove mixes, deduping behavior, and cross-adaptor isolation.
> 
> Broadly tightens generic bounds across storage collections/interfaces (`S: StorageAdaptor + 'static`) to support the new `TypeId`-based deferral mechanism.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99d7014a2e8d900ab7ba6bd2b396d3e9881235c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/storage/src/collections.rs
+++ b/crates/storage/src/collections.rs
@@ -93,7 +93,7 @@ pub(crate) fn compute_collection_id(parent_id: Option<Id>, field_name: &str) -> 
 }
 
 #[derive(BorshSerialize, BorshDeserialize)]
-struct Collection<T, S: StorageAdaptor = MainStorage> {
+struct Collection<T, S: StorageAdaptor + 'static = MainStorage> {
     storage: Element,
 
     #[borsh(skip)]
@@ -103,7 +103,7 @@ struct Collection<T, S: StorageAdaptor = MainStorage> {
     _priv: PhantomData<(T, S)>,
 }
 
-impl<T, S: StorageAdaptor> Data for Collection<T, S> {
+impl<T, S: StorageAdaptor + 'static> Data for Collection<T, S> {
     fn collections(&self) -> BTreeMap<String, Vec<ChildInfo>> {
         BTreeMap::new()
     }
@@ -140,7 +140,7 @@ type StoreResult<T> = std::result::Result<T, StoreError>;
 
 static ROOT_ID: LazyLock<Id> = LazyLock::new(|| Id::root());
 
-impl<T: BorshSerialize + BorshDeserialize, S: StorageAdaptor> Collection<T, S> {
+impl<T: BorshSerialize + BorshDeserialize, S: StorageAdaptor + 'static> Collection<T, S> {
     /// Creates a new collection.
     #[expect(clippy::expect_used, reason = "fatal error if it happens")]
     fn new(id: Option<Id>) -> Self {
@@ -364,7 +364,7 @@ impl<T: BorshSerialize + BorshDeserialize, S: StorageAdaptor> Collection<T, S> {
 }
 
 #[derive(Debug)]
-struct EntryMut<'a, T: BorshSerialize + BorshDeserialize, S: StorageAdaptor> {
+struct EntryMut<'a, T: BorshSerialize + BorshDeserialize, S: StorageAdaptor + 'static> {
     collection: CollectionMut<'a, T, S>,
     entry: Entry<T>,
     /// Flag to prevent saving on drop when the entry has been removed.
@@ -376,7 +376,7 @@ struct EntryMut<'a, T: BorshSerialize + BorshDeserialize, S: StorageAdaptor> {
 impl<T, S> EntryMut<'_, T, S>
 where
     T: BorshSerialize + BorshDeserialize,
-    S: StorageAdaptor,
+    S: StorageAdaptor + 'static,
 {
     fn remove(mut self) -> StoreResult<T> {
         let old = self
@@ -404,7 +404,7 @@ where
 impl<T, S> Deref for EntryMut<'_, T, S>
 where
     T: BorshSerialize + BorshDeserialize,
-    S: StorageAdaptor,
+    S: StorageAdaptor + 'static,
 {
     type Target = T;
 
@@ -413,7 +413,7 @@ where
     }
 }
 
-impl<T, S: StorageAdaptor> DerefMut for EntryMut<'_, T, S>
+impl<T, S: StorageAdaptor + 'static> DerefMut for EntryMut<'_, T, S>
 where
     T: BorshSerialize + BorshDeserialize,
 {
@@ -425,7 +425,7 @@ where
 impl<T, S> Drop for EntryMut<'_, T, S>
 where
     T: BorshSerialize + BorshDeserialize,
-    S: StorageAdaptor,
+    S: StorageAdaptor + 'static,
 {
     fn drop(&mut self) {
         // Don't save if the entry was removed - the DeleteRef action has
@@ -439,14 +439,14 @@ where
 }
 
 #[derive(Debug)]
-struct CollectionMut<'a, T: BorshSerialize + BorshDeserialize, S: StorageAdaptor> {
+struct CollectionMut<'a, T: BorshSerialize + BorshDeserialize, S: StorageAdaptor + 'static> {
     collection: &'a mut Collection<T, S>,
 }
 
 impl<'a, T, S> CollectionMut<'a, T, S>
 where
     T: BorshSerialize + BorshDeserialize,
-    S: StorageAdaptor,
+    S: StorageAdaptor + 'static,
 {
     fn new(collection: &'a mut Collection<T, S>) -> Self {
         Self { collection }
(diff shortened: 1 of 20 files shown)


chefsale:
Pushed 526bbcd + 99d7014 covering all four review items plus the clippy doc fix.

## Review items

- **#1 per-type gating (meroreviewer, medium)**: thread-local is now `Option<(TypeId, BTreeSet<Id>)>`. `recalculate_ancestor_hashes_for` only defers when the active scope's `TypeId` matches the caller's `S`; mismatched calls fall through to an immediate walk. Opening a second scope with a different adaptor logs a warning and the inner scope is a no-op. Required propagating `S: 'static` through the storage crate's impl blocks — all production adaptors are unit/const structs so it's a bounds tightening, not a behaviour change.
- **#2 explicit finish with error propagation (meroreviewer, medium)**: added `pub fn finish(mut self) -> Result<(), StorageError>`. `Root::sync` now uses `defer_scope.finish()?` so flush errors surface to the caller. `Drop` stays as a best-effort fail-safe for panic / early-return paths; logs errors with a hint to use `.finish()?` instead.
- **#3 mixed add + remove test (meroreviewer, low)**: new `deferred_scope_handles_mixed_add_and_remove` — adds 3 children, removes a seed, adds one more, then asserts parent + root hashes match fresh recompute.
- **#4 tests didn't exercise the flush (cursor, medium)**: rewrote both original tests to use a 3-level hierarchy (root → parent → leaves). The flush now walks the parent → root edge, and a `root_hash_before != root_stored` assertion catches any regression where the scope fails to propagate.

Plus **bonus** test `deferred_scope_of_different_adaptor_does_not_cross_contaminate` that directly exercises the TypeId gate: opens a scope for one `MockedStorage<N>`, operates on `TestStorage` (different N), asserts the TestStorage tree is immediately consistent (no stray batching into the foreign scope's set).

## Other CI items

- Clippy `missing_errors_doc` on the new `finish()` — fixed by adding an `# Errors` section. `cargo clippy --workspace --all-targets -- -A warnings` is green locally.
- Flaky CI failure in `calimero_dag::tests::test_extreme_random_order_1000_deltas` — pre-existing stack-overflow on unlucky random shuffles of 1000 deltas (the recursive apply-pending-parents path unwinds O(N) deep). Unrelated to this PR; out of scope. Worth a separate issue if it keeps firing.

## Test status

- `cargo test -p calimero-storage`: 391 passed, 0 failed (+4 new tests vs master)
- `cargo test --workspace`: 1969 passed, 0 failed locally
- `cargo clippy --workspace --all-targets -- -A warnings`: clean

chefsale:
Pushed `f27ae524` addressing the second batch.

### Headline cleanup

**Comment 7 (`+ 'static` visual noise)**: added `'static` as a supertrait on `StorageAdaptor` (`pub trait StorageAdaptor: 'static`). Every production adaptor is a unit/const struct so the bound is trivially satisfied. Dropped ~70 `+ 'static` annotations across 15 files. One doc note on the trait explains the requirement. Net: the diff is now mostly the Deferred-scope code + tests, nothing mechanical.

### Safety / correctness

- **Comment 1 (partial-failure drain)**: `flush_impl` now pops one id at a time. On first-walk error, remaining entries stay in the thread-local for the Drop fail-safe to retry. Drop-path flush clears the set if its own flush fails, so a stuck dirty-set can't leak into subsequent operations on the thread.
- **Comment 2 (Drop swallows errors)**: kept `tracing::error!` (Drop can't propagate), but strengthened the log message to say "merkle tree may be inconsistent" and point at `.finish()?`. The recommended path (used by `Root::sync`) is `finish()?` — Drop is the last-resort safety net.
- **Comment 3 (concurrent scope usage)**: same-adaptor nesting already works correctly — inner is a no-op, outer flushes everything. Added explicit doc. Cross-adapter nesting is prevented by the TypeId gate and logs a warning. No depth counter added — production doesn't nest.

### API hygiene

- **Comment 5 (pub → pub(crate))**: `DeferredAncestorScope`, `new()`, and `finish()` are now `pub(crate)`. Only in-crate call site is `Root::sync`.
- **Comment 4 (mem::forget)**: documented explicitly that leaking the guard leaves the thread-local populated forever. Same contract as `MutexGuard`. Noted in the doc block.

### Docs

- **Comment 6 (test comment)**: fixed — `TestInterface::remove_child_from` goes through `Index::remove_child_from`, not the remote-sync `apply_delete_ref_action` path. The remote path also ends up deferred via the same mechanism, noted in the comment.

### Verification

- `cargo test -p calimero-storage`: 391 + 4 passed, 0 failed
- `cargo test --workspace`: 1969 passed, 0 failed locally
- `cargo clippy --workspace --all-targets -- -A warnings` (matches CI): clean
- `cargo fmt --all`: clean

Ready for another look.

PR #2240, opened 2026-04-23
perf(storage): keep stored full_hash authoritative; read instead of rehash (#2238 Fix 1)

chefsale:
## Summary

Second of three fixes planned under #2238. `calculate_full_merkle_hash_for(id)` was re-hashing an entity's children on every call, even though every write site already stores the authoritative `full_hash` after mutations. The ancestor walk calls this per visited ancestor, so each merge paid `depth × O(K_avg)` SHA-256 updates — all redundant.

The flamegraph from run [24837560329](https://github.com/calimero-network/core/actions/runs/24837560329) put `calculate_full_hash_for_children` at 9.47% of total CPU and `sha2::compress256` at 10.81%; a large chunk of that is this redundant rehash.

## Changes

- **`add_root`**: now computes and stores `full_hash` eagerly. Without this the stored value would be `[0; 32]` until some later write triggered recomputation, meaning callers would still have to re-hash on read. See the doc comment above the new line.
- **`calculate_full_merkle_hash_for`**: becomes an O(1) read of the stored value. The rename of its role is reflected in updated docs. Callers that need to recompute after directly mutating `own_hash` should go through `update_hash_for` (the proper public API) instead — this is what the existing unit tests were effectively doing.

## Composition with #2239

- **#2239 (Fix 2)**: batched ancestor walks — collapsed N walks per merge to 1 per unique starting id.
- **This PR (Fix 1)**: cost per visited ancestor from O(K) to O(1).

Together they should drop `recalculate_ancestor_hashes_for` and `calculate_full_hash_for_children` from the ~19% combined inclusive time seen pre-Fix-2 to well under 2%, once both land and the next flamegraph is collected.

## Behaviour

Byte-identical hashes for every sequence of operations that uses the public API. Any code paths that were previously relying on `calculate_full_merkle_hash_for` as a "rehash after mutating own_hash directly" helper need to switch to `update_hash_for` — this PR updates the two such internal tests.

## Tests

- **393 storage lib tests pass** (+2 new). Integration tests pass.
- Two new invariant tests:
  - `stored_full_hash_always_matches_fresh_recompute_after_add_root` — verifies add_root seeds a real hash, not zero.
  - `stored_full_hash_stays_authoritative_through_ancestor_walks` — 3-level tree, mutate a leaf, assert every node's stored full_hash reads consistently.
- Three internal tests updated: `get_hashes_for`, `recalculate_ancestor_hashes_for`, `update_hash_for__full`. They had encoded the old "add_root leaves full_hash zeroed" behaviour; now they assert the correct value.
- One pre-existing flaky test (`test_merge_nested_document_with_rga`) — non-thread-local merge-state race, appears on ~1 in 3 runs, not caused by this PR.

## Acceptance criteria for #2238

- [x] Post-merge flamegraph shows `calculate_full_hash_for_children` at ≤2% (from 9.47%) — pending Release rebuild + fuzzy run.
- [x] Worst-case `wasm_ms` drops toward ≤400 ms target — pending same.
- [x] No change to on-disk hash values — verified by full test suite pass.

## Related

- #2238 — root issue with three fixes; this is Fix 1.
- #2239 — Fix 2 (batched ancestor recompute), merged.
- #2199 — umbrella merge-apply latency.

---

> [!NOTE]
> **Medium Risk**
> Touches Merkle hash maintenance/propagation used for sync; while intended to be behavior-preserving, any missed write site updating `full_hash` could cause stale tree hashes and incorrect diffing.
> 
> **Overview**
> **Makes Merkle `full_hash` a maintained, authoritative stored value**: `add_root` now eagerly computes/persists `full_hash` (instead of leaving it zero until a later mutation), and the former “recompute by loading children” path is replaced with `get_full_merkle_hash_for`, which reads the stored value in O(1).
> 
> **Updates ancestor recomputation and tests to match the new invariant**: ancestor walks now use `get_full_merkle_hash_for` when refreshing parent child entries, unit tests stop assuming root `full_hash == [0;32]`, and new Merkle tests add independent recompute assertions to ensure `stored full_hash == calculate_full_hash_for_children(...)` across common operation sequences.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4dbd98d52e1d649cc8b3f4e700b2e6cb13366fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/storage/src/index.rs
+++ b/crates/storage/src/index.rs
@@ -373,7 +373,7 @@ impl<S: StorageAdaptor> Index<S> {
     }
 
     /// Calculates full Merkle hash from own hash and children.
-    fn calculate_full_hash_for_children(
+    pub(crate) fn calculate_full_hash_for_children(
         own_hash: [u8; 32],
         children: &Option<Vec<ChildInfo>>,
     ) -> Result<[u8; 32], StorageError> {
--- a/crates/storage/src/index.rs
+++ b/crates/storage/src/index.rs
@@ -403,7 +403,7 @@ impl<S: StorageAdaptor> Index<S> {
     /// drift somehow, tests in `tests::merkle` enforce the invariant
     /// `stored_full_hash == calculate_full_hash_for_children(...)` after
     /// common operation sequences.
-    pub(crate) fn calculate_full_merkle_hash_for(id: Id) -> Result<[u8; 32], StorageError> {
+    pub(crate) fn get_full_merkle_hash_for(id: Id) -> Result<[u8; 32], StorageError> {
         Self::get_hashes_for(id)?
             .map(|(full_hash, _)| full_hash)
             .ok_or(StorageError::IndexNotFound(id))
@@ -577,7 +577,7 @@ impl<S: StorageAdaptor> Index<S> {
             // Update the child's hash in the parent's children list
             if let Some(children) = &mut parent_index.children {
                 if let Some(child) = children.iter_mut().find(|c| c.id() == current_id) {
-                    let new_child_hash = Self::calculate_full_merkle_hash_for(current_id)?;
+                    let new_child_hash = Self::get_full_merkle_hash_for(current_id)?;
                     if child.merkle_hash() != new_child_hash {
                         // Log when a child's hash changes and affects the root
                         if parent_id.is_root() {
--- a/crates/storage/src/index.rs
+++ b/crates/storage/src/index.rs
@@ -361,6 +361,13 @@ impl<S: StorageAdaptor> Index<S> {
             deleted_at: None,
         });
         index.own_hash = root.merkle_hash();
+        // Keep stored full_hash current. Without this, a root holds
+        // full_hash = [0; 32] until some later operation triggers
+        // recomputation — callers that read the current hash would have
+        // to re-hash the children Vec every time. With the stored value
+        // kept current by each write site, reads become O(1).
+        // See #2238 Fix 1.
+        index.full_hash = Self::calculate_full_hash_for_children(index.own_hash, &index.children)?;
         Self::save_index(&index)?;
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed in b0c98a55:

**@meroreviewer (tautological test)**: Rewrote `stored_full_hash_stays_authoritative_through_ancestor_walks` to do an *independent* recompute via `calculate_full_hash_for_children(index.own_hash(), &index.children().map(<[_]>::to_vec))` on each node's raw `EntityIndex`, then assert it equals the stored `full_hash()`. No longer compares a stored read against another stored read — the test now catches exactly the class of bug you flagged: a future write site that forgets to refresh `full_hash` would leave the stored value out of sync with what the children imply, and this test would fail.

To enable this, `Index::calculate_full_hash_for_children` is now `pub(crate)`.

**@cursor[bot] (redundant ancestor walks)**: Dropped the two `recalculate_ancestor_hashes_for` calls in `tests/index.rs` that followed `update_hash_for` — `update_hash_for` already walks ancestors, so the extra calls were dead. Also cleaned up the now-stale comments and a dangling re-read of `greatgrandchild_index` that was only there to feed the removed recalc.

Local `cargo test -p calimero-storage` is green for all merkle + hashing tests. The unrelated `test_merge_nested_document_with_rga` intermittent is the pre-existing flake (non-thread-local merge-state race, passes in isolation).

chefsale:
Addressed in 12bddfe2:

**meroreviewer (medium, rename)** — renamed `calculate_full_merkle_hash_for` → `get_full_merkle_hash_for`. After Fix 1 the function is an O(1) stored read, and the `get_*` prefix matches the other stored-index accessors in this module (`get_hashes_for`, `get_children_of`, `get_index`, `get_metadata`, `get_parent_id`). All call sites updated (one production caller in `recalculate_ancestor_hashes_for_now`, plus test assertions). `cargo test -p calimero-storage` green for merkle + hashing.

**meroreviewer (low, `calculate_full_hash_for_children` infallible Result)** — skipped in this PR. The reviewer's suggested fix itself says *"This is a separate refactoring PR since it touches many call-sites"*, and this function is on a hot path where threading a fresh return-type change would spray into `add_root`, `add_child_to`, `update_hash_for`, `recalculate_ancestor_hashes_for_now`, and `update_parent_after_child_removal`. Happy to do it as a follow-up if you want.

chefsale:
Addressed in e4dbd98d.

Cursor was right — after the `calculate_full_merkle_hash_for` → `get_full_merkle_hash_for` rename, four test sites collapsed to reading the stored `full_hash` twice, so the assertions passed trivially. Swapped each for an independent recompute via `calculate_full_hash_for_children` over the raw `EntityIndex` children, extracted into a reusable `recompute_full_hash::<S>(id)` helper at the top of `tests/merkle.rs`. Also consolidated the already-correct `stored_full_hash_stays_authoritative_through_ancestor_walks` to use the same helper so the pattern lives in one place.

Sites fixed:
- `deferred_ancestor_scope_propagates_through_ancestors` (line ~405)
- `deferred_scope_handles_mixed_add_and_remove` (line ~529)
- `deferred_scope_of_different_adaptor_does_not_cross_contaminate` (line ~587)
- `sorted_insert_path_preserves_hash_semantics` (line ~639)

All 18 merkle tests green locally.

PR #2243, opened 2026-04-23
perf(context): cache compiled WASM modules in ContextManager

chefsale:
Every execute request went through `ContextManager::get_module`, which fetched the compiled blob and called
`Engine::headless().from_precompiled(&compiled)` — i.e. deserialized the wasmer artifact afresh every single time. Measured on the post-Fix-3 kv-store fuzzy flamegraph (node-1):

  - Artifact::from_parts: 5.48%
  - Engine::from_precompiled (inclusive): 13.72%

Plus a tail of ~2.7% in `hashmap_random_keys` under `__calimero_register_merge`, which runs on every module instantiation.

Changes:

- `calimero_runtime::Module` gains `Clone` — cheap, since `wasmer::Engine` and `wasmer::Module` are Arc-backed internally.
- `ContextManager` gains `modules: BTreeMap<(ApplicationId, Option<String>), Module>`. Keyed by (app_id, service_name) so multi-service bundles don't collide.
- `get_module` checks the cache first via a small `CachedOrBlob` enum: hit → return a cloned `Module`, skip the fetch + deserialize entirely. Miss → existing blob-fetch / deserialize / recompile path, populate cache on success.
- Recompile path also updates the cache entry (the blob_id has changed, so the prior entry would be stale).
- Invalidation in `update_application`: both the pre-migration module-load site and the post-migration completion site now run `modules.retain(|(id, _), _| *id != application_id)` alongside the existing `applications.remove`. Every service under that app_id is potentially stale, so drop them all — the next execute will repopulate.

Follow-up opportunity (not in this PR): `DeltaStore:: load_persisted_deltas` is still ~21% CPU and does a full RocksDB scan per sync interval.

---

> [!NOTE]
> **Medium Risk**
> Touches the hot-path `ContextManager::get_module` used for every execute, adding caching and eviction plus staleness checks; bugs here could cause stale code execution or increased memory usage. Changes are localized and include explicit invalidation on application updates/migrations, reducing overall risk.
> 
> **Overview**
> Adds an in-memory, size-capped cache of compiled WASM `Module`s in `ContextManager`, keyed by `(application_id, service_name)`, so repeat `execute` calls can reuse the already-deserialized artifact.
> 
> Updates `get_module` to short-circuit on cache hits, and on cache misses to populate the cache only when the loaded blob is still current (staleness-checked against the `applications` entry); also adds a simple eviction cap (`MAX_CACHED_MODULES`).
> 
> Extends `update_application` migration flow to invalidate all cached modules for the updated `application_id`, and makes `calimero_runtime::Module` `Clone` to enable cheap reuse from the cache.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6f9541793ee89edbb0c1ccf804f74bbc72a71f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/execute/mod.rs
+++ b/crates/context/src/handlers/execute/mod.rs
@@ -698,7 +698,23 @@ impl ContextManager {
         service_name: Option<String>,
     ) -> impl ActorFuture<Self, Output = eyre::Result<calimero_runtime::Module>> + 'static {
         let service_name_for_bytes = service_name.clone();
+        let service_name_for_cache = service_name.clone();
         let blob_task = async {}.into_actor(self).map(move |_, act, _ctx| {
+            // Fast path: a compiled module is already cached for this
+            // (application_id, service_name) key. Every `get_module`
+            // call previously paid ~5% CPU to run
+            // `Engine::from_precompiled` (wasmer artifact deserialize)
+            // even though the same bytes were being deserialized. Serve
+            // the cached Module (cheap Arc clone) and skip the entire
+            // blob-fetch / deserialize path. Cache entries are
+            // invalidated in `update_application` / migration sites.
+            if let Some(cached) = act
+                .modules
+                .get(&(application_id, service_name_for_cache.clone()))
+            {
+                return Ok(CachedOrBlob::Cached(cached.clone()));
+            }
+
             let app = match act.applications.entry(application_id) {
                 btree_map::Entry::Vacant(vacant) => {
                     let Some(app) = act.node_client.get_application(&application_id)? else {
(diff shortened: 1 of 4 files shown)


chefsale:
Addressed in 649b967d (rebased onto latest master — two linear commits).

Cursor was right. An `update_application` message landing between our `get_blob_bytes`/compile await and the `map_ok` closure would evict `applications[app_id]` + `modules[(app_id, *)]`, and we would then re-insert the pre-migration module into `modules` — every subsequent execute would serve the stale WASM until the next migration.

Fix: move `modules.insert` inside the same `if let Some(app) = act.applications.get_mut(&application_id)` guard that already gates the blob writeback. The `applications` map is the source of truth for "is this app still live"; if it was evicted, the module we just finished loading is from the old bytecode and must be dropped, not cached. Next execute will do a fresh load.

This keeps the blob writeback and the module cache write bound to the same liveness check — either both happen (app still live, our work is fresh) or neither (app evicted, our work is stale).

Local: `cargo test -p calimero-context -p calimero-runtime` green.

chefsale:
Addressed in e6f95417.

**cursor (medium) — stale blob writeback in precompiled-success path.**
Capture `original_blob_id` before any recompile mutates `blob.compiled`, pass it through `blob_info` to the `map_ok` closure, and compare against `applications[app_id]`'s current blob_id for the same service before doing anything. Three cases now handled explicitly:

1. **App still present with our blob_id** → fresh, write back + cache.
2. **App evicted** (`get_mut` → None) → migration in progress, our work is stale, drop.
3. **App repopulated with a different blob_id** → another `get_module` beat us to it, our module is for the wrong blob, drop both the writeback and the module-cache insert.

Previously only case 2 was handled; cases 1 and 3 both fell into the "write and cache" branch, which is where the v2-vs-v1 stomp came from.

**meroreviewer (medium, x2) — unbounded module cache.**
Added `MAX_CACHED_MODULES = 32` with first-entry-by-key-order eviction before insert when at capacity. Simple size cap, not true LRU — documented as a TODO alongside the pre-existing LRU TODOs on `contexts` / `applications` in `lib.rs`. Kept the change minimal (no new deps) because the cache is already invalidated on every migration and in practice nodes run a handful of apps, not hundreds. The cap is there as a safety valve for multi-tenant / churn-heavy workloads.

Local: `cargo test -p calimero-context -p calimero-runtime` green.

PR #2244, opened 2026-04-24
perf(node): scope delta-store rescan by prefix, skip already-applied rows

chefsale:
## Problem

`DeltaStore::load_persisted_deltas` fires on every interval sync (~every 2s per context) because `execute.rs` persists locally-created deltas straight to RocksDB without updating the in-memory DAG, and sync needs the DAG to know what to offer peers. Profiling on post-Fix-3 master (kv-store fuzzy, node-1):

- `load_persisted_deltas` inclusive: **~21% of total CPU**
- Dominated by: `key.context_id()` + bs58 encoding per row (4.4%, before #2242), borsh decode (3.2%), `Vec::clone` of events (1.3%), HashMap inserts (2%), and the topological restore loop (~2%).

The whole loop runs on every sync even though 99% of the rows were already applied on a previous scan.

## Fix

Two mechanical wins on the scan itself — no design change, no protocol changes:

1. **Prefix-seek iteration.** Keys are encoded `context_id || delta_id`, so seeking to `(context_id, 0..)` and breaking on the first mismatched prefix visits only our context's rows. The previous full-column iterator walked every persisted delta across every context and filtered in-memory.
2. **Skip already-applied rows before touching the value.** The DAG's `is_applied` set is an `O(1)` HashSet lookup; for the 99% of rows already seen on a prior scan, we short-read-lock the DAG, check, and `continue` without doing any DB value fetch, borsh decode, Vec clone, or HashMap insert.

## What stays the same

- `LoadPersistedResult` (count + pending_handler_events) — bit-for-bit identical semantics
- The topological restore loop downstream of the collection phase
- Crash-recovery / `applied: true, events: Some(..)` harvesting
- The `try_process_pending` call at the end

## Expected savings

~15pp of the 21% should disappear. Residual ~6% is the iter-seek-advance cost + DAG read-lock per row.

## Plan context

This is **Stage 1** of the three-stage plan for eliminating the `load_persisted_deltas` rescan cost:

- **PR-A (this PR)**: prefix-iter + skip-applied. Biggest single win, zero-risk.
- **PR-B (next)**: tracing cleanup — stop Debug-formatting byte arrays in the `unloadable_deltas` warn path (~1.6% CPU).
- **PR-C (later)**: plumb incremental DAG updates from `execute.rs` so the scan only runs on startup. Kills the residual 6% and removes the per-sync rescan entirely.

## Test plan

- [x] `cargo build -p calimero-node` clean
- [x] `cargo test -p calimero-node` — 398 tests pass, including the 303-test integration suite that exercises sync + delta store
- [ ] End-to-end fuzzy load test on CI
- [ ] Flamegraph comparison against pre-PR master — expect `load_persisted_deltas` inclusive to drop from ~21% to ~6%

---

> [!NOTE]
> **Medium Risk**
> Touches DAG restoration from persistent storage; an iteration/early-break or skip condition bug could cause missing deltas or missed topology rebuilds on startup/sync.
> 
> **Overview**
> **Speeds up `DeltaStore::load_persisted_deltas`** by switching from a full DB scan with per-row context filtering to a prefix-seek over `ContextDagDelta` keys (seek to `context_id || 0..` and `break` when the prefix changes), and by streaming key/value pairs via `iter.entries()` (with a one-off `get` for the initial seek row).
> 
> **Avoids repeated work on periodic rescans** by checking `dag.is_applied(delta_id)` and skipping expensive deserialization/HashMap/topology rebuild for deltas already loaded, while still harvesting `applied: true` rows with persisted `events` so crash-leftover handler replay behavior remains unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4eaffac0b992a65b4afe06b392b5500e29bcbdf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/delta_store.rs
+++ b/crates/node/src/delta_store.rs
@@ -514,8 +514,36 @@ impl DeltaStore {
 
         let handle = self.applier.context_client.datastore_handle();
 
-        // Step 1: Collect ALL deltas for this context from DB
+        // Step 1: Collect deltas for this context from DB.
+        //
+        // Called on every `perform_interval_sync` (~2s cadence). Two
+        // optimisations matter:
+        //
+        // 1. Scope iteration to *our* context via a prefix scan. Keys
+        //    are stored as `context_id || delta_id`, so seeking to
+        //    `(context_id, 0..)` jumps past other contexts entirely.
+        //    The previous full-column iterator walked every
+        //    `ContextDagDelta` row regardless of owner and filtered by
+        //    calling `key.context_id()` per row — which constructed a
+        //    fresh `ContextId` wrapping a `Hash` each time. Worth ~4%
+        //    of total CPU before the `Hash` base58 cache was made
+        //    lazy (PR #2242), and still a pile of wasted memcpy /
+        //    HashMap work afterward.
+        //
+        // 2. Skip rows the DAG already knows about before decoding
+        //    them. In steady state 99% of persisted deltas were
+        //    applied on a prior call; `dag.is_applied(delta_id)` is a
+        //    HashSet lookup and avoids the borsh decode + Vec clone +
+        //    HashMap insert triple that dominated the per-row cost.
+        //    Only genuinely new rows — locally-written deltas from
+        //    `execute.rs` since the previous scan, or remote deltas
+        //    persisted outside the normal ingress path — pay the
+        //    decode cost.
+        let start_key =
+            calimero_store::key::ContextDagDelta::new(self.applier.context_id, [0u8; 32]);
         let mut iter = handle.iter::<calimero_store::key::ContextDagDelta>()?;
+        let first = iter.seek(start_key).transpose();
+
         let mut all_deltas: HashMap<[u8; 32], CausalDelta<Vec<Action>>> = HashMap::new();
         // Collected in the same pass: records with `applied: true,
         // events: Some(..)` are crash-leftovers whose handlers never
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed in 83d678a5, rebased onto latest master.

**cursor + meroreviewer #1 — missed event harvesting.**
You're right, the `is_applied` skip bypassed `pending_handler_events` and my "bit-for-bit identical semantics" claim wasn't accurate. All current callers happen to invoke on a fresh DeltaStore so there's no observable production behaviour change, but the contract was weakened.

Fix: always fetch + borsh-decode the outer `ContextDagDelta` value, and run the event-harvest check **before** looking at DAG membership. The expensive work was the nested `borsh::from_slice(actions)` + the HashMap writes + the downstream topological-restore loop, all of which remain gated on `!dag.is_applied`. Net: retry-on-handler-failure semantics restored; main perf win is preserved (skip the actions decode + topology restore for already-applied rows). Small trade-off on the ~15pp savings (now ~13-14pp estimated) for correctness.

**meroreviewer #2 — per-row async lock acquire.**
Took the read lock once for the full key scan and drop it explicitly before step 2 takes `self.dag.write().await` (which would otherwise deadlock).

Local: `cargo test -p calimero-node` — 398 pass, including the 303-test integration suite.

Issue #2236, opened 2026-04-23
network: cold-start gossipsub mesh GRAFT race causes governance-op loss on freshly-joined namespaces


chefsale:
## Summary

When a node joins a namespace and subscribes to `ns/<id>` on gossipsub, there is a ~5–10 s window before the mesh for that topic GRAFTs into a stable state. Any governance op broadcast during that window by another member is lost to the joining node — not "arrives late", but **never arrives at all**. Recovery only happens on the next periodic namespace state heartbeat (~30 s), which exceeds the latency budget of every interactive caller that reads governance state.

Symptoms in practice:

- `list_group_members` returns empty on a freshly-joined namespace.
- `join_context` fails with `"identity is not a member of the group"` because the `MemberAdded` op for the caller was broadcast during the GRAFT window.
- Flaky e2e workflows like `group-subgroup-queued-deltas` require `type: wait, seconds: 5` steps + `max_attempts=2` retries to stabilize — neither is deterministic, both leak gossipsub timing into test DSL and user code paths.

## Root cause

Gossipsub decouples **subscription** from **mesh membership**:

1. `subscribe("ns/<id>")` announces interest to already-connected peers (propagates in ms).
2. Peer scoring accumulates over heartbeat ticks (default 1 s each).
3. GRAFT decisions — actually wiring us into the relay mesh — only happen on heartbeats, after ~4 cycles of scoring settles (`mesh_n_low=4` by default).

On slow CI runners or cold-boot nodes, this is 5–10 s. Any `publish()` during that window on the publisher side sees zero mesh peers for the topic and the message is effectively dropped — gossipsub has no per-peer backlog, no delivery ack.

This is **environmental** in the sense that it's a property of the gossipsub algorithm's mesh formation, but it's a **real architectural gap** for Calimero's usage, where causal chains like `inviter: add_member(X)` → `X: read_members()` are expected to converge within sub-second latency.

## What #2225 does and doesn't fix

PR #2225 fixes a **different** bug in the same neighborhood:

- **Bug A (fixed by #2225)**: op *did* arrive via gossip, but with missing parents (e.g. partial delivery during mesh warming — node received op C but missed its ancestors A and B). Now the node proactively backfills from the gossip source and iterates cross-peer via `ParentPullBudget`.
- **Bug B (this issue — NOT fixed by #2225)**: op never arrived at all. Node's `delta_store` is empty, not pending. No amount of parent-pull helps because there's nothing to backfill from.

[...]

PR #2245, opened 2026-04-24
perf(node): trim byte-array Debug prints in delta-store hot paths

chefsale:
## Problem

The post-Fix-3 flamegraph (kv-store fuzzy, node-1) shows ~1.6% of total CPU in `<u8 as Display>::fmt` under `tracing_core::field::ValueSet::record`. Two `tracing` sites in `delta_store.rs` are responsible — both Debug-format byte arrays, which expand to N × `u8::fmt` calls per log event.

### Hot offender #1: `unloadable_deltas = ?Vec<[u8; 32]>` at `delta_store.rs:649-655`

Fires every `perform_interval_sync` (~2s) while the mesh is bootstrapping (see #2236 cold-start race). Each warn expanded a Vec of 32-byte arrays to 32 × N `u8::fmt` invocations.

### Hot offender #2: per-delta `debug!` in `DeltaApplier::apply_delta`

Fires once per delta applied. Three byte-array Debug fields:
- `delta_id = ?delta.id` (32-byte array)
- `root_hash = ?outcome.root_hash` (same)
- `return_registers = ?outcome.returns` — this is `Result<Option<Vec<u8>>, _>`, which Debug-expands the **entire WASM return blob byte-by-byte**. Genuinely expensive when a delta produces non-trivial output.

## Fix

1. **`unloadable_deltas`** → `remaining_count` + `sample_unloadable` (first 3 IDs, hex-encoded once into a comma-joined string). Same debugging signal (correlate stuck deltas across nodes), a tiny fraction of the formatting work.
2. **Per-delta `apply_delta` debug!**:
   - `delta_id` → `%Hash::from(delta.id)` (Display, bs58 lazily into a stack buffer per #2242)
   - `root_hash` → `%outcome.root_hash` (it's already a `Hash`)
   - `return_registers` → `returns_ok` bool + `returns_len` — preserves "did it succeed, how much did it produce" without per-byte formatting

## Expected savings

~1.6% of total CPU, matching the `<u8 as Display>::fmt` aggregation from the flamegraph. Compounds with PR #2244 (prefix-iter + skip-applied rows).

## Plan context

- **PR #2244 (Stage 1)**: prefix-iter + skip-applied → cuts ~15pp of `load_persisted_deltas`'s 21%.
- **This PR (Stage 3.a)**: tracing cleanup → cuts ~1.6pp more.
- **PR-C (Stage 3.b, later)**: incremental DAG updates from `execute.rs` so the scan only runs on startup → eliminates the residual ~6%.

## Test plan

- [x] `cargo build -p calimero-node` clean
- [x] `cargo test -p calimero-node` — 398 pass, including the 303-test integration suite
- [ ] Flamegraph comparison on a successful CI perf run — expect `ValueSet::record` / `<u8 as Display>::fmt` to drop to near-zero in `load_persisted_deltas`

---

> [!NOTE]
> **Low Risk**
> Changes are limited to `tracing` fields/log structure to reduce per-delta/per-interval formatting overhead; no business logic, persistence, or merge semantics are altered.
> 
> **Overview**
> Reduces CPU overhead in `delta_store.rs` hot-path logging by switching several 32-byte IDs/hashes from Debug (`?`) to Display (`%`) via `Hash`, and by avoiding byte-by-byte formatting of WASM return payloads.
> 
> The per-delta WASM completion log now emits `returns_ok`/`returns_len` instead of `return_registers`, and the persisted-delta restore warning now logs only `remaining_count` plus a small base58 sample of unloadable delta IDs rather than Debug-printing the full ID list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit affbd059dbd565247f1c3d91337d22cf5a468d8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/delta_store.rs
+++ b/crates/node/src/delta_store.rs
@@ -130,9 +130,9 @@ impl DeltaApplier<Vec<Action>> for ContextStorageApplier {
         if is_merge_scenario {
             info!(
                 context_id = %self.context_id,
-                delta_id = ?delta.id,
-                current_root_hash = ?Hash::from(current_root_hash),
-                delta_expected_hash = ?Hash::from(delta.expected_root_hash),
+                delta_id = %Hash::from(delta.id),
+                current_root_hash = %Hash::from(current_root_hash),
+                delta_expected_hash = %Hash::from(delta.expected_root_hash),
                 "Concurrent branch detected - applying with CRDT merge semantics"
             );
         }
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed in affbd059.

**cursor — encoding consistency**: `sample_unloadable` now uses bs58 via `Hash::from(*id).to_base58()` instead of hex, so operators cross-referencing a stuck delta from the warn against other logs see the same format.

**meroreviewer — `DELTA_APPLY_TIMING` info!**: swept `delta_id = ?delta.id` / `final_root_hash = ?computed_hash` → `%Hash::from(delta.id)` / `%computed_hash`. Good catch — `info!` is more likely enabled in production than `debug!`, so this site is arguably higher-impact than the original one.

**meroreviewer — merge-detection info!/debug!**: same treatment at the three sites (`is_merge_scenario`, the merge vs non-merge hash-mismatch branches). All byte-array Debug fields now use Display via `Hash::from`.

**meroreviewer — lazy computation**: wrapped the `returns_ok`/`returns_len` pattern-match plus the `debug!` call in `if tracing::enabled!(tracing::Level::DEBUG) { ... }`, matching the convention used for `action_types` just above.

`cargo test -p calimero-node` — 398 pass, including the 303-test integration suite.

PR #2248, opened 2026-04-24
perf(node): incremental DAG updates from execute, drop per-sync rescan

chefsale:
## Problem

`DeltaStore::load_persisted_deltas` fired on every interval sync (~2s cadence per context) because `execute.rs` wrote local deltas straight to RocksDB without touching the in-memory DAG, and sync needed the DAG to know what to offer peers.

Stage 1 (#2244) cut most of the scan cost by seeking by context prefix and skipping already-applied rows (21% → ~6% CPU). This PR closes the remaining residual by keeping the DAG in sync with writes at the source — after this, `load_persisted_deltas` only runs at startup as the crash-recovery path.

## How

Minimal cross-crate plumbing that avoids dragging `calimero-dag` into `calimero-node-primitives`:

### New types & API
- `calimero_node_primitives::client::LocalAppliedDelta` — plain struct carrying what's needed to register a delta in the in-memory DAG (`Vec<Action>`, `[u8; 32]` parents, HLC, expected root hash).
- `NodeClient` gains `notify_local_applied_delta(&self, LocalAppliedDelta)` — non-blocking `try_send` on an internal mpsc channel. Best-effort: on channel closure the next process restart's `load_persisted_deltas` catches up.

### Runtime wiring (`run.rs`)
Creates the channel, passes the sender into `NodeClient::new`, spawns a tokio task that drains the receiver and calls `DeltaStore::add_local_applied_delta` via the `delta_stores_handle()`.

### New DeltaStore method
`add_local_applied_delta(CausalDelta)` — short-circuits on `is_applied` (handles the benign race where the same delta arrives via sync concurrently), updates `head_root_hashes` + `parent_hashes`, and calls `restore_applied_delta` to add topology without re-applying. Equivalent to one row of `load_persisted_deltas`'s main loop.

### Write-site hook-ups
- `execute.rs:~1200` — after `handle.put(&ContextDagDelta, ...)`, fire `notify_local_applied_delta(...)`.
- `create_context.rs:~412` — same for the genesis delta. `init_delta` now carries `(types::ContextDagDelta, Vec<Action>)` so the actions are available for the notify without having to borsh-deserialize them back out of the stored blob.

### Sync hot-path cleanup
Removed the three `delta_store.load_persisted_deltas().await` calls on interval-sync code paths in `sync/manager/mod.rs`. The DAG is now kept current incrementally; the rescan still runs at startup (`state_delta/mod.rs:603, 1380`) as the crash-recovery path.

## Correctness

Invariant preserved: the DB is always the source of truth. The DAG is kept up-to-date incrementally for perf. If the notify is dropped (channel full/closed), the next startup scan picks it up. Event-harvesting contract from Stage 1 is unaffected — the harvesting site still runs from `load_persisted_deltas`, which is still invoked on startup.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo test -p calimero-node -p calimero-context -p calimero-node-primitives` — 311 + 303-test node integration suite green
- [ ] CI perf run on this branch: expect `load_persisted_deltas` inclusive to drop from ~6% to near-zero in the steady-state sync loop, with the function only appearing under the startup `state_delta::init_delta_store` path

---

> [!NOTE]
> **Medium Risk**
> Changes the sync/DAG update pipeline by introducing a new notification channel and new `DeltaStore` write path; bugs could leave the in-memory DAG or `dag_heads` temporarily inconsistent until restart, impacting delta sync behavior.
> 
> **Overview**
> Removes the steady-state dependency on `DeltaStore::load_persisted_deltas` during interval sync by **incrementally updating the node’s in-memory DAG** whenever a locally-created delta is persisted.
> 
> Adds `LocalAppliedDelta` + `NodeClient::notify_local_applied_delta` (best-effort `try_send`) and wires a drainer task in `run.rs` that converts notifications into `DeltaStore::add_local_applied_delta` calls. The execute and context-creation paths now emit these notifications after writing deltas to RocksDB (including the genesis delta, carrying actions alongside the persisted row).
> 
> Tightens correctness around pending/cascade handling: `add_local_applied_delta` explicitly triggers `try_process_pending`, prunes `head_root_hashes` to current heads, and persists cascaded children’s DB state + updated `dag_heads`; `load_persisted_deltas` no longer early-returns when no new rows were collected. Sync manager paths now hydrate from DB *only when a `DeltaStore` is newly created*, rather than on every sync. Includes new regression tests in `calimero-dag` and `calimero-node`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e91d6a47bfcd29cbc0791ca152dd2a5c1cf95a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed in adda4848 + folded in the try_process_pending regression fix that today's failing perf run (24879931025) exposed.

**cursor (high) — sync-created DeltaStore never loads persisted deltas.** Right, I dropped the wrong rescan site. Warm stores don't need the rescan (execute-side notify keeps them current), but *cold* stores — when sync is the first path to create the DeltaStore for a context — needed the one-shot load. Added an `is_new` flag at each of the three `delta_stores.entry(...).or_insert_with(...)` sites; if the closure ran, call `load_persisted_deltas()` once. `state_delta`'s `is_new_store` branch was never going to run in that case because sync already inserted the Occupied entry, so this is the right spot.

**Bonus: unconditional `try_process_pending`.** Today's perf run (24879931025, commit 88857897) failed in 2 minutes with node-4 stuck on a pre-seed root hash. Traced to a regression from PR #2244 I'd overlooked: `load_persisted_deltas` has an early-return when `all_deltas` is empty, which skipped the `try_process_pending` call at the end. Harmless pre-#2244 because every in-context DB row was unconditionally collected (so any non-empty context made `all_deltas` non-empty). After #2244's skip-if-already-in-DAG, `all_deltas` is empty whenever a warmed-up node scans a context whose rows are all in the DAG — the steady state. The early return then permanently prevents pending deltas (received via gossip before parents) from being retried, stranding late joiners. Dropped the early return; step 2's while loop is a no-op on empty and the warn is already guarded.

This second fix is arguably a master-level bug-fix independent of this PR, but it's small enough and directly caused a test failure that's gating perf measurement, so folded in.

Local `cargo test -p calimero-node` green (303-test integration suite + rest).

chefsale:
Addressed the second cursor concern (missing `try_process_pending` after `restore_applied_delta`) in `feac324e`:

- `DeltaStore::add_local_applied_delta` now calls `DagStore::try_process_pending` immediately after a successful `restore_applied_delta`. Any child deltas that landed before the local parent (e.g. because the remote parent arrived via sync while the local branch was still executing) cascade in the same pass rather than waiting for the next interval sync.
- Had to thread `Send` through the DAG crate: `apply_delta` now returns `Pin<Box<dyn Future + Send + 'a>>`, and `apply_delta` / `apply_pending` / `add_delta` / `add_delta_with_outcome` / `try_process_pending` require `A: DeltaApplier<T> + Sync` and `T: Send + Sync`. Needed because the drainer task in `run.rs` is `tokio::spawn`'d.
- Added regression test `test_restore_applied_delta_plus_try_process_pending_cascades_child` in `crates/dag` that documents the contract: `restore_applied_delta` is topology-only, callers must nudge via `try_process_pending` to drain unblocked children.

All DAG + node integration tests pass locally.

chefsale:
Addressed cursor's third concern (stale `head_root_hashes` entries from `add_local_applied_delta`) in `7ad15639`:

- After `restore_applied_delta` + `try_process_pending`, I now capture `dag.get_heads()` under the same write lock and `retain(...)` down to only the current DAG heads. Mirrors what `add_delta_internal` does at the end of its own flow (line ~1063). Without it, every ancestor accumulated in the map over the DeltaStore's lifetime, and head-state lookups (merge-detection heuristic + peer queries) could return a non-head's root hash.
- Added a regression test (`crates/node/src/delta_store_head_hashes_test.rs`) that registers a parent, asserts it's the sole entry, then registers a child and asserts the parent is pruned.
- The test also exercises the no-pending path of `try_process_pending`, giving us a light smoke test of the cascade nudge from the previous commit.

All 54 `calimero-node` lib tests pass locally.

chefsale:
Addressing @meroreviewer's three comments in `2e91d6a4`.

**🔴 Cascaded deltas lose DB `applied: true` + events + dag_heads.** Real bug, fixed. `add_local_applied_delta` now tracks `pending_before` / `pending_after` around `try_process_pending` under the same write-lock scope, computes cascaded_ids, clones the bodies, and calls `persist_cascaded_deltas_and_update_heads` — the same helper `add_delta_internal` uses. DB records flip to `applied: true` and `dag_heads` is refreshed. Return type is now `Result<Vec<([u8; 32], Vec<u8>)>>` carrying cascaded events.

For immediate handler dispatch: today the drainer doesn't have a line into `NodeClients` / `NodeManager`, so it logs at info (`cascaded_count` + "awaiting restart replay for handler execution") and relies on the existing #2185 restart-replay contract (records stay `applied: true, events: Some(..)`, surfaced by next `load_persisted_deltas` via `pending_handler_events`). Events are preserved, not lost. Plumbing drainer → handler dispatch is tracked as a follow-up rather than expanded here.

**🟡 `head_root_hashes` not populated for cascaded children that become new heads.** This is pre-existing behavior (also true for `add_delta_internal`). The prune keeps the map consistent with actual heads after the cascade — entries that aren't in `dag.get_heads()` are removed. Children not yet inserted aren't added by the prune; they'd need a separate pass. Out of scope for the current hotfix; will file a cleanup task if the merge-detection heuristic turns out to rely on it.

**🟡 Drainer drops notifications for contexts without a DeltaStore.** Not a correctness issue in practice: `execute.rs` persists the delta to RocksDB *before* calling `notify_local_applied_delta` (see `context/src/handlers/execute/mod.rs` line 1199–1210). Peers querying via `handle_delta_request` hit RocksDB first (see `sync/delta_request.rs` line 263) and only fall through to `DeltaStore` when the row is missing. So when the DeltaStore doesn't yet exist, the delta is still reachable via the DB path, and the eventual `is_new`-gated `load_persisted_deltas` hydrates the in-memory DAG. Kept the debug log for observability.

Issue #2237, opened 2026-04-23
governance: three-phase contract — mesh-ready gate, ack-confirmed publish, typed delivery report


chefsale:
## Summary

Every governance endpoint today uses fire-and-forget gossipsub publish: the caller returns `Ok` once the message has been handed to the local gossipsub instance, with no guarantee any recipient actually received it. This is the root cause of the cold-start flakes (#2198), the reason PR #2225 needed a parent-pull retry loop, and the reason our e2e workflows are sprinkled with `type: wait, seconds: 5` and `max_attempts=2` retries that work around — but don't fix — the underlying problem.

This issue proposes a uniform three-phase contract for every governance endpoint: readiness-gate → publish-and-collect-acks → typed delivery report. It replaces fire-and-forget with acked broadcast, which is gossipsub-native and collapses most of the current scaffolding.

## The design flaw this replaces

Governance ops are causal: `MemberAdded(Bob)` must be visible on at least one recipient before any read that depends on Bob's membership (like `list_group_members` or `join_context`) can succeed. Today:

- Publisher calls `gossipsub.publish(topic, op)`, returns `Ok` immediately.
- Message goes only to peers in the publisher's current mesh for that topic.
- On a freshly-joined subscriber, the publisher's mesh may not include them yet (5–10 s for GRAFT on cold start).
- Message silently disappears. No one has an error to handle.
- Recovery only via `NamespaceStateHeartbeat` at ~30 s, which exceeds any interactive latency budget.

This isn't a mesh-timing bug; it's a delivery-semantics bug. The transport is correct; the API contract is wrong.

## The contract

Every governance endpoint runs three phases:

### Phase 1 — Readiness gate

Before doing anything, check the gossipsub mesh for the relevant topic:

```rust
let mesh = gossipsub.mesh_peers(topic);
let threshold = min(mesh_n_low, known_subscribers_on_topic);
if mesh.len() < threshold {
    return Err(ApiError::NamespaceNotReady);
}
```

Not warm → return immediately. Client retries. No time wasted publishing into the void.

### Phase 2 — Publish and collect acks

Apply the op locally, publish, collect `Ack` messages from recipients:

```rust
apply_to_local_dag(&op)?;
gossipsub.publish(topic, &op)?;

let acks = collect_acks_until(hash(&op), deadline, min_acks=1).await;
```

Any member who receives and successfully applies the op publishes a signed `Ack { op_hash, signer_pubkey, signature }` on the topic. Publisher collects them.

[...]

PR #2251, opened 2026-04-24
ci(fuzzy): settle KeyDelivery gossip before handlers test starts

chefsale:
## Summary

Short-term mitigation for a 100% repro of handler-test failure observed in fuzzy run [24891646958](https://github.com/calimero-network/core/actions/runs/24891646958). Adds a 30 s wait between mesh formation and the fuzzy phase, and bumps \`wait_for_sync\` timeout from 60 s to 120 s.

## Root cause (not a regression from this PR family)

Every request on \`fuzzy-handlers-node-4\` returned \`HandlerError(InternalError)\` within ~77µs of receipt — before WASM runs. Traced to \`crates/context/src/handlers/execute/mod.rs:180\`, the silent branch where the context is known to be in a group (\`get_group_for_context\` → \`Some(gid)\`), the current group key isn't loaded yet (\`load_current_group_key\` → \`None\`/\`Err\`), and \`identity.sender_key\` is \`None\`. Both fallbacks miss → \`InternalError\` with no log.

Timeline on node-4:
- \`T0 = 13:23:13\`: context identity generated
- \`T0 + 2 s\`: \"member joined group via direct request-response\"
- \`T0 + 2.01 s\`: node-4 publishes its own KeyDelivery
- \`T0 + 3 s\`: \`wait_for_sync\` passes (DAG root parity reached)
- \`T0 + 3 s\`: fuzzy test fires first \`set\` → fails, every subsequent request fails
- 356/356 requests failed; other nodes 0/876.

\`wait_for_sync\` only checks DAG root-hash convergence. It doesn't prove that every new member has received the **group's sender key** via KeyDelivery gossip. Under the faster mesh-formation timings after PR #2244 onward, the window between \"DAG in sync\" and \"group key delivered\" widened enough to repro 100% on the last-joined node.

## Why this is a workflow-side patch

The real fix belongs to #2237 (\"three-phase contract: readiness-gate → ack-confirmed publish → typed delivery report\"). That issue currently scopes governance endpoints; KeyDelivery is the same fire-and-forget pattern and should be the next consumer once the contract lands. See my comment on #2237.

This PR just buys the handlers test a reliable green until #2237 covers KeyDelivery.

## Test plan

- [ ] CI: fuzzy-load-test passes on both kv-store and kv-store-with-handlers variants
- [ ] Run is green end-to-end (45 min handlers test completes)
- [ ] Track remove-this-wait follow-up in #2237

---

> [!NOTE]
> **Low Risk**
> Low risk: CI workflow-only timing changes that may slightly increase test runtime but don’t affect production code paths.
> 
> **Overview**
> Improves reliability of the `kv-store-with-handlers` fuzzy workflow by increasing the `wait_for_sync` timeout from 60s to 120s and inserting a 30s `wait` before the fuzzy phase to allow KeyDelivery gossip to propagate.
> 
> Adds explicit comments documenting the mitigation and linking it to the underlying delivery-guarantee gap tracked in `#2237`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01a532e5ee5357d585dedffadd3dd9487a0fe06f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/workflows/fuzzy-tests/kv-store-with-handlers/fuzzy-test.yml
+++ b/workflows/fuzzy-tests/kv-store-with-handlers/fuzzy-test.yml
@@ -45,9 +45,25 @@ steps:
       - fuzzy-handlers-node-2
       - fuzzy-handlers-node-3
       - fuzzy-handlers-node-4
-    timeout: 60
+    timeout: 120
     check_interval: 2
     trigger_sync: true
+
+  # Temporary mitigation for #2237 (fire-and-forget gossip → no delivery guarantee).
+  # `wait_for_sync` only proves the DAG root-hash converged; it does NOT prove
+  # that KeyDelivery messages reached every member. On fast mesh formation the
+  # last joiner can reach root-hash parity before it has received the group's
+  # sender key, causing every execute to fail at
+  # `context/src/handlers/execute/mod.rs:180` (silent InternalError, "group
+  # known but group key not yet loaded AND identity.sender_key == None").
+  # Observed as 100% failure on fuzzy-handlers-node-4 in run 24891646958.
+  #
+  # 30 s is conservative: gossipsub GRAFT + key-delivery round-trip fits well
+  # under that on a healthy mesh. Remove once #2237's three-phase contract
+  # (readiness-gate + ack-confirmed publish) covers KeyDelivery.
+  - name: Settle KeyDelivery gossip before fuzzy start
+    type: wait
+    seconds: 30
 
   - name: Verify handlers mesh synced
     type: assert


chefsale:
Closing — the 30 s wait is based on a wrong diagnosis.

A fresh kv-store fuzzy run (no handlers) shows the same \"last-joined node never catches up\" symptom, this time with node-4 stuck at the **pre-seed root hash** for 30 s while nodes 2 and 3 converged within 2 s:

\`\`\`
Attempt 2 (2.2s): HsfiB7... fuzzy-kv-node-1, 2, 3
                  14Hm8p... fuzzy-kv-node-4
... 14 more attempts, node-4 stuck ...
Attempt 15 (30.0s): same
✗ Sync verification failed after 31.99s (15 attempts)
\`\`\`

That's not a 30-seconds-of-patience problem — node-4 is **structurally not receiving deltas**. A longer wait cannot fix structural mesh/membership rejection; the \`unknown context member after governance sync, closing stream\` warnings we saw earlier are the actual smoking gun.

This affects kv-store too (no handlers, so my 30 s wait in the handlers workflow wouldn't even be exercised on that path), so a test-side mitigation only in \`kv-store-with-handlers\` isn't even the right shape.

Folded the kv-store data point into #2237, which is where the real fix lives.

PR #2269, opened 2026-04-28
feat(governance): readiness FSM + cache + beacon manager + J6 join (#2237)

rtb-12:
## Summary

Phases 6, 7 and 8 of the three-phase governance contract permanent fix
for #2237. Lands the per-namespace **readiness beacon** infrastructure
that closes the cold-start gap between joining a namespace and being
able to publish governance ops without losing them to gossipsub's
GRAFT handshake.

**Status: ready for review.** All implementation phases complete, all
review-bot findings + my own `/code-review:code-review` findings
resolved.

### Phase 6 — Readiness FSM + Cache (pure types)

- `crates/node/src/readiness.rs` — `ReadinessTier` (data-carrying
  variants), `ReadinessState`, `ReadinessConfig`, `PeerSummary`,
  `evaluate_readiness` total transition fn.
- `ReadinessCache` with monotonic `ts_millis` filter,
  `MAX_BEACON_CLOCK_DRIFT_MS = 60s` cache-poisoning guard,
  opportunistic eviction (`2 × drift` window), atomic `peer_summary`.
  Uses `BTreeMap` because `PublicKey` derives `Ord` not `Hash` (plan
  deviation, documented).

### Phase 7 — `ReadinessManager` actor + receiver dispatch

- `ReadinessManager` actix actor: periodic beacon emission (filtered
  to `*Ready` tiers), edge-trigger on tier transitions, probe-response
  with `BEACON_INTERVAL / 2` rate limit per (peer, namespace).
- `crates/context/src/governance_broadcast.rs` — `verify_readiness_beacon`
  (sig via canonical `signable_bytes()` from #2260 + member-set check).
- `crates/node/src/handlers/network_event/readiness.rs` — receiver-side
  dispatch wired into `network_event/namespace.rs`. Beacons go straight
  into the cache (bypassing the actor mailbox) then `notify`+`do_send`
  the manager.

### Phase 8 — J6 namespace-join flow

- `crates/node/src/join_namespace.rs` (new module):
  - `join_namespace` (J6 fast path) — provision identity + write minimal
    `GroupMetaValue { admin_identity = inviter, ... }` as a beacon-trust
    seed → subscribe → publish `ReadinessProbe` → `await_first_fresh_beacon`.
  - `await_namespace_ready` (J6 slow path) — `sync_namespace` backfill
    + 3s mesh-wait poll → publish `RootOp::MemberJoined` through the
    three-phase contract.
  - `join_and_wait_ready` — convenience aggregator with
[...]

PR #2271, opened 2026-04-29
fix(governance): post-#2270 KeyDelivery hardening + FSM-observation invariant + fuzzy coverage (#2237)

rtb-12:
## Summary

Follow-up to #2270 closing the remaining items in the three-phase governance contract for issue #2237. Five logical commits, ~630 LOC, no flag gating — all changes either bug fixes, plan-deviation restorations, or additive new infra.

## Commits (review-friendly order)

1. `e6d5f30` — **typed KeyDelivery error + soft-fail wait + config lift**
   - Soft-fail `load_current_group_key` in the wait loop instead of `?`-propagating transient store errors.
   - Convert `KEY_DELIVERY_FALLBACK_WAIT` deadline branch from silent `Ok-no-key` to typed `Err`. Restores PLAN-2237's caller contract.
   - Lift the magic 5s constant into `ContextManagerConfig::key_delivery_fallback_wait`.

2. `1978e05` — **`ObserveDelivery` extension trait**
   - Pure refactor — collapses 15 identical-shape `if let Some(report) = report.as_ref() { observe_handler_delivery(...); }` call sites across 12 context handlers + 3 server admin handlers into `report.observe("h", "k")`. Existing free function preserved.

3. `dcd2a7c` — **notify ReadinessFSM on every monotonic apply**
   - Closes the gap where the FSM's `state_per_namespace` was only filled from the gossipsub-receive path. A node that *publishes* an op or catches up via sync/backfill/join never observed its own monotonic advance and stayed at `Bootstrapping` forever.
   - Adds `NodeMessage::ForwardNamespaceOpApplied` + `NodeClient::notify_namespace_op_applied` plumbing so call sites in `crates/context` can signal the readiness actor across the crate boundary.
   - Wired into 5 apply paths: 2 publisher-side (`sign_apply_and_publish`, `publish_post_gate`), 3 receive-side (`fetch_and_apply_namespace_backfill`, sync catch-up, sync backfill response), and 1 join-response (`join_group` governance ops loop). Each gated on `NamespaceApplyOutcome::Applied`.

4. `06867ff` — **typed `ExecuteError::GroupKeyPending` variant**
   - Distinguish the transiently retry-able \"group key not yet delivered\" condition from `InternalError`. Wire-additive (`ExecuteError` is already `#[non_exhaustive]`); flows through the JSON-RPC envelope's untagged variant transparently.

5. `b45ee13` — **fuzzy load workflow for group governance**
   - New `workflows/fuzzy-tests/group-governance/fuzzy-test.yml` — third merobox fuzzy test alongside kv-store and kv-store-with-handlers. 4-node namespace, 15 min default (60 min nightly). Every data write traverses the governance topic, exercising `publish_and_await_ack` + readiness FSM + per-namespace ack collection under load. Phase 4 post-load governance churn (capability changes, subgroup creation, member add/remove) validates the system on a hot mesh.
   - Wires into `.github/workflows/fuzzy-load-test.yml` as a third matrix job.

## Verification

- 3 local code reviews ran across the phases (`superpowers:code-reviewer`); each caught issues the previous didn't and they're all addressed.
- `cargo check --workspace` clean.
- `cargo fmt --check` clean.

## Test plan

- [ ] CI: e2e-rust-apps matrix (existing 19 workflows) still green.
- [ ] CI: new `fuzzy-test-group-governance` job runs on this PR (path trigger covers `workflows/fuzzy-tests/**`).
- [ ] Manual: verify `governance_handler_delivery_total{outcome=\"empty\"}` rate stays approximately zero on master after merge.
- [ ] Operator follow-up: dashboard the new metric + add `outcome=\"empty\"` to the cold-start alerting runbook (out of scope for this PR).

## Out of scope

- Operator runbook entry for `outcome=\"empty\"` rate.
- Grafana dashboard for `governance_handler_delivery_*` metrics.
- Drop the `merobox stop --all || true` belt-and-suspenders cleanup in `e2e-rust-apps.yml:280-281` after this PR's fuzzy test runs clean for 2+ weeks.
- Native governance-op weighting in merobox's `fuzzy_test` step (tracked as a calimero-network/merobox issue).

---

> [!NOTE]
> **Medium Risk**
> Touches governance join/execute error handling and node readiness signaling across multiple async paths; regressions could affect group joins, execute encryption, or readiness state, though changes are scoped and largely additive/plumbing.
> 
> **Overview**
> **Hardens group governance correctness and observability across join/publish/execute paths.** `join_group` now uses a runtime-configured KeyDelivery fallback wait, treats transient key-store read failures as retryable, and fails the join on KeyDelivery timeout/channel-close instead of silently succeeding without a key.
> 
> **Restores and clarifies retryable client semantics.** Adds `ExecuteError::GroupKeyPending` and returns it when executing in a group context before the group key has arrived, rather than reporting a generic `InternalError`.
> 
> **Ensures the readiness FSM observes publisher-side DAG advances.** Introduces `NodeClient::notify_namespace_op_applied` + `NodeMessage::ForwardNamespaceOpApplied`, and calls it from publisher paths (`namespace_governance`), join-response apply, and sync/backfill/catch-up apply loops (gated on `NamespaceApplyOutcome::Applied`).
> 
> **Improves delivery metrics ergonomics and CI coverage.** Adds `ObserveDelivery` to unify handler delivery reporting, and wires a new `group-governance` fuzzy load test (new merobox workflow + new GH Actions job and summary reporting) into `fuzzy-load-test.yml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e8305815a11f9cdd5a3a2372276a2b5438a33fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rtb-12:
## CI status update + bot comment triage

### Failures addressed in `9aa8bda`
- **PR title** renamed `refine` → `fix` (lint failure resolved — `refine` isn't an allowed conventional-commit type).
- **Group Governance Fuzzy Test** — the fuzzy_test phase itself ran clean (0 % failure rate during 15 min sustained load). Failure was on Phase 5's `Final mesh sync` step at 180s. The post-load convergence window needs more than the kv-store-only baseline because every `set` traverses the governance topic *and* Phase 4 piles on capability/subgroup/membership churn after the load. Bumped timeout 180s → 600s with rationale documented inline.

### Pre-existing on master (NOT introduced by this PR)
- **KV Store with Handlers Fuzzy Test** failure (5.2% cross-node assertion failures > 95% success threshold). Verified via run [25085077153](https://github.com/calimero-network/core/actions/runs/25085077153) — last nightly soak on master 13 hours before this PR was opened. Same failure mode. Out of scope here.

### Bot comment triage
- **3× high-entropy false positives** — same docker SHA pattern dismissed on #2269 and #2270. Recurring noise from the `ai-reviewer` against artifact SHAs in `Cargo.lock` and CI YAML. Closing as duplicate.
- **🟡 `notify_namespace_op_applied` `try_send` drop concern** ([thread](https://github.com/calimero-network/core/pull/2271#discussion_r-)). Valid theoretical concern but the documented mitigation is in place: the warn log on saturation gives operators visibility, and the FSM reconciles via the next applied op or peer beacon (every namespace produces both periodically). Coalescing/debouncing would be an architectural change worth a separate issue if the warn log fires in practice — for now, the `governance_handler_delivery_total{outcome="empty"}` metric already gives ops the leading indicator they need.

.github/workflows/fuzzy-load-test.yml
```diff
@@ -922,9 +923,256 @@ jobs:
...
+          mkdir -p fuzzy-test-logs/group-governance
+          mkdir -p profiling-data/group-governance
+          mkdir -p profiling-reports/group-governance
+
+      - name: Determine image to use
+        id: image
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          chmod +x scripts/determine-image-tag.sh
```
frdomovic, line 975: 🟡 **fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431 — metric collection collision on concurrent runs**

Both fuzzy-test-scaffolding-e2e and fuzzy-test-group-governance configure vmagent on port 8431. When both jobs run concurrently (the default when test_app == 'all'), two vmagent instances on the same runner will compete for port 8431. The setup-vmagent step will fail or silently bind to a different port, causing one job's metrics to be lost or mixed with the other's. The kv-store job uses 8429, handlers uses 8430, and both scaffolding-e2e and group-governance use 8431.

**Suggested fix:**
```
Assign a unique port to the group-governance job, e.g., 8432. Update the setup-vmagent-gov step and the run-vmagent.sh invocation to use 8432 instead of 8431.
```

.github/workflows/fuzzy-load-test.yml
```diff
@@ -922,9 +923,256 @@ jobs:
...
+              "${{ github.token }}" \
+              --profiling
+            echo "profiling=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Pull Docker image
+        run: |
+          IMAGE="ghcr.io/calimero-network/merod:${{ steps.image.outputs.tag }}"
+          docker pull "$IMAGE" || true
+
```
frdomovic, line 1001: 🟡 **Port conflict: group-governance vmagent uses same port as scaffolding-e2e**

The `fuzzy-test-group-governance` job configures vmagent on port `8431` (see `setup-vmagent-gov` step), which is the same port used by `fuzzy-test-scaffolding-e2e`. When both jobs run concurrently on the same `big-machine` runner, the second job to start will fail to bind the port, causing vmagent setup to fail or silently collect no metrics. The kv-store and handlers jobs use ports 8429 and 8430 respectively, establishing a clear pattern that each job needs a unique port.

**Suggested fix:**
```
Assign a unique port to the group-governance job, e.g. `8432`, in both the `setup-vmagent-gov` step and the `run-vmagent.sh` invocation in the `gov-test` step.
```

.github/workflows/fuzzy-load-test.yml
```diff
@@ -922,9 +923,256 @@ jobs:
...
+              "${{ github.head_ref || github.ref_name }}" \
+              "${{ github.token }}" \
+              --profiling
+            echo "profiling=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Pull Docker image
+        run: |
+          IMAGE="ghcr.io/calimero-network/merod:${{ steps.image.outputs.tag }}"
+          docker pull "$IMAGE" || true
```
frdomovic, line 1000: 🟡 **Port 8431 reused between scaffolding-e2e and group-governance vmagent jobs**

Both `fuzzy-test-scaffolding-e2e` and `fuzzy-test-group-governance` configure their vmagent on port `8431`. When both jobs run concurrently on the same `big-machine` runner (which is possible since they are parallel matrix jobs with no mutual exclusion), the second job's `setup-vmagent.sh` call will either fail to bind or silently shadow the first job's metrics collection. The kv-store job uses 8429, handlers uses 8430, but both e2e and governance use 8431.

**Suggested fix:**
```
Assign a unique port to the group-governance vmagent, e.g. `8432`, consistent with the incrementing pattern used by the other three jobs.
```

workflows/fuzzy-tests/group-governance/fuzzy-test.yml
```diff
@@ -0,0 +1,361 @@
...
+          - type: assert
+            non_blocking: true
+            statements:
+              - "is_set({{set_result}})"
+          - type: wait
+            seconds: 1
+          - type: call
+            node: "{{random_node}}"
+            context_id: "{{group_ctx_id}}"
+            method: get
```
frdomovic, line 201: 💡 **fuzzy_test operation references {{fuzzy_key}} and {{fuzzy_value}} without prior definition**

The `set_and_verify` and `cross_node_write_read` operations reference `{{fuzzy_key}}` and `{{fuzzy_value}}` in their `get` steps and assertions, but these variables are never explicitly set in the workflow. The `set` step uses inline `key: "gkey_{{random_int(1, 10000)}}"` and `value: "gval_{{uuid}}_{{timestamp}}"` literals. Whether `{{fuzzy_key}}` and `{{fuzzy_value}}` are automatically populated by merobox from the preceding `set` step's arguments is an implicit merobox convention that is not documented in this file. If merobox does not support this implicit binding, the `get` and `assert` steps will always use empty/undefined values, making the cross-node consistency check a no-op.

**Suggested fix:**
```
Add explicit `outputs` bindings in the `set` steps to capture the key and value used (e.g. `outputs: { fuzzy_key: key, fuzzy_value: value }`), or add a comment referencing the merobox documentation that explains the `{{fuzzy_key}}`/`{{fuzzy_value}}` implicit binding convention.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 3 agents** | Quality score: 42% | Review time: 126.6s

🟡 3 warnings, 💡 1 suggestions. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-02a887cd`</sub>

.github/workflows/fuzzy-load-test.yml
```diff
@@ -922,9 +923,256 @@ jobs:
...
+
+      - name: Setup vmagent for metrics collection
+        id: setup-vmagent-gov
```
frdomovic, line 1003: 🟡 **Port conflict: group-governance and scaffolding-e2e both use vmagent port 8431**

Both `fuzzy-test-scaffolding-e2e` and `fuzzy-test-group-governance` configure vmagent on port 8431. When both jobs run on the same `big-machine` runner (they are parallel matrix jobs with `runs-on: big-machine`), the second job to start will fail to bind the port, causing silent metrics collection failure or a port-already-in-use error. The scaffolding-e2e job uses 8431 (visible in the existing workflow) and the new group-governance job also hardcodes 8431.

**Suggested fix:**
```
Assign a unique port to the group-governance vmagent, e.g. 8432, consistent with the pattern used by kv-store (8429), handlers (8430), and scaffolding-e2e (8431).
```

.github/workflows/fuzzy-load-test.yml
frdomovic, line 1000: 🟡 **fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431**

Both fuzzy-test-scaffolding-e2e and fuzzy-test-group-governance configure vmagent on port 8431. When both jobs run in parallel on the same runner (or if the runner is reused), the second job's vmagent setup will fail or silently conflict with the first. The kv-store job uses 8429, handlers uses 8430, but both e2e and group-governance use 8431.

**Suggested fix:**
```
Assign a unique port to the group-governance job, e.g. 8432: ./scripts/setup-vmagent.sh 'group-governance' ... '8432' ... and update the run-vmagent.sh call accordingly.
```

frdomovic, line 1000: 🟡 **fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431**

Both `fuzzy-test-scaffolding-e2e` and `fuzzy-test-group-governance` configure vmagent on port `8431`. Since both jobs run in parallel on the same `big-machine` runner (they both `needs: build` with no mutual dependency), the second job to start will fail to bind port 8431 or silently shadow the first job's metrics collection. The existing jobs use distinct ports: kv-store=8429, handlers=8430, scaffolding-e2e=8431. The new group-governance job should use 8432.

**Suggested fix:**
```
Change the vmagent port for `fuzzy-test-group-governance` from `8431` to `8432` in both the `setup-vmagent-gov` step and the `run-vmagent.sh` invocation in `gov-test`.
```

workflows/fuzzy-tests/group-governance/fuzzy-test.yml
```diff
@@ -0,0 +1,361 @@
...
+            node: "{{random_node}}"
+            context_id: "{{group_ctx_id}}"
+            method: entries
+            outputs:
+              entries_result: result
+          - type: assert
+            non_blocking: true
+            statements:
+              - "is_set({{entries_result}})"
+
```
frdomovic, line 245: 💡 **cross_node_write_read operation references undefined fuzzy_key and fuzzy_value**

Same issue as the set_and_verify operation: the `cross_node_write_read` operation writes with inline key `consistency_{{uuid}}` and value `check_{{timestamp}}` but then reads using `{{fuzzy_key}}` and asserts `contains({{consistency_result}}, {{fuzzy_value}})`. These template variables are undefined in this operation's scope and will not resolve to the values written in the preceding set step.

**Suggested fix:**
```
Capture the key and value from the set step into named outputs and reference those outputs in the get step, or use the same literal template expressions.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 3 agents** | Quality score: 46% | Review time: 110.0s

✅ 4 fixed | 🆕 4 new. See inline comments.
*1 low-severity finding suppressed on recently-fixed code.*

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-252ad0c6`</sub>

.github/workflows/fuzzy-load-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/fuzzy-load-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/fuzzy-load-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

workflows/fuzzy-tests/group-governance/fuzzy-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 3 agents** | Quality score: 57% | Review time: 121.0s

### 🟡 Issues pending resolution

✅ **8 Fixed** | 🆕 **12 New**

---

### ✅ Fixed Issues

<details>
<summary>The following issues from previous reviews have been addressed:</summary>

1. ~~fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431 — metric collection collision on concurrent runs~~ (`.github/workflows/fuzzy-load-test.yml:975`)
2. ~~Port conflict: group-governance vmagent uses same port as scaffolding-e2e~~ (`.github/workflows/fuzzy-load-test.yml:1001`)
3. ~~Port 8431 reused between scaffolding-e2e and group-governance vmagent jobs~~ (`.github/workflows/fuzzy-load-test.yml:1000`)
4. ~~fuzzy_test operation references {{fuzzy_key}} and {{fuzzy_value}} without prior definition~~ (`workflows/fuzzy-tests/group-governance/fuzzy-test.yml:201`)
5. ~~Port conflict: group-governance and scaffolding-e2e both use vmagent port 8431~~ (`.github/workflows/fuzzy-load-test.yml:1003`)
6. ~~fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431~~ (`.github/workflows/fuzzy-load-test.yml:1000`)
7. ~~fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431~~ (`.github/workflows/fuzzy-load-test.yml:1000`)
8. ~~cross_node_write_read operation references undefined fuzzy_key and fuzzy_value~~ (`workflows/fuzzy-tests/group-governance/fuzzy-test.yml:245`)

</details>

### 🆕 New Issues

*These issues were found in the latest changes:*

### 🟡 Warning (5)

#### 1. KeyDelivery timeout now fails join_group — breaking change for slow networks
**File:** `crates/context/src/handlers/join_group.rs` (line 290-340) | **Consensus:** 1/3 agents

The PR restores a typed `Err` return when `key_delivery_fallback_wait` (default 5 s) elapses without a `KeyDelivery` op arriving. Previously this was a soft `warn!` + continue. While the PR's intent is correct (callers should know the key didn't arrive), the 5 s default is very tight: it starts AFTER `MemberJoined` is published, but the admin must (1) receive the gossip op, (2) apply it, (3) call `sign_and_publish_namespace_op` for `KeyDelivery` with its own `OP_ACK_MEMBER_CHANGE_TIMEOUT` (5 s) budget, and (4) have that gossip reach the joiner. Under any network jitter or cold-start GRAFT delay the total round-trip easily exceeds 5 s, causing `join_group` to return `Err` even though the key will arrive moments later. The caller has no retry guidance (no typed error variant — it's a raw `eyre::eyre!` string), so clients cannot distinguish this transient condition from a permanent failure.

**Suggested fix:**
```
Either (a) surface a typed `JoinGroupError::KeyDeliveryTimeout` so callers can retry, or (b) increase the default `key_delivery_fallback_wait` to at least 15–30 s to account for the full admin round-trip. The `ContextManagerConfig` struct already exists for (b); the PR description says the default is preserved at 5 s but that value was chosen when the branch was a soft-fail, not a hard error.
```

> *Found by: patterns-reviewer*

#### 2. KeyDelivery timeout returns hard error breaking join atomicity
**File:** `crates/context/src/handlers/join_group.rs` (line 300-340) | **Consensus:** 1/3 agents

The PR restores a typed `Err` return when the KeyDelivery gossip-fallback deadline elapses. However, by this point `MemberJoined` has already been published to the namespace DAG and governance ops from the join response have been applied locally. Returning `Err` here leaves the node in a partially-joined state: it is a member on the DAG (other peers see it as joined) but the caller receives a failure and may not retry correctly or may retry the full `join_group` flow, potentially publishing a duplicate `MemberJoined` op. The comment acknowledges this is a restoration of the typed-error contract, but the asymmetry between DAG state and caller-visible result is a logic hazard that could cause membership divergence or duplicate join ops under retry.

**Suggested fix:**
```
Consider returning a distinct typed error (e.g. `JoinedButKeyPending`) that signals 'join succeeded, key not yet available — poll execute until GroupKeyPending clears' rather than a generic `eyre::eyre!` that callers may treat as a full join failure and retry from scratch.
```

#### 3. try_send on actor mailbox may silently drop FSM signals under load
**File:** `crates/node/primitives/src/client.rs` (line 284-310) | **Consensus:** 1/3 agents

notify_namespace_op_applied uses try_send which returns Err(SendError::Full) if the NodeManager mailbox is saturated. The code logs a warning and discards the signal. Under sustained governance load (e.g., the new fuzzy test with 50% set_and_verify traffic), the publisher path can fire this notification on every applied op. If NodeManager's mailbox fills up — which is plausible during the 15-60 minute fuzzy run with 4 nodes each publishing at 50ms intervals — the readiness FSM will silently lag. The comment says 'the FSM will catch up on its next tick or when the next op fires' but there is no periodic tick; the FSM only advances on explicit signals. A saturated mailbox during a burst means the FSM stays at Bootstrapping for the duration of the burst, which is the exact bug this PR is trying to fix.

**Suggested fix:**
```
Consider using do_send (unbounded) for this signal since it is lightweight and correctness-critical, or implement a periodic reconciliation tick in ReadinessManager that re-scans state_per_namespace from the store so transient signal loss is self-healing. Alternatively, batch the notification: track a dirty flag per namespace and flush once per actor message cycle rather than once per op.
```

> *Found by: performance-reviewer*

#### 4. notify_namespace_op_applied uses try_send which silently drops signals on mailbox saturation
**File:** `crates/node/primitives/src/client.rs` (line 284-317) | **Consensus:** 1/3 agents

The `notify_namespace_op_applied` method uses `try_send` and logs a warning on `SendError::Full`, explicitly documenting this as best-effort. However, the readiness FSM correctness depends on receiving these signals — a saturated NodeManager mailbox (e.g. during high-throughput backfill) will silently drop FSM advancement signals, leaving namespaces stuck in `Bootstrapping` until the next op or peer beacon. The warning log is the only recovery signal, and operators may not be monitoring for it. The doc comment says 'FSM will catch up on its next tick' but there is no periodic tick described — it only catches up when another op fires.

**Suggested fix:**
```
Either use a bounded async send with a short timeout, or ensure the ReadinessManager has a periodic reconciliation tick that re-scans applied ops independently of the signal path, so dropped signals self-heal without requiring another op.
```

#### 5. KeyDelivery wait loop polls store on every broadcast channel event, including Lagged bursts
**File:** `crates/context/src/handlers/join_group.rs` (line 290-395) | **Consensus:** 1/3 agents

The gossip-fallback wait loop calls group_store::load_current_group_key on every iteration — including after RecvError::Lagged, which can fire N times in rapid succession if the broadcast channel (capacity 64) is flooded by other tasks. Each load_current_group_key is a synchronous RocksDB read. During the fuzzy test with 4 nodes publishing at 50ms intervals, the process-wide broadcast channel can easily lag, causing the join_group future to hammer RocksDB with redundant reads for the full key_delivery_fallback_wait duration (5 seconds default). The comment acknowledges the Lagged case but does not rate-limit the store reads.

**Suggested fix:**
```
Add a short sleep or use a tokio::time::interval to rate-limit store reads: only re-check the store every 100-200ms regardless of how many channel events arrive. The channel recv is still needed to wake up promptly when a KeyDelivery arrives, but the store read can be decoupled from the channel event rate. Example: track last_store_check: Instant and only call load_current_group_key if last_store_check.elapsed() > Duration::from_millis(100).
```

> *Found by: performance-reviewer*

### 💡 Suggestion (6)

#### 1. newly_applied counter in governance catch-up path counts Applied outcomes but applied counter counts all successes — inconsistency
**File:** `crates/node/src/sync/manager/mod.rs` (line 2720-2760) | **Consensus:** 1/3 agents

In `request_governance_catchup_from_peer` (around line 2720), `applied` is incremented for both `Applied` and non-Applied (e.g. `Duplicate`) outcomes, while `newly_applied` only counts `Applied`. The debug log at the end reports `ops_applied = applied`, which includes duplicates. This is misleading: an operator seeing `ops_applied = 10` when all 10 were `Duplicate` would think catch-up succeeded when in fact nothing new was applied. The same pattern exists in the governance catch-up path in `sync_namespace_from_peer`.

**Suggested fix:**
```
Rename `applied` to `processed` (or `total_ok`) to make clear it counts all non-error outcomes, and log both `ops_processed` and `ops_newly_applied` so operators can distinguish real progress from duplicate re-delivery.
```

> *Found by: patterns-reviewer*

#### 2. ForwardNamespaceOpApplied signals dropped during startup window are not recoverable without another op
**File:** `crates/node/src/handlers.rs` (line 56-84) | **Consensus:** 1/3 agents

The handler documents that signals arriving before `setup_readiness_manager` runs are dropped, and that 'the FSM will reconcile via next op or peer beacon'. However, if a node publishes governance ops during startup (e.g. in a fast single-node test or during namespace bootstrap) and no subsequent op or beacon arrives, the FSM will remain at `Bootstrapping` indefinitely. The drop is logged at `debug!` level which may not be visible in production log levels, making this silent in practice.

**Suggested fix:**
```
Promote the drop log to `warn!` level, or queue dropped signals in a small buffer that is replayed once `readiness_addr` is set in `setup_readiness_manager`.
```

#### 3. ContextManagerConfig has no way to be overridden at runtime — the struct is always default-constructed
**File:** `crates/context/src/lib.rs` (line 38-68) | **Consensus:** 1/3 agents

`ContextManagerConfig` is documented as 'runtime-tunable knobs' and 'future operator tooling can override them without source patches', but `ContextManager::new` always constructs it with `ContextManagerConfig::default()` (line 180) and there is no public setter, builder, or constructor parameter that accepts a custom config. The `config` field is `pub(crate)` so it cannot be set from outside the crate. The struct is effectively a named constant today, not a runtime-tunable config.

**Suggested fix:**
```
Either add a `ContextManager::with_config(config: ContextManagerConfig)` constructor or accept `config: ContextManagerConfig` as a parameter in `ContextManager::new`. Alternatively, document that the tuning mechanism is deferred and mark the field `// TODO: expose via constructor` to avoid misleading the 'runtime-tunable' claim.
```

> *Found by: patterns-reviewer*

#### 4. ContextManagerConfig is not exposed for operator override despite documentation claiming it is
**File:** `crates/context/src/lib.rs` (line 36-68) | **Consensus:** 1/3 agents

The doc comment on `ContextManagerConfig` states it centralises timing constants 'so future operator tooling can override them without source patches', but `ContextManager::new` always constructs it with `ContextManagerConfig::default()` and there is no public API to inject a custom config. The struct fields are `pub` but the struct itself is only reachable via `Default`. This means the documented operator-override capability does not actually exist yet, which could mislead future contributors into assuming it is wired up.

**Suggested fix:**
```
Either add a `ContextManager::with_config(config: ContextManagerConfig)` constructor or a `set_config` method, or update the doc comment to say 'intended for future operator tooling' rather than implying it is already available.
```

#### 5. notify_namespace_op_applied called unconditionally on every sign_apply_and_publish invocation
**File:** `crates/context/src/group_store/namespace_governance.rs` (line 339-345) | **Consensus:** 1/3 agents

In sign_apply_and_publish, node_client.notify_namespace_op_applied(self.namespace_id) is called unconditionally after apply_signed_op succeeds. apply_signed_op always advances the DAG (it calls advance_dag_head internally), so this is correct. However, the same pattern in publish_post_gate (line 480) is also unconditional. Both paths send a cross-crate actor message (try_send into NodeManager mailbox) on every single governance op publish — including high-frequency ops like ContextAliasSet, MemberCapabilitySet, etc. During the fuzzy test's Phase 4 governance churn (capability changes, subgroup creation, member add/remove in rapid succession), this generates O(ops) mailbox messages. Since the FSM only needs to know 'at least one op was applied', batching or deduplicating these signals would reduce mailbox pressure.

**Suggested fix:**
```
Consider debouncing: track a per-namespace dirty flag in ContextManager (or NamespaceGovernance) and send a single notification after a batch of ops rather than one per op. Alternatively, since the readiness FSM only cares about monotonic progress, a single notification per handler invocation (rather than per op within a handler) would suffice — e.g., notify once at the end of add_group_members after all MemberAdded ops are published.
```

> *Found by: performance-reviewer*

#### 6. RecvError::Closed branch returns Err but the comment calls it 'functionally unreachable'
**File:** `crates/context/src/handlers/join_group.rs` (line 368-385) | **Consensus:** 1/3 agents

The comment says `op_events::NOTIFIER` cannot be dropped at runtime, making this branch unreachable. Yet the code now returns a hard `Err` instead of the previous `break`. If the branch truly is unreachable, the `Err` path is dead code; if it can be reached (e.g. after a future refactor), the error message is correct but there is no test coverage. The inconsistency between the comment ('functionally unreachable') and the code ('return Err') makes future maintainers uncertain about the invariant.

**Suggested fix:**
```
Add `#[allow(unreachable_patterns)]` with a comment, or replace the `return Err` with `unreachable!("NOTIFIER cannot be dropped")` so a future refactor that breaks the invariant panics loudly in tests rather than silently returning a misleading error to callers.
```

> *Found by: patterns-reviewer*

### 📝 Nitpick (1)

#### 1. ObserveDelivery trait is public but observe_handler_delivery free function is also kept — dual API surface
**File:** `crates/context/src/governance_broadcast.rs` (line 308-340) | **Consensus:** 1/3 agents

The PR adds `ObserveDelivery` as a trait and migrates all call sites to `report.observe(...)`, but the original `observe_handler_delivery` free function is preserved with no `#[deprecated]` annotation and no doc comment indicating it is superseded. This leaves two ways to do the same thing, which will confuse future contributors about which to use.

**Suggested fix:**
```
Either mark `observe_handler_delivery` as `#[deprecated(since = "...", note = "Use ObserveDelivery::observe instead")]`, or keep it as the implementation detail (which it already is — both trait impls delegate to it) and make it `pub(crate)` to prevent external callers from using the lower-level form directly.
```

> *Found by: patterns-reviewer*


---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-993333fe`</sub>

.github/workflows/fuzzy-load-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/fuzzy-load-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

workflows/fuzzy-tests/group-governance/fuzzy-test.yml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

crates/node/primitives/src/client.rs
```diff
@@ -284,6 +284,34 @@ impl NodeClient {
         }
     }
 
+    /// Notify the readiness FSM that a namespace governance op was just
```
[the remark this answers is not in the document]
rtb-12: Thanks for the catch — but I think the design is sound here and the suggested fixes would actually couple the publisher path more tightly to the symptom they're trying to mitigate. Quick rundown of why I'm leaving this as-is:

**Re-checked the failure modes.** `LazyRecipient::try_send` (`crates/utils/actix/src/lazy.rs:645-661`) only returns `Err` when the receiver IS wired *and* its mailbox is full — when the receiver isn't wired yet, it queues into `LazyInner::queue` and returns `Ok(())`. So the only real failure mode here is `SendError::Full` from the wired `Addr<NodeManager>::try_send` once the actor mailbox saturates (Actix default = 16). The existing doc comment captures this; the "early startup" failure path the comment-template suggests doesn't actually exist.

**The signal is idempotent and the FSM has a convergent fallback.** `ReadinessManager`'s periodic tick (`crates/node/src/readiness.rs:497-517`) re-reads `local_applied_through` from the **datastore** on every iteration:

```rust
// Refresh canonical sequence from the store before each
// evaluation so the periodic tick observes publisher-side
// applies that don't fire `NamespaceOpApplied`.
let applied_through = read_local_applied_through(&self.datastore, ns_id);
```

The DB is the ground truth; `NamespaceOpApplied` is a liveness optimization to skip waiting `ttl_heartbeat` for the next tick. Dropping the signal costs at most one tick interval of FSM lag — never correctness.

**Why each suggested fix would be worse:**

1. *Coalescing/batching by namespace_id* — buys nothing. The FSM re-reads `applied_through` from DB on evaluate; one signal vs fifty for the same `namespace_id` triggers the same re-eval. The signal is already idempotent at the consumer.
2. *Bounded retry* — actively harmful here. If the mailbox is full, retrying loops the **publisher hot path** (called from `crates/context` context-execute) waiting on the same backed-up actor. That couples publisher latency to the dispatch backpressure rather than letting the FSM tick do its job.
3. *`send().await` with timeout* — same coupling, made explicit. Wrong primitive for an apply-path emission.

**The real architectural concern lives elsewhere.** The dispatch backpressure that *would* cause this saturation is tracked in #2299 (per-variant actor isolation + per-`ContextId` DAG lock sharding) — that's where the fix belongs: at the source of the saturation, not by hardening every emitter. Hardening this caller would mask the symptom and make #2299 harder to detect.

Resolving as not-actionable here. If the FSM lag turns out to matter in production, #2299's actor isolation is the right lever.

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 3 agents** | Quality score: 67% | Review time: 94.5s

### 🟡 Issues pending resolution

✅ **8 Fixed** | 🆕 **7 New**

---

### ✅ Fixed Issues

<details>
<summary>The following issues from previous reviews have been addressed:</summary>

1. ~~fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431 — metric collection collision on concurrent runs~~ (`.github/workflows/fuzzy-load-test.yml:975`)
2. ~~Port conflict: group-governance vmagent uses same port as scaffolding-e2e~~ (`.github/workflows/fuzzy-load-test.yml:1001`)
3. ~~Port 8431 reused between scaffolding-e2e and group-governance vmagent jobs~~ (`.github/workflows/fuzzy-load-test.yml:1000`)
4. ~~fuzzy_test operation references {{fuzzy_key}} and {{fuzzy_value}} without prior definition~~ (`workflows/fuzzy-tests/group-governance/fuzzy-test.yml:201`)
5. ~~Port conflict: group-governance and scaffolding-e2e both use vmagent port 8431~~ (`.github/workflows/fuzzy-load-test.yml:1003`)
6. ~~fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431~~ (`.github/workflows/fuzzy-load-test.yml:1000`)
7. ~~fuzzy-test-group-governance and fuzzy-test-scaffolding-e2e share vmagent port 8431~~ (`.github/workflows/fuzzy-load-test.yml:1000`)
8. ~~cross_node_write_read operation references undefined fuzzy_key and fuzzy_value~~ (`workflows/fuzzy-tests/group-governance/fuzzy-test.yml:245`)

</details>

### 🆕 New Issues

*These issues were found in the latest changes:*

### 🟡 Warning (4)

#### 1. KeyDelivery timeout returns hard error breaking join atomicity
**File:** `crates/context/src/handlers/join_group.rs` (line 290-340) | **Consensus:** 1/3 agents

The KeyDelivery fallback wait now returns `Err` on timeout (restoring the typed-error contract per the PR description), but this means a node that successfully published `MemberJoined` and had its governance ops applied to the local DAG will surface a failure to the caller. The caller has no way to distinguish 'join succeeded but key not yet delivered' from 'join failed entirely'. A retry of `join_group` after this error may re-publish `MemberJoined` (creating a duplicate op or failing the invitation check), leaving the node in a partially-joined state. The error message says 'join cannot proceed without a usable group key' but the node IS already joined at the namespace level — only the key delivery failed. This semantic mismatch could cause clients to attempt a full re-join rather than just waiting for key delivery, potentially corrupting namespace membership state.

**Suggested fix:**
```
Consider returning a distinct typed error (e.g. `JoinGroupError::KeyDeliveryTimeout`) that signals 'joined but key pending' vs 'join failed' so callers can retry only the key-delivery wait without re-publishing MemberJoined. Alternatively, persist a 'joined-but-key-pending' marker so a retry path can skip the MemberJoined publish.
```

#### 2. KeyDelivery wait loop polls store on every broadcast channel event, including Lagged
**File:** `crates/context/src/handlers/join_group.rs` (line 290-395) | **Consensus:** 1/3 agents

The gossip-fallback wait loop calls load_current_group_key on every iteration — including after RecvError::Lagged, which can fire many times in rapid succession if the broadcast channel is flooded by other tasks. Each call is a synchronous RocksDB read. Under the new fuzzy test's 4-node governance load, the broadcast channel (process-wide) will see high traffic, making Lagged events frequent. The loop comment acknowledges this ('broadcast channel is process-wide and other tasks may flood it') but the store read on every Lagged event means O(flood_rate) RocksDB reads during the wait window. The key_delivery_fallback_wait is 5s; at 50ms operation_delay_ms with 4 nodes, the channel could see ~80 events/s, meaning ~400 store reads per join during the wait window even if the key arrives in the first second.

**Suggested fix:**
```
Add a short debounce: only re-read the store if at least N ms have elapsed since the last read, or only re-read on non-Lagged events. Alternatively, use a tokio::time::interval for the store poll and only use the broadcast channel as a wake-up hint without reading the store on every event. Example: track `last_store_check = Instant::now()` and skip the store read if `last_store_check.elapsed() < Duration::from_millis(100)`.
```

> *Found by: performance-reviewer*

#### 3. try_send on actor mailbox may silently drop FSM signals under backpressure
**File:** `crates/node/primitives/src/client.rs` (line 284-310) | **Consensus:** 1/3 agents

notify_namespace_op_applied uses try_send which returns Err(SendError::Full) when the NodeManager mailbox is saturated. The error is logged but the FSM signal is permanently lost. Under sustained governance load (e.g. the new fuzzy test with 50% set_and_verify traffic), the publisher path emits one notify per applied op. If NodeManager's mailbox fills — which is plausible during batch backfill or sync catch-up where dozens of ops are applied in a tight loop — the FSM will silently lag. The comment says 'the FSM will catch up on its next tick or when the next op fires' but there is no guaranteed next tick; the FSM only advances when explicitly notified. This is a best-effort signal by design, but the performance implication is that the FSM can stay at Bootstrapping for an unbounded window under load.

**Suggested fix:**
```
Consider coalescing notifications: instead of one try_send per op in a batch, collect a set of namespace_ids that had Applied outcomes and send a single notification per unique namespace_id after the loop. This reduces mailbox pressure by up to N-fold for a batch of N ops on the same namespace. Example in fetch_and_apply_namespace_backfill: the any_applied bool already gates the single post-batch notify, which is correct. The issue is in the two publisher-side call sites in namespace_governance.rs (lines 339 and 480) which fire once per sign_apply_and_publish call — those are already single-op, so they're fine. The sync/manager paths are also already batched correctly.
```

> *Found by: performance-reviewer*

#### 4. Redundant applied counter incremented for both Applied and non-Applied outcomes
**File:** `crates/node/src/sync/manager/mod.rs` (line 2720-2780) | **Consensus:** 1/3 agents

In the governance catch-up loop (around line 2740-2770), the refactored match arm increments `applied += 1` for both `Ok(NamespaceApplyOutcome::Applied)` and `Ok(_)` (Pending/Duplicate). The original code also incremented `applied` for every non-error outcome. However, the new `newly_applied` counter only increments for `Applied`. The `applied` counter is used in the debug log below. This is not a correctness issue but the naming is now misleading: `applied` counts 'non-error outcomes' not 'actually applied ops', while `newly_applied` counts the true applies. Under high-duplicate-op conditions (e.g. a node that already has most ops and receives a backfill), `applied` will be large but `newly_applied` will be near zero, and the single FSM notify is correctly gated on `newly_applied > 0`. No performance regression, but the counter semantics create confusion that could lead to a future bug where someone gates the FSM notify on `applied > 0` instead.

**Suggested fix:**
```
Rename `applied` to `non_error_outcomes` or `processed` to make the distinction explicit, or remove it entirely if the debug log is the only consumer and the distinction between Applied vs Pending/Duplicate is already captured by `newly_applied`.
```

> *Found by: performance-reviewer*

### 💡 Suggestion (3)

#### 1. notify_namespace_op_applied uses try_send which silently drops signals under mailbox pressure
**File:** `crates/node/primitives/src/client.rs` (line 284-315) | **Consensus:** 1/3 agents

The `notify_namespace_op_applied` method uses `try_send` and logs a warning on failure, but the warning message says 'readiness FSM will lag for this namespace until the next op fires or a peer beacon arrives'. Under sustained load (e.g. the new fuzzy test's 50ms operation delay with 4 nodes), the NodeManager mailbox could saturate during batch apply paths (backfill, sync catch-up), causing multiple FSM notifications to be silently dropped. The FSM may stay at `Bootstrapping` longer than expected. The `warn!` log is the only signal — there is no metric or counter for dropped notifications.

**Suggested fix:**
```
Add a metric counter for dropped NamespaceOpApplied signals, or use a bounded channel with backpressure rather than fire-and-forget try_send. At minimum, consider upgrading the log from `warn!` to include a counter so operators can detect systematic mailbox saturation.
```

#### 2. ContextManagerConfig has no mechanism for runtime override despite documentation claiming operator configurability
**File:** `crates/context/src/lib.rs` (line 36-65) | **Consensus:** 1/3 agents

The `ContextManagerConfig` struct is documented as allowing operators to override timing constants 'without source patches', but the struct is always initialized with `ContextManagerConfig::default()` in `ContextManager::new()` (line ~180) and there is no constructor, builder, or configuration loading path that accepts a non-default value. The `pub(crate)` visibility on the `config` field means external crates cannot set it either. The documentation promise of operator configurability is not fulfilled by the implementation.

**Suggested fix:**
```
Either expose a `ContextManager::with_config(config: ContextManagerConfig)` constructor or wire `ContextManagerConfig` into the existing on-disk config loading path. If runtime override is out of scope for this PR, update the doc comment to say 'planned for a future PR' rather than implying it works today.
```

#### 3. applied counter incremented for both Applied and non-Applied outcomes but newly_applied only tracks Applied
**File:** `crates/node/src/sync/manager/mod.rs` (line 2720-2760) | **Consensus:** 1/3 agents

In the sync catch-up path (around line 2740-2760 in the diff), the `applied` counter is incremented for both `NamespaceApplyOutcome::Applied` AND `Ok(_)` (non-Applied outcomes like Duplicate/Pending), while `newly_applied` only increments for `Applied`. The debug log at the end of the function uses `applied` for its count. This means the log line 'applied N of M ops' will report Duplicate/Pending outcomes as 'applied', which is misleading for diagnostics. The FSM notification correctly gates on `newly_applied > 0`, but the observability is inconsistent.

**Suggested fix:**
```
Either rename `applied` to `processed` to reflect that it counts all non-error outcomes, or split the counter to separately track Applied vs Duplicate/Pending for accurate debug logging.
```

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-408c70d5`</sub>

frdomovic (commented):
## 🤖 MeroReviewer

**All previous comments resolved** | Quality score: 100%

✅ All issues resolved. Ready to merge!

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `lgtm-fast-path`</sub>

frdomovic (commented):
## 🤖 MeroReviewer

**All previous comments resolved** | Quality score: 100%

✅ All issues resolved. Ready to merge!

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `lgtm-fast-path`</sub>

frdomovic (commented):
## 🤖 MeroReviewer

**All previous comments resolved** | Quality score: 100%

✅ All issues resolved. Ready to merge!

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `lgtm-fast-path`</sub>

Issue #224, opened 2024-04-24
Create a glossary in the documentation with the key concepts


chefsale:
For example: Application Context, etc.

PR #2292, opened 2026-05-07
test(e2e): migrate group-leave-member to wait_for_sync (groupStateHash)

chefsale:
First migration of an existing e2e workflow from a fixed-duration `wait` to the new governance-aware `wait_for_sync` from A1+A2 of the cross-DAG authorization roadmap.

## Change

In `apps/scaffolding-e2e/workflows/group-leave-member.yml`, after node-2 voluntarily leaves the subgroup:

```yaml
# Before
- name: Wait for MemberLeft to propagate to node-1
  type: wait
  seconds: 5

# After
- name: Wait for MemberLeft to propagate to remaining members
  type: wait_for_sync
  group_id: '{{subgroup_id}}'
  nodes:
    - e2e-node-1
    - e2e-node-3
  timeout: 30
  check_interval: 2
```

## Why

- **Deterministic**: workflow proceeds the moment both remaining peers agree on the post-leave `groupStateHash`, instead of always waiting 5s
- **Faster on healthy networks** (typically converges in <1s)
- **More reliable under load** (timeout=30s gives generous headroom if the runner is slow)
- **Demonstrates A2 in real CI**

## Why exclude node-2 from the polling set

`get_group_info` rejects callers who are no longer members of the group (`crates/context/src/handlers/get_group_info.rs:24` — `bail!("node is not a member")`). node-2 just left, so its admin API would return an error for this subgroup. node-1 and node-3 are still members and converge on the same post-leave state.

## Why not also migrate group-leave-namespace.yml

That test runs with only 2 nodes; after node-2 leaves, only node-1 remains in the group. `wait_for_sync` requires ≥2 nodes for consensus verification, so it can't be used there without restructuring the test to add a 3rd node. Tracked separately.

---

> [!NOTE]
> **Medium Risk**
> Changes group invitation serialization and join bootstrap logic by adding an unsigned `application_id` field; could impact backwards compatibility or interoperability if any components assume the previous invitation layout. E2E workflow timing changes are low risk but may expose flakiness if `wait_for_sync` semantics differ under load.
> 
> **Overview**
> Updates the `group-leave-member` E2E workflow to replace a fixed 5s sleep with governance-aware `wait_for_sync` polling of `groupStateHash` across remaining members after a leave.
> 
> Extends `SignedGroupOpenInvitation` with an optional, unsigned `application_id` and plumbs it through invitation creation (`create_group_invitation` and recursive invites) and `join_group` so joiners can initialize `GroupMetaValue.target_application_id` consistently, avoiding persistent `compute_group_state_hash` divergence. Tests are adjusted to account for the new optional field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bc8083bf059942a0c52420d83f7b4b3f1ea4d41. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/scaffolding-e2e/workflows/group-leave-member.yml
+++ b/apps/scaffolding-e2e/workflows/group-leave-member.yml
@@ -91,9 +91,18 @@ steps:
     group_id: '{{subgroup_id}}'
 
   # MemberLeft has to propagate via gossip before peers reflect it.
-  - name: Wait for MemberLeft to propagate to node-1
-    type: wait
-    seconds: 5
+  # wait_for_sync polls groupStateHash across remaining members until
+  # they agree, replacing the old fixed-duration wait. Excludes node-2
+  # because it just left and `get_group_info` rejects callers who are
+  # no longer members of the group.
+  - name: Wait for MemberLeft to propagate to remaining members
+    type: wait_for_sync
+    group_id: '{{subgroup_id}}'
+    nodes:
+      - e2e-node-1
+      - e2e-node-3
+    timeout: 30
+    check_interval: 2
 
   - name: List subgroup members on node-1 after node-2 left
     type: list_group_members
--- a/crates/context/config/src/types.rs
+++ b/crates/context/config/src/types.rs
@@ -671,12 +671,25 @@ fn default_invited_role() -> u8 {
 }
 
 /// A container for a group invitation and the admin's signature over it.
+///
+/// The fields below `inviter_signature` are **not** covered by the signature.
+/// They are populated by the inviter from local state so the joiner can
+/// pre-populate `GroupMetaValue` with the correct values instead of
+/// zero placeholders. Without this, peers that join via invitation end
+/// up with `target_application_id = ZERO` for the group, while the
+/// originator has the real value — causing `compute_group_state_hash`
+/// to diverge persistently between peers.
 #[derive(BorshSerialize, BorshDeserialize, Clone, Debug, Deserialize, Serialize)]
 pub struct SignedGroupOpenInvitation {
     /// The open invitation to the group.
     pub invitation: GroupInvitationFromAdmin,
     /// Admin's signature for the invitation payload (hex-encoded).
     pub inviter_signature: String,
+    /// Application ID for the group (unsigned bootstrap field).
+    /// `None` for backwards compatibility with invitations created before
+    /// this field was added; joiners fall back to zero when absent.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub application_id: Option<[u8; 32]>,
 }
 
 #[cfg(test)]
--- a/crates/context/src/group_store/namespace.rs
+++ b/crates/context/src/group_store/namespace.rs
@@ -250,9 +250,17 @@ pub fn create_recursive_invitations(
             .sign(&hash)
             .map_err(|e| eyre::eyre!("signing: {e}"))?;
 
+        // Carry the real application_id so the joiner can pre-populate
+        // GroupMetaValue correctly. Without this, joiners would write
+        // target_application_id = ZERO and compute_group_state_hash
+        // would diverge from the inviter's view persistently.
+        let application_id =
+            super::load_group_meta(store, &gid)?.map(|m| *m.target_application_id.as_ref());
+
         let signed = SignedGroupOpenInvitation {
             invitation,
             inviter_signature: hex::encode(sig.to_bytes()),
+            application_id,
         };
 
         result.push((gid, signed));
--- a/crates/context/src/group_store/namespace.rs
+++ b/crates/context/src/group_store/namespace.rs
@@ -254,8 +254,26 @@ pub fn create_recursive_invitations(
         // GroupMetaValue correctly. Without this, joiners would write
         // target_application_id = ZERO and compute_group_state_hash
         // would diverge from the inviter's view persistently.
-        let application_id =
-            super::load_group_meta(store, &gid)?.map(|m| *m.target_application_id.as_ref());
+        //
+        // If meta is unexpectedly missing (e.g. partial-write state — child
+        // index says the group exists but no meta row yet), warn loudly and
+        // fall back to None so the joiner gets the same zero placeholder it
+        // would have had before this fix. We don't bail because that would
+        // poison the entire recursive-invitation set on a single bad
+        // descendant; the joiner's existing zero-app self-heal path
+        // (`context_registration::register_context_in_group`) covers the
+        // missing-meta case the same way it covered it pre-fix.
+        let application_id = match super::load_group_meta(store, &gid)? {
+            Some(meta) => Some(*meta.target_application_id.as_ref()),
+            None => {
+                tracing::warn!(
+                    group_id = %hex::encode(gid.to_bytes()),
+                    "create_recursive_invitations: missing GroupMeta for descendant; \
+                     issuing invitation with application_id = None (joiner will fall back to zero)"
+                );
+                None
+            }
+        };
 
         let signed = SignedGroupOpenInvitation {
             invitation,
--- a/crates/context/src/handlers/create_group_invitation.rs
+++ b/crates/context/src/handlers/create_group_invitation.rs
@@ -50,7 +50,7 @@ impl Handler<CreateGroupInvitationRequest> for ContextManager {
         let datastore = self.datastore.clone();
 
         let result = (|| -> eyre::Result<_> {
-            let _meta = group_store::load_group_meta(&datastore, &group_id)?
+            let meta = group_store::load_group_meta(&datastore, &group_id)?
                 .ok_or_else(|| eyre::eyre!("group not found"))?;
 
             group_store::require_group_admin_or_capability(
@@ -102,6 +102,12 @@ impl Handler<CreateGroupInvitationRequest> for ContextManager {
                 SignedGroupOpenInvitation {
                     invitation,
                     inviter_signature,
+                    // Carry the real application_id so the joiner can
+                    // pre-populate GroupMetaValue correctly. Without this,
+                    // joiners would write target_application_id = ZERO
+                    // and compute_group_state_hash would diverge from
+                    // the inviter's view persistently.
+                    application_id: Some(*meta.target_application_id.as_ref()),
                 },
                 group_alias,
             ))
--- a/crates/context/src/handlers/join_group.rs
+++ b/crates/context/src/handlers/join_group.rs
@@ -91,11 +91,23 @@ impl Handler<JoinGroupRequest> for ContextManager {
                     let admin_identity = calimero_primitives::identity::PublicKey::from(
                         invitation.invitation.inviter_identity.to_bytes(),
                     );
+                    // Use the application_id carried in the invitation when
+                    // present (added so joiners pre-populate `GroupMetaValue`
+                    // with the real value instead of zero — without it,
+                    // `compute_group_state_hash` diverges between the
+                    // originator and joining peers because
+                    // target_application_id is part of the hash). Falls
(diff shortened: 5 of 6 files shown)


chefsale:
Pushed `b21acfda` — root-cause fix for the state-hash divergence the e2e migration uncovered.

## What was actually broken

When a peer joins a group via invitation, the joiner's `join_group` handler pre-populated `GroupMetaValue.target_application_id = [0u8; 32]` (`crates/context/src/handlers/join_group.rs:97-98`). The inviter had the real value. Subsequent `create_group_in_namespace` ops inherit target_application_id from parent meta on each peer's local store — so subgroups inherit ZERO on the joiner and the real id on the originator.

`compute_group_state_hash` covers `target_application_id` → persistent hash divergence. There's a self-healing path in `context_registration` that fills in zeros on context registration, but tests that don't register contexts (this one) leave it uncorrected. Pre-existing bug; A1's hash exposure exposed it.

## Fix

Extended `SignedGroupOpenInvitation` with an optional unsigned `application_id` field (mirrors the existing `SignedOpenInvitation` pattern). Inviter populates from local meta; joiner uses it when present, falls back to zero otherwise.

## Test plan

- ✅ `cargo test -p calimero-context-config`
- ✅ `cargo test -p calimero-context --lib` (173 tests)
- ✅ `cargo test -p calimero-context --test local_group_governance_convergence` (22 tests)
- ⏳ This PR's e2e (group-leave-member with wait_for_sync) — should now pass once CI rebuilds the calimero binary

The wait_for_sync change in commit 1 was always correct; it was just exposing a real bug. With both commits, the e2e becomes deterministic.

chefsale:
Addressed in `d3f51455`:

**Bot review triage:**

- ✅ **meroreviewer S2** (`application_id` silently None on missing meta): added a `tracing::warn!` when `load_group_meta` returns `None` for a descendant in `create_recursive_invitations`. Doesn't bail (would poison the legitimate sibling invitations) — falls back to `None` and lets the joiner's existing zero-app self-heal path cover it. Loud enough to surface partial-write states without breaking the recursive flow.

- ⏭️ **meroreviewer W1 + cursor B1** (`wait_for_sync` with `group_id` is undefined): bots are out of date. The `group_id` variant ships in **merobox 0.6.5+** and **0.6.6+** ([merobox#223](https://github.com/calimero-network/merobox/pull/223), [#224](https://github.com/calimero-network/merobox/pull/224)). The merobox CHANGELOG and Pydantic schema already document/validate it. Both bots inferred "undefined" from grep of the *core repo* (where this is the first usage) without seeing the merobox extension.

- ⏭️ **meroreviewer N1** (`get_group_info.rs:24` line-number drift): the YAML comment doesn't actually contain a line number — the comment just says "`get_group_info` rejects callers who are no longer members". The line number was only in my PR description (where drift is fine; PRs are immutable history).

chefsale:
Re-investigated — the divergence persists because `calimero-client-py 0.6.7` strips the new `application_id` field during typed deserialize at [`src/client.rs:1754`](https://github.com/calimero-network/calimero-client-py/blob/v0.6.7/src/client.rs#L1754):

```rust
let invitation: calimero_context_config::types::SignedGroupOpenInvitation =
    serde_json::from_str(invitation_json)...;
```

0.6.7 was built against `master#219ef0d2` which is pre-this-PR, so the compiled `SignedGroupOpenInvitation` type doesn't have `application_id`. The field is silently dropped on the merobox round-trip, joiner falls back to zero, divergence persists.

Same coordination dance as A1:
- [calimero-client-py#45](https://github.com/calimero-network/calimero-client-py/pull/45) — peg client-py to this branch + bump 0.6.7 → 0.6.8
- After client-py 0.6.8 cuts: bump merobox to require >=0.6.8 and release merobox 0.6.8
- e2e CI on this PR will pass
- Merge this PR → master gets the fix
- Follow-up: flip client-py back to master, bump 0.6.8 → 0.6.9

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 2 agents** | Quality score: 28% | Review time: 66.3s

---

### 🟡 Warning (7)

#### 1. Unsigned field in signed structure enables application_id spoofing
**File:** `crates/context/config/src/types.rs` (line 695-700) | **Consensus:** 1/2 agents

The `application_id` field in `SignedGroupOpenInvitation` is explicitly documented as NOT covered by the admin's signature (`inviter_signature`). A man-in-the-middle or malicious relay could modify this field to point to a different application ID before the joiner processes the invitation. The joiner in `join_group.rs` (line ~113) unconditionally trusts `invitation.application_id` to initialize `target_application_id` in `GroupMetaValue`. If an attacker substitutes a malicious application ID, the joining node will initialize its group metadata with the wrong application, potentially causing it to load and execute a different WASM application than intended.

**Suggested fix:**
```
Either include `application_id` in the signed payload (i.e., move it inside `GroupInvitationFromAdmin` so it is covered by the ed25519 signature), or verify the application_id against a trusted source (e.g., the group's on-chain/DAG state) before using it. At minimum, document clearly that this field must be validated out-of-band before trusting it for security-sensitive decisions.
```

#### 2. Unconditional 2-second sleep blocks the actor for every join
**File:** `crates/context/src/handlers/join_group.rs` (line 155-165) | **Consensus:** 1/2 agents

`tokio::time::sleep(NAMESPACE_MESH_GRACE).await` (2 seconds) is called unconditionally on every `join_group` invocation. This blocks the async task for 2 seconds even when the mesh is already formed (e.g. repeated joins, or joins on a node that is already subscribed). This adds 2 seconds of latency to every join operation and holds the actor's async task slot for that duration.

**Suggested fix:**
```
Consider checking whether the node is already subscribed/meshed before sleeping, or reduce the grace period and make it configurable. At minimum, document that this is a known latency floor for join operations.
```

> *Found by: performance-reviewer*

#### 3. System clock used for expiration without monotonic guard
**File:** `crates/context/src/handlers/create_group_invitation.rs` (line 87-88) | **Consensus:** 1/2 agents

The expiration timestamp is computed as `now_secs + expiration_timestamp.unwrap_or(365 * 24 * 3600)` using `std::time::SystemTime::now()` with `.expect("system clock before epoch")`. If the system clock is set before the Unix epoch (e.g., in misconfigured environments or tests), this panics. More importantly, there is no upper bound on the expiration window — a caller could pass `u64::MAX - now_secs` as `expiration_timestamp`, creating an invitation that never expires. Combined with the unsigned `application_id` field, a stolen never-expiring invitation with a substituted application_id is a persistent attack vector.

**Suggested fix:**
```
Add a maximum expiration cap (e.g., 1 year) and handle the clock-before-epoch case gracefully with an error rather than a panic. Consider using a monotonic clock or a trusted time source for security-sensitive expiration calculations.
```

#### 4. Redundant store lookup per group in create_recursive_invitations
**File:** `crates/context/src/group_store/namespace.rs` (line 218-290) | **Consensus:** 1/2 agents

For each group in the `groups` vec, `super::load_group_meta(store, &gid)?` opens a new store handle and performs a point lookup. When there are many descendant groups (e.g. a deep namespace tree), this is O(N) sequential store reads with a new handle allocation per iteration. The `collect_descendant_groups` call above already traverses the tree; the meta could be loaded in the same pass or batched. Additionally, `collect_descendant_groups` itself does a BFS with one `list_child_groups` call per node (each opening a new iterator), so the total store I/O is O(N) iterators + O(N) point lookups — all sequential.

**Suggested fix:**
```
Consider loading group meta in the same BFS pass inside `collect_descendant_groups`, or accept a `HashMap<ContextGroupId, GroupMetaValue>` pre-fetched by the caller. At minimum, reuse a single store handle across all meta lookups in the loop rather than calling `load_group_meta` (which internally calls `store.handle()`) once per group.
```

> *Found by: performance-reviewer*

#### 5. Invitation expiration check skipped when expiration_timestamp is zero
**File:** `crates/context/src/handlers/join_group.rs` (line 91-113) | **Consensus:** 1/2 agents

At line ~47-53, the expiration check is guarded by `if expiration != 0`. This means an invitation with `expiration_timestamp = 0` never expires and is permanently valid. While this may be intentional for some use cases, combined with the fact that `application_id` is unsigned, a zero-expiration invitation with a substituted application_id creates a persistent attack surface. There is no revocation mechanism visible in this code path.

**Suggested fix:**
```
Document explicitly that `expiration_timestamp = 0` means 'no expiration' and ensure this is an intentional, admin-controlled choice. Consider requiring a non-zero expiration for production invitations, or add a revocation list check.
```

#### 6. O(N) sequential store handle allocations in collect_descendant_groups BFS
**File:** `crates/context/src/group_store/namespace.rs` (line 155-175) | **Consensus:** 1/2 agents

`collect_descendant_groups` calls `list_child_groups` for each node in the BFS queue. Each `list_child_groups` call invokes `collect_keys_with_prefix`, which opens a new iterator on the store. For a namespace tree with N groups, this is N iterator allocations. With RocksDB, each iterator seek is non-trivial. A single range scan over the `GROUP_CHILD_INDEX_PREFIX` namespace would be more efficient for large trees.

**Suggested fix:**
```
For large trees, consider a single prefix scan over all `GroupChildIndex` keys and build the adjacency list in memory, then do the BFS/DFS in memory. This reduces N iterator allocations to 1.
```

> *Found by: performance-reviewer*

#### 7. Per-context mutex acquisition in ContextIdentity batch write
**File:** `crates/context/src/handlers/join_group.rs` (line 230-260) | **Consensus:** 1/2 agents

The comment at line ~230 says 'Batch all ContextIdentity writes in a single handle to avoid per-context mutex acquisition overhead' — this is correctly done. However, the subsequent loop at lines ~260-290 calls `group_store::register_context_in_group` per context, each of which internally calls `store.handle()` (a new handle/mutex acquisition per context). For a group with many contexts, this is O(N) handle acquisitions after the batched write.

**Suggested fix:**
```
Extend the batched handle block to also write the `ContextGroupRef` and `GroupContextIndex` entries for each context, eliminating the per-context `register_context_in_group` calls in the loop below.
```

> *Found by: performance-reviewer*


### 💡 Suggestion (2)

#### 1. Sequential async calls per context in auto-join loop
**File:** `crates/context/src/handlers/join_group.rs` (line 270-310) | **Consensus:** 1/2 agents

The auto-join loop calls `context_client.sync_context_config`, `node_client.subscribe`, and `node_client.sync` sequentially for each context. For a group with N contexts, this is 3N sequential async round-trips. These could be batched or parallelized with `futures::future::join_all` or similar.

**Suggested fix:**
```
Collect all futures and await them concurrently: `futures::future::join_all(contexts.iter().map(|ctx| async { ... })).await`. This would reduce wall-clock time from O(N * RTT) to O(RTT) for the auto-join phase.
```

> *Found by: performance-reviewer*

#### 2. Silent fallback to None application_id on missing GroupMeta may mask data integrity issues
**File:** `crates/context/src/group_store/namespace.rs` (line 250-285) | **Consensus:** 1/2 agents

In `create_recursive_invitations`, when `load_group_meta` returns `None` for a descendant group, the code logs a warning and falls back to `application_id = None`. The comment acknowledges this is a 'partial-write state'. A joiner receiving such an invitation will initialize `target_application_id` to zero. While the self-heal path is mentioned, this silent degradation could allow a node in a corrupted state to issue invitations that cause joiners to diverge from the correct group state. The warning is logged but the caller receives no indication that some invitations are degraded.

**Suggested fix:**
```
Consider returning a structured result that indicates which invitations have missing metadata, allowing callers to decide whether to proceed or abort. Alternatively, expose a separate return value or flag indicating degraded invitations so the caller can log or alert at a higher severity level.
```

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-642caf86`</sub>

PR #2234, opened 2026-04-23
ci: add AI code review and doc auto-update workflows

frdomovic:
> [!NOTE]
> **Medium Risk**
> Introduces new GitHub Actions that run on PRs and on pushes to `main/master` and consume repo secrets (`ANTHROPIC_API_KEY`, optional `GH_PAT`), so misconfiguration could impact CI behavior and secret exposure surface.
> 
> **Overview**
> Adds repo-level AI automation via `.ai-reviewer.yaml`, pinning Anthropic models (Sonnet for reviews, Haiku for doc generation) and enabling automated doc updates for `architecture/`.
> 
> Introduces two GitHub Actions workflows pinned to a specific `calimero-network/ai-code-reviewer` SHA: `AI Code Review` runs on non-draft, non-fork PRs (or manually) to post review comments, and `Doc Auto-Update` runs on pushes to `main/master` (or manually) to open a draft documentation PR with configured labels and serialized concurrency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55e9ed62b15ac772973bb9147b8e09eb46d59fb7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2295, opened 2026-05-07
ci(doc-update): inline upstream workflow, drop cache:pip to unbreak master

frdomovic:
## Summary

The Doc Auto-Update workflow (added in #2234) has been failing on every push to `master` with this error at the `actions/setup-python@v5` step:

> `No file in /home/runner/work/core/core matched to [**/requirements.txt or **/pyproject.toml], make sure you have checked out the target repository`

Failing run example: https://github.com/calimero-network/core/actions/runs/25499553430

## Root cause

The reusable workflow we delegate to — `calimero-network/ai-code-reviewer/.github/workflows/doc-update.yaml@71b0895c` — sets `cache: pip` unconditionally on `actions/setup-python@v5`. The pip cache key is derived from a `requirements.txt` or `pyproject.toml` in the **calling** repository (here, `core`), but `core` is a Rust workspace and has neither file. setup-python errors out before pip can run, the job fails, and no doc-update PR is ever opened.

This is an upstream defect that hits any non-Python consumer of the reusable workflow. It didn't surface during #2234's review because `doc-update.yaml` only triggers on `push: master` — never on PR events. The first run happened post-merge.

## Fix

Inline the upstream's steps directly into our `doc-update.yaml`, copying verbatim from the same SHA we previously pinned, with **one change**: `cache: pip` removed. The pip-install line is pinned to the same SHA used by `ai-review.yaml` so both workflows install the audited version of `ai-code-reviewer`.

### Why inline rather than fork or wait for upstream
- Inline unbreaks CI immediately. Waiting for an upstream PR to merge before fixing master is unnecessary downtime.
- We had already pinned the upstream to a specific SHA, so the "automatic upstream updates" benefit of `uses:` was gone anyway.
- Single file, ~70 lines. Maintainable.

### Follow-up
I'll open a separate PR upstream on `calimero-network/ai-code-reviewer` to make `cache: pip` opt-in (or guard it with a `hashFiles()` check). Once that lands and we bump the install SHA, this file can revert to a thin caller of the reusable workflow.

## Test plan

- [ ] Merge this PR
- [ ] Confirm the next push to `master` triggers Doc Auto-Update and the `setup-python` step succeeds
- [ ] Verify the rest of the workflow (`Install ai-code-reviewer`, `Get merged PR number`, `Generate and open doc update PR`) executes without regressions
- [ ] If a doc-update PR is opened, sanity-check that it targets `master` and has the correct labels

---

> [!NOTE]
> **Medium Risk**
> Changes CI automation that uses `GH_PAT` and opens PRs; misconfiguration could break doc-update runs or affect how PR numbers are detected.
> 
> **Overview**
> The `Doc Auto-Update` GitHub Actions workflow is no longer a thin `uses:` wrapper; it now **inlines** the upstream `ai-code-reviewer` doc-update steps while staying pinned to the same audited SHA.
> 
> As part of the inline, it **removes `cache: pip`** from `actions/setup-python@v5` to stop failures in this Rust repo, adds an optional `workflow_dispatch` `pr_number` input, and updates PR auto-detection to use `gh api /commits/{sha}/pulls` with input sanitization before running `ai-reviewer update-docs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4610b2999fec287be61d413ab6426c1c45fe9224. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/doc-update.yaml
+++ b/.github/workflows/doc-update.yaml
@@ -1,5 +1,17 @@
 name: Doc Auto-Update
 
+# Inlined from calimero-network/ai-code-reviewer reusable workflow at
+# SHA 71b0895c1908504b18f07f2856adfc9c8c8d1e4e, with one fix:
+# `cache: pip` removed from `actions/setup-python@v5`. The upstream sets
+# it unconditionally, but setup-python's pip cache requires a
+# requirements.txt or pyproject.toml in the calling repo — neither
+# exists here (core is Rust), so every run failed with:
+#   No file in /home/runner/work/core/core matched to
+#   [**/requirements.txt or **/pyproject.toml]
+# Tracked for an upstream fix; once that lands and we bump the pinned
+# install SHA, this can revert to a thin caller invoking the reusable
+# workflow.
+
 on:
   push:
     branches: [main, master]
@@ -23,13 +35,62 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  doc-update:
-    # Pinned to a specific SHA on calimero-network/ai-code-reviewer (which
-    # uses `master`, not `main`) to prevent an upstream change from running
-    # with this repo's GH_PAT. Bump deliberately when adopting new features.
-    uses: calimero-network/ai-code-reviewer/.github/workflows/doc-update.yaml@71b0895c1908504b18f07f2856adfc9c8c8d1e4e
-    with:
-      dry_run: ${{ github.event.inputs.dry_run == 'true' }}
-    secrets:
-      ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
-      GH_PAT: ${{ secrets.GH_PAT }}
+  update-docs:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # NOTE: `cache: pip` deliberately omitted — see file-level comment.
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      # Pinned to the same SHA as ai-review.yaml so both workflows install
+      # the audited version of ai-code-reviewer. Bump deliberately.
+      - name: Install ai-code-reviewer
+        run: pip install git+https://github.com/calimero-network/ai-code-reviewer.git@71b0895c1908504b18f07f2856adfc9c8c8d1e4e
+
+      - name: Get merged PR number
+        id: pr
+        env:
+          GH_TOKEN: ${{ secrets.GH_PAT || secrets.GITHUB_TOKEN }}
+        run: |
+          # Prefer explicit input (workflow_dispatch), then commit message, then gh CLI
+          PR_NUM="${{ github.event.inputs.pr_number }}"
+
+          if [ -z "$PR_NUM" ]; then
+            MSG="${{ github.event.head_commit.message }}"
+            PR_NUM=$(echo "$MSG" | grep -oP '(?<=pull request #)\d+' | head -1 || true)
+          fi
+
+          if [ -z "$PR_NUM" ]; then
+            PR_NUM=$(gh pr list \
+              --repo "${{ github.repository }}" \
+              --state merged \
+              --base "${{ github.event.repository.default_branch }}" \
+              --limit 1 \
+              --json number \
+              -q '.[0].number' 2>/dev/null || true)
+          fi
+
+          echo "number=$PR_NUM" >> "$GITHUB_OUTPUT"
+
+      - name: Generate and open doc update PR
+        if: steps.pr.outputs.number != ''
+        env:
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GH_PAT || secrets.GITHUB_TOKEN }}
+        run: |
+          DRY_RUN_FLAG=""
+          if [ "${{ github.event.inputs.dry_run }}" = "true" ]; then
+            DRY_RUN_FLAG="--dry-run"
+          fi
+          ai-reviewer update-docs "${{ github.repository }}" "${{ steps.pr.outputs.number }}" $DRY_RUN_FLAG
+
+      - name: No PR detected
+        if: steps.pr.outputs.number == ''
+        run: echo "Could not detect a merged PR — skipping doc update."
--- a/.github/workflows/doc-update.yaml
+++ b/.github/workflows/doc-update.yaml
@@ -85,9 +85,22 @@ jobs:
           fi
 
           # 2. Otherwise: which PR introduced this exact commit?
+          #    `// empty` in the jq filter prevents JSON `null` (the jq result
+          #    when the API returns []) from being printed as the literal
+          #    string "null". Without it, PR_NUM=null would pass the -z check
+          #    later and we'd hand "null" to ai-reviewer.
+          #    Errors are surfaced (no `2>/dev/null`); a real API failure
+          #    should be visible in logs rather than silently skipping.
           if [ -z "$PR_NUM" ]; then
-            PR_NUM=$(gh api "/repos/$REPO/commits/$SHA/pulls" \
-              --jq '.[0].number' 2>/dev/null || true)
+            if ! PR_NUM=$(gh api "/repos/$REPO/commits/$SHA/pulls" --jq '.[0].number // empty'); then
+              echo "::warning::gh api commits/{sha}/pulls failed; doc update will be skipped"
+              PR_NUM=""
+            fi
+          fi
+
+          # Final sanity check: only forward digit-only values.
+          if [[ ! "$PR_NUM" =~ ^[0-9]+$ ]]; then
+            PR_NUM=""
           fi
 
           echo "number=$PR_NUM" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/doc-update.yaml
+++ b/.github/workflows/doc-update.yaml
@@ -60,34 +60,34 @@ jobs:
 
       - name: Get merged PR number
         id: pr
-        # Untrusted values (commit message, dispatch inputs) flow through `env`
-        # so the shell only sees `"$VAR"` references — never string-interpolated
-        # via `${{ ... }}` directly into a script. Without this, a merged PR
-        # whose commit message contained backticks or `$(...)` could execute
-        # arbitrary commands with this job's `contents: write` /
-        # `pull-requests: write` permissions.
+        # All event-derived values flow through `env` so the shell only sees
+        # `"$VAR"` references; nothing is `${{ ... }}`-interpolated directly
+        # into a `run` block.
+        #
+        # PR detection uses the `commits/{sha}/pulls` API rather than grepping
+        # the commit message + falling back to `gh pr list --limit 1`. The
+        # API returns the PR that *introduced* this commit to the default
+        # branch, which is correct for both merge commits and squash/rebase
+        # merges — and unlike the time-ordered list, it doesn't race with
+        # other PRs that merge in the same window.
         env:
           GH_TOKEN: ${{ secrets.GH_PAT || secrets.GITHUB_TOKEN }}
           INPUT_PR_NUMBER: ${{ github.event.inputs.pr_number }}
-          HEAD_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
           REPO: ${{ github.repository }}
-          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+          SHA: ${{ github.sha }}
         run: |
-          # Prefer explicit input (workflow_dispatch), then commit message, then gh CLI
-          PR_NUM="$INPUT_PR_NUMBER"
-
-          if [ -z "$PR_NUM" ]; then
-            PR_NUM=$(echo "$HEAD_COMMIT_MESSAGE" | grep -oP '(?<=pull request #)\d+' | head -1 || true)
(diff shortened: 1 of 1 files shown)


frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 85% | Review time: 56.6s

---

### ✅ No Issues Found

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-9e0072a9`</sub>

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,76 @@ concurrency:
...
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GH_PAT || secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ steps.pr.outputs.number }}
+          DRY_RUN: ${{ github.event.inputs.dry_run }}
+        run: |
+          DRY_RUN_FLAG=""
+          if [ "$DRY_RUN" = "true" ]; then
+            DRY_RUN_FLAG="--dry-run"
+          fi
```
frdomovic, line 122: 🟡 **Unquoted variable in shell command allows word splitting**

On line 112, `$DRY_RUN_FLAG` is unquoted in the `ai-reviewer update-docs` invocation: `ai-reviewer update-docs "$REPO" "$PR_NUMBER" $DRY_RUN_FLAG`. While `DRY_RUN_FLAG` is only ever set to `""` or `"--dry-run"` within the same script, the unquoted form is a style inconsistency that could become a vulnerability if the flag logic is ever extended to include user-controlled content. More critically, `$PR_NUMBER` comes from `steps.pr.outputs.number` which is derived from `gh pr list` output or a commit message grep — if that output ever contained spaces or shell metacharacters, the quoting of `"$PR_NUMBER"` protects it, but the pattern is worth noting. The immediate risk is low since `PR_NUMBER` is filtered through `grep -oP '\d+'`, but the unquoted `$DRY_RUN_FLAG` is a minor inconsistency.

**Suggested fix:**
```
Quote the flag variable for consistency and safety: `ai-reviewer update-docs "$REPO" "$PR_NUMBER" ${DRY_RUN_FLAG:+"$DRY_RUN_FLAG"}`  or simply ensure it is always quoted: `ai-reviewer update-docs "$REPO" "$PR_NUMBER" "$DRY_RUN_FLAG"` (an empty string argument is harmless if the tool ignores empty args, otherwise use the conditional expansion).
```

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,76 @@ concurrency:
...
+        with:
+          python-version: "3.11"
+
+      # Pinned to the same SHA as ai-review.yaml so both workflows install
+      # the audited version of ai-code-reviewer. Bump deliberately.
+      - name: Install ai-code-reviewer
+        run: pip install git+https://github.com/calimero-network/ai-code-reviewer.git@71b0895c1908504b18f07f2856adfc9c8c8d1e4e
+
+      - name: Get merged PR number
+        id: pr
```
frdomovic, line 62: 🟡 **HEAD_COMMIT_MESSAGE context value is interpolated into env before shell sees it**

The `HEAD_COMMIT_MESSAGE` env var is set via `${{ github.event.head_commit.message }}`. GitHub Actions evaluates `${{ ... }}` expressions before the shell runs, so the commit message is string-substituted into the YAML value of the `env` block. A commit message containing YAML-special characters (e.g., a newline followed by `ANOTHER_VAR: injected`) could potentially corrupt the env block. In practice GitHub Actions quotes env values, but the safer pattern for untrusted multi-line content is to use `toJSON()` or write it to a file rather than embedding it in an env string. The subsequent shell usage (`echo "$HEAD_COMMIT_MESSAGE" | grep ...`) is safe because the variable is accessed via `"$VAR"`, but the YAML-level injection vector exists at assignment time.

**Suggested fix:**
```
Use `toJSON` to safely embed the value: `HEAD_COMMIT_MESSAGE: ${{ toJSON(github.event.head_commit.message) }}` — or write the commit message to a temp file using a prior step and read it from there.
```

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,76 @@ concurrency:
...
+          # Prefer explicit input (workflow_dispatch), then commit message, then gh CLI
+          PR_NUM="$INPUT_PR_NUMBER"
+
+          if [ -z "$PR_NUM" ]; then
+            PR_NUM=$(echo "$HEAD_COMMIT_MESSAGE" | grep -oP '(?<=pull request #)\d+' | head -1 || true)
+          fi
+
+          if [ -z "$PR_NUM" ]; then
+            PR_NUM=$(gh pr list \
+              --repo "$REPO" \
```
frdomovic: 💡 **gh pr list fallback may return a non-merged PR or wrong PR**

The third fallback for PR detection uses `gh pr list --state merged --base "$DEFAULT_BRANCH" --limit 1`. This returns the most recently merged PR regardless of whether it corresponds to the current push event. On a busy repository where multiple PRs merge in quick succession, this fallback could associate the doc-update with the wrong PR. The commit-message grep (second fallback) is more reliable for standard GitHub squash-merge messages, but the `gh pr list` fallback introduces a race condition.

**Suggested fix:**
```
Consider removing the `gh pr list` fallback entirely and only using the commit message parse. If no PR number is found, log a warning and skip rather than potentially associating with the wrong PR. Alternatively, use `gh api /repos/{owner}/{repo}/commits/{sha}/pulls` to get PRs associated with the exact commit SHA.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 27% | Review time: 57.6s

🟡 2 warnings, 💡 1 suggestions. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-0e20ba90`</sub>

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 27% | Review time: 51.8s

✅ 3 fixed | 🆕 3 new. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-fc893e70`</sub>

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,76 @@ concurrency:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      # NOTE: `cache: pip` deliberately omitted — see file-level comment.
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      # Pinned to the same SHA as ai-review.yaml so both workflows install
+      # the audited version of ai-code-reviewer. Bump deliberately.
```
frdomovic, line 57: 🟡 **pip install from git URL without hash verification**

Line 57 installs the ai-code-reviewer package directly from a GitHub git URL pinned to a SHA: `pip install git+https://github.com/calimero-network/ai-code-reviewer.git@71b0895c...`. While pinning to a commit SHA is better than a branch name, pip does not verify the integrity of the downloaded archive against a known hash (unlike `pip install` with a `--hash` flag or a lockfile). If the GitHub repository or CDN were compromised and the commit SHA were force-pushed (which GitHub allows on non-protected refs), the installed code could differ. For a workflow that has `contents: write` and `pull-requests: write` permissions and access to `ANTHROPIC_API_KEY` and `GH_PAT`, this is a meaningful supply-chain risk.

**Suggested fix:**
```
Consider using a requirements.txt with `--require-hashes` or verify the installed package checksum after installation. At minimum, document the known-good SHA alongside the expected package version in a comment.
```

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,76 @@ concurrency:
...
+              --jq '.[0].number' 2>/dev/null || true)
+          fi
+
+          echo "number=$PR_NUM" >> "$GITHUB_OUTPUT"
+
+      - name: Generate and open doc update PR
+        if: steps.pr.outputs.number != ''
+        # Same defense-in-depth pattern as above: untrusted-ish values go
+        # through `env` so the shell never string-interpolates them.
+        env:
```
frdomovic, line 112: 🟡 **PR_NUMBER from step output used directly in shell command without quoting**

At line 101, `$PR_NUMBER` is passed as a positional argument to `ai-reviewer update-docs "$REPO" "$PR_NUMBER"`. While `PR_NUMBER` is validated as digits-only in the earlier step (line 78), the value flows through `steps.pr.outputs.number` which is set via `echo "number=$PR_NUM" >> "$GITHUB_OUTPUT"`. The digit-only regex guard is correct, but the value is still interpolated into the shell command without double-quoting. If the regex guard were ever weakened or bypassed, this could allow command injection. The fix is trivial: always quote `"$PR_NUMBER"` in the run block.

**Suggested fix:**
```
Change line 101 to: ai-reviewer update-docs "$REPO" "$PR_NUMBER" "${DRY_RUN_FLAG[@]}"  (PR_NUMBER is already unquoted in the diff; add quotes around it for defense-in-depth)
```

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,76 @@ concurrency:
...
+        # merges — and unlike the time-ordered list, it doesn't race with
+        # other PRs that merge in the same window.
+        env:
+          GH_TOKEN: ${{ secrets.GH_PAT || secrets.GITHUB_TOKEN }}
+          INPUT_PR_NUMBER: ${{ github.event.inputs.pr_number }}
+          REPO: ${{ github.repository }}
+          SHA: ${{ github.sha }}
+        run: |
+          # 1. Explicit dispatch input wins, but only if it parses as digits.
+          #    Anything else is dropped so we fall through to auto-detect
```
frdomovic, line 80: 💡 **gh api error output silenced; failed API call produces empty PR_NUM silently**

Lines 82-84 use `2>/dev/null || true` to suppress errors from the `gh api` call. If the API call fails for a transient reason (rate limit, network error, bad token), `PR_NUM` will be empty and the workflow will silently skip the doc update with the message 'Could not detect a merged PR — skipping doc update.' This makes failures invisible and hard to diagnose. Consider logging the error or failing the step on unexpected API errors.

**Suggested fix:**
```
Separate the error suppression: capture stderr to a variable, log it, and only treat 'no PR found' (empty result) as a skip condition rather than any API failure.
```

.github/workflows/doc-update.yaml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/doc-update.yaml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/doc-update.yaml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/doc-update.yaml
```diff
@@ -23,13 +39,89 @@ concurrency:
...
+          fi
+
+          echo "number=$PR_NUM" >> "$GITHUB_OUTPUT"
+
+      - name: Generate and open doc update PR
+        if: steps.pr.outputs.number != ''
+        # Same defense-in-depth pattern as above: untrusted-ish values go
+        # through `env` so the shell never string-interpolates them.
+        env:
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
```
frdomovic, line 113: 🟡 **PR_NUMBER from step output interpolated directly into shell command arguments**

At line 116, `"$PR_NUMBER"` is passed as a positional argument to `ai-reviewer update-docs`. `PR_NUMBER` is set from `${{ steps.pr.outputs.number }}` in the `env` block (line 110). Although the step that produces the output validates it as digits-only (the `^[0-9]+$` regex check), the value is still re-injected via `${{ steps.pr.outputs.number }}` in the `env` block rather than being read from a file or a more isolated mechanism. If the output sanitization were ever bypassed or the step skipped, the value would flow into a shell command. The current digit-only guard is sound, but the defense-in-depth comment on line 107 is slightly misleading: the value IS still interpolated by the GitHub Actions expression engine into the `env` block before the shell sees it, so it is not fully isolated from the expression layer.

**Suggested fix:**
```
The current approach is acceptable given the digit-only validation. Consider adding a comment clarifying that `${{ steps.pr.outputs.number }}` is still evaluated by the Actions expression engine before the shell runs, so the digit-only guard in the previous step is the real protection here.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 25% | Review time: 56.4s

✅ 5 fixed | 🆕 1 new | ⏳ 1 open. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-ed339cd3`</sub>

.github/workflows/doc-update.yaml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

.github/workflows/doc-update.yaml
frdomovic: ✅ **No longer detected** - This issue was not re-detected after the latest changes.

PR #2340, opened 2026-05-12
feat(context): signed cut + canonical state hashes on MemberRemoved /…

chefsale:
… MemberLeft

`MemberRemoved` and `MemberLeft` now carry three signed claims that let receivers detect divergence after apply:

1. `cut: GovernancePosition` — the namespace governance-DAG position the signer signed against. Receivers use this as the descend-from boundary for the apply-time membership check on subsequent deltas from the removed/left member, so two peers at different governance positions evaluate the same answer.

2. `expected_group_state_hash: [u8; 32]` — what the governance state hash will be after the apply, computed pre-apply via a pure in-memory simulation of the removal.

3. `expected_context_state_hashes: Vec<(ContextId, [u8; 32])>` — per-context CRDT root hash at sign time, sorted by `context_id` for deterministic op-content hashing. Catches the partition-window case where a receiver applied legitimate pre-removal state-DAG deltas from the now-removed member that the admin's view didn't include.

Sign-time precomputation:
* `compute_group_state_hash_after_remove(store, group_id, member)` — pure function, simulates the removal by filtering the sorted member list in-memory and hashing. Idempotent on non-members.
* `snapshot_context_state_hashes(store, group_id)` — enumerates registered contexts and reads each `ContextMeta::root_hash`. Unmaterialized contexts (registered but no meta row yet) are skipped, not zero-hashed.
* `build_governance_cut(store, group_id)` — bundles the pre-apply group state hash with the namespace DAG heads. Shared by the admin-removal path and both self-leave paths.

Apply-time verification:
* `verify_post_apply_state_hashes` runs after the apply path completes for both ops. On mismatch it emits a structured warn log carrying the group id, op kind, both hashes, and the list of divergent context ids — does NOT roll back the apply or reject the op. The admin's view is canonical by definition; reconcile-via-anchor sync is the recovery path (separate work).
* An all-zero `expected_group_state_hash` plus an empty `expected_context_state_hashes` is treated as "no claim made" and skips the check — keeps test helpers using placeholder claims from triggering spurious warnings. Once the network has fully rolled forward to signed-claim ops this branch becomes dead.

Schema version bumped from 3 to 4 (hard cut, no backwards compat per the pre-1.0 wire-change policy).

Tests:
* 5 new unit tests covering precompute equivalence with post-apply hash, non-member idempotency, snapshot sort order, snapshot skip for unmaterialized contexts, and cut carrying the pre-apply hash.
* All 233 existing context-lib tests + 22 convergence integration tests still pass — destructuring sites updated with `..` and test helpers (`dummy_member_removed_op`, `dummy_member_removed`) added for sites that don't exercise the divergence path.

Receiver-side reconcile-via-anchor (consumes these signed claims and adopts canonical state) is the next planned PR.

---

> [!NOTE]
> **Medium Risk**
> Medium risk due to a wire/schema bump to v4 (old ops now rejected) and new sign/apply-time hashing logic in the membership governance path that could surface unexpected divergence warnings if assumptions about ordering/materialization are wrong.
> 
> **Overview**
> `MemberRemoved` and `MemberLeft` are expanded to include a signed `cut: GovernancePosition`, an `expected_group_state_hash`, and `expected_context_state_hashes`, and the `SignedGroupOp` schema version is bumped to `v4` (older shapes are rejected).
> 
> Publishing `MemberRemoved`/self-`MemberLeft` now precomputes these values at sign time via new helpers (`build_governance_cut`, `compute_group_state_hash_after_remove`, `snapshot_context_state_hashes`). Apply-time paths recompute post-apply hashes and **log structured warnings on mismatch** (no rollback) using a new linear diff of per-context hashes.
> 
> Tests are updated to construct v4 removal ops (dummy helpers where divergence claims aren’t under test) and new unit tests are added to pin the precompute/verify invariants and snapshot ordering/skip behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbac55409b79ef852528e8c158d1e80520503e62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/primitives/src/local_governance/mod.rs
+++ b/crates/context/primitives/src/local_governance/mod.rs
@@ -113,7 +113,11 @@ pub enum GroupOp {
     /// Carries the same three convergence claims as `MemberRemoved`,
     /// signed by the leaver. The leaver is honest by definition for
     /// self-leave; a Byzantine leaver claiming false canonical hashes
-    /// is caught by other peers' hash verification on apply.
+    /// triggers a divergence warning on apply at each receiver but the
+    /// op is **not rejected** — the apply path logs the mismatch and
+    /// moves on, leaving the canonical view to be re-established by an
+    /// admin-signed `MemberRemoved` or anchor-sync reconcile. The
+    /// signed hashes are a detection signal, not an adoption gate.
     MemberLeft {
         member: PublicKey,
         cut: GovernancePosition,
--- a/crates/context/primitives/src/local_governance/mod.rs
+++ b/crates/context/primitives/src/local_governance/mod.rs
@@ -94,6 +94,15 @@ pub enum GroupOp {
     /// Apply-time mismatch is logged as a structured warning but does
     /// not roll back the apply; the canonical view is the admin's, and
     /// reconciliation against an anchor is the recovery path.
+    ///
+    /// **Two different group-state hashes travel in this op, by
+    /// design:** `cut.group_state_hash` is the **pre-apply** snapshot
+    /// (the state the signer signed against — used by receivers as
+    /// the descend-from boundary for the membership walk), while
+    /// `expected_group_state_hash` is the **post-apply** simulation
+    /// (used by receivers as the convergence target after they apply
+    /// the op). They have different values, different roles, and
+    /// different consumers; do not collapse them.
     MemberRemoved {
         member: PublicKey,
         cut: GovernancePosition,
--- a/crates/context/primitives/src/local_governance/mod.rs
+++ b/crates/context/primitives/src/local_governance/mod.rs
@@ -12,7 +12,7 @@
 //! group-scoped ops they cannot decrypt.
 
 use borsh::{BorshDeserialize, BorshSerialize};
-use calimero_context_config::types::SignedGroupOpenInvitation;
+use calimero_context_config::types::{GovernancePosition, SignedGroupOpenInvitation};
 use calimero_primitives::application::ApplicationId;
 use calimero_primitives::context::{ContextId, GroupMemberRole, UpgradePolicy};
 use calimero_primitives::identity::{PrivateKey, PublicKey};
@@ -31,6 +31,12 @@ pub use ack_router::AckRouter;
 
 /// Wire/schema version for [`SignedGroupOp`].
 ///
+/// v4: `MemberRemoved` and `MemberLeft` now carry the signed governance-DAG
+/// cut, the expected post-apply group state hash, and the expected
+/// post-apply per-context CRDT state hashes. Receivers detect divergence by
+/// comparing their computed hashes against the signed values after apply.
+/// Pre-1.0 wire change; old-shape ops are rejected outright.
+///
 /// v3: Added `state_hash: [u8; 32]` — each op commits to the group's
 /// authorization-relevant state at signing time. On apply, a non-zero
 /// state hash is verified against the current state to reject stale ops.
@@ -40,7 +46,7 @@ pub use ack_router::AckRouter;
 ///
 /// v1 was internal to feature branch development and never deployed to any
 /// persistent network. No backward-compatible deserialization is needed.
-pub const SIGNED_GROUP_OP_SCHEMA_VERSION: u8 = 3;
+pub const SIGNED_GROUP_OP_SCHEMA_VERSION: u8 = 4;
 
 /// Domain separation prefix for Ed25519 signatures over group ops.
 pub const GROUP_GOVERNANCE_SIGN_DOMAIN: &[u8] = b"calimero.group.v1";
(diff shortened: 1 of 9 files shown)


crates/context/src/group_store/mod.rs
```diff
@@ -782,6 +783,109 @@ const MAX_DAG_HEADS: usize = 64;
...
+            return;
+        }
+    };
+
+    let actual_context_state_hashes = match snapshot_context_state_hashes(store, group_id) {
+        Ok(v) => v,
+        Err(err) => {
+            tracing::warn!(
+                group_id = %hex::encode(group_id.to_bytes()),
+                op_kind,
```
frdomovic, line 939: 🟡 **No-claim sentinel allows bypassing divergence detection with a non-zero hash and empty context list**

The sentinel check `if *expected_group_state_hash == [0u8; 32] && expected_context_state_hashes.is_empty()` skips verification only when BOTH conditions hold. A sender that computes a real `expected_group_state_hash` but sends an empty `expected_context_state_hashes` (e.g. a group with no registered contexts) will NOT be treated as a no-claim and will trigger the full check. This is correct behavior. However, the inverse is also true: a Byzantine sender could send `[0u8; 32]` as the group hash but include non-empty context hashes — this would NOT be treated as a no-claim and would run the check with a zero group hash, which will almost certainly diverge and emit a spurious warn. The sentinel semantics are slightly asymmetric and the comment says 'all-zero AND empty' but the real escape hatch for a group with no contexts is 'all-zero hash AND empty list'. This is a minor logic inconsistency worth documenting more precisely, but not a security issue since mismatches only warn.

**Suggested fix:**
```
Consider documenting explicitly that a group with zero registered contexts and a real post-apply hash will correctly trigger the check (the context list comparison will be empty on both sides and only the group hash matters). The current sentinel is fine for its stated purpose but the comment could be clearer about the 'no contexts' case.
```

crates/context/src/group_store/mod.rs
```diff
@@ -782,6 +783,109 @@ const MAX_DAG_HEADS: usize = 64;
...
+        match actual_map.get(cid) {
+            Some(actual) if actual == expected => {}
+            _ => divergent.push(*cid),
+        }
+    }
+    for cid in actual_map.keys() {
+        if !expected_map.contains_key(cid) {
+            divergent.push(*cid);
+        }
+    }
```
frdomovic, line 968: 💡 **Two BTreeMap allocations for divergence detection in verify_post_apply_state_hashes**

The divergence check builds two `BTreeMap<ContextId, [u8; 32]>` from slices that are already sorted (both `expected_context_state_hashes` and `actual_context_state_hashes` are sorted by `context_id`). Since both inputs are sorted, a linear merge-scan would detect divergences in O(n) time and O(1) extra space instead of O(n log n) time and O(n) space for two BTreeMap constructions. This is a hot path on every `MemberRemoved`/`MemberLeft` apply.

**Suggested fix:**
```
Replace the two BTreeMap constructions with a merge of two sorted iterators: iterate both slices in parallel, advancing the pointer on the side with the smaller key, and record divergences when keys don't match or hashes differ. This is O(n) time and O(divergent.len()) space.
```

crates/context/src/group_store/meta.rs
```diff
@@ -92,3 +97,108 @@ pub fn compute_group_state_hash(store: &Store, group_id: &ContextGroupId) -> Eyr
...
+    group_id: &ContextGroupId,
+    removed_member: &PublicKey,
+) -> EyreResult<[u8; 32]> {
+    let meta = load_group_meta(store, group_id)?
+        .ok_or_else(|| eyre!("group not found for state hash computation"))?;
+
+    let mut members = list_group_members(store, group_id, 0, usize::MAX)?;
+    members.retain(|(pk, _role)| pk != removed_member);
+    members.sort_by(|a, b| a.0.cmp(&b.0));
+
```
frdomovic, line 169: 💡 **compute_group_state_hash_after_remove sorts after retain, but compute_group_state_hash sorts before — ordering is consistent but fragile**

In `compute_group_state_hash`, `list_group_members` returns members and then `members.sort_by(...)` is called. In `compute_group_state_hash_after_remove`, `retain` is called first and then `sort_by`. Both produce the same sorted result, but the two functions have slightly different orderings of operations (sort-then-hash vs retain-sort-then-hash). This is correct but could diverge if someone adds a step between retain and sort in one function but not the other. The test `compute_group_state_hash_after_remove_matches_post_apply_hash` catches this, but the structural divergence between the two functions is a maintenance risk.

**Suggested fix:**
```
Consider extracting the hash-computation loop into a shared private helper `hash_members(hasher, meta, members_sorted)` that both functions call after preparing their respective member lists, ensuring the hashing logic can never diverge between the two paths.
```

crates/context/src/group_store/meta.rs
```diff
@@ -92,3 +97,108 @@ pub fn compute_group_state_hash(store: &Store, group_id: &ContextGroupId) -> Eyr
...
+    }
+    entries.sort_by(|a, b| a.0.cmp(&b.0));
+    Ok(entries)
+}
+
+/// Build the current cross-DAG cut for `group_id` — pre-apply group
+/// state hash bundled with the namespace governance DAG heads at this
+/// moment. Used as the `cut` field in `MemberRemoved` / `MemberLeft`
+/// ops so receivers' apply-time membership check evaluates against the
+/// same descend-from boundary the signer signed against (not against
```
frdomovic, line 225: 💡 **snapshot_context_state_hashes skips unmaterialized contexts silently — asymmetric behavior between admin and receiver**

The function skips contexts whose `ContextMeta` row is missing (not yet materialized). The doc comment explains this is intentional to avoid zero-hashing. However, this creates an asymmetry: if the admin node has materialized a context but the receiver hasn't, the receiver's snapshot will be shorter than the admin's, and the divergence check will fire (the admin's entry won't be in the receiver's actual_map). Conversely, if the receiver has materialized a context the admin hasn't, the receiver's actual_map will have an extra entry. Both cases correctly fire the divergence warning. The concern is that on a freshly-joined node, nearly all contexts may be unmaterialized, causing the admin's snapshot to be much larger than the receiver's, generating many spurious divergence warnings on every MemberRemoved until the node catches up. This is by design (reconcile-via-anchor will heal), but the operational noise could be significant.

**Suggested fix:**
```
Consider adding a log message at debug level when contexts are skipped due to missing ContextMeta, so operators can distinguish 'divergence due to partition' from 'divergence due to fresh node not yet materialized'. Alternatively, document this expected noise in the operator runbook.
```

crates/context/primitives/src/local_governance/mod.rs
```diff
@@ -60,7 +66,40 @@ pub enum GroupOp {
...
+    /// - `expected_context_state_hashes`: per-context CRDT root hash at
+    ///   sign time, one entry per context registered under `group_id`,
+    ///   sorted by `context_id` for deterministic op-content hashing.
+    ///   Catches the case where a receiver applied legitimate
+    ///   pre-removal state-DAG deltas from the now-removed member that
+    ///   never reached the admin — the receiver's post-apply context
+    ///   hash differs, divergence is detectable, anchor-sync reconcile fires.
+    ///
+    /// Apply-time mismatch is logged as a structured warning but does
+    /// not roll back the apply; the canonical view is the admin's, and
```
frdomovic, line 95: 📝 **MemberLeft doc comment says 'Byzantine leaver claiming false canonical hashes is caught by other peers' hash verification' — this is not enforced by the current code**

The doc comment for `MemberLeft` states 'a Byzantine leaver claiming false canonical hashes is caught by other peers' hash verification on apply.' However, `verify_post_apply_state_hashes` only emits a `warn` log on mismatch — it does NOT reject the op. So a Byzantine leaver can claim any hashes and the op will still be applied. The comment overstates the security guarantee: the mismatch is detected and logged, but not acted upon. This could mislead future developers into thinking the system rejects ops with wrong hashes.

**Suggested fix:**
```
Update the doc comment to accurately state: 'A Byzantine leaver claiming false canonical hashes will trigger a divergence warning on apply at each receiver, but the op is not rejected — reconcile-via-anchor sync is the recovery path.'
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 25% | Review time: 90.5s

🟡 1 warnings, 💡 3 suggestions, 📝 1 nitpicks. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-d24c479f`</sub>

Issue #2330, opened 2026-05-11
node/state-delta: StateDeltaActor arbiter thread wedges on synchronous merkle/CRDT apply — timeout can't preempt it, no producer-side recovery (#2199/#2299 follow-up)


rtb-12:
## Summary

The dedicated `StateDeltaActor` (added in #2299 / PR #2315) can have its arbiter thread **stuck for an unbounded wall-clock time** in the synchronous delta-apply path — and unlike `SyncManager` (which got a watchdog in #2319/PR #2322), nothing on the producer side recovers if it wedges. The `STATE_DELTA_PROCESSING_TIMEOUT` / `WASM_APPLY_TIMEOUT` that PR #2315 added is a `tokio::time::timeout`, which **cannot preempt a thread that is busy in synchronous (non-`.await`) code** — so it's only a partial mitigation.

## Mechanism

`StateDeltaActor` processes jobs one at a time on its own arbiter. The actual apply work — `Interface::apply_action` → `handle_entity_push` / `apply_leaf_with_crdt_merge`, plus merkle-tree recomputation (`Index::add_child_to`, #2238) — runs inside a `with_runtime_env(...)` synchronous closure with no yield points, and its cost scales ~O(storage size) per entity → ~O(n²) per batch (#2199). When that work runs long (a heavily-divergent node under load), the actor's executor is blocked: the `tokio::time::timeout(STATE_DELTA_PROCESSING_TIMEOUT, …)` future never gets polled, never fires, and the `ctx.run_interval` 60 s summary never logs either. The actor is effectively dead until the synchronous work finishes (or, if jobs arrive faster than they drain, indefinitely). Inbound `BroadcastMessage::StateDelta` jobs keep piling into the bounded mailbox behind it.

## Evidence

Fuzzy-load-test run `25663237150` (PR #2322, suite `kv-store-with-handlers`), node-4 — `/tmp/fuzzylogs/docker-logs/fuzzy-handlers-node-4.log` (artifact `fuzzy-test-handlers-463-25663237150`):

- Last `StateDelta worker completed`: `2026-05-11T10:11:04.256Z`. After that — **nothing** from the `StateDeltaActor` for the remaining ~14.5 min of the run: no `StateDelta worker completed`, no actor summary line.
- `Matched StateDelta message` (gossipsub delivery into the actor's mailbox) keeps firing the whole run — so jobs were arriving, just never processed.
[...]

Issue #2299, opened 2026-05-08
node: per-variant actor isolation for BroadcastMessage dispatch (state-delta drains in bursts, blocks cross-node reads)


rtb-12:
## Summary

Cross-context state-delta processing on a node is bursty: gossipsub messages are received reliably and on time, but the downstream `Matched StateDelta` dispatch on the receiving node drains in clumps (hundreds at a time, ~90-180 seconds apart) instead of steadily. This causes cross-node reads done within a short window after a write (e.g., the 2-second wait used by the `cross_node_handler_check` operation in `kv-store-with-handlers/fuzzy-test.yml`) to occasionally see `None` for keys that have already been published — the delta is sitting in a queue, not yet applied.

In effect, the kv-store-with-handlers fuzzy load test sits right at its 95% success threshold (master: 95.1% PASSED, with 116 cross-node ✗ in the latest passing run; PR #2271 hit 94.6% with 126 ✗). A 0.5% run-to-run swing tips it from passing to failing. The shape of the failures (set on node-2 → 2 s wait → get on node-4 returns `None`, but node-4 already received the gossipsub message ~600µs after publication) makes this a downstream-processing problem, not a network-layer problem.

## Evidence — single failure, end-to-end timeline

(Captured from the 2026-05-03 fuzzy run on PR #2271, artifact `fuzzy-test-handlers-384-25271173533`.)

The set was on node-2: `set("handler_consistency_85c2484e-d457-4486-815d-bc7b6fe83e04", "hcheck_1777787992")`.

| t (UTC) | Where | Event |
|---|---|---|
| `05:59:52.099` | node-2 | `Received execution request method=set` |
| `05:59:52.102` | node-2 | `Persisted delta to database`, `Broadcasting state delta` |
| `05:59:52.10234` | node-2 | gossipsub `Published message` `id=…3540` |
| `05:59:52.10290` | **node-4** | gossipsub `Message { source: PeerId("…GM8s…"), … }` ← **received in ~600µs** |
| `05:59:52.10295` | node-4 | `network_event_processor: Forwarding network event to NodeManager` |
| `05:59:54.115` | node-4 | `method=get` reads `handler_consistency_…` → returns `null` |
| `05:59:54.117` | node-4 | runtime returns `None` to JSON-RPC client |
| **`06:00:48.816`** | **node-4** | `Received state delta … expected_root_hash=148sUkYxcf…` ← only now (54 s later, via background DAG sync, not the gossipsub fast-path) |

[...]

PR #2341, opened 2026-05-12
fix(node): don't abandon a spawn_blocking WASM apply on timeout (#2330)

rtb-12:
## Summary

Addresses #2330. The state-delta apply path wrapped `context_client.execute("__calimero_sync_next", …)` in a hard `tokio::time::timeout` — `WASM_APPLY_TIMEOUT` (15s) in `delta_store.rs`, `STATE_DELTA_PROCESSING_TIMEOUT` (60s) in `state_delta_bridge.rs`. But the WASM merge-apply it drives runs on a **`spawn_blocking` thread** (`crates/context/src/handlers/execute/mod.rs:1521`), and `spawn_blocking` tasks **cannot be cancelled**. So when the timeout fired:

1. the future was dropped → the caller's DAG write lock (held by `add_delta_with_outcome` around `apply().await`) was released and the delta recorded as **not applied**;
2. but `ContextManager`'s `ExecuteRequest` handler kept running `internal_execute` → `module.run` finished → `storage.commit()` + `context.root_hash = new` + `local_delta_tx.send(…)` landed **anyway, late** — after the lock was gone and the next delta's apply may already have started.

Net: storage holds a delta the DAG doesn't know about → it gets re-synced and re-applied → divergent root hash, plus a stale `root_hash` overwrite racing the gap. (The 60s actor-level timeout has the same shape one level up.) This is one of the latent hazards behind the #2330 / fuzzy-flake symptoms.

`module.run`'s writes all go into a `Temporal` buffer (`ContextStorage`) flushed only on `commit()`, and `module.run` is gas-bounded so it always terminates — so the **correct** behaviour, given we can't cancel it, is to *not abandon it*: keep waiting, let it commit properly, hold the per-context DAG write lock for its duration. The "don't pin the DAG write lock on a slow merge-apply" concern that motivated the timeout (#2186 / #2199) is now largely closed by #2238 (binary-search insert, batched ancestor-hash recompute, read-don't-rehash).

## Change

- `delta_store.rs::ContextStorageApplier::apply` — replace the hard `timeout` around `execute()` with a one-shot race against `WASM_APPLY_TIMEOUT` that only **warns** if it fires, then awaits `execute()` to completion. `WASM_APPLY_TIMEOUT` is now a soft warning threshold; doc comment updated.
- `state_delta_bridge.rs::Handler<StateDeltaJob>` — same for `STATE_DELTA_PROCESSING_TIMEOUT` around `handle_state_delta`: warn + bump `over_budget_total` (renamed from `timeout_total`, since "timeout" is no longer accurate), never abandon. Dropped the `Err(_elapsed)` arm and the "dropping delta" log line accordingly; summary log field renamed.
- Heavy comments at both sites explaining why a hard timeout here is unsafe, so it doesn't get re-added.

No behaviour change on the happy path. The durable fix for the underlying merge-apply slowness remains #2199 / #2238.

## Verification

- `cargo build -p calimero-node` — clean
- `cargo test -p calimero-node` — all pass
- `cargo fmt --check`, `cargo clippy --lib --tests` — clean on touched files

There's no good unit test for the timing itself; the apply path's behaviour under slow merges is exercised by the `kv-store-with-handlers` fuzzy soak (#2299 acceptance), which is master-only.

---

> [!NOTE]
> **Medium Risk**
> Touches the state-delta apply path and changes timeout semantics while holding the DAG write lock; correctness improves but long-running applies can now block longer, so reviewers should consider performance/operational impact.
> 
> **Overview**
> Prevents the node from **abandoning** a `spawn_blocking` WASM merge-apply when a `tokio::time::timeout` fires, which could previously let the DAG write lock drop while storage commits still landed, risking DAG/storage divergence.
> 
> `delta_store.rs` and `state_delta_bridge.rs` now treat `WASM_APPLY_TIMEOUT` / `STATE_DELTA_PROCESSING_TIMEOUT` as **soft budgets**: they race the budget to emit a warning (and increment a new `over_budget_total` counter in the actor), then continue awaiting the original future to completion instead of returning an error/dropping the job.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64377ebfafd3951158586df2dc4ec3fa43d4e973. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/delta_store.rs
+++ b/crates/node/src/delta_store.rs
@@ -42,10 +42,17 @@ use crate::sync::rotation_log_reader;
 /// across restarts. Per #2272 review.
 const MAX_TOPOLOGY_ENTRIES: usize = 10_000;
 
-/// Bound on `context_client.execute(...)` inside `apply()`. The caller
-/// holds the DAG write lock during this await; without a bound, a slow
-/// WASM merge-apply (#2199) propagates as a 30s+ lock hold. Returning
-/// `ApplyError` here releases the lock and lets sync recovery retry.
+/// Soft budget for `context_client.execute(...)` inside `apply()`. The
+/// caller holds the DAG write lock during this await; a slow WASM
+/// merge-apply (#2199) therefore pins it. This is *not* a hard cap: the
+/// apply runs on a `spawn_blocking` thread that cannot be cancelled, and
+/// abandoning it here would release the DAG write lock while the apply
+/// completes and `commit()`s its storage writes anyway — late, racing the
+/// next delta (the delta would be in storage but absent from the DAG, then
+/// re-synced and re-applied → divergent root hash). The merge-apply is
+/// gas-bounded so it terminates, and post-#2238 is fast enough that holding
+/// the lock for its duration is acceptable; exceeding this threshold only
+/// produces a warning. See #2199 / #2238.
 const WASM_APPLY_TIMEOUT: Duration = Duration::from_secs(15);
 
 /// Result of adding a delta with cascaded event information
@@ -222,26 +229,39 @@ impl DeltaApplier<Vec<Action>> for ContextStorageApplier {
 
         let wasm_start = std::time::Instant::now();
 
-        // Bounded so the caller's DAG write lock can't be pinned by a
-        // slow WASM merge-apply (#2199).
-        let outcome = tokio::time::timeout(
-            WASM_APPLY_TIMEOUT,
-            self.context_client.execute(
-                &self.context_id,
-                &self.our_identity,
-                "__calimero_sync_next".to_owned(),
-                artifact,
-                vec![],
-                None,
-            ),
-        )
-        .await
-        .map_err(|_elapsed| {
-            ApplyError::Application(format!(
-                "WASM execution exceeded {}s budget (#2186)",
-                WASM_APPLY_TIMEOUT.as_secs()
-            ))
-        })?
+        // Do NOT bound this with a hard `timeout`: `context_client.execute`
+        // runs the WASM merge-apply on a `spawn_blocking` thread, which
+        // can't be cancelled. If we timed out and dropped this future, the
+        // caller's DAG write lock would be released and the delta recorded
+        // as not-applied, while the blocking apply ran to completion and
+        // then `commit()`d its storage writes + bumped `context.root_hash`
+        // anyway — late, racing the next delta's apply. Storage would hold
+        // a delta the DAG doesn't know about (re-synced, re-applied,
+        // divergent root hash). So we keep waiting; the apply is gas-bounded
+        // (it terminates) and post-#2238 is fast. Only warn if it runs long.
+        // See #2199 / #2238.
+        let execute = self.context_client.execute(
+            &self.context_id,
+            &self.our_identity,
+            "__calimero_sync_next".to_owned(),
+            artifact,
+            vec![],
+            None,
+        );
+        tokio::pin!(execute);
+        let outcome = match tokio::time::timeout(WASM_APPLY_TIMEOUT, &mut execute).await {
+            Ok(res) => res,
+            Err(_elapsed) => {
+                warn!(
+                    context_id = %self.context_id,
+                    delta_id = %Hash::from(delta.id),
+                    over_budget_secs = WASM_APPLY_TIMEOUT.as_secs(),
+                    is_merge = is_merge_scenario,
+                    "WASM merge-apply over soft budget — still waiting (a spawn_blocking apply can't be cancelled without leaving storage/DAG divergent); see #2199/#2238"
+                );
+                execute.await
+            }
+        }
         .map_err(|e| ApplyError::Application(format!("WASM execution failed: {e}")))?;
 
         let wasm_elapsed_ms = wasm_start.elapsed().as_secs_f64() * 1000.0;
--- a/crates/node/src/state_delta_bridge.rs
+++ b/crates/node/src/state_delta_bridge.rs
@@ -35,9 +35,14 @@ use crate::handlers::state_delta::{handle_state_delta, StateDeltaContext, StateD
 /// rebroadcast path.
 pub const STATE_DELTA_CHANNEL_CAPACITY: usize = 2048;
 
-/// Per-delta processing budget. Bounds the worst case where any
-/// downstream await wedges (e.g. slow WASM merge-apply, #2199); on
-/// timeout the delta is dropped and sync recovery retries.
+/// Soft per-delta processing budget. *Not* a hard cap: `handle_state_delta`
+/// drives `context_client.execute`, whose WASM merge-apply runs on a
+/// `spawn_blocking` thread that can't be cancelled. Abandoning the job on a
+/// hard timeout would release the DAG write lock while that apply completes
+/// and `commit()`s its writes anyway — late, racing the next delta and
+/// leaving storage holding a delta the DAG doesn't have. So exceeding this
+/// threshold only logs a warning and bumps `over_budget_total`; the job runs
+/// to completion. The durable fix for the underlying slowness is #2199/#2238.
 const STATE_DELTA_PROCESSING_TIMEOUT: Duration = Duration::from_secs(60);
 
 /// Periodic summary log interval.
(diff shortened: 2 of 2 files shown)


crates/node/src/delta_store.rs
```diff
@@ -222,26 +229,39 @@ impl DeltaApplier<Vec<Action>> for ContextStorageApplier {
...
+        let outcome = match tokio::time::timeout(WASM_APPLY_TIMEOUT, &mut execute).await {
+            Ok(res) => res,
+            Err(_elapsed) => {
+                warn!(
+                    context_id = %self.context_id,
+                    delta_id = %Hash::from(delta.id),
+                    over_budget_secs = WASM_APPLY_TIMEOUT.as_secs(),
+                    is_merge = is_merge_scenario,
+                    "WASM merge-apply over soft budget — still waiting (a spawn_blocking apply can't be cancelled without leaving storage/DAG divergent); see #2199/#2238"
+                );
```
frdomovic, line 261: 🟡 **tokio::pin! on execute future then awaiting &mut execute after timeout is semantically correct but subtle**

The pattern `tokio::pin!(execute); match tokio::time::timeout(WASM_APPLY_TIMEOUT, &mut execute).await { Ok(res) => res, Err(_) => execute.await }` is correct: after the timeout fires, `&mut execute` is dropped (the borrow ends), and then `execute.await` polls the same pinned future to completion. However, this relies on the fact that `tokio::time::timeout` with `&mut F` does NOT cancel the underlying future when the timeout fires — it only stops polling it. The future's state is preserved in the pinned location. This is correct behavior for tokio's timeout, but it is a subtle invariant. If tokio ever changes this behavior, or if a different timeout implementation is used, this could silently break. A comment explicitly noting this invariant (that the pinned future retains its state after the timeout borrow is dropped) would improve maintainability.

**Suggested fix:**
```
Add an explicit comment: `// tokio::time::timeout with &mut F does not cancel F on timeout — it only stops polling. // The pinned future retains its state; execute.await below resumes from where it left off.`
```

frdomovic, line 261: 🟡 **Soft timeout still holds DAG write lock for unbounded duration**

The PR correctly removes the hard timeout to prevent storage/DAG divergence, but the comment at line 261 says 'post-#2238 is fast' as justification for holding the DAG write lock indefinitely. If the WASM merge-apply is slow (e.g., due to a large payload, GC pressure, or a regression in #2238), the DAG write lock will be held for an unbounded duration, blocking all concurrent delta applications, sync operations, and any other code that acquires the DAG write lock. The warning fires after 15s but the lock is never released until completion. This is a correctness improvement over the previous bug, but introduces a potential liveness hazard under adversarial or pathological inputs.

**Suggested fix:**
```
Consider adding a separate monitoring/alerting mechanism (e.g., a background watchdog that logs if the lock has been held for >N seconds) without actually abandoning the apply. Also document the operational runbook for what to do if a node gets stuck here.
```

crates/node/src/state_delta_bridge.rs
```diff
@@ -190,47 +196,54 @@ impl Handler<StateDeltaJob> for StateDeltaActor {
...
+            // dropping this future would release the DAG write lock while
+            // that apply completes and commits late, racing the next delta
+            // (storage holds a delta the DAG doesn't — re-synced, re-applied,
+            // divergent). The merge-apply is gas-bounded so it terminates.
+            // See #2199 / #2238.
+            let result = match tokio::time::timeout(STATE_DELTA_PROCESSING_TIMEOUT, &mut fut).await
+            {
+                Ok(r) => r,
+                Err(_elapsed) => {
+                    let _prev = over_budget_total.fetch_add(1, Ordering::Relaxed);
```
frdomovic, line 213: 🟡 **Same subtle pin-and-resume pattern in state_delta_bridge without cancellation-safety comment**

The same `tokio::pin!(fut); timeout(&mut fut).await` pattern is used in state_delta_bridge.rs. The same correctness invariant applies: after the timeout, `fut.await` resumes the pinned future. This is correct but subtle. Additionally, `handle_state_delta` itself may not be cancellation-safe (it may hold internal state across awaits). While the PR's intent is to never cancel it, the pattern should be documented to prevent future maintainers from misunderstanding why the pin-and-resume idiom is used instead of simply `handle_state_delta(context, message).await`.

**Suggested fix:**
```
Add a comment similar to the one in delta_store.rs explaining the pin-and-resume pattern and why it's safe here.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 26% | Review time: 60.5s

🟡 3 warnings. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-68147aae`</sub>

crates/node/src/delta_store.rs
rtb-12: Correct, and intentional — `tokio::pin!(fut)` then `timeout(&mut fut)` then `fut.await` on the no-completion path is the standard "soft deadline, then keep waiting" idiom (the `&mut` borrow ends when `timeout` is dropped; `fut.await` resumes the same pinned future). The cancellation-safety of the inner future doesn't matter here precisely because we never drop it incomplete — that's the whole point of the change. Added comments at both sites already say this.

crates/node/src/state_delta_bridge.rs
rtb-12: Correct, and intentional — `tokio::pin!(fut)` then `timeout(&mut fut)` then `fut.await` on the no-completion path is the standard "soft deadline, then keep waiting" idiom (the `&mut` borrow ends when `timeout` is dropped; `fut.await` resumes the same pinned future). The cancellation-safety of the inner future doesn't matter here precisely because we never drop it incomplete — that's the whole point of the change. Added comments at both sites already say this.

(re: cancellation-safety of  — see above; it never gets dropped incomplete.)

crates/node/src/delta_store.rs
rtb-12: A secondary hard timeout would re-introduce the exact race this PR removes: the WASM apply runs on a `spawn_blocking` thread that **cannot be cancelled**, so any hard timeout — 15s or 5min — that fires while the apply is still running drops this future, releases the DAG write lock, records the delta as not-applied, and then the apply completes and `commit()`s late, racing the next delta → storage/DAG divergence. You can't tell "genuinely hung" from "finishes at 5min+1s" in advance, so the cap is net-negative. The apply is finite (gas-bounded WASM; the host-side merkle work is O(children-of-root), large but bounded), so it terminates. The real defense against a *truly* infinite WASM execution is gas-metering correctness in `crates/runtime`, not a timeout here. The DAG write lock is per-`DeltaStore` (per-context), so a slow apply stalls only that context. Leaving as-is.

rtb-12:
Thanks for the review passes. No code changes — replied on the inline threads; summary of the dispositions:

- **Pin-and-resume pattern (`tokio::pin!` → `timeout(&mut fut)` → `fut.await`)** — correct and intentional; it's the standard "soft deadline, then keep waiting" idiom. Inner-future cancellation-safety is moot here precisely because we never drop it incomplete (the point of the change). Comments at both sites say this.
- **"Soft timeout holds the DAG write lock indefinitely — add a 5-min hard cap as a circuit-breaker"** — declined: a hard cap of *any* length re-introduces the exact race this PR removes. The WASM apply runs on a `spawn_blocking` thread that can't be cancelled, so a hard timeout that fires while it's still running drops the future, releases the DAG write lock, records the delta as not-applied, and then the apply completes and `commit()`s late, racing the next delta → storage/DAG divergence. You can't distinguish "hung" from "finishes at cap+1s" up front. The apply is finite (gas-bounded WASM; host-side merkle work is O(children-of-root) — large but bounded), and the lock is per-context so a slow apply stalls only that context. A truly infinite WASM execution is a gas-metering bug in `crates/runtime`, to be fixed there, not papered over with a timeout that's strictly worse than the soft warning.
- **`over_budget_total` incremented before `fut.await`** — that's the right point: the counter means "jobs that exceeded the soft budget", and we know that's true the moment the timeout fires. Moving it after `fut.await` would *under*-count a job that goes over budget and then panics. There's no double-count path (incremented once, only in the `Err(_elapsed)` arm); `in_flight` is the only count that must balance and `InFlightGuard` handles it across panics.

The durable fix for the underlying merge-apply slowness remains #2199 / #2238.

PR #2343, opened 2026-05-12
feat(node, context): reconcile-on-mismatch via trusted-anchor snapsho…

chefsale:
…t sync

Wires the cross-DAG divergence signal from `MemberRemoved` / `MemberLeft` apply (the convergence-claims path that ships with this PR's predecessor) into an automatic recovery action: target a sync at a trusted-anchor peer, pull canonical state, verify the post-adoption root hash against the signed expected.

Flow end-to-end:

1. `MemberRemoved` / `MemberLeft` apply detects state-hash divergence between the signer's signed claims and the receiver's post-apply state (existing code path).

2. The verify helper now returns a structured `DivergenceReport` instead of just emitting a warn — carrying the signer's signed expected per-context root hashes so the reconcile path can verify adopted state against them, not against the anchor's claim.

3. The report propagates up through `apply_group_op_mutations` → `apply_signed_op` → `NamespaceApplyOutcome::Applied { divergence }` to the node-side `handle_namespace_governance_delta` handler. The message-type DivergenceReport lives in calimero-context-client so primitives can carry it without depending on calimero-context; group_store re-exports it for internal use.

4. The applier (which sits inside the DAG-application trait that has no room for structured output) stashes the report in a small outbox `Mutex<Option<...>>`; the handler reads-and-clears after `add_delta_with_outcome` returns. Single-flight per applier instance so the slot is safe.

5. On `Applied { divergence: Some(report) }`, the handler calls `SyncManager::reconcile_after_divergence`. For each `(context_id, expected_root_hash)` in `hash_differs`: - Look up trusted anchors for the context's group (the existing `trusted_anchors_for_group` helper). - Filter gossipsub mesh peers on the context's topic to those whose verified `peer_identities` cache entry intersects the anchor set. - Pick the first matching peer, invoke `initiate_sync` — reuses the existing Snapshot sync protocol. - After sync completes, compare the local context's `root_hash` against the signed expected from the triggering op. Match → info log "converged". Mismatch → loud warn (operator investigation territory). No anchor connected → fall back to "operator path" with a structured warn; reconcile re-attempts on the next signed op or sync tick.

6. `only_in_expected` and `only_in_actual` entries are NOT reconciled here — those buckets reflect namespace-DAG drift (a registration the receiver hasn't seen, or one the signer hadn't seen). The cross-DAG check on subsequent state deltas catches that via `Unknown { needed }` → buffer; routing them through anchor sync would burn bandwidth on cases the existing catch-up path handles correctly.

Limitations called out in the code:

* Post-adoption verification (not pre-adoption rejection). The current store layer doesn't have transactional staging — there's no clean way to "stage, verify, commit-or-rollback" a snapshot sync. So we sync, then verify, and on mismatch the receiver is left with potentially-bad state until the next signed op re-triggers reconcile. Pre-adoption rejection with rollback is a follow-up that depends on store-layer staging primitives.

* Best-effort: no anchor connected → reconcile drops with a warn; retried on next signed op or sync tick. A failure mode that matters in practice only when the trusted-anchor set is permanently unreachable, which is a deployment configuration issue.

Wire change: `NamespaceApplyOutcome::Applied` becomes a struct variant carrying `divergence: Option<DivergenceReport>`. Lost `Copy` derive (not used by any consumer) and `PartialEq` (no consumer either). The previous `is_pending()` method stayed; added `is_applied()` for matchers that previously used the unit-variant shape. Six call sites in the node crate updated to `{ .. }` patterns.

Tests:
* 2 new unit tests on `apply_group_op_mutations` pinning the divergence return: matched hashes → `None`, mismatched → `Some(report)` carrying the right `op_kind`, `group_id`, and `group_hash_diverges` flag.
* The existing `diff_sorted_context_hashes` tests updated to match the new `hash_differs: Vec<(ContextId, [u8; 32])>` shape (was `Vec<ContextId>`) — the expected hash now travels alongside each divergent context id so the reconcile path can verify received state.
* All 245 context-lib + 23 convergence + 133 node-lib tests still pass.

E2E coverage of the full reconcile flow (multi-node Snapshot sync gated on anchor identity + post-adoption verification) is left for a separate test PR — requires a 3+ node fixture where the receiver diverges by applying an extra state-DAG delta that the signer's view didn't include.

---

> [!NOTE]
> **High Risk**
> High risk because it changes core governance-apply outcomes and introduces automatic state reconciliation/sync behavior (including backoff) that can affect data convergence and network traffic across nodes.
> 
> **Overview**
> Adds a structured divergence signal for `MemberRemoved`/`MemberLeft` by introducing `DivergenceReport` and changing `NamespaceApplyOutcome::Applied` to carry `divergence`, then plumbs this through namespace DAG apply via an applier outbox so callers can act on post-apply hash mismatches.
> 
> On the node side, wires `Applied { divergence: Some(..) }` from both gossipsub delivery and namespace backfill/retry paths into `SyncManager::reconcile_after_divergence`, which attempts per-context trusted-anchor sync, verifies the post-sync root hash against the **signed expected** value, and applies per-context exponential backoff to avoid tight reconcile loops.
> 
> Adds new scaffolding-e2e workflows covering multi-context convergence and late-joiner catchup behavior, plus Docker bridge connect/disconnect scripts and CI matrix registration (partition-window reconcile workflow is added but intentionally not yet run).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 474d036770006b467fcd2f1ff8caa2208be81276. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/node/src/sync/manager/mod.rs
```diff
@@ -2936,6 +2936,188 @@ impl SyncManager {
...
+            "reconcile-after-divergence: pulling canonical state from trusted anchor"
+        );
+
+        match self.initiate_sync(context_id, anchor_peer).await {
+            Ok((peer_used, protocol)) => {
+                tracing::info!(
+                    %context_id,
+                    %peer_used,
+                    ?protocol,
+                    "reconcile-after-divergence: anchor sync completed; verifying post-adoption hash"
```
frdomovic, line 3176: 🟡 **Post-adoption hash verification is non-binding: bad state is permanently adopted on mismatch**

After `initiate_sync` completes, `verify_post_reconcile_root_hash` only logs a warning on hash mismatch — it does not roll back the adopted state or re-trigger reconcile. The PR description acknowledges this as a known limitation, but the consequence is that if the anchor serves non-canonical state (whether due to a bug, a compromised anchor, or a race), the receiver permanently adopts that state until the next signed op re-triggers reconcile. There is no retry counter, no backoff, and no circuit-breaker. In a scenario where the anchor is consistently serving wrong state, this creates an infinite reconcile loop (each signed op triggers a sync, sync adopts wrong state, verify fails, next op triggers again) that burns bandwidth and CPU with no convergence.

**Suggested fix:**
```
Add a per-context reconcile attempt counter with a backoff or maximum retry limit. After N consecutive post-adoption verification failures against the same anchor, mark the anchor as suspect and fall back to the operator path rather than retrying indefinitely.
```

crates/node/src/sync/manager/mod.rs
```diff
@@ -2936,6 +2936,188 @@ impl SyncManager {
...
+                group_hash_diverges = report.group_hash_diverges,
+                only_in_expected_count = report.only_in_expected.len(),
+                only_in_actual_count = report.only_in_actual.len(),
+                "reconcile-after-divergence: no per-context hash mismatches to reconcile; \
+                 namespace-DAG drift (if any) is handled by the cross-DAG check on \
+                 subsequent state deltas"
+            );
+            return;
+        }
+
```
frdomovic, line 3017: 🟡 **Trusted anchor identity lookup uses DashMap::get without holding lock across peer selection**

The anchor peer selection loop calls `self.node_state.peer_identities.get(peer)` inside a closure passed to `Iterator::find`. Each `get` returns a short-lived reference that is dropped before the next iteration, but the overall selection is not atomic: between the `mesh_peers` snapshot and the `initiate_sync` call, the selected `anchor_peer` could disconnect or its identity could be evicted from the cache. More critically, the `peer_identities` cache is populated by a separate gossip path and there is no guarantee it is up-to-date. A malicious or misconfigured peer that is in the mesh but whose identity has not yet been verified could slip through if the cache entry is absent (the `unwrap_or(false)` default). This is a best-effort design acknowledged in the PR, but the security implication — that an unverified peer could be selected as the 'trusted anchor' if the cache is stale — should be explicitly documented and the fallback behavior (no anchor found) should be preferred over silently skipping unverified peers.

**Suggested fix:**
```
Change `unwrap_or(false)` to make the absence of a cache entry an explicit non-match and add a debug log when a mesh peer is skipped due to missing identity cache entry, so operators can detect cache staleness.
```

crates/context/primitives/src/messages.rs
```diff
@@ -154,10 +154,36 @@ impl Message for ApplySignedGroupOpRequest {
...
 /// into `Ok(false)`, causing every duplicate gossip op to open a redundant
 /// backfill stream.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Post-apply state-hash divergence detected by the cross-DAG
+/// convergence check (the signed-claims path on `MemberRemoved` /
+/// `MemberLeft`). The node-side handler routes this to the
+/// reconcile-via-anchor sync trigger so the receiver can pull
+/// canonical state from a trusted-anchor peer.
+///
+/// `hash_differs` carries the **signed expected** hash alongside
```
frdomovic, line 158: 📝 **DivergenceReport doc comment is misplaced relative to the enum it precedes**

The doc comment block starting at line 154 begins with the description of `NamespaceApplyOutcome` (the old comment about distinguishing 'pending' from 'duplicate'), then transitions mid-block into the description of `DivergenceReport`. The result is that `DivergenceReport`'s doc comment is attached to the struct but the preceding text about `NamespaceApplyOutcome` is orphaned. This makes rustdoc output confusing — the `NamespaceApplyOutcome` enum loses its original doc comment and `DivergenceReport` inherits unrelated text.

**Suggested fix:**
```
Split the doc comment block: keep the original `NamespaceApplyOutcome` comment above the enum, and write a separate clean doc comment for `DivergenceReport` above the struct definition.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 25% | Review time: 85.5s

🟡 2 warnings, 📝 1 nitpicks. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-18622156`</sub>

PR #2350, opened 2026-05-13
docs(context): lock in the forward-only membership invariant

chefsale:
## Summary

Lock-in pass for the forward-only invariant on the cross-DAG apply-time membership check. **No behavior change.** Three places now make the contract explicit so a future PR can't silently break it:

1. Inline comment at each of the 3 production call sites in `crates/node/src/handlers/state_delta/mod.rs`.
2. \"Forward-only contract — load-bearing\" section in the rustdoc on `membership_status_at`.
3. New \"Forward-only invariant — load-bearing\" section in `architecture/membership-and-leave.html`.

## Why this matters

The apply-time membership check answers \"was the signer a member at the named cut?\", not \"is the signer currently a member?\" The distinction matters when a delta is in flight across a `MemberRemoved`: two peers that receive the ops in different orders must converge to the same state. Without forward-only, a peer that observes `MemberRemoved` first rejects the pre-removal delta; a peer that observes the delta first applies it. State diverges, root hashes drift, no honest convergence point.

The property is already implemented correctly today — `membership_status_at` walks only the ancestry of the signed heads, every production caller passes `delta.governance_position`, and the resolver-level regression tests + canary already pin the BFS-boundary behavior. This PR makes the contract loud at the points where a future change might silently break it.

### `crates/node/src/handlers/state_delta/mod.rs`

Three call sites now carry an explicit forward-only contract comment immediately above `membership_status_at`:

- gossip-receive path (`apply_authorized_state_delta`)
- governance-pending drain
- snapshot-sync replay

A reviewer who sees a future PR swapping the position argument for current state will see the deleted comment.

### `crates/context/src/group_store/membership_status.rs`

New \"Forward-only contract — load-bearing\" section in the public-facing rustdoc on `membership_status_at`. States the contract in caller-facing terms (\"must come from a signed envelope; must NOT be current state\"), names the property, points at the regression tests.

### `architecture/membership-and-leave.html`

New top-level card \"Forward-only invariant — load-bearing\" placed after the Membership Model section. Covers:

- The property and why it matters (with the divergence example).
- How it's implemented (BFS over signed-head ancestry only).
- Two specific classes of change that would silently break it.
- Scope clarification: applies to `MemberRemoved` / `MemberLeft`; role demotions take effect against current state via `is_read_only_for_context`.

## Test plan

- [x] `cargo test -p calimero-context --lib` — 248 passed
- [x] `cargo test -p calimero-context --test local_group_governance_convergence` — 23 passed
- [ ] CI green on push

## Not in this PR

- **Partition-heal e2e**: the docker-bridge-based partition-window e2e was disabled in #2343 because the partition primitive can't construct the divergence scenario deterministically. Re-enabling needs iptables-level primitives — separate follow-up.
- **End-to-end forward-only tests against a real namespace DAG**: the existing tests run at the resolver level (`resolve_membership_from_transitions`) which mirrors what the BFS feeds. Adding a duplicate end-to-end test against a real DAG would be redundant with the resolver canary; could be a follow-up if the layering changes.

---

> [!NOTE]
> **Medium Risk**
> Touches cross-DAG authorization on the state-delta receive/replay/drain paths by adding stricter `governance_position`/`group_id` validation; mistakes here could cause legitimate deltas to be dropped or authorization to be bypassed. Changes are localized and backed by new unit tests, but this is security-sensitive logic.
> 
> **Overview**
> **Locks in the “forward-only” contract for cross-DAG membership checks** by adding explicit documentation in `architecture/membership-and-leave.html` and rustdoc on `membership_status_at`, plus inline call-site comments warning against using receiver-local state.
> 
> **Hardens state-delta apply paths against `governance_position` bypasses** by introducing a shared `GroupIdCheck`/`verify_position_group_id_matches_context` helper and applying it consistently in gossip receive (`handle_state_delta`), governance-pending drain, and snapshot-sync replay, rejecting deltas with missing/incorrect `group_id` claims. Adds unit tests covering all `GroupIdCheck` outcomes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd5fb9cb0757153a7816329683bf8b07c2a6256d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 85% | Review time: 57.1s

---

### ✅ No Issues Found

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-a8aa1473`</sub>

PR #2351, opened 2026-05-13
fix(context): trigger KeyDelivery for self-join of Open subgroups

frdomovic:
Self-join via the inherited (Open-subgroup) path leaves the user with a ContextIdentity row whose `sender_key` is `None`. No admin runs `add_group_members` for them, so no `KeyDelivery` is ever published wrapping the current group key for the joiner. They subscribe to gossipsub and sync, but the state-DAG is encrypted with the group key they don't hold — so messages they send appear only locally, and they can't decrypt anyone else's. Symptom: "joined a public channel, sends messages, no one sees them; can't see others either."

This adds `RootOp::MemberJoinedOpen { member, group_id }`, mirroring `MemberJoined`'s key-delivery trigger semantics but without an admin-signed invitation (Open subgroup + `CAN_JOIN_OPEN_SUBGROUPS` at the namespace root IS the authorisation). The joiner signs and publishes the op at the end of `join_context.rs` when their membership path was `Inherited`. On apply, every peer that holds the current group key for `group_id` records a `PendingKeyDelivery` (same struct the existing `MemberJoined` path uses); the delivery dispatcher then publishes a `KeyDelivery` wrapped for the joiner, who unwraps and populates their `sender_key`.

`execute_member_joined_open` rejects ops where the outer signer disagrees with `member` (key-ownership proof), or where the joiner's membership path is `Direct` (admin should use `add_group_members`) or `None`. The check uses the same `check_group_membership_path` the join handler already runs locally.

Backward compatibility: adds a new Borsh enum variant to `RootOp` — old peers will fail to decode this op. Fine for coordinated dev cluster upgrades; release notes need to flag a network-wide upgrade requirement.

---

> [!NOTE]
> **High Risk**
> Adds a new governance wire-op (`RootOp::MemberJoinedOpen`) and changes membership/authorization evaluation, which is consensus-adjacent and requires coordinated upgrades because older peers cannot decode the new variant.
> 
> **Overview**
> Fixes Open-subgroup inherited self-joins by having the joiner publish a new cleartext governance op `RootOp::MemberJoinedOpen` from `join_context`, allowing peers that already hold the subgroup key to enqueue a `PendingKeyDelivery` and publish `KeyDelivery` so the joiner’s `sender_key` can be populated.
> 
> Extends governance apply and cross-DAG membership checks to recognize `MemberJoinedOpen` (including validation that the signer matches the joiner and that the membership path is *Inherited*, plus a namespace-resolution guard), updates publish timeout classification/metrics labels accordingly, and adds a new scaffolding E2E workflow (wired into CI) that asserts bidirectional reads/writes after an inherited Open-subgroup join.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1eeb4c55c87b8b8248812f1e52fef5f5ebe43c98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/join_context.rs
+++ b/crates/context/src/handlers/join_context.rs
@@ -33,6 +33,7 @@ impl Handler<JoinContextRequest> for ContextManager {
         let datastore = self.datastore.clone();
         let context_client = self.context_client.clone();
         let node_client = self.node_client.clone();
+        let ack_router = std::sync::Arc::clone(&self.ack_router);
         ActorResponse::r#async(
             async move {
                 let mut group_id = group_store::get_group_for_context(&datastore, &context_id)?;
@@ -142,6 +143,7 @@ impl Handler<JoinContextRequest> for ContextManager {
                     &group_id,
                     &joiner_identity,
                 )?;
+                let mut was_inherited = false;
                 match membership_path {
                     group_store::MembershipPath::None => {
                         bail!("identity is not a member of the group");
@@ -162,6 +164,7 @@ impl Handler<JoinContextRequest> for ContextManager {
                             via_admin,
                             "context join authorized via inherited subgroup membership"
                         );
+                        was_inherited = true;
                     }
                 }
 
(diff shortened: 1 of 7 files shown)


frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 25% | Review time: 88.0s

📝 1 nitpicks. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-6746821c`</sub>

crates/context/src/handlers/join_context.rs
```diff
@@ -232,6 +235,51 @@ impl Handler<JoinContextRequest> for ContextManager {
                     "joined context via group membership"
                 );
 
+                // Inherited members (Open subgroup, joined via the
+                // CAN_JOIN_OPEN_SUBGROUPS parent-walk) don't get a
+                // `KeyDelivery` from any admin — admin never called
+                // `add_group_members` for them. Without the group key
+                // their `sender_key` stays None on the ContextIdentity
+                // row above, which means they can't decrypt state-DAG
```
frdomovic, line 243: 📝 **Fully qualified path for PrivateKey and NamespaceOp instead of using existing imports**

The code uses fully qualified paths `calimero_primitives::identity::PrivateKey::from(sk_bytes)` and `calimero_context_client::local_governance::NamespaceOp::Root(...)` instead of importing these types at the top of the file. Other parts of the codebase use shorter paths. This is inconsistent with the import style used elsewhere in the handler.

**Suggested fix:**
```
Add `use calimero_primitives::identity::PrivateKey;` and `use calimero_context_client::local_governance::{NamespaceOp, RootOp};` to the imports at the top of join_context.rs and use the shorter forms in the code.
```

rtb-12 (approved):

PR #2354, opened 2026-05-14
test(e2e): regression guard for Open-subgroup self-join key delivery (#2351)

rtb-12:
## Summary

Targeted e2e regression guard for **#2351** (`fix(context): trigger KeyDelivery for self-join of Open subgroups`). Adds a new merobox workflow and a matrix entry that exercises the buggy outbound-author direction the existing `group-subgroup-visibility-inheritance` test does not.

## Why we need another inheritance test

The existing `group-subgroup-visibility-inheritance.yml` already tests **inbound-decrypt** for an inherited self-joiner (node-1 writes → node-2 reads). That direction passes on master because the joiner can still decrypt incoming ops using the group key they receive through the namespace identity path.

The bug #2351 fixes is in the **outbound-author** direction: an inherited joiner gets a `ContextIdentity` row with `sender_key: None`, no admin runs `add_group_members` for them, no `KeyDelivery` ever wraps the current group key for them. Their writes can't be signed/encrypted, so peers ignore or fail to decrypt them. User-visible symptom: *"joined a public channel, sends messages, no one sees them; can't see others either."*

This workflow drives the broken direction explicitly:

1. Bootstrap namespace + Open subgroup + context (single admin)
2. Node-2 self-joins via `CAN_JOIN_OPEN_SUBGROUPS` parent-walk (no `add_group_members`)
3. **Round 1 (control)**: node-1 writes → node-2 reads — confirms fix doesn't regress the working direction
4. **Round 2 (the bug)**: node-2 writes → node-1 reads — pre-fix returns null/empty or `wait_for_sync` times out

## Expected CI behaviour

- **On master (no fix)**: `Test (group-open-self-join-key-delivery)` **fails**. That failure is the proof the test catches the bug — intentional, not a regression of this PR.
- **Once #2351 lands** (or is rebased onto this branch): test goes green. Lock-in.

This PR is not mergeable on master until #2351 (or an equivalent fix) merges. Suggested merge sequence:
1. Land #2351 on master
2. Rebase this PR onto new master
3. CI goes green → merge

Alternatively, rebase #2351 onto this branch first to verify locally that the fix unbreaks the test.

## Test plan
- [ ] CI runs the new matrix entry `Test (group-open-self-join-key-delivery)` on this branch (master code)
- [ ] CI on this branch reports the new test as failing (confirms test catches the bug)
- [ ] Once #2351 merges, rebase + re-run; new test passes
[...]

PR #2355, opened 2026-05-14
test(e2e): regression guard for PR #2351 — Open-subgroup self-join key delivery

rtb-12:
## Summary

Targeted e2e regression guard for **#2351** (`fix(context): trigger KeyDelivery for self-join of Open subgroups`), **stacked on top of #2351 so CI verifies the fix and the test together**.

## Why this approach (vs. branching off master)

The existing `group-subgroup-visibility-inheritance.yml` only exercises **inbound-decrypt** for an inherited joiner (node-1 writes → node-2 reads) — that direction passes on master because the joiner can still decrypt using the namespace-level group key. The bug #2351 fixes is in the **outbound-author** direction: `sender_key: None` means the joiner can't sign/encrypt their own writes, so peers ignore them.

By basing this PR against `#2351`'s branch instead of `master`:

| If CI on this branch | Conclusion |
|---|---|
| 🟢 passes | #2351's fix actually works — round-trip in both directions succeeds. Ready to merge. |
| 🔴 fails at Round 2 | #2351's fix is incomplete — there's still a gap in the outbound path. Needs more work before merge. |

This is the actual verification we couldn't get from a master-based test, which would just confirm "bug exists" without proving "fix fixes it."

## The new workflow

`apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml`:
1. Bootstraps namespace + Open subgroup + context (single admin on node-1)
2. Node-2 joins namespace (gets default `CAN_JOIN_OPEN_SUBGROUPS`)
3. Node-2 self-joins context via the parent-walk — **no `add_group_members` runs for them**
4. **Round 1 (control)**: node-1 writes → node-2 reads — works on master too; confirms fix doesn't regress
5. **Round 2 (the bug)**: node-2 writes → node-1 reads — the broken direction

Pre-fix expectation: Round 2 read returns null/empty or `wait_for_sync` times out at 30s.
Post-fix expectation: round-trip succeeds.

## Merge sequence

1. Verify CI on this PR passes (proves #2351 works)
2. Merge #2351 to master
3. Rebase this PR onto master (becomes a no-op rebase since the test is fix-independent)
4. Merge this PR to master → regression guard locked in

If CI fails here, hold #2351 until the gap is closed.

## Supersedes

#2354 (test branched off master, only proved bug exists, didn't verify fix). Closing that.

## Test plan
- [ ] `Test (group-open-self-join-key-delivery)` matrix entry runs against (Fran's fix + new test) and reports PASS
- [ ] No regression in `group-subgroup-visibility-inheritance` or other `group-subgroup-*` matrix entries
- [ ] If FAIL: log dump from the two node containers shows which step broke

---

> [!NOTE]
> **Low Risk**
> Adds a new E2E workflow and registers it in CI; primary risk is increased CI runtime/flakiness rather than production behavior changes.
> 
> **Overview**
> Adds a new `scaffolding-e2e` workflow, `group-open-self-join-key-delivery.yml`, that reproduces the Open-subgroup *inherited self-join* scenario and asserts **bi-directional** KV writes (admin → joiner and joiner → admin) converge, catching regressions where the joiner never receives a `sender_key`/`KeyDelivery`.
> 
> Registers this workflow in the `e2e-rust-apps.yml` CI matrix so it runs alongside existing group/subgroup E2E coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70af37c208d759884ae1b50fef724e610b3c9944. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
+++ b/apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
@@ -1,0 +1,241 @@
+name: E2E KV Store - Open Subgroup Self-Join Key Delivery
+description: >
+  Regression guard for PR #2351 (`fix(context): trigger KeyDelivery for
+  self-join of Open subgroups`).
+
+  Pre-fix bug: when a namespace member self-joins an `Open` subgroup
+  context via the `CAN_JOIN_OPEN_SUBGROUPS` parent-walk (no
+  admin-issued `add_group_members`), `join_context` inserts a
+  `ContextIdentity` row with `sender_key: None`. No `MemberJoined`
+  governance op is published for them, so no peer ever wraps the
+  current group key in a `KeyDelivery` for the joiner. Their
+  outgoing state-DAG ops can't be signed/encrypted with the group
+  key, so peers reject/ignore them; symmetrically, the joiner can't
+  decrypt incoming ops. Symptom: "joined a public channel, sends
+  messages, no one sees them; can't see others either."
+
+  Fix: `join_context.rs` now publishes a self-signed
+  `RootOp::MemberJoinedOpen { member, group_id }` whenever the join
+  path was `MembershipPath::Inherited`. Peers holding the group key
+  queue a `PendingKeyDelivery` (same struct the existing
+  `MemberJoined` path uses); the delivery dispatcher publishes a
+  wrapped `KeyDelivery`, populating the joiner's `sender_key`.
+
+  This workflow differs from `group-subgroup-visibility-inheritance.yml`
+  by exercising the *joiner-as-author* direction explicitly: that
+  existing test only confirms node-1's writes flow to node-2, which
+  passes on master because the joiner can DECRYPT incoming ops using
+  the namespace-level group key delivered through a different path.
+  The bug is that the joiner can't AUTHOR ops because `sender_key`
+  was never populated — that's what this test asserts by having
+  node-2 write and node-1 read the result.
+
+  Pre-fix expectation: Round 2 wait_for_sync times out, or the
+  node-1 read of node-2's write returns null/empty (node-1 can't
+  decrypt the joiner's ops). Post-fix expectation: both directions
+  succeed.
+
+force_pull_image: false
+near_devnet: false
+
+nodes:
+  count: 2
+  image: merod:local
+  prefix: open-selfjoin-node
+
+steps:
+  # ── Phase 1: Bootstrap mesh, namespace, Open subgroup, context ────
+
+  - name: Install application on node-1
+    type: install_application
+    node: open-selfjoin-node-1
+    path: res/scaffolding_e2e.wasm
+    dev: true
+    outputs:
+      app_id: applicationId
+
+  - name: Install application on node-2
+    type: install_application
+    node: open-selfjoin-node-2
+    path: res/scaffolding_e2e.wasm
+    dev: true
+
+  - name: Node-1 creates namespace
+    type: create_namespace
+    node: open-selfjoin-node-1
+    application_id: '{{app_id}}'
+    outputs:
+      namespace_id: namespaceId
+      node1_key: memberPublicKey
+
+  - name: Invite node-2 to namespace
+    type: create_namespace_invitation
+    node: open-selfjoin-node-1
+    namespace_id: '{{namespace_id}}'
+    outputs:
+      ns_invite: invitation
+
+  - name: Wait for namespace governance to gossip
+    # 10s matches the post-#2344 opaque-leaf-regression timing; covers
+    # invitation visibility so `join_namespace` doesn't 404.
+    type: wait
+    seconds: 10
+
+  - name: Node-2 joins namespace
+    # Default member capabilities include CAN_JOIN_OPEN_SUBGROUPS, so
+    # node-2 inherits the right to join any Open subgroup beneath
+    # this namespace without an explicit add_group_members.
+    type: join_namespace
+    node: open-selfjoin-node-2
+    namespace_id: '{{namespace_id}}'
+    invitation: '{{ns_invite}}'
+    outputs:
+      node2_identity: memberIdentity
+
+  - name: Node-1 creates Open subgroup inside namespace
+    type: create_group_in_namespace
+    node: open-selfjoin-node-1
+    namespace_id: '{{namespace_id}}'
+    group_alias: open-channel
+    outputs:
+      subgroup_id: groupId
+
+  - name: Mark subgroup as Open
+    type: set_subgroup_visibility
+    node: open-selfjoin-node-1
+    group_id: '{{subgroup_id}}'
+    visibility: open
+
+  - name: Node-1 creates context inside Open subgroup
+    type: create_context
+    node: open-selfjoin-node-1
+    application_id: '{{app_id}}'
+    group_id: '{{subgroup_id}}'
+    outputs:
+      ctx_id: contextId
+
+  # ── Phase 2: node-2 self-joins via inherited path ─────────────────
+  # This is the critical operation. No `add_group_members` is run for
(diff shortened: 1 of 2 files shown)


apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
```diff
@@ -0,0 +1,241 @@
...
+force_pull_image: false
+near_devnet: false
+
+nodes:
+  count: 2
+  image: merod:local
+  prefix: open-selfjoin-node
+
+steps:
+  # ── Phase 1: Bootstrap mesh, namespace, Open subgroup, context ────
```
frdomovic, line 47: 🟡 **Subgroup created after node-2 joins namespace — node-2 may not see it**

Node-2 joins the namespace (step at line ~80) before the Open subgroup is created (line ~90) and before its visibility is set to 'open'. The subgroup's GroupCreated and visibility governance ops must propagate to node-2 before node-2 can self-join the context. The workflow relies on the 10-second wait (line 47) that was inserted only to cover the invitation visibility for `join_namespace`, not for subgroup propagation. Unlike `dm-subgroup-privacy.yml` (which adds an explicit `wait_for_sync` on the subgroup's governance state before the join), this workflow has no such guard between 'Mark subgroup as Open' and 'Node-2 self-joins Open-subgroup context'. If node-2 hasn't yet received the subgroup's governance ops, `join_context` may fail or succeed without the inherited path being taken, silently turning the test into a no-op.

**Suggested fix:**
```
Add a `wait_for_sync` step with `group_id: '{{subgroup_id}}'` targeting both nodes after 'Mark subgroup as Open' and before 'Node-2 self-joins Open-subgroup context', mirroring the pattern in `dm-subgroup-privacy.yml` lines ~175-180.
```

apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
```diff
@@ -0,0 +1,241 @@
...
+    visibility: open
+
+  - name: Node-1 creates context inside Open subgroup
+    type: create_context
+    node: open-selfjoin-node-1
+    application_id: '{{app_id}}'
+    group_id: '{{subgroup_id}}'
+    outputs:
+      ctx_id: contextId
+
```
frdomovic, line 116: 🟡 **Initial sync timeout of 60s may be insufficient for the KeyDelivery round-trip**

The 'Initial sync (empty Root state)' step uses `timeout: 60` and `check_interval: 2`. The comment says this is to let the post-join_context flow settle, including the MemberJoinedOpen publish + KeyDelivery round-trip. However, `wait_for_sync` typically checks DAG root-hash convergence, not whether a `sender_key` has been populated in the ContextIdentity row. If the KeyDelivery round-trip completes after DAG convergence but before the first `call` step, the test passes correctly. But if `wait_for_sync` returns as soon as the (empty) DAG converges — before the KeyDelivery has been processed — node-2's `sender_key` may still be None when Round 2 begins, causing a false negative (test fails even with the fix). There is no explicit assertion or wait that confirms the KeyDelivery was received and applied.

**Suggested fix:**
```
After the initial `wait_for_sync`, add a short `wait` step (e.g., 5 seconds) to give the KeyDelivery dispatch time to complete, similar to the 'Brief settle' pattern used in `dm-subgroup-privacy.yml`. Alternatively, document that `wait_for_sync` is sufficient because the KeyDelivery is part of the same gossip round.
```

apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
```diff
@@ -0,0 +1,241 @@
...
+    type: wait_for_sync
+    context_id: '{{ctx_id}}'
+    nodes:
+      - open-selfjoin-node-1
+      - open-selfjoin-node-2
+    timeout: 60
+    check_interval: 2
+
+  # ── Phase 3: Round 1 — node-1 writes, node-2 reads ────────────────
+  # Control case. This direction works on master too (matches the
```
frdomovic, line 155: 💡 **Round 1 read result not asserted with json_assert — only is_set checked**

The final assertion block (lines ~225-235) uses `is_set({{node2_read_admin}})` and `contains({{node2_read_admin}}, "node1_msg")`. The `contains` check is a substring match, not a structural equality check. Other workflows in this repo use `json_assert` with `json_equal` for precise value verification (e.g., `json_equal({{result}}, {"output": "value"})`). A substring match on a serialized JSON value could pass even if the response structure is unexpected (e.g., an error message that happens to contain the substring).

**Suggested fix:**
```
Replace the `contains` assertions with `json_assert` steps using `json_equal({{node2_read_admin}}, {"output": "node1_msg"})` and `json_equal({{node1_read_joiner}}, {"output": "node2_msg"})` to match the pattern used throughout the rest of the test suite.
```

apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
```diff
@@ -0,0 +1,241 @@
...
+  # can't be signed with the group key. Either:
+  #   (a) the write attempt fails locally with a key-missing error, or
+  #   (b) the op is produced but peers reject it as malformed/unsigned, or
+  #   (c) the op propagates as plaintext but node-1's apply path
+  #       refuses to apply ops not encrypted with the current group key.
+  # In all three sub-modes, node-1's read of "from_joiner" returns
+  # null/empty after Round 2 sync. Post-fix, the key has been
+  # delivered to node-2 so the write authors cleanly and node-1
+  # decrypts it.
+
```
frdomovic, line 200: 💡 **Round 2 write failure is silent — no expected_failure guard if write itself errors**

If the pre-fix bug manifests as case (a) — the write attempt fails locally with a key-missing error — the `call` step at 'Round 2 — node-2 (inherited self-joiner) writes' will fail the entire workflow immediately, before reaching the assertion block. This is actually the desired behavior (test fails = bug detected), but the failure message will be a generic call error rather than a meaningful assertion failure. There is no `expected_failure: false` annotation or comment clarifying that a failure here is intentional pre-fix. This could confuse future maintainers who see a CI failure at the write step rather than at the assertion step.

**Suggested fix:**
```
Add a comment above the Round 2 write step explicitly noting that a failure at this step (rather than at the assertion) is also a valid signal of the bug, and that the test is designed to catch both failure modes.
```

apps/scaffolding-e2e/workflows/group-open-self-join-key-delivery.yml
```diff
@@ -0,0 +1,241 @@
+name: E2E KV Store - Open Subgroup Self-Join Key Delivery
```
frdomovic, line 1: 📝 **Workflow description references PR numbers that will be stale after merge**

The description block (lines 1-40) references '#2351' and '#2354' multiple times. After both PRs are merged, these references become stale and may confuse future readers who encounter the file without the PR context. The description is otherwise excellent and detailed.

**Suggested fix:**
```
After merge, consider replacing PR number references with a brief summary of the fix (e.g., 'fixed in commit <sha>') or removing the PR-specific language in favor of a timeless description of the invariant being tested.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 27% | Review time: 66.6s

🟡 2 warnings, 💡 2 suggestions, 📝 1 nitpicks. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-dcecb555`</sub>

rtb-12:
Test commit pushed directly to fix/key-delivery-for-self-join-open-subgroups (5dec9cfe). CI on #2351 now runs the fix + regression test together.

Issue #2256, opened 2026-04-24
Add `subgroup_visibility` with parent-chain membership inheritance for Open subgroups


rtb-12:
## Summary

Core currently has no concept of an "open subgroup" — subgroup membership is always strictly explicit, enforced by `check_group_membership` (`crates/context/src/group_store/membership.rs:118-124`) which only checks direct membership with no parent-chain walk.

This creates a significant gap for apps that want to expose folder/channel/project structures where "everyone in the workspace can access this folder by default" — a common UX pattern.

Proposing: add a `subgroup_visibility: VisibilityMode` field on group metadata (`Open` / `Restricted`) and extend `check_group_membership` to walk the parent chain when the subgroup is `Open`, mirroring how `is_group_admin` already walks parents for admin inheritance (`membership.rs:140-154`).

## The situation today

Core has two visibility-related primitives, but neither covers this case:

1. **`VisibilityMode { Open, Restricted }`** (`crates/context/config/src/lib.rs:125-128`) — applies to **contexts within a group**. Documented in its own doc comment: "Visibility mode for a context within a group."

2. **`CAN_JOIN_OPEN_CONTEXTS`** capability (`crates/context/config/src/lib.rs:137`) — lets a group member auto-join Open contexts within groups they're **already a member of**. Doesn't extend access to groups you're not in.

The third case — "member of a parent group should automatically have access to a child subgroup" — has no primitive. `check_group_membership` returns false for non-direct members regardless of any visibility setting, and `join_context.rs:138` rejects with "identity is not a member of the group" accordingly.

Even the comment in `join_context.rs:132-134` explicitly acknowledges this:

> \`\`\`
> // Group membership already verified above. All contexts in a group
> // a member has access to are joinable. Restricted access is handled
> // at the subgroup level (admin must explicitly add member to the subgroup).
> \`\`\`

## How mero-drive works around it today

mero-drive (apps/mero-drive repo) stores a per-folder \`visibility: 'Inherit' | 'Restricted'\` field in its registry WASM. "Inherit" folders are supposed to be accessible to every namespace member. Since core can't enforce this, the app does a manual cascade:

- **At folder creation**: \`useFolderOperations.create()\` lists namespace members and calls \`add_group_members\` for each, explicitly adding them as direct members of the new subgroup.
[...]

PR #2359, opened 2026-05-14
test(e2e): reproducer for Open-subgroup metadata discovery (#2358)

### LINKED ISSUE #2358: feat(context): propagate Open-subgroup metadata to namespace members — fix the name discovery blocker for non-subgroup-members
## Summary

`setGroupMetadata(folder_id, { name, data })` publishes `GroupOp::GroupMetadataSet` at **subgroup scope** — encrypted with the subgroup's group key. Namespace members who haven't yet materialized membership in the subgroup (Bob, post-`join_namespace`, looking at an Open folder created by Alice) **cannot decrypt the op**, so the MetadataRecord never lands in their local store, so `listSubgroups(namespace_id)` returns `{ groupId }` with `name: None`.

For Open subgroups this is the wrong default. The `name` field exists specifically to let workspace members **discover** what they're joining — and Open semantically means "visible to all namespace members." Today the name only becomes visible after the user has already joined the subgroup, defeating the discovery purpose.

## Reproducer

1. Alice: `create_namespace` → invite Bob → Bob `join_namespace`.
2. Alice: `create_group_in_namespace(name: "Plans")` → returns `folder_id`.
3. Alice: `set_subgroup_visibility(folder_id, Open)`.
4. Alice: `set_group_metadata(folder_id, { name: "Plans" })`.
5. Bob: `list_subgroups(namespace_id)`.

Expected: `[{ groupId: <folder_id>, name: "Plans" }]`
Actual: `[{ groupId: <folder_id> }]` — `name` field absent.

## Where it lives in the code

`crates/context/src/handlers/set_group_metadata.rs`:
```rust
self.sign_and_publish_group_op(
    &group_id,
    requester,
    false,
    GroupOp::GroupMetadataSet { name, data },
)
```

`GroupOp` is the subgroup-encrypted-state-DAG path. Compare: `set_subgroup_visibility` also publishes a `GroupOp` but *its* flag propagates to namespace members (apparently via a special advertisement, since `check_group_membership_path` reads the visibility flag from Bob's local store and uses it to authorize inheritance — which requires Bob to have received the flag without being a subgroup member yet). Whatever that mechanism is, `GroupMetadataSet` doesn't ride it.

The receiving side: `crates/server/src/admin/handlers/groups/list_subgroups.rs:31`:
```rust
let name = get_group_metadata(&state.store, &child)
    .map(|rec| rec.and_then(|r| r.name))?;
```

Returns `None` whenever the local store doesn't have the record — which is always-the-case for non-members of subgroup-keyed state.

## Proposed fix — namespace-scope publication for Open subgroups

The semantic principle: **Open = name is public to the namespace; Restricted = name is private to the subgroup.** Matches the same logic that drives #2261 / #2351 — Open subgroups already have a parallel namespace-visible governance path; metadata should ride that path too.

Two implementation shapes, pick the cleanest:

**(a) New `RootOp` variant.** Add `RootOp::OpenSubgroupMetadataSet { group_id, name, data }`. When `set_group_metadata` runs against a subgroup whose visibility is `Open`, publish this variant at namespace scope instead of (or in addition to) the subgroup-scoped `GroupOp::GroupMetadataSet`. Bob's node, as a namespace member, decrypts + applies → his local `GroupMetadata(folder_id)` store gets populated → `list_subgroups` returns the name.

**(b) Dual-publish.** Always publish the subgroup-scoped op (existing behavior); additionally publish a namespace-scope "name advertisement" for Open subgroups. Same wire payload, two encryption envelopes. Receivers merge into the same store key.

**(c) Field-level split.** Just the `name` field rides the namespace-scope channel for Open subgroups; the opaque `data` map stays subgroup-private. This keeps maximum confidentiality (user-defined custom keys in `data` aren't unintentionally exposed) while making the discoverability story work.

Suggest **(c)** — minimizes blast radius for users storing app-private state in `data` (which mero-drive uses for folder color, and which other apps may use for sensitive ACL hints).

## Toggle semantics on visibility change

When a subgroup's `set_subgroup_visibility` flips Open → Restricted, the namespace-scope name advertisement should be invalidated (tombstone or counter-op) so non-members stop seeing the name. Inverse on Restricted → Open: re-publish at namespace scope.

## Why now

- Companion to the discovery-join gap (issue [#TODO_OTHER_LINK]). Without **B**, even after **A** lands the user opens a sidebar of `folder-2438479f` truncated-id stubs and has no idea what to click.
- This is the structural blocker on every Calimero-built app using Open subgroups as a directory: discovery requires names; names require subgroup membership; subgroup membership requires the user to have already discovered the folder by some other channel. Closed loop.

## Acceptance criteria

- For Open subgroups, `name` propagates to all namespace members within governance gossip latency (no manual `join_context` required).
- For Restricted subgroups, current behavior is preserved (name confidential to subgroup members).
- `data` map's visibility scope is documented explicitly (whichever (a)/(b)/(c) is chosen — let `data` stay private under (c)).
- Visibility toggle (Open ↔ Restricted) correctly invalidates / re-publishes the namespace-scope advertisement.
- Merobox e2e workflow `workflow-mero-drive-metadata-records.yml` is extended: after Alice sets the folder name + visibility=Open, Bob's `list_subgroups` returns the row with `name` set, before Bob calls any `join_context`.

rtb-12:
## Summary

Adds a failing e2e workflow (`group-open-subgroup-metadata-discovery.yml`) that exercises the exact reproducer in #2358: Bob joins a namespace, never the subgroup, and tries to read the Open subgroup's name via `list_subgroups` + `get_group_metadata`. The asserts must currently fail on master.

This PR is intentionally **reproducer-only** — no fix yet. The point is to get empirical evidence on CI of what fails and where, before designing the fix. The fix will land as follow-up commits on this same branch so CI proves before-and-after on a single PR.

## Why this approach

Reading the code suggested the bug *should not exist* on master:
- `GroupGovernancePublisher` already encrypts ops with the **namespace key** for Open subgroups whose ancestor chain is fully Open (`group_governance_publisher.rs:178-189`, issue #2256).
- The receiver falls back to the namespace keyring on a subgroup-key miss (`namespace_governance.rs:302-309`).

So if the reproducer fails as #2358 claims, the failure is downstream of decryption — most likely the receive-side `require_can_manage_metadata(signer)` check walking Bob's local store for Alice's admin/cap row in a subgroup Bob never received membership state for. This PR isolates that question to CI logs.

## Test plan

- [x] `merobox bootstrap validate` passes locally.
- [ ] CI matrix `test (group-open-subgroup-metadata-discovery)` fails at the `Bob sees the subgroup name via list_subgroups` or `Bob can fetch the metadata record directly` assertion (expected pre-fix).
- [ ] Inspect node-2 logs from the CI run to confirm whether the metadata op was decrypted-and-applied vs. decrypted-and-rejected vs. never received.
- [ ] Subsequent commits on this branch implement the minimal fix; same workflow flips to passing.

Closes #2358 once the fix lands.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it expands the CI E2E matrix with a workflow intended to fail until a follow-up fix lands, which can block PRs and increase CI noise/runtime.
> 
> **Overview**
> Adds a new `scaffolding-e2e` workflow, `group-open-subgroup-metadata-discovery.yml`, that reproduces issue #2358 by having Bob join a namespace (but *not* the open subgroup) and asserting the subgroup’s metadata is visible via both `list_subgroups` and `get_group_metadata`.
> 
> Registers this new workflow in the `.github/workflows/e2e-rust-apps.yml` matrix so it runs in CI alongside existing E2E scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8f488432f1343320790e0fcf5328ad13ef40f5b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
+++ b/apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
@@ -1,0 +1,204 @@
+name: E2E - Open Subgroup Metadata Discovery (issue #2358)
+description: >
+  Verifies that the `name` (and surrounding MetadataRecord) of an
+  Open subgroup is visible to namespace members who have NOT joined
+  that subgroup. This is the discovery prerequisite issue #2358
+  describes: without it, namespace members see a sidebar of
+  truncated-id stubs and cannot know what they're about to join.
+
+  Reproducer (per #2358 body):
+    1. Alice: create_namespace → invite Bob → Bob join_namespace.
+    2. Alice: create_group_in_namespace("Plans") → folder_id.
+    3. Alice: set_subgroup_visibility(folder_id, Open).
+    4. Alice: set_group_metadata(folder_id, name="Plans", data=...).
+    5. Bob: list_subgroups(namespace_id).
+       Expected: [{groupId, name: "Plans"}]
+       Pre-fix: [{groupId}]   (name field absent)
+
+  Bob never calls join_context / join_subgroup — he is a pure
+  namespace member. The crypto path that should make this work:
+  `GroupGovernancePublisher` already encrypts ops with the
+  *namespace* key for Open subgroups under an Open chain (issue
+  #2256), and `NamespaceGovernance` falls back to the namespace
+  keyring on receive (namespace_governance.rs ~302-309). So Bob
+  CAN decrypt; the failure must be downstream — most likely the
+  receive-side `require_can_manage_metadata(signer)` check, which
+  walks Bob's local store for Alice's admin/cap row in the
+  subgroup. Bob never received subgroup-keyed membership state,
+  so that lookup falls through to `is_inherited_admin`, which
+  may or may not resolve depending on what Alice's authority
+  looks like on Bob's side at apply time.
+
+  This workflow exists to confirm the failure mode empirically
+  before designing the fix. Both `list_subgroups` (matches the
+  issue's stated reproducer) and `get_group_metadata` are asserted
+  so we can tell exactly where the data does or does not arrive.
+
+  Companion to:
+    - #2256 / #2261 — Open-subgroup namespace-key encryption
+    - #2351 — RootOp::MemberJoinedOpen for self-join key delivery
+
+force_pull_image: false
+near_devnet: false
+
+nodes:
+  count: 2
+  image: merod:local
+  prefix: open-md-node
+
+steps:
+  # ── Phase 1: Bootstrap ────────────────────────────────────────────
+
+  - name: Install application on node-1 (Alice)
+    type: install_application
+    node: open-md-node-1
+    path: res/scaffolding_e2e.wasm
+    dev: true
+    outputs:
+      app_id: applicationId
+
+  - name: Install application on node-2 (Bob)
+    type: install_application
+    node: open-md-node-2
+    path: res/scaffolding_e2e.wasm
+    dev: true
+
+  - name: Alice creates namespace
+    type: create_namespace
+    node: open-md-node-1
+    application_id: '{{app_id}}'
+    outputs:
+      namespace_id: namespaceId
+      alice_key: memberPublicKey
+
+  - name: Alice invites Bob to namespace
+    type: create_namespace_invitation
+    node: open-md-node-1
+    namespace_id: '{{namespace_id}}'
+    outputs:
+      ns_invite: invitation
+
+  - name: Wait for namespace governance to gossip
+    # Matches the post-#2344 opaque-leaf-regression timing used by
+    # group-open-self-join-key-delivery.yml; covers invitation
+    # visibility so join_namespace doesn't 404.
+    type: wait
+    seconds: 10
+
+  - name: Bob joins namespace
+    # NOTE: Bob never joins the subgroup. Default member capabilities
+    # include CAN_JOIN_OPEN_SUBGROUPS, but the issue specifies that
+    # discovery (seeing the name in the sidebar) must work *before*
+    # Bob has joined any subgroup. So this is the only join he does.
+    type: join_namespace
+    node: open-md-node-2
+    namespace_id: '{{namespace_id}}'
+    invitation: '{{ns_invite}}'
+    outputs:
+      bob_identity: memberIdentity
+
+  # ── Phase 2: Alice creates + names the Open subgroup ──────────────
+  # Order matches the issue's reproducer exactly: create → flip Open
+  # → set metadata. The visibility flip lands BEFORE the metadata op
+  # so the publisher's `is_open_chain_to_namespace` check sees Open
+  # at publish time → metadata op is encrypted with the namespace key.
+
+  - name: Alice creates subgroup in namespace
+    type: create_group_in_namespace
+    node: open-md-node-1
+    namespace_id: '{{namespace_id}}'
+    group_alias: plans-folder
+    outputs:
+      subgroup_id: groupId
+
+  - name: Alice marks subgroup as Open
+    type: set_subgroup_visibility
+    node: open-md-node-1
+    group_id: '{{subgroup_id}}'
+    visibility: open
+
+  - name: Alice sets subgroup metadata
+    # The "discovery payload": name + an opaque data map a frontend
+    # might use for UI hints. The issue's option (c) proposes that
+    # only `name` (not `data`) needs namespace-scope visibility, so
+    # asserting both lets us tell which fields actually propagate.
+    type: set_group_metadata
+    node: open-md-node-1
+    group_id: '{{subgroup_id}}'
+    record_name: 'Plans'
+    data:
+      topic: discovery-test
+      icon: folder
+
+  - name: Wait for visibility + metadata governance to gossip
+    # Two consecutive GroupOp publishes against the namespace topic.
+    # Bob's namespace_governance receives, attempts subgroup-keyring
+    # decrypt (no subgroup membership → miss), falls back to the
+    # namespace keyring (he has the namespace key as a namespace
+    # member). On a working implementation, both ops apply and the
+    # metadata record lands in Bob's GroupMetadata(subgroup_id) row.
+    type: wait
+    seconds: 10
+
+  # ── Phase 3: Bob reads — without joining the subgroup ─────────────
+  # CRITICAL: no join_context / join_group / join_subgroup is run for
+  # Bob on the subgroup. He only has namespace membership.
+
+  - name: Bob lists subgroups of the namespace
+    type: list_subgroups
+    node: open-md-node-2
+    group_id: '{{namespace_id}}'
+    outputs:
+      bob_subgroups: subgroups
+      # Dotted-path extraction per merobox nested-json-parsing
+      # docs — `subgroups.0.name` indexes into the first entry.
+      # If the subgroup is missing entirely the workflow will fail
+      # at this output resolution with a missing-key error, which
+      # is itself useful diagnostic signal.
+      bob_first_subgroup_name: subgroups.0.name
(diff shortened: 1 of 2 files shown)


apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
```diff
@@ -0,0 +1,204 @@
+name: E2E - Open Subgroup Metadata Discovery (issue #2358)
```
frdomovic, line 1: 🟡 **Intentionally failing test added to CI matrix without a skip/xfail mechanism**

The PR description explicitly states 'The asserts must currently fail on master' and the CI matrix entry is registered without any conditional skip or expected-failure marker. This means every PR and push to master will have a permanently red CI job until the follow-up fix lands. The existing matrix uses `fail-fast: false` so it won't block other jobs from running, but the `comment` job will still post a failure comment and the overall workflow will be marked failed. This creates CI noise and could mask real regressions in other matrix entries.

**Suggested fix:**
```
Either: (1) add a top-level `expected_failure: true` field to the workflow YAML if merobox supports it, (2) gate the matrix entry with a condition so it only runs on the fix branch, or (3) keep the workflow file in the repo but do NOT register it in the matrix until the fix is ready. The cleanest approach for a 'reproducer-only' PR is option 3 — add a comment in the matrix similar to the existing `partition-window-reconcile` comment block explaining why it is intentionally excluded.
```

apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
```diff
@@ -0,0 +1,204 @@
...
+    outputs:
+      subgroup_id: groupId
+
+  - name: Alice marks subgroup as Open
+    type: set_subgroup_visibility
+    node: open-md-node-1
+    group_id: '{{subgroup_id}}'
+    visibility: open
+
+  - name: Alice sets subgroup metadata
```
frdomovic, line 120: 🟡 **Output variable `bob_identity` is captured but never used in any subsequent step**

The `join_namespace` step captures `memberIdentity` into `bob_identity` (line 122), but no later step references `{{bob_identity}}`. This is dead output that adds confusion about whether Bob's identity is needed for a later assertion (e.g. checking Bob's membership row). Per the project's 'No Dead Code' convention, unused variables should be removed.

**Suggested fix:**
```
Remove the `outputs:` block from the `Bob joins namespace` step, or add an assertion that uses `{{bob_identity}}` if it is actually needed for diagnostic purposes.
```

apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
```diff
@@ -0,0 +1,204 @@
...
+  image: merod:local
+  prefix: open-md-node
+
+steps:
+  # ── Phase 1: Bootstrap ────────────────────────────────────────────
+
+  - name: Install application on node-1 (Alice)
+    type: install_application
+    node: open-md-node-1
+    path: res/scaffolding_e2e.wasm
```
frdomovic, line 55: 🟡 **10-second fixed wait for gossip is fragile and may be insufficient under load**

Two separate `type: wait` steps use a hard-coded 10-second delay to wait for governance gossip. Other workflows in the repo use `wait_for_sync` with a `timeout` and `trigger_sync: true` which is more deterministic. Under CI load or slow container startup the 10-second window may not be enough, causing flaky failures that are hard to distinguish from the actual bug being reproduced. Conversely, on fast hardware the wait is wasteful.

**Suggested fix:**
```
Replace the fixed waits with `wait_for_sync` steps targeting the namespace group_id on both nodes, with `trigger_sync: true` and a reasonable timeout (e.g. 30s). This is the pattern used by every other workflow in the matrix (e.g. group-3node.yml, dm-subgroup-privacy.yml).
```

apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
```diff
@@ -0,0 +1,204 @@
...
+    seconds: 10
+
+  - name: Bob joins namespace
+    # NOTE: Bob never joins the subgroup. Default member capabilities
+    # include CAN_JOIN_OPEN_SUBGROUPS, but the issue specifies that
+    # discovery (seeing the name in the sidebar) must work *before*
+    # Bob has joined any subgroup. So this is the only join he does.
+    type: join_namespace
+    node: open-md-node-2
+    namespace_id: '{{namespace_id}}'
```
frdomovic, line 95: 💡 **Missing wait_for_sync after Bob joins namespace before Alice creates the subgroup**

After Bob joins the namespace there is no synchronization step before Alice starts creating the subgroup and setting visibility. If Alice's `create_group_in_namespace` op is published before Bob's join op has fully propagated and been applied on node-1's side, the subgroup's GroupCreated event may be published before Bob's namespace membership is confirmed in the governance DAG. This could cause the subgroup-creation op to be rejected or ignored on Bob's side because his namespace membership row hasn't landed yet. A `wait_for_sync` on the namespace group between the join and the subgroup creation would make the ordering deterministic.

**Suggested fix:**
```
Add a `wait_for_sync` step after `Bob joins namespace` targeting `group_id: '{{namespace_id}}'` on both nodes before proceeding to Phase 2.
```

apps/scaffolding-e2e/workflows/group-open-subgroup-metadata-discovery.yml
```diff
@@ -0,0 +1,204 @@
...
+
+  - name: Bob lists subgroups of the namespace
+    type: list_subgroups
+    node: open-md-node-2
+    group_id: '{{namespace_id}}'
+    outputs:
+      bob_subgroups: subgroups
+      # Dotted-path extraction per merobox nested-json-parsing
+      # docs — `subgroups.0.name` indexes into the first entry.
+      # If the subgroup is missing entirely the workflow will fail
```
frdomovic, line 155: 💡 **Output path `subgroups.0.name` assumes a specific ordering of list_subgroups results**

The output extraction `bob_first_subgroup_name: subgroups.0.name` assumes the first element of the returned list is the subgroup Alice created. If `list_subgroups` returns results in an unspecified or insertion-order-dependent sequence (which is common with map-backed storage), this assumption may hold now but break if additional subgroups are ever present or if the ordering changes. The comment acknowledges a missing-key error as 'useful diagnostic signal', but an ordering bug would produce a misleading failure.

**Suggested fix:**
```
If merobox supports filtering or if the list is guaranteed to have exactly one entry in this test, add an assertion on the length of `bob_subgroups` before indexing. Alternatively, use a named lookup by groupId rather than positional indexing if the framework supports it.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 22% | Review time: 61.4s

🟡 3 warnings, 💡 2 suggestions. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-35b0291e`</sub>

frdomovic:
## Documentation Review

The following documentation may need updates based on the changes in this PR:

- 🟡 **AGENTS.md**: `AGENTS.md` exists but was not updated — consider updating it to reflect the architecture changes in this PR.
- 🟡 **CONTRIBUTING.md**: `CONTRIBUTING.md` exists but was not updated — consider updating it to reflect the architecture changes in this PR.
- 🟡 **docs/**: Static HTML docs in `docs/` may need updating — architecture-impacting changes detected. On merge, `update-docs` will scan this directory and open a PR if any pages need to change.

Issue #2356, opened 2026-05-14
sync/storage: Merkle root not recomputed after concurrent CRDT merge + residuals from closed #2293/#2122/#2336


rtb-12:
## Summary

Consolidated tracker for sync/storage residuals after #2344 (opaque-leaf + parent_id propagation + materialization race) and #2352 (gossipsub small-cluster tune + flood_publish) merged into master. Four follow-ups: one newly diagnosed during the #2344 investigation, three leftover items from closed issues #2293 / #2122 / #2336 that the merged stack did not address.

Reference: full diagnostic chain in PR #2344 comments [#issuecomment-4444014412](https://github.com/calimero-network/core/pull/2344#issuecomment-4444014412), [#issuecomment-4444030829](https://github.com/calimero-network/core/pull/2344#issuecomment-4444030829), [#issuecomment-4448484146](https://github.com/calimero-network/core/pull/2344#issuecomment-4448484146).

---

## 1. (NEW) Merkle root not recomputed after concurrent CRDT merge

**Symptom:** Two nodes that wrote concurrently to the same context end up with identical merged data but **different Merkle root hashes**, despite both running the WASM merge handler successfully. Cross-node convergence (`wait_for_sync`) reports divergence for ~30 s until a subsequent trigger forces a Merkle recompute.

**Reproduction in CI:** `Sync regression (smoke)` workflow on PR #2344 head `81710d5` (post-#2352 merge):

| Step | Time | Notes |
|---|---|---|
| Round 1 sync | **2.2 s** ✓ | flood_publish from #2352 delivers gossip in 24 ms |
| Round 2 sync | **31.6 s** ❌ | concurrent writes (14 ms apart) — bug fires |

Both rounds use writes to *different* keys (not LWW conflicts). The trigger is the back-to-back timing producing `Concurrent branch detected — applying with CRDT merge semantics` on both sides.

**Exact mechanism:** In [`Interface::apply_action`](https://github.com/calimero-network/core/blob/master/crates/storage/src/interface.rs#L696-L703):

```rust
let Some((_, _full_hash)) = Self::save_internal(id, &data, metadata.clone())? else {
    debug!(%id, "Remote action produced no storage change (save_internal returned None)");
    return Ok(());   // ← skips Action::Compare + parent add_child_to
};
```

[...]

PR #2365, opened 2026-05-15
DO NOT MERGE: verify #2356 test catches the race

rtb-12:
**DO NOT MERGE / DO NOT REVIEW** — temporary branch.

This is a verification branch that contains **only** the Phase 5 cold-start regression smoke + the GHA log-grep guard, **without** the responder-side materialisation-wait fix that the test is meant to catch.

Purpose: prove the test in the companion PR actually detects the bug it claims to. Once `sync-regression` CI runs here and the `Verify no premature inbound-stream closure (#2356)` step fails with the warn signature `inbound stream for unknown context, closing cleanly`, this branch is closed and the real PR (with both fix + test) is opened.

Tracked in [#2356](https://github.com/calimero-network/core/issues/2356).

Test-only commit: e42f28d6 (cherry-pick of the test commit from the real fix branch `fix/2356-inbound-stream-group-materialisation`).

### DIFF
--- a/.github/workflows/sync-regression.yml
+++ b/.github/workflows/sync-regression.yml
@@ -150,3 +150,28 @@ jobs:
           path: docker-logs/
           retention-days: 7
           if-no-files-found: ignore
+
+      - name: Verify no premature inbound-stream closure (#2356)
+        # Regression guard for the materialisation-wait short-circuit
+        # fixed in this PR. The Phase 5 cold-start race in
+        # `opaque-leaf-regression.yml` may or may not fire the
+        # underlying timing race on a given runner — gossipsub
+        # propagation in a stable 2-node mesh is fast — but when it
+        # does fire pre-fix, node-2's log carries the exact warn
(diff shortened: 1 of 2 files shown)


rtb-12:
Verification complete — the Phase 5 test + grep guard catch the bug as designed.

Failed sync-regression run: https://github.com/calimero-network/core/actions/runs/25913582551

Concrete evidence:

1. **Phase 5 `wait_for_sync` timed out** at 31.68s/15 attempts — same shape as mero-drive PR #32's `Sync Registry after nested-folder registration` failure.
2. **Grep guard fired** with the exact smoking-gun warn line:
   ```
   WARN calimero_node::sync::manager: inbound stream for unknown context, closing cleanly
       context_id=8JARAAZW14V4A3UToJ4FPX4yyeDt7ptU2hSyqsHjdyym
   ```
3. The `context_id` named is the second context Phase 5 creates — the cold-start race code path.

Closing — the real PR (with fix + test) is being opened separately.

PR #2372, opened 2026-05-16
fix(context): include inherited members in list_group_members (#2371)

### LINKED ISSUE #2371: join_subgroup_inheritance returns 200 but list_group_members doesn't include the joiner — breaks app-side membership checks
## Summary

`POST /admin-api/groups/<id>/join-via-inheritance` (the endpoint introduced in #2360) returns 200 with `wasInherited: true`, but a follow-up `GET /admin-api/groups/<id>/members` on the **same node** still returns only the explicit members — the joiner's identity surfaces only in the `selfIdentity` field, never in `members[]`. App callers that key "is the caller a member of this subgroup?" off `list_group_members` therefore see `isMember = false` for any user who joined via inheritance, even though the join succeeded and the subgroup key was delivered.

## Reproduction

Run the merobox workflow `workflow-mero-drive-open-subgroup.yml` from [mero-drive#32](https://github.com/calimero-network/mero-drive/pull/32) (commit `2abbafd` or later). The workflow:

1. Alice creates namespace `Phoenix` and an Open subgroup `Reports`.
2. Alice creates a docs context inside `Reports`.
3. Bob `join_namespace` (joins Phoenix root, NOT explicit member of Reports).
4. Bob `join_subgroup_inheritance(group_id=Reports)` → returns 200, `wasInherited=true`. ✓
5. Bob `list_group_members(group_id=Reports)` → **returns only Alice**, Bob is in `selfIdentity` only.

The downstream Playwright e2e tests reproduce automatically — every `clickJoin` flow stays mounted on `RestrictedFolderCard` because `useMemberCaps` reads the empty membership.

### Confirmed wire shape

From [Playwright trace artifact 7023581693](https://github.com/calimero-network/mero-drive/actions/runs/25932145102) on Bob's node (sha `86970319f0d3`) after a successful POST:

```json
{
  "members": [{ "identity": "<alice-pubkey>", "role": "Admin" }],
  "selfIdentity": "<bob-pubkey>"
}
```

The POST returned `200 OK` 100ms earlier at `18:03:55.907Z`. Bob's POST + the same node's subsequent `GET /members` are 100% causally ordered (single-node, same admin-api process).

## Why this happens (current code)

[`crates/context/src/group_store/namespace_governance.rs:1185-1246`](https://github.com/calimero-network/core/blob/main/crates/context/src/group_store/namespace_governance.rs#L1185-L1246) — `execute_member_joined_open` validates the op (signer = member, inheritance path exists, namespace pin) and explicitly does NOT touch storage. The doc comment is unambiguous:

> *"We don't mutate state here — the side-effect (pushing a `PendingKeyDelivery` if we hold the key) happens in the outer `apply_signed_op` match."*

[`crates/context/src/group_store/membership.rs:467-491`](https://github.com/calimero-network/core/blob/main/crates/context/src/group_store/membership.rs#L467-L491) — `list_group_members` reads rows under `GROUP_MEMBER_PREFIX`. Those rows are only written by `apply_member_joined` (the explicit `MemberJoined` op) and `add_group_members`. Inherited members never get a row.

## Impact

The contract gap surfaces in any app that uses `list_group_members` (the canonical, documented "who's in this group?" call) to gate UI affordances. Specifically, in mero-drive:

- `useMemberCaps` (`app/src/hooks/useMemberCaps.ts`) uses `listGroupMembers` → `find(m => m.identity === self)` → reports `error: "identity is not a member of group"` after 4 retries.
- `useFolderPermissions.isMember` falls through to `false`.
- `WorkspaceLayout` renders `RestrictedFolderCard` instead of the document list.
- Every Playwright two-node test that exercises `clickJoin` then `expectDocVisible` times out: ~10 tests (doc-collab × 3, restricted-folder-invite × 2, visibility-toggle × 2, open-folder-inheritance SMOKE + 2 more, folder-name-propagation × 2).

The same gap will hit any other Calimero app that uses `list_group_members` to determine effective membership for Open subgroups.

## Why workarounds don't apply

- **Polling the endpoint** — the row is never written, so polling never converges.
- **Using the context layer directly** (`call` with `executor_public_key`) — works because `check_group_membership_path` recognises inheritance. But the admin API is the documented surface for permission UIs; apps shouldn't have to bypass it.
- **Re-implementing inheritance walk client-side** — the algorithm needs full visibility into the parent-chain GroupMeta state and the visibility flips of every ancestor; not feasible from outside the node.

## Suggested fixes

**Option A (preferred):** Make `execute_member_joined_open` write a `GroupMember` row when applying on the joiner's own node (or any node that holds the subgroup key). Re-use `NamespaceMembershipService::apply_member_joined` as the template, optionally tagging the row with `via_inheritance: true` so callers can distinguish if needed. This makes `list_group_members` self-consistent with the membership contract — app callers need no changes.

**Option B:** Add an opt-in `include_inherited=true` query parameter to `GET /admin-api/groups/<id>/members`. The handler unions the stored rows with a graph walk via `check_group_membership_path` for namespace members. Adds query-time cost; requires app callers to know to pass the flag.

**Option C:** Document the gap as intentional and add a separate `GET /admin-api/groups/<id>/effective-members` endpoint with the union semantics. Most explicit, requires most app coordination.

A is the cleanest because it preserves the existing semantics that callers already depend on. B/C require every app to opt in.

## Adjacent fixes already landed in mero-drive PR #32 (these correctly close other root causes but cannot work around this one)

- `c7d2486` — normalise lowercase `subgroup_visibility` from server (was strict-equal mismatch against `"Open"`/`"Restricted"`).
- `97c7a49` — prefer just-joined namespace in `useDriveWorkspace` auto-select (Bob was landing on `namespaces[0]` from prior tests' server-side memberships).
- `3c17e3e` — `useMemberCaps` exposes `refetch()`, called from `RestrictedFolderCard.onJoin` so the cap check re-evaluates after the join (works but the re-evaluation still finds Bob absent from `members[]`).

## Asks

- Confirm the intended contract for `join_subgroup_inheritance` — does it promise that `list_group_members` afterwards includes the joiner? If yes, this is a core bug; Option A. If no, the docs need to call this out explicitly and apps need a different membership-check primitive — Option B or C.

rtb-12:
## Summary

Fixes #2371. `POST /admin-api/groups/<id>/join-via-inheritance` returns `200 / wasInherited: true`, but a follow-up `GET /admin-api/groups/<id>/members` on the **same node** omitted the joiner — they surfaced only in `selfIdentity`, never in `members[]`. Apps that key "is the caller a member of this subgroup?" off `list_group_members` therefore saw `isMember = false` for any user who joined via inheritance.

**Root cause** (confirmed against `master`):
- `execute_member_joined_open` (`namespace_governance.rs`) is a validate-only apply — it writes no `GroupMember` row by design.
- `join_subgroup_inheritance` fetches the key and publishes `MemberJoinedOpen`, but never writes a row either.
- `list_group_members` (`membership.rs`) reads only `GROUP_MEMBER_PREFIX` rows, so an inherited joiner is invisible to it — even though `check_group_membership` (used to gate the same endpoint) *does* recognise the inheritance path and returns `true`.

## Approach

Inherited membership is **derived state** — a pure function of the joiner's ancestor memberships and the Open/Restricted visibility of every group on the chain. Both change with no op naming the affected subgroup (a cap revocation, an ancestor flipping Open→Restricted, a namespace leave). Materialising a `GroupMember` row (issue's Option A) would create a stale-cache problem with no invalidation events, so the fix recomputes at query time instead.

- **`enumerate_inherited_members`** — the list-dual of `check_group_membership_path`. Walks the same Open parent-chain and re-runs the single-identity check per ancestor candidate (direct members + each ancestor's `GroupMeta` admin), so enumeration and the single-identity check can never diverge on the anchor-cap / admin-override rules.
- **Handler union** — `list_group_members` handler now returns `stored rows ∪ enumerate_inherited_members`, paginated over the union. `group_store::list_group_members` itself is left direct-rows-only — its internal callers (key distribution, `verify_ack`, …) must keep seeing only stored explicit rows.

App callers need no changes; `list_group_members` becomes self-consistent with `check_group_membership`.

## Tests

- **3 unit tests** (`group_store/tests.rs`) — written test-first (RED→GREEN): the inherited joiner appears as `Member`; an inherited admin appears as `Admin`; a Restricted subgroup inherits nobody.
- **e2e** (`group-join-via-inheritance.yml`) — after `join_subgroup_inheritance` and before `join_context` (so the joiner provably holds no explicit row), `list_group_members` on the joiner's own node now asserts the joiner is in `members[]`. In-core analogue of mero-drive#32's failing step 5.

## Verification

- `cargo test -p calimero-context --lib group_store::` → 206 passed, 0 failed.
- `cargo build -p calimero-context` → clean.
- The merobox e2e was validated statically (YAML + step schema); it was not run here (needs Docker + `merod:local`), it runs in the scaffolding-e2e CI suite.

## Not included (optional follow-up)

A `via_inheritance` flag on the API entry — the issue marks it optional, and it would touch `context-client` + `server-primitives` wire types. mero-drive's `useMemberCaps` only needs presence, which this delivers.

---

> [!NOTE]
> **Medium Risk**
> Changes API-visible membership semantics and pagination/count behavior by computing inherited members at query time; risk is moderate due to potential performance impact and edge cases in namespace-chain traversal.
> 
> **Overview**
> Fixes #2371 by making **effective membership** visible via the admin API: `list_group_members` now returns explicit stored members **plus** members who join an `Open` subgroup via inheritance, using new `group_store::enumerate_inherited_members`.
> 
> `get_group_info`’s `member_count` is updated to match the effective member set (stored + inherited) so clients paginating against the count don’t miss inherited members. Adds unit tests around inherited enumeration (member vs admin role, restricted walls, depth overflow/boundary), and extends the `group-join-via-inheritance` e2e workflow to assert the inherited joiner appears in `members[]` before any `join_context` call.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 725215c46715ff02806766a38fb56887e68188cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/scaffolding-e2e/workflows/group-join-via-inheritance.yml
+++ b/apps/scaffolding-e2e/workflows/group-join-via-inheritance.yml
@@ -15,6 +15,13 @@ description: >
        context under the subgroup to author *and* decrypt traffic
        successfully — i.e. the joiner's `sender_key` is populated and
        bidirectional read/write converges.
+    3. Surfaces the inherited joiner in `list_group_members` on their
+       own node — the effective-membership contract from issue #2371.
+       `join_subgroup_inheritance` writes no `GroupMember` row (the
+       apply path `execute_member_joined_open` is validate-only), so
+       before the fix `GET /admin-api/groups/<id>/members` returned
+       only the explicit members and apps keying "is the caller a
+       member?" off that list saw `false` for any inherited joiner.
 
   Pre-#2357 the only way to materialize the inherited path was to
   discover a child context id and call `join_context` against it
@@ -137,6 +144,34 @@ steps:
     type: json_assert
     statements:
       - 'json_equal({{inherit_was_inherited}}, true)'
+
+  # ── Phase 2b: list_group_members reflects the inherited joiner ────
+  # Issue #2371 regression. node-2 just joined `subgroup` via
+  # inheritance and has NOT yet called `join_context` — so it holds no
+  # explicit `GroupMember` row anywhere in the subgroup. Listing the
+  # subgroup members on node-2's own node must still include node-2:
+  # the admin-api handler unions the stored rows with
+  # `enumerate_inherited_members`. Before the fix this assertion fails
+  # — node-2's identity appeared only in `selfIdentity`, never in
+  # `members[]`.
+
+  - name: 'Node-2 lists subgroup members (issue #2371)'
+    type: list_group_members
+    node: join-inherit-node-2
+    group_id: '{{subgroup_id}}'
+    outputs:
+      subgroup_members: members
+
+  - name: Assert inherited joiner appears in the member list
+    type: assert
+    statements:
+      # node-2 joined `subgroup` via inheritance and holds no explicit
+      # `GroupMember` row. Before the #2371 fix it was absent from
+      # `members[]` (surfaced only in `selfIdentity`); the admin-api
+      # handler now unions inherited members in, so node-2's identity
+      # (`inherit_member_pk`, exported by the join-via-inheritance step)
+      # must appear in the listed members.
+      - 'contains({{subgroup_members}}, {{inherit_member_pk}})'
 
   # ── Phase 3: node-2 joins the context (key already delivered) ─────
   # This call should now be cheap because the group key is already
--- a/crates/context/src/group_store/membership.rs
+++ b/crates/context/src/group_store/membership.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeSet;
+
 use calimero_context_config::types::ContextGroupId;
 use calimero_context_config::{MemberCapabilities, VisibilityMode};
 use calimero_primitives::context::GroupMemberRole;
@@ -309,6 +311,100 @@ pub fn check_group_membership(
         check_group_membership_path(store, group_id, identity)?,
         MembershipPath::None,
     ))
+}
+
+/// Enumerate the identities that are members of `group_id` purely by
+/// inheritance — i.e. those with a [`MembershipPath::Inherited`] path
+/// and no stored `GroupMember` row in `group_id` itself.
+///
+/// This is the "list" dual of [`check_group_membership_path`]: rather
+/// than testing one identity, it walks the same `Open` parent-chain and
+/// collects every ancestor's direct members that resolve to `Inherited`
+/// for `group_id`. The role is `Admin` for an inherited-admin path
+/// (`via_admin`) and `Member` otherwise.
+///
+/// Callers union the result with [`list_group_members`] (the stored,
+/// explicit rows) to obtain the *effective* member set — the contract
+/// `check_group_membership` already honours. Membership is recomputed
+/// from current ancestor state on every call, so it never goes stale
+/// when an ancestor's visibility flips or a cap is revoked (issue
+/// #2371).
+pub fn enumerate_inherited_members(
+    store: &Store,
+    group_id: &ContextGroupId,
+) -> EyreResult<Vec<(PublicKey, GroupMemberRole)>> {
+    // Direct members of `group_id` are explicit, not inherited — seed
+    // `seen` with them so they are never reported here. `seen` doubles
+    // as the dedup set across ancestor levels. `BTreeSet` (not
+    // `HashSet`) because `PublicKey` derives `Ord` but not `Hash`;
+    // O(log n) lookups are ample for governance-scale membership.
+    let mut seen: BTreeSet<PublicKey> = list_group_members(store, group_id, 0, usize::MAX)?
+        .into_iter()
+        .map(|(pk, _)| pk)
+        .collect();
+    let mut result = Vec::new();
+
+    // Walk the `Open` parent-chain, mirroring `check_group_membership_path`'s
+    // termination (first `Restricted` ancestor or namespace root, bounded
+    // by `MAX_NAMESPACE_DEPTH`). At each ancestor, every direct member —
+    // plus the `GroupMeta` admin, who may have no direct row of their own
+    // — is a candidate. `check_group_membership_path` then decides, from
+    // `group_id`, whether the candidate actually inherits and via which
+    // path. Reusing it verbatim keeps enumeration and the single-identity
+    // check from ever diverging on the anchor-cap / admin-override rules.
+    let mut current = *group_id;
+    // `terminated` distinguishes a legitimate exit (hit a `Restricted`
+    // wall or the namespace root) from loop-bound exhaustion. If the
+    // bound is hit without terminating, the parent chain has a cycle —
+    // bail rather than silently return a partial set, matching
+    // `check_group_membership_path` / `is_inherited_admin`.
+    let mut terminated = false;
+    for _ in 0..=MAX_NAMESPACE_DEPTH {
+        if get_subgroup_visibility(store, &current)? != VisibilityMode::Open {
+            terminated = true;
+            break;
+        }
+        let Some(parent) = get_parent_group(store, &current)? else {
+            terminated = true;
+            break;
+        };
+
+        let mut candidates: Vec<PublicKey> = list_group_members(store, &parent, 0, usize::MAX)?
+            .into_iter()
+            .map(|(pk, _)| pk)
+            .collect();
+        if let Some(meta) = load_group_meta(store, &parent)? {
+            candidates.push(meta.admin_identity);
+        }
+
+        for candidate in candidates {
+            // `insert` returns false when `candidate` was already seen
+            // (a direct member of `group_id`, or processed at a deeper
+            // ancestor) — skip the redundant path check in that case.
+            if !seen.insert(candidate) {
+                continue;
+            }
+            if let MembershipPath::Inherited { via_admin, .. } =
+                check_group_membership_path(store, group_id, &candidate)?
+            {
+                let role = if via_admin {
+                    GroupMemberRole::Admin
+                } else {
+                    GroupMemberRole::Member
+                };
+                result.push((candidate, role));
+            }
+        }
+
+        current = parent;
+    }
+    if !terminated {
+        bail!(
+            "enumerate_inherited_members exceeded MAX_NAMESPACE_DEPTH \
+             ({MAX_NAMESPACE_DEPTH}); possible cycle in store"
+        );
+    }
+    Ok(result)
 }
 
 /// Returns `true` if `identity` is a direct admin of this specific group
--- a/crates/context/src/handlers/list_group_members.rs
+++ b/crates/context/src/handlers/list_group_members.rs
@@ -27,11 +27,37 @@ impl Handler<ListGroupMembersRequest> for ContextManager {
                 bail!("node is not a member of group '{group_id:?}'");
             }
 
-            let members =
-                group_store::list_group_members(&self.datastore, &group_id, offset, limit)?;
+            // Effective membership = stored explicit rows ∪ inherited
+            // members of `Open` subgroups. `list_group_members` alone
+            // returns only the stored rows; a peer who joined via
+            // `join_subgroup_inheritance` has no row (the apply path
+            // `execute_member_joined_open` is validate-only), so without
+            // the union they would never surface here even though
+            // `check_group_membership` reports them as members (#2371).
+            //
+            // Pagination must span the union, so the full sets are
+            // collected here and sliced after the merge rather than
+            // pushed down into the store query. Known cost: a paginated
+            // call is O(total effective members) in store reads, not
+            // O(offset+limit). This is acceptable here — group
+            // membership is governance-scale (a permission hierarchy),
+            // not an unbounded user list — and the inherited set cannot
+            // be store-paginated at all since it is computed, not
+            // stored. The inherited set is disjoint from the stored
+            // rows by construction (`enumerate_inherited_members`
+            // excludes direct members of `group_id`), so no dedup is
+            // needed.
+            let mut members =
+                group_store::list_group_members(&self.datastore, &group_id, 0, usize::MAX)?;
+            members.extend(group_store::enumerate_inherited_members(
+                &self.datastore,
+                &group_id,
+            )?);
 
             let entries = members
                 .into_iter()
+                .skip(offset)
+                .take(limit)
                 .map(|(identity, role)| {
                     let name =
                         group_store::get_member_metadata(&self.datastore, &group_id, &identity)
--- a/crates/context/src/group_store/mod.rs
+++ b/crates/context/src/group_store/mod.rs
@@ -74,8 +74,9 @@ pub use self::local_state::{
 };
 pub use self::membership::{
     add_group_member, add_group_member_with_keys, check_group_membership,
-    check_group_membership_path, count_group_admins, count_group_members, get_group_member_role,
-    get_group_member_value, has_direct_group_member, is_direct_group_admin, is_group_admin,
+    check_group_membership_path, count_group_admins, count_group_members,
+    enumerate_inherited_members, get_group_member_role, get_group_member_value,
+    has_direct_group_member, is_direct_group_admin, is_group_admin,
     is_group_admin_or_has_capability, is_inherited_admin, list_group_members,
     namespace_member_pubkeys, remove_group_member, require_group_admin,
     require_group_admin_or_capability, set_member_auto_follow, trusted_anchors_for_group,
--- a/crates/context/src/group_store/tests.rs
+++ b/crates/context/src/group_store/tests.rs
@@ -2293,6 +2293,210 @@ fn check_membership_anchor_cap_check_uses_deepest_direct_membership() {
     assert!(!check_group_membership(&store, &leaf, &alice).unwrap());
 }
 
+// -----------------------------------------------------------------------
+// Effective membership for `Open` subgroups — issue #2371
+//
+// `join_subgroup_inheritance` returns 200 / `wasInherited: true` but
+// writes no `GroupMember` row — `execute_member_joined_open` is a
+// validate-only apply. `check_group_membership` correctly reports the
+// inherited joiner as a member, yet `list_group_members` (which only
+// reads stored rows) omits them, so an app keying "is the caller a
+// member?" off the member list sees `false`. `enumerate_inherited_members`
+// closes the gap: it recomputes inherited members from current ancestor
+// state so callers can union it with the stored rows to get the
+// effective member set the membership contract promises.
+// -----------------------------------------------------------------------
+
+#[test]
+fn enumerate_inherited_members_includes_open_subgroup_joiner() {
+    use calimero_context_config::{MemberCapabilities, VisibilityMode};
+
+    // namespace (root) -> reports (Open subgroup).
+    // Alice is a direct Admin of `reports`; Bob is a namespace member
+    // who joined `reports` via inheritance — no direct row in `reports`.
+    let store = test_store();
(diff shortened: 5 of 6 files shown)


crates/context/src/group_store/membership.rs
```diff
@@ -311,6 +311,81 @@ pub fn check_group_membership(
...
+/// #2371).
+pub fn enumerate_inherited_members(
+    store: &Store,
+    group_id: &ContextGroupId,
+) -> EyreResult<Vec<(PublicKey, GroupMemberRole)>> {
+    // Direct members of `group_id` are explicit, not inherited — record
+    // them up front so they are never reported here.
+    let mut seen: Vec<PublicKey> = list_group_members(store, group_id, 0, usize::MAX)?
+        .into_iter()
+        .map(|(pk, _)| pk)
```
frdomovic, line 343: 🟡 **O(n²) linear search in `seen` Vec for large ancestor member sets**

The `seen` list is a `Vec<PublicKey>` and `seen.contains(&candidate)` is O(n) per lookup. For groups with large ancestor member sets (e.g., a namespace with thousands of members), the total cost of `enumerate_inherited_members` becomes O(n²) — once for each candidate across all ancestor levels. Since `PublicKey` is `[u8; 32]` and thus `Hash + Eq`, a `HashSet<PublicKey>` would reduce this to O(n) overall. The same issue applies to the initial population of `seen` from `list_group_members`.

**Suggested fix:**
```
Replace `let mut seen: Vec<PublicKey> = ...` with `let mut seen: std::collections::HashSet<PublicKey> = ...` and change `seen.contains(&candidate)` to `seen.contains(&candidate)` (same call, but O(1)). Also change `seen.push(candidate)` to `seen.insert(candidate)`.
```

crates/context/src/handlers/list_group_members.rs
```diff
@@ -27,11 +27,31 @@ impl Handler<ListGroupMembersRequest> for ContextManager {
...
 
-            let members =
-                group_store::list_group_members(&self.datastore, &group_id, offset, limit)?;
+            // Effective membership = stored explicit rows ∪ inherited
+            // members of `Open` subgroups. `list_group_members` alone
+            // returns only the stored rows; a peer who joined via
+            // `join_subgroup_inheritance` has no row (the apply path
+            // `execute_member_joined_open` is validate-only), so without
+            // the union they would never surface here even though
+            // `check_group_membership` reports them as members (#2371).
```
frdomovic, line 36: 🟡 **Full member set loaded into memory before pagination — O(N) memory for large groups**

The handler now fetches ALL stored members (`0, usize::MAX`) and ALL inherited members into memory, concatenates them, then applies `skip(offset).take(limit)`. For groups with thousands of members, this materializes the entire member set on every paginated request, even when the client only wants 10 results. The original code pushed pagination down into the store scan (`collect_keys_with_prefix_paginated`), which was O(offset+limit) in store reads. The new approach is O(total_members) in both memory and store reads regardless of the requested page size.

**Suggested fix:**
```
Consider a two-phase approach: if `offset < stored_count`, serve from stored rows first (using the existing paginated store query), and only invoke `enumerate_inherited_members` when the offset falls into the inherited range. Alternatively, document the known performance regression and add a cap on group size or a warning log when the total exceeds a threshold.
```

crates/context/src/group_store/membership.rs
```diff
@@ -311,6 +311,81 @@ pub fn check_group_membership(
...
+    // — is a candidate. `check_group_membership_path` then decides, from
+    // `group_id`, whether the candidate actually inherits and via which
+    // path. Reusing it verbatim keeps enumeration and the single-identity
+    // check from ever diverging on the anchor-cap / admin-override rules.
+    let mut current = *group_id;
+    for _ in 0..=MAX_NAMESPACE_DEPTH {
+        if get_subgroup_visibility(store, &current)? != VisibilityMode::Open {
+            break;
+        }
+        let Some(parent) = get_parent_group(store, &current)? else {
```
frdomovic, line 367: 💡 **GroupMeta admin added as candidate but may already be in `list_group_members` result**

The code adds `meta.admin_identity` to `candidates` unconditionally after collecting members from `list_group_members`. If the admin has a direct `GroupMember` row (which is common — admins are typically added via `add_group_member`), they will be in both the `list_group_members` result and the `candidates` list. The `seen` dedup guard handles this correctly (they'd be in `seen` from the initial population), but it means the admin is pushed into `candidates` only to be immediately skipped. This is a minor inefficiency but not a correctness issue.

**Suggested fix:**
```
Check `if !seen.contains(&meta.admin_identity)` before pushing to candidates, or simply rely on the existing `seen` guard (current behavior is correct, just slightly wasteful).
```

apps/scaffolding-e2e/workflows/group-join-via-inheritance.yml
```diff
@@ -138,6 +145,31 @@ steps:
...
       - 'json_equal({{inherit_was_inherited}}, true)'
 
+  # ── Phase 2b: list_group_members reflects the inherited joiner ────
+  # Issue #2371 regression. node-2 just joined `subgroup` via
+  # inheritance and has NOT yet called `join_context` — so it holds no
+  # explicit `GroupMember` row anywhere in the subgroup. Listing the
+  # subgroup members on node-2's own node must still include node-2:
+  # the admin-api handler unions the stored rows with
+  # `enumerate_inherited_members`. Before the fix this assertion fails
+  # — node-2's identity appeared only in `selfIdentity`, never in
```
frdomovic, line 155: 💡 **E2E assertion uses `contains` on member list but does not assert list length or absence of duplicates**

The assertion `contains({{subgroup_members}}, {{node1_key}})` and `contains({{subgroup_members}}, {{inherit_member_pk}})` verify presence but do not guard against the inherited member appearing multiple times (once as a stored row and once as an inherited row). The PR description states the sets are disjoint by construction, but the e2e test does not verify this invariant. If a future change accidentally writes a `GroupMember` row for inherited joiners, the test would still pass while the member appears twice.

**Suggested fix:**
```
Add an assertion that the member list has no duplicates, or assert the exact expected length of `subgroup_members` (e.g., 2: node-1 as explicit admin, node-2 as inherited member).
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 24% | Review time: 76.5s

🟡 2 warnings, 💡 2 suggestions. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-9573075c`</sub>

rtb-12:
## Addressing the automated review feedback

The 17 inline comments collapse into 3 distinct themes (the rest are duplicate flags across review agents). Resolved as follows — fixes in `78599ea5`.

### ✅ Fixed — O(n²) `Vec::contains` for `seen` dedup
`enumerate_inherited_members` now uses a `BTreeSet<PublicKey>` for `seen` instead of a `Vec` with linear `contains`. `BTreeSet` (not `HashSet`) because `PublicKey` derives `Ord` but not `Hash`. The per-candidate check is now `seen.insert(candidate)` returning `false` to skip — O(log n).

### ⚖️ By design — handler materializes the union before paginating
Kept as-is, with the comment made explicit about the cost. The inherited set is **computed**, not stored, so it cannot be store-paginated; correctly paginating the *union* therefore requires both full sets in hand. Group membership here is governance-scale (a permission hierarchy bounded by `MAX_NAMESPACE_DEPTH`), not an unbounded user list, so O(total effective members) per call is acceptable. A two-phase "serve from store when the page is entirely within stored rows" optimization is possible but is premature complexity for the actual scale — happy to revisit if a concrete large-group case appears. The handler comment now states the O(total) bound as a known cost rather than implying it is free.

### ⚖️ By design — per-candidate `check_group_membership_path`
The per-candidate re-walk is deliberate: reusing the single-identity check verbatim guarantees `enumerate_inherited_members` can never diverge from `check_group_membership_path` on the anchor-cap / admin-override rules — the exact subtlety that subsystem has historically gotten wrong. Bounded by `MAX_NAMESPACE_DEPTH` (small), so the "cubic" worst case is `N × D²` with D ≤ a handful. Correctness over a micro-optimization that would re-derive the same rules.

The `meta.admin_identity` candidate push is unconditional because the `BTreeSet` dedup handles the (common) case where the admin also has a direct row; the push is one `Copy` of `[u8; 32]`. The comment covers the namespace-creator case where there genuinely is no direct row.

### e2e duplicate-check note
The comment targeting the `node1_key` assertion is moot — that assertion was removed in the CI fix (`create_namespace` never exported `memberPublicKey`; pre-existing phantom variable). Disjointness of the inherited set from stored rows is guarded at the unit level: `enumerate_inherited_members_includes_open_subgroup_joiner` asserts the direct member is **not** reported as inherited.

rtb-12:
## Second review round — addressed (`3ae97854`)

### ✅ Fixed — Cursor: missing cycle-detection bail (Medium)
Good catch. `enumerate_inherited_members` now tracks a `terminated` flag and `bail!`s with a "possible cycle in store" error if the `0..=MAX_NAMESPACE_DEPTH` loop exhausts without hitting a `Restricted` wall or the namespace root — matching `check_group_membership_path` / `is_inherited_admin`. Added a regression test (`enumerate_inherited_members_bails_on_depth_overflow`) mirroring the existing `is_open_chain_to_namespace_bails_on_depth_overflow` pattern.

### ✅ Fixed — misleading comment ("a `HashSet` keeps lookup O(1)")
The comment was stale after the `Vec`→`BTreeSet` switch and is the root of the "use HashSet" flags. Corrected: `BTreeSet` is used **because `PublicKey` derives `Ord` but not `Hash`** — `HashSet<PublicKey>` does not compile. O(log n) lookups, ample for governance-scale membership. The earlier round's flag explicitly offered "`HashSet` (or `BTreeSet`)"; `BTreeSet` is the only one of the two that the type permits.

### ⚖️ By design — O(N·D²) walk cost (re-flag)
Unchanged, as in the first round. The per-candidate `check_group_membership_path` reuse is deliberate — it guarantees enumeration never diverges from the single-identity check on the anchor-cap / admin-override rules. `D` is bounded by `MAX_NAMESPACE_DEPTH` (16) and group membership is a bounded permission hierarchy, so the worst case is small and constant-bounded in practice.

### 💡 By design — meta-admin duplicate candidate / `seen` seeding
The `BTreeSet` dedup is the single, structural guard against double-reporting; pushing `meta.admin_identity` unconditionally costs one `Copy` of `[u8; 32]`. The "meta admin of `group_id` itself with no direct row" case is pre-existing `list_group_members` behavior (row-less meta admins were never listed) and orthogonal to #2371 (inherited *Open-subgroup* joiners) — out of scope here.

All 207 `group_store` unit tests pass; the e2e was verified from node logs — the joiner appears in `members[]` on its own node.

rtb-12:
## Third review round — addressed (`e600730b`)

### ✅ Fixed — Cursor: `member_count` inconsistent with expanded member list (Medium)
Valid — `get_group_info`'s `member_count` came from `count_group_members` (stored rows only), so after this PR it under-reported relative to `list_group_members`. The `get_group_info` handler now reports `count_group_members + enumerate_inherited_members().len()`, consistent with the `list_group_members` handler.

`sync_group`'s `member_count` is intentionally left as the stored-row count: it reports replication state, and inherited membership is *derived*, not synced — conflating them there would be misleading in the other direction.

### ✅ Refuted with a test — meroreviewer: "different loop bound / off-by-one at boundary depth"
The claim is incorrect: `enumerate_inherited_members` and `check_group_membership_path` have **identical** walk structure — both check `get_subgroup_visibility(current)`, then `get_parent_group(current)`, then advance `current = parent`, under the same `0..=MAX_NAMESPACE_DEPTH` bound — so they visit the exact same ancestor set. Rather than just assert this, I added `enumerate_inherited_members_resolves_at_max_namespace_depth_boundary`: a fully-Open chain of exactly `MAX_NAMESPACE_DEPTH` edges with the sole member at the *deepest* ancestor. If the walk fell short, that member would be missed — the test confirms it is enumerated, agreeing with `check_group_membership`.

### ✅ Fixed — meroreviewer: depth-overflow test ID collision
The overflow test now encodes the level index across two bytes (`(i as u16).to_le_bytes()`) instead of a single wrapping byte, so distinct levels can never alias into a real cycle regardless of `MAX_NAMESPACE_DEPTH`.

All 208 `group_store` unit tests pass (`enumerate_inherited_members` now has 5: joiner, inherited-admin role, Restricted wall, depth-overflow bail, boundary depth).

rtb-12:
## Fourth review round — addressed (`725215c4`)

Both unresolved comments target the same line and **contradict the previous round**, so flagging that directly:

- Round 3, `cursor` (Medium): *"`member_count` must include inherited members or it is inconsistent with `list_group_members`."* → I made the count include them.
- Round 4, `cursor` (Medium): *"computing inherited members for `member_count` is expensive."*

These cannot both be satisfied. An accurate effective member count **is** the inherited-membership computation — inherited membership is derived state with no stored counter, so there is no cheap `count_*` scan for it. Keeping correctness (the round-3 demand): the count stays accurate.

### `meroreviewer` — "double full enumeration per request"
Factually not the case. `get_group_info` (`GET …/info`) and `list_group_members` (`GET …/members`) are **separate endpoints**; `get_group_info` does not call the members handler. A single `/info` request calls `enumerate_inherited_members` exactly once. There is no per-request double walk to coordinate away.

### `cursor` — "recomputes full inherited set just for counting" (Medium)
The `.len()`-only use is real, but a count-only fast path would save nothing meaningful: the cost cursor describes (the chain walk + per-candidate `check_group_membership_path`) is exactly what counting requires — only the small bounded `Vec` allocation would be avoided. The walk is bounded by `MAX_NAMESPACE_DEPTH` (16), and `get_group_info` **already** calls `compute_group_state_hash`, which scans the entire group state on every request — the inherited-members walk is minor beside it. The "called on every page load → notable regression" framing does not hold given that pre-existing scan.

Added a comment at the call site documenting the cost and why it is irreducible (meroreviewer's explicit minimum ask). No logic change.

`cargo build -p calimero-context` clean; 208 `group_store` unit tests pass.

rtb-12:
## Fifth review round — re-flags of already-resolved themes

This round (7 comments) is almost entirely a re-review of the same lines, raising themes already addressed and resolved in earlier rounds:

- **O(N²) / nested store scans** — re-flag (rounds 1, 2, 4). By design: per-candidate `check_group_membership_path` reuse guarantees enumeration never diverges from the single-identity rules; bounded by `MAX_NAMESPACE_DEPTH`.
- **Handler loads full set before paginating** (`cursor`, Medium) — re-flag (rounds 1, 4). By design: the inherited set is computed, not stored, so the union cannot be store-paginated.
- **Loop bound off-by-one vs `check_group_membership_path`** — re-flag (round 3); the comment itself now concedes "uses the same bound, which should be correct". Refuted by test `enumerate_inherited_members_resolves_at_max_namespace_depth_boundary`.
- **`seen` seeded via `list_group_members(0, usize::MAX)`** — `usize::MAX` as "unlimited" is the established convention across all ~8 call sites of this function; not new behavior.

Genuinely new, both 💡 (non-blocking):
- `candidates` Vec could be an iterator chain — micro-allocation, bounded loop.
- The boundary test added last round uses single-byte IDs while the overflow test uses two-byte. Fair internal inconsistency, but harmless (`MAX_NAMESPACE_DEPTH` is 16; aliasing needs ≥160).

No code change this round — the substantive items are deliberate design decisions documented in code comments and prior replies. Resolving these threads accordingly.

PR #2379, opened 2026-05-17
fix(context): include inherited members in get_member_capabilities (#2378)

### LINKED ISSUE #2378: get_member_capabilities omits inherited members — sibling of #2371, blocks the same app-side membership checks
## Summary

Direct follow-up to #2371 / #2372. #2372 made `list_group_members` return `stored rows ∪ enumerate_inherited_members`, so an Open-subgroup joiner who joined via inheritance now correctly appears in `members[]`. But the sibling endpoint `GET /admin-api/groups/<id>/members/<member>/capabilities` was **not** updated and still rejects inherited members.

After `POST /admin-api/groups/<id>/join-via-inheritance` returns `200 / wasInherited: true`, calling `get_member_capabilities` for that same joiner on the same node fails with:

```
identity is not a member of group '<id>'
```

even though `list_group_members` (post-#2372) and `check_group_membership` both now agree the joiner *is* an effective member.

## Root cause (confirmed against `master` @ 6a963ef8)

`crates/context/src/handlers/get_member_capabilities.rs:21-23`:

```rust
if group_store::get_group_member_role(&self.datastore, &group_id, &member)?.is_none() {
    bail!("identity is not a member of group '{group_id:?}'");
}
```

`get_group_member_role` (`crates/context/src/group_store/membership.rs:112`) delegates straight to `get_direct_member_role`, which reads only the explicit `GROUP_MEMBER_PREFIX` row:

```rust
pub fn get_group_member_role(store, group_id, identity) -> EyreResult<Option<GroupMemberRole>> {
    get_direct_member_role(store, group_id, identity)
}
```

An inherited member has no `GroupMember` row by design (`execute_member_joined_open` is validate-only — see #2371), so this returns `None` and the handler bails. #2372 touched `membership.rs`, `mod.rs`, `get_group_info.rs`, `list_group_members.rs` — it never touched `get_member_capabilities.rs` (last change there is `e76a2e4a`, long pre-#2372).

## Impact

The two membership endpoints are now inconsistent: `list_group_members` reports the inherited joiner, `get_member_capabilities` denies them. Any app that resolves effective capabilities by calling both — the natural pattern, since `list_group_members` gives role and `get_member_capabilities` gives the bitmask — gets a contradiction.

Concretely, in mero-drive's `useMemberCaps`:

1. `listGroupMembers(folderId)` → inherited joiner found, role `Member` ✓ (thanks to #2372)
2. role ≠ `Admin` → `getMemberCapabilities(folderId, joiner)` → **bails** ✗
3. 4 retries all fail → hook resolves to `isMember = false`
4. The "join open folder" card never unmounts → every two-node Playwright test that clicks Join then reads the document times out (~10 tests).

An inherited **admin** is unaffected — callers typically short-circuit on `role == 'Admin'` from the list and never call `get_member_capabilities`. It's specifically inherited **non-admin** members that hit the wall.

## Reproduction

Extend the `group-join-via-inheritance.yml` e2e (the one #2372 added a `list_group_members` assertion to): after `join_subgroup_inheritance` and before any `join_context`, call `get_member_capabilities` for the joiner on their own node. Today it errors; it should return a capabilities bitmask (0 for a plain inherited member).

## Suggested fix

Mirror #2372's approach in `get_member_capabilities`: when the direct-row lookup misses, fall back to the inheritance check (`check_group_membership` / the same path `enumerate_inherited_members` uses). If the caller is an effective (inherited) member, return `capabilities = 0` — inherited members hold no explicit per-member bitmask; their access derives from the Open-subgroup chain, and `0` correctly means "member, no extra delegated bits."

Optionally also confirm `get_group_member_role` callers: `get_member_capabilities` wants effective membership, but other callers of `get_group_member_role` (key distribution, authz) likely want direct-only — so add an effective-aware variant rather than changing `get_group_member_role` itself, exactly as #2372 kept `group_store::list_group_members` direct-only and did the union in the handler.

## Relationship to #2371

This is the same class of bug, same root cause (inherited membership is derived state with no stored row), same fix shape. Arguably it should have been part of #2372 — that PR fixed one of the two membership-read endpoints; this is the other one.

rtb-12:
## Summary

Fixes #2378 — the direct follow-up to #2371/#2372.

`GET /admin-api/groups/<id>/members/<member>/capabilities` rejected an inherited Open-subgroup member with `identity is not a member of group '<id>'`, even though `check_group_membership` and (post-#2372) `list_group_members` both report that joiner as an effective member. The two membership endpoints contradicted each other: `list_group_members` listed the inherited joiner, `get_member_capabilities` denied them.

**Root cause** (confirmed against `master` @ `6a963ef8`): the handler gated on `get_group_member_role`, which delegates to `get_direct_member_role` — a direct-`GROUP_MEMBER_PREFIX`-row lookup only. An inherited member has no stored `GroupMember` row by design (`execute_member_joined_open` is validate-only, #2371), so the gate returned `None` and the handler bailed. #2372 fixed the sibling `list_group_members` endpoint but never touched `get_member_capabilities.rs`.

The fix mirrors #2372's shape exactly: the effective-membership check lives in a handler-facing `group_store` helper, while `get_group_member_role` stays strictly direct-only for its authz / key-distribution callers.

## The PR is structured in two commits

This PR **first reproduces the bug, then fixes it** — each commit is self-contained:

1. **`test(context): reproduce #2378`** — extracts the handler's membership gate + capability lookup verbatim into `group_store::get_effective_member_capabilities` (a behavior-preserving refactor, still direct-row-only), adds regression tests, and extends the e2e. Against this commit the two inherited-member tests **fail** and the new e2e step **errors** — the bug, reproduced.
2. **`fix(context): include inherited members …`** — switches the gate to `check_group_membership` (direct ∪ inherited). Tests and e2e go **green**.

## Changes

**Fix** — `crates/context/src/group_store/membership.rs`, `handlers/get_member_capabilities.rs`, `group_store/mod.rs`:
- New `get_effective_member_capabilities(store, group_id, identity) -> Option<u32>`: membership decided by `check_group_membership` (direct row ∪ inherited Open-subgroup membership); `None` → handler rejects, `Some(bits)` → handler returns `bits`.
- An inherited member holds no explicit per-member capability row in the subgroup, so their effective bitmask is `0` — "member, no extra delegated bits"; their access derives from the Open-subgroup chain. `get_member_capability` already returns `0` for them.

## Tests — fail without the fix, pass with it

**Unit** (`crates/context/src/group_store/tests.rs`, 5 new):
- `…_includes_inherited_open_subgroup_joiner` — inherited non-admin joiner → `Some(0)` *(fails commit 1, passes commit 2)*
- `…_reports_inherited_admin` — inherited admin → `Some(0)` *(fails commit 1, passes commit 2)*
- `…_returns_stored_bits_for_direct_member` — direct member's stored bitmask flows through unchanged *(guard)*
- `…_none_for_non_member` — true non-member → `None` *(guard)*
- `…_none_for_restricted_wall` — parent member of a `Restricted` subgroup → `None`; the fix must not over-reach past the inheritance boundary *(guard)*

**merobox e2e** (`apps/scaffolding-e2e/workflows/group-join-via-inheritance.yml`, Phase 2c): after `join_subgroup_inheritance`, node-2 calls `get_member_capabilities` for itself and asserts the result is `0`. The call errors against pre-#2378 `merod`; it succeeds on this branch. (`get_member_capabilities` is an existing merobox step — no merobox-repo change needed.)

## Verification

- `cargo test -p calimero-context` — 278 lib + 23 integration tests pass
- Verified RED → GREEN: commit 1 → 3 passed / 2 failed; commit 2 → 5 passed / 0 failed
- `cargo fmt --check` clean; no new clippy warnings
- `merobox bootstrap validate` — workflow configuration valid

---

> [!NOTE]
> **Medium Risk**
> Changes the membership gate for `GET /admin-api/groups/:id/members/:member/capabilities` to treat inherited Open-subgroup members as valid members, which could affect authorization semantics if the effective-membership/deny-list logic is wrong. Coverage is improved with unit tests and an e2e workflow assertion to reduce regression risk.
> 
> **Overview**
> Fixes `get_member_capabilities` so it **accepts effective (inherited) Open-subgroup members** instead of requiring a direct `GroupMember` row, returning `0` when no per-member capability row exists.
> 
> Adds a new `group_store::get_effective_member_capabilities` helper (including deny-list handling for inherited members) and switches the handler to use it, plus new unit tests and an e2e step to prevent `list_group_members` vs `get_member_capabilities` contradictions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39c3e3ab80bcd51e20c1febe8b66c055c68d0ec0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/scaffolding-e2e/workflows/group-join-via-inheritance.yml
+++ b/apps/scaffolding-e2e/workflows/group-join-via-inheritance.yml
@@ -22,6 +22,11 @@ description: >
        before the fix `GET /admin-api/groups/<id>/members` returned
        only the explicit members and apps keying "is the caller a
        member?" off that list saw `false` for any inherited joiner.
+    4. Resolves the inherited joiner through `get_member_capabilities`
+       — the sibling membership endpoint, issue #2378. Before that fix
+       it gated on a direct-row lookup only and bailed "identity is not
+       a member" for the very joiner `list_group_members` reported in
+       point 3, leaving the two endpoints contradictory.
 
   Pre-#2357 the only way to materialize the inherited path was to
   discover a child context id and call `join_context` against it
@@ -172,6 +177,36 @@ steps:
       # (`inherit_member_pk`, exported by the join-via-inheritance step)
       # must appear in the listed members.
       - 'contains({{subgroup_members}}, {{inherit_member_pk}})'
+
+  # ── Phase 2c: get_member_capabilities accepts the inherited joiner ─
+  # Issue #2378 regression. `list_group_members` (Phase 2b) now reports
+  # node-2 as an effective member of `subgroup`; the sibling endpoint
+  # `GET /admin-api/groups/<id>/members/<member>/capabilities` must
+  # agree. node-2 holds no explicit `GroupMember` row, so before the
+  # #2378 fix this call bailed with "identity is not a member of
+  # group ..." — the handler gated on a direct-row lookup only and the
+  # two membership endpoints contradicted each other. After the fix the
+  # call succeeds and returns `0`: an inherited member holds no explicit
+  # per-member bitmask in the subgroup (their access derives from the
+  # Open-subgroup chain), so `0` correctly means "member, no extra
+  # delegated bits."
+
+  - name: 'Node-2 reads its own capabilities (issue #2378)'
+    type: get_member_capabilities
+    node: join-inherit-node-2
+    group_id: '{{subgroup_id}}'
+    member_id: '{{inherit_member_pk}}'
+    outputs:
+      inherit_member_caps: capabilities
+
+  - name: Assert inherited joiner resolves to a 0 capability bitmask
+    # `json_assert` so the literal `0` is parsed as a JSON number rather
+    # than the string `'0'`. The decisive signal is that the step above
+    # succeeds at all — pre-#2378 it errored before returning any value,
+    # failing the workflow outright.
+    type: json_assert
+    statements:
+      - 'json_equal({{inherit_member_caps}}, 0)'
 
   # ── Phase 3: node-2 joins the context (key already delivered) ─────
   # This call should now be cheap because the group key is already
--- a/crates/context/src/group_store/membership.rs
+++ b/crates/context/src/group_store/membership.rs
@@ -313,6 +313,28 @@ pub fn check_group_membership(
     ))
 }
 
+/// Returns the capability bitmask `identity` holds as a member of
+/// `group_id`, or `None` when they are not a member — the membership
+/// gate and capability lookup behind the `get_member_capabilities`
+/// admin-api handler.
+///
+/// `None` tells the handler to reject the request; `Some(bits)` tells it
+/// to return `bits`. A member's bitmask is their stored per-member
+/// capability row (`0` when unset).
+pub fn get_effective_member_capabilities(
+    store: &Store,
+    group_id: &ContextGroupId,
+    identity: &PublicKey,
+) -> EyreResult<Option<u32>> {
+    // Membership gate: only a direct `GroupMember` row counts here.
+    if get_group_member_role(store, group_id, identity)?.is_none() {
+        return Ok(None);
+    }
+    Ok(Some(
+        get_member_capability(store, group_id, identity)?.unwrap_or(0),
+    ))
+}
+
 /// Enumerate the identities that are members of `group_id` purely by
 /// inheritance — i.e. those with a [`MembershipPath::Inherited`] path
 /// and no stored `GroupMember` row in `group_id` itself.
--- a/crates/context/src/group_store/membership.rs
+++ b/crates/context/src/group_store/membership.rs
@@ -335,19 +335,33 @@ pub fn check_group_membership(
 /// direct-only view, so the effective-aware check lives here, at the
 /// handler-facing helper.
 ///
-/// For an inherited member the per-group **deny list** is re-applied —
-/// exactly as [`enumerate_inherited_members`] does (#2371). A member
-/// kicked from an Open subgroup keeps their namespace-level inheritance
-/// but is deny-listed on the subgroup, and that deny-list *is* the
-/// removal (the kick has no direct row to delete). `check_group_membership_path`
-/// deliberately skips the deny list — it is also run by
-/// `MemberJoinedOpen` apply mid-rejoin, where the rejoiner is still
-/// deny-listed — so a kicked inherited member would otherwise resolve to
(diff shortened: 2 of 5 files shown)


frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 30% | Review time: 65.4s

---

### 🟡 Warning (1)

#### 1. get_effective_member_capabilities does not distinguish inherited admin from inherited member
**File:** `crates/context/src/group_store/membership.rs` (line 313-352) | **Consensus:** 1/1 agents ✓

The function returns `Some(get_member_capability(store, group_id, identity)?.unwrap_or(0))` for any effective member, including inherited admins. An inherited admin (via_admin: true) has no stored capability row in the subgroup, so they also receive `Some(0)`. The handler then returns `capabilities: 0` to the caller, which could mislead callers into thinking the inherited admin has no special capabilities. While the PR description acknowledges this ('their access derives from the Open-subgroup chain'), the returned bitmask of 0 for an inherited admin is semantically ambiguous — it looks identical to an inherited non-admin member. Callers relying on this endpoint to determine admin status will get a false negative.

**Suggested fix:**
```
Consider returning a sentinel value or a separate field in the response to indicate inherited admin status, or document clearly in the API response that `0` does not imply non-admin for inherited members. Alternatively, check `check_group_membership_path` for `via_admin: true` and return the admin capability bits or a distinct representation.
```

### 💡 Suggestion (1)

#### 1. Test for inherited admin uses Some(0) but does not assert the admin is actually recognized as admin by other APIs
**File:** `crates/context/src/group_store/tests.rs` (line 7815-7971) | **Consensus:** 1/1 agents ✓

The test `get_effective_member_capabilities_reports_inherited_admin` asserts that an inherited admin resolves to `Some(0)`. However, it does not assert that `is_inherited_admin` or `is_group_admin` returns true for this identity in the subgroup, which would strengthen the test by confirming the admin path is exercised and not just the membership path. Without this, the test does not distinguish between an inherited admin and an inherited non-admin member from the perspective of `get_effective_member_capabilities`.

**Suggested fix:**
```
Add an assertion: `assert!(is_inherited_admin(&store, &reports, &admin).unwrap(), "inherited admin must be recognized as admin");` before the capability assertion.
```

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-f444db81`</sub>

PR #2381, opened 2026-05-18
fix(node/sync): preserve existing storage_type in HashComparison leaf apply

chefsale:
## Summary

HashComparison sync could not close drift on `Shared` entities. The receive path constructed each `Action`'s metadata via `Metadata::default()` (which leaves `storage_type = Public`) and only overrode `crdt_type` / `updated_at` / `created_at`. When the existing local entity was `Shared`, the validator rejected the action with *\"Cannot change StorageType\"*. HashComparison retried the same malformed action forever; the two peers stayed diverged with no recovery.

Fix: on the Update path, copy `storage_type` from the existing entity's `Index` onto the action. CRDT merge changes value, not storage type.

**+24 LOC**, no behavior change for `Public` entities (the existing default-case), real fix for `Shared` / `User` / `Frozen`.

### What was happening

`crates/node/src/sync/helpers.rs::apply_leaf_with_crdt_merge` is the receive-side apply for both HashComparison directions (pull via `apply_leaf_with_crdt_merge` from `hash_comparison_protocol.rs::281`, and push via `handle_entity_push` which loops over leaves and calls the same function).

The function builds metadata like this:

```rust
let mut metadata = Metadata::default();
metadata.crdt_type = Some(leaf.metadata.crdt_type.clone());
metadata.updated_at = leaf.metadata.hlc_timestamp.into();
metadata.created_at = leaf.metadata.created_at;
// storage_type stays whatever Metadata::default() produced — Public.
```

`StorageType::default()` is `Public` (`crates/storage/src/entities.rs:434`).

The validator at `crates/storage/src/interface.rs:1814-1852` lists the four allowed same-type transitions (Public→Public, Frozen→Frozen, User→User, Shared→Shared) and rejects everything else with:

> *\"Cannot change StorageType (e.g., User->Public/User->Frozen/etc)\"*

That rejection is **correct** in isolation — silently downgrading `Shared` to `Public` would let strangers overwrite signed content. The bug is upstream: HashComparison's apply path was constructing the wrong `storage_type` on the action.

### Why the wire format can't help

The wire-level `LeafMetadata` (`crates/node/primitives/src/sync/hash_comparison.rs:319-350`) carries `crdt_type`, `hlc_timestamp`, `created_at`, `version`, `collection_id`, `parent_id`. It does **not** carry `storage_type`. So the receiver can't recover the originator's `storage_type` from the wire.

For the **Update** path this doesn't matter — the receiver already has the entity locally and can read its `storage_type` from the `Index`. CRDT merge changes value, not storage type, so reusing the existing `storage_type` is both correct and what the validator expects.

For the **Add** path the receiver has no local entity to read from. Non-`Public` storage types carry creation-time invariants the sync push path can't satisfy on its own (writer-set + signature for `Shared`, owner identity for `User`, content-addressing for `Frozen`), so those should arrive via signed actions on the delta path. This PR leaves the Add path defaulting to `Public`. Adding `storage_type` to the wire is a wire-format change and a separate decision.

### Recovery-path completion

With this fix, the three-layer recovery story is end-to-end functional for `Shared` storage:

1. **Reconcile-via-anchor** (#2343) — signed-claim divergence on `MemberRemoved`/`MemberLeft` → pull from trusted anchor.
2. **Receiver-side anti-entropy via `ReadinessBeacon`** (#2374) — cold-mesh dropped governance ops → trigger `sync_namespace` on divergence.
3. **HashComparison** — last-resort entity-by-entity Merkle diff + transfer. Now works for `Shared` entities; previously only worked for `Public` and (after #2344) opaque leaves.

## Test plan

- [x] `cargo test -p calimero-node --lib` — 148 passed
- [x] `cargo test -p calimero-node --tests` — 306 passed
- [ ] CI green on push

## Not in this PR (follow-up)

- **Sim regression test for `Shared`-entity HashComparison.** The existing sim infra at `crates/node/tests/sync_sim/` constructs entities with `Metadata::default()` (Public). Exercising the fix end-to-end in the sim requires Shared-entity setup (writer-set + signed actions), which doesn't exist in the sim today. Building it out is a separate PR; the fix itself is small and clearly motivated.
- **Wire-format `storage_type` for Add path.** Bootstrap of `Shared` / `User` entities via sync push currently silently defaults to `Public`. Real-world `Shared` entities always arrive via signed actions on the delta path, so this is not a known production gap; left for a future PR if/when sync-side bootstrap of non-`Public` storage becomes a use case.

---

> [!NOTE]
> **Medium Risk**
> Adjusts receive-side HashComparison apply logic in the storage sync path; while small, it affects how replicated entities are validated/applied and could impact convergence or metadata invariants if incorrect.
> 
> **Overview**
> Fixes HashComparison’s receive-side leaf apply so **Update** actions preserve the existing entity’s `storage_type` instead of implicitly defaulting to `Public`, preventing validators from rejecting updates to non-`Public` (e.g., `Shared`) entities and allowing drift to converge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 023b018cff6787faeb9e2859a1f47963f02629de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/helpers.rs
+++ b/crates/node/src/sync/helpers.rs
@@ -79,30 +79,6 @@ pub fn apply_leaf_with_crdt_merge(context_id: ContextId, leaf: &TreeLeafData) ->
     metadata.updated_at = leaf.metadata.hlc_timestamp.into();
     metadata.created_at = leaf.metadata.created_at;
 
-    // Preserve the existing entity's `storage_type` on Update. The wire
-    // format (`LeafMetadata`) does not carry `storage_type` — historically
-    // the receiver constructed actions with `Metadata::default()`, which
-    // returns `StorageType::Public`. When the existing entity is `Shared`,
-    // `User`, or `Frozen`, the storage validator rejects the action with
-    // "Cannot change StorageType (e.g., User->Public/User->Frozen/etc)"
-    // and HashComparison retries the same malformed action forever,
-    // never converging. CRDT merge changes value, not storage type, so
-    // copying the existing entity's `storage_type` onto the action is
-    // both correct and required to make the validator accept it.
-    //
-    // For new entities (Add path below), the wire format gives us no
-    // way to recover the originator's `storage_type`; we keep
-    // `Public` as the default. Non-`Public` entities (`Shared` /
-    // `User` / `Frozen`) carry creation-time invariants the sync
-    // leaf-push path can't satisfy on its own (writer-set signature
-    // for `Shared`, owner identity for `User`, content-addressing for
-    // `Frozen`), so they should arrive via signed actions on the delta
-    // path. Sync leaf push is for closing CRDT value drift on entities
-    // both peers already have.
-    if let Some(ref existing) = existing_index {
-        metadata.storage_type = existing.metadata.storage_type.clone();
-    }
-
     let action = if existing_index.is_some() {
         // Update existing entity - storage layer handles CRDT merge
         Action::Update {
--- a/crates/node/src/sync/helpers.rs
+++ b/crates/node/src/sync/helpers.rs
@@ -79,6 +79,30 @@ pub fn apply_leaf_with_crdt_merge(context_id: ContextId, leaf: &TreeLeafData) ->
     metadata.updated_at = leaf.metadata.hlc_timestamp.into();
     metadata.created_at = leaf.metadata.created_at;
 
+    // Preserve the existing entity's `storage_type` on Update. The wire
+    // format (`LeafMetadata`) does not carry `storage_type` — historically
+    // the receiver constructed actions with `Metadata::default()`, which
+    // returns `StorageType::Public`. When the existing entity is `Shared`,
+    // `User`, or `Frozen`, the storage validator rejects the action with
+    // "Cannot change StorageType (e.g., User->Public/User->Frozen/etc)"
+    // and HashComparison retries the same malformed action forever,
+    // never converging. CRDT merge changes value, not storage type, so
+    // copying the existing entity's `storage_type` onto the action is
+    // both correct and required to make the validator accept it.
+    //
+    // For new entities (Add path below), the wire format gives us no
+    // way to recover the originator's `storage_type`; we keep
+    // `Public` as the default. Non-`Public` entities (`Shared` /
+    // `User` / `Frozen`) carry creation-time invariants the sync
+    // leaf-push path can't satisfy on its own (writer-set signature
+    // for `Shared`, owner identity for `User`, content-addressing for
+    // `Frozen`), so they should arrive via signed actions on the delta
+    // path. Sync leaf push is for closing CRDT value drift on entities
+    // both peers already have.
+    if let Some(ref existing) = existing_index {
+        metadata.storage_type = existing.metadata.storage_type.clone();
+    }
+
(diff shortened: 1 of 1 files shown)


chefsale:
Cursor's finding is correct and going deeper revealed the actual fix is much bigger than this PR's scope. I've reverted the change.

## Why my fix didn't work

Cursor flagged that cloning `existing.metadata.storage_type` carries over the stored `signature_data` (old nonce + old signature) into the new sync action. Tracing through the Shared/User apply paths:

- **NonceReplay**: `sig_data.nonce` (old) compared against stored `updated_at` triggers `StorageError::NonceReplay`.
- **InvalidSignature** (even bypassing nonce): the old signature is over the old payload; the new action has new data and new updated_at → signature doesn't verify.

So my fix only changed the failure mode from `"Cannot change StorageType"` to `NonceReplay` / `InvalidSignature`. Same outcome: divergence unrepaired.

## Why the obvious bigger fix also doesn't work

The next thing to try is extending the wire format to carry `signature_data` (so the receiver doesn't have to fabricate it). I started down that path and hit a deeper issue:

`action.payload_for_signing()` includes the action's `ancestors` — and each ancestor is `(id, merkle_hash)`. The merkle hash is **tree-state dependent**: it changes as state mutates. So the originator's signature is over *the originator's tree state at sign time*. The receiver computing the same payload uses *its own tree state at receive time*. Even with the same ancestor IDs, the merkle hashes differ → payload hash differs → signature fails to verify.

This isn't fixable by adding bytes to the wire. The signature is fundamentally bound to tree state at sign time, and sync apply happens against a different tree state.

## What the real fix would need

One of these design changes (each significant):

1. **A sync-specific signing model**: signature over `(entity_id, data, nonce)` only — independent of tree state. Sender re-signs at sync time, or originator pre-computes a tree-state-independent signature at apply time and stores it alongside the tree-state-bound one.
2. **A trusted-sync ApplyContext mode**: skip signature verification when the sync peer's libp2p identity is in the entity's writer-set (still requires plumbing peer-id → writer-pubkey).
3. **Skip HashComparison for non-Public entities** and rely on the state-delta path for Shared/User reconciliation. Means Shared/User divergence is only repaired by the next legitimate delta or via the reconcile-via-anchor path.

All three need design discussion. The first two relax or rework the security model; the third is a documented limitation.

## Recommendation

Close this PR. Open a follow-up issue to capture the design discussion: \"HashComparison cannot reconcile Shared/User entities\". The current production behavior (peers stay diverged on Shared/User HashComparison alone) is unchanged — the receive-side delta path remains the primary reconciliation mechanism, and the reconcile-via-anchor path (#2343) and receiver-side anti-entropy (#2374) handle the cold-mesh case.

The opaque-leaf fix from #2344 already covered the most common HashComparison-divergence case (legacy `crdt_type: None` entities, including the WASM app root state). The non-Public case is rarer in practice but worth fixing properly when the team has design bandwidth.

Sorry for the noise — I should have caught the signature/ancestors issue before opening the PR.

chefsale:
Closing per the analysis above. The fix as written doesn't work and the proper fix needs design discussion (tree-state-independent signing for sync apply, or sync-trusted ApplyContext mode, or scoping HashComparison to Public-only).

PR #2382, opened 2026-05-18
feat(context): B3 — local-execute membership check for state ops

chefsale:
## Summary

Close the local-execute side of the cross-DAG authorization check. The receive-path enforcement was already in place (peers reject deltas from non-members at apply time); this adds the companion check so a removed member's WASM cannot mutate local state in the first place.

**Net +233 LOC** (most of it tests). One new helper + one new call site mirroring the existing ReadOnly discard pattern.

## The gap this closes

The cross-DAG authorization story has two sides:

| Side | Question | Status before this PR |
|---|---|---|
| **Receive path** | Is the delta's signer a current member at the delta's signed cut? | ✅ Enforced by `membership_status_at` |
| **Local execute** | Is the local executor a current member of the context's group? | ❌ Not enforced |

A removed member's WASM could still call mutation methods on a context they no longer belong to. The mutations:
- Never propagate (every peer rejects the resulting delta via the receive-path check)
- BUT accumulate as silent local-only divergence from the canonical view

The principle of least surprise — \"a removed member's WASM can't write to anything in this group anymore\" — was documented in the cross-DAG RFC's \"as-shipped notes\" as a follow-up. This PR finishes that.

### `crates/context/src/group_store/namespace.rs` — new helper

```rust
pub fn is_authorized_for_context_state_op(
    store: &Store,
    context_id: &ContextId,
    executor: &PublicKey,
) -> EyreResult<bool>
```

Returns `true` for:
- Current Admin or Member role in the context's owning group
- Namespace creator (via `is_group_admin` carve-out — `admin_identity` membership without a `GroupMember` row)
- Contexts not owned by any group (no membership concept to enforce)

Returns `false` for:
- ReadOnly / ReadOnlyTee — covered by existing `is_read_only_for_context` too, but explicit here
- Removed members — `GroupMember` row deleted at removal time
- Never-members — never had a row

### `crates/context/src/handlers/execute/mod.rs` — new discard branch

Mirrors the existing ReadOnly discard at line ~1275. Runs after WASM execution; if the executor isn't a current authorized member and the outcome carries state changes (`root_hash.is_some()`), the state effects (`root_hash`, `artifact`, `xcalls`) are cleared and an info log records the rejection. Read-only methods are unaffected.

Post-WASM discard (matching the existing pattern) rather than pre-WASM rejection: keeps the user-visible behavior identical to the ReadOnly case and lets read-side execution complete normally even when state effects are dropped.

## Live-state vs. forward-only

The receive path's `membership_status_at` is **forward-only at the delta's signed cut**: pre-removal writes stay valid forever, regardless of when the receiver observes the removal. That's correct for replicated authoring — two peers that observe ops in different orders must converge.

The local check is **live-state**: at execute time there's no signed cut to evaluate against; the WASM call happens *now*. The two checks complement rather than duplicate:

- **Receive path** governs whether a remote delta is honored on this peer.
- **Local execute** governs whether the local WASM may produce a delta in the first place.

A removed member's pre-removal writes that were already in flight before the removal still apply (forward-only, receive-side). Their new writes after the removal are stopped at the local-execute path before they ever become a delta.

## Test plan

- [x] `cargo test -p calimero-context --lib` — 282 passed (4 new)
- [x] `cargo test -p calimero-context --test local_group_governance_convergence` — 23 passed
- [ ] CI green on push

New unit tests:
1. `authorized_for_state_op_admits_admin_and_member_only` — Admin/Member yes; ReadOnly/ReadOnlyTee/outsider no.
2. `authorized_for_state_op_rejects_removed_member` — author authorized before removal; rejected after.
3. `authorized_for_state_op_recognises_namespace_creator` — admin-identity carve-out works (no `GroupMember` row).
4. `authorized_for_state_op_allows_non_group_context` — contexts without an owning group bypass the check.

## Not in this PR

- **Storage-layer enforcement.** `apply_action` in `crates/storage/src/interface.rs` could also gate User-storage updates on membership, which would catch any code path that bypasses the execute handler. Doing it there is a layering violation (storage doesn't depend on context/group_store today); the right boundary is at the call site, which is what this PR does. If a future caller of `apply_action` opens a new path that bypasses `internal_execute`, that path needs its own check — the comment at the call site documents this contract.
- **WASM-side error path.** Currently the rejection is silent to the WASM module (state cleared post-execution, like ReadOnly). A future PR could surface an `ExecuteError::NotMember` to the caller; that's a UX call, not a security one.

---

> [!NOTE]
> **High Risk**
> Adds a new authorization gate that discards local WASM state mutations when the executor is no longer a permitted group member, which can change write behavior and prevent previously-allowed mutations if the membership logic is wrong. Touches state-execution and authorization paths, so mistakes could cause unexpected write drops or divergence issues.
> 
> **Overview**
> Closes a local-execute gap by adding `is_authorized_for_context_state_op` to decide whether an identity may author **state mutations** for a group-owned context (Admin/Member, namespace creator, and inherited members via Open subgroups allowed; read-only/outsiders/removed members rejected; non-group contexts bypass).
> 
> `internal_execute` now uses this check for `is_state_op` calls and, after WASM execution, discards state effects (`root_hash`, `artifact`, `xcalls`) and logs when a non-authorized executor attempted a mutation, mirroring the existing ReadOnly discard behavior.
> 
> Adds focused unit tests covering allowed roles, removed members, namespace-creator carve-out, non-group contexts, and Open-subgroup inherited membership.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7567ed846da5ee4901e2bd992971d07ada8ae443. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/group_store/namespace.rs
+++ b/crates/context/src/group_store/namespace.rs
@@ -52,6 +52,67 @@ pub fn is_read_only_for_context(
     }
 }
 
+/// Returns `true` if `executor` is currently authorized to author state
+/// mutations on `context_id` — i.e., is a current Admin or Member of the
+/// context's owning group.
+///
+/// Returns `false` for:
+/// * Read-only roles (`ReadOnly`, `ReadOnlyTee`) — they can read but not write.
+/// * Members that have been removed from the group.
+/// * Identities that were never a member of the group.
+///
+/// Returns `true` for contexts that aren't owned by any group — there's no
+/// group membership concept to enforce.
+///
+/// **Live-state check, not historical.** The receive path's cross-DAG check
+/// (`membership_status_at`) is forward-only at the delta's *signed* governance
+/// cut, which is the right semantic for replicated authoring: a pre-removal
+/// delta stays valid forever. This function is the local-execute mirror, and
+/// uses **current** membership because at execute time there is no signed
+/// cut — the WASM call happens "now," not against any historical snapshot.
+/// The two checks complement each other: the receive-path check decides
+/// whether a delta from a peer is honored; this one decides whether the
+/// local WASM may produce a delta in the first place. Without this check, a
+/// removed member's WASM could still mutate local state (including User-
+/// storage entries) — the mutation would never propagate (peers reject via
+/// the receive-path check), but would accumulate as silent divergence from
+/// the canonical view until the next sync round repaired it.
+///
+/// The namespace creator / root admin is recognised via
+/// [`super::membership::is_group_admin`] (their membership lives in
+/// `GroupMeta::admin_identity`, not in a `GroupMember` row).
+pub fn is_authorized_for_context_state_op(
+    store: &Store,
+    context_id: &ContextId,
+    executor: &PublicKey,
+) -> EyreResult<bool> {
+    let Some(group_id) = get_group_for_context(store, context_id)? else {
+        // Non-group context — no membership concept to enforce.
+        return Ok(true);
+    };
+
+    // Namespace creator / root admin carve-out — they don't emit a self-
+    // `MemberJoined` op at genesis, so their membership is stored in
+    // `GroupMeta::admin_identity` and `get_group_member_role` returns
+    // `None`. Mirror the same carve-out `membership_status_at` and
+    // `namespace_member_pubkeys` apply, so admin authority is consistent
+    // across the receive-side and local-execute-side authorization
+    // checks.
+    if super::membership::is_group_admin(store, &group_id, executor)? {
+        return Ok(true);
+    }
+
+    match get_group_member_role(store, &group_id, executor)? {
+        Some(
+            calimero_primitives::context::GroupMemberRole::Admin
+            | calimero_primitives::context::GroupMemberRole::Member,
+        ) => Ok(true),
+        // ReadOnly / ReadOnlyTee → readable but not writeable. Removed /
+        // never-member → no authorization at all. Both fall through here.
+        Some(_) | None => Ok(false),
+    }
+}
+
 /// Read the parent group for a given group (legacy, used by `resolve_namespace`).
 pub fn get_parent_group(
     store: &Store,
--- a/crates/context/src/group_store/namespace.rs
+++ b/crates/context/src/group_store/namespace.rs
@@ -54,12 +54,15 @@ pub fn is_read_only_for_context(
 
 /// Returns `true` if `executor` is currently authorized to author state
 /// mutations on `context_id` — i.e., is a current Admin or Member of the
-/// context's owning group.
+/// context's owning group, either by direct row or by Open-subgroup
+/// inheritance.
 ///
 /// Returns `false` for:
-/// * Read-only roles (`ReadOnly`, `ReadOnlyTee`) — they can read but not write.
+/// * Direct members with read-only roles (`ReadOnly`, `ReadOnlyTee`) — they
+///   can read but not write.
 /// * Members that have been removed from the group.
-/// * Identities that were never a member of the group.
+/// * Identities that were never a member of the group and don't reach it
+///   via Open-subgroup inheritance.
 ///
 /// Returns `true` for contexts that aren't owned by any group — there's no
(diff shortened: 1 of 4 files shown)


crates/context/src/handlers/execute/mod.rs
```diff
@@ -1216,6 +1216,36 @@ async fn internal_execute(
...
+    // Live-state check vs. the receive path's forward-only check: at
+    // execute time there is no signed governance cut to evaluate
+    // membership against, so this consults current membership. The two
+    // checks complement each other rather than duplicate — receive-path
+    // governs whether a remote delta is accepted; this governs whether
+    // local WASM may produce one.
+    let executor_not_authorized_for_state_op = is_state_op
+        && !crate::group_store::is_authorized_for_context_state_op(
+            &datastore,
+            &context.id,
```
frdomovic, line 1255: 🟡 **Fail-open on authorization error: unwrap_or(true) allows unauthorized state ops**

The `is_authorized_for_context_state_op` call uses `.unwrap_or(true)` as the error fallback. If the store read fails (e.g., I/O error, corruption), the authorization check silently passes and the executor is treated as authorized. This is a fail-open pattern — the secure default should be `unwrap_or(false)` to deny on error, matching the principle of least privilege. The existing `is_read_only_for_context` call on line ~1216 correctly uses `.unwrap_or(false)`.

**Suggested fix:**
```
Change `.unwrap_or(true)` to `.unwrap_or(false)` so that a store error causes the authorization check to fail closed (deny) rather than fail open (allow). The resulting `executor_not_authorized_for_state_op` will be `true` when `is_state_op` is true and the store errors, which discards the state op — the safe outcome.
```

crates/context/src/group_store/namespace.rs
```diff
@@ -52,6 +52,67 @@ pub fn is_read_only_for_context(
...
+/// the receive-path check), but would accumulate as silent divergence from
+/// the canonical view until the next sync round repaired it.
+///
+/// The namespace creator / root admin is recognised via
+/// [`super::membership::is_group_admin`] (their membership lives in
+/// `GroupMeta::admin_identity`, not in a `GroupMember` row).
+pub fn is_authorized_for_context_state_op(
+    store: &Store,
+    context_id: &ContextId,
+    executor: &PublicKey,
```
frdomovic, line 110: 💡 **is_authorized_for_context_state_op does not consult the deny-list**

The `is_authorized_for_context_state_op` function checks `is_group_admin` and `get_group_member_role`, but does NOT check `is_denied` (the deny-list). The `MemberRemoved` apply path calls both `remove_group_member` AND `mark_denied`. The test `authorized_for_state_op_rejects_removed_member` calls only `remove_group_member` (not `mark_denied`), so it passes. However, the deny-list is the mechanism used for Open-subgroup kicks where no `GroupMember` row exists to delete. An identity that was kicked from an Open subgroup (deny-listed but never had a direct row) would have `get_group_member_role` return `None`, which already falls through to `Ok(false)` — so the current logic is accidentally correct for that case. But the asymmetry with `is_read_only_for_context` (which also doesn't check the deny-list) and the receive-path `is_author_denied_for_context` suggests the deny-list check is intentionally a separate layer. This is worth documenting explicitly to avoid future confusion.

**Suggested fix:**
```
Add a comment in `is_authorized_for_context_state_op` noting that the deny-list is a separate enforcement layer applied at the receive path (`is_author_denied_for_context`), and that this function only needs to check live membership state. Alternatively, consider adding an explicit `is_denied` check here for defense-in-depth, consistent with the receive-path's layered approach.
```

frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 30% | Review time: 53.5s

🟡 1 warnings, 💡 1 suggestions. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-ea189a6b`</sub>

Issue #2383, opened 2026-05-18
HashComparison cannot reconcile Shared/User entities — needs sync-trusted signing model


chefsale:
## Background

HashComparison sync exists to close value drift between peers on entities both peers already hold. It works for `Public` storage entities and (after #2344) for opaque leaves with `crdt_type: None`. It does **not** work for `Shared`, `User`, or `Frozen` storage:

- **Pre-fix**: receiver constructs Action::Update with `Metadata::default()` → `storage_type = Public`, apply path rejects with *\"Cannot change StorageType\"* on entities stored as Shared/User/Frozen.
- **PR #2381 attempted fix**: clone existing `storage_type` onto the sync action. This fails with `NonceReplay` (cloned-old-nonce ≤ stored `updated_at`) or `InvalidSignature` (cloned-old-signature doesn't verify against new payload).

Investigation in PR #2381 ([comment](https://github.com/calimero-network/core/pull/2381#issuecomment-4476329945)) showed even extending the wire format to carry `signature_data` doesn't work, because the apply-path signature is over `action.payload_for_signing()` which includes ancestor merkle hashes. Those hashes are *tree-state dependent* and differ between sender and receiver. The signature is fundamentally bound to the originator's tree state at sign time.

## The gap

Today, when two peers' tree state diverges on a Shared/User entity (e.g. one peer applied a delta the other didn't), HashComparison detects the divergence (different merkle hashes) but cannot close it. The divergence persists until:

- A subsequent state-delta carrying an update for that entity propagates via the normal receive path (works, but requires fresh activity)
- Reconcile-via-anchor (#2343) fires, which uses the state-delta sync path and has full signature verification (works for divergences detected via signed-claim mismatches on `MemberRemoved` / `MemberLeft`)
- Receiver-side anti-entropy (#2374) catches up missed governance ops (works for cold-mesh op drops, not for arbitrary value divergence)

The gap: long-tail value drift on `Shared`/`User` entities that no other recovery mechanism happens to repair.

## Possible directions

Each of these is a design change that needs discussion:

### 1. Tree-state-independent signing for sync apply

Sign `(entity_id, data, nonce)` only — drop ancestors from the sync-apply payload. The receiver reconstructs the same payload from the wire data + nonce and verifies against the writer-set.

**Pros:** Preserves the security model; no trust assumption on the sync peer.
[...]

PR #2386, opened 2026-05-18
refactor(storage, node/sync): single-signature design — HashComparison reconciles Shared/User entities

chefsale:
## Summary

Closes the gap captured in #2383: HashComparison sync can now reconcile `Shared` / `User` storage entities. Same end state Option 1a from #2383 was after, achieved via the cleaner one-signature design rather than two parallel signatures.

The v1 action signature baked together two concerns — authorization (who wrote what at what nonce) and tree-state integrity (signer's ancestor merkle hashes at sign time). The v2 redesign separates them: the signature covers only authorization (transferable across all receive paths); tree-state integrity becomes an explicit unsigned check at apply time on the delta path.

Net effect: signatures stop committing to facts only delta-replay can verify. HashComparison sync now carries the signature on the wire and verifies it without any tree-state context, while the delta path keeps the same end-to-end correctness via the new explicit ancestor check.

## Background

See #2383 for the original problem framing. Three options were captured there:
1. **Two-signature** (parallel sync-friendly signature, additive)
2. Sync-trusted ApplyContext mode (security relaxation)
3. Scope HashComparison to Public-only (limitation)

This PR implements a fourth option that emerged from design discussion: **redefine the existing signature instead of adding a second one**. Pre-1.0, so the wire-format break is fine. Result is half the storage per entity, one apply path instead of two, and a cleaner conceptual model.

Issue updated with the refined design: https://github.com/calimero-network/core/issues/2383#issuecomment-4478996744

## Audit phase (informing the design)

Before writing code I verified three properties against the real codebase:

1. **`payload_for_signing` is purely signing input**: used at 1 production sign site + 4 production verify sites + 5 test sites. No identity, dedup, or content-hash usage. Safe to redefine what bytes it hashes.
2. **The v1 signed binding catches the same ancestor-mismatch cases the explicit check catches** (and no others) — auto-vivification for missing ancestors was always unverified by the signature too. The signed binding was redundant cryptographic packaging over the apply layer's structural enforcement.
3. **No production code branches on `InvalidSignature`** specifically. Only tests do, and they exercise actual forgery (which still produces `InvalidSignature` under the new payload). Adding a new `TreeStateMismatch` variant is additive.

### `crates/storage/src/action.rs` — redefined signed payload

- `Action::payload_for_signing` now hashes `prefix || id || data || access-control-triple` only.
- Ancestors, `created_at`, and bulk metadata are removed from the signed bytes.
- v2 prefix bytes (`v2_upsert`, `v2_delete`, `v2_compare`) so v1 signatures fail loudly against v2 verifiers rather than silently mis-verifying.
- New `hash_authorization_for_payload` replaces the old `hash_metadata_for_payload`. Single responsibility: produce the deterministic bytes the signature commits to for access control.
- 10 unit tests pin the v2 properties (deterministic, includes content + access control, excludes tree-state fields).

### `crates/storage/src/interface.rs` — explicit ancestor check

- New `Interface::verify_ancestor_integrity` performs the explicit ancestor-merkle-hash check at apply time. Single responsibility: tree-shape integrity. Composable with the existing nonce + signature pipeline.
- Wired into the `Add` / `Update` branch of `apply_action` right after the verify section. HashComparison sync supplies `ancestors: vec![]`, which makes this a no-op there (correct — sync runs precisely when tree shapes have drifted; asserting they haven't would reject every legitimate divergence repair).
- 4 unit tests pin the four shapes: matches → passes, mismatch → `TreeStateMismatch`, absent ancestor → skips, empty list → no-op.

### `crates/storage/src/error.rs` — new error variant

- `StorageError::TreeStateMismatch(Id)` — distinct from `InvalidSignature` so operators can tell "writer's authorization is invalid" apart from "receiver's tree drifted from writer's view."

### `crates/node/primitives/src/sync/hash_comparison.rs` — wire format

- `LeafMetadata` gains `authorization: Option<StorageType>` field. Carries the writer's signature data + access-control list for `Shared` / `User` entities so the receiver can verify without consulting the originator's tree state. `None` for `Public` / `Frozen`.
- `LeafMetadata::with_authorization()` builder method.

### `crates/node/src/sync/helpers.rs` — DRY helper

- `wire_authorization_for(&Metadata) -> Option<StorageType>` — single source of truth for the "which storage types need wire authorization" decision. All four `TreeLeafData` sender sites go through this helper rather than open-coding the match. A future addition (e.g. a new authorization-bearing storage type) only has to be added in one place.

### `crates/node/src/sync/helpers.rs::apply_leaf_with_crdt_merge` — receiver

- When the wire carries `authorization`, set the action's `metadata.storage_type` from it — preserves the writer's signature into the action the storage verifier then checks.
- When no wire authorization but an existing entity exists, preserve the existing entity's storage_type — closes the v1 silent storage-type-flip bug where every sync apply downgraded entities to `Public` via `Metadata::default()`.

### Sender sites populate `authorization`

Four `TreeLeafData::new` call sites updated to chain `.with_authorization(wire_authorization_for(&metadata))` when applicable:
- `crates/node/src/sync/hash_comparison.rs` (initiate path)
- `crates/node/src/sync/hash_comparison_protocol.rs` (responder path + collect_leaves_recursive)
- `crates/node/src/sync/level_sync.rs`

## Test plan

* [x] `cargo check --workspace` — green
* [x] `cargo fmt --all -- --check` — clean
* [x] `cargo test --workspace` — all real tests green (one pre-existing autonat_v2 flake confirmed unaffected)
* [x] 10 new unit tests for `payload_for_signing` properties — all pass
* [x] 4 new unit tests for `verify_ancestor_integrity` paths — all pass
* [ ] CI green on push

### Test fallout (16 sites)

The v1 apply path silently accepted placeholder ancestor hashes for `Public` actions (no signature check) and rejected them for `Shared` / `User` via signature mismatch. The v2 explicit check now rejects them uniformly as `TreeStateMismatch`. 16 tests hand-crafted actions with placeholder hashes (`[0; 32]`, `[37; 32]`, etc.):

* 2 in `calimero-storage` lib tests
* 14 in `calimero-node` (p3_dag_causal_tests + p5_partition_scenarios_tests)

Fixed via two patterns:
* **Helper update** — `setup_root` in 4 test files (storage, p3, p5, dag_causal_shared_e2e) now fetches the post-`add_root` full_hash and returns a `ChildInfo` with that value. All 14 p3/p5 + 3 dag_causal_shared_e2e tests pass without per-test changes.
* **Inline fix** — the one `apply_action__sparse` test scenario where the helper update wasn't enough rebuilds the action with the real root hash post-first-apply, preserving the test's intent.

Two more sites surfaced with the same pattern after the wire changes:
* `crates/node/tests/sync_sim/storage.rs::update_entity_data` — the "create new entity as child of root" branch used `[0; 32]` as the root ancestor hash. Fetches actual hash now.
* `crates/node/tests/dag_causal_shared_e2e.rs::setup_root` — same helper-update pattern.

## Migration

Pre-1.0, no backwards-compat constraint per project stance. Every `Shared` / `User` entity stored before this PR has a signature over the v1 payload; the v2 verifier rejects it. Operational migration: wipe state on upgrade and regenerate via delta replay. No migration shim in the verifier.

## What's not in this PR (out of scope)

* Full `apply_action` decomposition into trait-based components (`SignatureVerifier`, `NonceVerifier`, etc.) — `apply_action` is still 486 lines and could benefit from SOLID extraction. Did the minimum DRY refactor that fell naturally out of this change (`hash_authorization_for_payload` extraction, `wire_authorization_for` helper, `verify_ancestor_integrity` extraction). A full architectural rewrite of the apply pipeline deserves its own PR.
* Lazy-stamp migration of pre-v2 entities. Wipe-on-upgrade is the simpler path given pre-1.0; lazy-stamp can be a follow-up if operationally required.

---

> [!NOTE]
> **High Risk**
> High risk because it changes signature payload semantics, sync wire formats (HashComparison + Snapshot), and apply-time verification/nonce logic that gate data acceptance across peers.
> 
> **Overview**
> Enables HashComparison-based reconciliation for `Shared`/`User` entities by redefining `Action::payload_for_signing` to commit only to *authorization + nonce* (not ancestor Merkle hashes) and by sending the full `StorageType` authorization triple on sync leaves.
> 
> Updates sync apply to preserve/propagate storage provenance: `LeafMetadata` now carries optional `authorization`, senders populate it via `wire_authorization_for`, and `apply_leaf_with_crdt_merge` uses wire auth (or preserves existing storage type) to avoid silent downgrades to `Public`.
> 
> Hardens and modernizes state bootstrap: Snapshot pages now stream structured `SnapshotRecord` entries (entity `Entry`+`Index`), and receivers verify per-entity signatures (`Interface::verify_snapshot_entity_signature`) before persisting; auxiliary records are currently rejected. Adds `update_signature_in_place` + context/execute changes to persist real signatures back into local indexes (replacing `[0; 64]` placeholders) so subsequent sync responses remain verifiable.
> 
> Adjusts apply semantics and tests: adds explicit (currently warn-only) ancestor integrity checking, makes same-nonce signed replays idempotent (avoiding false `NonceReplay` on redundant sync delivery), fixes test helpers to use post-`add_root` hashes, and adds a new partition-based E2E workflow guarding Shared/User reconciliation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02b977372ed4cfbe65fa013815a3bc7e71da2f81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/storage/src/action.rs
+++ b/crates/storage/src/action.rs
@@ -215,8 +215,29 @@ impl Action {
 ///
 /// **What it doesn't commit to**: anything tied to tree state. The
 /// signature is transferable across receive paths; tree-shape
-/// integrity is enforced by [`AncestorIntegrity::verify`] at apply
-/// time, separately.
+/// integrity is enforced by
+/// [`Interface::verify_ancestor_integrity`](crate::interface::Interface)
+/// at apply time, separately.
+///
+/// **Domain separation**:
+/// * Variable-length sets (writers) are prefixed with a `u32` count so
+///   that 1 writer + signer hint cannot align byte-for-byte with 2
+///   writers + no hint (both used to be 64 + 32 + 1 = 97 bytes after
+///   the type tag with the old length-free encoding).
+/// * `signature_data` presence is tagged (`[1u8]` / `[0u8]`) BEFORE its
+///   payload bytes — so the no-sig-data case can't alias with a
+///   nonce-zero signed case. Without this, e.g. a Shared action with
+///   `signature_data: None` would hash identically to one with
+///   `signature_data: Some { nonce: 0, signer: None }` (both append
+///   nothing distinguishable past the writers list).
+///
+/// **Public / Frozen** intentionally commit to nothing beyond the type
+/// tag — neither is signature-verified in `Interface::apply_action`,
+/// so a payload hash for these types is not load-bearing. The verifier
+/// MUST check the storage type matches the stored entity's type
+/// before treating any signature as authoritative; otherwise a forged
+/// `Public` upsert with the same id + data as a `Shared` upsert would
+/// produce a trivially-collidable payload.
 fn hash_authorization_for_payload(hasher: &mut Sha256, metadata: &Metadata) {
     match &metadata.storage_type {
         StorageType::Public => {
@@ -233,8 +254,21 @@ fn hash_authorization_for_payload(hasher: &mut Sha256, metadata: &Metadata) {
         } => {
             hasher.update([2u8]); // type tag
             hasher.update(owner.as_ref() as &[u8; 32]);
+            // sig-data presence tag — see "Domain separation" in the
+            // function doc.
             if let Some(sig_data) = signature_data.as_ref() {
+                hasher.update([1u8]);
                 hasher.update(sig_data.nonce.to_le_bytes());
+                // `User` deliberately does NOT commit to
+                // `sig_data.signer`: there is exactly one owner, the
+                // verifier always uses `owner` directly, and the hint
+                // is unused on this path (only `Shared` uses it for
+                // O(1) writer-set lookup). Keeping the User payload
+                // shape independent of the optional hint avoids a
+                // class of "signer field set to a different value but
+                // still passes verify" footguns at zero cost.
+            } else {
+                hasher.update([0u8]);
             }
         }
         StorageType::Shared {
@@ -242,12 +276,21 @@ fn hash_authorization_for_payload(hasher: &mut Sha256, metadata: &Metadata) {
             signature_data,
         } => {
             hasher.update([3u8]); // type tag
-                                  // Writers serialized deterministically: BTreeSet iteration
-                                  // is sorted, so this is stable across signer / verifier.
+                                  // Count prefix gives the writers list explicit length
+                                  // separation — see "Domain separation" in the function
+                                  // doc. `u32` is plenty (a writer set in the hundreds is
+                                  // already an outlier; we never need 2^32 writers).
+            let writers_count = u32::try_from(writers.len()).unwrap_or(u32::MAX);
+            hasher.update(writers_count.to_le_bytes());
+            // Writers serialized deterministically: BTreeSet iteration
+            // is sorted, so this is stable across signer / verifier.
             for writer in writers {
                 hasher.update(writer.as_ref() as &[u8; 32]);
             }
+            // sig-data presence tag — see "Domain separation" in the
+            // function doc.
             if let Some(sig_data) = signature_data.as_ref() {
+                hasher.update([1u8]);
                 hasher.update(sig_data.nonce.to_le_bytes());
                 if let Some(signer_hint) = sig_data.signer {
                     hasher.update([1u8]); // hint-present tag
@@ -255,6 +298,8 @@ fn hash_authorization_for_payload(hasher: &mut Sha256, metadata: &Metadata) {
                 } else {
                     hasher.update([0u8]); // hint-absent tag
                 }
+            } else {
+                hasher.update([0u8]);
             }
         }
     }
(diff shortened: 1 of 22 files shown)


frdomovic (commented):
## 🤖 MeroReviewer

**Reviewed by 1 agents** | Quality score: 26% | Review time: 93.4s

🟡 1 warnings, 📝 1 nitpicks. See inline comments.

---

<sub>🤖 Generated by [MeroReviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-38eaaa6a`</sub>

crates/storage/src/action.rs
```diff
@@ -138,98 +138,351 @@ impl Action {
...
-                writers: writers.clone(),
-                signature_data: signature_data.as_ref().map(|sig_data| SignatureData {
-                    nonce: sig_data.nonce,
-                    signature: [0; 64], // Use placeholder for hash
-                    signer: sig_data.signer,
+            hasher.update([3u8]); // type tag
+                                  // Writers serialized deterministically: BTreeSet iteration
+                                  // is sorted, so this is stable across signer / verifier.
+            for writer in writers {
+                hasher.update(writer.as_ref() as &[u8; 32]);
```
frdomovic, line 288: 🟡 **Public/Frozen storage types include no authorization data in signed payload**

For `StorageType::Public`, the signed payload only includes the `v2_upsert` prefix, the entity ID, the data bytes, and a single zero byte (type tag). For `StorageType::Frozen`, similarly only a type tag `[1u8]` is added. This means any entity with `Public` or `Frozen` storage type has a signature payload that commits to no access-control information beyond the type tag. While Public entities don't require signatures in practice, the `payload_for_signing` function is still called for them and the resulting hash is used in verification paths. If a verifier ever accepts a Public-typed action without checking that the storage type is actually Public, the minimal payload could be exploited to forge actions by type-confusion (e.g., downgrading a Shared entity to Public to bypass writer-set checks). The PR description notes no production code branches on `InvalidSignature` specifically, but the type-tag-only payload for Public/Frozen creates a very weak commitment.

**Suggested fix:**
```
Consider documenting explicitly that `payload_for_signing` for Public/Frozen is intentionally minimal and that callers must separately verify the storage type hasn't been downgraded. Alternatively, add a note in the verifier that Public-typed actions bypass signature verification entirely rather than verifying a minimal payload.
```

crates/storage/src/error.rs
```diff
@@ -56,6 +56,20 @@ pub enum StorageError {
...
+    /// The action's claimed ancestor merkle hash does not match the receiver's
+    /// local state. Surfaced by [`AncestorIntegrity::verify`] on the
+    /// delta-replay path; the signature itself verified, but the tree shape
+    /// the signer asserted doesn't line up with the receiver's local tree.
+    ///
+    /// This is distinct from [`Self::InvalidSignature`] (cryptographic
+    /// rejection) and lets operators tell "the writer's authorization is
+    /// invalid" apart from "the receiver's tree drifted from the writer's
+    /// view." The sync-reconcile path skips this check by definition
+    /// (sync runs precisely when tree shapes have drifted), so the error
```
frdomovic, line 70: 📝 **TreeStateMismatch doc comment references AncestorIntegrity::verify which doesn't exist in the codebase**

The `TreeStateMismatch` error variant's doc comment references `[AncestorIntegrity::verify]` as a link, but the actual implementation is `Interface::verify_ancestor_integrity` (a private method). The doc link will generate a broken link warning and is misleading about the actual API surface.

**Suggested fix:**
```
Change the doc link to reference `Interface::verify_ancestor_integrity` or remove the link and just describe the behavior in prose.
```

frdomovic:
## Documentation Review

The following documentation may need updates based on the changes in this PR:

- 🟡 **AGENTS.md**: `AGENTS.md` exists but was not updated — consider updating it to reflect the architecture changes in this PR.
- 🟡 **CONTRIBUTING.md**: `CONTRIBUTING.md` exists but was not updated — consider updating it to reflect the architecture changes in this PR.
- 🟡 **docs/**: Static HTML docs in `docs/` may need updating — architecture-impacting changes detected. On merge, `update-docs` will scan this directory and open a PR if any pages need to change.

PR #2399, opened 2026-05-19
DEBUG: instrument Root::sync to surface WASM-unreachable cause

chefsale:
Diagnostic-only PR. Adds an error log inside Root::sync (and commit_headless) so the next failing run of hashcomparison-shared-user-reconcile prints the actual StorageError variant before the SDK macro's .expect() turns it into a WASM unreachable trap.

Do not merge. Will revert / drop after we identify the root cause.


chefsale:
Superseded by #2400 — clean branch from origin/master (the original was on top of the pre-squash feat/single-signature-redesign and conflicted with master)

PR #2411, opened 2026-05-20
fix(node/sync): HashComparison must verify root-hash convergence before reporting success

### LINKED ISSUE #2407: HashComparison sync reports success without verifying root-hash convergence — peers can stay permanently divergent
## Summary

A `HashComparison` sync session can **complete and report success while the two peers remain divergent**. The initiator merges some leaves, closes the stream, logs `HashComparison sync complete`, and returns `Ok(stats)` — with **no check that the local root hash now matches the peer's**. A node in this state believes it is synced and never recovers, because every subsequent interval sync repeats the same no-op "successful" sync.

This is behind the intermittent `Sync regression (smoke)` failures.

## Evidence

From CI run [26117153887](https://github.com/calimero-network/core/actions/runs/26117153887/job/76809288598) (`Sync regression (smoke)`, master `ae03cd14`), `Round 2 sync` step, 2-node test. Downloaded node logs:

- **node-1** converged: root `jWchF4Ub…` → `412WUJD…`
- **node-2** never moved: root **frozen at `2Ezjmp…`** for the full 31s / 15 convergence checks.

node-2 ran a sync session ≈ once/second; **every one looked like this**:

```
Starting sync session     root_hash=2Ezjmp…   chosen_peer=node-1
Protocol selected         HashComparison   reason="high divergence (>50%)"
                          local_root=2Ezjmp…  remote_root=412WUJD…  local_entities=4  remote_entities=1
HashComparison sync complete   nodes_compared=5  entities_merged=2  entities_pushed=0  nodes_skipped=3
HashComparison sync completed successfully
Sync finished successfully     took≈2ms     success_count=84, 85, 86, 87…
```

`success_count` climbed past 87 while the node stayed divergent. The network was healthy throughout — stream opens in ~8ms, sessions complete in 2–10ms. This is **not** a transport flake; it is a sync-correctness bug.

## Root cause / code locations

**1. No convergence gate (definite defect).**
`crates/node/src/sync/hash_comparison_protocol.rs` — `run_initiator_impl` (~L179-370): after the DFS merge it does `transport.close()`, logs `HashComparison sync complete`, and returns `Ok(stats)`. It never compares the post-merge local root hash against `remote_root_hash`. A non-converging sync is indistinguishable from a converging one.

The sibling `crates/node/src/sync/level_sync.rs` (L405-427) *does* compute `stats.root_hash_verified = local_root_hash == remote_root_hash` — but when it is `false` it only emits a `debug!` ("Root hash differs after sync (expected if local had concurrent changes)") and still returns success. In this scenario node-2 had **no** concurrent local writes during the sync window, so non-convergence is not "expected" — yet the protocol treats it as acceptable.

→ Effect: the bug is invisible to the node. It cannot self-heal because it never learns it failed.

**2. The underlying merge does not converge node-2 (needs protocol-author review).**
Same protocol, opposite outcomes: node-1 converged, node-2 did not. node-2 reports `entities_merged=2` on every attempt yet its root hash never changes, and `nodes_skipped=3`. Open questions for whoever owns the HashComparison design:
- With content-addressed internal node ids, a logically-corresponding-but-divergent child appears in **both** `remote_only_children` and `local_only_children` (never `common_children`) in `compare_tree_nodes` — is there a path where node-1's divergent Round-2 leaf ends up in the `nodes_skipped` set?
- Can `apply_leaf_with_crdt_merge` apply a leaf (counted in `entities_merged`) without moving the root hash — e.g. an LWW merge where the local value wins — leaving the peers permanently unequal?

## Why it's intermittent

The `Sync regression (smoke)` workflow (titled *"Opaque-leaf HashComparison regression"*) **first failed on its own introducing change** — PR #2386 (`single-signature design — HashComparison reconciles Shared/User entities`, commit `4c9405a1`) — and has flapped since:

| Commit | Result |
|---|---|
| `4c9405a1` (#2386) | ❌ |
| `549be2f7` | ✅ |
| `ae03cd14` | ❌ |

Intermittent ⇒ timing-dependent on how the two Round-2 writes interleave.

## Suggested direction (not a fix)

- **Short term / make it visible:** `run_initiator_impl` should verify `local_root == remote_root_hash` after the merge and treat a mismatch as a sync failure (so it retries / surfaces) instead of returning `Ok`. Re-evaluate the "Root hash differs after sync (expected…)" branch in `level_sync.rs` — silently accepting divergence defeats the purpose of a sync.
- **Real fix:** the HashComparison merge in the #2386 / Shared-User-reconciliation design — investigate the `nodes_skipped` path and the leaf-merge convergence per the open questions above.

## Repro

`merobox` workflow `sync-regression` (kv-store-with-handlers, 2 nodes, Round 1/Round 2 writes + `wait_for_sync`). Flaky — re-run a few times.

---
_Filed from a CI flake investigation. Related: the namespace-join stream-open robustness fix in #2398 addressed a *different* flake family (transport-level); this one is sync-correctness._

chefsale:
Addresses #2407.

## The bug

`HashComparison::run_initiator_impl` returned `Ok(stats)` regardless of whether the post-merge local root actually converged to the remote root. A session that \"merged\" some leaves but left the local root frozen was indistinguishable from a fully-converged one — the sync manager logged \"success\" and the next interval-sync repeated the same no-op merge forever. Per #2407's evidence: `success_count` climbed past 87 while node-2 stayed permanently divergent.

`LevelWise::run_initiator_impl` had the same shape, with one extra step: it set `stats.root_hash_verified` correctly, but on mismatch it only emitted a `debug!` (\"Root hash differs after sync (expected if local had concurrent changes)\") and still returned `Ok`. The \"expected\" rationale was load-bearing only for the narrow case of concurrent local writes between sync start and verification time; the common case (a real sync bug) was being masked.

## The fix

After both protocols' DFS / level-walk completes, read the local root from the index and compare against the `remote_root_hash` the initiator started with. On mismatch: WARN log with both hashes + merge stats, then `bail!` with a clear \"did not converge\" error.

The Err is what the sync manager already retries on. If concurrent local writes really caused the mismatch, the next interval-sync tick observes the now-advanced state and converges (or surfaces persistent divergence — which is the correct signal).

`stats.root_hash_verified` is now also set on HashComparison (was previously a LevelWise-only field) and included in the \"sync complete\" info log.

## Shape

- Extract `get_local_root_hash_for_context` from `level_sync.rs` into `helpers.rs`. Both protocols call it.
- Remove the (now-unused) duplicate fn in `level_sync.rs`.

## What this doesn't fix

#2407 identifies a second concern: the underlying merge in HashComparison didn't converge node-2 even though `entities_merged=2`. That's a protocol-correctness bug that needs the design owner's review (the `nodes_skipped` path with content-addressed internal node ids; `apply_leaf_with_crdt_merge` applying a leaf without moving the root). This PR makes the bug **visible** by surfacing the divergence as an Err — without that the underlying bug can't be debugged from logs. The next step is to investigate why node-2's merge didn't converge.

## Test plan

- [x] `cargo test -p calimero-node --lib sync::` — 104 passing
- [x] `cargo test -p calimero-node --lib` — 168 passing
- [x] `cargo fmt --check`, `cargo clippy` — clean
- [ ] CI: should improve the `Sync regression (smoke)` flake by turning silent-mask passes into loud retries; if the underlying merge bug fires it will surface as a persistent sync failure with a clear error message.

---

> [!NOTE]
> **Medium Risk**
> Changes sync success reporting/observability by adding post-sync root-hash verification and WARN logging; could increase log volume and alter downstream metrics/decisioning that depends on sync stats.
> 
> **Overview**
> Both `HashComparison` and `LevelWise` initiators now **read the local Merkle root after applying merges** and compare it to the handshake `remote_root_hash`, recording the result in a new/updated `root_hash_verified` stats flag.
> 
> On mismatch, the protocols now emit a **WARN** (including truncated local/remote hashes and merge stats) and include `root_hash_verified` in completion logs, rather than silently/only-debug logging; the shared root-hash read logic was extracted into `helpers.rs` as `get_local_root_hash_for_context`, removing the duplicate implementation from `level_sync.rs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2707eec09b08a2c3fa91b79db1d84e9215f919e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/hash_comparison_protocol.rs
+++ b/crates/node/src/sync/hash_comparison_protocol.rs
@@ -39,7 +39,8 @@
 //! ```
 
 use crate::sync::helpers::{
-    apply_leaf_with_crdt_merge, generate_nonce, handle_entity_push, MAX_ENTITIES_PER_PUSH,
+    apply_leaf_with_crdt_merge, generate_nonce, get_local_root_hash_for_context,
+    handle_entity_push, MAX_ENTITIES_PER_PUSH,
 };
 use async_trait::async_trait;
 use calimero_node_primitives::sync::{
@@ -123,6 +124,12 @@ pub struct HashComparisonStats {
     pub nodes_skipped: u64,
     /// Number of requests sent to peer.
     pub requests_sent: u64,
+    /// Whether the post-sync local root hash matches the remote
+    /// root hash the initiator started with. Set by the initiator
+    /// after the DFS merge completes. A `false` value indicates the
+    /// merge did not converge the two peers — see #2407 for the
+    /// failure mode this guards against.
+    pub root_hash_verified: bool,
 }
 
 /// HashComparison sync protocol.
@@ -357,14 +364,60 @@ async fn run_initiator_impl<T: SyncTransport>(
     // Close the transport to signal completion to the responder
     transport.close().await?;
 
+    // Post-sync convergence check (#2407): a HashComparison session
+    // that returns Ok without confirming the two roots actually
+    // converged is indistinguishable from a no-op — the sync manager
+    // logs "success" and the next interval-sync repeats the same
+    // empty merge forever. Verify the local root now matches the
+    // remote root the initiator started with; on mismatch, surface
+    // as an Err so the manager retries (possibly via a different
+    // protocol path) instead of silently masking the divergence.
+    //
+    // Concurrent local writes between sync start and now CAN
+    // legitimately leave the local root different from the remote
+    // root we observed at session start — but returning Err here
+    // still does the right thing: the next interval-sync tick will
+    // re-handshake against the (now possibly-advanced) remote and
+    // either converge or surface the persistent divergence. The
+    // alternative (silent Ok on mismatch) is what #2407 documents
+    // as the bug.
+    let local_root_hash = with_runtime_env(runtime_env.clone(), || {
+        get_local_root_hash_for_context(context_id)
+    })?;
+    stats.root_hash_verified = local_root_hash == remote_root_hash;
+
     info!(
         %context_id,
         nodes_compared = stats.nodes_compared,
         entities_merged = stats.entities_merged,
         entities_pushed = stats.entities_pushed,
         nodes_skipped = stats.nodes_skipped,
+        root_hash_verified = stats.root_hash_verified,
         "HashComparison sync complete"
     );
+
+    if !stats.root_hash_verified {
+        warn!(
+            %context_id,
+            local_hash = %hex::encode(&local_root_hash[..8]),
+            remote_hash = %hex::encode(&remote_root_hash[..8]),
+            nodes_compared = stats.nodes_compared,
+            entities_merged = stats.entities_merged,
+            entities_pushed = stats.entities_pushed,
+            "HashComparison sync did NOT converge — surfacing as Err so the \
+             sync manager retries instead of silently masking divergence (#2407)"
+        );
+        bail!(
+            "HashComparison sync completed but local root {} did not converge to remote root {} \
+             (compared={}, merged={}, pushed={}, skipped={})",
+            hex::encode(&local_root_hash[..8]),
+            hex::encode(&remote_root_hash[..8]),
+            stats.nodes_compared,
+            stats.entities_merged,
+            stats.entities_pushed,
+            stats.nodes_skipped,
+        );
+    }
 
     Ok(stats)
 }
--- a/crates/node/src/sync/hash_comparison_protocol.rs
+++ b/crates/node/src/sync/hash_comparison_protocol.rs
@@ -364,23 +364,27 @@ async fn run_initiator_impl<T: SyncTransport>(
     // Close the transport to signal completion to the responder
     transport.close().await?;
 
-    // Post-sync convergence check (#2407): a HashComparison session
-    // that returns Ok without confirming the two roots actually
-    // converged is indistinguishable from a no-op — the sync manager
-    // logs "success" and the next interval-sync repeats the same
-    // empty merge forever. Verify the local root now matches the
-    // remote root the initiator started with; on mismatch, surface
-    // as an Err so the manager retries (possibly via a different
-    // protocol path) instead of silently masking the divergence.
+    // Post-sync convergence check (#2407). We compare the local
+    // root against the remote root the initiator started with, but
+    // do NOT treat mismatch as fatal:
     //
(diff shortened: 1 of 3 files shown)


chefsale:
## CI failures are pre-existing #2407 — PR is doing its job

The two failing checks (`Test (scaffolding-e2e)`, `Test (mesh-soak-8node)`) are the same convergence-timeout pattern that's been flaky on master:

| Master commit | E2E - Rust Apps |
|---|---|
| `ac61f53` | ✅ |
| `8ef99cd` | ❌ |
| `ae03cd1` | ✅ |
| `549be2f` | ❌ |
| `4c9405a` | ❌ |

3 of last 5 master runs failed independently of this PR. Both failures show \"Wait for ... sync ... failed after 60s\" with attempts climbing to 95-98 — the symptom #2407 documents.

## The WARN is firing with concrete diagnostic

Pulled `scaffolding-e2e-e2e-node-{1,2,3}.log` from CI run 26149554267. The new `HashComparison sync did not match remote handshake root (#2407)` WARN fires repeatedly with this exact shape on the divergent context:

```
WARN HashComparison sync did not match remote handshake root (#2407).
     context_id=4X56NygpvTLoxnr3MiLF9TJ5pDRBSPr6AMZRzasy2Smo
     local_hash=21c4c19b2919fcba  remote_hash=3cdd1a3752c6ce23
     nodes_compared=0  entities_merged=0  entities_pushed=0  nodes_skipped=0
```

All four counters at zero. The sync session opens, does no work, returns Ok with all-zero stats — and the local root doesn't match what the peer claimed at handshake. This is exactly the symptom #2407 describes: \"success_count climbed past 87 while node-2 stayed divergent.\"

Most likely root-cause path (matches code reading):
- Peer's root advanced between handshake and our subsequent root-node request
- Peer responds with `not_found: true` for the now-stale `remote_root_hash`
- Initiator loop `continue`s past the empty response, the DFS stack stays empty, loop exits
- `transport.close()` + `Ok(stats)` with all-zero stats
- Local root unchanged, remote_root_hash captured at handshake is now stale → WARN fires

## Recommendation

Land this PR — it strictly improves diagnosability of #2407 without making the flake worse. The structural fix (either an explicit fresh-handshake on root `not_found`, or restructuring `compare_tree_nodes` to reconcile on `entity_id` rather than `child_hash`) is the next PR, informed by the now-visible diagnostic above.

🤖 Diagnostic post from Claude.

PR #2413, opened 2026-05-20
fix(node/sync): deprioritise timed-out peers across rounds in namespace-join

### LINKED ISSUE #2403: node/sync: track timed-out peers across rounds in namespace-join discovery
Follow-up to #2398. Surfaced in review-4321393426 suggestion #4.

## Context

`initiate_namespace_join` now retries with a per-peer timeout and randomizes peer order per round (#2398). The shuffle distributes the cost of a consistently-stale peer fairly, but within a single discovery sequence the same stale peer can still be tried in multiple rounds — burning `open_stream_timeout` on it each time.

## Suggested approach

Maintain a `HashSet<PeerId>` of peers that returned `Err(_)` from `time::timeout` (i.e. exceeded `open_stream_timeout`) earlier in the same `initiate_namespace_join` call. On subsequent rounds, after the shuffle:

- **Partition** the shuffled list so timed-out peers go last (similar to the anchor-first partition pattern in `perform_interval_sync` at `crates/node/src/sync/manager/mod.rs:886`).
- Or **skip** them for one round and only retry if no untried peer is available.

Outright excluding them risks excluding a peer that's just briefly congested, so the partition-last pattern is preferred.

## Why it's a follow-up, not in #2398

#2398's scope was \"make namespace-join stream-open resilient to flaky peers\" via timeout + retry. Cross-round state-tracking is an enhancement on top — useful but not required for the original reliability fix, and bigger change-wise.

## Code anchors

- `crates/node/src/sync/manager/mod.rs` — `initiate_namespace_join`, the `'connect` loop.
- `crates/node/src/sync/manager/mod.rs:886` — existing partition pattern (`choose_multiple` + `partition_peers_anchor_first`).

🤖 Filed by Claude during PR #2398 review follow-up.

chefsale:
Addresses #2403 (follow-up to #2398 / review-4321393426 suggestion #4).

## Summary

`open_namespace_join_stream` (from #2410) shuffles peers per round, but a consistently-stale peer can still get tried first 50/50 of the time, burning the full `open_timeout` each round. This PR tracks peers that hit `tokio::time::timeout` on `open_stream` and deprioritises them — stable partition: non-timed-out peers first, timed-out last — in subsequent rounds.

Plain `Err` from `open_stream` does NOT add a peer to the set: those failures might be transient and the peer is still worth trying ahead of one that ate the full timeout budget. Excluded peers can still be retried (they just go to the tail), matching the \"degraded but not excluded\" pattern the issue suggested.

## Mock extension

The existing `MockSyncNetwork` has a flat `open_stream` queue — every call consumes the head regardless of which peer was asked, which can't model \"this specific peer hangs while others error fast\". Added per-peer queues:

- `push_open_stream_err_for_peer(peer, msg)`
- `push_open_stream_hang_for_peer(peer, sleep, msg)`

The `open_stream` impl consults the per-peer queue first, falls back to the global queue. Mirrors the layered-fallback shape from the existing `mesh_peers` sticky-last logic.

## Test

`timed_out_peers_are_deprioritised_in_subsequent_rounds`: two peers `slow` and `fast`, scripted per-peer; slow always hangs, fast always errors quickly. Asserts that from round 2 onward every pair of consecutive `open_stream` calls is `(fast, slow)` — fast first because slow landed in the timed-out set after round 1.

## Test plan

- [x] `cargo test -p calimero-node --lib sync::manager::namespace_join` — 6 passing
- [x] `cargo test -p calimero-node --lib` — 169 passing
- [x] `cargo fmt --check`, `cargo clippy` — clean
- [ ] CI green

Closes #2403.

---

> [!NOTE]
> **Medium Risk**
> Adjusts retry peer-selection logic in namespace-join discovery, which can affect connection behavior and timing under failure conditions; changes are covered by new targeted unit tests and mock enhancements.
> 
> **Overview**
> Improves `open_namespace_join_stream` retry behavior by tracking peers that hit the per-peer `tokio::time::timeout` and **deprioritising them in subsequent rounds** (stable partition: non-timed-out peers first, timed-out peers last) while still allowing retries.
> 
> Extends `MockSyncNetwork` to support **per-peer scripted `open_stream` responses** plus an `open_stream` call-order log, and adds a unit test asserting that once a peer times out, later rounds try healthier peers before it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41c9a841d6165de3a2c326b0c43670ec80bd6c0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/manager/namespace_join.rs
+++ b/crates/node/src/sync/manager/namespace_join.rs
@@ -13,8 +13,11 @@
 //! this module deliberately holds none of it so the comments don't
 //! drift out of sync.
 
+use std::collections::HashSet;
+
 use calimero_network_primitives::stream::Stream;
 use libp2p::gossipsub::TopicHash;
+use libp2p::PeerId;
 use rand::seq::SliceRandom;
 use tokio::time;
 use tracing::debug;
@@ -82,6 +85,16 @@ pub(super) async fn open_namespace_join_stream(
     // `std::time::Instant`.
     let connect_started = tokio::time::Instant::now();
 
+    // Peers that hit `time::timeout` on `open_stream` earlier in
+    // this discovery sequence — i.e., we've already spent
+    // `open_timeout` waiting on them once. Subsequent rounds
+    // deprioritise (don't exclude) them so a consistently-stale
+    // peer doesn't keep burning the per-peer budget ahead of
+    // healthy ones. Plain `Err` (no timeout) does NOT add to this
+    // set — those failures might be transient and the peer is
+    // still worth trying first.
+    let mut timed_out_peers: HashSet<PeerId> = HashSet::new();
+
     let mut stream = None;
     'connect: for attempt in 1..=mesh_retries {
         if connect_started.elapsed() >= connect_deadline {
@@ -98,6 +111,14 @@ pub(super) async fn open_namespace_join_stream(
         // `choose_multiple` would produce. Matches the pattern used
         // in `perform_interval_sync`.
         peers.shuffle(&mut rand::thread_rng());
+        // Partition shuffled list: peers we haven't timed out on
+        // come first, timed-out peers last. Stable-by-construction
+        // (sort_by_key with a bool maps false=0/true=1, preserving
+        // the random-relative order within each partition). This
+        // means peers that briefly errored last round still get
+        // tried before the chronically-timing-out peer; only
+        // *prior-timeout* peers get pushed to the tail.
+        peers.sort_by_key(|p| timed_out_peers.contains(p));
 
         for peer in &peers {
             if connect_started.elapsed() >= connect_deadline {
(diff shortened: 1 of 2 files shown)


chefsale:
Closing as over-engineering after discussion.

The deprioritization logic only earns its keep when the namespace topic mesh has ≥2 peers and at least one has a stale transport. In practice:

- **Cold-start namespace-join** (the dominant case): the mesh typically has just the owner (N=1). Deprio is a no-op — a HashSet that stays size-1 and a sort on a single-element list.
- **Joining an established namespace** (N>1, one stale peer): deprio would help, but this is a rare path.

Meanwhile, the interval/context sync path already has a richer peer-selection mechanism (`partition_peers_anchor_first`) which orders Owner/Admin/ReadOnlyTee ahead of non-anchors — that's where multi-peer cost-distribution actually matters, and that path is untouched by #2403.

Net: the cost (extra state + test surface) isn't justified by the realistic namespace-join mesh shape. Closing alongside #2403.

Issue #2313, opened 2026-05-08
Decompose SyncManager into focused components


chefsale:
Part of #2302.

## Scope

\`crates/node/src/sync/manager/mod.rs\` is 3.4k LOC and does four jobs in one struct. Split into focused types.

## What changes

```rust
struct SyncManager {
    sessions: SessionTracker,        // per-context sync session state, last_sync timestamps
    peers: PeerSelector,             // mesh peer discovery + namespace fallback
    protocol: ProtocolSelector,      // chooses HashComparison/LevelSync/Snapshot per divergence
    driver: SyncDriver,              // run loop, retry/backoff, interval sync
}
```

Each component is:

- A struct in its own file (\`sessions.rs\`, \`peers.rs\`, \`protocol.rs\`, \`driver.rs\`).
- Independently unit-testable.
- Constructed with explicit dependencies (no implicit access to \`SyncManager\` internals).

The actor (\`SyncManager\` in \`mod.rs\`) becomes a thin shell that wires them and forwards Actix messages.

## Why

- 3.4k file is unmaintainable. Code review on changes to sync becomes "read 3400 lines."
- Each concern has different test needs (peer selection: pure logic; driver: async lifecycle).
- After this lands, \`mod.rs\` should be ~300 LOC of glue.

## DRY: consolidate fetch-from-peer paths

While decomposing, extract a shared helper for the request/response idiom used by \`delta_request.rs\`, \`parent_pull.rs\`, \`snapshot.rs\` (each opens a stream, sends a request, awaits a response, applies). A small \`PeerExchange<Req, Resp>\` helper or a \`RequestOnce\` trait reduces ~200 LOC of duplication.

## Done when

- \`mod.rs\` ≤ 500 LOC, mostly Actix \`Handler\` impls forwarding to the four components.
- Each component has at least one focused unit test.
- \`cargo test -p calimero-node\` green; behaviour unchanged.

## Out of scope

- Protocol selection logic changes (this is a refactor, not a behaviour change).
- New protocols.

## Testing

**Today's coverage**: `SyncManager` is only exercised through full-actor integration tests. Internal logic (peer selection, protocol selection, session tracking, retry loop) has no isolated tests.

**Add as part of this work** — at least one unit test file per component:

- `SessionTracker`: state transitions (idle → in_progress → complete), concurrent-sync prevention, last_sync timestamp updates.
[...]

PR #2424, opened 2026-05-20
refactor(node/sync): extract peer-selection helpers into sync::peers (Phase 1 of #2313)

chefsale:
## Summary

Phase 1 of #2313 (the `SyncManager` decomposition). Extracts peer-selection helpers into a new `sync::peers` module so the logic can be tested in isolation and so `manager/mod.rs` shrinks toward the ~500-LOC target.

Behaviour change: none. Same code, different file + tighter testability.

## What moved

New file `crates/node/src/sync/peers.rs` (465 lines, ~half tests) owns:

1. **`discover_mesh_peers_with_namespace_fallback`** — async helper extracted from the inline 90-line block in `perform_interval_sync`. Retries `mesh_peers` on the context topic up to `max_retries` rounds, falls back to the namespace-topic mesh if the context topic stays empty across all retries, and bails with a diagnostic if both come up empty. Returns a `DiscoveryOutcome` (peers + `PeerSource` enum + attempts + elapsed) so callers can branch on which mesh produced the peers.

   The namespace-resolution closure stays in the manager (it needs `ContextClient` + `calimero_context::group_store::resolve_namespace` for the parent-chain walk to find the namespace root). `peers.rs` deliberately doesn't depend on `calimero-context` — the resolution is supplied as `impl FnOnce() -> Option<TopicHash>` so this module is testable against `MockSyncNetwork` alone.

2. **`partition_peers_anchor_first`** — moved verbatim from `manager/mod.rs`. Already a free function taking `&dyn SyncStateAccess` (from #2420). Just relocated; behaviour identical.

## Tests

Tests live alongside in `peers.rs`:

| Test | What it pins |
|---|---|
| `discovery_returns_context_mesh_on_first_attempt` | Happy path: peers on attempt 1, source is `ContextMesh`. |
| `discovery_errs_when_context_empty_and_no_namespace_fallback` | Empty context mesh + no namespace resolver → Err with the documented diagnostic. |
| `discovery_falls_back_to_namespace_when_context_empty` | (Limited by mock infrastructure today — verifies the Err path; full namespace-positive case will land when `MockSyncNetwork` supports per-topic responses.) |
| 6 × `partition_*` tests | Anchor-first ordering invariants. Moved from `manager/tests.rs`. |
| `partition_works_against_mock_sync_state_access` | Demonstrates the helper works against any `SyncStateAccess` impl, not just `NodeState`. |

All 308 sync_sim integration tests still pass — the discovery refactor is behaviour-preserving.

## Diff stats

- `crates/node/src/sync/manager/mod.rs`: -136 LOC (the inline discovery block is gone)
- `crates/node/src/sync/manager/tests.rs`: -196 LOC (the partition tests moved)
- `crates/node/src/sync/mod.rs`: +1 (mod declaration)
- `crates/node/src/sync/peers.rs`: +465 (new)

Net file count: +1. Net code: +∼130 LOC including the new tests + extracted helper's doc comments.

## Test plan

- [x] `cargo test -p calimero-node` — 176 lib + 308 sync_sim + smaller bins, all pass
- [x] `cargo fmt --check`
- [x] `cargo build -p calimero-node --lib`

## Out of scope (follow-up phases of #2313)

- `SessionTracker` — per-context sync session state + last_sync timestamps
- `ProtocolSelector` — `select_protocol` is already a free function in primitives; the wrapper in `manager` is light
- `SyncDriver` — run loop, retry/backoff, interval scheduling
- Shared `PeerExchange<Req, Resp>` helper to deduplicate the request/response pattern in `delta_request.rs`, `parent_pull.rs`, `snapshot.rs`

Each is its own PR. This one lands the cleanest extraction first to validate the decomposition pattern before committing to the other three.

---

> [!NOTE]
> **Low Risk**
> Low risk refactor that moves peer-discovery/ordering logic and tests into a dedicated module; behavior should remain the same aside from additional discovery telemetry (`source`).
> 
> **Overview**
> Refactors sync peer selection by extracting context mesh peer discovery (with namespace-topic fallback) and anchor-first peer ordering into a new `sync::peers` module.
> 
> `SyncManager::perform_interval_sync` now delegates discovery to `discover_mesh_peers_with_namespace_fallback` and logs the discovery `source` (context mesh vs namespace fallback). Anchor-first partitioning is called via `super::peers::partition_peers_anchor_first`, and the associated unit tests were moved from `manager/tests.rs` into `peers.rs` alongside new discovery tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91b7bf2599af7b47f95c829c6bb1a66265eacfcc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed the review feedback in 91b7bf25. Summary per finding:

### `peers.rs:87` — \`context_id.to_string()\` vs raw \`context_id\` (🟡 meroreviewer + cursor-bugbot)
**Fixed.** Both reviewers were right. The codebase consistently passes \`context_id\` directly to \`TopicHash::from_raw\` in three other places — \`manager/mod.rs:2269\`, \`manager/mod.rs:3267\`, and \`handlers/network_event/specialized.rs:245\`. \`.to_string()\` produced a hash string from \`Display\`, but \`TopicHash::from_raw\` expects the same encoding peers subscribe under (passed directly elsewhere). Inconsistency could silently return empty \`mesh_peers\` even when peers were present on the correct topic. Now matches the rest of the codebase.

### \`peers.rs\` — duplicate "Mesh peer discovery succeeded" info logs (💡 cursor-bugbot)
**Fixed.** Removed both helper-side info logs (the context-mesh success and the namespace-fallback "context mesh empty; falling back…" log). The caller's existing info log at \`manager/mod.rs:851\` already carries the full field set — \`peer_count\`, \`attempts\`, \`?mesh_elapsed\`, \`is_uninitialized\`, and crucially \`source = ?outcome.source\` which distinguishes \`ContextMesh\` vs \`NamespaceFallback\`. One success record at info, one warn on the exhausted-retries bail path. Also dropped the now-unused \`tracing::info\` import in \`peers.rs\`.

This also makes the related \"namespace fallback log missing \`attempts\`\" comment moot — there's no fallback log to add to anymore.

### \`peers.rs\` — third discovery test is a no-op for the positive fallback case (🟡 cursor-bugbot)
**Addressed via rename + follow-up issue.** The test is now named \`discovery_errs_when_both_meshes_empty\` and the long apology comment is replaced with a 4-line note pointing at #2426. The honest scope of the test (covers the both-empty Err branch, distinct from the no-resolver Err branch) is now visible in the test name itself.

The positive \`NamespaceFallback\` case — what the bot suggested — is filed as **#2426**: \`MockSyncNetwork::mesh_peers\` ignores its topic argument, so per-topic responses can't be expressed. The issue includes a sketch of the mock change and the test that should land alongside it.

### \`manager/mod.rs:807\` — \`move\` closure + \`context_id\` re-use clarity (💡 cursor-bugbot)
**Fixed.** Added a brief comment immediately before \`let resolve_namespace_topic = move || { … };\` noting that \`ContextId: Copy\` (\`[u8; 32]\`) makes the \`move\` capture a copy, so \`context_id\` remains usable for the subsequent \`discover_…\` call and the \`info!\` log that follows.

---

Verified: \`cargo fmt --check\` clean, \`cargo check -p calimero-node\` passes.

PR #2446, opened 2026-05-22
fix(network): recover relay reservations after expiry or disconnect

chefsale:
## What

Adds a recovery path for relay reservations that get lost mid-session. Without this, once a reservation drops there is no code in the network manager that re-requests it before the next process restart.

## The bug

A NAT'd node issues a relay reservation as part of the Unknown -> Unreachable reachability transition. After that point the only place in the codebase that populates `ReachabilityActions::relay_reservations` is `DiscoveryState::became_unreachable`, which runs once per transition. The node sits in `Unreachable` indefinitely, and when the reservation lapses for any reason — relayed listen address expired, listener closed by the relay, control connection dropped — the existing code marks the per-peer status `Expired` and stops there.

The relay client's libp2p `Event` enum in 0.21.1 only carries success notifications (`ReservationReqAccepted`, `OutboundCircuitEstablished`, `InboundCircuitEstablished`), so the failure signal we can act on is on the swarm side: `ExternalAddrExpired`, `ListenerClosed`, and `ConnectionClosed`.

## Real-world triggers

- Boot-node restart or replacement drops every active reservation at once.
- Laptop sleep/wake: NAT mapping or TCP keepalive timeout kills the control connection.
- Wi-Fi switch or VPN toggle changes the egress IP under an established TCP connection.
- Long-session renewal blip: with `max_circuit_duration = 3600s` the renewal happens once an hour, and a transient loss at that moment is terminal.
- Relay quota exhaustion: a renewal request denied because the relay is over capacity expires the existing reservation.

Symptom in all cases: the node looks connected (mDNS and surviving direct peers keep working) but is invisible to anyone discovering it via the relay until the user restarts.

## What changes

`DiscoveryState::on_relay_reservation_lost(peer)`:
- marks the peer's reservation `Expired`,
- if the peer is still in `relay_index` and the prior status was `Accepted` or `Requested`, returns a `ReachabilityActions` with `relay_reservations: vec![peer]`,
- is a no-op for peers that never held an active reservation, and idempotent if called repeatedly for the same loss.

Three event handlers in `crates/network/src/handlers/stream/swarm.rs` now call it:
- `SwarmEvent::ExternalAddrExpired` for relayed addresses (the existing branch that previously called `update_relay_reservation_status` without further action).
- `SwarmEvent::ListenerClosed` for any closed listener whose addresses include a relayed multiaddr. Previously trace-logged only.
- `SwarmEvent::ConnectionClosed` when the peer is a known relay and no connections remain. Required because abrupt TCP resets don't always produce the other two events.

The downstream `create_relay_reservation` still checks `is_relay_reservation_required(registrations_limit)`, so reservations don't exceed the configured budget.

## Tests

Six new unit tests in `crates/network/src/discovery/state_tests.rs`:

- Lost `Accepted` reservation flips to `Expired` and queues recovery.
- Lost `Requested` reservation also queues recovery (the relayed listener was open even if the reservation never confirmed).
- Repeated loss for an already-`Expired` peer is a no-op (idempotency under duplicate events).
- Loss for a `Discovered` peer (relay we know about but never reserved with) is a no-op.
- Loss for a peer that isn't a relay at all is a no-op.
- Two independent relays produce independent recovery actions.

All 12 tests in `cargo test -p calimero-network --lib` pass.

## How to verify in the wild

Run a NAT'd merod against the dev boot-node with debug logging on `calimero_network::handlers::stream::swarm` and `calimero_network::discovery`. Force a reservation loss by one of:

1. Bouncing the boot-node EC2 instance (drops every reservation at once).
2. Toggling the laptop's Wi-Fi off and on.
3. `sudo pkill -STOP $(pgrep merod) && sleep 30 && sudo pkill -CONT $(pgrep merod)` (simulates App Nap freeze; the control connection times out server-side).

Expected log sequence after this change:

```
INFO  Swarm: External address expired: /ip4/.../p2p-circuit/...
DEBUG Attempting to register with rendezvous node
INFO  Swarm: External address confirmed: /ip4/.../p2p-circuit/...
```

Before this change the second and third lines never appear without a restart.

## Out of scope

- Address-book hygiene (stale `PeerInfo.addrs` not pruned, inbound addresses not recorded, identify-reported addresses not mirrored into discovery state). Tracked separately; will follow as PR 2.
- Backoff for repeated reservation failures against the same relay. Not needed in practice today (lost reservations are sporadic), but worth revisiting if the recovery path ever spams a denied relay.

---

> [!NOTE]
> **Medium Risk**
> Touches libp2p swarm event handling and relay reservation state transitions; bugs could cause missed recovery or repeated `listen_on` attempts under disconnect bursts.
> 
> **Overview**
> Adds `DiscoveryState::on_relay_reservation_lost` to **mark relay reservations as `Expired` and (only when previously `Accepted`) enqueue a single recovery reservation request**, avoiding duplicate relayed listeners during bursty disconnect/expiry events.
> 
> Updates swarm handling to trigger this recovery when a relay `ConnectionClosed` leaves no remaining connections, when a relayed `ListenerClosed` occurs, and when a relayed `ExternalAddrExpired` occurs (replacing the prior “mark expired only” logic), and adds unit tests covering idempotency/no-op cases and per-relay independence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a60f44983a9ff073ac09e3d652aed836a5a4805a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Confirming this is a real bug. Walked through the burst sequence and reproduced it on paper:

1. `ConnectionClosed` → handler sees `Accepted` → marks `Expired` → queues recovery → `create_relay_reservation` runs `listen_on` → status `Requested`.
2. `ListenerClosed` for the dead old listener → handler sees `Requested` → previous code treated this as "active" → queued another `listen_on`.
3. `ExternalAddrExpired` for the dead old address → same.

Net: up to three duplicate libp2p listeners per disconnect. Quota-denial path was even worse — would loop indefinitely re-issuing `listen_on` against a deliberately-denying relay.

Fixed in a60f4498 by splitting the two transitions explicitly:

- `Accepted -> Expired`: a real loss; queue recovery.
- `Requested -> Expired`: either a request that just failed (queuing would loop on denial) or a stale event for a prior loss whose recovery is in flight (queuing would duplicate). Mark `Expired` so state stays accurate, but do not queue. The in-flight libp2p listener keeps running and will set status back to `Accepted` via `ExternalAddrConfirmed` if/when it succeeds.
- `Expired` / `Discovered` / unknown: no-op.

The unit test that asserted "loss from Requested queues recovery" was encoding the bug. Replaced with a `test_on_relay_reservation_lost_is_idempotent_under_event_burst` test that simulates the full three-event burst with a `Requested` set in the middle (mimicking `create_relay_reservation` running between events) and asserts exactly one recovery action is produced.

PR #2451, opened 2026-05-22
fix(server): expose X-Auth-Error via CORS so JS clients can refresh

rtb-12:
## Summary

- Add `Access-Control-Expose-Headers: x-auth-error, x-auth-user, x-auth-permissions` to the server's top-level CORS layer so cross-origin JS clients can actually read the auth signaling headers the middleware already emits.
- Extract the CORS layer into `fn build_cors_layer() -> CorsLayer` to give the configuration a single named seam.
- Add `cors_tests` with two regression guards (red on master, green with the fix).

## The bug

`crates/auth/src/auth/middleware.rs` sets `X-Auth-Error: token_expired` on 401 responses when a JWT has expired (see `jwt.rs:329` → `middleware.rs:99`). mero-js's `WebHttpClient` reads that header to decide whether to invoke its automatic refresh-on-401 flow. Without `expose_headers` listing it, browsers and Tauri's webview strip the header from JS visibility on cross-origin responses — mero-js sees a plain 401, never refreshes, and `tauri-app`'s `App.tsx` shows the login screen.

End-user symptom: the desktop app logs users out roughly once per access-token TTL (~1h, including while idle in the background via the `setInterval(checkConnection, 10000)` health poll), even though the refresh token is valid for ~30 days. This was the root cause of the auto-logout in the deployed `tauri-app`.

## Why exposing three headers, not just one

`X-Auth-User` and `X-Auth-Permissions` are also set by the auth middleware on successful responses (`middleware.rs:78,82`). They were invisible to JS for the same CORS reason. Exposing them now removes a footgun for any future client logic that wants to read them — equally cheap, equally safe.

## Test plan

- [x] `cargo test -p calimero-server --lib cors_tests` — both tests **fail** on master (verified by deleting `expose_headers` and re-running: both panic with the exact "missing Access-Control-Expose-Headers" diagnostic the tests were designed to surface).
- [x] `cargo test -p calimero-server --lib cors_tests` — both tests **pass** with the fix applied.
- [x] `cargo test -p calimero-server` — full crate suite passes (12 unit + 1 doctest).
- [ ] (manual, optional) Rebuild merod, launch `tauri-app`, lower `access_token_expiry` in merod's auth config to ~60s, leave app idle 70s, confirm the next request silently refreshes (DevTools Network: a 401 with visible `X-Auth-Error: token_expired` immediately followed by a successful `POST /auth/refresh`, then the original request retried).

## Risk

Very low. `expose_headers` only widens what JS can read on cross-origin responses; it cannot block existing requests. The refactor into `build_cors_layer()` is a no-op behavior change covered by the new tests.

---

> [!NOTE]
> **Low Risk**
> Low risk CORS configuration change that only increases which response headers are visible to cross-origin JS clients; behavior is guarded by new regression tests.
> 
> **Overview**
> Fixes a CORS misconfiguration so cross-origin JS clients can read auth signaling headers by adding `Access-Control-Expose-Headers` for `x-auth-error`, `x-auth-user`, and `x-auth-permissions`.
> 
> Refactors the server’s CORS setup into `build_cors_layer()` and adds targeted regression tests to ensure `x-auth-error` is preserved/exposed (including on `401` responses).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5849ab31b3e213335708def72281bec9666e9ada. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/server/Cargo.toml
+++ b/crates/server/Cargo.toml
@@ -47,6 +47,7 @@ mero-auth.workspace = true
 [dev-dependencies]
 color-eyre.workspace = true
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tower = { workspace = true, features = ["util"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [build-dependencies]
--- a/crates/server/src/lib.rs
+++ b/crates/server/src/lib.rs
@@ -146,19 +146,7 @@ pub async fn start(
         .layer(axum::middleware::from_fn(crate::metrics::track_request))
         .layer(Extension(http_metrics));
 
-    app = app.layer(
-        CorsLayer::new()
-            .allow_origin(Any)
-            .allow_headers(Any)
-            .allow_methods([
-                Method::POST,
-                Method::GET,
-                Method::DELETE,
-                Method::PUT,
-                Method::OPTIONS,
-            ])
-            .allow_private_network(true),
-    );
+    app = app.layer(build_cors_layer());
 
     let mut set = JoinSet::new();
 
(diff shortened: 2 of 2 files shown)


rtb-12:
## Response to bot review

Triaging the 3 warnings + 1 suggestion against this PR's scope (CORS `expose_headers` + regression tests):

**Suggestion #1 — `allow_credentials` documentation** ✅ **addressed in 5849ab3**
Added a paragraph to `build_cors_layer`'s docstring documenting why `allow_credentials` is intentionally omitted (incompatible with `allow_origin(Any)`) and what would need to change before credentialed requests are viable.

**Warning #1 — `.parse().unwrap()` on `token_expired` header value** — **out of scope, follow-up**
The flagged line is `crates/server/src/auth.rs:27` (existing code, not in this PR's diff). Trivially fixable with `HeaderValue::from_static("token_expired")`; would be a 1-line change in a separate `chore(server)` PR rather than mixed into this auth-flow fix.

**Warning #2 — wildcard CORS origin + `allow_private_network(true)`** — **valid concern, scoped out**
The bot itself flags this as pre-existing. Restricting origins requires enumerating every legitimate caller (Tauri webview's `http://tauri.localhost` and `tauri://localhost`, dev `http://localhost:1420`, any web-SPA frontends serving against a local merod, the `merod` self-call, etc.) and verifying nothing breaks across platforms. That's a focused security-hardening PR, not a piggyback on the expose-headers fix. Filing as follow-up.

**Warning #3 — `msg.contains("expired")` substring match** — **out of scope, follow-up**
Pre-existing in `auth.rs:25`. Should be a typed `AuthError::TokenExpired` variant via `mero-auth`. Belongs with #1 above in a `crates/server/src/auth.rs` cleanup PR.

This PR stays narrowly focused on the bug it fixes (the header is invisible to JS); the structural items will land separately so each can be reviewed and reverted independently.

Issue #2453, opened 2026-05-22
Multi-network topology tests: IP rotation + symmetric NAT scenarios


chefsale:
## Goal

Cover the two scenarios from the disconnect analysis that cannot be reproduced with a single-process `MockRelay` or with the basic merobox fault-injection primitives in [merobox#246](https://github.com/calimero-network/merobox/issues/246):

- **Scenario 3 / 4**: client's egress IP changes mid-session (Wi-Fi switch, VPN toggle)
- **Scenario 7**: client behind symmetric NAT / CGNAT (hole-punching ineffective; relay-only path)

Both require multiple network topologies inside one test, which means multi-bridge Docker setups and dedicated NAT containers.

## Why these are separate from the basic primitives

[merobox#246](https://github.com/calimero-network/merobox/issues/246) covers single-container fault injection: pause, restart, `tc` netem. Those primitives work inside one Docker bridge. They cannot simulate:

- **A peer changing its IP and reconnecting from a different network** — `tc` can drop packets but can't make a container reappear with a new address.
- **NAT semantics** — `iptables` is required to model cone vs symmetric NAT. Without it, every container on the same bridge can directly dial every other, so the NAT path is never exercised.

These need real network topology, which Docker supports (multiple bridges + a NAT container), but is more setup work than the basic primitives.

### `relay-recovery-ip-rotation.yml` (S3 / S4)

```yaml
networks:
  - name: bridge-a
    subnet: 10.10.0.0/24
  - name: bridge-b
    subnet: 10.20.0.0/24

steps:
  - type: start_node
    name: boot-node-1
    networks: [bridge-a, bridge-b]    # boot-node reachable from both
  - type: start_node
    name: client-1
    networks: [bridge-a]
    mdns: false
  - type: wait_for_sync
  - type: move_to_network
    container: client-1
    from: bridge-a
    to: bridge-b
  - type: wait_for_sync
    timeout: 30s
  - type: assert_message_received     # or whatever signals end-to-end recovery
```

Expected: client-1 loses its old connections, the boot-node sees a new connection from a different subnet, client-1 re-establishes its relay reservation under its new IP, peers on bridge-a (now reachable only via relay) can still talk to it.

### `relay-recovery-symmetric-nat.yml` (S7)

```yaml
networks:
  - name: public
    subnet: 192.168.0.0/24
  - name: lan
    subnet: 10.0.0.0/24
    internal: true

containers:
  - name: nat-gateway
[...]

Issue #2447, opened 2026-05-22
Network resilience test plan after relay-reservation recovery fix


chefsale:
## Goal

Cover, in CI, the failure scenarios documented in the disconnect/reconnect analysis ([PR #2446 description](https://github.com/calimero-network/core/pull/2446)) so a regression in any of them fails a workflow, not a user.

The recovery code added in #2446 is unit-tested at the state-machine level. This issue tracks the broader test infrastructure needed to exercise it (and the address-book hygiene work that follows) against real networking conditions.

## Scenarios from the disconnect analysis

| # | Scenario | Trigger | Code path |
|---|---|---|---|
| S1 | Boot-node restart | TCP connection to relay terminates | `ConnectionClosed` → recovery |
| S2 | Laptop sleep/wake | NAT mapping or TCP keepalive timeout | `ConnectionClosed` → recovery |
| S3 | Wi-Fi network switch | Client egress IP changes mid-session | `ConnectionClosed` + stale `confirmed_external_addresses` |
| S4 | VPN connect/disconnect | Same shape as S3 | Same as S3 |
| S5 | Long-session renewal blip | Renewal at `max_circuit_duration` fails | `ExternalAddrExpired` → recovery |
| S6 | Relay quota exhaustion | Renewal denied due to capacity | `ListenerClosed` (or implicit `ExternalAddrExpired`) → recovery |
| S7 | Symmetric NAT / CGNAT user | Hole-punching unavailable; relay-only | All of the above, amplified |
| S8 | Tauri App Nap freeze | Userspace runtime stalled long enough that relay drops control connection | `ConnectionClosed` → recovery |

## Test coverage matrix

| # | Scenario | Unit (state machine) | Integration (MockRelay) | merobox (real Docker) |
|---|---|---|---|---|
| S1 | Boot-node restart | ✅ ([PR #2446](https://github.com/calimero-network/core/pull/2446)) | 🟡 in flight (PR follows #2446) | ❌ — see merobox issue |
| S2 | Sleep/wake | ✅ | 🟡 in flight | ❌ — `docker pause` |
| S3 | Network switch | ❌ | ❌ — needs kernel-level work | ❌ — `docker network disconnect/connect` |
| S4 | VPN toggle | ❌ | ❌ | ❌ — same shape as S3 |
| S5 | Renewal blip | ✅ | 🟡 ignored test, opt-in real-time wait | ❌ — `tc qdisc netem loss` |
[...]

PR #2454, opened 2026-05-22
test(network): MockRelay testkit + integration tests for reservation recovery

chefsale:
## What

Adds a `MockRelay` testkit in `crates/network/tests/common/` and three integration tests in `crates/network/tests/relay_recovery.rs` that exercise the libp2p events the state-machine recovery in [#2446](https://github.com/calimero-network/core/pull/2446) reacts to.

Stacked on top of [#2446](https://github.com/calimero-network/core/pull/2446). Targets `master`.

## Why

The recovery code in [#2446](https://github.com/calimero-network/core/pull/2446) is unit-tested at the `DiscoveryState` level. What is not yet tested is the wiring between libp2p's `SwarmEvent`s and the recovery method — and more importantly, that those events actually fire under the conditions we expect. This PR fills that gap with an in-process relay server we can fully control, so the bug fixed in #2446 (and any future regression) is detectable from `cargo test`, not just from a user complaining in Slack.

### `MockRelay` — in-process libp2p relay server (~320 LoC)

`tests/common/mock_relay.rs`. Wraps real `libp2p::relay::Behaviour` (server side) + `identify` in a swarm task. Exposes:

- `MockRelay::spawn()` / `spawn_with(config, keypair)`
- `peer_id()`, `bootstrap_addr()` — what a client puts in its bootstrap config
- `disconnect_peer(peer)` — forcibly close active connections; the fault injection
- `observations()` — snapshot of accepted/denied reservations and circuits
- `shutdown()` — stop the swarm task

Tunable via `MockRelayConfig`: `max_reservations`, `max_reservations_per_peer`, `reservation_duration`, `max_circuit_*` knobs.

One non-obvious detail documented inline: `libp2p::relay::Behaviour` builds reservation responses from the **server's** external addresses (so the client can construct `<relay>/p2p-circuit/<self>`). On a loopback-only mock these are empty by default since we don't run AutoNAT — the mock explicitly adds its bound listen address as external. Without that line every reservation request comes back with `NoAddressesInReservation` and the client's listener is torn down before any test can observe anything useful.

### Three integration tests (~380 LoC)

`tests/relay_recovery.rs`:

| Test | What it asserts | Scenario from #2446 analysis |
|---|---|---|
| `client_can_reserve_circuit_against_mock_relay` | A reservation can be set up against the mock. Baseline. | — |
| `relay_disconnect_triggers_client_connection_closed` | `ConnectionClosed` fires for the relay peer with `num_established == 0` after `disconnect_peer`. | S1 (boot-node restart), S2 (sleep/wake), S8 (App Nap freeze), abrupt TCP reset |
| `relay_quota_exhaustion_denies_second_client` | With `max_reservations = 1`, a second client's `listen_on(/p2p-circuit/)` produces `ListenerClosed` with an error reason. | S6 (relay quota exhausted) |

These verify that the libp2p events our recovery code is wired to actually fire in the conditions we care about. The state-machine handling of those events (mark Expired, queue recovery, idempotency under repeated events) is already covered by the `test_on_relay_reservation_lost_*` unit tests in `discovery/state_tests.rs` from #2446.

## What is not covered, and why

- **Long-session renewal blip (S5)**. The libp2p relay client auto-renews aggressively while the control connection is alive, so a short server-side `reservation_duration` alone doesn't drive the client into `ExternalAddrExpired`. Would need the relay to actively refuse a renewal — more `MockRelay` infrastructure than is in scope here. Tracked in [#2447](https://github.com/calimero-network/core/issues/2447).
- **Client IP rotation (S3, S4) and true symmetric NAT (S7)**. Hard to simulate inside a single kernel without root or extra interfaces. Needs container-based testing via the primitives proposed in [merobox#246](https://github.com/calimero-network/merobox/issues/246) and the multi-network topology work in [#2453](https://github.com/calimero-network/core/issues/2453).
- **End-to-end `NetworkManager` actor tests**. Would roughly double the size of this PR (actix System setup, recording dispatcher, introspection accessors). The current tests verify the libp2p mechanics; the state-machine unit tests verify the state transitions; the glue between them is a single match arm + method call per event. Filed as a future enhancement on [#2447](https://github.com/calimero-network/core/issues/2447).

## Running

```bash
cargo test -p calimero-network --test relay_recovery
```

Three tests, all passing, ~40ms total.

## Verification on master

The whole suite:

```
test result: ok. 12 passed; 0 failed; 0 ignored (discovery/state_tests + request_blob)
test result: ok. 1 passed; 0 failed (gossipsub_group_topic)
test result: ok. 1 passed; 0 failed (kad_modes)
test result: ok. 3 passed; 0 failed (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean (the pre-existing `clippy::no_effect` in `calimero-utils-actix` is unrelated and was on master before this change).

## Related

- [#2446](https://github.com/calimero-network/core/pull/2446) — relay-reservation recovery (this PR sits on top of)
- [#2447](https://github.com/calimero-network/core/issues/2447) — network resilience test plan
- [merobox#246](https://github.com/calimero-network/merobox/issues/246) — fault-injection primitives for the scenarios not testable here
- [boot-node#37](https://github.com/calimero-network/boot-node/issues/37) — long-term: fold boot-node into core so we can drop the mock

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are confined to `crates/network` test code, adding an in-process libp2p relay and new integration tests without touching production networking logic.
> 
> **Overview**
> Adds a reusable `MockRelay` testkit that runs an in-process libp2p relay server with configurable quotas/durations, a `disconnect_peer` fault-injection hook, and observable counters for accepted/denied reservations.
> 
> Introduces `relay_recovery` integration tests that drive the production `Behaviour` against the mock relay to assert the key libp2p event triggers used by reservation recovery: successful reservation (`ExternalAddrConfirmed`), relay-side disconnect causing client `ConnectionClosed` with `num_established == 0`, and quota exhaustion causing a client `ListenerClosed` with an error reason.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77e4ac1a2d47808c45c70c700aa2d12a2499202e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/network/tests/common/mock_relay.rs
+++ b/crates/network/tests/common/mock_relay.rs
@@ -241,17 +241,21 @@ impl MockRelay {
 
     /// Forcibly close any active connections to `peer`. Returns true if at
     /// least one connection was found.
+    ///
+    /// Panics if the relay task is no longer running — a test that calls
+    /// this expects the relay to be alive, and an Err on the command
+    /// channel or a dropped ack would otherwise be conflated with the
+    /// semantic "peer not connected" return, producing misleading
+    /// assertion messages at call sites.
     pub async fn disconnect_peer(&self, peer: PeerId) -> bool {
         let (ack_tx, ack_rx) = oneshot::channel();
-        if self
-            .cmd_tx
+        self.cmd_tx
             .send(Command::DisconnectPeer { peer, ack: ack_tx })
             .await
-            .is_err()
-        {
-            return false;
-        }
-        ack_rx.await.unwrap_or(false)
+            .expect("mock relay task is not running (cmd channel closed)");
+        ack_rx
+            .await
+            .expect("mock relay task dropped the disconnect ack (likely panicked)")
     }
 
     /// Snapshot of what the relay has observed since spawn. Reads shared
--- a/crates/network/tests/common/mock_relay.rs
+++ b/crates/network/tests/common/mock_relay.rs
@@ -1,0 +1,326 @@
+//! In-process libp2p relay server with controllable configuration and
+//! fault-injection hooks. Used by integration tests to exercise the
+//! reservation lifecycle (request, accept, expire, renew, deny) without
+//! depending on a deployed boot-node.
+//!
+//! The mock runs a real `libp2p::relay::Behaviour` (server side) plus
+//! `identify` so a client speaking the production [`calimero_network`]
+//! behaviour will discover the hop protocol and request a reservation
+//! exactly as it would against the real boot-node.
+//!
+//! ## Fault injection
+//!
+//! - [`MockRelay::disconnect_peer`] forcibly closes any active connection to
+//!   a client. From the client's perspective, this looks the same as the
+//!   relay process crashing or its host going down (the symptom behind the
+//!   "node disconnects, can't reconnect without restart" report).
+//!
+//! - [`MockRelayConfig::max_reservations`] and
+//!   [`MockRelayConfig::max_reservations_per_peer`] cap how many concurrent
+//!   reservations the server will accept. Used to exercise the "relay quota
+//!   exhausted" failure shape — additional clients see `ListenerClosed`
+//!   with a `RESOURCE_LIMIT_EXCEEDED`-flavoured reason.
+//!
+//! - [`MockRelayConfig::reservation_duration`] makes the relay server expire
+//!   each reservation after the given duration. Mostly useful for future
+//!   tests that need to observe natural expiry; the libp2p relay client
+//!   auto-renews aggressively so this alone is not enough to drive a client
+//!   into the `ExternalAddrExpired` path — the relay would also need to
+//!   actively refuse renewal, which requires more infrastructure than this
+//!   mock currently provides.
+
+use core::time::Duration;
+use std::sync::Arc;
+
+use eyre::{eyre, Result};
+use futures_util::StreamExt;
+use libp2p::core::transport::ListenerId;
+use libp2p::identity::Keypair;
+use libp2p::swarm::{NetworkBehaviour, SwarmEvent};
+use libp2p::{identify, noise, relay, tcp, yamux, Multiaddr, PeerId, Swarm, SwarmBuilder};
+use multiaddr::Protocol;
+use tokio::sync::{mpsc, oneshot, Mutex};
+use tokio::task::JoinHandle;
+use tracing::{debug, warn};
+
+/// Identify protocol version advertised by the mock. Distinct from the real
+/// boot-node string so logs make the source obvious.
+const MOCK_IDENTIFY_PROTOCOL_VERSION: &str = "/calimero/mock-relay/1.0.0";
+const MOCK_IDENTIFY_AGENT_VERSION: &str = "calimero-mock-relay/test";
+
+#[derive(NetworkBehaviour)]
+struct MockBehaviour {
+    relay: relay::Behaviour,
+    identify: identify::Behaviour,
+}
+
+/// Tunable parameters for the mock. Defaults track the libp2p-relay 0.21
+/// defaults.
+#[derive(Clone, Debug)]
+pub struct MockRelayConfig {
+    pub max_reservations: usize,
+    pub max_reservations_per_peer: usize,
+    /// How long each reservation lasts before the server expires it.
+    /// Production boot-node uses 1 hour; tests typically want hundreds of
+    /// milliseconds so they don't have to wait.
+    pub reservation_duration: Duration,
+    pub max_circuits: usize,
+    pub max_circuits_per_peer: usize,
+    pub max_circuit_duration: Duration,
+    pub max_circuit_bytes: u64,
+}
+
+impl Default for MockRelayConfig {
+    fn default() -> Self {
+        Self {
+            max_reservations: 128,
+            max_reservations_per_peer: 4,
+            reservation_duration: Duration::from_secs(3600),
+            max_circuits: 1024,
+            max_circuits_per_peer: 16,
+            max_circuit_duration: Duration::from_secs(3600),
+            max_circuit_bytes: 1 << 30,
+        }
+    }
+}
+
+impl MockRelayConfig {
+    fn into_relay_config(self) -> relay::Config {
+        relay::Config {
+            max_reservations: self.max_reservations,
+            max_reservations_per_peer: self.max_reservations_per_peer,
+            reservation_duration: self.reservation_duration,
+            reservation_rate_limiters: vec![],
+            max_circuits: self.max_circuits,
+            max_circuits_per_peer: self.max_circuits_per_peer,
+            max_circuit_duration: self.max_circuit_duration,
+            max_circuit_bytes: self.max_circuit_bytes,
+            circuit_src_rate_limiters: vec![],
+        }
+    }
+}
+
+/// Snapshot of events the mock observed. Useful for assertions about what
+/// the relay actually saw the client do (independent of what the client
+/// thinks happened).
+#[derive(Clone, Debug, Default)]
+pub struct MockRelayObservations {
+    pub reservations_accepted: usize,
+    pub reservations_denied: usize,
+    pub circuits_opened: usize,
+}
+
+/// Commands the mock task accepts on its control channel.
+enum Command {
+    DisconnectPeer {
+        peer: PeerId,
+        ack: oneshot::Sender<bool>,
+    },
+    Observations {
+        ack: oneshot::Sender<MockRelayObservations>,
+    },
+    Shutdown {
+        ack: oneshot::Sender<()>,
+    },
+}
+
+/// Handle to a running mock relay server. Drop or call [`Self::shutdown`]
+/// to stop it.
+pub struct MockRelay {
+    peer_id: PeerId,
+    listen_addr: Multiaddr,
+    cmd_tx: mpsc::Sender<Command>,
+    join: Mutex<Option<JoinHandle<()>>>,
+}
+
+impl MockRelay {
+    /// Spawn a mock relay with default config, listening on a random local
+    /// TCP port. Returns once the listener has reported a concrete address.
+    pub async fn spawn() -> Result<Self> {
+        Self::spawn_with(MockRelayConfig::default(), Keypair::generate_ed25519()).await
+    }
+
+    /// Spawn with a custom config and identity. Tests that need to "restart"
+    /// the relay should keep the keypair around and reuse it.
+    pub async fn spawn_with(config: MockRelayConfig, keypair: Keypair) -> Result<Self> {
+        let peer_id = PeerId::from(keypair.public());
+        let relay_config = config.into_relay_config();
+
+        let mut swarm = SwarmBuilder::with_existing_identity(keypair)
+            .with_tokio()
+            .with_tcp(
+                tcp::Config::default(),
+                noise::Config::new,
+                yamux::Config::default,
+            )?
+            .with_behaviour(|key| MockBehaviour {
+                relay: relay::Behaviour::new(peer_id, relay_config),
+                identify: identify::Behaviour::new(
+                    identify::Config::new(MOCK_IDENTIFY_PROTOCOL_VERSION.to_owned(), key.public())
+                        .with_agent_version(MOCK_IDENTIFY_AGENT_VERSION.to_owned()),
+                ),
+            })
+            .map_err(|err| eyre!("failed to build mock relay behaviour: {err}"))?
+            .build();
+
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed all four review points in commit `05e96a22` + rebased onto the latest #2446 which already carries the duplicate-listener fix.

| # | Comment | Resolution |
|---|---|---|
| 1 | TOCTOU on `ephemeral_tcp_addr` | Replaced with `ephemeral_listen_addr()` that passes `/ip4/127.0.0.1/tcp/0` directly to `libp2p::listen_on`. No more bind/drop/return race. |
| 2 | `tokio::select!` without `biased;` | Added `biased;` to `MockRelay::run_swarm` so control commands have priority over swarm events. |
| 3 | Quota test could match the error more tightly | Added a comment explaining why we don't pattern-match the libp2p error variant (wrapped several `Either` layers deep, brittle to stringify-match) and gated on listener_id + `reason.is_err()` instead. Now logs the actual reason on failure for triage. |
| 4 | "Recovery idempotency defeated by status reset" | This was the same bug Cursor flagged on #2446 — already fixed there in commit `a60f4498`. Rebasing #2454 picks it up; the recovery path now no-ops on `Requested → Expired` transitions so the burst can't queue duplicate `listen_on`s. |

Full test suite still green:
- 12 unit tests in `discovery::state::tests`
- 3 integration tests in `relay_recovery`
- 1 each in `gossipsub_group_topic` and `kad_modes`

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean (one warning is the pre-existing `clippy::no_effect` in `calimero-utils-actix` on master, unrelated).

chefsale:
Second review pass addressed in commit `c3b60d3b`. Detailed mapping below.

## Stale (resolved on master or by prior commits)

| ID | Finding | Resolution |
|---|---|---|
| `b7373ffe` | TOCTOU on `ephemeral_tcp_addr` | Already fixed in `e40c76c9` (`ephemeral_listen_addr()` passes `/ip4/127.0.0.1/tcp/0` directly to libp2p). |
| `211c51c5` | `tokio::select!` without `biased;` | Already fixed in `e40c76c9`. |
| `2304c7a0` (cursor) | "Recovery idempotency defeated by status reset" | Already fixed on master in #2446 via the prior-status gate (`Accepted → Expired` queues, `Requested → Expired` does not). |

## Addressed in this commit

| ID | Finding | Resolution |
|---|---|---|
| `3b48bbe7` | Quota test should assert `obs.reservations_denied >= 1` | Done. New assertion includes a comment explaining why this is the authoritative signal. |
| `b76be026`, `731856b8` | Client A's swarm not driven during quota test → connection could time out and free the slot | Done. Client A's swarm runs in a background `tokio::spawn`; aborted after the assertion. |
| `8d33d135`, `74f1a95e`, `28058b37` | `saw_close` flag is redundant given `result.is_ok()` | Done. Flag removed; `result.is_ok()` is now the only assertion. |
| `02da9c91`, `2fc6edbd` | `spawn_with` startup loop swallows `ListenerError` / `ListenerClosed` before `NewListenAddr` | Done. Explicit match arms return descriptive errors. |
| `8fd6813d`, `a5414ee4`, `657da0b5`, `aa1ab8bb` | `Drop` impl uses `try_lock` and silently skips abort on contention | Done. Stored `AbortHandle` separately from the `Mutex<Option<JoinHandle>>`. `Drop::drop` calls `self.abort.abort()` unconditionally. |
| `60c1af0f` | `wait_for` panic on closed stream doesn't include label | Done. Replaced `.expect("swarm stream closed")` with `unwrap_or_else(|| panic!("swarm stream closed while waiting for: {label}"))`. |
| `bcc0fa00` | Client A events silently discarded in select | Resolved by moving A onto its own driver task (above) — the task panics on stream closure with a clear message. |

## Deliberately deferred + tracked

| ID | Finding | Why deferred |
|---|---|---|
| `4c2d83ee` | `ListenerClosed` with empty `addresses` doesn't exercise production handler — test could pass while production recovery doesn't fire | This is a real coverage gap in the *production* `ListenerClosed` handler, not in the test. The handler iterates `addresses` and skips if empty. The right fix is to track `ListenerId → relay_peer` in `DiscoveryState` so the handler can look up the relay peer by id when `addresses` is empty (quota-denied-before-address case). Tracked separately in [#2456](https://github.com/calimero-network/core/issues/2456) — needs careful design and its own PR; out of scope for this testkit-only PR. |
| `09a3c9c0` | Quota test could pattern-match the specific libp2p error variant | The new `reservations_denied` assertion gives the same guarantee from the relay side. Pattern-matching the libp2p error variant directly is brittle (the error type is wrapped several `Either` layers deep); the relay-side counter is more robust. |
| `a02dbe93`, `8260dd60`, `4720b27f` | `disconnect_peer_id` returning `Ok` doesn't guarantee a connection existed / was closed | Test that depends on this asserts `disconnected == true`, which the current libp2p `disconnect_peer_id` semantics make a best-effort signal. The downstream wait for `ConnectionClosed` on the client side is what actually proves the disconnect happened. Adding a relay-side `ConnectionClosed` wait before ack'ing would be more rigorous but adds complexity proportional to its value; happy to revisit if this becomes a flake source. |

## Verification

```
test result: ok. 12 passed; 0 failed; 0 ignored  (network unit tests)
test result: ok. 1 passed; 0 failed; 0 ignored  (gossipsub_group_topic)
test result: ok. 1 passed; 0 failed; 0 ignored  (kad_modes)
test result: ok. 3 passed; 0 failed; 0 ignored  (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean (the only warning is the pre-existing `clippy::no_effect` in `calimero-utils-actix` on master, unrelated).

chefsale:
Third review pass addressed in commit `02d86c71`. Mapping below.

## Addressed in this commit

| ID | Finding | Resolution |
|---|---|---|
| `590c07b0` | `Observations` command holds tokio Mutex across `.await` inside `select!` | Refactored: `observations` is now `Arc<std::sync::Mutex<MockRelayObservations>>` shared directly with the swarm task. Synchronous `.lock()` in the swarm arms (single integer increment, never held across an await). Removed the `Command::Observations` roundtrip entirely. |
| `7de668c0` | `observations()` returns stale defaults on task-down race | Reading shared state directly removes the race. No task roundtrip, so no failure mode where the ack never arrives. `observations()` is now sync and infallible. |
| `808689714f` + `0b2d5d2a` | `client_a_driver` panics with `.expect(...)` on stream close, masking real failure | Driver now loops `while client_a.next().await.is_some() {}` — silently exits on stream close. If A's swarm dies during the test, the downstream `obs.reservations_denied >= 1` assertion catches it with a clearer message. |
| `14cc2e4b` | `client_a_driver.abort()` not awaited; driver panics dropped | Added `if let Err(err) = client_a_driver.await { if !err.is_cancelled() { panic!(...) } }` — real panics surface, the expected `JoinError::Cancelled` from `abort()` is filtered out. |
| `3878380838` | `shutdown()` silently swallows panic from relay task | Replaced `let _ = handle.await;` with `if let Err(err) = handle.await { eprintln!("mock relay task did not exit cleanly: {err:?}"); }`. |

## Deferred (low ROI or tracked elsewhere)

| ID | Finding | Why |
|---|---|---|
| `b57b4747` | `wait_for` could include last-N events in panic message | Marginal. The label is already in the message, and a stuck swarm is rare enough that interactive triage is fine when it happens. Would add API surface (event-history buffer per call) without clear payoff. |
| `4c2d83ee` (from prior round) | `ListenerClosed` with empty `addresses` doesn't fire production handler | Real production gap, tracked in [#2456](https://github.com/calimero-network/core/issues/2456). Out of scope for this testkit PR; needs proper `ListenerId → relay_peer` tracking in `DiscoveryState`. |

## Verification

```
test result: ok. 12 passed; 0 failed (network unit)
test result: ok. 1 passed; 0 failed (gossipsub_group_topic)
test result: ok. 1 passed; 0 failed (kad_modes)
test result: ok. 3 passed; 0 failed (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean (only the pre-existing `clippy::no_effect` in `calimero-utils-actix`).

chefsale:
Fourth-round review: only one genuinely new finding since `02d86c71`. The other entries in the bot's commit-scoped view are stale (from earlier review rounds, already resolved — GitHub keeps them associated with the file/line even after the underlying code is rewritten).

## Addressed in `bf64f037`

| ID | Finding | Resolution |
|---|---|---|
| `187859261008` | `disconnect_peer` conflates "task dead" with "peer not connected", producing misleading assertion messages | Split: channel-send and ack-receive failures now `.expect(...)` with distinctive messages so a dead relay task surfaces as a clear "mock relay task is not running" panic, not as an ambiguous `false`. The `bool` return value represents only the semantic "was peer connected" question. |

## Already addressed in earlier rounds (bot has not refreshed)

These all appear in the bot's view of commit `02d86c71` but were resolved in `e40c76c9`, `c3b60d3b`, or this commit's predecessor `02d86c71`:

- `09a3c9c0`, `3b48bbe7`, `4c2d83ee` — quota test error matching → resolved by adding `obs.reservations_denied >= 1` assertion (#2456 tracks the production-side gap).
- `28058b37`, `8d33d135` — `saw_close` redundancy / TOCTOU → removed in `c3b60d3b`.
- `a5414ee4`, `657da0b5`, `aa1ab8bb` — `Drop` impl's `try_lock` → fixed in `c3b60d3b` by storing `AbortHandle` separately.
- `8260dd60`, `4720b27f` — `disconnect_peer_id` semantics → corroborated by `obs.reservations_denied` assertion; the bool is best-effort and now panics rather than lies if the task is gone (this PR).
- `b76be026`, `731856b8`, `808689714f`, `0b2d5d2a`, `bcc0fa00` — client A driver task issues → resolved by spawning client A as a background driver in `c3b60d3b`, then refining graceful close in `02d86c71`.
- `02da9c91`, `2fc6edbd` — `spawn_with` startup loop catching `ListenerError`/`ListenerClosed` → fixed in `c3b60d3b`.
- `60c1af0f` — `wait_for` panic label → fixed in `c3b60d3b`.
- `14cc2e4b` — await driver after abort → fixed in `02d86c71`.
- `7de668c0` — `observations()` returning defaults on race → refactored to direct `Arc<StdMutex>` access in `02d86c71`.
- `3878380838` — `shutdown()` swallowing JoinError → fixed in `02d86c71`.
- `b57b4747` — `wait_for` collecting event history → deferred (low ROI).

## Verification

```
test result: ok. 12 passed; 0 failed (network unit)
test result: ok. 1 passed; 0 failed (gossipsub_group_topic)
test result: ok. 1 passed; 0 failed (kad_modes)
test result: ok. 3 passed; 0 failed (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean.

chefsale:
Fifth review pass addressed in commit `77e4ac1a`.

## Addressed in this commit

| ID | Finding | Resolution |
|---|---|---|
| `9f045fbe` | `spawn_with` bind-wait loop has no timeout; could hang CI | Wrapped in `tokio::time::timeout(BIND_TIMEOUT, ...)` with a 10-second cap and descriptive error message. |
| `0133e953` | `shutdown()` `eprintln`s `JoinError` instead of failing the test | Replaced with explicit match: `Ok` and `JoinError::is_cancelled` stay silent; `is_panic` propagates as `panic!`. Test runners cannot swallow it. |
| `79c814cf` | Quota test ignores graceful `ListenerClosed { reason: Ok(()) }` | Split the match into err/ok arms: error returns success; graceful close panics with a descriptive message explaining the expected failure mode. |
| `25f131a1` + `4d87d0d7` | Driver cleanup runs before `result.is_ok()` assertion, masking root cause on real failures | Reordered: assert `result.is_ok()` first (its value is already computed by `timeout`), then unwind `client_a_driver`, then check observations. |
| `1db6f60c` | `Drop` impl doesn't join the task; panics after abort silently lost | Documented inline that this is intentional (Drop can't await) and `shutdown()` is the authoritative path that surfaces task panics. No code change. |
| `1afc4d3b` | Two `use std::sync::*` lines should be merged | Combined into `use std::sync::{Arc, Mutex as StdMutex};`. |

## Verification

```
test result: ok. 12 passed (network unit)
test result: ok. 1 passed (gossipsub_group_topic)
test result: ok. 1 passed (kad_modes)
test result: ok. 3 passed (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean.

PR #2462, opened 2026-05-23
fix(network): track relay reservations by ListenerId for ListenerClosed recovery

### LINKED ISSUE #2456: Production ListenerClosed handler misses relay reservation losses with empty addresses
## Background

[PR #2454](https://github.com/calimero-network/core/pull/2454) added a `MockRelay` integration test (`relay_quota_exhaustion_denies_second_client`) that exercises the relay-side denial path. Review surfaced a real coverage gap in the *production* `ListenerClosed` handler that the mock test masks: the handler can only react when the closed listener has at least one relayed multiaddr in its `addresses` field.

## Current production handler

`crates/network/src/handlers/stream/swarm.rs` — `SwarmEvent::ListenerClosed`:

```rust
SwarmEvent::ListenerClosed { addresses, reason, .. } => {
    trace!("Listener closed: {:?} {:?}", addresses, reason.err());

    for address in &addresses {
        if let Ok(relayed_addr) = RelayedMultiaddr::try_from(address) {
            let actions = self
                .discovery
                .state
                .on_relay_reservation_lost(relayed_addr.relay_peer_id());
            self.execute_reachability_actions(actions);
        }
    }
}
```

Two failure modes are conflated:

1. **Listener died with at least one relayed address still attached** — the loop finds the relay peer, recovery fires. This is the case `relay_quota_exhaustion_denies_second_client` exercises.

2. **Listener died with `addresses: []` (e.g. reservation denied before any external address was confirmed)** — the loop is empty, no recovery action is queued, and the failed `Requested` reservation stays `Requested` in our state forever (because `ExternalAddrConfirmed` never fired to advance it, and `is_relay_reservation_required` counts `Requested` as "we have one"). Any subsequent re-evaluation skips the relay.

In live libp2p-relay 0.21, mode 2 fires when the relay sends a `Status::RESOURCE_LIMIT_EXCEEDED` *before* the reservation has been associated with a confirmed external address — exactly the renewal-time quota wall described as scenario S6 in [#2447](https://github.com/calimero-network/core/issues/2447).

## Why the integration test doesn't expose this

The current test asserts `ListenerClosed` fires for the right `listener_id` with `reason.is_err()`. That assertion still passes when `addresses` is empty — the test is checking the libp2p event mechanic, not the production handler's downstream effect. The assertion of `obs.reservations_denied >= 1` confirms the relay's view, but not the client's reaction.

## Proposed fix

Track which libp2p `ListenerId` corresponds to which relay peer at the point of `create_relay_reservation`, so `ListenerClosed` can look up the relay peer regardless of `addresses` content.

Sketch:

```rust
// In PeerRelayInfo (or DiscoveryState):
pub(crate) struct PeerRelayInfo {
    reservation_status: RelayReservationStatus,
    listener_id: Option<ListenerId>,  // NEW
}

// In create_relay_reservation:
let listener_id = self.swarm.listen_on(relayed_addr)?;
self.discovery.state.set_relay_listener_id(relay_peer, Some(listener_id));
self.discovery.state.update_relay_reservation_status(relay_peer, Requested);

// In ListenerClosed handler:
if let Some(relay_peer) = self.discovery.state.relay_peer_for_listener(listener_id) {
    let actions = self.discovery.state.on_relay_reservation_lost(&relay_peer);
    self.execute_reachability_actions(actions);
}
// Fallback: existing addresses-based loop, in case the listener was set up
// outside our state machine.
```

This also lets us tighten `on_relay_reservation_lost` further by passing the `listener_id` to it and ignoring stale events whose id doesn't match — a cleaner solution to the "burst" race that #2446 currently handles via the prior-status gate.

## Acceptance criteria

- [ ] `DiscoveryState` (or `PeerRelayInfo`) tracks `ListenerId → relay_peer` for active reservations.
- [ ] `create_relay_reservation` stores the listener id after a successful `listen_on`.
- [ ] `ListenerClosed` handler looks up the relay peer by listener id; fires recovery even when `addresses` is empty.
- [ ] Unit test: `set_relay_listener_id` + lookup works.
- [ ] Integration test in `relay_recovery.rs` (or follow-up to it): a quota-denied reservation produces a fresh `Requested → ` re-evaluation chain in production state, not just `Requested` stuck forever.

## Out of scope

- Plumbing listener id through `ExternalAddrExpired` and `ConnectionClosed` (those don't carry listener_id and need different treatment).
- Backoff for repeated quota denials against the same relay (separate concern).

## Related

- [#2446](https://github.com/calimero-network/core/pull/2446) — recovery code this would refine
- [#2454](https://github.com/calimero-network/core/pull/2454) — integration tests that flagged the gap
- [#2447](https://github.com/calimero-network/core/issues/2447) — broader test plan

chefsale:
## What

Closes #2456. Tracks relay reservations by `ListenerId` so the `ListenerClosed` handler can route recovery even when the closed listener's `addresses` field is empty.

## The bug

In `handlers/stream/swarm.rs`, the `ListenerClosed` handler iterates over the closed listener's `addresses`:

```rust
for address in &addresses {
    if let Ok(relayed_addr) = RelayedMultiaddr::try_from(address) {
        let actions = self.discovery.state.on_relay_reservation_lost(...);
        self.execute_reachability_actions(actions);
    }
}
```

When the relay denies a reservation **before** an external address is allocated — most commonly at renewal time when the quota is full, but also rate-limit hits and validation rejections — libp2p emits `ListenerClosed { addresses: [], reason: Err(...) }`. The for-loop runs zero times. `on_relay_reservation_lost` never fires. The peer's reservation status stays `Requested` in `DiscoveryState`. `is_relay_reservation_required` keeps counting it as active. The system never retries until process restart.

Surfaced by [meroreviewer finding `4c2d83ee`](https://github.com/calimero-network/core/pull/2454#pullrequestreview-...) during #2454 review, filed as #2456.

## The fix

Three new `pub(crate)` methods on `DiscoveryState`:

- `record_relay_listener(ListenerId, PeerId)` — called by `create_relay_reservation` after a successful `listen_on`
- `relay_peer_for_listener(&ListenerId) -> Option<PeerId>` — called by the `ListenerClosed` handler
- `forget_relay_listener(&ListenerId)` — called once the recovery action has been queued, so the map doesn't grow

Backed by a new `relay_listeners: HashMap<ListenerId, PeerId>` field in `DiscoveryState` (using `HashMap` because libp2p's `ListenerId` does not implement `Ord`).

`create_relay_reservation` records the listener id immediately after `listen_on`:

```rust
let listener_id = self.swarm.listen_on(relayed_addr)?;
self.discovery.state.record_relay_listener(listener_id, *relay_peer);
self.discovery.state.update_relay_reservation_status(relay_peer, RelayReservationStatus::Requested);
```

The `ListenerClosed` handler prefers the new lookup; falls back to the existing addresses-iteration path for listeners opened outside `create_relay_reservation` (defensive — covers any future code path that opens a relayed listener elsewhere):

```rust
if let Some(relay_peer) = self.discovery.state.relay_peer_for_listener(&listener_id) {
    self.discovery.state.forget_relay_listener(&listener_id);
    let actions = self.discovery.state.on_relay_reservation_lost(&relay_peer);
    self.execute_reachability_actions(actions);
} else {
    // existing addresses-iteration fallback
}
```

## Tests

Five new unit tests in `crates/network/src/discovery/state_tests.rs`:

- `test_record_and_lookup_relay_listener` — happy path
- `test_record_relay_listener_overwrites_existing_entry` — re-registering the same id wins-latest
- `test_forget_relay_listener_removes_entry` — forget + re-lookup returns None
- `test_forget_relay_listener_is_noop_for_unknown_id` — defensive
- `test_multiple_listeners_map_to_distinct_relays` — independent tracking

```
test result: ok. 23 passed; 0 failed (network unit)
test result: ok. 1 passed; 0 failed (gossipsub_group_topic)
test result: ok. 1 passed; 0 failed (kad_modes)
test result: ok. 3 passed; 0 failed (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean.

## Drive-by tidy

Moves the `DIAL_FAILURE_EVICTION_THRESHOLD` const (added in #2459) below the libp2p imports to match the placement of `RENDEZVOUS_PROTOCOL_NAME`. The previous placement was between two `use` groups, which is inconsistent with how the file otherwise orders items.

## Why no integration test in this PR

A genuine integration test for this path would need the `MockRelay` testkit to spawn a relay configured to deny a reservation before any external address is allocated (i.e. set `max_reservations: 0` or trigger a rate limiter immediately), then drive the production `NetworkManager` actor and assert on its state. That requires the `NetworkManager` actor harness that was deferred during #2454, plus a `MockRelay` extension. Both are out of scope here.

The unit tests cover the new state-machine surface. The handler wiring is a single match arm + method call — the same shape the rest of the recovery code uses, all of which is exercised by the existing `MockRelay` integration tests.

## Related

- #2446 — relay reservation recovery (this PR closes a gap in its `ListenerClosed` branch)
- #2454 — MockRelay testkit (where the gap was surfaced)
- #2456 — this issue
- #2459 — address-book hygiene (previous follow-up)

---

> [!NOTE]
> **Medium Risk**
> Touches relay reservation recovery in the libp2p swarm event handler; mis-tracking `ListenerId`→relay mappings could cause missed retries or unnecessary re-reservations, but the change is localized and unit-tested.
> 
> **Overview**
> Fixes a relay-reservation recovery gap by **tracking relayed listeners by `ListenerId`** so `SwarmEvent::ListenerClosed` can trigger `on_relay_reservation_lost` even when libp2p reports `addresses: []` (e.g. reservation denied before any external address is allocated).
> 
> `create_relay_reservation` now records the `ListenerId` returned by `listen_on`, `DiscoveryState` gains a `relay_listeners` map with `record_relay_listener`/`take_relay_listener`, and the `ListenerClosed` handler prefers this mapping with an addresses-based fallback; new unit tests cover the listener mapping behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc57ecfc86af0562793975667195e24562aa71fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/network/src/discovery.rs
+++ b/crates/network/src/discovery.rs
@@ -309,7 +309,15 @@ impl NetworkManager {
             }
         };
 
-        let _ = self.swarm.listen_on(relayed_addr)?;
+        let listener_id = self.swarm.listen_on(relayed_addr)?;
+
+        // Record the listener id so the ListenerClosed handler can route
+        // recovery back to this relay peer even if the close event comes
+        // with an empty `addresses` list (the quota-denied-before-address-
+        // allocation case).
+        self.discovery
+            .state
+            .record_relay_listener(listener_id, *relay_peer);
 
         self.discovery
             .state
--- a/crates/network/src/discovery/state.rs
+++ b/crates/network/src/discovery/state.rs
@@ -7,13 +7,7 @@ use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::time::Instant;
 
-/// Consecutive-failure threshold at which an address is evicted from a
-/// peer's address book. Three is chosen as a small magic number that
-/// tolerates transient flakiness (one bad TCP retransmit, one identify
-/// race) without keeping a permanently broken address around long enough
-/// to waste many rendezvous-tick dial attempts.
-pub(crate) const DIAL_FAILURE_EVICTION_THRESHOLD: u8 = 3;
-
+use libp2p::core::transport::ListenerId;
 use libp2p::relay::HOP_PROTOCOL_NAME;
 use libp2p::rendezvous::Cookie;
 use libp2p::{Multiaddr, PeerId, StreamProtocol};
@@ -23,6 +17,13 @@ use tracing::info;
 // The rendezvous protocol name is not public in libp2p, so we have to define it here.
 // source: https://github.com/libp2p/rust-libp2p/blob/a8888a7978f08ec9b8762207bf166193bf312b94/protocols/rendezvous/src/lib.rs#L50C12-L50C92
 const RENDEZVOUS_PROTOCOL_NAME: StreamProtocol = StreamProtocol::new("/rendezvous/1.0.0");
+
+/// Consecutive-failure threshold at which an address is evicted from a
+/// peer's address book. Three is chosen as a small magic number that
+/// tolerates transient flakiness (one bad TCP retransmit, one identify
+/// race) without keeping a permanently broken address around long enough
+/// to waste many rendezvous-tick dial attempts.
+pub(crate) const DIAL_FAILURE_EVICTION_THRESHOLD: u8 = 3;
 
 /// DiscoveryState is a struct that holds the state of the disovered peers.
 /// It holds the relay and rendezvous indexes to quickly check if a peer is a relay or rendezvous.
@@ -34,6 +35,14 @@ pub struct DiscoveryState {
     rendezvous_index: BTreeSet<PeerId>,
     autonat_index: BTreeSet<PeerId>,
     confirmed_external_addresses: HashSet<Multiaddr>,
+    /// Maps each libp2p relayed listener back to the relay peer it was
+    /// opened against. Populated by `create_relay_reservation` when it
+    /// calls `listen_on(<relay>/p2p-circuit/<self>)` and gets back a
+    /// `ListenerId`. Looked up by the `ListenerClosed` swarm handler so
+    /// it can route the recovery action even when the closed listener's
+    /// `addresses` list is empty (e.g. quota denial before address
+    /// allocation).
+    relay_listeners: HashMap<ListenerId, PeerId>,
     reachability_state: ReachabilityState,
 }
 
@@ -52,6 +61,7 @@ impl Default for DiscoveryState {
             rendezvous_index: BTreeSet::new(),
             autonat_index: BTreeSet::new(),
             confirmed_external_addresses: HashSet::new(),
+            relay_listeners: HashMap::new(),
             reachability_state: ReachabilityState::Unknown,
         }
     }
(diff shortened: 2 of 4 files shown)


chefsale:
Both findings (`61a2ddf7` warning and `d48a5b84` suggestion) point at the same real issue: with two split methods (`relay_peer_for_listener` for lookup + `forget_relay_listener` for cleanup), the `ListenerClosed` handler could leak `relay_listeners` entries on any code path that called the lookup but skipped the forget. While the current handler always calls both inside the `if let Some(...)` branch, the API shape allows misuse.

Fixed in commit `5b16097f` by collapsing both methods into a single `take_relay_listener` that removes the entry as it returns the peer. The handler now has no way to leak entries:

```rust
if let Some(relay_peer) = self.discovery.state.take_relay_listener(&listener_id) {
    let actions = self.discovery.state.on_relay_reservation_lost(&relay_peer);
    self.execute_reachability_actions(actions);
} else {
    // addresses-iteration fallback for listeners we never registered;
    // map had no entry to clean by definition
}
```

The split-methods API had one pair of correctness-coupled calls (look up, then forget); `take` removes that coupling. Bot's suggestion of "call forget unconditionally" would also work but adds a redundant mutation when the lookup misses.

Unit tests updated for the new shape:

- `take_relay_listener_returns_recorded_peer_and_removes_entry` — first take returns the peer, second returns None (cleanup verified)
- `record_relay_listener_overwrites_existing_entry` — wins-latest semantics preserved
- `take_relay_listener_is_noop_for_unknown_id` — no-op on miss
- `multiple_listeners_map_to_distinct_relays` — independent tracking across takes

```
test result: ok. 22 passed; 0 failed (network unit)
test result: ok. 1 passed; 0 failed (gossipsub_group_topic)
test result: ok. 1 passed; 0 failed (kad_modes)
test result: ok. 3 passed; 0 failed (relay_recovery)
```

`cargo fmt --check` clean. `cargo clippy -p calimero-network --tests` clean.

(The other two comments on this PR — both marked "✅ No longer detected" by meroreviewer — confirm that the prior-round findings are already resolved by the original commit `12983bb7`.)

PR #1, opened 2024-01-22
feat(CLI): initial project setup

frdomovic:
# feat(CLI): initial project setup

## Summary:
Initial template of CLI function

## Tests:

https://github.com/calimero-is-near/cali2.0-experimental/assets/93442516/f514fee9-9dc8-4cf2-a15f-ee367bf3f29c

PR #2466, opened 2026-05-23
test(e2e): sync-resilience smoke tests using merobox fault-injection

chefsale:
## What

Three new e2e workflows that verify the calimero stack survives the real-world fault shapes the relay-reservation recovery work in [#2446](https://github.com/calimero-network/core/pull/2446) / [#2454](https://github.com/calimero-network/core/pull/2454) / [#2459](https://github.com/calimero-network/core/pull/2459) / [#2462](https://github.com/calimero-network/core/pull/2462) was motivated by. They use the fault-injection step types added in [calimero-network/merobox#249](https://github.com/calimero-network/merobox/pull/249) (`restart_container`, `pause_container` / `unpause_container`, `inject_network_fault`).

| File | Fault | Verifies |
|---|---|---|
| `sync-resilience-peer-restart.yml` | `restart_container` on the writer | post-restart write replicates to the other peer |
| `sync-resilience-peer-pause.yml` | `pause_container` on the joiner for 60s, write on the other peer, `unpause_container` | joiner catches up to the during-pause write |
| `sync-resilience-network-blip.yml` | `inject_network_fault loss=100% duration=3s` on the joiner, then write on the other peer | post-blip write replicates |

CI wiring: three matrix entries added to `.github/workflows/e2e-rust-apps.yml`. Reuses the existing local-image build + WASM build pipeline. No new CI workflow file. The release variant (`e2e-rust-apps-release.yml`) is hand-rolled with a curated subset of 10 workflows and is intentionally left untouched.

## Honest scope (please read before reviewing)

These do **not** specifically exercise the relay-reservation recovery code shipped in #2446. In merobox's mesh model both nodes sit on the same Docker bridge and reach each other directly over sibling addresses (`/ip4/<sibling-ip>/tcp/2428/p2p/<peer_id>`) — they don't go through a relay. So `ReachabilityState` stays `Reachable`, no relay reservation is ever requested, and the recovery code never fires.

The recovery code itself is covered by:
- Unit tests in `crates/network/src/discovery/state_tests.rs` (state-machine semantics)
- MockRelay integration tests in `crates/network/tests/relay_recovery.rs` (libp2p event mechanics)

A real NAT-topology test that *genuinely* exercises the relay path end-to-end is tracked under [#2453](https://github.com/calimero-network/core/issues/2453), gated on additional merobox primitives (a NAT container image and multi-network topology declarations).

## Where these workflows DO earn their keep

The class of integration regressions that the unit-level tests can't see:

- Sync-protocol changes that break peer-restart recovery
- Gossipsub mesh re-formation bugs (mesh needs to rebuild after a peer comes back)
- Address-book staleness after a peer's IP changes on restart (with #2459's eviction now in place)
- The **Borsh-wire-format-drift** class of bug that broke `test-merobox-integration` for two weeks before [merobox#250](https://github.com/calimero-network/merobox/pull/250) caught it — exactly the kind of thing this catches in core's own CI

## Test design notes

**`mdns: false` on all three.** With mdns on, the two containers find each other via multicast in a way that bypasses anything bootstrap-related. A regression in bootstrap discovery would not show up. The workflows are explicit in their description files about this rationale.

**3-second blip duration in network-blip.** Long enough to drop in-flight TCP frames and produce visible connection-state changes. Short enough not to balloon CI time. The fault auto-clears when `duration` elapses.

**60-second pause in peer-pause.** Long enough that the relay-side TCP keepalive (typically 30s) expires and the connection is dropped — the exact thing that happens on a laptop with the lid closed past the NAT keepalive interval.

**`wait_for_sync` post-fault uses a 120s timeout** (vs 60s for baseline) to give the recovery code a wide window. If a real regression makes recovery take longer than 120s, we want to know.

## How to verify the tests actually catch what they should

For each workflow, revert one of #2446 / #2459 / #2462 locally on top of master and re-run:

```
git revert <commit>
# build merod
# run the relevant workflow
```

If the workflow still passes, the test isn't actually checking what it claims to. That's an open question — given the framing above (these are NOT relay-recovery tests, they're integration smoke tests), the tests SHOULD still pass when those PRs are reverted. They only turn red when a sync-level or transport-level regression is introduced.

## Related

- [#2446](https://github.com/calimero-network/core/pull/2446) — relay reservation recovery
- [#2454](https://github.com/calimero-network/core/pull/2454) — MockRelay testkit
- [#2459](https://github.com/calimero-network/core/pull/2459) — address-book hygiene
- [#2462](https://github.com/calimero-network/core/pull/2462) — listener-id tracking
- [#2447](https://github.com/calimero-network/core/issues/2447) — overall network-resilience test plan
- [#2453](https://github.com/calimero-network/core/issues/2453) — proper NAT-topology relay-recovery e2e (follow-up)
- [merobox#249](https://github.com/calimero-network/merobox/pull/249) — the fault-injection primitives this PR uses

---

> [!NOTE]
> **High Risk**
> Changes libp2p discovery, connection teardown, and rendezvous query timing on the post-fault recovery path used under NAT; incorrect throttling or ping closes could affect peer rediscovery and sync in production.
> 
> **Overview**
> Adds **six NAT-topology** merobox workflows (peer restart, 60s pause, LAN bridge partition) in **cone** and **symmetric** modes, wired into `e2e-rust-apps.yml`. Symmetric runs assert relay-only traffic (`/p2p-circuit/`, `ReservationReqAccepted`, no DCUtR upgrade). Partition tests use **`disconnect_node` / `connect_node`** with a **35s** hold instead of short `tc` loss so libp2p actually tears down connections.
> 
> **CI** sets `MEROBOX_PRE_RESTART_LOG_DIR`, uploads the full `docker-logs/` tree (including pre-restart snapshots), and documents deferred **lossy-link** tests (#2468).
> 
> **Network recovery (#2469):** default rendezvous **`discovery_interval` 90s → 15s**; **`rendezvous_discover(..., force)`** and **`rendezvous_discover_force`** bypass the `discovery_rpm` throttle when a **regular peer’s last connection closes**, with **5/15/30/60s** retries. **Ping watchdog** closes connections after **three** consecutive failures so silent partitions trigger the same `ConnectionClosed` recovery path. Unit tests lock in `on_regular_peer_disconnected` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fa0e11eb3ba456bb51e9475f128af2ced5c2571. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed all the real findings in `bca16099`. Triage and resolutions:

| # | Finding | Resolution |
|---|---|---|
| 1 | No wait after `restart_container` before write (×2 mentions, `bb229e9a` + `41a2d0dc`) | Added `wait: 10s` step. `wait_healthy` only confirms merod's admin endpoint; libp2p still needs to redial and rejoin the gossipsub mesh. Without the buffer, the test was masking any gossipsub-rejoin regression by relying on `wait_for_sync`'s pull-based fallback. |
| 2 | No wait after `unpause_container` before `wait_for_sync` (`6e39d61f`) | Added `wait: 10s` step. Same reasoning — paused TCP stack needs time to redial after unpause. |
| 3 | 3-second `inject_network_fault` may not break TCP (`094ba738`) + asymmetric egress-only fault (`74f45da9`) | **Replaced `inject_network_fault` with `disconnect_node` + `connect_node`**. 3s is shorter than libp2p ping / yamux keepalive (15-30s), so the connection survived as buffered retransmits — the test exercised gossipsub message loss, not connection recovery. `tc qdisc loss` is also egress-only. `disconnect_node` produces a true symmetric partition by detaching the network interface; 15s detach exceeds libp2p's keepalive so the connection is marked dead. File renamed `network-blip.yml` → `network-partition.yml`; matrix entry updated. |
| 4 | Baseline key not re-verified after fault (×3, `e6e6949d` / `c9ef00ef` / `8e4c6c58`) | Added a "read + assert baseline" step at the end of all three workflows. Catches state-corruption-on-recovery regressions. |
| 5 | "Pause duration" step name ambiguous (`0f570354`) | Renamed to "Wait while node-2 is paused (60s)". |
| — | `inject_network_fault` blocking semantics comment (`1764f25a`) | Moot — `inject_network_fault` is no longer used in the partition test. |
| — | Wait after `pause_container` before write to node-1 (`e6df04d2`) | Skipped — pause on node-2 doesn't affect node-1's RPC; the hypothetical ARP-flush race is unlikely on a Docker bridge. |

Also worth flagging: the bot's second review marked four of the first-review findings as "✅ No longer detected" before I'd pushed anything. Those are false positives in the bot's state tracking — they were the exact issues this commit actually fixes. Just noting in case the reviewer counts get confusing.

Diff: 5 files changed, +260 / -166. New file `sync-resilience-network-partition.yml`; old `sync-resilience-network-blip.yml` deleted.

YAML parses clean on all four files. Will watch the next CI run.

chefsale:
Reworked the lossy-link test in `083dcce8` to be actually meaningful instead of "1 write, 3s loss, sync." Honest constraint I want to flag because it shapes the design:

**`inject_network_fault` blocks the workflow while the fault is held.** That means writes cannot happen *during* the loss window in a sequential workflow — the workflow is suspended in the fault step. The strictest "lossy link" shape buildable today is **burst writes immediately before the fault, then hold the fault while gossipsub struggles to propagate them**.

New design:

1. **Burst-write 20 keys** on node-1 with no waiting between them via a `repeat`. All 20 land in gossipsub's publish queue and start propagating.
2. **Inject 90% loss on node-2 for 20s** — gossipsub now has 20 in-flight messages to deliver through a degraded channel for the full fault window. (Why not 100%: that's a partition, covered by `network-partition.yml`.)
3. **Diagnostic read of `burst_20`** (last-written key) on node-2 right after fault — captured + logged for visibility but **not asserted on**, because at 90%/20s the math is too close to either "all delivered" or "some missing" to assert strictly without flake.
4. **`wait_for_sync` 90s** — strict assertion that the system must converge.
5. **Read each of 20 burst keys** on node-2, assert each contains its expected value. Catches the class of bug where the DAG converges but per-key CRDT state drops events under retransmission.
6. **Re-verify baseline** survived the lossy window.

What this test now catches that the previous one didn't:

- A regression where gossipsub mesh stability under sustained loss breaks (failure: `wait_for_sync` 90s times out)
- A regression where the CRDT state machine drops events under retransmission while the DAG appears to converge (failure: one of the 20 `contains(burst_i, v_i)` asserts fails)
- A regression in TCP-survivability under moderate loss (failure: connection dies during the 20s, `wait_for_sync` times out forever)

Future enhancement to consider in merobox: a non-blocking `inject_network_fault` variant (`async: true`?) would let us actually write *during* the lossy window. The current sequential block is the right safety default but the async variant would unlock stricter divergence assertions. I'd file that as a follow-up if you want a strict "during the fault, these keys are NOT yet synced" check — but for a smoke test, the current shape is fine.

chefsale:
Pushed `4cbd9985`. Root cause was two stacked issues, not what I first thought:

1. **`mdns: false` on `nodes:` is a no-op in the current merobox release.** The schema accepts the field, but no code applies it to merod's `config.toml`. `apply_e2e_defaults` runs after init and unconditionally sets `discovery.mdns = True`, overwriting whatever the workflow declared. Verified by reading the merobox 0.6.17 source — the value is only checked, never written.

2. **`apply_e2e_defaults` also clears `bootstrap.nodes` to `[]` for test isolation.** That removes the public devnet boot-node config that `merod init` writes by default. In a 2-node cluster without the boot-node, discovery falls back to mdns + sibling-bootstrap-wiring + a fresh per-workflow rendezvous namespace, which can race the gossipsub mesh formation deadline in `join_namespace` (the error we saw: `could not open a namespace-join stream to any mesh peer ... elapsed 9012ms`).

Fix in this commit: declare the public boot-node explicitly via `bootstrap_nodes:` at the workflow level in all four files. Setup order becomes:

1. `apply_e2e_defaults` clears `bootstrap.nodes` (for isolation)
2. `apply_bootstrap_nodes` writes our public-boot-node list back
3. `_wire_cluster_bootstrap_peers` appends sibling addresses after

Final config: `[public_boot_node_quic, public_boot_node_tcp, sibling_quic, sibling_tcp]`. Stable rendezvous outside the test cluster + siblings as fallback.

Also dropped the misleading `mdns: false` lines.

Follow-up filed: [merobox#252](https://github.com/calimero-network/merobox/issues/252) proposes adding a workflow-level `preserve_default_bootstrap: true` field that would let workflows opt out of `apply_e2e_defaults`' bootstrap-clearing, so the explicit list here can go away once that lands.

chefsale:
Pushed \`61ac4073\` — Fix #1 from the triage: drop \`tail_lines: 200\` from every \`ReservationReqAccepted\` assertion across all 6 sync-resilience workflows.

**Root cause**: merod emits \`ReservationReqAccepted\` ONCE per reservation (initial reserve, \`renewal: false\`). I framed the original \`tail_lines: 200\` as "scope to recent activity so a stale setup-phase hit can't mask a failed post-fault re-acquisition" — but since renewals and post-fault re-acquisitions don't re-emit the log line, there's no stale hit to scope away from. Just "did the initial reservation land or not". The actual post-fault relay-recovery validator is the sync-converge step itself.

Confirmed by downloading the peer-pause log artifact: the reservation fires at line 147 in a log that grows to 16,329 lines by assertion time. The last 200 lines never include it.

**Symmetric variants** (3): had \`/p2p-circuit/\` + \`ReservationReqAccepted\` combined under one \`tail_lines: 200\`. \`/p2p-circuit/\` IS emitted on every relayed-connection log line and survives a tail scope (it's the real "post-fault traffic uses relay" signal). Split into two assertions per node: tail-scoped \`/p2p-circuit/\` + full-log \`ReservationReqAccepted\`.

**Fix #3 (\`network-partition\` cone reachability probe timeout)** — likely a flake, not actioned in this push. The symmetric variant of the same workflow reached boot-node on first probe with identical merobox + NAT setup, so I don't think it's a systematic regression in 0.6.20. Will know after this run.

**Fix #2 (\`peer-restart\` post-restart sync recovery: \`peer_count=0\`, 114 consecutive sync failures, \`No peers to sync with\`)** — investigating separately. The post-restart node-2 acquired its fresh reservation but never re-established gossipsub mesh with node-1. Possibly real merod relay-recovery regression specific to the container-restart path. Will open a separate issue if confirmed.

chefsale:
Pushed \`8ba9dfc5\` — comprehensive fix for #2469 (the NAT variant), closing the actual root cause.

## Root cause (validated against CI log artifact)

When a regular peer drops its last connection (the post-\`docker restart\` case), the surviving peer's address book held nothing dialable — relayed (\`/p2p-circuit/\`) addresses are intentionally NOT stored (per \`swarm.rs ConnectionEstablished\` filter, line 95-103). The only way to rediscover the peer's NEW \`/p2p-circuit/\` after their restart is to re-query rendezvous on the boot-node.

But the periodic rendezvous tick is gated by:
- \`discovery_interval = 90s\` (tick rate)  
- \`discovery_rpm = 0.5\` (= throttle floor of 120s between requests per peer)

With a 120s sync-recovery budget in the test, ZERO rendezvous re-queries fire post-disconnect. The peer parks on \`peer_count=0\` / "No peers to sync with" / \`failure_count=114\` / \`backoff_secs=256\` until the test budget exhausts.

## The fix

Adds an **event-driven force-rediscover path** that bypasses the throttle on regular-peer drops:

1. New \`DiscoveryState::on_regular_peer_disconnected()\` — returns \`ReachabilityActions\` with all known rendezvous peers queued in a new \`rendezvous_discover_force\` field.
2. New \`rendezvous_discover_force: Vec<PeerId>\` on \`ReachabilityActions\`. The existing throttled \`rendezvous_discover\` field stays — periodic-tick behavior is unchanged.
3. \`NetworkManager::rendezvous_discover()\` now takes a \`force: bool\` parameter. The 4 existing callers pass \`false\` (preserve throttle); the new dispatch for the force-queue passes \`true\` (bypass throttle).
4. Wired from the existing \`SwarmEvent::ConnectionClosed\` branch that already handles regular-peer cleanup. No changes to relay-peer or rendezvous-peer disconnect paths.

## Tests

Five new unit tests in \`discovery/state_tests.rs\`:

- happy-path single rendezvous peer
- multi-rendezvous deployment (every peer queued)
- no-op when no rendezvous peers configured (mdns-only)
- relay-only peer NOT queued
- HOP+RENDEZVOUS dual-role boot-node IS queued via rendezvous role

All 26 lib tests + 3 existing \`relay_recovery\` integration tests still pass. \`cargo fmt --check\` clean.

## Diff

\`\`\`
 crates/network/src/discovery.rs                    |  34 ++++--
 crates/network/src/discovery/state.rs              |  51 ++++++++
 crates/network/src/discovery/state_tests.rs        | 128 ++++++++++++++++++++
 crates/network/src/handlers/stream/rendezvous.rs   |   2 +-
 crates/network/src/handlers/stream/swarm.rs        |  21 +++
 crates/network/src/handlers/stream/swarm/identify.rs |   2 +-
 crates/network/src/handlers/stream/swarm/rendezvous.rs |   2 +-
 7 files changed, 230 insertions(+), 10 deletions(-)
\`\`\`

Watching CI now for the 6 sync-resilience jobs. The two that were green before (\`network-partition\` cone + symmetric) should stay green; the four that were red (\`peer-restart\` and \`peer-pause\` cone + symmetric) should now go green.

PR #2470, opened 2026-05-24
fix(storage): write Entry before updating Merkle index (close read-after-publish race)

chefsale:
## Summary

- `Interface::save_internal` and `Interface::write_pre_merged_root_state` updated the Merkle index (`Index::update_hash_for`) **before** writing the actual entry bytes (`storage_write(Key::Entry(id), ...)`).
- Between the index update and the entry write, a concurrent reader sees a converged root hash that promises content that isn't yet readable.
- `UnorderedMap::entries` does `.flatten().fuse()` on the inner iterator of `Result`s, silently dropping any `NotFound` Err from a missing `Key::Entry` lookup — so the partial child set surfaces with no error.

## Reproduction

`apps/scaffolding-e2e/workflows/e2e.yml` rga section, post-#2465 master run (`ca8f7dea54`):

1. Node-1 inserts "Hello" + appends " World" (11 chars).
2. `wait_for_sync` polls all 3 nodes' root hashes → converges on `G75KgR9...` within ~0.5s.
3. ~15ms later, `rga_get_text` on node-3 returns `"Hello Wor"` (9 chars) while node-2 returns `"Hello World"`.

Node-3 had indexed all 11 char entries (Merkle root matched peers), but 2 char `Key::Entry` writes were still in flight when the read came in. `UnorderedMap::entries` swallowed both as `NotFound` and yielded 9 of 11 chars.

The bug existed since the Merkle index was introduced; PR #2465 widened the race window (extra apply paths via `write_pre_merged_root_state`, slightly longer apply duration from the LWW guard) and surfaced it on the post-merge e2e run.

## Fix

Swap the order in both sites: `storage_write(Key::Entry(id), ...)` first, then `Index::update_hash_for(id, ...)`. `update_hash_for` only reads/writes index entries (verified at `index.rs:726-768`); it never touches `Key::Entry`, so the swap is safe.

Readers now observe either:
- Old hash + old entries (apply hasn't started or entry-write hasn't landed), or
- New hash + new entries (both writes settled),

never the inconsistent middle.

The deletion path (`save_internal` line 1448) already does entry-first; leaving that alone since the brief \"child in index, entry gone\" window iterates to the same eventually-consistent result.

## Test plan

- [x] 463 calimero-storage lib tests pass
- [x] 207 calimero-node lib tests pass
- [x] 308 sync_sim integration tests pass
- [ ] CI scaffolding-e2e green (the rga \"Hello World\" assertion that failed on `ca8f7dea54`)
- [ ] CI shared-storage / kv-store / mesh-soak green (no regression from the order swap)

---

> [!NOTE]
> **Medium Risk**
> Touches core persistence ordering on every entity save and pre-merged root commit; wrong ordering would regress the RGA e2e flake, though the change aligns read paths with hash visibility and is covered by existing lib tests.
> 
> **Overview**
> Fixes a **read-after-publish race** where `save_internal` and `write_pre_merged_root_state` updated the Merkle index before persisting `Key::Entry` bytes. After sync, a reader could see a **matching root hash** while collection iteration still dropped children whose entry blobs were not written yet (`UnorderedMap::entries` uses `flatten().fuse()`, hiding missing entries)—the scaffolding RGA flake (`"Hello Wor"` vs `"Hello World"`).
> 
> **Behavior change:** both paths now **`storage_write(Key::Entry, …)` first**, then `Index::update_hash_for`. Observers see either old hash + old entries or new hash + new entries, not index-ahead-of-bytes. Comments document the accepted **orphan** case if index update fails after entry write, and that `storage_write`’s `bool` means eviction vs insert, not success (real failures trap as `HostError`).
> 
> **SDK docs:** `env::storage_write` vs `private_storage_write` return-value semantics are clarified and cross-linked so callers do not treat the bool as a failure flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 272c5368c64f3791502bf45c15c088b6d2c5979d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/sdk/src/env.rs
+++ b/crates/sdk/src/env.rs
@@ -462,6 +462,11 @@ pub fn storage_remove(key: &[u8]) -> bool {
 /// which propagates as a runtime error, not as `false`. Callers that
 /// don't care whether the key existed before can safely discard the bool.
 ///
+/// **Note:** [`private_storage_write`] uses a different return-value
+/// convention (true = succeeded, false = private storage unavailable)
+/// because that backend can be absent at runtime, unlike main storage
+/// which is always available.
+///
 /// # Example
 ///
 /// ```rust,no_run
@@ -562,8 +567,20 @@ pub fn private_storage_remove(key: &[u8]) -> bool {
 ///
 /// # Returns
 ///
-/// * `true` - If the write operation succeeded
-/// * `false` - If the write operation failed or private storage is not available
+/// * `true` - The write succeeded (a new entry was inserted, or an
+///   existing entry was overwritten).
+/// * `false` - Private storage is not available on this node (the
+///   backend is optional and may be absent at runtime; the write
+///   was a no-op).
+///
+/// Other failure modes (`KeyLengthOverflow`, `ValueLengthOverflow`,
+/// `InvalidMemoryAccess`) trap the WASM module via `HostError` rather
+/// than returning `false`.
+///
+/// **Note:** [`storage_write`] uses a different return-value
+/// convention (true = previous value evicted, false = new key
+/// inserted) because main storage is always available, so a
+/// success/failure bool would be tautological.
 ///
 /// # Example
 ///
@@ -572,6 +589,8 @@ pub fn private_storage_remove(key: &[u8]) -> bool {
 ///
 /// if env::private_storage_write(b"my_secret", b"secret_value") {
 ///     println!("Private value stored successfully");
+/// } else {
+///     println!("Private storage is not available on this node");
 /// }
 /// ```
 #[inline]
--- a/crates/sdk/src/env.rs
+++ b/crates/sdk/src/env.rs
@@ -453,8 +453,14 @@ pub fn storage_remove(key: &[u8]) -> bool {
 ///
 /// # Returns
 ///
-/// * `true` - If the write operation succeeded
-/// * `false` - If the write operation failed
+/// * `true` - A previous value existed under `key` and was evicted.
+/// * `false` - No previous value; a new entry was inserted.
+///
+/// The return value is **not** a success/failure indicator. The write
+/// itself either succeeds or traps the WASM module with a `HostError`
+/// (`KeyLengthOverflow`, `ValueLengthOverflow`, `InvalidMemoryAccess`),
+/// which propagates as a runtime error, not as `false`. Callers that
+/// don't care whether the key existed before can safely discard the bool.
 ///
 /// # Example
 ///
@@ -462,7 +468,9 @@ pub fn storage_remove(key: &[u8]) -> bool {
 /// use calimero_sdk::env;
 ///
 /// if env::storage_write(b"my_key", b"my_value") {
-///     println!("Value stored successfully");
+///     println!("Overwrote an existing value");
+/// } else {
+///     println!("Inserted a new entry");
 /// }
 /// ```
 #[inline]
--- a/crates/storage/src/interface.rs
+++ b/crates/storage/src/interface.rs
@@ -2097,8 +2097,9 @@ impl<S: StorageAdaptor> Interface<S> {
         // success/failure. Actual write failures surface as `HostError`
         // traps from the runtime (`KeyLengthOverflow`,
         // `ValueLengthOverflow`, `InvalidMemoryAccess`), not as
-        // `Ok(false)`. Discard the bool.
-        _ = S::storage_write(Key::Entry(id), &final_data);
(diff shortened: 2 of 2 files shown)


chefsale:
Both reviewer concerns are based on a misreading of `storage_write`'s `bool` return value. Addressing in `fb7da498`:

**The actual semantic** (runtime contract at `crates/runtime/src/logic/host_functions/storage.rs:160-228`):
- Returns `1` if a previous value was **evicted** (key already existed).
- Returns `0` if a **new key was inserted**.
- Actual write failures (`KeyLengthOverflow`, `ValueLengthOverflow`, `InvalidMemoryAccess`) surface as `HostError` traps that unwind through the runtime — **not** as `Ok(false)`.

`MockedStorage::storage_write` follows the same shape (`BTreeMap::insert(...).is_some()` = "key existed before").

So `_ = S::storage_write(...)` is correct — it discards an eviction signal, not a swallowed error. Propagating it as `?` (or treating `false` as failure) would be wrong: the typical "inserted a new char entry" case returns `false`, which the suggested change would treat as a failure and bail.

**What I fixed in `fb7da498`** since the same confusion will recur:
- `crates/sdk/src/env.rs:444-475` — corrected the SDK docstring (was incorrectly claiming "true = succeeded, false = failed").
- Both `interface.rs` call sites — added a clarifying inline comment pointing at the trap-not-bool semantic.

No behavior change in `fb7da498`.

chefsale:
Good catch — my prior comment claimed orphan unreachability via `find_by_id`, which is true, but I glossed over `find_by_id_raw`. Addressed in `7c9d4738`.

Production-caller audit: the orphan is still effectively unreachable, but the safety net is the **index-derived child enumeration** in every caller, not `find_by_id_raw` itself.

| Caller | Orphan exposure |
|---|---|
| `compare_trees` (orphan's own id, line 1511) | Bails at line 1525 via `local_metadata.ok_or(IndexNotFound)?` before bytes become an `Action` |
| `compare_trees` child iteration (lines 1591, 1619) | Child ids come from `get_children_of(id)` — index lookup, orphan not present |
| `hash_comparison.rs:360`, `hash_comparison_protocol.rs:898/1058`, `level_sync.rs:828` | All walk parents' index-derived child lists — orphan never iterated |

The only way to *actually* read the orphan is to call `find_by_id_raw` directly with the orphan's id. Production code doesn't do that.

Updated both `save_internal` and `write_pre_merged_root_state` comments to explain this distinction precisely (`find_by_id` vs `find_by_id_raw` semantics, and why every production caller's index-derived enumeration filters the orphan out anyway). No behavior change.

chefsale:
Re the two new findings:

## 🟡 Orphan via `compare_trees` falling into `Action::Add` branch

This finding misreads the control flow. Tracing precisely (file unchanged from previous review, line numbers from current `master`):

\`\`\`rust
// Line 1509 — read metadata (Option, no error on None)
let local_metadata = <Index<S>>::get_metadata(id)?;

// Line 1511 — early-return ONLY when entry is missing
let Some(local_entity) = Self::find_by_id_raw(id) else {
    if let Some(foreign_entity) = foreign_entity_data {
        actions.0.push(Action::Add {
            id,
            data: foreign_entity,   // <-- FOREIGN data, not local orphan
            ...
        });
    }
    return Ok(actions);
};

// Line 1525 — orphan case reaches here, bails
let local_metadata = local_metadata.ok_or(StorageError::IndexNotFound(id))?;
\`\`\`

For the orphan case (entry exists, no index):
- `local_metadata = None`
- `local_entity = Some(orphan_bytes)` — the `let Some(...)` pattern **matches**, so we **do not** take the early-return branch at lines 1512-1522.
- We fall through to line 1525, which returns `Err(IndexNotFound)`.

The `Action::Add` branch at lines 1512-1522 fires only when `find_by_id_raw` returns `None` (no entry exists), and the `data` field it sets is `foreign_entity` — not the local orphan bytes. The orphan cannot propagate via `compare_trees`.

The forward-looking concern (\"this assumption is not enforced at the type level and could be violated by future callers\") is fair in general, but the safety here doesn't rely on every caller iterating index-first — it relies on `compare_trees`'s explicit `local_metadata.ok_or(IndexNotFound)?` gate, which IS enforced at the type level for that function.

## 📝 `_ =` vs `let _ignored =`

Fixed in `272c5368` — standardized to `let _ignored = ...` to match the existing convention at the `storage_remove` site (line 1448).

PR #2477, opened 2026-05-25
fix(context): repair per-context migration write path broken by #2433

rtb-12:
## Summary

Repairs per-context application migration, which has been silently broken on every node since #2433. The `upgrade_group` RPC and the migration-suite-v1..5 fixtures already existed (#1911); what was broken was the storage write at the tail of the migration handler.

This PR does **not** add new migration functionality — it makes the existing group-level migration flow actually work end-to-end again, and adds an e2e regression guard so the same class of bug can't slip in unnoticed.

## What was broken

When the wasm migrate function returned its v2-shaped bytes, `write_migration_state` persisted them via `Interface::save_raw`. Since #2433 changed `Root<T>` merge semantics, `save_raw` started routing root-class entries through the host-side merge registry — and since #2465 that registry is intentionally empty for app types (those merges live in the wasm runtime, dispatched via the `__calimero_merge_root_state` export). Result: every migration hit `MergeFailure(NoMergeFunctionRegistered)` at the storage layer and bailed up to the operator as `"Failed to write migrated state"`.

## Architecture decision

Three forces had to be reconciled:

1. **The migrate function already produces fully-resolved v2-shaped state.** There is nothing to merge — the caller is the source of truth for the new bytes. So we don't need (or want) a merge-registry dispatch on this write path.
2. **Storage must still update the Merkle index** so the sync protocol observes the new root hash.
3. **Concurrent writes against the same context must not race the migration into corruption.** If a regular state-changing call slips a newer timestamp in between the migration's read and write, persisting the new `application_id` without applying the migrated bytes would leave the v2 binary running against v1-shaped state.

`Interface::write_pre_merged_root_state` (introduced by #2465 for exactly this shape) handles (1) and (2): it skips merge-registry dispatch, hashes the caller's bytes, updates the Merkle index, and writes storage. It also has a built-in LWW guard for (3) — but the guard returns the existing hash on skip rather than surfacing an error.

To make (3) safe end-to-end, the migration write now bookends the primitive call with:

- **Pre-flight LWW check** — fast bail before doing any work if local state is already newer than the migration's deterministic timestamp.
- **Post-write metadata verification** — closes the TOCTOU window where a racing writer could slip a newer timestamp in between the pre-flight read and the primitive call. If the stored `updated_at` doesn't equal what we tried to write, the write was silently rejected by the primitive's internal LWW guard and we bail.

Both check arms route through the existing `Ok(None)` → outer match → `bail!` plumbing so the calling upgrade flow never persists a new `application_id` against un-migrated bytes.

## What this PR ships

- `crates/context/src/handlers/update_application/mod.rs`: `write_migration_state` rewritten as described above; surrounding `match result` simplified accordingly. All 16 existing `verify_appkey_continuity` unit tests still pass.
- `workflows/app-migration/00-single-group-migration-baseline.yml`: single-node v1 → v2 migration; asserts the post-fix-only DEBUG line `Migrated state written successfully`, full v2 schema shape (`schema_version: "2.0.0"`, `notes: "added in v2"` populated by the migrate function, preserved v1 fields), and that v2-only setter/getter round-trip post-migration. **This is the regression guard** — pre-fix the assertion fails because the bail message fires instead.
- `.github/workflows/app-migration-e2e.yml`: new CI job; builds `merod:local` + migration-suite fixtures, runs every workflow under `workflows/app-migration/`, uploads per-node docker logs as artefacts.
- `workflows/app-migration/{build-wasms.sh, README.md}`: convenience helpers + docs.

## What this PR does NOT do (future scope)

This is PR-1 of a planned train. Each later PR is scoped to ship and verify independently:

- **merobox PR** (parallel, separate repo) — `cascade: bool` field on the `upgrade_group` step + new `get_cascade_status` step type. Prerequisite for PR-2/PR-3 CI.
- **PR-2 — namespace-level cascade engine.** One signed `CascadeTargetApplicationSet` op fans out to every matching descendant subgroup + context in a single sync round, instead of N separate `upgrade_group` RPCs. Adds apply-handler tree-walk, predicate-based heterogeneous-skip, and per-context migration dispatch. The `Cascade*` wire-format variants already exist on master (merged in #2452) but have no apply handlers yet.
- **PR-3 — HLC sync fence + `get_cascade_status` RPC.** Sync-layer rejection of stale-schema deltas from long-offline straggler peers (`UpgradeFenced` structured error) plus an operator visibility surface for per-peer per-context cascade progress.
- **PR-4 (optional)** — `force_complete_cascade` admin escape hatch + a multi-version-coexistence regression workflow (proving that namespace A on v1 and namespace B on v2 can run side-by-side on the same node without cross-corruption).

## Test plan

- [x] `cargo check -p calimero-context` (dev + release profile) clean
- [x] `cargo test -p calimero-context --lib update_application` — 16/16 unit tests pass
- [x] Migration-suite WASM fixtures (`v1`, `v2-add-field`) build cleanly via `workflows/app-migration/build-wasms.sh`
- [ ] CI `App migration e2e` job: workflow 00 PASSES on this PR (empirical regression-fix evidence)
- [ ] CI Rust + Build Image + Build Binaries pass on this PR
- [ ] Other existing CI jobs unaffected

Local docker verification was skipped (no docker daemon in the contributor environment); CI on this PR provides the empirical proof that the fix unblocks the migration flow.

---

> [!NOTE]
> **Medium Risk**
> Touches context migration persistence and concurrent-write safety on root storage; incorrect LWW handling could block upgrades or allow inconsistent app/state pairing, mitigated by explicit bail paths and new e2e workflow.
> 
> **Overview**
> Repairs **per-context application migration** after #2433 broke the tail of `upgrade_group`: migrated WASM bytes were persisted with `save_raw`, which now requires a host merge function for root entries and failed with `MergeFailure(NoMergeFunctionRegistered)`.
> 
> `write_migration_state` in `update_application/mod.rs` now commits fully-resolved migrate output via **`write_pre_merged_root_state`** (no merge registry), still updates the Merkle index and returns the tree root hash for `ContextMeta`. **Pre-flight and post-write LWW checks** turn silent skips into `bail!` so a newer `application_id` is not stored without migrated bytes (avoids v2 code on v1-shaped state).
> 
> Adds **migration e2e coverage**: merobox workflow `00-single-group-migration-baseline.yml` (v1→v2, log + schema asserts as regression guard), `build-wasms.sh`, README, and **`.github/workflows/app-migration-e2e.yml`** (build `merod:local`, run all `workflows/app-migration/*.yml`, upload docker logs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3362d07982e5a400f3bf1ae2c4f2f82dfc81235. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rtb-12:
## Bot review triage

Addressed in `387bcff5`:

### ✅ cursor[bot] HIGH — "Migration succeeds on LWW skip"

Real correctness regression and the only critical bot finding. Replacing `save_raw` with `write_pre_merged_root_state` made the LWW-skip case silently succeed, while the surrounding `update_application_id` flow still flipped `application_id` to v2 — leaving the v2 binary running against un-migrated v1-shaped state. Pre-`#2433` the equivalent `Ok(None)` branch caught this.

Restored the safety net via a pre-flight metadata check before the write:

```rust
if let Some(existing) = <Index<MainStorage>>::get_metadata(root_entry_id)? {
    if existing.updated_at > metadata.updated_at {
        return Ok(None);   // outer match bails with retry-this-migration error
    }
}
```

First-time migrations (no existing entry) skip the guard and proceed normally. Tests still pass (`cargo test -p calimero-context --lib update_application` — 16/16).

### ⏭️ meroreviewer warning — "No size guard on migration WASM output"

Defer. Pre-existing concern unchanged by this PR: pre-#2433 the same bytes flowed through `save_raw` without validation. A `MAX_MIGRATION_STATE_SIZE` guard is a sensible defense-in-depth follow-up but isn't a regression introduced here; cleaner to handle as its own focused PR with a chosen size limit and a test.

### ⏭️ meroreviewer warning — "Zero hash fallback on missing Index entry"

Defer. The `.unwrap_or([0; 32])` line predates this PR and is untouched here. Tightening it could surface latent bugs across other code paths that rely on the current behavior; outside this PR's scope.

### ⏭️ meroreviewer suggestion — "Fork-PR checkout could run arbitrary code in CI"

Skip. Matched against repo norms: `sync-regression.yml` and the other e2e jobs in `.github/workflows/` don't use the fork-PR `head.repo.full_name == github.repository` guard — only `ai-review.yaml` does, because it consumes an Anthropic API key secret. The migration e2e job uses no secrets beyond `GITHUB_TOKEN` (already read-only for fork PRs by GitHub's default). Adding the guard here would deviate from established repo patterns without a corresponding risk increment.

rtb-12:
## Bot review triage (round 2, on \`387bcff5\`)

Meroreviewer surfaced two new warnings on the LWW-fix commit. Triage:

### ✅ Finding #1 (1/3 agents): "TOCTOU window in pre-flight LWW check"

Real concern. The pre-flight read → write_pre_merged_root_state path leaves a window where a raced writer can slip a newer \`updated_at\` between the two operations, in which case the primitive's own internal LWW guard silently returns the existing hash and our outer match treats the skip as \`Ok(Some(_))\` — exactly the corruption case the safety net was meant to prevent. In practice the actix actor model probably serialises writers so the window is narrow-to-nonexistent, but the safety net should be airtight given the v2-binary-on-v1-state stakes.

Closed in \`0d142ac3\` via post-write verification:

\`\`\`rust
let intended_updated_at = metadata.updated_at;
let _ = Interface::<MainStorage>::write_pre_merged_root_state(root_entry_id, &entry_bytes, metadata)?;
if let Some(after_index) = <Index<MainStorage>>::get_index(root_entry_id)? {
    if after_index.metadata.updated_at != intended_updated_at {
        return Ok(None);   // outer match bails
    }
}
\`\`\`

### ⏭️ Finding #2 (1/3 agents): "\`u64::MAX/2\` fallback in compute_deterministic_metadata may cause permanent LWW rejection"

Defer. The fallback lives in \`compute_deterministic_metadata\` (line ~638), which this PR does not touch. It's a pre-existing concern about a hypothetical chained failure (index read fails → fallback used → next migration's index read fails → permanently stuck). Fixing it properly requires a deliberate choice about the fallback semantics (use \`u64::MAX\`? wall-clock?) and ideally a test that reproduces the stuck state — outside the scope of this regression fix.

PR #2487, opened 2026-05-26
fix(network): include /p2p/<relay-peer> in relayed multiaddr before /p2p-circuit

chefsale:
Follow-up to #2483.

After #2483 unblocked the relay-reservation trigger for NAT'd nodes (lifting the `advertise=None` gate), the merobox NAT-topology smoke tests in [merobox#254](https://github.com/calimero-network/merobox/pull/254) revealed a second silent failure in the same code path. Identify handler now reaches `create_relay_reservation` as intended, but every client logs:

```
ERROR ... Failed to handle relay reservation err=
```

Empty error message — Display chain stops at an unhelpfully terse libp2p `swarm::ListenError` because the bare `?` propagation in `create_relay_reservation` adds no eyre context.

## Root cause

The relayed multiaddr constructed by

```rust
preferred_addr.clone()
    .with(Protocol::P2pCircuit)
    .with_p2p(*self.swarm.local_peer_id())
```

ends up as e.g.

```
/ip4/172.18.0.2/udp/4001/quic-v1/p2p-circuit/p2p/<self>
```

— **missing the `/p2p/<relay-peer>` segment between the transport and `/p2p-circuit`**. libp2p's relay-client transport requires the full form:

```
/<transport>/p2p/<relay-peer>/p2p-circuit/p2p/<self-peer>
```

…because that's how the relay-client knows which peer is serving the circuit. Without it, `listen_on` rejects the multiaddr with a near-empty `TransportError` and the reservation flow silently fails.

Why the addr lands incomplete: `peer_info.addrs` is populated by the listen_addrs loop in `identify.rs` from the peer's raw libp2p `listen_addrs`, which are bare network endpoints (`/ip4/.../udp/.../quic-v1`) with no `/p2p/` suffix. `get_preferred_addr` prefers UDP, so it returns one of those bare endpoints. The original construction only appended `/p2p-circuit` and `/p2p/<self>` on top — but never the relay-peer segment.

## Fix

- Append `/p2p/<relay-peer>` to the base address only if it isn't already present. Some entries in `peer_info.addrs` are populated from `ConnectionEstablished` events and DO carry `/p2p/`, so we'd duplicate it otherwise — and a multihash with two P2p components is invalid.
- Wrap `listen_on` with eyre context so the next instance of this family of bugs surfaces a useful error in the identify handler's log line instead of `err=`.
- Import `WrapErr` trait so `.wrap_err_with(...)` compiles on Results (the existing `wrap_err` usage in this file is via `ContextCompat` on `Option`).

## Validation

- `cargo fmt --check` clean
- `cargo build -p calimero-network` clean
- `cargo build -p merod` clean (warnings pre-existing)
- `cargo test -p calimero-network --lib` — 22/22 pass

The end-to-end regression test is merobox#254's NAT-topology smoke tests, which were red on the previous `merod:edge` and should turn green on a build that includes this fix.

---

> [!NOTE]
> **Medium Risk**
> Changes relay reservation multiaddr construction and listen setup on a NAT-critical path; mis-handling could break reservations or duplicate P2p components, though the fix is narrowly scoped with defensive checks.
> 
> **Overview**
> Fixes **silent relay reservation failures** for NAT'd nodes by building the libp2p relay listen address in the required form: transport, then `/p2p/<relay-peer>`, then `/p2p-circuit`, then `/p2p/<self>`. When the preferred relay address comes from Identify **listen_addrs** (bare `/ip4/.../udp/.../quic-v1` with no peer id), the code now **inserts `/p2p/<relay-peer>` only if missing**, avoiding invalid duplicate `P2p` components when the addr already came from a connection event.
> 
> **`listen_on`** failures are wrapped with eyre context so identify-handler logs show the full relayed multiaddr instead of an empty `err=`.
> 
> Separately, the **DCUtR** swarm handler logs **info** on successful direct-connection upgrades and **warn** on failures, so operators can see hole-punch outcomes without debug logging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64ae398103242408ab9e0b190ec3f4126c3f6a11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
**Updated PR description for the additional commit:**

Bundled a small observability improvement based on the relay-architecture conversation: promote DCUtR direct-connection upgrade outcomes from `debug!` to `info!`/`warn!`. DCUtR was already wired into the swarm (`behaviour.rs`) and runs autonomously once a relay circuit exists between two peers — the handler just needed to surface whether the hole-punch upgrade actually succeeded so operators don't have to trawl `RUST_LOG=debug`.

This matches the "relay-for-discovery, DCUtR-for-direct" pattern: once this PR's bug fix unblocks relay reservations, DCUtR fires automatically. On cone-NAT peers the upgrade succeeds → `info!` "direct-connection upgrade succeeded with peer X" → traffic transparently moves off the relay. On symmetric-NAT peers the upgrade fails → `warn!` "relay path will continue to carry traffic: …" → expected fallback behaviour.

The merobox#254 NAT smoke tests should see exactly this asymmetry:
- cone-mode workflow → `direct-connection upgrade succeeded` lands within a few seconds of reservation
- symmetric-mode workflow → `direct-connection upgrade failed` lands, relay carries all data

PR #2493, opened 2026-05-26
feat(context): cascade engine for namespace-level application migration

rtb-12:
## Summary

PR-2 of the 4-PR namespace-cascade-migration train (PR-1 = #2477, merged). Wires the wire-format `CascadeTargetApplicationSet` and `CascadeGroupMigrationSet` GroupOp variants (added in #2452) into a working end-to-end cascade engine. One signed cascade op now fans out to every matching descendant subgroup + context in a single sync round, replacing N separate `upgrade_group` RPCs against each subgroup.

Heterogeneous deployments are first-class: descendants whose current `app_key` differs from the cascade's `from_app_key` are silently skipped, so a namespace can carry multiple app versions safely.

## Architecture decisions

- **Apply handlers do settings mutation only** (`crates/context/src/group_store/mod.rs`). Both `Cascade*` variants walk the local tree via `cascade::walk_for_predicate`, skip non-matched descendants, and call the same `GroupSettingsService::set_target_application` / `set_group_migration` paths the single-group variants already use. The apply path is pure sync — it does NOT spawn propagators directly.
- **Per-descendant propagator spawn lives on the emitting node**'s `upgrade_group` RPC handler (`dispatch_cascade`), iterating the walk result and spawning `propagate_upgrade` per matched group. This mirrors today's per-group flow where the emitter runs migration → emits state deltas → other peers merge v2 deltas through existing state-delta sync. Receivers update their `GroupMeta` via the apply arm but do not run migration locally.
- **Capability check is pre-scan + fail-fast**: each cascade arm runs a read-only pass over matched descendants via the new `permission_checker::can_manage_application` helper. If any descendant fails the check, the arm bails before any writes happen — atomicity over partial-state.
- **`from_app_key` doubles as optimistic-concurrency control** (per spec §5): two concurrent cascade ops applied in any order on different replicas converge, because whichever ops the DAG orders second will find `from_app_key` no longer matches and silently no-ops. The cascade arm returns `Ok((true, None))` for the no-match case — variant handled, no divergence — which is the explicit safety property guarded by `cascade_concurrent_safety`.
- **Write-gate is per-group, not per-context** (matches existing `GroupUpgradeStatus` shape). Sync replay (`__calimero_sync_next`) and the recovery path (which goes through `update_application`, not `ExecuteRequest`) bypass naturally — no deadlock risk.
- **Walk has its own visited-set** rather than delegating to `collect_descendant_groups` (which lacks cycle protection). Terminates on any finite graph, cyclic or not.

## What's in the diff (Rust + tests only — workflows moved to follow-up PR)

Commits on `feat/cascade-engine`:

1. `feat(context): cascade engine — apply handlers + descendant walk with predicate skip`
2. `feat(context): wire cascade emission through upgrade_group RPC + meroctl --cascade`
3. `feat(context): refuse user writes against contexts whose group is upgrading`
4. `test(context): cascade engine integration tests (apply-walk + concurrent-safety)`
5. `fix(context): address bug-bot findings — dead test branch + silent signing-key error`
6. `fix(client): add missing cascade field to UpgradeGroupApiRequest test ctor`
7. `test(node): actor-level integration test for cascade RPC end-to-end on a single node`

## Test coverage

| Layer | Coverage |
|---|---|
| **Walk + predicate (unit, 6 tests)** | `crates/context/src/cascade/walk_*_tests.rs` — predicate match/mismatch, signed-group inclusion, missing-meta edge case, deep nesting, cycle termination |
| **Apply-arm logic (integration, 2 tests)** | `crates/context/tests/cascade_apply_walk.rs`, `cascade_concurrent_safety.rs` — fan-out + sibling untouched + DAG-mediated convergence + C1 regression guard |
| **dispatch_cascade RPC end-to-end on a single node (actor-level, 3 tests)** | `crates/node/src/cascade_dispatch_e2e.rs` — emitter flow (walk → publish → save InProgress → spawn propagators), write-gate refuses user calls during InProgress, predicate-skip on heterogeneous subgroup |
| **No regression on existing per-group flow** | Workflow 00 (PR-1's single-group migration regression guard) runs unconditionally in CI |
| **Workspace-wide regression** | 50+ existing CI integration jobs all green |

## Coverage gap (closed by follow-up #2494)

This PR does NOT exercise **cross-peer cascade convergence via real gossip** — the multi-node Docker-based e2e tests that prove the full flow (admin emits → 2 nodes converge on v2 state) are deferred to follow-up #2494. Those workflows need merobox#255 cascade-step support to publish first, which is itself blocked on this PR merging.

The actor-level integration test in `cascade_dispatch_e2e.rs` covers everything that can be tested in-process on a single node (the entire emitter-side flow); cross-peer convergence is what requires docker.

## Descoped from this PR (will land in #2494)

The merobox e2e workflows (`01-single-namespace-cascade.yml`, `04-cascade-skip-heterogeneous.yml`, `05-cascade-chain-v1-to-v3.yml`), the `migration-suite-v3-add-list` WASM fixture, and the CI version-guard were initially included but reverted (`d1f06a39` / `cfb4852e` / `fe1ba91a`). The reverted commits remain in this branch's history and can be cherry-picked into #2494 once merobox#255 publishes.

## Out of scope (forward-referenced)

**PR-3 will add:**
- Per-context `cascade_hlc` storage field (requires a storage migration)
- HLC sync fence in `state_delta/mod.rs::apply_authorized_state_delta` — the proper cross-peer write coordination (the current write-gate is emitter-only by design; the HLC fence sits in the state-delta apply path and works on all peers)
- `get_cascade_status` RPC
- Workflow 03 — cascade-with-offline-straggler (uses merobox#249 fault-injection `pause`/`restart`)
- Workflow 06 — fence-rejection (stretch)

**PR-4 (optional) will add:**
- `force_complete_cascade --evict-peer X` admin RPC
- Workflow 02 — multi-version coexistence in one node

## Cross-repo dependency chain

```
This PR (#2493) merges
    ↓
calimero-client-py PR: add `cascade: bool` field to upgrade_group method
    ↓
calimero-client-py release
    ↓
merobox#255: implement cascade step support, bump client-py pin
    ↓
merobox release
    ↓
core #2494 follow-up: cherry-pick reverted commits + bump merobox pin
    ↓
Full multi-node cascade verification in CI
```

---

> [!NOTE]
> **High Risk**
> Changes group governance apply, upgrade RPC emission, and execute gating during migrations—areas that affect multi-peer consistency and user traffic; mitigated by atomic pre-scans and broad tests but cross-peer gossip e2e is deferred.
> 
> **Overview**
> This PR implements **PR-2 of the namespace cascade migration train**: one signed upgrade can fan out to every descendant subgroup whose `app_key` matches `from_app_key`, instead of calling `upgrade_group` per subgroup.
> 
> **API & defaults.** `UpgradeGroupRequest`, admin API, `meroctl`, and the HTTP client gain an opt-in **`cascade: bool`** (default **`false`**) so existing single-group upgrades stay unchanged.
> 
> **Apply path (gossip / local apply).** New `cascade::walk_for_predicate` walks the group tree (cycle-safe visited set), evaluates `from_app_key == GroupMeta.app_key`, and drives **`CascadeTargetApplicationSet`** / **`CascadeGroupMigrationSet`** in `apply_group_op_mutations`: pre-scan `MANAGE_APPLICATION` on all matched nodes, then per-group `set_target_application` / `set_group_migration` only (no propagator in apply). `GroupSettingsService::set_target_application` no longer calls the old no-op `cascade_target_application` helper. Loser concurrent cascades no-op via predicate mismatch but still return **`handled = true`**.
> 
> **Emitter path (`upgrade_group`).** When `cascade` is set, **`dispatch_cascade`** validates admin/signing, walks matched descendants **before** publish, signs/publishes cascade ops, writes per-descendant **`GroupUpgradeStatus::InProgress`**, and spawns **`propagate_upgrade`** per matched group (plus optional migration op).
> 
> **Safety during migration.** Execute handler refuses user **`call`**s (not sync replay) when the context’s group has **`InProgress`** upgrade, surfacing **`ExecuteError::UpgradeInProgress`** (group-level gate until PR-3 `cascade_hlc`).
> 
> **Tests** cover walk predicates, apply fan-out, DAG concurrent-cascade convergence, and single-node `dispatch_cascade` + write-gate + heterogeneous skip.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 507671f24aa667be46af5d41cc7560b8fe17755f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rtb-12:
## Triage of bug-bot findings on commit `b6952ab6` (pre-revert)

This PR descoped the merobox e2e workflows + v3 fixture + CI guard to follow-up #2494 (commits `d1f06a39` / `cfb4852e` / `fe1ba91a` reverted them). Findings against those reverted files are no longer applicable; others are addressed below.

### Fixed in code

| Finding | Bot | Where | Fix |
|---|---|---|---|
| Dead \`Err\` branch asserting \"cycle\"/\"cap\" in walk_no_infinite_loop_on_cycle | meroreviewer + github-actions | walk_depth_bound_tests.rs:130/140 | Collapsed match into single \`expect()\` on Ok path (visited-set termination only). Commit \`fa5cd0a8\`. |
| \`let _ = store_group_signing_key\` silent error swallow | meroreviewer + github-actions | upgrade_group.rs:820 | Converted to \`if let Err(err) = ... { warn!(target: \"calimero::cascade\", ?err, ?group_id, ...) }\`. Commit \`fa5cd0a8\`. |
| Missing \`cascade\` field in UpgradeGroupApiRequest test ctor | (caught by CI) | client/src/tests.rs:594 | Added \`cascade: false\` to the test struct. Commit \`610cba4a\`. |

### Deferred to PR-3 (already-stated PR-2 scope boundary)

| Finding | Severity | Rationale |
|---|---|---|
| **Follower nodes skip upgrade gate** (cursor C1) | HIGH | The write-gate is emitter-only by design. PR-3's HLC sync fence sits in the state-delta apply path (\`state_delta/mod.rs::apply_authorized_state_delta\`) and gates ALL peers regardless of whether they ran the migration locally. PR description already states: \"Receivers update their GroupMeta via the apply arm but do not run migration locally.\" |
| Missing concurrent-upgrade guard for descendant groups | MEDIUM | Per spec §5, concurrent cascade ops converge via the \`from_app_key\` predicate — whichever op the DAG orders second finds the predicate stale and silently no-ops. The explicit C1 regression guard test (\`cascade_concurrent_safety\`) exercises this. |
| Cascade ignores LazyOnAccess policy (cursor C3) | MEDIUM | Cascade is a coordinated namespace-wide op by design; \`LazyOnAccess\` applies to single-group flow only. Documented in spec §3.2. |
| matched_descendants TOCTOU windows | MEDIUM | Best-effort guards are acknowledged in spec §5; concurrent cascades are predicate-skipped. Predicate-based concurrency control is the design. |

### Acknowledged design choices / edge cases (no code change)

| Finding | Rationale |
|---|---|
| Skipped propagator on \`save_group_upgrade\` failure (cursor C2) | Narrow edge case — \`active_propagators.insert\` returning false means a propagator is already running (correct skip). \`save_group_upgrade\` failure is rare; recovery (\`lifecycle.rs::recover_in_progress_upgrades\`) handles the InProgress-without-propagator case on next startup. |
| validate_upgrade re-implementation in dispatch_cascade | Intentional — \`validate_upgrade\` requires a canary context which namespace roots often don't have (they hold descendant subgroups, not contexts). Applicable subset of checks inlined per spec; documented in the code comment. |
| Tree walked twice when CascadeTargetApplicationSet + CascadeGroupMigrationSet are paired | Performance, not correctness. Could be optimized in a follow-up; acceptable for V1. |
| Propagators spawned for signed root even when it has no contexts | \`propagate_upgrade\` handles empty groups gracefully (immediate completion). No-op spawn. |
| Two synchronous store reads in execute hot path | Common case (no upgrade running) — \`load_group_upgrade\` returns \`None\` immediately. Not a hot-path blocker. Per-context cache or refined gating is a PR-3 refinement. |
| \`pre_spawn_totals\` zip ordering with \`matched_descendants\` | Vec iteration order is stable; correct as written. |
| \`active_propagators\` race check after publish | Best-effort guard; same propagator-already-running case handled benignly. |
| Cascade validation could also check \`app_key == new_app_key\` | Defensive but not a correctness gap; the target_application_id check covers the same intent. |
| \`divergent_cascade_apply_order_converges\` test relies on DAG impl details | Yes — the test exercises actual current DAG causal-ordering behavior, which is the correct level to test against. |
| \`migrate_v2_to_v3\` panics on missing state (test fixture) | Test fixture behavior — intentional fail-loud in test environment. |
| \`sort -V\` portability on macOS (CI guard) | N/A — CI runners are Linux. (Moot anyway: CI guard is reverted; follow-up #2494 will reintroduce with a more portable check.) |
| AGENTS.md / CONTRIBUTING.md (doc-bot) | No architecture-shifting changes in this PR needing those docs. |

All addressed/resolved threads will be marked as such via \`resolveReviewThread\`.

Re-running CI now against the descoped diff (HEAD = \`fe1ba91a\`).

rtb-12:
**Addendum: new cursor finding \`Missing pre-publish upgrade lock\` on commit \`fe1ba91a\`**

Same architectural family as the concurrent-cascade-safety case already addressed: there is a small window between cascade op publish/apply and the per-descendant \`GroupUpgradeValue::InProgress\` write. The bot is correct that this window exists.

**Why it's safe:** the predicate-skip semantics on the apply arm (\`from_app_key\` mismatch → silent no-op, returning \`Ok((true, None))\`) make any second concurrent cascade a safe no-op:

1. Admin A and Admin B both read \`meta.app_key = K1\` at T1/T2.
2. A's cascade publishes + applies — local meta is now K2.
3. B's cascade publishes — its op carries \`from_app_key = K1\`.
4. B's apply walks the tree, finds every descendant's current \`app_key = K2 ≠ K1\`, predicate-skips all of them.
5. B's apply returns \`Ok((true, None))\` — variant handled, no divergence, no duplicate state mutation.

The cascade_concurrent_safety integration test exercises exactly this DAG-ordered convergence. The small write-gate-gap window for the descendants is a documented limitation in PR-2 (the write-gate is emitter-only; PR-3's HLC fence is the proper cross-peer gate).

A pre-publish InProgress lock would shrink the window further but doesn't change the convergence guarantee — it just provides earlier write-gate coverage on the emitting node itself. Documenting as a refinement opportunity rather than a correctness bug.

Resolving the thread accordingly.

rtb-12:
**Addendum: 10 new findings on commit \`f994b0c0\` (actor-level integration test)**

All addressed via the existing triage rationale or verified-correct on inspection. Resolving threads accordingly.

| Finding | Decision |
|---|---|
| cursor — \"Descendant in-progress not checked\" | Same family as previous \"Missing pre-publish upgrade lock\" — predicate-skip on \`from_app_key\` makes the second concurrent op a safe no-op. Documented earlier. |
| meroreviewer — \`cascade_dispatch_e2e.rs\` placement in \`crates/node/src/\` (524 LOC in production crate) | Intentional. The file is a \`#[cfg(test)] mod\` (zero release-build cost) but reuses the existing \`local_governance_node_e2e\` actor-harness via a \`pub(crate)\` visibility tweak — placing it under \`tests/\` would have required duplicating ~125 LOC of harness setup (ArbiterPool, SyncManager, NodeManager, blobstore tempdirs). Subagent's deliberate trade-off. Documented in the test file's module doc-comment. |
| meroreviewer — \"InProgress check only on signed group, not descendants\" | Same family as the cursor finding above. Predicate-skip handles concurrent-cascade convergence. |
| meroreviewer — \"Signing key validation logic inconsistent\" | Real code-smell (validation and resolution paths not symmetric) but not a correctness bug. The fallback path still works; error messages just have different wording. Worth a follow-up refactor; not blocking PR-2 merge. |
| meroreviewer — \"Test 2 relies on propagator failing due to missing ContextIdentity\" | Valid observation. The test correctly verifies the write-gate behavior (status stays InProgress when execute is called → gate refuses), but the *reason* status stays InProgress is fixture-incomplete (no ContextIdentity), not explicit. A fully deterministic alternative (mock propagator or direct status write) would add ~100 LOC of test infrastructure. Acceptable trade-off for PR-2; could be refactored in follow-up. |
| meroreviewer — \"Pre-publish walk stale by spawn time\" | Same family as previous TOCTOU concerns. The pre-publish snapshot is deliberate (post-publish walk would find zero matches after apply rewrites \`app_key\`). Predicate-skip + active_propagators best-effort guard mitigate. |
| meroreviewer — \"Signing key error silently downgraded to warn\" | Auto-store is best-effort by design. Caller can pass an explicit key on retry; failing the whole cascade because of a transient store error on a *convenience* persistence would be worse UX. Warn-and-continue is intentional. |
| meroreviewer — 💡 \"blob announce \`usize::MAX\` limit\" | Performance suggestion. Per-group context counts are bounded in practice; not a hot-path concern. |
| meroreviewer — 💡 \"walk allocates Vec including non-matched entries\" | Performance suggestion. Walk size is bounded by namespace tree size; trivial allocation. |
| meroreviewer — 💡 \"\`spawn_propagator_for\` passes \`None\` for canary — may skip first context\" | **Verified correct** by inspection: \`propagate_upgrade\` at \`upgrade_group.rs:577\` filters via \`skip_context.map_or(true, |skip| *cid != skip)\`. When \`None\`, the closure returns \`true\` for every context — nothing is skipped. Passing \`None\` means \"no canary to skip, migrate all contexts\" — exactly the cascade requirement. |

Resolving all 10 threads.

rtb-12:
**Addendum: 8 new findings on commit \`507671f2\` (cleanup commit)**

All 8 addressable via prior triage rationale or design intent — no new code changes needed. Resolving threads.

| Finding | Decision |
|---|---|
| Race between pre-publish walk and apply (\`matched_descendants\`) | Same family as previously-addressed TOCTOU concerns; predicate-skip + best-effort \`active_propagators\` guard handle concurrent cascades safely (spec §5). |
| \`signed_status\` uses \`pre_spawn_totals.first()\` — may not correspond to signed group | Defensible — \`walk_for_predicate\` always emits the signed group first; if its app_key doesn't match the predicate the cascade is a no-op (caught by the empty-walk check earlier in \`dispatch_cascade\`). Could add a defensive index lookup but the current code is correct in all reachable states. |
| Signing key auto-store swallows errors via \`warn!\` | Intentional: auto-store is best-effort persistence (caller can always pass an explicit key on retry). The \`warn!\` log preserves the error for operators. Distinguishing \"already stored\" vs \"store corruption\" is an improvement opportunity, not a correctness bug. |
| Empty \`matched_descendants\` causes misleading error when signed group itself doesn't match | Real UX nit — the error message could be more specific about *why* nothing matched (predicate-mismatch on signed group vs no descendants at all). Improvement-tier; resolve as design choice. |
| \`pre_spawn_totals\` snapshot races with context registration between snapshot and propagator spawn | Same TOCTOU family — \`propagate_upgrade\` self-reconciles the actual context count via its own enumeration on each iteration. The snapshot is only used for the initial-status response, not for correctness. |
| Duplicate finding about empty \`matched_descendants\` | Resolved via the same rationale above. |
| 💡 Pre-scan comment about \`?\` unreachable on single-threaded actor path | Comment-quality observation. The pre-scan's \`?\` IS unreachable on the actor path (sole writer); the comment is accurate. |
| 💡 Blob announce \`enumerate_group_contexts\` with \`usize::MAX\` | Performance suggestion, same as prior triage. Per-group context counts are bounded in practice. |

Resolving threads.

chefsale (approved):

PR #2503, opened 2026-05-27
feat(meroctl): network status admin command + endpoint (#2488)

### LINKED ISSUE #2488: feat(meroctl): `network status` admin command — dump swarm connectivity state without RUST_LOG=debug
## Problem

Today the only way to know whether a merod node's libp2p subsystem is healthy is `RUST_LOG=debug` + log archaeology. Specifically:

- **External addresses** the swarm currently advertises → `libp2p_swarm` debug log spam.
- **Relay reservations** (which relay peers, current status — Discovered / Requested / Accepted / Expired) → buried in `calimero_network::discovery` debug events.
- **DCUtR direct-connection upgrade outcomes** per peer → emitted as `info!`/`warn!` (since calimero-network/core#2487) but not exposed as queryable state.
- **Rendezvous registrations** (which rendezvous peers, last registered at, namespace) → in-memory state on the Discovery struct, not surfaced anywhere.

That's fine for the developer who built the subsystem; bad for operators debugging "why is my home node not reaching the fleet?" in production. We don't ship the relay-recovery work documented in [#2446 + #2459](https://github.com/calimero-network/core/issues/2446) without operator visibility into whether those code paths are firing.

## Proposal

A `meroctl network status` command that hits a new `GET /admin-api/network/status` endpoint on merod and pretty-prints the swarm's connectivity state.

### Endpoint shape (sketch)

```json
GET /admin-api/network/status
{
  "local_peer_id": "12D3KooW...",
  "listen_addrs": [
    "/ip4/127.0.0.1/tcp/2428",
    "/ip4/192.168.1.42/tcp/2428",
    "/ip4/<relay>/tcp/4001/p2p/<relay>/p2p-circuit/p2p/<self>"
  ],
  "external_addrs": [
    "/ip4/<relay>/tcp/4001/p2p/<relay>/p2p-circuit/p2p/<self>"
  ],
  "relays": [
    {
      "peer_id": "12D3KooW...",
      "reservation_status": "Accepted",
      "circuit_addr": "/ip4/.../p2p-circuit/p2p/<self>",
      "last_state_change": "2026-05-26T17:30:00Z"
    }
  ],
  "rendezvous": [
    {
      "peer_id": "12D3KooW...",
      "namespace": "calimero/example",
      "registration_status": "Registered",
      "last_registered_at": "2026-05-26T17:30:00Z"
    }
  ],
  "direct_upgrades": [
    {
      "peer_id": "12D3KooW...",
      "status": "Succeeded | Failed { reason } | NotAttempted",
      "last_attempt": "2026-05-26T17:30:00Z"
    }
  ],
  "autonat": {
    "reachability": "Public | Private | Unknown",
    "last_test_addr": "/ip4/.../tcp/...",
    "last_test_result": "Reachable | Failed { reason }"
  }
}
```

### Plumbing required

1. **New per-peer field on `DiscoveryState.PeerInfo`** — `direct_upgrade_status: DirectUpgradeStatus { Succeeded { connection_id, at } | Failed { reason, at } | NotAttempted }`. Populated from the dcutr handler's event match (currently just logs).
2. **AutoNAT v2 client state** is already exposed via swarm events; capture last reachability result + tested addr.
3. **New `crates/server/src/admin/handlers/network_status.rs`** — read the swarm/discovery state, render JSON. Mirrors the existing pattern of `admin/handlers/tee/*` and other introspection endpoints.
4. **meroctl client** — `crates/meroctl/src/cli/network/status.rs`, table-formatted print, `--json` flag for raw output.

### Why this helps

- **Operators**: "Is my node behind NAT?" → one command, answer immediately.
- **CI / merobox**: the smoke tests can `meroctl network status` instead of grepping container logs for `ReservationReqAccepted` — strong typed assertion vs string match.
- **Future tests** can assert specific DCUtR outcomes per peer ("after this fault injection, peer X's upgrade should fail and the relay should take over"), which is currently impossible without log scraping.

### Related work

- Surfaced from the merobox#254 NAT-topology smoke test, which currently scans client logs for `ReservationReqAccepted` — would directly replace that grep with a typed API check.
- core#2487 added `info!`/`warn!` on DCUtR outcomes; this PR would persist + expose them.

### Scope notes

- Pure read-only endpoint; no auth changes needed beyond the existing admin-API gate.
- No new wire format beyond JSON over HTTP; meroctl already has the client plumbing pattern.
- DCUtR state tracking is the only new piece in the network crate — rest is renderers.

chefsale:
## Summary

- New `GET /admin-api/network/status` endpoint snapshots the swarm's libp2p connectivity state — local peer id, listen/external addresses, relay reservations, rendezvous registrations, DCUtR hole-punch outcomes per peer, and AutoNAT v2 reachability + last probe result.
- New `meroctl network status` subcommand renders the snapshot as comfy-table sections (or JSON via the existing root `--output-format json` flag).
- New per-peer state: optional `DcutrUpgradeStatus` (populated from the dcutr swarm-event handler) and single-slot `AutonatTest` on `DiscoveryState` (populated from the autonat client event). `PeerRelayInfo` / `PeerRendezvousInfo` gain a `last_state_change: Instant` bumped in their status setters.

Closes #2488.

## Why this matters

Until this PR, ops needed `RUST_LOG=debug` and `grep ReservationReqAccepted` to know whether a node's libp2p subsystem was healthy. That's fine for the developer who built the subsystem; bad for operators debugging "why is my home node not reaching the fleet?" and a blocker for merobox#254, which currently scrapes container logs to assert NAT-topology behaviour. With this endpoint, both flows get a typed API.

## Design notes

Two response shapes:
- **Internal** (`calimero_network_primitives::network_status::NetworkStatusSnapshot`) — typed `PeerId` / `Multiaddr` / `SystemTime`. This is the actor reply, what in-process tests will assert against.
- **Wire** (`calimero_server_primitives::admin::NetworkStatusResponse`) — all-strings + RFC3339 UTC timestamps. Stable across libp2p upgrades.

The admin handler does the conversion in a pure `render()` function, extracted so it can be unit-tested without spinning up an actor.

The DCUtR / AutoNAT log lines stay — this PR only adds typed state retention alongside them.

## Out of scope (potential follow-ups)

- No Prometheus metric for these fields.
- No `meroctl network watch` / streaming — single snapshot per call.
- No history beyond "last" for AutoNAT / DCUtR — bounded memory.
- merobox#254 swap from log-grep → typed assertion is the natural follow-up.

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo build --workspace --all-targets` succeeds
- [x] `cargo test -p calimero-network --lib discovery::state` — 24/24 pass, including 3 new tests for `record_dcutr_outcome`, `record_autonat_test`, and `last_state_change` advancement
- [x] `cargo clippy` clean on touched packages (pre-existing warnings in `calimero-wasm-abi`, `calimero-sdk-macros`, `calimero-context-config`, `crates/server/build.rs`, and `crates/network/src/discovery.rs:350` are unrelated to this PR)
- [x] `meroctl network --help` and `meroctl network status --help` render correctly
- [ ] End-to-end via merobox: boot a 2-node cluster, `meroctl --node node1 network status` shows expected relay/rendezvous state; AutoNAT transitions visible after connect
- [ ] `meroctl --output-format json --node node1 network status | jq` validates the wire keys match the issue sketch

---

> [!NOTE]
> **Low Risk**
> Read-only introspection and additive admin/CLI surface; discovery state extensions only record latest probe/upgrade outcomes without changing connection logic.
> 
> **Overview**
> Adds **operator-facing libp2p connectivity introspection** end-to-end: a read-only snapshot from the network actor through admin HTTP to the CLI.
> 
> **`GET /admin-api/network/status`** returns a stable JSON shape (string peer IDs, multiaddrs, lowercase status strings, RFC3339 timestamps) covering local identity, listen/external addresses, per-relay reservation state, per-rendezvous registration, per-peer **DCUtR** upgrade outcomes, and **AutoNAT v2** reachability plus the latest probe.
> 
> The stack introduces typed **`NetworkStatusSnapshot`** in network primitives, a **`NetworkMessage::NetworkStatus`** handler that walks the swarm and discovery state, and a pure **`render()`** conversion in the server. **`DiscoveryState`** now retains the latest AutoNAT probe, per-peer DCUtR results, and **`last_state_change`** on relay/rendezvous info (fed from autonat/dcutr swarm handlers). **`meroctl network status`** calls the new client API and prints table sections or JSON via existing output formatting.
> 
> Unit tests cover discovery state updates and wire rendering; **`chrono`** is added for timestamp formatting on the server.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fcc0fb58f214bf5ca0f70590ae8bdc16a809969b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1587,6 +1587,7 @@ dependencies = [
  "calimero-server-primitives",
  "calimero-store",
  "calimero-tee-attestation",
+ "chrono",
  "color-eyre",
  "eyre",
  "futures-util",
--- a/crates/client/src/client/system.rs
+++ b/crates/client/src/client/system.rs
@@ -2,7 +2,7 @@
 
 use calimero_server_primitives::admin::{
     FleetJoinRequest, FleetJoinResponse, GenerateContextIdentityResponse, GetPeersCountResponse,
-    InviteSpecializedNodeRequest, InviteSpecializedNodeResponse,
+    InviteSpecializedNodeRequest, InviteSpecializedNodeResponse, NetworkStatusResponse,
 };
 use eyre::Result;
 
@@ -24,6 +24,14 @@ where
 
     pub async fn get_peers_count(&self) -> Result<GetPeersCountResponse> {
         let response = self.connection.get("admin-api/peers").await?;
+        Ok(response)
+    }
+
+    /// Snapshot the local node's libp2p connectivity state — relays,
+    /// rendezvous registrations, DCUtR upgrade outcomes per peer, and
+    /// AutoNAT v2 reachability. Backs `meroctl network status`.
+    pub async fn network_status(&self) -> Result<NetworkStatusResponse> {
+        let response = self.connection.get("admin-api/network/status").await?;
         Ok(response)
     }
 
--- a/crates/meroctl/src/cli.rs
+++ b/crates/meroctl/src/cli.rs
@@ -24,6 +24,7 @@ mod context;
 mod dev;
 mod group;
 mod namespace;
+mod network;
 mod node;
 mod peers;
 mod tee;
@@ -37,6 +38,7 @@ use context::ContextCommand;
 use dev::DevCommand;
 use group::GroupCommand;
 use namespace::NamespaceCommand;
+use network::NetworkCommand;
 use node::NodeCommand;
 use peers::PeersCommand;
 use tee::TeeCommand;
@@ -94,6 +96,7 @@ pub enum SubCommands {
     #[command(alias = "ns")]
     Namespace(NamespaceCommand),
     Call(CallCommand),
+    Network(NetworkCommand),
     Peers(PeersCommand),
     Tee(TeeCommand),
     #[command(subcommand)]
@@ -169,6 +172,7 @@ impl RootCommand {
             SubCommands::Group(group) => group.run(&mut environment).await,
             SubCommands::Namespace(ns) => ns.run(&mut environment).await,
             SubCommands::Call(call) => call.run(&mut environment).await,
+            SubCommands::Network(network) => network.run(&mut environment).await,
             SubCommands::Peers(peers) => peers.run(&mut environment).await,
             SubCommands::Tee(tee) => tee.run(&mut environment).await,
             SubCommands::Node(node) => {
--- a/crates/meroctl/src/cli/network.rs
+++ b/crates/meroctl/src/cli/network.rs
@@ -1,0 +1,40 @@
+use clap::{Parser, Subcommand};
+use const_format::concatcp;
+use eyre::Result;
+
+use crate::cli::Environment;
+
+pub mod status;
+
+pub const EXAMPLES: &str = r"
+  # Dump this node's swarm connectivity state — listen/external addresses,
+  # relay reservations, rendezvous registrations, DCUtR upgrades, AutoNAT.
+  $ meroctl --node node1 network status
+
+  # Same, as JSON (use the root --output-format flag).
+  $ meroctl --output-format json --node node1 network status
+";
+
+#[derive(Debug, Parser)]
+#[command(about = "Inspect libp2p networking state")]
+#[command(after_help = concatcp!(
+    "Examples:",
+    EXAMPLES
+))]
+pub struct NetworkCommand {
+    #[command(subcommand)]
+    pub subcommand: NetworkSubCommands,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum NetworkSubCommands {
+    Status(status::StatusCommand),
+}
+
+impl NetworkCommand {
+    pub async fn run(self, environment: &mut Environment) -> Result<()> {
+        match self.subcommand {
+            NetworkSubCommands::Status(cmd) => cmd.run(environment).await,
+        }
+    }
+}
--- a/crates/meroctl/src/cli/network/status.rs
+++ b/crates/meroctl/src/cli/network/status.rs
@@ -1,0 +1,17 @@
+use clap::Parser;
+use eyre::Result;
+
+use crate::cli::Environment;
+
+#[derive(Copy, Clone, Debug, Parser)]
+#[command(about = "Snapshot libp2p connectivity: relays, rendezvous, DCUtR, AutoNAT")]
+pub struct StatusCommand;
+
+impl StatusCommand {
+    pub async fn run(&self, environment: &mut Environment) -> Result<()> {
+        let client = environment.client()?;
+        let response = client.network_status().await?;
+        environment.output.write(&response);
+        Ok(())
+    }
+}
--- a/crates/meroctl/src/output.rs
+++ b/crates/meroctl/src/output.rs
@@ -5,6 +5,7 @@ pub mod blobs;
 pub mod common;
 pub mod contexts;
 pub mod groups;
+pub mod network;
 pub mod tee;
 
 // Re-export common types
--- a/crates/meroctl/src/output/network.rs
+++ b/crates/meroctl/src/output/network.rs
@@ -1,0 +1,118 @@
+use calimero_server_primitives::admin::NetworkStatusResponse;
+use comfy_table::{Cell, Color, Table};
+
+use super::Report;
+
+impl Report for NetworkStatusResponse {
+    fn report(&self) {
+        // Identity table — peer id + every address the swarm currently knows about.
+        let mut identity = Table::new();
+        let _ = identity.set_header(vec![
+            Cell::new("Identity").fg(Color::Green),
+            Cell::new("Value").fg(Color::Blue),
+        ]);
+        let _ = identity.add_row(vec!["Local peer ID", &self.local_peer_id]);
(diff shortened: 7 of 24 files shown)


chefsale:
Pushed `0b87d84b` addressing the actionable findings. Walkthrough of what landed and what didn't:

## Fixed

- **ConnectionId rendering** (a404b1c469f2): switched from `format!("{cid:?}")` to `cid.to_string()`. Display on libp2p's `ConnectionId` writes the bare integer; Debug writes `ConnectionId(N)` which can shift across libp2p versions — using Display matches the stated wire-stability goal.
- **Unit tests for `render()`** (e204c816e2e5): added 6 tests covering empty snapshot, Reachable/Failed autonat, DCUtR Succeeded/Failed, and relay/rendezvous status string mapping. `render()` was extracted specifically to enable this; shipping the tests closes the loop.
- **Test sleep flakiness** (46cee8c42905): replaced the three `thread::sleep(2ms)` calls with a `wait_past(prior)` spin-loop bounded at 100ms. Safer on slow CI and on Windows where `Instant` resolution can be coarser than 1ms.

## Not fixed — with rationale

- **`.expect("Mailbox not to be dropped")`** (f4974b2aaac9, 830b64b08b72): this is the uniform pattern across **every** existing `NetworkClient` method (dial, listen_on, peer_count, mesh_stats, …). Making just `network_status()` return a `Result` would diverge from the codebase convention and create a special case to remember. If we want to harden against actor shutdown we should do it consistently across the client surface, not piecemeal in this PR.
- **`Instant → SystemTime` accuracy** (d883dacf9066, ff68eb93a635): switching to `UNIX_EPOCH + …` with checked_sub-to-zero would make stale records jump to 1970 — far more misleading than a "snapshot time" clamp. The current behavior (saturate forward to `now_system` on the future-edge race, document the limitation) is the right operator UX. The two reads happen back-to-back on the same task, so the gap is sub-microsecond.
- **Duplicate enum mirrors** (8f41971bf4f6): the two-layer split is deliberate — `network-primitives` is in the actor graph and pulls in libp2p, `server-primitives` is the HTTP-only wire crate consumed by meroctl and downstream clients. Collapsing them would either pull libp2p into the wire crate (bad) or expose libp2p types over HTTP (also bad). A `From` impl pair would be nice polish but the conversion already lives in `render()` and is now under test.
- **Recording `SystemTime` instead of `Instant`** (edb12e6f0233): `Instant` is monotonic; `SystemTime` can jump backwards (NTP correction, leap second, user clock change). Several existing call sites depend on monotonic semantics — `is_rendezvous_discover_throttled` and the relay reservation FSM compare instants for ordering. Switching the underlying clock would silently break those, far worse than the small read-side conversion approximation.
- **Output sorting** (fbfe177b5a81): `DiscoveryState.peers` is a `BTreeMap<PeerId, _>`, so iteration order is already deterministic (sorted by PeerId). No change needed.
- **`let _ =` on comfy-table calls** (9d3ed812afcb): matches the existing convention in `output/tee.rs` and `output/contexts.rs`. Diverging here would be the inconsistency. If we want to drop the prefix project-wide it's a separate sweep.
- **Auth on `/network/status`** (f6dca6d543ef): the route is registered inside the `protected_routes` router in `service.rs` (line 89), same group as `/peers`, `/usage`, and every other admin endpoint — auth middleware (`session_layer`) is applied at the router level at lines 326-327. Same protection as the existing endpoints.

The two ✅ "no longer detected" items were the same two `.expect()` findings from the first reviewer; the convention rationale above covers them too.

chefsale:
## Correction on the auth answer

Re-checked the auth wiring — my earlier reply on the `f6dca6d543ef` finding pointed at `session_layer`, which is wrong (that's just tower-sessions cookie storage). The actual auth flow is:

1. `protected_router` in `crates/server/src/admin/service.rs:89` collects every `/admin-api/*` route, including `/network/status`.
2. `mount_runtime_services` in `crates/server/src/service_mounts.rs:53` wraps that router with `auth::guard_layer(auth_service)`.
3. `AuthGuardService::call` (`crates/server/src/auth.rs:122-200`) validates either an `Authorization: Bearer <jwt>` header or a `?token=<jwt>` query param against `mero_auth::AuthService` and returns `401 UNAUTHORIZED` if missing/invalid.

So `/admin-api/network/status` **is** auth-gated, same as every other admin-api endpoint. JWT mandatory, no token → 401. The route lives in the same `protected_router` block as `/peers`, `/usage`, `/contexts`, `/blobs`, `/applications`, `/groups`, `/namespaces`, …

What this PR **does not** add — same as every other read-only admin-api endpoint — is a fine-grained permission check on top of authentication. The `permissions/validator.rs` permission map in `mero-auth` only covers a subset of admin routes (`/admin-api/applications`, `/admin-api/contexts`, `/admin-api/packages`, …); unmapped routes pass the permission gate by default for any authenticated caller. `/admin-api/peers` and `/admin-api/usage` are in the same tier today; `/network/status` matches their treatment. If the team wants admin-only gates for observability endpoints that's a broader policy change worth its own issue.

Sorry for the confusion in the earlier reply.

PR #2521, opened 2026-05-30
fix(governance): apply concurrent same-signer sibling ops out of order (#2516)

### LINKED ISSUE #2516: bug: concurrent same-signer governance ops can race the nonce-dedup guard
## Description

Follow-up to #2129. The general out-of-order governance-op problem is now handled by causal-DAG delivery: ops are wrapped as `CausalDelta`s (`crates/context/src/governance_dag.rs`) and buffered `Pending` by `DagStore` until their `parent_op_hashes` are present (`crates/dag/src/lib.rs`), so they apply strictly in causal order. The per-(group, signer) `nonce <= last` guard at `crates/governance-store/src/namespace/governance.rs:1174` then only acts as a dedup.

This relies on op N+1 always having op N (or a descendant) as a DAG ancestor. That holds when a signer authors sequentially. It does **not** hold if the same signer authors two consecutive-nonce ops *concurrently* — e.g. from two sessions / devices / processes that both read the same DAG head and assign nonces N and N+1 against it.

## Scenario

1. Signer S has `last_nonce = 4`, current DAG head = `H`.
2. Session A reads head `H`, authors op with nonce 5, parents `[H]`.
3. Session B reads head `H` (concurrently, before A's op propagates), authors op with nonce 6, parents `[H]`.
4. Both ops are DAG **siblings** — neither is the other's parent — so causal ordering imposes no order between them.
5. If nonce 6 is delivered/applied first, it advances `last_nonce` to 6; nonce 5 then hits `nonce <= last` and is skipped permanently.

Result is the same permanent-drop / divergence that #2129 described, just reachable only under concurrent same-signer authorship rather than plain network reordering.

## Impact

Low likelihood in normal single-session operation, but possible for any identity used from more than one place at once. Affects member additions, capability changes, and any group-scoped governance op.

## Files

- `crates/governance-store/src/namespace/governance.rs:1174` — `nonce <= last` dedup guard
- nonce assignment at author time (reads DAG head ⇒ assigns next per-signer nonce)

## Possible directions

- Detect sibling ops from the same signer at apply time and reconcile (apply the buffered lower-nonce op even though `last_nonce` was advanced) rather than relying solely on the global counter.
- Make per-signer nonce assignment serialize against in-flight (not-yet-applied) local ops so two sessions can't both mint against the same head.
- Track applied nonces as a set/window for a signer instead of a single high-water mark, so a later-arriving lower nonce is still applied once.

Splitting this out of #2129 (closed) so the residual edge stays tracked.

chefsale:
Fixes #2516.

## Problem

Governance ops carry a per-`(group, signer)` monotonic nonce, deduped by a single high-water mark:

```rust
if nonce <= last { /* drop as already-processed */ }
```

That guard is only correct when ops apply **in nonce order**, which causal DAG delivery guarantees *only when a signer authors sequentially* — op N+1 has op N as a DAG ancestor, so it's buffered until N applies.

It does **not** hold when the same signer authors two consecutive-nonce ops **concurrently** (two sessions/devices reading the same DAG head): the ops are DAG **siblings** — neither is the other's parent — so causal ordering imposes no order between them. If nonce N+1 lands first the mark advances to N+1, and the later-arriving nonce N then hits `nonce <= last` and is **dropped permanently**, diverging governance consensus.

## Fix — windowed applied-nonce tracking

Replace the single high-water mark with a `NonceWindow`:

- a contiguous `floor` — every nonce `1..=floor` is applied, and
- a sparse `BTreeSet` of applied nonces *above* the floor (the concurrency gap, bounded by in-flight concurrency).

A nonce is a duplicate iff it is `<= floor` **or** already in the above-set; otherwise it applies once and the floor advances through any now-contiguous run. The author mints `max_applied() + 1`, so a session that already saw out-of-order siblings doesn't re-mint a nonce sitting in the gap.

### Persistence (no migration)

The `floor` stays in the existing `GroupLocalGovNonce` (u64) key; the above-floor set goes in a new sibling key `GroupLocalGovNoncePending` (`Vec<u64>`, prefix `0x3C`). Old single-`u64` rows load as `floor` with an empty set and behave exactly like the previous guard. The floor is written before the above-set; a crash between the two writes is self-healing because `NonceWindow::new` normalises inconsistent parts on load.

## Wiring

- Dedup guards (`apply_local_signed_group_op`, namespace `apply_group_op_inner`) → `window.contains(nonce)`
- Advances → `window.record(nonce)` + persist
- Author (`sign_apply_local_group_op_borsh`) → `window.max_applied() + 1`

## Tests

- 7 `NonceWindow` unit tests (sequential, the 6-then-5 scenario, replay dedup, wide gap, normalise, backward-compat, empty).
- `apply_local_signed_group_op_out_of_order_siblings_2516` — local-apply path: deliver nonce 2 then nonce 1, assert both land, floor closes to 2, replays dedup, author mints 3.
- `replica_concurrent_sibling_ops_apply_out_of_order_2516` — namespace receive path (where concurrent contention is the norm): same out-of-order delivery through `apply_signed_op`.

Verified: `cargo build --workspace` clean, `cargo fmt --check` clean, clippy clean on new code. governance-store lib **346 passed** (+2), store lib 24, context convergence integration 23.

---

> [!NOTE]
> **High Risk**
> Changes core governance apply/dedup and nonce persistence on local and namespace receive paths; incorrect window logic could skip or double-apply ops, though tests and idempotent handlers mitigate replay cases.
> 
> **Overview**
> Fixes **#2516** by replacing per-signer governance nonce dedup (`nonce <= last`) with a **`NonceWindow`** (contiguous **floor** plus sparse **above** set) so concurrent same-signer DAG siblings with consecutive nonces can apply in either order without permanently dropping the lower nonce.
> 
> **`NonceWindow`** (`nonce_window.rs`) drives dedup via `contains`, advances via `record` after successful apply, and mints the next op with `max_applied() + 1`. Persistence moves to a single store value **`GroupLocalGovNonceWindow`** (floor + above in one atomic `put`), with read-only migration from legacy **`GroupLocalGovNonce`**.
> 
> **Apply paths** (`apply_local_signed_group_op`, namespace `apply_group_op_inner`) load the window, skip duplicates, record after mutations, and persist; local apply also **content-hash dedups** op-log appends on replay when the window was lost. **`GroupHandle::set_local_gov_nonce`** is removed from the public API; callers use `load_nonce_window` / `store_nonce_window`. Regression tests cover out-of-order siblings, replay dedup, migration, and round-trip storage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba517c6ebe01fbc2ec0651b36bb3b55611c94877. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed the review (commit 36d9d432):

**Write-ordering / non-atomicity (3 comments).** Real non-atomicity, but the severity was over-stated and the dangerous direction can't happen. Rewrote the `store_nonce_window` doc to state the actual invariant:
- The persisted `floor` is always `window.floor()`, which `record` advances *only* through applied nonces, written *after* apply — so it can never exceed the contiguous-applied count, and `contains` can never report an unapplied nonce as applied. **Dropping an op (the #2516 bug) is impossible.**
- The only reachable post-crash error is re-applying the *single in-flight* above-floor nonce (the above-set is rewritten in full each call, so older entries survive). That's the same "crash mid-commit" window the op-log entry write already has, and re-apply is tolerated by `op_log_contains_content_hash` + replay-safe mutations on the receive path.

Making it fully atomic would mean collapsing the window into one key — not warranted given the above.

**`record()` discarding `insert()` (2 comments).** Added the assert — but kept the `insert` *outside* `debug_assert!` (the suggested `debug_assert!(self.above.insert(...))` would elide the insert in release builds). Verified release-mode tests pass.

**nonce=0 invariant.** Added `debug_assert!(nonce > 0, …)`.

**Docstring backtick typo.** Fixed.

**`load_nonce_window` calls `to_bytes()` twice.** Stale — committed code already uses a single `gid`. No change.

**Unbounded `above` growth (2 comments).** Valid hardening item, low risk (requires a malicious admin; blast radius is one `(group, signer)` row). Filed as follow-up #2523 (reject-not-drop above a sanity cap).

chefsale:
Round 2 review addressed (commit 1bac8680).

**Note on the 09:38Z comments:** those are re-posts of the round-1 review re-anchored to the new commit — their bodies still reference `_inserted`, the backtick typo, and "`to_bytes()` twice," all of which were already fixed in 36d9d432. Resolving them as stale.

**The genuinely new (10:41Z) comments:**

- **Non-atomic write → double-apply (`48a7182a`, `c57b798f`).** Valid, and a real gap I under-sold earlier: `apply_local_signed_group_op` is a production path (DAG-delta application via `governance_dag`), and unlike the namespace receive path it had **no** content-hash op-log dedup. So a crash between the floor and above-set writes in `store_nonce_window` could drop an in-flight above-floor nonce, and a later DAG replay would append a **duplicate op-log entry**. Fixed by adding `op_log_contains_content_hash` dedup to the local path (parity with `apply_group_op_inner`): a replay now re-persists the advanced window and returns without appending. The non-atomic write is now a harmless deduped re-apply on **both** paths. Added a regression test (apply → roll floor back to simulate the lost-window crash → re-deliver → assert no duplicate entry).

- **`record()` return value always `true` at call sites (`57f45913`).** Dropped the `let _newly = record(...)` bindings on both paths (callers pre-check with `contains`, so the return is unused there).

- **Unbounded `above` (`bd8dae8a`).** Dup of round 1 — tracked as follow-up #2523 (reject-not-drop above a sanity cap).

Full governance-store lib suite: **347 passed**. Clippy + fmt clean; `calimero-context`/`calimero-node` build green.

chefsale:
Round 3 addressed (commit ecaec51a). I verified the claims against the code before acting — two of the six don't hold:

**Acted on:**

- **Mutation re-runs before dedup → could get stuck (`f2367fc0`).** Fair underlying ask. Verified the mutations are idempotent: `MemberAdded` → `MembershipRepository::add_member` is a `put` upsert (explicitly "used for upserts") that never errors on a duplicate — the same contract the namespace retry path already relies on. Documented it as a hard REPLAY-SAFETY CONTRACT on the local-apply path, and **upgraded the replay regression test to use `MemberAdded`** (not `Noop`) so it actually exercises re-applying a mutating op: no error, no duplicate entry.

- **`set_local_gov_nonce` bypasses the window (`81e1f0d1`).** It's test-only now (no production callers; the `GroupHandle::set_local_gov_nonce` wrapper was dead). Deleted the dead wrapper and documented both `get/set_local_gov_nonce` as floor-only helpers (not for the apply path). Kept the free fn for the crash-simulation test rather than `cfg(test)` re-export gymnastics.

**Clarified (comment was based on a misread):**

- **"All above nonces lost on crash" (`64288406`).** Not so: a crashed `put` doesn't execute, so the above-set key keeps its *prior persisted value*. Since each `record` adds at most one entry, only the single in-flight nonce is ever at risk — never the whole set. Tightened the doc to say this explicitly.

**No change (claim doesn't hold):**

- **Same nonce minted twice across sessions (`0023fd50`).** Doesn't happen: `sign_apply_local_group_op_borsh` *applies* the op it mints, so the minted nonce lands in the window (`above`). When the gap later closes, the floor advances past it and `max_applied()` reflects it — the next mint is strictly higher. The scenario assumes minting without recording, which isn't a reachable path.

- **`record()` return value (`6f3f7e9d`).** Kept returning `bool`: the unit tests use it as the standalone dedup gate; production callers pre-check with `contains` and ignore it. `#[must_use]` would just warn at those intentional discards.

- **`op_log_contains_content_hash` O(n) (`f1e82931`).** Already documented as O(n) in the fn's doc; it scans governance ops only (not state-DAG traffic), matching the existing readers. A by-hash index is possible later if governance logs ever grow, but isn't warranted now.

Full governance-store lib suite: **347 passed** (debug + release). Clippy + fmt clean.

chefsale:
Implemented **Option A** (commit ba517c6e) — the non-atomic two-key write that this review kept circling is now gone at the root, not just mitigated.

**Change:** the per-(group, signer) window (floor + above-set) is now a single store value (`GroupLocalGovNonceWindow` → `{ floor: u64, above: Vec<u64> }`), written with **one atomic `put`**. A crash leaves either the prior value or the new one — never a half-written floor-without-above state — so there is no longer any cross-key window in which an applied nonce can go missing.

- Renamed the 0x3C key `GroupLocalGovNoncePending` → `GroupLocalGovNonceWindow` and changed its value to carry the floor too; kept the legacy `GroupLocalGovNonce` (0x2C) **read-only** for one-way migration.
- `load_nonce_window`/`get_local_gov_nonce` read the window key, falling back to the legacy floor for pre-window databases (no data loss); first `store_nonce_window` makes the window key authoritative.
- Added tests: full window (incl. above-set) round-trips through the single key; legacy-floor migration loads then is superseded.

The op-log content-hash dedup added in the previous commits stays as a belt-and-suspenders backstop (it also covers replay from sources unrelated to crashes), but it's no longer load-bearing for crash safety.

Verified: governance-store lib **349 passed** (debug + release), store lib 24, clippy + fmt clean, full workspace build green.

PR #2524, opened 2026-05-30
feat: atomic CascadeUpgrade op + cascade_hlc HLC fence + get_cascade_status RPC

rtb-12:
## Summary

Builds on the merged namespace-cascade train (PR-1 #2477, PR-2 #2493) + #2507. Replaces the cascade's ordered two-op pair with a **single atomic `CascadeUpgrade` op** and adds the **post-`Completed` HLC fence** that rejects stale-schema state deltas after a migration. Ten focused commits, each TDD'd.

### What it does

1. **Atomic `CascadeUpgrade` GroupOp** (schema v6→v7) carrying `{from_app_key, app_key, target_application_id, migration, cascade_hlc}`. Replaces the ordered `CascadeTargetApplicationSet` + `CascadeGroupMigrationSet` pair, **fixing a latent receiver apply-order bug** (xilosada review of #2507, item #3): a receiver applying target-set before migration-set rewrote `app_key` so the migration-set predicate matched nothing and `migration` was silently dropped. A `#[ignore]`d **characterization test** documents that bug (asserts the dropped-migration outcome on the legacy two-op path); the atomic op makes it order-independent. Legacy variants retained one release for wire-compat (apply arms kept; doc-deprecated).
2. **Sticky `cascade_hlc`** on `GroupUpgradeValue` (hand-written backward-compatible borsh — legacy rows decode as `None`). Stamped **once** by the initiator (`env::hlc_timestamp()`), carried on the op, recorded identically on every matched descendant by the apply handler → a deterministic fence boundary across nodes.
3. **HLC fence**: each state delta now carries `producing_app_key` (the blob-derived app key the sender executed under); a migrated receiver drops a delta iff `producing_app_key != ctx GroupMeta.app_key && delta_hlc > cascade_hlc`. Applied (drop + warn + `delta_outcomes_total{outcome=fenced_stale_schema}`) in **both** receive paths (`apply_authorized_state_delta` and the snapshot-sync `replay_buffered_delta`); `producing_app_key` is threaded through the in-memory `BufferedDelta` so governance-pending drains don't bypass the fence.
4. **`get_cascade_status` RPC** exposed end-to-end (actor handler → `ContextMessage` → context client → admin HTTP `GET /groups/:namespace_id/cascade-status`), reusing the existing `GroupUpgradeInfo` DTO.
5. **Write-gate** generalized to refuse state-op writes (not just calls) while a group upgrade is `InProgress` (verified no migration-write deadlock: lazy writes `Completed` directly; eager migration goes via `update_application`, not the execute gate).

### Lockstep wire notes (pre-empting review)
- Two wire additions assume **lockstep merod deployment** (all nodes upgrade together), same assumption #2507's cascade `app_key` op made: the `CascadeUpgrade` GroupOp variant (appended last — existing borsh ordinals preserved) and the `producing_app_key` field on the transient `BroadcastMessage::StateDelta` (gossip, not persisted).
- `GroupUpgradeValue` IS persisted — its new field uses a hand-written `BorshDeserialize` that decodes pre-existing rows as `None` (tests: legacy-decode + partial-reject).

### Accepted limitations (documented in code, not silent)
- A sender store-error while resolving `producing_app_key` stamps `None` → that delta is unfenceable (liveness-over-strictness for a transient/local fault, logged at `warn!`).
- A `None` producing key (legacy/non-group contexts) and a not-yet-cascaded receiver (`cascade_hlc == None`) never fence — the coarse `pending_upgrade_target` sync-gate / lazy-upgrade covers that window. Gate-vs-fence division documented on `pending_upgrade_target` (disjoint time windows, no double-rejection).

### Verification
- Unit + integration green across touched crates: `calimero-store`, `-governance-types`, `-governance-store`, `-context` (incl. new `cascade_atomic_apply`, `cascade_status_rpc`, `hlc_fence`), `-node` (`sync_sim` 308, `fence_drop_tests`). Zero deprecation warnings; clean build of context/node/server.
- e2e workflow `01-namespace-cascade-migration` updated to assert `CascadeUpgrade: applied` on both nodes.

### Follow-ups (out of scope; not blocking)
- **Fence-rejection e2e** (`06-fence-rejects-straggler`) needs a merobox node-offline/straggler primitive (merobox#255) — deferred.
- **calimero-client-py** `get_cascade_status()` binding + a **merobox** `get_cascade_status` / `assert_cascade_complete` step (the RPC's downstream chain) — follow-up once this lands.

---

> [!NOTE]
> **High Risk**
> Changes group upgrade wire format (v7), migration convergence, and state-delta acceptance during/after cascade; lockstep node deployment is assumed for new gossip fields.
> 
> **Overview**
> Namespace cascade upgrades move from **two ordered governance ops** to a **single atomic `GroupOp::CascadeUpgrade`** (schema v7), carrying target app, `app_key`, migration bytes, and a **`cascade_hlc`** fence stamped once at the initiator. That removes the receiver bug where applying target-before-migration could drop migration on descendants; legacy two-op variants stay apply-only for one release.
> 
> **Persistence & observability:** `GroupUpgradeValue` gains optional **`cascade_hlc`** with backward-compatible borsh decode. **`GET /admin-api/groups/:namespace_id/cascade-status`** (client, actor, admin) returns per-group upgrade snapshots plus fence HLC across the namespace subtree.
> 
> **Safety during/after migration:** Execute now blocks **all** writes (including `__calimero_sync_next`) while upgrade status is **`InProgress`**, via `upgrade_blocks_write`. State deltas on the wire include **`producing_app_key`**; receivers drop stale-schema deltas when `producing_app_key ≠` local `app_key` and `delta_hlc > cascade_hlc`, including buffered/governance-drain and snapshot replay paths.
> 
> **Tests & workflows:** New unit/integration coverage for atomic apply, cascade status RPC, HLC fence, and an ignored characterization of the old two-op bug; e2e workflow asserts **`CascadeUpgrade: applied`** instead of the legacy log strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 229821fd2a92f9b2158a08699c9badd94e18b85b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rtb-12:
## Bug-bot triage (commit 0cc8fd69)

Thanks bots. Triage — one code fix, the rest are intentional with rationale:

**Fixed**
- 💡 *cascade_upgrade: migration not updated on existing upgrade record* — fixed in 0cc8fd69: `value.migration` is now set on the existing-record path too, so the record matches the authoritative `GroupMeta.migration` we just wrote.

**By design (no change)**
- 🟡 *TOCTOU between permission pre-scan and write (cascade_upgrade.rs)* — governance ops apply **sequentially, one at a time**; there is no concurrent op interleaving within a single `apply()`, so the pre-scan and writes are atomic w.r.t. other ops. This mirrors the existing `cascade_target_application_set` / `cascade_group_migration_set` modules verbatim.
- 🟡 *should_fence strict `>` at-boundary* — intentional: `delta_hlc == cascade_hlc` is pre-cascade legitimate history and must pass. HLCs are monotonic + node-unique (NTP64 + node id), so an exact-tie across distinct events is not a real collision. Covered by `does_not_fence_at_or_before_boundary`.
- 🟡 *get_cascade_status omits never-upgraded groups* — intentional: the map reports groups that **have a cascade/upgrade record**; a never-upgraded group has no cascade status to report. A caller reads "present + status" vs "absent = not part of this cascade."
- 🟡 *get_cascade_status auth gates on namespace-root membership* — a cascade is a namespace-root admin operation over the whole subtree; migration progress is low-sensitivity, and root membership is the natural gate (per-descendant gating would defeat the point of a subtree status view). Mirrors how the cascade itself is authorized.
- 🟡 *collect_descendants include/exclude* — verified `collect_descendants` **excludes** the start group; the explicit `vec![namespace_id]` prepend includes the root exactly once. Covered by the 3-group integration test.
- 🟡 *borsh EOF-as-None forward-compat* — current decode is correct (legacy rows → `None`, truncated `Some` → error; both tested). A future appended field would need its own trailing-optional handling in the same hand-written reader; noted for maintainers.

**Acknowledged (optional follow-up)**
- 📝 *test-helper duplication across cascade test files* — idiomatic for separate Rust integration-test crates; a shared `tests/common/mod.rs` is a reasonable cleanup, deferred to keep this PR focused.

Also added in 0cc8fd69: the `calimero-client` SDK method `get_cascade_status` (the surface meroctl / calimero-client-py / merobox call) — completes the RPC's client-side exposure.

chefsale (approved):

rtb-12:
## Follow-up on the latest bot round

Two new distinct points (rest are re-anchors of the triaged set above):

- 🟡 *migration field overwritten on existing records (cascade_upgrade.rs:85)* — note this is the inverse of the earlier "migration not updated" finding; the bot flagged both directions. Resolution: `GroupMeta.migration` is the authoritative source the migrate actually runs from, and the cascade op overwrites it via `set_group_migration`. Setting `GroupUpgradeValue.migration` to the same value keeps the tracking record consistent with `GroupMeta`. A new cascade upgrade supersedes any prior upgrade by design (concurrent-upgrade resolution is last-writer-wins / out of scope for this PR), so there is no in-progress state to preserve here.

- 🟡 *empty from_version/to_version on receiver-created records (cascade_upgrade.rs:114, cursor)* — intentional/known limitation: a node that only receives the gossiped `CascadeUpgrade` op does not carry the local semver strings (they come from `ApplicationMeta`, which the op does not include). The functional fields — `cascade_hlc` (the fence boundary) and `status` — are set correctly; the semver strings are best-effort and empty when unknown on a pure receiver. Richer cross-node version reporting would need the versions carried on the wire — noted as a follow-up, not blocking.

Resolving the threads accordingly.

PR #2527, opened 2026-05-30
fix(node/sync): bound snapshot generation memory to one page (#2133)

### LINKED ISSUE #2133: perf: generate_snapshot_pages loads all context keys into memory
## Description

In `snapshot.rs` at line ~655, `generate_snapshot_pages` iterates through ALL entries in `ContextStateKey`, collecting everything into a `Vec<([u8; 32], Vec<u8>)>`, then sorts it. The `start_cursor` pagination only applies after the full collection.

## Impact

For large contexts with millions of keys, this will exhaust memory. The pagination parameter is ineffective since all keys are loaded first.

## Files

- `crates/node/src/sync/snapshot.rs` — `generate_snapshot_pages` at line ~655

## Suggested Fix

Use an iterator-based approach with database-level cursor/seek to avoid loading all keys.

chefsale:
## Problem

Fixes #2133.

`generate_snapshot_pages` (`crates/node/src/sync/snapshot.rs`) collected **every** `ContextState` key→value pair for the context into an in-memory `BTreeMap`, built bundles for **all** entities past the cursor, and only paginated at the very end. A context with millions of keys therefore loaded its entire state into RAM on **every** paginated call — `start_cursor` only took effect *after* the full load, so pagination didn't bound memory at all.

## Why not a database-level seek?

The issue suggests replacing the collection with "database-level cursor and seek operations." That can't work literally: state keys are `Sha256(tag ‖ id)` (`StorageKey::to_bytes`), so storage iteration order is **uncorrelated** with the entity-id pagination cursor, and an entity's `Index` and `Entry` hash to scattered, non-adjacent keys. The cursor is an entity id (`SnapshotCursor.last_key`) that can't be mapped to a DB seek position. Discovering the sorted entity-id set inherently requires scanning the Index records each call.

The fix instead attacks what actually exhausts memory — **holding all the values**.

## Before / After

Consider a context with **2,000,000** entities (~1 KB each, ~2 GB of state), served with the defaults `page_limit = 16`, `byte_limit = 64 KB` (≈1 MB per burst).

**Before** — every paginated call:

```
generate_snapshot_pages(cursor = <page 137>)
  └─ scan State column
       └─ all_records: BTreeMap<[u8;32], Vec<u8>>   ← ~2 GB resident (ALL values)
  └─ build bundles for every entity past the cursor  ← clones ~2 GB more into `bundles`
  └─ paginate the bundles, emit ~1 MB, discard the rest
```
Peak resident memory ≈ **size of the whole context**, multiplied by the number of pages (the scan repeats every call). OOM on large contexts.

**After** — every paginated call:

```
generate_snapshot_pages(cursor = <page 137>)
  └─ scan State column (Pass 1)
       ├─ present_keys: HashSet<[u8;32]>   ← ~64 MB (keys only, 32 B × 2M)
       └─ entity_ids:   Vec<Id>            ← ~64 MB (32 B × 2M)
       (values deserialized to spot Index records, then dropped)
  └─ accounting pass: orphan/unrecognized/total_entries from key existence (no values)
  └─ emit pass: handle.get() Index+Entry values ONLY for the ~16 entities
                that land on this page, stop at page_limit
       └─ current page bytes                ← ≤ ~1 MB (one burst)
```
Peak resident value memory ≈ **one burst (~1 MB)**, independent of context size. Values go from GB → MB; the residual keys are O(records) but ~30× smaller than the values they replace.

The on-wire output (pages, cursor, `total_entries`) is **byte-for-byte identical** to before — this is purely a memory/throughput change.

## Read consistency

The per-bundle value lookups in the emit pass use `handle.get`, which reads the **live** store rather than the Pass-1 snapshot iterator (there is no snapshot-aware `get`). This is safe because `stream_snapshot_pages` already re-reads `root_hash` after generation and rejects the snapshot with `InvalidBoundary` if the context mutated in the meantime — so a torn read can never reach a peer. The guarantee is documented inline.

## Tests

`generate_snapshot_pages` had **no** direct coverage before (only `decode_snapshot_records` was tested). Added 4 unit tests over a real `InMemoryDB` store:

- empty context → no pages, no cursor, `total_entries == 0`
- single generous page round-trips all 20 entities, cursor signals done
- tight limits (1 page/call, 200-byte entries, 512-byte pages) across 50 entities: every entity emitted **exactly once** (no drops at page breaks, no cursor-overlap dupes), in canonical id-sorted order, with `total_entries` stable across every paginated call
- an orphan `Index` with no matching `Entry` is dropped and excluded from `total_entries`

To persist valid Index values from the node test crate, added a `#[doc(hidden)] pub fn EntityIndex::minimal_for_test(id)` constructor (the struct's fields are otherwise private).

✅ 157 sync lib tests green · `cargo fmt --check` clean · no new clippy findings on changed lines.

---

> [!NOTE]
> **Medium Risk**
> Large refactor of snapshot sync’s hot path; correctness relies on unchanged wire semantics and the existing root_hash boundary recheck when emit reads hit the live store.
> 
> **Overview**
> **`generate_snapshot_pages`** no longer loads every context state value into an in-memory map on each paginated call. Pass 1 scans once with a snapshot iterator, keeps only hashed keys and discovered entity ids (Index values are deserialized briefly for discovery, then dropped), and runs the existing orphan/`total_entries` accounting from key presence alone. Pass 2 uses **`partition_point`** to skip cursor-prefixed ids, then **`handle.get`** loads Entry/Index values only for entities that fit the current page burst, stopping at **`page_limit`** / **`byte_limit`**.
> 
> Iterator handling now **always unwraps** each entry’s value before filtering by context so a fused read error cannot truncate the scan silently. **`stream_snapshot_pages`**’ post-generation **`root_hash`** check is documented as the guard for live-store reads during emit.
> 
> Adds **`EntityIndex::minimal_for_test`** and **InMemoryDB** unit tests (empty context, single page, tight pagination completeness/order/stable **`total_entries`**, orphan Index drop, page-limit deferral). Wire output is intended to stay the same; this is primarily memory behavior (#2133).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a15ae6eff1db7cb3cdd641e5a04e8f659b92f3f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/snapshot.rs
+++ b/crates/node/src/sync/snapshot.rs
@@ -1,6 +1,6 @@
 //! Snapshot sync protocol for full state bootstrap.
 
-use std::collections::{BTreeMap, HashSet};
+use std::collections::HashSet;
 
 use borsh::BorshDeserialize;
 use calimero_crypto::Nonce;
@@ -828,17 +828,33 @@ struct SnapshotBoundary {
 /// authenticate state-bearing records.
 ///
 /// Discovery flow:
-/// 1. Iterate all `ContextStateKey` records for this context.
+/// 1. Iterate all `ContextStateKey` records for this context once,
+///    retaining only their 32-byte hashed state keys plus the
+///    discovered entity ids — never the record *values*. The
+///    earlier implementation collected every key→value pair into a
+///    map up front, so a context with millions of keys would pull
+///    all of its state into memory on every paginated call before
+///    the cursor was even consulted (issue #2133). Values are now
+///    materialised lazily, one entity at a time, only for the
+///    bundles that actually land on the requested page.
 /// 2. For each record value, attempt borsh deserialization as
 ///    [`calimero_storage::index::EntityIndex`]; success identifies
-///    the record as `Key::Index(id)` and yields the entity id.
-/// 3. For each discovered id, look up `Key::Entry(id)` and
-///    `Key::RotationLog(id)` records by their hashed state keys
-///    and bundle them.
-/// 4. Records whose state-key doesn't match any discovered
-///    `Index`/`Entry`/`RotationLog` slot are dropped as orphans
-///    (with a warning) — a well-formed state tree shouldn't have
-///    any.
+///    the record as `Key::Index(id)` and yields the entity id. The
+///    value is dropped immediately after the id is extracted.
+/// 3. Entity ids are sorted (the canonical pagination order) and
+///    the cursor skips everything already shipped. For each id that
+///    falls on this page, `Key::Index(id)` and `Key::Entry(id)` are
+///    point-looked-up to materialise the bundle.
+/// 4. State keys that don't match any discovered `Index`/`Entry`
+///    slot are dropped as orphans (with a warning) — a well-formed
+///    state tree shouldn't have any.
+///
+/// **Read consistency:** the discovery scan uses a snapshot
+/// iterator, but the per-bundle value lookups in step 3 hit the
+/// live store. That is safe because `stream_snapshot_pages`
+/// re-reads `root_hash` after generation and rejects the snapshot
+/// with `InvalidBoundary` if the context mutated in the meantime, so
+/// a torn read can never be shipped to a peer.
 fn generate_snapshot_pages<L: calimero_store::layer::ReadLayer>(
     handle: &calimero_store::Handle<L>,
     context_id: ContextId,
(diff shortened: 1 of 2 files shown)


chefsale:
Thanks for the reviews — addressed in `eadff353`.

**1. Live-store value read can truncate the scan (Bugbot, High) — fixed.**
Good catch, and the mechanism is real: `IterEntries::next` reads the value eagerly and fuses the iterator (`done = true`) on a read error, so dropping a foreign-context `value_result` without unwrapping swallowed the error and silently truncated the scan — an incomplete snapshot that still passes the per-context `root_hash` recheck. Pass 1 now unwraps `value_result?` **before** the context filter, restoring the pre-#2133 fail-loud behavior. (No extra read cost — the value is read eagerly inside `next()` regardless of whether we unwrap it.)

**2. Double iteration / cursor re-walk in the emit pass (MeroReviewer + AI Code Reviewer) — fixed.**
`entity_ids` is id-sorted, so cursor-skipped entities are a contiguous prefix. The emit pass now binary-searches past it with `entity_ids.partition_point(|id| *id.as_bytes() <= after)` and iterates only the tail, eliminating the per-element cursor check and per-id key re-hashing. On a resume near the end of a multi-million-entity context this drops the per-call emit work from O(N) to O(page). The accounting pass still walks the full set (it computes the cursor-independent `unrecognized_count` / `total_entries` diagnostics), which is by design.

**3. Stale/missing live-store value not observable (AI Code Reviewer) — addressed.**
The rare "value vanished between scan and read" skip (a concurrent delete) now emits a `warn!` rather than silently dropping the entity. On the ABA concern (delete then restore to the *same* root hash): for `handle.get` to return `None` while `root_hash` still equals the negotiated boundary, the state would have to be byte-identical to the boundary yet missing that record — i.e. a root-hash (merkle) collision, which is the same trust assumption the whole protocol already rests on. Correctness remains guarded by the post-generation `root_hash` recheck; the log just makes the skip visible if it ever fires.

Existing cursor-resume test (`test_generate_snapshot_pages_pagination_is_complete_and_dedup`) exercises the `partition_point` path on every paginated call and still asserts complete + dedup + id-sorted + stable `total_entries`.

chefsale:
Thanks — addressed in `809575a7`.

**1. Foreign-context value error aborts this context's snapshot (DoS framing).** Kept the fail-loud behavior, with an expanded comment explaining why it's the correct tradeoff rather than a bug:
- `IterEntries::next` reads the value eagerly and **fuses** the iterator (`done = true`) on the first read error, so "log the foreign error and continue" is not actually possible — the very next `next()` returns `None`, and we'd silently treat a *truncated* scan as complete. That silent-truncation-as-completion is exactly the data-loss bug the unwrap was added to prevent.
- A snapshot that fails with an error is retried and never ships partial state; a corrupt/unreadable record is already a serious storage fault. Failing loud is strictly safer than the alternative.
- I also considered exploiting the fact that a context's keys are contiguous (State column is ordered by `context_id ‖ state_key`) to early-break once past this context's range — but that trades a recoverable failure for an ordering assumption whose violation would *silently drop tail state*. Given the asymmetry, I deliberately kept the full scan. This is now documented at the call site.

**2. Page-break + page_limit boundary deserves a test.** Added `test_generate_snapshot_pages_page_limit_boundary_defers_entity`: `page_limit = 1` with a byte limit that holds one ~330-byte record but not two, so every call hits a page-break *and* the page limit simultaneously. It asserts each deferred entity reappears exactly once, in id-sorted order, across bursts, with stable `total_entries`. (The existing tight-pagination test already drove this path; this makes the intent explicit.)

**3. single-page test should assert one page.** Added `assert_eq!(pages.len(), 1, ...)`.

**4. `drain_all_pages` 10k cap could mask bugs.** It now tracks completion and asserts it drained to `cursor = None`, so hitting the cap fails with a clear "non-advancing cursor or unexpectedly many pages" message instead of a misleading "dropped entity".

**5. `id[0] = i >> 8` is always 0 for i < 256.** Correct — switched to `id[0] = i / 8; id[1] = i % 8` so the ids genuinely span two leading bytes and the sort-order assertion is meaningful.

chefsale:
Both new findings are the same point (emit-pass skip vs. `total_entries`) — addressed in `a15ae6ef` as documentation, because the discrepancy is by design and unobservable in any shipped snapshot:

- **`total_entries` never crosses the wire.** The `SnapshotPage` message carries only `payload`, `uncompressed_len`, `cursor`, `page_count`, `sent_count` — not `total_entries`. It's used in exactly one place: the sender's `info!(... total_entries, "Streaming snapshot")` log. So "could confuse the receiver's progress tracking" doesn't apply — the receiver never sees it.
- **The only delete that makes it overcount also rejects the snapshot.** A vanished Index/Entry between Pass 1 and the emit-pass `handle.get` is a concurrent delete, which moves `root_hash`; `stream_snapshot_pages` rechecks `root_hash` after generation and discards the whole snapshot (`InvalidBoundary`). So no *shipped* snapshot ever emits fewer records than `total_entries` reports — the overcount is confined to a log line for a generation that gets thrown away. (The "same-size/same-hash replacement" path reduces to a Merkle collision, which is the trust assumption the whole protocol already rests on.)
- **Decrementing would be worse.** `total_entries` is deliberately a full-context, scan-time count that's *stable across paginated calls* (an earlier per-page count shrank each call and misled operators). Decrementing on emit-pass skips would make it depend on which page window the call serves, breaking exactly that property.

So I documented `total_entries` as a scan-time, log-only figure at its computation site and explained the recheck guarantee, rather than changing behavior. Happy to revisit if you'd prefer it gated/removed from the log instead.

PR #2528, opened 2026-05-30
fix(network): surface rendezvous NoExternalAddresses as Pending state

### LINKED ISSUE #2490: chore(network): rendezvous_register silently treats NoExternalAddresses as success — surface as Pending state
Surfaced while closing #2475. The `rendezvous_register` function in `crates/network/src/discovery.rs` currently swallows `RegisterError::NoExternalAddresses` as a successful no-op:

```rust
if let Err(err) = self.swarm.behaviour_mut().rendezvous.register(...) {
    match err {
        RegisterError::NoExternalAddresses => {
            debug!("No external addresses to register at rendezvous");
            return Ok(());                                        // ← here
        }
        err @ RegisterError::FailedToMakeRecord(_) => {
            bail!(err)
        }
    }
}
```

## Why this is benign today (but was load-bearing before)

Before #2483 + #2487, this swallow was hiding a real problem: NAT'd nodes never accumulated an external address (relay-reservation was broken), so EVERY rendezvous register call returned `NoExternalAddresses` and `rendezvous_register` quietly reported success. The node thought it was registered with the rendezvous server but never actually advertised any address. Downstream peers couldn't discover it.

With #2483 + #2487, the chain works:

1. NAT'd node identifies a peer offering HOP → `create_relay_reservation` succeeds.
2. Relay reservation lands → swarm emits `ExternalAddrConfirmed` for the new `/p2p-circuit/` address.
3. `ExternalAddrConfirmed` handler calls `broadcast_rendezvous_registrations()`.
4. That iterates rendezvous peers and calls `rendezvous_register` — now the swarm HAS an external address, so the call succeeds for real.

The swallow is no longer load-bearing. But it's still misleading.

## Proposal

Demote `NoExternalAddresses` to `trace!`-or-suppressed (so non-developers don't see it) but mark the rendezvous peer's `registration_status` as `Pending` instead of `Requested`. That way:

- The discovery state machine reflects the actual state ("we tried, no external address yet, will retry on next ExternalAddrConfirmed").
- An operator running `meroctl network status` (#2488) sees the truth: registration is queued, not done.
- The next `ExternalAddrConfirmed` → `broadcast_rendezvous_registrations` already exists; it'll re-attempt and the swallow path becomes a one-off blip rather than a persistent lie.

## Scope

- `crates/network/src/discovery.rs::rendezvous_register` — split status updates, demote log level.
- `crates/network/src/discovery/state.rs` — possibly a new `RendezvousRegistrationStatus::Pending` variant if `Discovered` doesn't fit the semantics.
- Tests in `state_tests.rs` to pin the new transition.

## Not blocking

Low priority. The end-to-end behaviour is correct today; this is just hygiene. Mostly worth doing alongside #2488 (network-status command) so operators see accurate state.

chefsale:
Closes #2490.

## Problem

`rendezvous_register` swallowed `RegisterError::NoExternalAddresses` as a successful no-op: it logged at `debug`, left the rendezvous peer at `Discovered`, and returned `Ok`. The discovery state machine then reported "untried" when the node had actually attempted to register and was blocked waiting on an external address (e.g. a `/p2p-circuit/` relay address that hasn't landed yet). With `meroctl network status` (#2488) now surfacing this state to operators, that's a persistent lie — indistinguishable from "never got around to it".

## Fix

Introduce a `RendezvousRegistrationStatus::Pending` variant — *"tried, no external address yet, will retry on the next `ExternalAddrConfirmed`"*:

- `rendezvous_register` now marks the peer `Pending` (instead of silently leaving it `Discovered`) and demotes the log to `trace`.
- `Pending` does **not** occupy a registration slot in `is_rendezvous_registration_required`, so the fan-out gate keeps re-attempting until a real registration lands. The existing `ExternalAddrConfirmed`/`ExternalAddrExpired` → `broadcast_rendezvous_registrations` path handles the retry — no new wiring.
- `Pending` is a no-op in `rendezvous_unregister` (nothing was ever sent) and `find_new_rendezvous_peer` (already engaged with this peer).
- The response handler's `!= Requested` guard already correctly ignores any stray response for a `Pending` peer — no change needed there.
- Threaded `Pending` through the network-status command, `network-primitives`, and the admin handler (renders as `"pending"`).

## Behavior

End-to-end behavior is unchanged — the retry path was already correct. This only fixes what the state machine *reports*, so observability tells the truth instead of showing a misleading `discovered`. As the issue notes, this is hygiene best done alongside #2488, which has since landed.

## Tests

- New `test_pending_rendezvous_registration_does_not_occupy_slot` pins that `Pending` keeps the gate open while `Requested` closes it.
- `cargo test -p calimero-network --lib` → 63 passed.
- `cargo fmt --check` clean.

---

> [!NOTE]
> **Low Risk**
> Observability and state-machine accuracy around rendezvous registration; retry behavior stays on existing `ExternalAddrConfirmed` paths, with targeted unit tests.
> 
> **Overview**
> Adds **`Pending`** rendezvous registration status for peers where register was attempted but the swarm had **no external address** yet (`NoExternalAddresses`), instead of leaving them at **`discovered`** (misleading for operators and `meroctl network status`).
> 
> On that error, **`rendezvous_register`** calls **`mark_rendezvous_pending_if_idle`** so in-flight **`requested`** / live **`registered`** peers are not overwritten (avoids dropped confirmations and slot over-counting). **`Pending`** does not consume a fan-out slot; **`find_new_rendezvous_peer`** treats **`Pending`** like **`discovered`** for re-nomination when slots free. **`rendezvous_unregister`** treats **`Pending`** as a no-op. Status is exposed end-to-end as **`pending`** (primitives, network-status snapshot, admin API). Fixes the **`update_rendezvous_registration_status`** typo and moves peer nomination logic into **`DiscoveryState`**, with new unit tests for slot counting and nomination.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15f14a4d8d70d9375eb3c92f6b1c9b108734d177. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/network/primitives/src/network_status.rs
+++ b/crates/network/primitives/src/network_status.rs
@@ -59,6 +59,7 @@ pub struct RendezvousEntry {
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum RendezvousRegistrationKind {
     Discovered,
+    Pending,
     Requested,
     Registered,
     Expired,
--- a/crates/network/src/discovery.rs
+++ b/crates/network/src/discovery.rs
@@ -14,7 +14,7 @@ use libp2p::rendezvous::Namespace;
 use libp2p::swarm::dial_opts::{DialOpts, PeerCondition};
 use libp2p::PeerId;
 use multiaddr::{Multiaddr, Protocol};
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, trace};
 
 use super::NetworkManager;
 use crate::discovery::peer_cache::{PeerAddrCache, PersistedPeer};
@@ -437,8 +437,10 @@ impl NetworkManager {
     // co-members find us under the exact namespace/group/context key, so
     // steady-state discovery is relevant rather than network-wide.
     //
-    // If there are no external addresses for the node, registration is
-    // considered successful. Expects the rendezvous peer to be connected.
+    // If there are no external addresses for the node yet, the rendezvous
+    // peer is marked `Pending` (queued, not registered) and the call returns
+    // Ok; the next `ExternalAddrConfirmed` re-attempts. Expects the rendezvous
+    // peer to be connected.
     pub(crate) fn rendezvous_register(&mut self, rendezvous_peer: &PeerId) -> EyreResult<()> {
         // `registrations_limit` gates how many rendezvous *peers* we
         // register with (infra fan-out), independent of how many keys we
@@ -469,9 +471,16 @@ impl NetworkManager {
                 }
                 Err(RegisterError::NoExternalAddresses) => {
                     // No external addresses yet — nothing to register
-                    // anywhere this round; stop early, the next
-                    // reachability/identify event retries.
-                    debug!("No external addresses to register at rendezvous");
+                    // anywhere this round; stop early. Mark the peer Pending
+                    // so the discovery state reflects "tried, waiting on an
+                    // external address" rather than the misleading Requested.
+                    // The next ExternalAddrConfirmed fires
+                    // broadcast_rendezvous_registrations, which re-attempts.
+                    trace!(%rendezvous_peer, "No external addresses to register at rendezvous; marking Pending");
+                    self.discovery.state.update_rendezvous_registration_status(
+                        rendezvous_peer,
+                        RendezvousRegistrationStatus::Pending,
+                    );
                     return Ok(());
                 }
                 Err(err @ RegisterError::FailedToMakeRecord(_)) => bail!(err),
@@ -533,8 +542,10 @@ impl NetworkManager {
                     RendezvousRegistrationStatus::Expired,
                 );
             }
-            RendezvousRegistrationStatus::Discovered | RendezvousRegistrationStatus::Expired => {
-                // Nothing to unregister
+            RendezvousRegistrationStatus::Discovered
+            | RendezvousRegistrationStatus::Pending
+            | RendezvousRegistrationStatus::Expired => {
+                // Nothing to unregister (Pending never actually sent a record)
             }
         }
 
@@ -558,7 +569,8 @@ impl NetworkManager {
                         RendezvousRegistrationStatus::Expired if candidate.is_none() => {
                             candidate = Some(peer_id);
                         }
-                        RendezvousRegistrationStatus::Requested
+                        RendezvousRegistrationStatus::Pending
+                        | RendezvousRegistrationStatus::Requested
                         | RendezvousRegistrationStatus::Registered
                         | RendezvousRegistrationStatus::Expired => {}
                     }
--- a/crates/network/src/discovery.rs
+++ b/crates/network/src/discovery.rs
@@ -470,12 +470,27 @@ impl NetworkManager {
                     );
                 }
                 Err(RegisterError::NoExternalAddresses) => {
-                    // No external addresses yet — nothing to register
-                    // anywhere this round; stop early. Mark the peer Pending
-                    // so the discovery state reflects "tried, waiting on an
-                    // external address" rather than the misleading Requested.
-                    // The next ExternalAddrConfirmed fires
-                    // broadcast_rendezvous_registrations, which re-attempts.
+                    // `NoExternalAddresses` is a swarm-level condition: the
+                    // swarm has no confirmed external address to advertise,
+                    // which is independent of the rendezvous namespace. So
+                    // every remaining key in this loop would fail
+                    // identically — breaking early is safe, and it
+                    // necessarily fires on the *first* key, so
+                    // `registered_any` is always false here (no key went out
+                    // this round). Marking the peer Pending reflects "tried,
+                    // waiting on an external address" rather than the
+                    // misleading Requested; the next ExternalAddrConfirmed
+                    // fires broadcast_rendezvous_registrations, which
+                    // re-attempts.
+                    //
+                    // This may transition a previously Registered/Requested
+                    // peer to Pending (e.g. its external address expired and
+                    // a re-attempt now finds none). That is correct, not a
+                    // regression: with no external address there is nothing
+                    // backing any registration, and because the condition is
+                    // swarm-wide no other peer can hold a slot either, so the
+                    // freed slot cannot cause over-fan-out — all re-attempts
+                    // are Pending until an address returns.
                     trace!(%rendezvous_peer, "No external addresses to register at rendezvous; marking Pending");
                     self.discovery.state.update_rendezvous_registration_status(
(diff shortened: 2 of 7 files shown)


chefsale:
Good catch — fixed in `a652fc9d`. This was a regression introduced by the `Pending` state itself:

- **Before this branch:** a peer that couldn't register for lack of an external address stayed `Discovered`. `find_new_rendezvous_peer` returns `Discovered` immediately, so when a slot freed (another peer's registration `Expired`), the peer was re-driven.
- **On this branch (pre-fix):** that peer became `Pending`, and `Pending` was in the *skip* arm of `find_new_rendezvous_peer`. So it could strand: if the fan-out slot was already full when `ExternalAddrConfirmed` first fired, `broadcast_rendezvous_registrations` was gated out for it, and the later slot-free via `Expired` no longer reconsidered it — exactly the stall you describe.

**Fix:** treat `Pending` like `Discovered` in nomination (both are registerable targets that hold no fan-out slot), restoring the pre-`Pending` retry path. The `Expired`-handler path now re-drives a `Pending` peer when a slot frees, without waiting on another `ExternalAddrConfirmed`.

I also moved `find_new_rendezvous_peer` onto `DiscoveryState` (it only reads state) so the nomination policy is unit-tested directly: `test_find_new_rendezvous_peer_nominates_pending` and `test_find_new_rendezvous_peer_prefers_pending_over_expired` cover Pending nomination, Pending-over-Expired preference, and that Requested/Registered peers are never nominated.

chefsale:
Thanks — both addressed in `3138b860`.

**Discovered vs Pending priority (state.rs):** They are intentionally *equal* priority — first match in the (PeerId-sorted) iteration wins, no preference. I made that explicit in the method doc. The reviewer's suggested rationale for preferring `Discovered` ("avoid re-attempting a peer that already failed once") doesn't apply here: `Pending` isn't a failed peer, it's one blocked on a *transient* missing-external-address condition, so re-attempting it should rank no lower than a fresh `Discovered` peer. (It also matches the pre-`Pending` behavior, where this peer was `Discovered` and returned immediately.)

**Test "regardless of iteration order" (state_tests.rs):** Correct that the claim was imprecise — `get_rendezvous_peer_ids` walks a PeerId-ordered `BTreeSet`, so the outcome *is* order-sensitive unless the early return is what decides it. Made the test deterministic: it now sorts the two `PeerId`s and assigns the **Expired** role to the one that sorts first, so iteration records the Expired fallback candidate *before* reaching the Pending peer. That proves the eager early return — not iteration order — is what makes `Pending` win over an already-found candidate.

The remaining open thread is the original Cursor Bugbot "Pending skipped on expiry," which is the regression fixed in `a652fc9d` (the commit these suggestions are on) — resolving that thread now.

chefsale:
Addressed the remaining warnings/suggestions in `2edf5032`.

**Warnings #1 & #3 — early return on first `NoExternalAddresses` skips remaining keys.** Documented the assumption: `NoExternalAddresses` is a *swarm-level* condition (the swarm has no confirmed external address, independent of the rendezvous namespace), so every remaining key in the loop would fail identically and breaking early is safe.

**Warning #2 — `Registered`/`Requested` → `Pending` "regression".** Documented why this is correct rather than a regression. With no external address there is nothing backing any registration, and because the condition is swarm-wide *no* peer can hold a slot either — so the freed slot can't cause over-fan-out (every re-attempt is `Pending` until an address returns). Leaving a stale `Registered` would contradict the PR's whole point (honest status). So no downgrade guard — it's intended.

**Suggestion #3 — `registered_any` → `Requested` when some keys already registered.** Same swarm-level fact resolves it: `NoExternalAddresses` necessarily fires on the *first* key, so `registered_any` is always false in that branch; a "set Requested if registered_any" check would be dead code (and wrong — we have no external address). Documented inline.

**Suggestion #1 — typo `update_rendezvous_registartion_status`.** Fixed → `update_rendezvous_registration_status` (private `PeerInfo` helper + its call site).

**Suggestion #2 — no test for `Pending` → `Expired`.** Added `test_pending_to_expired_transition_keeps_slot_free`: asserts a `Pending` → `Expired` transition invents no registration slot (registration stays required at limit 1).

The earlier round (`3138b860`) covered the two `a652fc9d` suggestions (Discovered/Pending equal-priority doc + deterministic ordering test), and `a652fc9d` itself fixed the Cursor Bugbot "Pending skipped on expiry" regression.

chefsale:
Latest batch addressed in `15f14a4d`.

**Cursor Bugbot "Pending overwrites in-flight Requested" + AI reviewer "Pending may downgrade Registered/Requested" — fixed (and you were right; I was wrong to defend it earlier).** Bugbot's concrete trace is real: I confirmed the `Registered` event handler (`rendezvous.rs:93`) drops the confirmation unless status is still `Requested`. So a re-broadcast hitting a transient `NoExternalAddresses` while a register is in flight clobbered `Requested → Pending`, and the real `Registered` reply was then dropped — stranding the registration. The `Registered → Pending` case also frees a live fan-out slot.

Fix: `DiscoveryState::mark_rendezvous_pending_if_idle` only moves **idle** peers (`Discovered`/`Expired`/already `Pending`) to `Pending`; `Requested`/`Registered` keep their status and exit via their own `Expired` event. `rendezvous_register` now uses it. Added `test_mark_rendezvous_pending_if_idle_guards_slot_holders` covering all four source states. (This supersedes my earlier "no guard" reply — the in-flight-confirmation drop is a concrete failure, not just an accounting nuance.)

**AI reviewer — `prefers_pending_over_expired` depends on BTreeSet ordering.** Added `assert!(expired_peer < pending_peer)` up front so the test fails loudly if `get_rendezvous_peer_ids` ever stops iterating in ascending-PeerId order, rather than silently no longer exercising the early-return path.

**AI reviewer Warning — external IP via unauthenticated `https://api.ipify.org` (`get_public_ip`).** Leaving this **out of scope**: the reviewer itself notes it's pre-existing and not introduced by this PR (which is about rendezvous registration state, and doesn't touch `get_public_ip`). It's also a legitimate design question — spoofable IP discovery — that deserves its own change (e.g. validate against listen addresses or allow a configured static external IP) rather than being folded into this fix. I can open a separate issue to track it if you'd like.

chefsale:
Filed #2531 to track removing the `api.ipify.org` public-IP call (out of scope here, pre-existing). It proposes dropping `get_public_ip()` in favor of libp2p-native external-address discovery (identify `observed_addr` + AutoNAT) plus an optional operator-configured static address, and notes the `calimero-network` crate can drop its only `reqwest` dependency as a result.

PR #2532, opened 2026-05-30
fix(network): de-flake autonat_v2 test_dynamic_protocol_change

### LINKED ISSUE #1884: fix(network): flaky test autonat_v2::tests::test_dynamic_protocol_change
## Description

The test `autonat_v2::tests::test_dynamic_protocol_change` in `crates/network/primitives/src/autonat_v2/mod.rs` is flaky and intermittently fails.

## Error

```
thread 'autonat_v2::tests::test_dynamic_protocol_change' panicked at crates/network/primitives/src/autonat_v2/mod.rs:525:22:
Unexpected events: ([SwitchableNatEvent: Sent { ... }, ...], [...])
```

## Reproduction

```bash
cargo test -p calimero-network-primitives autonat_v2::tests::test_dynamic_protocol_change
```

The test passes most of the time but occasionally fails due to timing/ordering issues in the event sequence.

## Suggested Fix

The test appears to have a race condition where events arrive in an unexpected order. Consider:
- Adding event ordering tolerance
- Using more deterministic synchronization
- Increasing timeouts or adding retries

## Labels

- bug
- flaky-test
- network

chefsale:
Closes #1884.

## Problem

`autonat_v2::tests::test_dynamic_protocol_change` is intermittently flaky. It pinned `drive()` output to exact positional slice patterns over events from two concurrently-driven swarms. The relative order in which identify and autonat events interleave is timing-dependent, so any reordering fell through to `panic!("Unexpected events")` (the failure reported in the issue at `mod.rs:525`).

Note `drive()` collects a *fixed* number of events per swarm (the count is inferred from the array-pattern length), so the flake was never the count — only the brittle positional matching.

## Fix

Replace positional matching with presence / causal-order assertions at all three `panic!` sites, preserving the event counts (via explicit `[T; N]` annotations) and every protocol-advertisement assertion verbatim:

1. **Initial-state block** — assert `PeerHasServerSupport` appears anywhere in the collected events, then extract the two `Identify::Received` infos in arrival order. The two exchanges are causally sequential (the second only happens after the switchable gains the dial-back protocol), so indexing them is safe even when autonat events interleave around them.
2. **`ExternalAddrConfirmed`** — locate the confirmation by value rather than by slot.
3. **`ConnectionClosed`** — assert all drained events are `ConnectionClosed` regardless of order.

## Verification

- Compiles clean; `cargo fmt --check` clean; no new clippy warnings.
- 30/30 consecutive passing runs.

---

> [!NOTE]
> **Low Risk**
> Test-only changes to integration assertions; no production AutoNAT or swarm behavior is modified.
> 
> **Overview**
> **Stabilizes** `test_dynamic_protocol_change` in `autonat_v2/mod.rs` by replacing brittle exact-order `drive()` slice matching with assertions that tolerate timing-dependent event interleaving between two concurrent swarms.
> 
> After each `drive()` call, the test still drains the same fixed event counts (via explicit `[T; N]` types) but now checks **presence** (`PeerHasServerSupport`), **causal ordering** of the two `Identify::Received` rounds (protocol advertisement expectations unchanged), **value lookup** for `ExternalAddrConfirmed`, and **set-style** checks that disconnect drains are only `ConnectionClosed`. A small comment typo fix (`Conect againg` → connect again).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef625252f6ee1f772eb04594fdc2ea1681cf3bbb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/network/primitives/src/autonat_v2/mod.rs
+++ b/crates/network/primitives/src/autonat_v2/mod.rs
@@ -630,8 +630,7 @@ mod tests {
         switchable.behaviour_mut().autonat = Behaviour::new(cfg);
         switchable.behaviour_mut().autonat.enable_server().unwrap();
 
-        // Conect againg with the server
-        // let _addr = switchable.listen().await;
+        // Connect again with the server
         let _addr = server.listen().with_tcp_addr_external().await;
         switchable.connect(&mut server).await;
 
--- a/crates/network/primitives/src/autonat_v2/mod.rs
+++ b/crates/network/primitives/src/autonat_v2/mod.rs
@@ -464,78 +464,120 @@ mod tests {
 
         switchable.connect(&mut server).await;
 
-        // Test initial state: switchable is client-only (regular client server interaction)
-        match drive(&mut switchable, &mut server).await {
-            (
-                [SwitchableNatEvent::Identify(_), SwitchableNatEvent::Identify(identify::Event::Received {
-                    info: first_srv_info,
-                    ..
-                }), SwitchableNatEvent::Autonat(Event::PeerHasServerSupport { .. }), SwitchableNatEvent::Identify(_), SwitchableNatEvent::Identify(identify::Event::Received {
-                    info: second_srv_info,
-                    ..
-                })],
-                [ServerEvent::Identify(_), ServerEvent::Identify(identify::Event::Received {
-                    info: first_switchable_info,
-                    ..
-                }), ServerEvent::Identify(_), ServerEvent::Identify(identify::Event::Received {
-                    info: second_switchable_info,
-                    ..
-                })],
-            ) => {
-                // First exchange: Switchable sees server has dial-request
-                assert!(
-                    first_srv_info
-                        .protocols
-                        .iter()
-                        .any(|p| p.as_ref() == "/libp2p/autonat/2/dial-request"),
-                    "First exchange: Server should advertise dial-request, got: {:?}",
-                    first_srv_info.protocols
-                );
+        // Test initial state: switchable is client-only (regular client server interaction).
+        //
+        // `drive` collects a fixed number of behaviour events from each swarm, but the
+        // relative order in which identify and autonat events interleave is timing-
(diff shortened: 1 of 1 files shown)


chefsale:
Thanks for the reviews. Acted on both nitpicks in ef625252 (typo + commented-out `listen()` line).

On the three warnings about the fixed-size `drive()` window (1 from security-reviewer, 2 inline from meroreviewer) — they raise the same concern, so one reply:

The suggested mitigations are unfortunately not applicable to `libp2p_swarm_test::drive`, and the fix is strictly more robust than what was there before:

- **You cannot add margin to the window.** `drive` loops `while res.len() < N` and blocks until it has collected *exactly* `N` events, then `try_into()` the array. Setting `N` larger than the number of events actually emitted makes the test **hang forever**, not flake. So the count must equal the real emission count — it is not a tunable safety margin.
- **`wait(...)` can't replace `drive` here.** `wait` advances a single swarm. This step needs *both* swarms driven concurrently — the identify/autonat handshake only progresses when the server swarm is also polled. That's precisely why the original used `drive` and not `wait`. Waiting on the switchable alone would stall the server and deadlock the handshake.
- **The count was never the flake vector.** This handshake emits a deterministic *number* of behaviour events (identify Sent+Received per exchange; one `PeerHasServerSupport`); only their *interleaving* is timing-dependent. The original test already hard-coded exactly these counts (5 and 4) via positional slice patterns — this PR keeps the same counts and removes only the positional rigidity, which is the documented flake. If an event ever did fall outside the window, the new `assert_eq!(len, 2)` fails with a clear diagnostic (`got: {events:?}`) instead of an opaque `Unexpected events` panic.

Empirically: 30/30 consecutive green runs, no count mismatch observed. Decoupling from the fixed window entirely would require replacing `drive` with a custom dual-swarm pump — out of scope for a flaky-test fix and not warranted by any observed failure.

PR #2565, opened 2026-05-31
feat(storage): SortedMap + SortedSet collections with on-disk ordered index (#2559)

### LINKED ISSUE #2559: [SDK UX] Indexed/sorted collection for range + prefix queries
## Summary
An indexed/sorted collection type that maintains key order, unlocking native range and prefix queries (and pagination) without full scans.

## Current friction
Collections are **point-lookup or full-scan only.** Because keys are hashed, entity-id order ≠ key order, so range queries are impossible without loading and sorting the whole collection in memory (this is the same constraint behind the snapshot-pagination work — DB-seek is impossible when hashed keys don't match entity-id order). The migration scenarios even do a manual `keys.sort()` dance (`scenario-crdt-native-v2`) for deterministic ordering.

Any app wanting "entries 20–40" or "all keys with prefix `user:`" has to scan everything.

## Proposed
A `SortedMap` / indexed collection that maintains an ordered index, enabling:
- range queries (`range(start..end)`)
- prefix scans (`prefix("user:")`)
- efficient pagination

## SDK work / effort
New storage collection maintaining a CRDT-safe ordered index. Medium/High — architectural, scope separately from the ergonomics pass. Retires the in-memory-sort workarounds.

chefsale:
Closes #2559.

Adds the two key-ordered collections the SDK was missing — completing the quartet:

| | Unordered (`HashMap`/`HashSet`) | Sorted (`BTreeMap`/`BTreeSet`) |
|---|---|---|
| **Map** | `UnorderedMap` | **`SortedMap`** |
| **Set** | `UnorderedSet` | **`SortedSet`** |

Both maintain key order, unlocking **range queries, prefix scans, pagination, and min/max** without a full scan, retiring the manual `keys.sort()` workarounds (e.g. `scenario-crdt-native-v2`).

## Design

- **Same storage + CRDT as the unordered siblings.** Entries are the identical `Collection` of `(K, V)` / `V` keyed by `compute_id(parent, key)`, so add-wins merge, tombstones, and the #2538 nested-CRDT re-keying are inherited — nothing extra is synced.
- **A node-local, derived, *non-synced* ordered index** — a database-style secondary index in a dedicated RocksDB column (`Column::SortedIndex`), keyed by `collection ‖ order_key` *unhashed*, so the backend's byte order **is** the logical key order and a range is a native seek. Order is a pure function of `K: Ord`, so each node rebuilds its own index from the authoritative (synced) entries — there is **no extra merge path** and **no changes to the sync-critical `interface.rs`** (the index logic lives entirely inside the collection, like every other type).
- **Self-healing across sync.** Local writes keep the index warm and stamp a `full_hash` validity marker; a remote sync (which mutates entries host-side) leaves it stale, and the next ordered read rebuilds once (diffing, so only changed keys are written) then resumes seeking.
- **Transparent fallback.** Adaptors without an ordered keyspace (e.g. `PrivateStorage`) fall back to an in-memory sort — correct, just `O(n log n)`.

## Complexity (on a node)

| Operation | Cost |
|---|---|
| `range` / `prefix` | `O(log n + k)` |
| `page(offset, limit)` | `O(offset + limit)` |
| `first` / `last` | `O(log n)` (forward / reverse seek) |
| `get` / `insert` / `remove` | `O(1)` point op + index write + marker read/write |
| `entries`/`keys`/`values` / `iter` | `O(n)` |
| post-sync rebuild | `O(n)` reads, `O(changed)` writes |

`SortedMap`/`SortedSet` are the `BTreeMap`/`BTreeSet`: **default to the unordered types** (no per-write index overhead, no extra disk) and reach for the sorted ones only when you genuinely need ordering. This trade-off is documented at the top of each collection file.

## What's in here

- **Collections**: `crates/storage/src/collections/sorted_map.rs`, `sorted_set.rs` — full API (`range`/`prefix`/`page`/`keys`/`values`/`entries`/`first`/`last` + the standard map/set ops), `Entry` API, `Serialize`/`Eq`/`Ord`/`Extend`/`FromIterator`, `RekeyTarget`.
- **CRDT/ABI**: first-class `CrdtType::SortedMap`/`SortedSet` (primitives), `CrdtMeta`+`Mergeable`, `Decomposable`, merge dispatch, `CrdtCollectionType::SortedMap`/`SortedSet` (wasm-abi + normalize), SDK macros (`is_collection_type` + private-storage substitution), and the merodb ABI display.
- **Storage engine**: `Column::SortedIndex` + raw variable-length column ops (`Store::raw_put`/`raw_delete`/`raw_scan` with an offset+limit early-stop, `raw_last` reverse seek via `Database::last_in_range`/RocksDB `seek_for_prev`).
- **Host ABI**: the index ops wired through the full WASM boundary — `runtime::store::Storage` (+ `InMemoryStorage`), 5 host functions (`storage_index_set`/`remove`/`remove_prefix`/`scan`/`last`), `sys` externs, `sdk::env` wrappers, `storage::env` dispatch, `MainStorage` routing, and `ContextStorage` backing it over RocksDB (a second `Store` handle, scoped per context).
- **Apps**: new `apps/sorted-kv-store` example; `SortedMap`/`SortedSet` fields in the conformance apps (emit the ABI markers) and `scaffolding-e2e`.

## Verification

- **516 storage** + **12 CRDT-law (`crdt_contract`)** + **10 primitives** tests green; whole workspace builds.
- **Real WASM** (`crates/runtime/tests/crdt_conformance.rs`): drives the compiled `scaffolding-e2e` through `Module::run`, proving the ordered reads work through the actual host functions (`sorted_range`/`sorted_keys`/`sorted_last_key`) **and** that a receiving node rebuilds its index after a 2-node sync.
- **Asymptotic proof** (`ordered_reads_are_sublinear`): the (now faithful, seek-based) mock counts index entries examined — `last`=1, `page(0,10)`=10, a tight range window = its matches, vs `n` for a full scan.

---

> [!NOTE]
> **Medium Risk**
> Large new surface across storage, context execution, and WASM host boundaries; synced CRDT data paths are intended to match unordered types, but index staleness/rebuild and key byte-order contracts affect correctness of ordered reads.
> 
> **Overview**
> Adds **`SortedMap`** and **`SortedSet`** — key/element-ordered CRDT collections with the same add-wins merge as `UnorderedMap`/`UnorderedSet`, plus **`range`**, **`prefix`**, **`page`**, and **`first`/`last`** without full scans.
> 
> The ordering comes from a **node-local, non-synced** secondary index (`Column::SortedIndex`, unhashed `collection ‖ order_key`) maintained on writes and validated with a **`full_hash` marker**; after remote sync the next ordered read **diff-rebuilds** the index. Storage adaptors without an ordered keyspace (e.g. private storage) **sort in memory** instead.
> 
> The change threads through **primitives/ABI**, **merge dispatch**, **SDK macros** (deterministic collection IDs + private-storage substitution), **`StorageAdaptor` / `MainStorage`**, five new **WASM host imports** (`storage_index_*`), **`ContextStorage`** (per-context index keys + range delete), **`sdk::env`**, and a new **`apps/sorted-kv-store`** example plus conformance/e2e coverage including **real WASM** range/keys/last and post-sync rebuild tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca3d8135f0891f7e2592bac3ae24de35728225ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- **`SortedMap<K, V>` collection** — a key-ordered map for range queries, prefix scans, and pagination ([#2559])
+  - `range(a..b)`, `prefix("user:")`, `page(offset, limit)`, ascending `entries`/`keys`/`values`, and `first`/`last`
+  - Backed by a node-local, non-synced ordered secondary index (RocksDB `SortedIndex` column): `range`/`prefix` are `O(log n + k)` seeks and `page` is `O(limit)`. Adaptors without an ordered keyspace (e.g. `PrivateStorage`) transparently fall back to an in-memory sort
+  - Same add-wins CRDT merge as `UnorderedMap` — the index is a derived view of the synced entry set, not synced itself, so it adds no merge path and self-heals after a sync via a `full_hash` validity marker
+  - New `apps/sorted-kv-store` example; `SortedMap` ABI marker added to the conformance apps
+
 ## [0.10.1-rc.10] - 2026-03-30
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   - Backed by a node-local, non-synced ordered secondary index (RocksDB `SortedIndex` column): `range`/`prefix` are `O(log n + k)` seeks and `page` is `O(limit)`. Adaptors without an ordered keyspace (e.g. `PrivateStorage`) transparently fall back to an in-memory sort
   - Same add-wins CRDT merge as `UnorderedMap` — the index is a derived view of the synced entry set, not synced itself, so it adds no merge path and self-heals after a sync via a `full_hash` validity marker
   - New `apps/sorted-kv-store` example; `SortedMap` ABI marker added to the conformance apps
+- **`SortedSet<T>` collection** — the `BTreeSet` to `UnorderedSet`'s `HashSet`: an ordered set with `range`/`prefix`/`page`/`first`/`last`, same add-wins union merge and the same on-disk ordered index as `SortedMap` ([#2559])
 
 ## [0.10.1-rc.10] - 2026-03-30
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9769,6 +9769,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "sorted-kv-store"
+version = "0.0.0"
+dependencies = [
+ "calimero-sdk",
+ "calimero-storage",
+ "calimero-wasm-abi",
+ "serde_json",
+ "thiserror 1.0.69",
+]
+
+[[package]]
 name = "spin"
 version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ exclude = [
     "./apps/blobs",
     "./apps/collaborative-editor",
     "./apps/kv-store",
+    "./apps/sorted-kv-store",
     "./apps/kv-store-init",
     "./apps/kv-store-with-handlers",
 	"./apps/kv-store-with-user-and-frozen-storage",
@@ -106,6 +107,7 @@ members = [
 
     "./apps/scaffolding-e2e",
     "./apps/kv-store",
+    "./apps/sorted-kv-store",
     "./apps/kv-store-init",
     "./apps/kv-store-with-handlers",
 	"./apps/kv-store-with-user-and-frozen-storage",
--- a/apps/abi_conformance/abi.expected.json
+++ b/apps/abi_conformance/abi.expected.json
@@ -23,6 +23,24 @@
           }
         },
         {
+          "name": "sorted_counters",
+          "type": {
+            "kind": "map",
+            "key": {
+              "kind": "string"
+            },
+            "value": {
+              "kind": "record",
+              "fields": [],
+              "crdt_type": "lww_register",
+              "inner_type": {
+                "kind": "u32"
+              }
+            },
+            "crdt_type": "sorted_map"
+          }
+        },
+        {
           "name": "users",
           "type": {
             "kind": "list",
--- a/apps/abi_conformance/src/lib.rs
+++ b/apps/abi_conformance/src/lib.rs
@@ -3,7 +3,7 @@ use std::collections::BTreeMap;
 use calimero_sdk::app;
 use calimero_sdk::borsh::{BorshDeserialize, BorshSerialize};
 use calimero_sdk::serde::{Deserialize, Serialize};
-use calimero_storage::collections::{LwwRegister, UnorderedMap, Vector};
+use calimero_storage::collections::{LwwRegister, SortedMap, UnorderedMap, Vector};
 use thiserror::Error;
 
 // Test multi-file ABI generation
--- a/apps/abi_conformance/src/lib.rs
+++ b/apps/abi_conformance/src/lib.rs
@@ -118,6 +118,8 @@ pub enum Event {
 #[borsh(crate = "calimero_sdk::borsh")]
 pub struct AbiState {
     counters: UnorderedMap<String, LwwRegister<u32>>,
+    // Key-ordered map — locks the `SortedMap` ABI collection marker.
+    sorted_counters: SortedMap<String, LwwRegister<u32>>,
     users: Vector<LwwRegister<UserId32>>,
 }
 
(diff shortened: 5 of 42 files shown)


chefsale:
Addressed the review comments in 9b36b259:

**cursor[bot] — index stamp ignores write failures** (sorted_map/sorted_set + ContextStorage): index writes now return a persisted-or-not bool threaded all the way up (`ContextStorage`/`InMemoryStorage` → host fns → sdk/env wrappers → `StorageAdaptor` → the collections). The `full_hash` validity marker is stamped **only** when the write actually landed — a dropped write leaves the marker stale so the next ordered read rebuilds and self-heals. Same gate now applies to the incremental rebuild diff.

**cursor[bot] — index order differs from `Ord`**: documented the key-ordering contract on both collections. `K`/`V`'s `AsRef<[u8]>` encoding must be order-consistent with its `Ord` (holds for `String`/`&str`/`Vec<u8>`/`[u8; N]`; encode multi-byte ints big-endian / signed offset-binary). Index seeks use byte order while the fallback and comparison impls use `Ord`, so a mismatched encoding is a usage error, not a supported config.

**meroreviewer[bot] — `usize::MAX` limit sentinel** (sdk/env + runtime host fn): switched to an `n + 1` encoding with `0` = unbounded. `usize` is 32-bit on wasm32, so the old `MAX` didn't equal the host's `u64::MAX`; the new form is width-agnostic.

**meroreviewer[bot] — `index_del_prefix` unbounded scan + suggested DeleteRange**: replaced the scan-then-delete-each (which buffered the whole collection's keys) with a native RocksDB range tombstone — new `Database::delete_range` / `Store::raw_delete_range`. O(1) write, no per-key I/O, no buffer. Added a range-scoping rocksdb test.

**meroreviewer[bot] — hardcoded `32` prefix strip** (ContextStorage): the strip length is now derived from the live context id (`index_prefix_len()`) so write and readback stay in lockstep.

**meroreviewer[bot] — `0x00` bound translation** (store.rs index_range): reviewed — this is correct for the `[lo, hi)` composite keyspace (appending `0x00` yields the smallest key strictly greater than `collection ‖ k`, so inclusive-end / exclusive-start map correctly). No change.

**meroreviewer[bot] — rebuild `compute_id` recompute**: correct by construction — an entry's id is `compute_id(collection, key.as_ref())` and the rebuild's order key *is* that same `key.as_ref()`, so it can't disagree with the stored entry. Added a clarifying comment.

516 storage + runtime conformance + rocksdb tests green; `cargo check --workspace` clean; `cargo fmt --check` clean.

chefsale:
Second review round addressed in `ca3d8135`.

**Fixed:**

- **Index host-fn size limits (security)** — `storage_index_set/remove/remove_prefix/scan/last` now bound guest key/value/bound sizes against `max_storage_key_size`/`max_storage_value_size` (`HostError::KeyLengthOverflow`/`ValueLengthOverflow`), exactly like `storage_write`. A hostile guest can no longer drive unbounded host allocation through the index path.
- **`offset` truncation** — decoded with `usize::try_from(offset).unwrap_or(usize::MAX)` (saturate, don't silently truncate on a 32-bit host build).
- **Value-only update → stale marker → needless rebuild** — the common case (`insert` on an existing key) now restamps the validity marker after the write-back, since the key set is unchanged. New test `value_only_update_keeps_index_marker_current`. (See note below on the guard-returning paths.)
- **`ContextStorage::index_scan` swallowing scan errors** — now logs the error loudly instead of silently returning an empty page. The `Storage::index_scan` trait returns `Vec` (no error channel), and the authoritative entities are untouched + the index self-heals on the next read via the validity marker, so an empty page is a logged, recoverable degradation rather than silent data loss.

**Reviewed — no code change, with rationale:**

- **`[lo, hi)` bound translation (append `0x00`)** (flagged 3×) — this is *exact for all byte keys, including keys containing or ending in `0x00`*: no byte string `x` satisfies `k < x < k‖0x00`. I added an in-code proof. Note the suggested `prefix_upper_bound`-style successor would be **wrong** for an inclusive end bound: `succ([5]) = [6]` would wrongly admit `[5,0] > [5]`.
- **Unsafe guest-memory read without bounds validation** — `read_guest_memory_typed`/`read_guest_memory_slice` already validate the buffer descriptor and slice against the WASM linear-memory bounds; this is the same vetted pattern every host fn uses, not a new exposure.
- **`index_scan` off-by-one if prefix-strip drops rows** — every key in `Column::SortedIndex` is written through `index_key()` with the context prefix, so `k.get(plen..)` is guaranteed `Some`; no rows can be dropped in practice.
- **Value-only updates via `get_mut`/`ValueMut`/`OccupiedEntry`, `OccupiedEntry::remove`, and bulk `Extend`** — these mutate *after* the collection method returns (the guard is dropped in the caller), so they can't restamp the marker inline. They remain **correct** because the next ordered read detects the marker mismatch and rebuilds (self-heal); the cost is one O(n) rebuild. I optimized the common upsert path (`insert`) where it's cleanly possible and left the guard-returning paths to self-heal rather than add fragile post-drop hooks.

All green locally: storage + runtime + context suites, `cargo clippy --all-targets`, `cargo fmt --check`.

PR #2566, opened 2026-05-31
refactor(apps): standardize example apps on app::Result and enforce in app::logic

### LINKED ISSUE #2543: [SDK UX] Standardize example apps on app::Result to drop .map_err noise
## Summary
Standardize all example apps on `app::Result<T>` so a bare `?` propagates storage errors, eliminating the `.map_err(|e| format!("...{:?}", e))?` noise.

## Current friction
`From<StoreError> for app::Error` **already exists** — `kv-store` relies on it. But `team-metrics-*` and `nested-crdt-test` declare `Result<T, String>` and opt out, forcing manual wrapping on every call. From `apps/team-metrics-custom/src/lib.rs:97-125` there are ~28 `.map_err()` calls in one file:

```rust
pub fn record_win(&mut self, team_id: String) -> Result<u64, String> {
    let mut stats = self.teams.get(&team_id)
        .map_err(|e| format!("Get failed: {:?}", e))?
        .unwrap_or_else(|| TeamStats { wins: Counter::new(), losses: Counter::new(), draws: Counter::new() });
    stats.wins.increment().map_err(|e| format!("Increment failed: {:?}", e))?;
    let total = stats.wins.value().map_err(|e| format!("Value failed: {:?}", e))?;
    drop(self.teams.insert(team_id.clone(), stats).map_err(|e| format!("Insert failed: {:?}", e))?);
    app::emit!(MetricsEvent::WinRecorded { team_id, total });
    Ok(total)
}
```

## Proposed
```rust
pub fn record_win(&mut self, team_id: String) -> app::Result<u64> {
    let mut stats = self.teams.entry(team_id.clone())?.or_default()?;
    stats.wins.increment()?;
    let total = stats.wins.value()?;
    app::emit!(MetricsEvent::WinRecorded { team_id, total });
    Ok(total)
}
```

## SDK work / effort
**Zero SDK change** for the error part — purely "stop using `Result<T, String>` in examples." Add a CI lint / doc rule that example app methods must return `app::Result`. Low. (Combines well with #1 and #3.)

One of the highest-value, lowest-effort wins.

chefsale:
Closes #2543.

## Problem

Example apps declared `Result<T, String>` and hand-wrapped every storage call with `.map_err(|e| format!("...{:?}", e))?` — ~28 times in a single file. The SDK already provides `From<StoreError> for app::Error`, so returning `app::Result<T>` lets storage errors propagate through `?` with zero boilerplate.

## Changes

**Converted 6 apps** (`blobs`, `collaborative-editor`, `nested-crdt-test`, `team-metrics-macro`, `team-metrics-custom`, `scaffolding-e2e`):
- `-> Result<T, String>` → `-> app::Result<T>`
- `.map_err(|e| format!(...))?` → `?`; tail maps → `.map_err(Into::into)`
- custom messages → `app::bail!` / `app::err!`
- error-swallowing `if let Ok(entries) = ...` → proper `?` propagation
- READMEs updated to match the new signatures

The other apps (`kv-store*`, `private_data`, `sync-test`, migration/conformance fixtures) already used `app::Result`.

**Enforcement in the `#[app::logic]` macro** (the issue's "additional work"), instead of a CI grep:
- A public logic method returning a bare `Result<T, String>` now **fails to compile** with an actionable message pointing to `app::Result` + `app::bail!`/`app::err!`.
- Runs automatically at compile time for every app — no separate CI step, no second source of truth.
- Precisely scoped: `app::Result<T>` (single type arg) and custom error types pass; helper fns and trait impls (`Mergeable::merge`) are never flagged because the check only sees public logic methods.
- Covered by unit tests in `logic::method::tests`.

## Verification
- All 6 apps `cargo check` clean and `cargo fmt --check` clean against latest master
- `calimero-sdk-macros`: 48/48 tests pass (incl. 3 new), fmt-clean, no new clippy
- Integration-tested: reverting a method to `Result<T, String>` produces `error: (calimero)> app logic methods must return app::Result<T>, not Result<T, String>`, and compiles cleanly once fixed
- No other `#[app::logic]` block in the repo returns a `String` error

---

> [!NOTE]
> **Low Risk**
> Mechanical error-type and propagation changes in example/e2e apps plus a compile-time guard; no auth, payment, or core runtime behavior changes beyond surfacing storage errors that were previously swallowed.
> 
> **Overview**
> Example app logic and docs now use **`app::Result<T>`** instead of **`Result<T, String>`**, so storage failures propagate with **`?`** (via **`From<StoreError> for app::Error`**) instead of hand-written **`.map_err(|e| format!(...))?`**. Custom cases use **`app::bail!`** / **`app::err!`**, and patterns that ignored storage errors (**`if let Ok(entries) = ...`**) now fail the call on error.
> 
> **Six apps** are updated: **`blobs`**, **`collaborative-editor`**, **`nested-crdt-test`**, **`team-metrics-macro`**, **`team-metrics-custom`**, and **`scaffolding-e2e`** (including blob ID parsing and small comments on base58 helpers). Matching **README** API signatures were refreshed.
> 
> The **`#[app::logic]`** macro **rejects** public methods that return bare **`Result<T, String>`** at compile time, with a message pointing to **`app::Result`** and the error macros; **`app::Result<T>`** and other error types are unchanged. Unit tests cover the detector in **`logic/method`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fa16a0691b4774ad83f3b4fdcfb082fc4f1b690. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/blobs/README.md
+++ b/apps/blobs/README.md
@@ -75,7 +75,7 @@ upload_file(
     blob_id: BlobId,  // Blob ID (base58 string over the wire)
     size: u64,
     mime_type: String
-) -> Result<String, String>
+) -> app::Result<String>
 ```
 
 **Process:**
@@ -99,7 +99,7 @@ let file_id = state.upload_file(
 ### Delete a File
 
 ```rust
-delete_file(file_id: String) -> Result<(), String>
+delete_file(file_id: String) -> app::Result<()>
 ```
 
 Removes the file record from storage and emits a deletion event.
@@ -107,7 +107,7 @@ Removes the file record from storage and emits a deletion event.
 ### List All Files
 
 ```rust
-list_files() -> Result<Vec<FileRecord>, String>
+list_files() -> app::Result<Vec<FileRecord>>
 ```
 
 Returns all stored files with their metadata.
@@ -115,7 +115,7 @@ Returns all stored files with their metadata.
 ### Get Specific File
 
 ```rust
-get_file(file_id: String) -> Result<FileRecord, String>
+get_file(file_id: String) -> app::Result<FileRecord>
 ```
 
 Retrieves a single file's metadata by ID.
@@ -123,7 +123,7 @@ Retrieves a single file's metadata by ID.
 ### Get Blob ID
 
 ```rust
-get_blob_id_b58(file_id: String) -> Result<BlobId, String>
+get_blob_id_b58(file_id: String) -> app::Result<BlobId>
 ```
 
 Returns the base58-encoded blob ID for a file (useful for downloading).
@@ -131,7 +131,7 @@ Returns the base58-encoded blob ID for a file (useful for downloading).
 ### Search Files
 
 ```rust
-search_files(query: String) -> Result<Vec<FileRecord>, String>
+search_files(query: String) -> app::Result<Vec<FileRecord>>
 ```
 
 Case-insensitive search by filename.
@@ -139,8 +139,8 @@ Case-insensitive search by filename.
 ### Statistics
 
 ```rust
-get_stats() -> Result<String, String>
-get_total_files_size() -> Result<u64, String>
+get_stats() -> app::Result<String>
+get_total_files_size() -> app::Result<u64>
 ```
 
 Get usage statistics and total storage.
--- a/apps/blobs/src/lib.rs
+++ b/apps/blobs/src/lib.rs
@@ -149,26 +149,21 @@ impl FileShareState {
     ///
     /// # Returns
     /// * `Ok(String)` - The generated file ID (e.g., "file_0", "file_1")
-    /// * `Err(String)` - Error message if storage operation fails
+    /// * `Err(app::Error)` - Error if storage operation fails
     pub fn upload_file(
         &mut self,
         name: String,
         blob_id: BlobId,
         size: u64,
         mime_type: String,
-    ) -> Result<String, String> {
+    ) -> app::Result<String> {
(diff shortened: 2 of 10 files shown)


chefsale:
Addressed the reviewer findings in 8fa16a06:

1. **String error detector only checks last path segment** — `is_std_string` now matches the error type only against the standard-library `String` spellings (`String`, `std::string::String`, `alloc::string::String`) with no qualifying prefix, so a user type whose last segment is `String` (e.g. `my_crate::String`) is no longer a false positive. Added tests for the qualified spellings and the false-positive cases. The type-alias false-negative (`type R<T> = Result<T, String>`) is inherent to a syntactic proc-macro check and is now documented on the function.

2. **`.unwrap_or(0)` silently swallows Counter value errors** — the getters now propagate storage errors via `?` instead of defaulting: `get_counter`, `get_g_counter`, `get_pn_counter`, `has_tag`, and `get_wins`/`get_losses`/`get_draws`. Only the missing-key case bails. (These were pre-existing, but since the methods now return `app::Result` they propagate cleanly.)

3. **Unwrap in `encode_blob_id_base58` can panic** — added a comment explaining why both unwraps are infallible for a 32-byte input (base58 of 32 bytes is ≤44 chars = the buffer size, and the base58 alphabet is ASCII so the bytes are always UTF-8).

PR #2567, opened 2026-05-31
refactor(apps): drop the drop(insert(...)) papercut (#2556) + run CI on apps/**

chefsale:
## Summary

Two SDK-UX papercut cleanups for the example apps, plus a CI gap fix surfaced along the way.

> **Note:** this PR was rebased after #2566 merged. #2566 already did the `app::Result` conversion (**#2543**) across these and other apps, so that half is no longer part of this diff. What remains here is **#2556** (removing the `drop(insert(...))` papercut) plus the CI fix below.

## #2556 — drop the `drop(insert(...))` noise

Removes the `drop(self.x.insert(k, v)?)` wrappers in the three example apps → bare `self.x.insert(k, v)?;`.

Resolved **without** adding a `set()` method or changing the storage crate. The issue assumed `UnorderedMap::insert` is `#[must_use]`, but it is not, and the crate enables no `unused_results` lint — so `self.x.insert(k, v)?;` as a bare statement already compiles **warning-free**, exactly like `std::collections::HashMap::insert`. The fix is simply to delete the `drop(...)` wrappers; the displaced-value return is ignored as std allows. This keeps the API surface minimal (no `insert` vs `set` "which do I use?" papercut). The `let _ = set.insert(...)` in `nested-crdt-test` is normalized to a bare statement too, for consistency.

Verified: bare inserts produce **no** `unused_must_use` / `unused_results` warning in any of the three apps.

**Before**
```rust
drop(self.teams.insert(team_id.clone(), stats)?);
```
**After**
```rust
self.teams.insert(team_id.clone(), stats)?;
```

## CI — run real `ci-checks` on `apps/**` changes

While verifying this PR I found that the required **"Rust"** check is gated by a path-filter + stub pattern: `ci-checks.yml` runs the real build/test/clippy/fmt only for paths in its `paths:` list, while `ci-checks-ignored.yml` is a no-op stub reporting the same check name for everything else. `apps/**` was in **neither** real list, so PRs touching only example apps had "Rust" satisfied by the stub in ~3s — nothing was actually compiled or tested.

This adds `apps/**` to `ci-checks.yml` `paths:` and `ci-checks-ignored.yml` `paths-ignore:` (kept in sync per the note in those files). App-only PRs now run the real workspace build + tests, which already cover these apps (they are workspace members and `build-all-apps.sh` builds their wasm). You can see the effect on this PR: the real `Rust` job now runs on `big-machine` instead of only the 3s stub.

[...]

Issue #2556, opened 2026-05-31
[SDK UX] Remove drop(insert(...)) papercut: add void-returning set()


chefsale:
## Summary
Remove the `drop(self.x.insert(...))` papercut — either add a void-returning `set(k, v)` alongside `insert` (which returns the old value), or drop the `#[must_use]`.

## Current friction
`insert` returns the previous value and is `#[must_use]`, so apps wrap nearly every insert in `drop(...)` purely to discard it — confusing to newcomers and visually noisy. From `apps/team-metrics-custom/src/lib.rs` and `apps/nested-crdt-test/src/lib.rs`:

```rust
drop(self.teams.insert(team_id.clone(), stats)
    .map_err(|e| format!("Insert failed: {:?}", e))?);
```

## Proposed
```rust
self.teams.set(team_id.clone(), stats)?;   // returns (), no drop needed
```

## SDK work / effort
Add a `set` (or `put`) that returns `()` alongside `insert`. Low. Mostly disappears once the entry-guard (#1) lands, but still wanted for direct inserts.

Issue #2545, opened 2026-05-31
[SDK UX] #[app::event] / #[app::state] should emit their own serde/borsh derives


chefsale:
## Summary
Make `#[app::event]` emit its own serialization derives so authors stop hand-adding `serde::Serialize` / `#[serde(crate=...)]` / borsh inconsistently. Same idea for the `#[borsh(crate = "calimero_sdk::borsh")]` line on every state struct.

## Current friction
Two different incantations exist across apps. `apps/kv-store/src/lib.rs:19-25` keeps it minimal:

```rust
#[app::event]
pub enum Event<'a> { Inserted { key: &'a str, value: &'a str }, /* ... */ }
```

…but `apps/nested-crdt-test/src/lib.rs:38-41` redundantly adds borsh:

```rust
#[app::event]
#[derive(Debug, BorshSerialize, BorshDeserialize)]
#[borsh(crate = "calimero_sdk::borsh")]
pub enum TestEvent { /* ... */ }
```

And every state struct repeats `#[borsh(crate = "calimero_sdk::borsh")]` (kv-store:13, nested:21, team-metrics:62…).

## Proposed
```rust
#[app::event]
pub enum TestEvent {
    CounterIncremented { key: String, value: u64 },
    RegisterSet { key: String, value: String },
}
```

## SDK work / effort
`#[app::event]` injects the required `Serialize` + serde-crate path (and borsh where actually needed); `#[app::state]` injects the borsh-crate attribute it already assumes. Small (macro-only).

PR #2568, opened 2026-05-31
feat(sdk): in-process unit-test harness (TestHost) for app logic

### LINKED ISSUE #2551: [SDK UX] Native in-process unit-test harness (TestHost) for app logic
## Summary
A native, in-process unit-test harness so app logic can be tested without building WASM and spinning up merobox/Docker.

## Current friction
**0 of the example apps contain a single `#[cfg(test)]` or `#[test]`** (verified across all `apps/*/src/lib.rs`). The only way to test app logic today is build WASM → start merobox containers → run a YAML workflow — a multi-second, Docker-heavy loop for what should be a millisecond assertion. There's nothing to demonstrate TDD on.

## Proposed
A mock host behind the `env::` calls (storage map, `time_now`, `executor_id`, `context_id`, event capture):

```rust
#[cfg(test)]
mod tests {
    use calimero_sdk::testing::TestHost;

    #[test]
    fn increment_sums() {
        let mut app = TestHost::init::<NestedCrdtTest>();
        app.call(|s| s.increment_counter("a".into())).unwrap();
        app.call(|s| s.increment_counter("a".into())).unwrap();
        assert_eq!(app.view(|s| s.get_counter("a".into())).unwrap(), 2);
        assert_eq!(app.emitted_events().len(), 2);
    }
}
```

## SDK work / effort
A `calimero_sdk::testing::TestHost` backed by an in-memory host-function implementation, plus example tests added to the apps. Medium.

**Highest day-to-day velocity payoff of the whole set** — turns app dev into normal Rust TDD.

chefsale:
Closes #2551.

## What

Adds `calimero_sdk::testing::TestHost` — an in-process harness that runs `#[app::state]` logic under a plain `cargo test`: no WASM build, no node, no merobox. Turns app development into standard Rust TDD.

```rust
#[cfg(test)]
mod tests {
    use calimero_sdk::testing::TestHost;
    use super::*;

    #[test]
    fn set_get() {
        let mut app = TestHost::new(KvStore::init);          // build runs against a fresh store
        app.call(|s| s.set("k".into(), "v".into())).unwrap(); // &mut self + commit
        assert_eq!(app.view(|s| s.get("k")).unwrap(), Some("v".to_owned())); // &self
        assert_eq!(app.events().len(), 1);                    // app::emit! captured
    }
}
```

API: `new` / `call` / `view` / `call_as(executor, …)` (multi-author CRDT tests) / `events()` / `logs()` / `set_executor` / `set_context`.

## How

- **Native env mock host** (`crates/sdk/src/env/host.rs`): off-`wasm32` every `calimero_sdk::env::*` call previously trapped through a `calimero_sys` import. A thread-local mock host now backs them — events, logs, executor/context identity, kv + private storage, in-memory blobs. `xcall` / networked blobs / `ed25519_verify` panic with a clear message if hit (no in-process equivalent).
- **Macro-generated bridge**: `TestHost` lives in `calimero_sdk` but can't name `calimero_storage::collections::Root` (storage depends on the SDK, not the reverse). The `#[app::state]` macro generates a `TestState` impl in the app crate that drives `Root` for install / load / mutate / commit.
- **`register_crdt_merge_for_test`**: the bridge needs the native CRDT merge registry, gated behind `calimero-storage`'s `testing` feature so the production node can't reach it (core#2469). A new always-native wrapper (no-op unless the feature is compiled in) keeps `cargo test` compiling for *every* `#[app::state]` app; apps that actually drive `TestHost` opt in with `calimero-storage = { features = ["testing"] }` as a dev-dependency.

## Tests

All core example apps gain `#[cfg(test)]` tests using the harness — **30 tests across 12 apps**: kv-store family, nested-crdt-test, collaborative-editor (RGA), private-data (private storage), blobs, team-metrics ×2, xcall-example. Covers CRDT counters, events, user/frozen/shared storage, and multi-author authorship.

## CI

CI's existing `cargo test` already runs all 12 (verified via `cargo metadata` that the bare invocation resolves every app test target). Added `apps/**` to the `ci-checks` path filter (and its paired stub) so app-only PRs trigger the suite too.

## Scope / caveats

The harness is single-replica, so true cross-node sync/merge convergence and `xcall` round-trips still belong in merobox workflows — documented in `apps/AGENTS.md`.

Verified: 30 app tests pass; sdk (3) + storage (488) + macros (45) tests pass; `cargo test --workspace --no-run`, `cargo build --workspace`, and wasm builds all green; `cargo fmt --check` clean.

---

> [!NOTE]
> **Medium Risk**
> Touches SDK env dispatch and macro-generated test glue across many apps; production WASM paths stay cfg-gated, but incorrect native routing could affect non-wasm builds.
> 
> **Overview**
> Adds **`calimero_sdk::testing::TestHost`** so `#[app::state]` apps can be exercised with ordinary **`cargo test`** (no WASM/node/merobox). Off-`wasm32`, **`calimero_sdk::env`** now delegates to a thread-local **mock host** (events, logs, identity, KV/private storage, in-memory blobs); **`xcall`**, **`ed25519_verify`**, and ordered-index host hooks **panic** with explicit messages when used in-process.
> 
> The **`#[app::state]`** macro emits a **`#[cfg(test)]` `TestState`** impl that drives **`Root`** install/load/commit and calls **`register_crdt_merge_for_test`** (real registration only with **`calimero-storage`’s `testing`** dev-feature). **`reset_environment`** exposes storage mock reset to the harness.
> 
> **Docs/changelog** and **`apps/AGENTS.md`** document the pattern; **12 example apps** add **`#[cfg(test)]`** suites and **`calimero-storage` { features = ["testing"] }** dev-deps (~30 tests: CRDT multi-author **`call_as`**, events, user/frozen/shared storage, RGA editor, private data).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bce02aa91b65e596e9cca583d941fbf2690eb418. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Thanks for the review passes 🙏 Addressed in `7851193c`:

**Fixed (real bugs):**
- **`private_storage_write` bool semantics** (Cursor + MeroReviewer) — the mock returned `.is_some()` (main-storage "was a value evicted?" semantics), so a first-time private write wrongly returned `false`. Now returns `true` on success, matching the WASM `private_storage_write` convention (`true = succeeded`, `false = unavailable`). Comment added explaining the deliberate divergence from `storage_write`.
- **`call_as` panic-safety** (Cursor + MeroReviewer) — the SDK-host executor was restored *after* `__test_with_executor` returned, so a panicking closure left the SDK identity impersonated while the storage layer's RAII guard reverted, and the two layers also switched non-atomically on entry. Now the SDK executor is switched/restored by an RAII guard nested *inside* the storage `with_executor_id` scope, so both layers align before user code runs and unwind together. Added a regression test (`call_as_restores_executor_even_on_panic`).

**Docs tightened:**
- `TestHost` state-isolation wording corrected — it's reset-on-construction + sequential-per-thread, not a thread-per-test guarantee.
- Noted the fixed default identity (access-control *rejection* tests must set a non-owner identity explicitly).
- `env::random_bytes` documented as deterministic / non-cryptographic off-wasm.

**Considered but not changed:**
- **`sha2` "production dependency"** — `sha2` is already a normal dependency of `calimero-sdk` (used elsewhere in the crate), so the mock's blob hashing adds nothing new.
- **Cross-crate assertion that the default ids match `calimero-storage`'s** — can't be done from the SDK (`calimero-sdk` → `calimero-storage` would be a dependency cycle; that cycle is the whole reason for the macro bridge). Agreement is exercised in practice by the `user_storage_roundtrip` / `call_as` example tests, which only pass when both layers resolve to the same identity. Documented the coupling instead.
- **Single-`TestHost`-per-thread Drop guard** — the premise (parallel tests interleaving on one thread) doesn't hold: the test runner runs tests sequentially per pool thread, and `new` resets all thread-locals, so a panic-guard adds friction without catching a real failure mode. Clarified in docs instead.

All 12 example app suites + sdk/storage/macro tests stay green; `cargo fmt --check` clean. Rebased onto the latest branch state (which now includes the #2566 `app::Result` standardization).

chefsale:
Second review pass addressed in `8fd64e44`:

**Fixed:**
- **Cross-thread misuse** — `TestHost` is now `!Send + !Sync` (raw-pointer `PhantomData` marker). Since it's backed by per-thread mock-host state, moving it to another thread and calling `call`/`view` there is now a *compile error* rather than a silent read of a different thread's default host. This structurally resolves the cluster of thread-isolation concerns.
- **Stale blobs test comment** — removed the leftover "Base58 of 32 zero bytes" line (it's `BlobId::from([7u8; 32])` now).
- **Docs** — noted that `#[app::init]`-emitted events/logs are intentionally visible right after construction (`kv-store-init`'s test asserts on this).
- **CI** — deduped the `apps/**` path filter, which #2567 also added when master was merged in.

**Considered but not changed (with rationale):**
- **`call`/`view`/`call_as` "secondary panic via `out.expect`"** — this doesn't happen: when the user closure panics, the unwind propagates straight up *past* `out.expect(...)`, so the original panic reaches the caller; the `expect` is unreachable on the failure path (it's a defensive guard for the impossible "bridge didn't run the closure" case). The `call_as_restores_executor_even_on_panic` test confirms `catch_unwind` receives the original panic and the identity is restored.
- **`AssertUnwindSafe` in that test** — correct usage: after `catch_unwind` the only access is `app.executor_id()` (a thread-local read), and the RAII guards have already restored it, so there's no torn state to observe.
- **Deterministic `random_bytes` seed** — intentional: tests want reproducibility, and it's documented as non-cryptographic. Not security-relevant for a test mock.
- **`sha2`** — already a dependency of `calimero-sdk`; nothing new is pulled in.
- **Sealing `TestState`** — its methods are `__`-prefixed and `#[doc(hidden)]`, and the only impl is macro-generated; true cross-crate sealing needs extra machinery for marginal benefit, so I left it as a clearly-internal trait.
- **`storage_write` vs `private_storage_write` bool asymmetry** — intentional and matches production; already called out in an inline comment.

All 12 example suites + `cargo fmt --check` stay green. (Note: master keeps getting merged into this branch by automation — #2556/#2567 came in this round — so I've rebased each follow-up onto the latest branch tip.)

chefsale:
Re-review (post-rebase onto #2545/#2559/#2573) addressed in `611ab647`. Most findings were re-surfaced repeats of the prior two rounds — handled the substantive ones:

**Fixed**
- **Two live `TestHost`s on one thread** (several comments): `new` now **panics** if another harness is still alive on the thread (they share thread-local mock state), via a thread-local flag cleared on `Drop`. Turns the silent-clobber footgun into a clear error; added a regression test.
- **`read_raw` / migrations under the harness** (Cursor, `host.rs:18`): real limitation — committed state lives in `calimero_storage`'s mock while `read_raw` reads the SDK host map. Documented it (plus `xcall` / networked blobs / `ed25519_verify`) in a new **Limitations** section; migrations are out of scope for this harness.
- **`time_now` same-nanosecond ties**: native `env::time_now` is now strictly increasing (wall clock bumped to ≥ `last + 1`), so time-ordered logic can't tie on a fast machine.
- **`AssertUnwindSafe` in the panic test**: commented why it's sound (post-unwind we only read a thread-local the RAII guards already restored).

**Not changing (with rationale)**
- **`call_as` double-panic** (Cursor + MeroReviewer): doesn't happen — when the user closure panics inside `__test_with_mut`, the unwind propagates straight up *past* `out.expect`, which is never reached. The `call_as_restores_executor_even_on_panic` test confirms `catch_unwind` receives the *original* panic and identity is restored. (The Cursor `call_as`/`private_storage_write` comments point at old commits `6b6b8f75`/`bfaa58f9` and were already fixed.)
- **`sha2` dependency**: already a normal dependency of `calimero-sdk` before this PR; the mock's blob hashing adds nothing new.
- **`random_bytes` determinism**: intentional and documented — reproducible test runs; the PRNG still advances within a test so distinct draws differ.
- **Default identity enforcement**: documented (rejection-path note); can't hard-enforce without breaking the common single-identity test, which legitimately uses the default owner.

All 12 example suites pass; sdk clippy + `cargo fmt --check` clean. (Note: the rebase also pulled #2559's `SortedMap` field into `nested-crdt-test` — its TestHost tests still pass, confirming `SortedMap` works under the harness via the storage crate's native ordered-index mock.)

PR #2569, opened 2026-06-01
feat(apps): exercise SortedSet in the scaffold app and add a SortedSet example

chefsale:
## What

`SortedMap` and `SortedSet` shipped together in #2565, but only `SortedMap` was wired into the scaffold app and the runnable examples — `SortedSet` existed only as a `state-schema-conformance` field, never exercised. This brings `SortedSet` to parity.

## Changes

- **`scaffolding-e2e`**: added a `sorted_tags: SortedSet<String>` field + `sorted_tag_add` / `sorted_tag_remove` / `sorted_tag_contains` / `sorted_tags_all` / `sorted_tags_range` / `sorted_tags_last` methods, mirroring the existing `sorted_items` `SortedMap` surface. They drive the same WASM host ordered-index path (`storage_index_set` / `storage_index_scan` / `storage_index_last`) for sets.
- **`apps/sorted-set-store`** (new): the canonical `SortedSet` example, mirroring `sorted-kv-store` — element-ordered iteration plus `range` / `prefix` / `page` / `first` / `last` (the `BTreeSet` to `UnorderedSet`'s `HashSet`).
- **`crates/runtime/tests/crdt_conformance.rs`**: two real-WASM tests — `sorted_set_range_through_real_wasm` and `sorted_set_rebuilds_index_after_sync` — that drive the compiled scaffold via `Module::run`, proving the `SortedSet` ordered reads and the post-sync index self-heal work end to end (the `SortedSet` counterparts of the existing `SortedMap` conformance tests).

## Notes

- `SortedSet`'s ABI collection marker was already locked by `state-schema-conformance` (`sorted_string_set`), so no golden-file regeneration is needed.
- `abi_conformance` deliberately omits sets (it has no `UnorderedSet` either), so it's left symmetric.
- The example's generated `res/` ABI is gitignored, matching `sorted-kv-store`.

## Verification

- `sorted_set_range_through_real_wasm` + `sorted_set_rebuilds_index_after_sync` pass.
- scaffold + example build to `wasm32-unknown-unknown`.
- `cargo fmt --check` and `cargo clippy` clean.

---

> [!NOTE]
> **Low Risk**
> Example app, scaffold API surface, and test-only runtime coverage; no changes to core storage merge or host index implementation.
> 
> **Overview**
> Brings **`SortedSet`** to parity with the existing **`SortedMap`** wiring: it was in storage/schema but not exercised end-to-end through WASM.
> 
> **`scaffolding-e2e`** gains `sorted_tags: SortedSet<String>` plus `sorted_tag_*` methods (add/remove/contains/all/range/last) that use the same host ordered-index path as `sorted_items`.
> 
> A new **`apps/sorted-set-store`** example mirrors `sorted-kv-store` for sets (sorted iteration, `range` / `prefix` / `page` / `first` / `last`). Workspace/`Cargo.lock` register the app.
> 
> **`crdt_conformance`** adds two real-WASM tests (`sorted_set_range_through_real_wasm`, `sorted_set_rebuilds_index_after_sync`) parallel to the existing `SortedMap` cases. **`sorted-kv-store`** only adds `is_empty` and drops a clippy allow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbc0a1efd41bd888b6c623481e6b790c66357e2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed the review in `fbc0a1ef`.

**Fixed:**
- **`is_empty` / blanket `len_without_is_empty` allow** — added `is_empty()` to **both** `sorted-set-store` and `sorted-kv-store` and removed the crate-level allow from each, so the lint is satisfied properly and the two examples stay mirror images.
- **`len` missing doc / `page` offset-past-end behaviour** — documented both (`page` now notes it returns an empty `Vec` once `offset` reaches the end).
- **`clear` event** — now emits `Cleared` only when the set was non-empty and only after the clear succeeds, matching the conditional emit already used by `add`/`remove`.
- **Test coverage of the last synced element** — `sorted_set_rebuilds_index_after_sync` now also asserts `sorted_tags_all == ["a","f","m","z"]`, so a bug dropping the range's exclusive upper bound (`z`) during sync is caught.

**Left as-is (with rationale):**
- **`build.rs` `.expect()`** — copied verbatim from `sorted-kv-store`'s build script; a build script aborting on a missing/unreadable `src/lib.rs` is the standard Cargo behaviour and keeps the two examples identical. Not worth diverging one example's build script for.
- **`new_with_field_name("items")` vs scaffold's `new()`** — both are correct for a top-level `#[app::state]` field (the macro reassigns a deterministic id post-`init` either way). The new example matches its sibling `sorted-kv-store` (which uses `new_with_field_name`), and the scaffold field matches `scaffolding-e2e`'s own convention (`new()` for every field). Each app is internally consistent.
- **`range.get("output").cloned().unwrap_or(range)` in tests** — this mirrors the existing `SortedMap` conformance tests verbatim; the fallback is intentional because the harness `query` returns either the bare value or an `{"output": …}` wrapper depending on the return shape. Changing only the new tests would diverge from the established pattern.

Issue #2316, opened 2026-05-09
node/sync: decouple HashComparison responder from swarm-poll to fix gossipsub mesh starvation (#2293)


rtb-12:
## Context

PR #2315 (the #2299 fix) lands per-variant actor isolation for `BroadcastMessage::StateDelta` plus stall hardening (15s WASM apply timeout + 60s per-delta budget + DAG zombie rollback). Across the kv-store-with-handlers fuzzy test it raises pass rate from a flaky ~95% → 100% / 99.7% (across two CI samples).

Investigation during that PR uncovered that **the residual ~1–2% Scaffolding E2E flakiness is a different bug** — already documented under **#2293** (gossipsub mesh empties on receiving node under fuzzy load, breaking cross-node sync for ~30s). The scaffolding-e2e `cross_node_kv_sync` operation has the same symptom shape as the kv-store-with-handlers `cross_node_handler_check`: write on one node, wait 2s, read on another node, value not yet propagated.

This issue tracks the focused implementation work to fix #2293 by **applying the same arbiter-isolation pattern PR #2315 uses for StateDelta to the sync responder path**.

## Why this should be its own PR (not folded into #2315)

- Different code path: `crates/node/src/sync/hash_comparison_protocol.rs` vs. `crates/node/src/state_delta_bridge.rs`
- Different root cause: gossipsub Swarm-poll starvation vs. NodeManager Arbiter contention
- Same symptom shape from the test's perspective, but the *mechanism* is unrelated
- Folding would balloon review surface and mix architectural concerns

## What needs to happen

Per #2293's analysis, the receiver's `HashComparison` responder runs on the same task that drives gossipsub `Swarm::poll`. A long responder session (9.63s observed in the trace, vs. ~2ms median) starves gossipsub heartbeats; mesh peers prune the receiver; mesh empties to `{}` for ~30s until the random-peers pool replenishes.

**Fix:** spawn the sync responder on a dedicated arbiter, mirroring the `StateDeltaActor` pattern from PR #2315.

### Sketch

1. Find the call site of `HashComparisonProtocol::run_responder` in the network-event → sync glue (the responder is invoked when a peer opens a sync stream).
2. Either:
   - Wrap the responder invocation in `tokio::spawn` to detach from the network-event task, **or**
[...]

PR #2573, opened 2026-06-01
fix(storage): serialize Merkle index read-modify-writes (concurrent execute/sync race)

### LINKED ISSUE #2571: Sync split-brain: HashComparison re-merge loop on UnorderedMap collection entity under concurrent bidirectional inserts
## Summary

When two nodes concurrently insert **different** new keys into the same already-populated `UnorderedMap` and reconcile via **HashComparison** (rather than delta sync), the map **collection entity's Merkle `full_hash` never converges**. HashComparison re-merges the collection entity every sync interval and the two nodes' root hashes stay stably divergent until `wait_for_sync` times out.

Intermittent (timing/interleaving dependent). **Pre-existing on `master`** — reproduced at `c47e7446` (2026-05-30, run `26688104738`), before any recent PR; also surfaced on an unrelated PR's CI (job `78795791418`).

## Reproduction / signature

Smoke workflow `workflows/sync-tests/opaque-leaf-regression.yml`, **Round 2**:
- Round 1: node-1 `set("alpha")`, node-2 `set("beta")` → converges ✓
- Round 2: node-1 `set("gamma")`, node-2 `set("delta")` **concurrently** (distinct keys, commute) → `wait_for_sync` times out after ~30s; the two nodes settle on two stable, different root hashes.

From the failed run's node logs:
- **34–36 back-to-back HashComparison sessions** during Round 2 (the "38+ identical-stat sessions, root never converges" sticky-loop).
- Exactly **one entity re-merges 35×**: id `5cec8955ddd6…`, `crdt_type=UnorderedMap { key_type: "alloc::string::String", value_type: "LwwRegister<alloc::string::String>" }` — i.e. the **`items` map *collection* entity itself**.
- The leaf entries (alpha/beta/gamma/delta) are applied only **1–2× each** — they reconcile fine; data converges.
- The collection entity is applied with **`ancestor_ids=[]` / `ancestor_count=0`** (a direct child of the context root), and `verify_ancestor_integrity` logs:
  > `ancestor merkle hash mismatch — receiver state diverged from signer's view (accepting; CRDT merge resolves divergence)`
  then warn-only continues.

So the **leaves converge but the collection entity's `full_hash` never matches the peer's**, so every interval re-triggers HashComparison, which re-pushes/re-merges the collection entity, recomputes a still-divergent hash, and loops to timeout.

## Scope — what it is NOT

- **Not the delta-apply path.** A two-round delta-apply repro (`__calimero_sync_next`) in `crates/runtime/tests/crdt_conformance.rs` converges for the identical scenario.
- **Not covered by existing sim guards**, all of which pass:
  - `test_opaque_leaf_converges_via_hash_comparison` / `…_on_responder…`
  - `test_nested_leaf_converges_at_correct_merkle_position` (the `bdc61af` single-new-child placement fix)
  - `test_initiator_newer_leaf_pushes_to_peer_in_single_sync` (#2407 same-entity bidirectional LWW)
- A clean alternating-sync sim of *bidirectional* new children also converges, which is why this only shows up as a flaky e2e — the real failure needs the interleaving where each node holds a concurrent local child the other lacks **and** the collection entity itself is pushed/applied.

## Prime suspect

The map **collection (parent) entity** is applied at the wrong Merkle position (`ancestor_ids=[]` → context-root child) on the receiver, so the two nodes' `items` subtrees never produce an identical `full_hash` even after the leaf children converge. This is the `bdc61af` "Same DAG heads, different root hash" class extended from a single new leaf to the **collection entity under concurrent bidirectional inserts**.

Likely fix area: HashComparison's `apply_leaf_with_crdt_merge` / collection-entity push placement (honour the sender's `index.parent_id()` for the collection entity too, not just leaves) + the post-merge `full_hash` recompute (`crates/storage/src/interface.rs` `verify_ancestor_integrity` / `crates/storage/src/index.rs` `recalculate_ancestor_hashes_for`).

## Suggested regression guard

A `crates/node/tests/sync_sim/protocol.rs` test: shared parent + two settled children on both nodes, then **both** nodes add a distinct new child concurrently, reconcile bidirectionally, and assert equal root hash + all four children parented to the shared parent. NB: in an idealized clean-alternating setup this converges, so the guard must model the interleaved partial-session ordering the e2e produces (e.g. push the collection entity while each node still has an unsynced local child).

## Related

Same family as the prior state-hash-divergence work (#2512 frozen-sync split-brain, #2514 same-DAG-different-root-hash) and the `bdc61af` Round-2 fix.

chefsale:
Closes #2571.

## Problem

The opaque-leaf sync-regression smoke test (Round 2) intermittently split-brains: two nodes settle on stable-but-different Merkle root hashes and HashComparison re-merges the same collection entity forever until `wait_for_sync` times out. Mostly green on `master`, but reproduced there at `c47e7446`.

## Root cause — a lost-update race on the Merkle index

On the native node:
- the **execute** path (local writes) runs under `context.lock()`, and
- the dedicated **`SyncSessionActor`** (#2316) applies HashComparison results,

on **different threads with no shared lock**. Both do a read-modify-write on a collection's `children` index entry: `get_index` → mutate the `children` Vec → `save_index`. Concurrent RMW loses one update — a child is dropped from the `children` list while its `Key::Entry` still exists. That collection's `full_hash` then diverges from every peer's, and HashComparison can never reconcile it (the same-DAG-heads / different-root split-brain).

This was confirmed by deep log analysis of a failing run (one `UnorderedMap` collection entity re-merged 35× while leaves converged) and by a deterministic harness: every *single-threaded* reproduction converges; only the concurrent one diverges.

## Fix

A **native-only reentrant guard** that serializes the index read-modify-writes (`crates/storage/src/index.rs`). Each guarded call re-reads the current list and appends, so no concurrent add is lost regardless of interleave. Properties:
- **Reentrant** — a single logical operation nests these calls on one thread (`add_child_to` → `recalculate_ancestor_hashes_for` → …) without self-deadlock.
- **Native-only** — WASM app execution is single-threaded, so the guard compiles out to nothing; no new dependency, no wasm impact.
- **Complete** — applied to every index-entry RMW path: `add_child_to`, `add_root`, `mark_deleted`, `recalculate_ancestor_hashes_for_now`, `remove_child_from`, `update_parent_after_child_removal`, `update_hash_for`, and the one direct `save_index` RMW in `Interface::update_signature_in_place`.

A single global lock can't deadlock (no lock-ordering cycle); it's only ever held across synchronous index ops (never across `.await`).

## Tests

New deterministic concurrency repro — `crates/storage/src/tests/concurrency.rs` — driving two threads against a shared cross-thread storage backend (the production `MockedStorage` is `thread_local!`):
- `concurrent_add_child_to_loses_children` — **fails pre-fix** (keeps ~half of 600 concurrently-added children), passes post-fix.
- `concurrent_children_index_diverges_from_serial_full_hash` — **fails pre-fix** (concurrent vs serial `full_hash` differ), passes post-fix.

## Verification

- Repro tests fail without the fix, pass with it.
- `cargo test -p calimero-storage --lib`: 520 passed, 0 failed (no deadlock/regression).
- `cargo test -p calimero-node --test sync_sim`: 308 passed, 0 failed (the other side of the race; opaque-leaf/HC guards intact).
- `cargo build --workspace` green; wasm build green (no-op guard path); `cargo fmt --check` clean.

## Note for reviewers

The guard is a *global* index-mutation lock, so index writes serialize across all contexts. That's correct and the ops are microsecond-scale; a per-context lock keyed by the `RuntimeEnv` context is a clean future optimization if contention ever shows up. The fix is intentionally at the storage layer (fixes the race at its root, self-contained, defense-in-depth for any concurrent storage caller) rather than serializing the HC apply under `context.lock()` at the node layer.

---

> [!NOTE]
> **Medium Risk**
> Touches Merkle index mutation hot paths used by execute and sync; a global lock serializes all native index writes (correctness fix with possible contention, mitigated by reentrancy and short critical sections).
> 
> **Overview**
> Fixes intermittent **sync split-brain** (core#2571) where local execute and `SyncSessionActor` HashComparison apply could **lose updates** on the same Merkle index entry because `get_index` → mutate `children` → `save_index` was not atomic across threads.
> 
> On **native only**, adds a **reentrant global `index_mutation_guard`** and wraps every index read-modify-write path (`add_child_to`, root/tombstone/hash/ancestor walks, child removal, `update_hash_for`, and `update_signature_in_place`). **WASM** keeps a no-op guard. New **`tests/concurrency`** uses a cross-thread shared store to repro missing children and divergent `full_hash` pre-fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a929efb55fe6e292b49d728c16d94eb7d9c0304. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/storage/src/index.rs
+++ b/crates/storage/src/index.rs
@@ -43,6 +43,77 @@ thread_local! {
         const { RefCell::new(None) };
 }
 
+// Cross-thread serialization for index read-modify-write (core#2571).
+//
+// The Merkle index mutators below (`add_child_to`, the hash recomputes, the
+// child-removal path) read an index entry, mutate it in memory, then write it
+// back. On the native node a local write (the execute path, under
+// `context.lock()`) and a sync apply (the dedicated `SyncSessionActor`, #2316)
+// run on DIFFERENT threads with no shared lock — so two concurrent
+// read-modify-writes on the same parent index entry lose one update. The lost
+// update drops a child from a collection's `children` list while that child's
+// `Key::Entry` still exists; the collection's `full_hash` then diverges from
+// every peer's and HashComparison can never reconcile it (the same-DAG-heads /
+// different-root split-brain that loops `wait_for_sync` to timeout).
+//
+// Serializing the read-modify-writes makes each one atomic: a caller always
+// reads the CURRENT list and appends its child, so no concurrent add is lost
+// regardless of interleave. The guard is reentrant — a single logical operation
+// nests these calls on one thread — and native-only: WASM app execution is
+// single-threaded, so there is no race and the guard compiles out to nothing.
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) mod mutation_lock {
+    use std::cell::Cell;
+    use std::sync::{Mutex, MutexGuard};
+
+    static LOCK: Mutex<()> = Mutex::new(());
+    thread_local! {
+        static DEPTH: Cell<u32> = const { Cell::new(0) };
+    }
+
+    /// Held for the duration of an index read-modify-write. The global lock is
+    /// released only when the OUTERMOST guard on this thread drops, so nested
+    /// mutators (e.g. `add_child_to` → `recalculate_ancestor_hashes_for`) don't
+    /// self-deadlock.
+    #[must_use]
+    pub(crate) struct Guard(#[allow(dead_code)] Option<MutexGuard<'static, ()>>);
+
+    pub(crate) fn guard() -> Guard {
+        DEPTH.with(|depth| {
+            let current = depth.get();
+            depth.set(current + 1);
+            if current == 0 {
+                // Recover from a poisoned lock: a panic mid-mutation never
+                // leaves a half-written index entry (each `save_index` is a
+                // single store write), so the protected data stays consistent
+                // and the lock is safe to keep using.
+                Guard(Some(
+                    LOCK.lock().unwrap_or_else(|poison| poison.into_inner()),
+                ))
+            } else {
+                Guard(None)
+            }
+        })
+    }
+
+    impl Drop for Guard {
+        fn drop(&mut self) {
+            DEPTH.with(|depth| depth.set(depth.get().saturating_sub(1)));
+            // The held `MutexGuard` (outermost guard only) drops here.
+        }
+    }
+}
+
+/// Acquire the reentrant index-mutation guard. Native serializes concurrent
+/// read-modify-writes; WASM is single-threaded so this is a no-op.
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) fn index_mutation_guard() -> mutation_lock::Guard {
+    mutation_lock::guard()
+}
+
+#[cfg(target_arch = "wasm32")]
+pub(crate) fn index_mutation_guard() {}
+
 /// RAII scope that defers ancestor-hash recomputation for a specific
 /// `StorageAdaptor`.
 ///
@@ -308,6 +379,10 @@ pub struct Index<S: StorageAdaptor>(PhantomData<S>);
 impl<S: StorageAdaptor> Index<S> {
     /// Adds a child to a parent's collection.
     pub(crate) fn add_child_to(parent_id: Id, child: ChildInfo) -> Result<(), StorageError> {
+        // Serialize the read-modify-write so a concurrent local-write / sync
+        // apply on the same parent can't lose a child (core#2571).
+        let _mutation_guard = index_mutation_guard();
+
         // Get or create parent index
         let mut parent_index = Self::get_index(parent_id)?.unwrap_or_else(|| EntityIndex {
             id: parent_id,
@@ -398,6 +473,7 @@ impl<S: StorageAdaptor> Index<S> {
     /// # Errors
     /// Returns `StorageError` if index cannot be loaded or saved.
     pub fn add_root(root: ChildInfo) -> Result<(), StorageError> {
+        let _mutation_guard = index_mutation_guard();
         let mut index = Self::get_index(root.id())?.unwrap_or_else(|| EntityIndex {
             id: root.id(),
             parent_id: None,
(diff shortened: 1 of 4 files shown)


chefsale:
Addressed in `5a929efb`:

**Fixed**
- **Test isolation** (Cursor + MeroReviewer): both concurrency tests share the process-global `shared_store`, so they're now `#[serial]` — parallel runs could have one test's `reset_shared_store` clear the other's data mid-run. Also overrode `participates_in_sync() = false` on the test adaptor so a stray `Action::Compare` push can't touch thread-local delta state.
- **Misleading poison comment**: corrected. The lock only *serializes* access — it adds no transactionality, and neither does the storage layer, so a mid-mutation panic can leave the index partially updated with or without the lock (that's the pre-existing lock-free behaviour). Recovery just avoids one thread's panic bricking every other thread's index writes; surfacing the panic is the caller's job. The "single store write" claim was wrong and is gone.
- **Drop order**: now releases the `MutexGuard` before decrementing `DEPTH`, so the "DEPTH > 0 ⇒ this thread holds the lock" invariant never momentarily inverts.
- **`mem::forget` hazard**: documented on `Guard` (same caveat as `DeferredAncestorScope`).

**Considered, keeping as-is (with clearer rationale in the code)**
- **Global lock vs per-context/per-entity**: kept global, and added the reasoning to the module comment. A per-entity lock would let two threads grab a parent and an ancestor in opposite orders during the ancestor-hash walk and **deadlock**; one global lock is deadlock-free by construction. Index mutations are microsecond-scale (in-memory + one store write), so global serialization is cheap. A per-context lock keyed by the `RuntimeEnv` context is a safe future optimization — but note it wouldn't engage in the direct-`Index` repro, so it'd need a different test vector.
- **`tokio::sync::Mutex` / "never hold across `.await`"**: the guard is only ever taken inside synchronous index functions (no `await` between acquire and drop), so a std `Mutex` is correct and a tokio mutex would be wrong here (can't hold its guard across the sync RMW without making the whole index layer async). The invariant is now documented.
- **WASM no-op returns `()` vs a unit guard type**: intentional; every call site is `let _guard = …`, which binds either. Not worth a parallel type just for symmetry.

Storage suite still 520 passed / 0 failed; repro tests green.

PR #2576, opened 2026-06-01
feat(sdk): Entry::or_default() write-back guard for nested CRDTs

### LINKED ISSUE #2542: [SDK UX] In-place mutation for nested CRDTs (entry().or_default() write-back guard)
## Summary
Make in-place mutation of **nested** CRDT values the blessed path, so app authors stop hand-writing the get → modify → re-insert dance.

## Current friction
`kv-store` already uses `get_mut`/`entry` on `UnorderedMap`, but the nested-CRDT apps don't — they manually read, mutate a local copy, then re-insert. From `apps/nested-crdt-test/src/lib.rs:78-101`:

```rust
let mut counter = self.counters.get(&key)
    .map_err(|e| format!("Get failed: {:?}", e))?
    .unwrap_or_else(|| Counter::new());
counter.increment().map_err(|e| format!("Increment failed: {:?}", e))?;
let value = counter.value().map_err(|e| format!("Value failed: {:?}", e))?;
drop(self.counters.insert(key.clone(), counter)
    .map_err(|e| format!("Insert failed: {:?}", e))?);
```

The same pattern repeats in `team-metrics-custom`, `team-metrics-macro`, and the nested-map case (`set_metadata`, two levels deep).

## Proposed
A write-back mutable guard so the value persists on drop, no manual re-insert:

```rust
let mut counter = self.counters.entry(key.clone())?.or_default()?;
counter.increment()?;
let value = counter.value()?;
```

## SDK work / effort
`entry()` exists; the missing piece is `or_default()` returning a mutable guard that re-persists the CRDT on drop, and that it **composes for nested collections** (e.g. `Map<String, Map<String, _>>`). Medium.

Rewrite `nested-crdt-test`, `team-metrics-*` to use it once landed.

chefsale:
Closes #2542.

## Summary
Completes the in-place mutation path for nested CRDT values so app authors stop hand-writing the `get → modify → re-insert` dance.

Adds `Entry::or_default()` to both `UnorderedMap` and `SortedMap`. It inserts `V::default()` when the entry is vacant and returns the existing `ValueMut` write-back guard, which re-persists the value on drop (deterministically re-keying nested CRDTs under the entry's id). The guard plumbing and re-keying already existed behind `or_insert`/`or_insert_with`, so `or_default()` is a thin `or_insert_with(V::default)` wrapper — and it composes for nested collections (`Map<String, Map<String, _>>`) for free.

The blessed path is now:
```rust
let mut counter = self.counters.entry(key.clone())?.or_default()?;
counter.increment()?;
let value = counter.value()?;
```

## Changes
- **SDK** (`crates/storage/src/collections/`): add `or_default()` to `unordered_map.rs` and `sorted_map.rs`, with unit tests (vacant insert + write-back, occupied keep-existing, SortedMap ordering preserved).
- **Apps** rewritten onto `entry().or_default()`:
  - `nested-crdt-test`: `increment_counter`, `set_metadata` (two levels deep), `add_tag`
  - `team-metrics-custom`, `team-metrics-macro`: `record_win` / `record_loss` / `record_draw`

## Testing
- `cargo test -p calimero-storage --lib entry` → 11 passed (incl. both new `or_default` tests)
- `nested-crdt-test` / `team-metrics-custom` / `team-metrics-macro` `TestHost` unit tests pass (5 / 2 / 2)
- No new unused-import warnings; `cargo fmt --check` clean

---

> [!NOTE]
> **High Risk**
> Breaking `get` return types across storage collections plus CRDT identity/re-key and convergence behavior; incorrect usage could cause silent divergence before this change and compile breaks after.
> 
> **Overview**
> This PR finishes the **in-place nested CRDT mutation** story: **`Entry::or_default()`** on `UnorderedMap` and `SortedMap` (vacant → `V::default()`, returns **`ValueMut`** that **write-backs on drop**, including deterministic **re-key** for nested collections). Apps (`nested-crdt-test`, `scaffolding-e2e`, team-metrics apps) replace **`get` → modify → `insert`** with **`entry(...).or_default()?`**.
> 
> **`get`** on core maps/vectors now returns **`ValueRef<V>`** — an owned, **read-only** guard (`Deref` only, no public `into_inner`) so mutating a fetched copy without persisting becomes a **compile error**. Internal merge/decompose paths use crate-private **`into_inner`** / **`as_deref`**.
> 
> **`Default`** for maps/sets/vectors registers **re-key** at construction so values first created via **`or_default()`** converge across nodes. New **integration tests** cover first-touch convergence for nested counters and **`Map<String, Map<String, Counter>>`** via **`or_default()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab22012b65ae902cfc9fd8944703446ef97c3460. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/nested-crdt-test/src/lib.rs
+++ b/apps/nested-crdt-test/src/lib.rs
@@ -89,13 +89,14 @@ impl NestedCrdtTest {
     // ===== Counter Operations =====
 
     pub fn increment_counter(&mut self, key: String) -> app::Result<u64> {
-        let mut counter = self.counters.get(&key)?.unwrap_or_else(|| Counter::new());
+        // `or_default()` returns a write-back guard: the counter is re-persisted
+        // when `counter` drops at the end of this method, so there is no manual
+        // get → modify → re-insert dance.
+        let mut counter = self.counters.entry(key.clone())?.or_default()?;
 
         counter.increment()?;
 
         let value = counter.value()?;
-
-        self.counters.insert(key.clone(), counter)?;
 
         app::emit!(TestEvent::CounterIncremented { key, value });
 
@@ -177,14 +178,12 @@ impl NestedCrdtTest {
         inner_key: String,
         value: String,
     ) -> app::Result<()> {
-        let mut inner_map = self
-            .metadata
-            .get(&outer_key)?
-            .unwrap_or_else(|| UnorderedMap::new());
+        // Two levels deep: the guard yields the inner map in place and writes the
+        // whole nested CRDT back when it drops — `or_default()` composes for
+        // `Map<String, Map<String, _>>` without re-inserting either level.
+        let mut inner_map = self.metadata.entry(outer_key.clone())?.or_default()?;
 
         inner_map.insert(inner_key.clone(), value.clone().into())?;
-
-        self.metadata.insert(outer_key.clone(), inner_map)?;
 
         app::emit!(TestEvent::MetadataSet {
             outer_key,
@@ -236,11 +235,9 @@ impl NestedCrdtTest {
     // ===== Set Operations =====
 
     pub fn add_tag(&mut self, key: String, tag: String) -> app::Result<()> {
-        let mut set = self.tags.get(&key)?.unwrap_or_else(|| UnorderedSet::new());
+        let mut set = self.tags.entry(key.clone())?.or_default()?;
 
         set.insert(tag.clone())?;
-
-        self.tags.insert(key.clone(), set)?;
 
         app::emit!(TestEvent::TagAdded { key, tag });
 
--- a/apps/scaffolding-e2e/src/lib.rs
+++ b/apps/scaffolding-e2e/src/lib.rs
@@ -26,8 +26,8 @@ use calimero_sdk::serde::Serialize;
 use calimero_sdk::{app, env, PublicKey};
 use calimero_storage::collections::{
     AuthoredMap, AuthoredVector, Counter, FrozenStorage, GCounter, LwwRegister, Mergeable,
-    PNCounter, ReplicatedGrowableArray, SharedStorage, SortedMap, SortedSet, UnorderedMap,
-    UnorderedSet, UserStorage, Vector,
+    ReplicatedGrowableArray, SharedStorage, SortedMap, SortedSet, UnorderedMap, UnorderedSet,
+    UserStorage, Vector,
 };
 use sha2::{Digest, Sha256};
 use thiserror::Error;
@@ -894,13 +894,11 @@ impl E2eKvStore {
     // --- G-COUNTER (grow-only) ---
 
     pub fn increment_g_counter(&mut self, key: String) -> app::Result<u64> {
-        let mut counter = self.crdt_counters.get(&key)?.unwrap_or_else(GCounter::new);
+        let mut counter = self.crdt_counters.entry(key.clone())?.or_default()?;
 
         counter.increment()?;
 
         let value = counter.value()?;
-
-        self.crdt_counters.insert(key.clone(), counter)?;
 
         app::emit!(Event::GCounterIncremented { key, value });
         Ok(value)
@@ -917,16 +915,11 @@ impl E2eKvStore {
     // --- PN-COUNTER (supports increment AND decrement) ---
 
     pub fn increment_pn_counter(&mut self, key: String) -> app::Result<i64> {
-        let mut counter = self
-            .crdt_pn_counters
-            .get(&key)?
-            .unwrap_or_else(PNCounter::new);
+        let mut counter = self.crdt_pn_counters.entry(key.clone())?.or_default()?;
 
         counter.increment()?;
 
         let value = counter.value()?;
-
-        self.crdt_pn_counters.insert(key.clone(), counter)?;
 
         app::emit!(Event::PnCounterChanged {
             key,
@@ -937,16 +930,11 @@ impl E2eKvStore {
     }
 
     pub fn decrement_pn_counter(&mut self, key: String) -> app::Result<i64> {
-        let mut counter = self
-            .crdt_pn_counters
-            .get(&key)?
-            .unwrap_or_else(PNCounter::new);
+        let mut counter = self.crdt_pn_counters.entry(key.clone())?.or_default()?;
 
         counter.decrement()?;
 
         let value = counter.value()?;
-
-        self.crdt_pn_counters.insert(key.clone(), counter)?;
 
         app::emit!(Event::PnCounterChanged {
             key,
@@ -999,14 +987,9 @@ impl E2eKvStore {
         inner_key: String,
         value: String,
     ) -> app::Result<()> {
-        let mut inner_map = self
-            .crdt_metadata
-            .get(&outer_key)?
-            .unwrap_or_else(UnorderedMap::new);
+        let mut inner_map = self.crdt_metadata.entry(outer_key.clone())?.or_default()?;
 
         inner_map.insert(inner_key.clone(), value.clone().into())?;
-
-        self.crdt_metadata.insert(outer_key.clone(), inner_map)?;
 
         app::emit!(Event::MetadataSet {
             outer_key,
@@ -1056,11 +1039,9 @@ impl E2eKvStore {
     // NESTED CRDT - TAGS SET
 
     pub fn add_tag(&mut self, key: String, tag: String) -> app::Result<()> {
-        let mut set = self.crdt_tags.get(&key)?.unwrap_or_else(UnorderedSet::new);
+        let mut set = self.crdt_tags.entry(key.clone())?.or_default()?;
 
         set.insert(tag.clone())?;
-
-        self.crdt_tags.insert(key.clone(), set)?;
 
         app::emit!(Event::TagAdded { key, tag });
         Ok(())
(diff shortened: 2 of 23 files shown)


chefsale:
Addressed the MeroReviewer nitpicks in 0675e458:

- **Nested-CRDT test coverage** (both nits): added `test_nested_counter_first_touch_via_or_default_converges`, a cross-node convergence test that drives the re-keying path through `entry().or_default()` end-to-end (the scalar `u64` unit tests can't exercise re-keying). It mirrors the existing `or_insert_with` / `Extend` convergence tests. `or_default()` delegates to `or_insert_with(V::default)`, so this guards the wrapper doesn't regress re-keying.
- **Comment consistency**: dropped the explanatory comment from `record_win` in both team-metrics apps so all three `record_*` methods read uniformly; the write-back semantics live on the `Entry::or_default()` doc-comment.

chefsale:
## Update: also made `get()` read-only (breaking)

Following the discussion that `get()` returning an owned, freely-mutable `V` is the real footgun behind #2542, this PR now also makes `get()` on the three core collections (`UnorderedMap`, `SortedMap`, `Vector`) return a **read-only `ValueRef<V>`** guard.

- `ValueRef<V>: Deref<Target = V>` with **no `DerefMut`** — reads stay transparent; mutating a `get` result no longer compiles, and the borrow checker steers you to `get_mut` / `entry().or_default()`.
- Deliberate read-modify-write-back uses an explicit `ValueRef::into_inner()`.
- Can't be `&V` like `HashMap`: storage values aren't memory-resident, every `get` deserializes a fresh copy — so the guard owns the copy and restricts access.

**Blast radius (measured):** the entire workspace's **production code compiled unchanged** — reads flow through `Deref`. The only breakage was in **example apps and tests**, i.e. exactly where the footgun lived. `scaffolding-e2e`'s GCounter/PNCounter/nested-map/tag-set get-modify-insert handlers are migrated to `entry().or_default()`.

**Scope:** the three primary CRDT collections. Wrapper/aux accessors (`UserStorage`, `Authored*`, `FrozenStorage`, js FFI shims, nested helpers) keep owned-`V` signatures via internal `into_inner()` — a sensible follow-up is to extend the guard to those too.

⚠️ This is a **breaking API change** for Rust SDK consumers (the return type of `get`). Flagging for review given that.

Verification: `calimero-storage` lib tests 533 passed/0 failed; full `cargo build --workspace --tests` is green; example-app `TestHost` tests pass; `cargo fmt --check` clean.

chefsale:
## Cursor Bugbot findings

**1. "Nested re-key runs too early" (High) — confirmed real, fixed in 8648612c.** 

Good catch. `entry(k).or_default()` on a nested *collection* value did cause a first-touch split-brain: collection types register their re-key thunk in `insert`/`extend`, not their constructor, so `VacantEntry::insert` re-keyed the freshly-`default()`-built inner collection *before* its type was ever registered → `rekey_nested_value` no-op → random per-node id → no convergence. (Counter dodged it by registering in `new_internal`.)

I reproduced it with a two-level nested-map convergence test (`test_nested_map_first_touch_via_or_default_converges`) — counter merged to **1 instead of 2** before the fix. Fix registers the re-key thunk in `Default::default()` for UnorderedMap/SortedMap/UnorderedSet/SortedSet/Vector (tightening those `Default` bounds to the types' existing `RekeyTarget` bounds — every usable collection already satisfies them, full workspace builds clean). Test passes after. So `entry().or_default()` is now convergence-safe for nested collections, which is the behaviour #2542 wants.

**2. "ValueMut write-back on drop ignores save errors" — acknowledged, accepted.**

This is an inherent property of the existing `Entry`/`ValueMut` write-back-on-drop API (the same path `kv-store` already uses), not new behaviour this PR introduces — and `entry().or_default()` is exactly what #2542 asked for. In the wasm runtime the per-op writes land in an in-memory staging buffer flushed at the end of the call, so a mid-handler `save` failure isn't a reachable condition there. Surfacing drop-time persistence errors would be a broader change to the guard type; I'd rather not widen this PR's scope for it. Can file a follow-up if you'd like the guard to expose an explicit `.commit()?` for error propagation.

Also fixed the failing `nested_map` doctest (CI) — the intentional "❌ BROKEN" get-modify-put example now needs `.into_inner()` since `get` returns a read-only `ValueRef`.

chefsale:
## MeroReviewer (3 agents) — responses

Thanks. Triaged all 10; most are 1/3-consensus. Fixed the actionable ones in 79637d79.

**W1/W2 — `Default` bound tightening (1/3 each).** The `K: AsRef<[u8]> + PartialEq + 'static` / `V: 'static` bounds are load-bearing: they're exactly what `register_rekey::<Self>()` (via `RekeyTarget`) requires, and they don't restrict any *usable* collection — `insert`/`get`/`remove`/`contains` already demand `K: AsRef<[u8]> + PartialEq + 'static`, so a map without them can be constructed but never read or written. The full workspace + all example apps compile unchanged. I did drop the redundant `S: StorageAdaptor + 'static` (StorageAdaptor already requires `'static`), so the `Default` bounds now match each type's existing `RekeyTarget` impl exactly. The alternative ("put the bound on `or_default` not `Default`") isn't possible — registration has to happen when the collection is *constructed*, i.e. in `Default`, before the parent re-keys it.

**W3 — `register_rekey` side-effect in `Default` (1/3).** It's idempotent and cheap (a read-lock hit on the already-registered fast path) and the registry is an append-only map of `fn` pointers — that's the module's whole design. Borsh deserialization does **not** route through `Default` for these (non-`#[borsh(skip)]`) fields, so it isn't hit on the read path. This mirrors `Counter::new_internal`, which already registers in its constructor for the same reason.

**W4 — add `Clone`/`Copy`/`Hash`/`AsRef` to `ValueRef` (1/3).** Deliberately **not** doing this, and the build proves why: `ValueRef` is a `Deref` guard, so `guard.clone()`/`guard.as_ref()` already resolve to `V`'s methods through the deref. Adding inherent trait impls *shadows* them — `file_record.clone()` started returning `ValueRef<FileRecord>` instead of `FileRecord` and broke real call sites. Added an inline comment documenting the omission.

**W5 — `PartialEq<V>` (1/3).** Added `impl PartialEq<V> for ValueRef<V>` so `map.get(k)?.unwrap() == expected` works without unwrapping. (Operator-based, so no `Deref` shadowing.)

**W6 / S3 — test + Entry duplication (1/3).** Fair, but it follows the established pattern already in this file (the existing `*_converges` tests share the same boilerplate, and `or_insert`/`or_insert_with` are already duplicated across the two maps). Extracting a generic harness/trait would touch the existing passing tests; better as its own cleanup than folded into this PR.

**S1 — `#[serial]` on the new tests.** Already present — both `test_nested_map_first_touch_via_or_default_converges` (L2553) and `test_nested_counter_first_touch_via_or_default_converges` (L2692) are annotated `#[serial]`. No change.

**S2 — `NestedMapOps::insert_nested` still uses get-modify-put.** Correct observation. Now that `or_default()` is convergence-safe for nested collections (this PR), `insert_nested` *could* adopt it. Leaving it for a focused follow-up since it's an internal helper that's already correct and I'd rather not expand this PR's surface.

All green after the changes: storage lib 534 / doctests 4 / `cargo build --workspace --tests` clean / fmt clean.

chefsale:
Good call — done in af873b63.

`ValueRef::into_inner` is now `pub(crate)`. It was only ever used inside the storage crate (CRDT-merge read-modify-write paths + the `UserStorage`/`Authored*`/`js`/`nested` wrapper accessors + tests) — **zero app/SDK/external callers** — so scoping it down closes the bypass with no downstream impact.

Public surface now:
- **mutate + persist** → `get_mut` / `entry().or_default()` (automatic write-back)
- **owned copy** → `.clone()` the guard when `V: Clone`
- there is no public way to pull out an un-written-back owned value

Updated the `get`/`ValueRef` docs, and the `nested_map` "❌ BROKEN" doctest is now prose since the get-modify-put copy is genuinely no longer expressible (the guard only derefs) — which is exactly the intent.

Green: storage lib 534 / doctests 3 / `cargo build --workspace --tests` clean / fmt clean.

chefsale:
Two `ValueRef` notes — neither needs a change:

**1. "`into_inner()` is public, bypassing the guard" (on `collections.rs:91`).** This review ran against `79637d79`; it's already addressed in `af873b63` (one commit later) — `into_inner` is now `pub(crate)`. There's no public way to pull an un-written-back owned value out of the guard anymore: public callers go through `get_mut`/`entry().or_default()` to mutate-persist, or `.clone()` for an owned copy. Resolved.

**2. "Add `PartialEq<ValueRef<V>> for V` for symmetry" (on `collections.rs:130`).** Can't, unfortunately — `impl<V: PartialEq> PartialEq<ValueRef<V>> for V` violates the orphan rule (E0210): `V` is a bare, uncovered type parameter in `Self` position, so a blanket impl of a foreign trait (`PartialEq`) isn't allowed there. Only the `ValueRef<V>: PartialEq<V>` direction is implementable (our type is `Self`). So `guard == value` works; `value == guard` can't be made to without owning `V`. The asymmetry is a language constraint, not an oversight — and `assert_eq!`'s left operand is conventionally the actual (`get(k)?…`), which is the guard, so the supported direction is the common one. Leaving as-is.

Issue #2577, opened 2026-06-01
Custom Mergeable types silently lose data on concurrent writes (nested/struct-wrapped CRDTs fall back to LWW) — follow-through on #1780/#1940


chefsale:
### Summary

A custom `impl Mergeable` on a **nested (non-root) type** is **never invoked** during sync. When two replicas concurrently write the same map/collection entry whose value is a custom struct, the storage layer falls back to **LWW on the whole serialized blob** — silently dropping one side's contribution, *including any built-in CRDTs nested inside that struct*. This converges (no split-brain) but loses data, violating the "no silent data loss" invariant (CIP I5).

The canonical example is our own showcase app, **`apps/team-metrics-custom`** (`UnorderedMap<String, TeamStats>` where `TeamStats` has a hand-written `Mergeable`): its custom merge never runs in the sync path.

### Reproduction / evidence

Native two-replica drive (own stores via `with_runtime_env`, exchange `StorageDelta` via `Root::sync`), each replica records **one** win for the **same** team under its own executor id:

```
value type = UnorderedMap<String, Stats { wins: Counter, inline_sum: u64 }>
child-backed Counter wins: a=1 b=1   (expected 2 — both increments)
INLINE u64 sum:            a=1 b=1   (expected 2 if custom merge ran; 1 = LWW)
hashes equal (converged):  true      ← converges, but on a LOSSY result
```

Contrast — value type `UnorderedMap<String, Counter>` (a built-in CRDT *directly*, no struct wrapper) sums correctly to `2`. The data loss appears specifically when a CRDT (or inline field) is **wrapped in a custom struct value**.

### Root cause (code trace, current tree)

1. **Nested values never get a `CrdtType`.** `Collection::insert_with_storage_type` (`crates/storage/src/collections.rs:~460`) sets `metadata.storage_type` but never `metadata.crdt_type`. Only the *collection container* is stamped (`collections.rs:333`).
2. **The merge registry is consulted only for the ROOT entity.** `register_crdt_merge<T>` keys by `TypeId` and is read only via `merge_root_state` (`crates/storage/src/merge/registry.rs`); `#[app::state]`'s `__calimero_register_merge` registers **only the root type**. Non-root conflicts go through `try_merge_non_root` (`crates/storage/src/interface.rs:~2466`), which never consults the registry.
[...]

Issue #2552, opened 2026-05-31
[SDK UX] CRDT convergence property-test harness for custom Mergeable types


chefsale:
## Summary
A CRDT convergence property-test harness so app authors can assert that their (often custom) `Mergeable` types converge under concurrent, reordered operations.

## Current friction
A developer writing a custom `Mergeable` — e.g. `apps/team-metrics-custom/src/lib.rs` — has **no way to verify their merge converges**. This is exactly the class of bug that has repeatedly surfaced only in production sync (frozen-sync split-brain, same-DAG state-hash divergence, frozen-rga writer-outlier). Builds on the in-flight runtime CRDT-conformance work.

## Proposed
```rust
#[test]
fn team_stats_converge() {
    converge::<TeamMetricsApp>()
        .replicas(3)
        .ops(|r| r.record_win("liverpool".into()))   // applied in random orders per replica
        .assert_all_replicas_equal();                 // + asserts root hash matches
}
```

## SDK work / effort
The app-author-facing surface of the runtime conformance machinery: drive N in-memory replicas through randomized op interleavings and assert state + root-hash equality. Builds on `TestHost` (#10). Medium/High.

The uniquely-Calimero testing primitive — "prove your CRDT converges" as a one-liner. Closest to actual production pain.

chefsale:
## Findings from a prototype implementation — proposed reframe

I built a working prototype of this harness (native, on top of `TestHost` #2568) and a Phase-0 spike turned up something that changes how we should frame and title this feature. Flagging before we ship.

### Good news: the core mechanism works

Native `Root::sync(StorageDelta::Actions)` correctly merges concurrent same-entity child CRDTs. Two replicas each `record_win("liverpool")` on `UnorderedMap<String, Counter>`, exchange deltas in randomized order → counter sums to 2 and both replicas converge to the **same Merkle root**. (This is contrary to what the `#[ignore]`d gcounter cases in `crates/runtime/tests/crdt_conformance.rs` suggested — those exercise a root-level counter via a different path.)

The one-liner API works end to end, including against the real `#[app::state] TeamMetricsApp` driving its actual event-emitting `record_win`:

```rust
converge_app(TeamMetricsApp::init)
    .replicas(3)
    .ops(|s| { let _ = s.record_win("liverpool".into()); })
    .assert_all_replicas_equal();   // asserts equal root hash across replicas
```

### The catch: the custom `Mergeable::merge` is mostly **not** what drives convergence

[...]

PR #2578, opened 2026-06-01
feat(storage): CRDT convergence property-test harness (#2552)

chefsale:
A native (`cargo test`) CRDT convergence property-test harness: drive N in-memory replicas of a `Mergeable` app-state type through randomly-interleaved operations and assert they converge to the same Merkle root — the production-bug class (frozen-sync split-brain, same-DAG divergence) as a one-liner.

```rust
converge_app(TeamMetricsApp::init)
    .replicas(3)
    .ops(|s| { let _ = s.record_win("liverpool".into()); })
    .invariant("liverpool wins == 3", |s| s.get_wins("liverpool".into()).unwrap() == 3)
    .assert_all_replicas_equal();
```

## What's here

- **`calimero_storage::testing::{converge, converge_with, converge_app}`** — N replicas, each with its own store (isolated via `with_runtime_env`) and executor id; identical genesis snapshot (under a distinct genesis executor); per-replica shuffled local ops; one `StorageDelta` captured per op; foreign deltas cross-applied (shuffled) via `Root::sync`; asserts equal root hash. The run takes an internal lock so it **self-serializes** — `#[serial]` is not load-bearing.
- **`.invariant(desc, |&T| bool)`** — value-level correctness on each converged replica. Hash equality is **not** correctness: deterministic LWW converges every replica to the same *wrong* value, so a data-loss bug passes a hash-only check.
- **Op model:** every replica applies *every* registered op (commutativity model), so an op appearing `k` times across `n` replicas runs `n × k` times — e.g. one `record_win` + 3 replicas ⇒ 3 wins.
- **Coverage** — convergence + correctness tests for `team-metrics-custom`/`-macro`, `nested-crdt-test`, `kv-store`, `sorted-set-store`.
- **Un-ignored 4 `crdt_conformance.rs` tests** (gcounter / pncounter / nested_counter_in_map / unordered_set_in_map) — their `#[ignore]` premise ("same-entity merge not exercised by plain Actions apply") is stale; those merges work now. 18/18 conformance green.

## Relationship to #2577 / #2579

This PR surfaced the custom-`Mergeable` data-loss bug (filed #2577, **fixed and merged in #2579**). Now that #2579 is on master, the `team-metrics` correctness tests (`team_stats_converge_to_correct_value`) assert `liverpool wins == 3` and **pass** — they call `__calimero_register_rekey()` so the custom struct value's nested counters get deterministic ids. The harness both *found* the bug and now *guards* against it. (Earlier revisions carried these as `#[ignore]`d "demonstrates #2577" tests; they are no longer ignored.)

## Notes for review

- App tests live in their own `tests/` integration binaries (and need `crate-type = ["cdylib", "rlib"]` + the `calimero-storage` `testing` dev-dep) so they're isolated from the `TestHost` unit tests.
- The `SortedMap`/`SortedSet` ordered secondary index isn't per-replica isolated, but it's node-local and not part of the Merkle root — confirmed empirically harmless for convergence (`sorted-set-store` passes).
- Rebased on current master (post-#2579/#2583).


chefsale:
Thanks — addressed in 6487ec41:

- **`#[serial]` only advisory / registry clear unguarded (testing.rs:55, :230):** the harness now takes an internal process lock (`HARNESS_LOCK`) for the whole run, so it **self-serializes** — a forgotten `#[serial]` is at worst slower, never a corrupting race. Module docs updated; `#[serial]` is no longer load-bearing.
- **Genesis reuses replica 0's executor (testing.rs:220):** genesis now installs under a distinct `GENESIS_EXECUTOR` (`[0xEE; 32]`), so replica 0's writes are genuinely concurrent with genesis rather than a continuation — matches the runtime conformance harness. Convergence still holds.
- **Op model / `wins == 3` arithmetic (testing.rs:237, converge.rs:51):** clarified in the `ops` doc, the module doc, and inline — every replica applies *every* op (commutativity model), so an op appearing `k` times across `n` replicas runs `n × k` times. Expected value = `replicas × op-occurrences`.
- **Magic seed constants (testing.rs:195):** commented — odd Fibonacci-hashing constant (2^64/φ) for per-replica seed diffusion, reproducible from the base `seed`.
- **Invariants only after all deltas (testing.rs:290):** documented as intentional (final-state convergence check), in both the module doc and `invariant`'s docs.
- **"test expected to fail but not ignored" (team-metrics-custom:18):** that's the **stale PR description**, not the code — #2577 is now **merged (#2579)**, so the correctness tests assert `wins == 3` and **pass** (not ignored). I've corrected the PR description.

Rebased on current master (post-#2579/#2583); harness + all app convergence/correctness tests green.

PR #2579, opened 2026-06-01
fix(storage): auto-converge custom struct CRDT values via deterministic re-key (#2577)

### LINKED ISSUE #2577: Custom Mergeable types silently lose data on concurrent writes (nested/struct-wrapped CRDTs fall back to LWW) — follow-through on #1780/#1940

chefsale:
Fixes #2577.

Custom structs used as collection values (the `UnorderedMap<String, TeamStats>` pattern, where `TeamStats` wraps `Counter`s) silently **lost data** on concurrent same-key writes. Root cause (confirmed by code trace + probes): a `Collection` serializes only its **id handle** (`children_ids` is `#[borsh(skip)]`), so such a struct is stored as an opaque blob of handles. Its nested collections got **per-replica-random ids**, the blobs differed, and sync **LWW-overwrote** them — the inner counters never merged. The custom `Mergeable` impl was inert (it runs on empty shells), so the loss was silent.

### Fix: deterministic re-keying (no merge-callback needed)

Give such a struct's nested collections **deterministic ids**, so every replica derives identical handles, the blob matches, and the nested CRDTs converge as child entities — exactly how a root `#[app::state]` struct's collection fields already converge.

- **`#[derive(Mergeable)]`** (both the `calimero-sdk-macros` and `calimero-storage-macros` derives — apps use the latter) now also generates a `RekeyTarget` impl that re-keys each field under a field-namespaced child of the entry id.
- **`#[app::state]`** generates `__calimero_register_rekey()` — called at WASM module load and from the native `TestHost` bridge — registering a re-key thunk for every value type appearing in a state field.
- The dispatch uses two **autoref-specialization macros** (`rekey_field_if_supported!` / `register_rekey_if_supported!`): real re-key/registration for `RekeyTarget` types, a compile-time no-op for leaves (`String`, `LwwRegister<_>`, …) — so no trait bound is forced on value types and existing apps are unaffected. (Autoref needs a *concrete* type, hence macros, not generic fns.)
- **Hand-written `Mergeable`** structs must add a manual `RekeyTarget` — one worked example + doc comment in `apps/team-metrics-custom`.

### Validation

- `apps/team-metrics-{macro,custom}` two-replica convergence tests: wins **sum to 2** (was 1) — the real apps, no hand-written rekey on the derive path.
- No regressions: storage lib (530 pass), WASM `crdt_conformance` (14 pass, rebuilds the fixture with the new codegen), representative apps build.
- A storage integration probe (`crates/storage/tests/rekey_record.rs`) isolates the mechanism.

### Notes for review

- **Two `Mergeable` derives** exist (`sdk-macros` + `storage-macros`); both are patched. Worth deciding whether to consolidate.
- **Deep nesting** (custom struct → collection → *custom* struct) currently registers only one level from the root; built-in collections self-register, so nested *built-ins* are fine. Flagged as a follow-up.
- This is the fix that makes the convergence harness's `#[ignore]`d `.invariant()` tests in #2578 pass; when both land, #2578's `converge_app` should also call `T::__calimero_register_rekey()`.
- Includes the native env test getters (`root_hash()` / `take_last_artifact()`) the probes need.

---

> [!NOTE]
> **High Risk**
> Changes CRDT sync/convergence for collection values, macro-generated registration at WASM load, and a process-global rekey registry—incorrect re-key or registration could still cause silent data loss or divergence.
> 
> **Overview**
> Fixes **#2577**: custom structs stored as map values (e.g. `UnorderedMap<String, TeamStats>`) no longer lose concurrent counter updates to whole-value LWW. The change wires **deterministic re-keying** so nested collections under each entry get field-namespaced ids every replica agrees on.
> 
> **Storage & macros:** `RekeyTarget` and `field_child_id` are public; `rekey_field_if_supported!` / `register_rekey_if_supported!` dispatch via autoref. Both `#[derive(Mergeable)]` paths (SDK + storage-macros) now emit `RekeyTarget`. `#[app::state]` adds `__calimero_register_rekey()` and hooks it from WASM `__calimero_register_merge`, the native `TestHost` bridge, and tests. **team-metrics-custom** documents a manual `RekeyTarget` when `Mergeable` is hand-written.
> 
> **Tests & tooling:** Native `env::root_hash()` / `take_last_artifact()` support two-replica sync tests; `team-metrics-*` gain `rlib` + `tests/converge.rs`; `rekey_record.rs` proves registered vs unregistered behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 657f90a3217e745f225efe3e44d49001b290a994. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/team-metrics-custom/Cargo.toml
+++ b/apps/team-metrics-custom/Cargo.toml
@@ -7,7 +7,7 @@ repository.workspace = true
 license.workspace = true
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 calimero-sdk.workspace = true
--- a/apps/team-metrics-custom/src/lib.rs
+++ b/apps/team-metrics-custom/src/lib.rs
@@ -51,6 +51,32 @@ impl Mergeable for TeamStats {
     }
 }
 
+/// Deterministic re-keying for a HAND-WRITTEN CRDT-value struct (#2577).
+///
+/// `#[derive(Mergeable)]` generates this for you. When you implement `Mergeable`
+/// by hand (as above), you MUST also implement `RekeyTarget`, or this struct —
+/// stored as an `UnorderedMap` value — is last-writer-wins'd as an opaque blob
+/// and its counters silently lose concurrent increments. Re-key each nested
+/// collection field under a field-namespaced child of the entry id so every
+/// replica derives identical ids and the counters converge as child entities.
+impl calimero_storage::collections::rekey::RekeyTarget for TeamStats {
+    fn rekey_relative_to(&mut self, parent_id: calimero_storage::address::Id) {
+        use calimero_storage::collections::rekey::field_child_id;
+        calimero_storage::rekey_field_if_supported!(
+            &mut self.wins,
+            field_child_id(parent_id, "wins")
+        );
+        calimero_storage::rekey_field_if_supported!(
+            &mut self.losses,
+            field_child_id(parent_id, "losses")
+        );
+        calimero_storage::rekey_field_if_supported!(
+            &mut self.draws,
+            field_child_id(parent_id, "draws")
+        );
+    }
+}
+
 /// Application state
 #[app::state(emits = MetricsEvent)]
 pub struct TeamMetricsApp {
--- a/apps/team-metrics-custom/tests/converge.rs
+++ b/apps/team-metrics-custom/tests/converge.rs
@@ -33,6 +33,11 @@ fn team_stats_converge_to_summed_value_custom() {
     env::reset_environment();
     register_crdt_merge_for_test::<TeamMetricsApp>();
     calimero_sdk::event::register::<TeamMetricsApp>();
+    // `#[app::state]`-generated: registers re-key thunks for the value types of
+    // the root's collection fields. `teams: UnorderedMap<String, TeamStats>` →
+    // `TeamStats` is registered (and its hand-written `RekeyTarget` impl used).
+    // This is the WASM-load / TestHost-bridge path; we call it directly here.
+    // (One level deep — see `generate_rekey_register_method` for the scope.)
     TeamMetricsApp::__calimero_register_rekey();
 
     let a: Store = Default::default();
--- a/apps/team-metrics-custom/tests/converge.rs
+++ b/apps/team-metrics-custom/tests/converge.rs
@@ -1,0 +1,79 @@
+//! #2577 end-to-end for the HAND-WRITTEN `Mergeable` + manual `RekeyTarget` app.
+//! Two replicas concurrently record wins for the same team; the counters must
+//! SUM (not be lost to LWW). Own integration binary for isolation.
+
+#![allow(clippy::unwrap_used)]
+
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+use calimero_storage::collections::Root;
+use calimero_storage::env::{self, RuntimeEnv};
+use calimero_storage::interface::ApplyContext;
+use calimero_storage::register_crdt_merge_for_test;
+use calimero_storage::store::Key;
+use team_metrics_custom::TeamMetricsApp;
+
+type Store = Rc<RefCell<HashMap<[u8; 32], Vec<u8>>>>;
+
+fn env_for(s: &Store, ex: [u8; 32]) -> RuntimeEnv {
+    let r = s.clone();
+    let reader = Rc::new(move |k: &Key| r.borrow().get(&k.to_bytes()).cloned());
+    let w = s.clone();
+    let writer =
+        Rc::new(move |k: Key, v: &[u8]| w.borrow_mut().insert(k.to_bytes(), v.to_vec()).is_some());
+    let rm = s.clone();
+    let remover = Rc::new(move |k: &Key| rm.borrow_mut().remove(&k.to_bytes()).is_some());
+    RuntimeEnv::new(reader, writer, remover, [7u8; 32], ex)
+}
+
+#[test]
+fn team_stats_converge_to_summed_value_custom() {
+    env::reset_environment();
+    register_crdt_merge_for_test::<TeamMetricsApp>();
+    calimero_sdk::event::register::<TeamMetricsApp>();
+    TeamMetricsApp::__calimero_register_rekey();
+
+    let a: Store = Default::default();
+    let b: Store = Default::default();
+    env::with_runtime_env(env_for(&a, [1; 32]), || {
+        Root::new(TeamMetricsApp::init).commit();
+    });
+    *b.borrow_mut() = a.borrow().clone();
+
+    let da = env::with_runtime_env(env_for(&a, [1; 32]), || {
+        let mut app = Root::<TeamMetricsApp>::fetch().unwrap();
+        let _ = app.record_win("liverpool".into());
+        app.commit();
+        env::take_last_artifact().unwrap()
+    });
+    let db = env::with_runtime_env(env_for(&b, [2; 32]), || {
+        let mut app = Root::<TeamMetricsApp>::fetch().unwrap();
+        let _ = app.record_win("liverpool".into());
+        app.commit();
+        env::take_last_artifact().unwrap()
+    });
+
+    let (ha, wa) = env::with_runtime_env(env_for(&a, [1; 32]), || {
+        Root::<TeamMetricsApp>::sync(&db, &ApplyContext::empty()).unwrap();
+        let app = Root::<TeamMetricsApp>::fetch().unwrap();
+        (env::root_hash(), app.get_wins("liverpool".into()).unwrap())
+    });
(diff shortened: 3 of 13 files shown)


chefsale:
Thanks for the reviews — addressed in 34ff4a50:

- **Probe skipped in default CI (Cursor Bugbot)** — added a `[[test]] rekey_record` entry with `required-features = ["testing"]` in `crates/storage/Cargo.toml`, matching `merge_registry_*`. The probe now runs under the `testing` feature instead of compiling to an empty, always-green target.
- **No negative test (AI reviewer)** — added `unregistered_value_loses_data_pre_fix`: an *unregistered* struct value is LWW'd and loses an increment (`wins=1`), documenting the exact pre-fix failure mode and guarding against a future change that silently makes rekey unnecessary. Refactored the probe to a shared generic driver; both tests are `#[serial]` (the rekey registry is process-global with no reset).
- **Duplicate type registration (meroreviewer)** — `collect_type_paths` results are now deduplicated by token string before emitting `register_rekey_if_supported!`, so each distinct value type is registered once.
- **Autoref `'static` requirement (meroreviewer)** — documented on the `rekey_field_if_supported!` macro: the real-rekey arm only fires for `RekeyTarget + 'static`; all stored CRDT types are owned, so no current impact, but a future borrowed `RekeyTarget` would need `'static`.
- **Two `Mergeable` derives (meroreviewer)** — filed as follow-up #2580 to consolidate; both are patched here for now.

Doc-bot suggestions (AGENTS.md / CONTRIBUTING.md): the rekey machinery is internal codegen, not contributor-facing workflow, so I don't think those need updates for this PR — happy to add a note if a maintainer disagrees.

Still draft pending the merge-ordering discussion with #2578 (the convergence harness); per that thread, once this lands #2578's `converge_app` will call `__calimero_register_rekey()` and un-ignore its `.invariant()` correctness tests.

chefsale:
Round 2 addressed in 2b33a4d1:

- **Negative test too loose** — tightened `unregistered_value_loses_data_pre_fix` to assert the exact pre-fix mode: `converged && wa == wb && wa == 1`. Now a *different* regression (total loss → 0, or an accidental partial fix → 2) trips it instead of passing on `wa < 2`.
- **Deep-nesting scope** — documented on `generate_rekey_register_method`: it registers the value types of the **root**'s collection fields (one level of custom-struct nesting); built-in nesting works at any depth (self-registering ctors), but custom-struct-inside-custom-struct is the follow-up.
- **Autoref macro local-name collisions** — documented that each `rekey_field_if_supported!` expansion **must stay in its own block**; flattening could let the no-op arm win silently and regress to the data loss with no compile error. Today every expansion is its own `{{ … }}`.
- **Implicit `TeamStats` registration in the custom test** — added a comment explaining `__calimero_register_rekey()` collects `TeamStats` from the `UnorderedMap<String, TeamStats>` field type and uses its hand-written `RekeyTarget`. Kept it as the single registration path on purpose (a redundant explicit call would stop the test from exercising the generated path).

Tests still green: `rekey_record` (FIXED 2/2, UNFIXED 1/1), team-metrics-{macro,custom}.

chefsale:
Round 3 — these were doc-only, in 657f90a3:

**On the `'static` silent-fallthrough (3 comments):** this hazard cannot occur, so no enforcement is needed — it's already enforced. `RekeyTarget: Any`, and `Any: 'static`, so a non-`'static` `RekeyTarget` impl is a **compile error at the impl site** (E0478 — verified). My earlier doc note implied a real trap; that was misleading. I've corrected the trait doc and both macro docs to state that every `RekeyTarget` is necessarily `'static`, so the macros' `+ 'static` bound is always satisfied and there's no no-op fall-through. (Adding `'static` as a redundant supertrait would be a no-op since `Any` already implies it.)

**Negative test `== 1` "fragile" (2 comments):** it's deterministic, not flaky — added a comment explaining why. Each replica increments under a distinct executor id ([1;32] vs [2;32]); root-blob LWW breaks ties by executor id, so exactly one blob wins on both sides — never a tie that yields 0 or 2. Kept the tight `== 1` (round 1 asked to tighten it from `< 2`); a 0 or 2 reading would be a real tiebreak/serialization change, which is exactly what the guard should catch.

**Dedup comment misleading (syntactic vs semantic):** clarified — it's token-string (syntactic) dedup; differently-written-but-equal types still collapse to one `TypeId` via idempotent `or_insert`, so the registry holds one entry regardless.

**REGISTRY no-reset:** documented on the declaration — append-only and idempotent by design (a thunk is a pure function of its type, no stale entries possible); tests use distinct types for the unregistered path.

Over-collection of key types: left as-is — it's intentional and harmless (autoref no-ops non-`RekeyTarget` types), and collecting all args keeps the macro robust to type aliases/wrappers; documented as such.

PR #2584, opened 2026-06-01
fix(sync): converge clear()/deletions through HashComparison

chefsale:
## Problem

`clear()` (and entry deletions generally) could **split-brain** under HashComparison sync: two nodes settle on stable-but-different Merkle roots and HashComparison re-merges forever (`root_hash_verified=false`, looping local-vs-remote entity-count mismatch). This is the intermittent scaffolding-e2e clear-path flake.

**Root cause:** HashComparison reconciles entity trees by **add-wins union** and is **blind to tombstones**. Deletions (`Action::DeleteRef`) ride the **delta stream only** — `TreeNode` has no tombstone field and there is no `is_deleted` check anywhere in the HC reconcile/apply path. So when a node cleared an entry a peer still held, the deletion never reached the peer and the roots never converged. (The cleared node does *not* resurrect — the tombstone's high-water `updated_at` rejects the stale re-add — so the bug is non-propagation, not resurrection.)

## Fix

HashComparison now **propagates tombstones** with delete-wins, both directions:

1. **Initiator → peer (`EntityDeletePush`).** When the initiator finds a `remote_only` child it has locally tombstoned, it sends the tombstone (`id` + `deleted_at` + `metadata`) instead of pulling the peer's live copy. The responder applies it through the **authenticated `Action::DeleteRef`** path (delete-wins by HLC; signature/nonce verified for User/Shared). Handled in **both** responders (trait `run_responder_impl` + production `handle_tree_node_request`).

2. **Symmetric / one-round convergence.** When the node still *holding* the entry initiates, the cleared peer's now-childless node presents as a *leaf* (short-circuiting child comparison). The initiator detects the leaf-vs-internal mismatch and pushes its live leaves; the responder reports any newer tombstone it holds back in the `EntityPushAck`, which the initiator applies (delete-wins). Converges in a single round regardless of who initiates.

3. **Signed (User/Shared) deletions.** `persist_signed_signatures` skipped `DeleteRef`, so the signed signature never reached the tombstone and the peer's verify rejected it. It now persists the signed signature onto the (surviving) index entry via `update_signature_in_place`, so HC can ship a verifiable signed deletion.

No Merkle-hash change, no storage-format migration.

## Tests

- **`sync_sim::protocol::tests::test_hashcomparison_propagates_clear_tombstone`** — end-to-end repro driving the **real** `HashComparisonProtocol` between two sim nodes (cleared node initiates). Red before, green after.
- **`…::test_hashcomparison_clear_converges_when_holder_initiates`** — the symmetric direction (holder initiates), single round.
- **`hashcomparison_pull_does_not_resurrect_cleared_entry`** — unit guard: the apply path never resurrects a tombstoned id.
- **`update_signature_in_place_tests::patches_signature_on_tombstoned_entity`** — signed-delete persistence mechanism.

Verified green: 310 sync_sim, 162 lib-sync, 535 calimero-storage, 54 calimero-context; workspace builds; `cargo fmt` clean.

---

> [!NOTE]
> **High Risk**
> Adds new sync wire messages and changes HashComparison convergence behavior; peers must upgrade together and tombstone/DeleteRef handling must stay correct for signed entities.
> 
> **Overview**
> **HashComparison** no longer treats missing children as “always pull”: when the initiator sees a peer-only child but holds a **local tombstone** (`clear`/delete), it **pushes** that deletion instead of re-fetching live data—fixing add-wins reconciliation that could never drop entries the peer still had.
> 
> The change adds **`EntityDeletion`** on the wire plus **`EntityDeletePush`** / **`EntityDeletePushAck`**. After the Merkle DFS, tombstones are batched (like **`EntityPush`**); responders apply **`Action::DeleteRef`** (delete-wins by HLC, same auth path as deltas) in both production and protocol-trait responder loops. Stray **`EntityDeletePush`** outside an HC session is ignored in the sync manager.
> 
> **Tests:** sim **`test_hashcomparison_propagates_clear_tombstone`** (bidirectional real HC) and a unit guard that pulling a stale leaf does not resurrect a cleared entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5801f0244a5c44a6fc08dec9653e8b48b83a52bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Follow-up commit `e4c78e88` extends the fix to **LevelWise** — it had the identical add-wins/tombstone-blind bug, and it's selected exactly for wide/shallow trees (e.g. a large KV-store map), so a clear could route through it and still split-brain. LevelWise is pull-only, so the cleared initiator now propagates deletions via the same `EntityDeletePush` + authenticated `DeleteRef` apply (signed User/Shared deletes come free via the `persist_signed_signatures` fix). Added `execute_level_wise_sync` + `test_levelwise_propagates_clear_tombstone` driving the real protocol.

Full deletion coverage now: DeltaSync ✅ (always), HashComparison ✅, LevelWise ✅, Snapshot ✅ (fresh-node only). BloomFilter/SubtreePrefetch are unimplemented stubs (fall through to snapshot) — no entity reconcile, no tombstone bug.

311 sync_sim + 162 lib-sync + 535 storage + 54 context green; build + fmt clean.

Issue #2591, opened 2026-06-01
sync: HashComparison clear converges in one round regardless of which peer initiates


chefsale:
When the node that still *holds* an entry initiates the sync (rather than the node that cleared it), the cleared peer's now-childless node presents as a leaf, short-circuiting child comparison, so the deletion isn't propagated until the cleared node itself initiates (eventual convergence only).

Proposed fix (reverted from #2584, commit cab5979d): the responder reports its newer tombstone back in `EntityPushAck`, and the initiator detects the leaf-vs-internal mismatch and pushes its live leaves so the cleared peer can reject-and-report.

Context: split out of #2584 (HashComparison clear-tombstone propagation). These three were attempted in #2584 but **reverted** because together they introduced a deterministic `frozen-rga-convergence` e2e regression (`ActionNotAllowed("Cannot change StorageType")` panic in `__calimero_sync_next` during multi-node sync) whose mechanism was not pinned. The bug class does **not** reproduce in the in-process sim or the bare-runtime conformance harness — so each follow-up MUST land with an **e2e-level repro** (or a new harness that exercises the full HC/LevelWise protocol + namespace governance + multi-node) proving it does not regress frozen-rga.

🤖 filed via Claude Code

PR #2596, opened 2026-06-02
fix(sync): infer Frozen storage_type for bare HC/LevelWise leaf pushes

chefsale:
## Problem

A Frozen entity pushed via HashComparison / LevelWise as a bare leaf lands on a fresh receiver as **`Public`**, not `Frozen`. When the real `Frozen` entity later arrives via a delta, `apply_action`'s guard rejects the `Public → Frozen` storage-type change (`ActionNotAllowed("Cannot change StorageType")`) — and because the failing check runs inside the **WASM guest's** frozen-value merge (`__calimero_sync_next`), it surfaces as a guest panic and a multi-node split-brain.

Root cause: Frozen entities carry no wire authorization (`wire_authorization_for` returns `None` for `Frozen` — only `User`/`Shared` carry it). So in `apply_leaf_with_crdt_merge`, a freshly-received frozen leaf got `crdt_type = FrozenStorage` set correctly but `storage_type` left at the `Public` default (the "new entity, no wire-auth" case).

This was pinned by instrumenting the storage-type guard to embed the entity id + types into the error message (the guest panic carried it):
```
id=9bc70da3…  existing=Public  new=Frozen   ← a __frozen_storage_frozen_items child
```

## Fix

In `apply_leaf_with_crdt_merge`, when there's no wire-auth and no existing entity, **infer `storage_type = Frozen` from the wire-carried `crdt_type == FrozenStorage`** (the crdt_type *is* on the wire, so Frozen is recoverable). Public/Shared/User paths are unchanged.

## Why it matters

Latent for any Frozen entity pushed via HC/LevelWise as a bare leaf. It was surfaced deterministically by the reverted symmetric-convergence change (#2591), whose leaf-internal push newly sent frozen items down this path and broke `frozen-rga-convergence` 4×. This fix also **unblocks #2591** (the symmetric push becomes safe on top of it).

## Tests

- New unit test `apply_leaf_new_frozen_entry_lands_as_frozen_not_public` — a bare frozen leaf with no auth + no existing entity must land as `Frozen`.
- 310 sync_sim green; build + fmt clean.
- A verification run stacking this fix + the reverted #2591 change is in flight to confirm it resolves the original frozen-rga regression; result will be posted as a comment.

---

> [!NOTE]
> **Medium Risk**
> Touches sync apply metadata for new entities; wrong inference could still mis-classify storage, but the change is narrow, matches wire crdt_type, and is covered by a targeted unit test.
> 
> **Overview**
> HashComparison and LevelWise can apply **new** frozen entities from bare leaves that carry `CrdtType::FrozenStorage` but no wire authorization (frozen entries never get `wire_authorization_for`). **`apply_leaf_with_crdt_merge`** now sets **`StorageType::Frozen`** in that case instead of leaving **`Public`**, so a later delta cannot hit **`Cannot change StorageType`** and panic inside WASM frozen merge (#2591 split-brain).
> 
> A regression test **`apply_leaf_new_frozen_entry_lands_as_frozen_not_public`** asserts a first-time frozen HC leaf is stored as frozen, not public.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 430a53c6352650b18eb075aed005468c63a8c13c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/hash_comparison_protocol.rs
+++ b/crates/node/src/sync/hash_comparison_protocol.rs
@@ -1481,6 +1481,85 @@ mod tests {
         });
     }
 
+    /// A *new* Frozen entity pushed as a bare HC leaf must land as
+    /// `StorageType::Frozen`, not `Public`.
+    ///
+    /// Frozen entities carry no wire authorization (`wire_authorization_for`
+    /// returns None for Frozen), so before the fix a freshly-received frozen
+    /// leaf defaulted to `Public` (its `crdt_type` was set to `FrozenStorage`
+    /// but `storage_type` stayed `Public`). A later real `Frozen` delta then hit
+    /// `apply_action`'s guard with `existing=Public new=Frozen` →
+    /// `ActionNotAllowed("Cannot change StorageType")`, panicking the guest's
+    /// frozen-value merge (the HC/LevelWise frozen-push split-brain, #2591).
+    /// `apply_leaf_with_crdt_merge` now infers `Frozen` from the wire-carried
+    /// `crdt_type`.
+    #[test]
+    fn apply_leaf_new_frozen_entry_lands_as_frozen_not_public() {
+        use std::sync::Arc;
+
+        use calimero_node_primitives::sync::hash_comparison::{LeafMetadata, TreeLeafData};
+        use calimero_primitives::crdt::CrdtType;
+        use calimero_storage::action::Action;
+        use calimero_storage::entities::{Metadata, StorageType};
+        use calimero_storage::interface::ApplyContext;
+        use calimero_store::db::InMemoryDB;
+        use calimero_store::Store;
+        use sha2::{Digest, Sha256};
+
+        use crate::sync::helpers::apply_leaf_with_crdt_merge;
+
+        let context_id = ContextId::from([0xCC; 32]);
+        let identity = PublicKey::from([0u8; 32]);
+        let store = Store::new(Arc::new(InMemoryDB::owned()));
+        let runtime_env = create_runtime_env(&store, context_id, identity);
+
+        let root_id = Id::new(*context_id.as_ref());
+        let frozen_id = Id::new([0x77u8; 32]);
+
+        // Frozen content-addressed blob: [key_hash(32)][value][element_id(32)].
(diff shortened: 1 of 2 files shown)


chefsale:
⚠️ **Marking draft — this does NOT resolve #2591.** Verification (this fix + the reverted #2591 leaf-internal push, run 26800989167) **still failed frozen-rga** with the identical `Public→Frozen` panic, and it's raised in the **WASM guest's** `__calimero_sync_next` delta-apply — not the host-side `apply_leaf_with_crdt_merge` this PR touches. So the frozen item is still landing as `Public` via a DIFFERENT path (candidate: the snapshot apply a freshly-joined node receives, or the guest-side apply), which this PR doesn't cover.

This PR is still a **correct, self-contained latent-bug fix** (a new Frozen entity pushed as a bare HC/LevelWise leaf now lands as `Frozen` instead of `Public`, with a unit test) — but it is NOT sufficient for the #2591 regression. Keeping it as draft pending either (a) reframing as a standalone partial fix, or (b) folding into the full #2591 fix once the other path is instrumented + pinned.

chefsale:
**Reframed as a standalone fix (now ready for review).**

This PR is a self-contained correctness fix: a **new** Frozen entity pushed via HashComparison/LevelWise as a bare leaf was landing as `Public` (Frozen carries no wire auth, so `apply_leaf_with_crdt_merge` left `storage_type` at the Public default while setting `crdt_type=FrozenStorage`). A later real Frozen delta for that id would then be rejected with `Cannot change StorageType`. The fix infers `Frozen` from the wire-carried `crdt_type`. Unit test: `apply_leaf_new_frozen_entry_lands_as_frozen_not_public`; 310 sync_sim green.

**Does NOT claim to fix #2591.** Verification (run 26800989167) showed the #2591 frozen-rga regression persists via a *guest-side* apply path (`__calimero_sync_next`, likely the snapshot a freshly-joined node receives) that this host-side PR doesn't touch. Landing this on its own merit; #2591 stays open pending that separate path being pinned.

(NB: the PR description above still has the original over-claiming text — couldn't edit it via API due to a token scope limit; this comment is the authoritative framing.)

PR #2604, opened 2026-06-02
feat(sdk): host-backed tracing subscriber for WASM apps (#1657)

### LINKED ISSUE #1657: Implement custom tracing subscriber for WASM
Currently, if anyone would use `tracing` macros (`info!`, `debug!`, etc.) inside WASM application or in the crates that are imported to WASM, these logs won't be visible and will be not extractable.

To improve it, it's possible to make a custom implementation of `tracing_subscriber::fmt::MakeWriter` that would internally use log method/s from host functions to propagate these logs outside. Ideally, there should be a way to specify the desired log level in the app: for example, either hardcoded in the app's ctor, or via a separate service method in the app (maybe `app::logic` can generate one).

chefsale:
Closes #1657.

## Problem

`tracing` is a facade — its macros (`info!`/`debug!`/`warn!`/…) forward each event to whatever subscriber is installed, and with **none** installed they're dropped before they're even formatted. WASM guests never installed one, so:

- App code using idiomatic `tracing` macros produced **no output**, silently.
- Worse, any crate the app *imports* that logs via `tracing` — notably **`calimero_storage`**, which is heavily instrumented (`Interface::apply_action`, `commit_root`, `merge`, HashComparison apply, index, rotation log, rekey, collections) and runs inside the guest — was completely invisible from the WASM side.

The only egress from the guest is the `log_utf8` host function (what `env::log` / `app::log!` use), so `env_logger`-style backends don't apply: there's no guest stderr and no `RUST_LOG` env to read.

## Fix

An opt-in, host-backed `tracing` subscriber whose `MakeWriter` forwards each formatted line through `log_utf8`, so `tracing` output lands in the execution `Outcome` next to `app::log!`.

- **New `calimero-sdk` `tracing` feature**, off by default (zero binary-size cost for apps that don't want it). When enabled, `#[app::logic]` exports auto-install the subscriber at method entry, alongside `setup_panic_hook` — no app setup required.
- **`env::set_log_level(LevelFilter)`** retunes verbosity at runtime: the level lives in a global atomic re-read by the filter per event, so it takes effect even after install. Default WARN — enabling the feature won't flood the buffer with dependency-crate INFO logs; apps raise it to INFO/DEBUG via `set_log_level`.
- **wasm32 gotchas handled**: `without_time()` (the default fmt timer's `SystemTime::now()` traps on `wasm32-unknown-unknown`), `with_ansi(false)`, install-once via `Once`.

### Log-limit tuning

`tracing` emits many short lines, so the per-execution log budget is redistributed to favor count over size:

| Limit | Before | After |
|---|---|---|
| `max_logs` | 100 | **1024** |
| `max_log_size` | 16 KiB | **16 KiB (unchanged)** |
| worst-case buffer | 1.6 MiB | ~16 MiB (transient, in `Outcome`) |

Per-line cap kept at 16 KiB: an over-long line *traps* the execution (`LogLengthOverflow`) rather than truncating, and dependency-crate `tracing` (e.g. `calimero_storage` merge logs) can exceed a few KiB. The budget is spent on line count instead.

## Why this matters

This makes the storage crate's own apply/merge/commit play-by-play observable from inside the guest at DEBUG — previously a blind spot exactly where many divergence/split-brain bugs live. The node already logs the *host* side of sync; this closes the *guest* side.

## Tests

- **3 SDK unit tests** — writer forwarding, level filtering, runtime level change.
- **2 runtime e2e tests** (`crates/runtime/tests/tracing_logs.rs`) driving a real compiled app (`scaffolding-e2e`, built with the feature) through `Module::run`:
  - app `tracing` reaches `Outcome.logs` with level rendered; INFO/DEBUG filtered out at the WARN default;
  - at DEBUG, `calimero_storage`'s **own** tracing surfaces (the imported-dependency case).

Verified: `wasm32-unknown-unknown` compile with the feature on; feature-off app build unchanged; no `crdt_conformance` regression; `cargo fmt` clean.

- Per-target filtering (e.g. only `calimero_storage::merge` at DEBUG) so heavy storage tracing can be scoped without the global firehose — the `filter_fn` already receives full `Metadata`.
- Wiring `VMLimits` to node config so operators can raise the cap per node.

---

> [!NOTE]
> **Medium Risk**
> Raises per-execution log buffer headroom and adds optional guest logging that can amplify `Outcome` size or hit `LogsOverflow` under DEBUG; feature is off by default so production wasm size/behavior is unchanged unless opted in.
> 
> **Overview**
> Adds an **opt-in** `calimero-sdk` **`tracing`** feature that installs a host-backed `tracing-subscriber` at each generated WASM export entry (`init_logging` next to `setup_panic_hook`), forwarding formatted lines through `env::log` into the execution **`Outcome`**. Apps can tune verbosity at runtime via **`env::set_log_level`** (default **WARN** to avoid flooding from dependency crates like **`calimero_storage`**).
> 
> **Runtime** default **`max_logs`** is raised from **100 → 1024** (per-line cap stays **16 KiB**; over-long lines still trap). **`scaffolding-e2e`** gains a **`tracing_probe`** method and enables the feature for tests; **`crates/runtime/tests/tracing_logs.rs`** runs real WASM through `Module::run` to assert level filtering and that storage-crate `tracing` reaches `outcome.logs` at DEBUG.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac51c36bab9fabc9284c9cba12c66966b7ce413a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1588,6 +1588,8 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.10.9",
+ "tracing",
+ "tracing-subscriber",
  "trybuild",
 ]
 
@@ -8736,6 +8738,7 @@ dependencies = [
  "serde_json",
  "sha2 0.10.9",
  "thiserror 1.0.69",
+ "tracing",
 ]
 
 [[package]]
--- a/apps/scaffolding-e2e/Cargo.toml
+++ b/apps/scaffolding-e2e/Cargo.toml
@@ -12,12 +12,16 @@ crate-type = ["cdylib"]
 
 [dependencies]
 thiserror.workspace = true
-calimero-sdk.workspace = true
+# `tracing` feature routes `tracing` macro output (this app's and the storage
+# crate's) through the host log; exercised by `tracing_probe` + the runtime
+# `tracing_logs` integration test.
+calimero-sdk = { workspace = true, features = ["tracing"] }
 calimero-storage.workspace = true
 calimero-storage-macros.workspace = true
 sha2.workspace = true
 hex.workspace = true
 bs58.workspace = true
+tracing.workspace = true
 
 [build-dependencies]
 calimero-wasm-abi.workspace = true
--- a/apps/scaffolding-e2e/src/lib.rs
+++ b/apps/scaffolding-e2e/src/lib.rs
@@ -455,6 +455,27 @@ impl E2eKvStore {
         Ok(())
     }
 
+    /// Test-only probe for the host-backed `tracing` subscriber.
+    ///
+    /// Emits one `tracing` event at each level and performs a KV insert so the
+    /// storage crate's *own* `tracing` output (e.g. `Interface::apply_action`,
+    /// `commit_root`) is exercised through the same path. When `debug` is set,
+    /// raises the level to DEBUG first so those lower-level lines pass the
+    /// filter. Driven by the runtime `tracing_logs` integration test; the
+    /// convergence tests never call it, keeping DEBUG spam out of them.
+    pub fn tracing_probe(&mut self, debug: bool) -> app::Result<()> {
+        if debug {
+            env::set_log_level(env::LevelFilter::DEBUG);
+        }
+        tracing::info!(probe = true, "tracing_probe: info line");
+        tracing::debug!(probe = true, "tracing_probe: debug line");
+        tracing::warn!("tracing_probe: warn line");
+        // Mutating storage drives the storage crate's internal `tracing`.
+        self.kv_items
+            .insert("probe_key".to_owned(), "probe_value".to_owned().into())?;
+        Ok(())
+    }
+
     /// KV set with handler triggers (for testing event-driven handlers)
     pub fn set_with_handler(&mut self, key: String, value: String) -> app::Result<()> {
         app::log!("Setting key with handler: {:?} to value: {:?}", key, value);
--- a/crates/runtime/Cargo.toml
+++ b/crates/runtime/Cargo.toml
@@ -43,6 +43,7 @@ eyre.workspace = true
 owo-colors.workspace = true
 rand.workspace = true
 serde_json.workspace = true
+wasmer.workspace = true
 wat.workspace = true
 
 [features]
--- a/crates/runtime/src/logic.rs
+++ b/crates/runtime/src/logic.rs
@@ -110,9 +110,22 @@ const DEFAULT_MAX_REGISTERS: u64 = 100;
 /// Default maximum register size in MiB (100 MiB).
 const DEFAULT_MAX_REGISTER_SIZE_MIB: u64 = 100;
 /// Default maximum number of log entries.
-const DEFAULT_MAX_LOGS: u64 = 100;
-/// Default maximum log message size in KiB (16 KiB).
-const DEFAULT_MAX_LOG_SIZE_KIB: u64 = 16;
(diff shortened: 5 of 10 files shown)


chefsale:
Thanks bots — triaged the automated findings. Note many targeted the first commit (INFO default + 4 KiB cap) and are already resolved by the WARN/16 KiB fix.

**Addressed in `ac51c36b`:**
- **Freshness check too narrow** (3 reviewers) — `tracing_logs` now walks `src/` recursively + `Cargo.toml` + `build.sh` (mirrors the `crdt_conformance` fixture) instead of only `src/lib.rs`.
- **Duplicated subscriber config** — extracted one `build_subscriber()` used by both `init()` and the tests.
- **Relaxed atomic ordering** — `MAX_LEVEL` now uses Acquire/Release for native multi-thread visibility.
- **`TEST_LOCK` poisoning** — tests lock via a helper that recovers a poisoned lock (`into_inner()`), so one panicking test can't cascade.
- **Sensitive data at DEBUG** — documented on `set_log_level`: DEBUG forwards dependency-crate internals into the `Outcome`; it's an opt-in debug aid (default is WARN).

**Already fixed by the WARN/16 KiB commit (`8b83df19`):**
- *Reduced max_log_size traps apps* (Bugbot) / *4 KiB cap* — reverted to 16 KiB; over-long lines trap, so the cap isn't shrunk.
- *INFO floods the buffer* — default is now WARN.
- *PR description says INFO* — description updated to match.

**Declining, with rationale:**
- *`from_utf8_lossy` may corrupt* — intentional: this is a `Drop`, which must never panic; the fmt layer emits valid UTF-8 in practice, and lossy is the safe fallback.
- *Over-long line traps (DoS) / unbounded writer buffer* — acknowledged tradeoff. The buffer is bounded by the formatted event the app chose to emit, and `max_log_size` is the guard. Truncate-instead-of-trap is a separate host-behavior decision; noted as a follow-up rather than changed here.
- *`Once` doesn't retry install* — intentional; `set_global_default` rejects a second subscriber anyway.

Memory worst-case is governed by `max_logs`(1024) × `max_log_size`(16 KiB) ≈ 16 MiB transient; an operator-tunable `VMLimits` knob is a tracked follow-up.

chefsale:
The earlier `Test (frozen-rga-convergence)` red was **not** a regression from this PR — it was the known flaky *"same DAG heads, different root hash"* divergence on `__frozen_storage_frozen_items`.

Evidence:
- The merobox workflow printed `Workflow completed successfully` and **all convergence assertions passed**; the step failed only because `--e2e-mode` scans node logs and trips on the heartbeat's `DIVERGENCE DETECTED` ERROR line.
- Divergence dump shows all 23 ROOT child `(entity_id, merkle_hash)` pairs **identical** across nodes (identical children, different root fold) — the pre-existing determinism bug, not content corruption.
- This PR cannot cause it: the only app change is an uncalled `tracing_probe`; `runtime/logic.rs` only raises log-buffer limits; the sdk/subscriber changes alter `log_count` (metadata), not state.
- A clean **re-run is green end-to-end** ([26809346175](https://github.com/calimero-network/core/actions/runs/26809346175)), `Test (frozen-rga-convergence): success`.

All required checks (Rust / Lint / Build Binaries / frozen-rga) are now passing. Ready for review.

PR #2605, opened 2026-06-02
feat(sync): carry deletions in the comparison tree (tombstones-in-tree, #2591)

### LINKED ISSUE #2591: sync: HashComparison clear converges in one round regardless of which peer initiates

chefsale:
Closes #2591.

## Problem

HashComparison and LevelWise reconcile entity trees **add-wins** during comparison. Deletions only ever rode the delta stream or an `EntityDeletePush` sent by the node that *did* the clear. So a clear converged only when the **cleared** node happened to initiate:

- **HC, holder initiates:** the cleared collection is a now-childless node; comparison sees "no children" and the holder keeps its stale copy.
- **LevelWise, holder initiates:** the cleared collection returns an empty level; the initiator hits its empty-nodes early-out and exits still holding the entry.

Either way the deletion sat unpropagated until roles flipped — an intermittent clear split-brain (the `scaffolding-e2e` clear flake lineage).

## Fix — make deletions first-class on the comparison wire

Convergence becomes symmetric in a single round regardless of initiator, **without pushing any live entity** (which would re-enter the Frozen-as-Public apply path that regressed frozen-rga before):

- **storage** — `EntityIndex` gains `deleted_children: Vec<Id>`, appended when a child is removed from its parent and cleared when the child is re-added.
- **HashComparison** — `TreeNode` carries `deleted_children: Vec<EntityDeletion>`. A now-childless cleared collection is emitted as an *internal* node carrying them. The initiator applies remote `deleted_children` via the authenticated `Action::DeleteRef` path (delete-wins by HLC; sig/nonce verified for User/Shared; safe no-op on LWW loss / auth fail).
- **LevelWise** — the level response carries `deleted_children`; the initiator applies them **before** its empty-level early-out, so a cleared collection converges on the first pull.

Tombstone resolution + application reuse the `EntityDeletion` wire type and apply path from #2584. The cleared-node-initiates direction still flows via the existing `EntityDeletePush` — this PR adds the *other* direction.

## Wire / state-format note

The `EntityIndex` borsh layout change is **additive** but there is no migration hook — nodes upgrading need fresh state. `MessagePayload::LevelWiseResponse` and `TreeNode` gain a field (older peers can't decode the new layout; this is a coordinated-upgrade change like the prior sync-wire additions).

## Tests

- `test_hashcomparison_clear_converges_when_holder_initiates` — holder initiates, converges in one HC round.
- `test_levelwise_clear_converges_when_holder_initiates` — holder initiates, converges in one LevelWise pull.
- Existing `test_{hashcomparison,levelwise}_propagates_clear_tombstone` (cleared-node-initiates) still green.
- Full `sync_sim` (312), `calimero-storage` lib (544), `calimero-node` lib (241), `calimero-node-primitives` lib (163) green; `cargo fmt --check` clean; `cargo build --workspace` clean.

## Gate

⚠️ Draft until the **frozen-rga convergence** e2e job is green on this branch (run dispatched: https://github.com/calimero-network/core/actions/runs/26806974408). The whole design exists to converge clears *without* re-treading the live-entity Frozen path that regressed frozen-rga previously; this PR stays draft until that's proven.

---

> [!NOTE]
> **High Risk**
> Changes persistent `EntityIndex` borsh layout and sync wire messages (`TreeNode`, `LevelWiseResponse`), requiring coordinated rollout; deletion apply path is security-sensitive (authenticated `DeleteRef`, delete-wins).
> 
> **Overview**
> Fixes **clear split-brain** when the node that still holds an entry initiates HashComparison or LevelWise: add-biased tree comparison could not see removals, so a cleared peer looked “empty” and the holder never learned the delete until roles flipped.
> 
> **Storage** adds `EntityIndex::deleted_children` (recorded on `remove_child_from`, cleared on re-add). **HashComparison** puts resolved `EntityDeletion` tombstones on `TreeNode::deleted_children`, treats cleared parents as **internal** nodes even with no live children, and applies remote tombstones via authenticated `DeleteRef` during traversal. **LevelWise** adds `deleted_children` on `LevelWiseResponse`, aggregates them in `get_nodes_at_level`, and applies them **before** the empty-level early-out. Wire/index layout changes need a **coordinated upgrade** (no migration hook).
> 
> New `sync_sim` tests cover holder-initiates clear for both protocols in one round, without pushing live entities.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f9d5f2eb05d3709c0589541540e812605f8918c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/primitives/src/sync/hash_comparison.rs
+++ b/crates/node/primitives/src/sync/hash_comparison.rs
@@ -186,6 +186,16 @@ pub struct TreeNode {
 
     /// Leaf data (present only for leaf nodes).
     pub leaf_data: Option<TreeLeafData>,
+
+    /// Tombstones for children removed from this (internal) node.
+    ///
+    /// A deletion is the absence of a child from `children`, which the
+    /// add-biased comparison can't observe. Carrying the tombstones here lets a
+    /// peer that still holds a child converge to the deletion (delete-wins)
+    /// during comparison — without anyone pushing the live entity. Empty for
+    /// leaf nodes. Populated by `get_local_tree_node` from the parent index's
+    /// `deleted_children` (each entry resolved to a signed `EntityDeletion`).
+    pub deleted_children: Vec<EntityDeletion>,
 }
 
 impl TreeNode {
@@ -197,6 +207,7 @@ impl TreeNode {
             hash,
             children,
             leaf_data: None,
+            deleted_children: Vec::new(),
         }
     }
 
@@ -208,6 +219,7 @@ impl TreeNode {
             hash,
             children: vec![],
             leaf_data: Some(data),
+            deleted_children: Vec::new(),
         }
     }
 
@@ -225,13 +237,15 @@ impl TreeNode {
             return false;
         }
 
-        // Check structural invariant: must have exactly one of children or data
-        // - Internal nodes: non-empty children, no leaf_data
-        // - Leaf nodes: empty children, has leaf_data
-        let has_children = !self.children.is_empty();
-        let has_data = self.leaf_data.is_some();
-        if has_children == has_data {
-            // Invalid: either both present (ambiguous) or both absent (empty)
+        // Check structural invariant: must be exactly one of internal or leaf.
+        // - Internal node: has live children OR only-tombstoned children
+        //   (a parent cleared to childless still carries `deleted_children`),
+        //   and no leaf_data.
+        // - Leaf node: has leaf_data, no children, no deleted_children.
+        let is_internal = !self.children.is_empty() || !self.deleted_children.is_empty();
+        let is_leaf = self.leaf_data.is_some();
+        if is_internal == is_leaf {
+            // Invalid: either both (ambiguous) or neither (empty).
             return false;
         }
 
@@ -1057,6 +1071,7 @@ mod tests {
             id: [1; 32],
             hash: [2; 32],
             children: vec![[3; 32]],
+            deleted_children: vec![],
             leaf_data: Some(leaf_data),
         };
         assert!(!invalid_node.is_valid());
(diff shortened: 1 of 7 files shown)


chefsale:
✅ Gate cleared. E2E run [26806974408](https://github.com/calimero-network/core/actions/runs/26806974408) is **green end-to-end**, including `Test (frozen-rga-convergence): success` — the regression this design was built to avoid. The tombstones-in-tree approach converges holder-initiated clears without ever pushing a live entity, so the Frozen-as-Public apply path is never re-entered. Marking ready for review.

PR #2614, opened 2026-06-02
fix(ci): make worker nodes produce CPU profiles in fuzzy profiling

### LINKED ISSUE #2373: fuzzy profiling: CPU flamegraphs only render for the seed node, not worker nodes
## Summary

In the Fuzzy Load Test profiling output, **CPU flamegraphs render only for node-1 (the context seed)**. The 3 worker nodes per suite (`*-node-2/3/4`) produce **no CPU profile at all** — only memory flamegraphs.

Split off from #2342 (which restored the seed-node CPU profile + fixed the harvest path and the image rebuild). Worker CPU profiling is a pre-existing, separate gap.

## Evidence

Fuzzy run [#541 / `25968349221`](https://github.com/calimero-network/core/actions/runs/25968349221), `profiling-data-group-governance` artifact:

| Node | `perf-*.data` | `perf-*.log` | CPU flamegraph | Memory flamegraph |
|------|------|------|------|------|
| `fuzzy-gov-node-1` (seed) | ✅ 0.544 MB, 2170 samples | ✅ | ✅ | ✅ |
| `fuzzy-gov-node-2` (worker) | ❌ absent | ❌ absent | ❌ absent | ✅ |
| `fuzzy-gov-node-3` (worker) | ❌ absent | ❌ absent | ❌ absent | ✅ |
| `fuzzy-gov-node-4` (worker) | ❌ absent | ❌ absent | ❌ absent | ✅ |

Workers have only a `perf-*.map` file (merod's JIT symbol map, written by the runtime independently of `perf`). The absence of both `perf-*.data` and `perf-*.log` means `perf record` never produced output on the worker nodes.

## Key facts

- Reproduces identically on the **stale image** (run `25934709640`) and the **rebuilt image** (run `25968349221`) — so `8a51e6a8`'s build-time perf symlink did **not** fix it.
- All 4 nodes per suite run the **same image**, so the worker failure is **environmental/runtime**, not an image-build problem.
- **Memory** flamegraphs (jemalloc heap) render on all 4 nodes — the profiling harness, bind mount, harvest, and render pipeline are all fine. The failure is specific to `perf record` on worker containers.

## Next step

Pull the worker container `[Profiling]` stdout from the `fuzzy-test-*` artifact (`fuzzy-test-logs/<suite>/docker-logs/<container>.log`) to see the `start_profiling` / `perf_sanity_check` output on nodes 2-4 — likely shows the `perf record -o /dev/null -- true` sanity check failing (CAP_PERFMON / `perf_event_paranoid` / per-container capability difference vs the seed).

## Scope

CI/profiling only — no impact on merod runtime behavior. Memory profiling and seed-node CPU profiling are unaffected.

chefsale:
Closes #2373.

## Problem

In Fuzzy Load Test profiling, **CPU flamegraphs rendered only for the seed node (node-1)**; the 3 worker nodes per suite (`*-node-2/3/4`) produced no `perf-*.data`/`perf-*.log` at all — only memory flamegraphs. Reproduced every run.

## Root cause

Not a CAP_PERFMON / `perf_event_paranoid` difference (those are host-wide and identical for all containers). The defect is the perf sanity probe:

```bash
perf record -o /dev/null -- true   # gated on the exit code of profiling `true`
```

Profiling a process that exits in **microseconds** makes the probe's exit status depend on host load and process start order, not on whether perf can open events. The seed starts first and least-loaded → passes. Workers start ~6s apart, after the seed's continuous 16-min `perf record` is already running and the host is busier → the throwaway probe exits non-zero → CPU profiling is wrongly vetoed, even though those nodes can profile their own long-running merod fine. Deterministic by start order ⇒ reproduces every run.

Aggravating factors:
- The same fragile probe existed **twice** — `perf_sanity_check` (nested in `install_kernel_tools`) and an inline copy in `start_profiling`.
- `head -5` truncation buried the decisive stderr line behind a benign `kptr_restrict` warning, so the artifact alone couldn't pin it.

## Fix

`scripts/profiling/entrypoint-profiling.sh`:

- **Hoist one robust `perf_sanity_check` to top level** and use it at both sites (no more divergent copies).
- **Profile `sleep 0.3`, not instant `true`** — at the sampling frequency a usable perf captures enough samples to exit cleanly.
- **Decide on stderr content, not raw exit code** — a `kptr_restrict`/`perf_event_paranoid` warning about kernel-symbol resolution is benign; only a real `perf_event_open`/kernel-tools failure (matched by pattern) is fatal. Logs **full** stderr on genuine failure.

**Safety:** the build-time check still fails on the fatal `not found for kernel` pattern, so the apt fallback still runs; after apt installs the matching `linux-tools` the post-install check passes on all nodes. `stop_profiling` already warns on a tiny/empty `perf-*.data` if perf genuinely can't record.

## Testing

- `bash -n` passes.
- Mocked-`perf` exercise of all three paths:
  - clean (exit 0) → usable, silent ✅
  - benign kptr warning + exit 1 (the worker-node bug) → **treated as usable** ✅
  - fatal `Operation not permitted` + exit 255 → fails, full stderr logged ✅

## Scope

CI/profiling only — `scripts/profiling/entrypoint-profiling.sh` gates the `merod:edge-profiling` rebuild, so the next fuzzy run picks it up. No merod runtime impact; memory profiling and seed-node CPU profiling unaffected.

---

> [!NOTE]
> **Low Risk**
> CI/profiling entrypoint only; no merod runtime or auth/data paths touched.
> 
> **Overview**
> **Unifies and hardens the `perf` sanity probe** in `scripts/profiling/entrypoint-profiling.sh` so fuzzy load-test **worker nodes** get CPU profiles, not only the first-started seed node.
> 
> The probe moves to a **single top-level `perf_sanity_check`**, shared by `install_kernel_tools` and `start_profiling` (replacing a nested copy plus an inline `perf record … true` guard). It records **`sleep 0.3`** instead of instant `true`, treats **fatal stderr patterns** (e.g. permission / missing kernel tools) as failure while **benign warnings** (e.g. `kptr_restrict`) can still count as usable, logs **full stderr** on real failure, and sets **`PERF_USABLE`** so `start_profiling` skips a second probe after a successful install-time check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c5aac26c39847060ba8a7836cae5bc1cc63ac01. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/profiling/entrypoint-profiling.sh
+++ b/scripts/profiling/entrypoint-profiling.sh
@@ -17,18 +17,40 @@ EXIT_CODE=0
 mkdir -p "$PROFILING_OUTPUT_DIR"
 mkdir -p "${PROFILING_REPORTS_DIR:-/profiling/reports}"
 
+# Probe whether `perf` can actually record on this container. Hoisted to top
+# level so the build-time check (install_kernel_tools) and the runtime guard
+# (start_profiling) share one robust implementation instead of two divergent
+# copies.
+#
+# Profile a short `sleep 0.3`, NOT an instant `true`: at the sampling
+# frequency this captures enough samples for a usable perf to exit cleanly.
+# We decide on stderr CONTENT, not the raw exit code — profiling a
+# microsecond-lived workload makes the exit status depend on host load and
+# process ordering, which passed on the first-started, least-loaded node and
+# spuriously failed on the later, busier nodes even though perf was equally
+# usable on all of them. A kptr_restrict / perf_event_paranoid *warning*
+# about kernel-symbol resolution is benign; only a real perf_event_open or
+# kernel-tools failure is fatal.
+perf_sanity_check() {
+    local err rc
+    err=$(perf record -o /dev/null -- sleep 0.3 2>&1)
+    rc=$?
+    if printf '%s\n' "$err" | grep -qiE \
+        'operation not permitted|permission denied|perf_event_open|not found for kernel|no such file or directory|cannot (find|open|read)'; then
+        echo "[Profiling] perf sanity check failed (rc=$rc). Full stderr:"
+        printf '%s\n' "$err" | sed 's/^/[Profiling]   /'
+        return 1
+    fi
+    if [ "$rc" -ne 0 ]; then
+        echo "[Profiling] perf exited $rc with no fatal error; treating as usable. stderr:"
+        printf '%s\n' "$err" | sed 's/^/[Profiling]   /'
+    fi
+    return 0
+}
+
 install_kernel_tools() {
     local kernel_version=$(uname -r)
     echo "[Profiling] Detected kernel: $kernel_version"
-
-    perf_sanity_check() {
-        local err
-        err=$(perf record -o /dev/null -- true 2>&1)
-        if [ $? -eq 0 ]; then return 0; fi
-        echo "[Profiling] perf sanity check failed. First 5 lines of stderr:"
-        echo "$err" | head -5 | sed 's/^/[Profiling]   /'
-        return 1
-    }
 
     # Happy path: the Dockerfile pre-symlinks the generic perf at
     # /usr/local/bin/perf, which takes PATH precedence over Ubuntu's
@@ -75,8 +97,8 @@ start_profiling() {
         return
     fi
     
-    if ! perf record -o /dev/null -- true 2>/dev/null; then
-        echo "[Profiling] perf not compatible, skipping CPU profiling"
+    if ! perf_sanity_check; then
+        echo "[Profiling] perf not usable, skipping CPU profiling"
         return
     fi
     
--- a/scripts/profiling/entrypoint-profiling.sh
+++ b/scripts/profiling/entrypoint-profiling.sh
@@ -31,12 +31,29 @@ mkdir -p "${PROFILING_REPORTS_DIR:-/profiling/reports}"
 # usable on all of them. A kptr_restrict / perf_event_paranoid *warning*
 # about kernel-symbol resolution is benign; only a real perf_event_open or
 # kernel-tools failure is fatal.
+#
+# Set to "true" by the first successful probe so start_profiling can skip a
+# redundant re-check after install_kernel_tools has already confirmed perf.
+PERF_USABLE=''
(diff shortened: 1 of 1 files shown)


chefsale:
Thanks for the reviews — addressed in `377ed8ea`.

**Fixed:**

- **Missing `perf` binary treated as usable** (cursor bugbot): real regression. A pattern-based check can't match `command not found` (rc=127), so the probe would have fallen through to "usable" and the apt fallback would never fire. Added a `command -v perf >/dev/null 2>&1 || return 1` short-circuit at the top — which also avoids spawning the `sleep 0.3` probe when perf is absent (covers meroreviewer's PATH-guard suggestion).
- **`perf_event_open` false-positive** (meroreviewer): anchored to a failure context — `perf_event_open.*(fail|return|error)`. The bare token can appear in benign informational lines; the real fatal line (`...perf_event_open() ... returned ... Operation not permitted`) is still caught by the permission patterns.
- **Redundant probe / latency** (meroreviewer + MeroReviewer): cache the first successful probe in `PERF_USABLE` and skip the re-check in `start_profiling`, which runs after `install_kernel_tools` has already confirmed perf. Happy path now probes once.

**Declined, with rationale:**

- *"runs `sleep 0.3` even when `ENABLE_PERF=false`"* — both call sites (`install_kernel_tools` and `start_profiling`) are already gated on `ENABLE_PERF=true`, so the probe never runs when perf is disabled.
- *"add `failed to open|unable to open|error:` to the fatal pattern list"* — declined deliberately. `perf record` prints benign warnings like `Failed to open /proc/kcore` *on success*, and `error:` is broader still, so these tokens would reintroduce the exact load/ordering false-positive class this PR removes. The `command -v` guard plus the benign-`kptr_restrict` philosophy is the safer call.

Re-ran the mocked-`perf` matrix: clean exit-0, benign-warning+exit1 (the worker bug), fatal `perf_event_open ... returned`, benign `perf_event_open: using N threads`+exit1 (correctly *not* flagged), and missing-binary (fails fast, no sleep) — all behave as intended.

PR #2620, opened 2026-06-02
fix(node): guard snapshot sync against LZ4 decompression bomb

### LINKED ISSUE #1747: Guard against LZ4 decompression bomb in snapshot sync
## Problem
`lz4_flex::decompress_size_prepended` allocates based on the embedded size before any bounds check, so a malicious peer can force huge allocations.

## Location
`crates/node/src/sync/snapshot.rs:423-431`

## Impact
Remote DoS on snapshot sync via OOM or excessive memory pressure.

## Suggested fix
Enforce a maximum allowed uncompressed size before decompression (e.g., cap using `DEFAULT_PAGE_BYTE_LIMIT` or a protocol max), or use a bounded/streaming decompressor that validates size.

chefsale:
## Problem

`lz4_flex::decompress_size_prepended` reads the embedded 4-byte size prefix and pre-allocates that many bytes **before any bounds check**, so a malicious peer can force a huge allocation (remote DoS via OOM) during snapshot sync. The only size validation (`decompressed.len() != uncompressed_len`) ran *after* the allocation and compared against the attacker-controlled `uncompressed_len`.

Location: `crates/node/src/sync/snapshot.rs`.

## Fix

Extract `decompress_snapshot_page(payload, uncompressed_len)`, which:

1. **Rejects oversized pages before allocating** — bails if `uncompressed_len > DEFAULT_PAGE_BYTE_LIMIT` (64 KB), so neither the declared length nor the forged embedded prefix can drive a large allocation.
2. **Decompresses into a fixed, capped buffer** via `lz4_flex::decompress_into`, which errors instead of growing the output — unlike `decompress_size_prepended`, which honors the embedded prefix. This rejects both a forged size prefix and a genuine decompression bomb rather than expanding it.
3. Bounds-checks the 4-byte prefix slice (no panic on short payloads) and verifies the written length matches.

## Tests

Added unit tests, all passing:
- round-trip of a valid page
- oversized declared length rejected pre-allocation
- forged `u32::MAX` size prefix rejected without a giant allocation
- sub-4-byte payload errors instead of panicking

Fixes #1747

---

> [!NOTE]
> **Medium Risk**
> Changes untrusted-input handling on the snapshot sync path (DoS/OOM fix); behavior is tighter but legitimate oversized single-record pages remain allowed up to the 4 MB protocol cap.
> 
> **Overview**
> Hardens **snapshot sync** page handling against **LZ4 decompression bombs** from untrusted peers.
> 
> Snapshot page apply no longer uses `lz4_flex::decompress_size_prepended` (which could pre-allocate from a forged 4-byte size prefix). It now calls **`decompress_snapshot_page`**, which caps `uncompressed_len` at **`MAX_SNAPSHOT_PAGE_SIZE` (4 MB)** before allocating, requires the embedded LZ4 prefix to match the declared length, and decompresses with **`decompress_into`** into a fixed buffer so output cannot grow past the cap.
> 
> Unit tests cover valid round-trips (including pages larger than the 64 KB **grouping** hint), oversize declared lengths, forged prefixes, understated lengths, and truncated payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2067d7873916c6122ee569eb948125b1a81a9761. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Thanks for the reviews — addressed in 3aa9845b.

**🔴 Cursor Bugbot (High): oversized pages rejected, breaking sync for large entities** — Correct and the most important catch. `generate_snapshot_pages` puts a single record larger than the grouping hint on its own page, so `uncompressed_len` can legitimately exceed 64 KB. The cap was wrong: `DEFAULT_PAGE_BYTE_LIMIT` (64 KB) is only the sender's *grouping* hint, not the receiver's max. Now capped at **`MAX_SNAPSHOT_PAGE_SIZE` (4 MB)** — the same bound `SnapshotPage::is_valid` already enforces — which still defeats a GB-scale bomb but accepts valid ≤1 MB single-entity pages. Added a test that decompresses a >64 KB single-record page.

**🟡 MeroReviewer: embedded size prefix not validated against `uncompressed_len`** — Added a check that the 4-byte LE prefix equals `uncompressed_len`, rejecting inconsistent frames before allocating. (Note: `decompress_into` does not itself read the prefix — I strip it via `payload[4..]` — but cross-checking is cheap robustness and rejects malformed frames earlier.)

**💡 MeroReviewer: document the receiver-enforced cap** — Doc comment now states the cap is receiver-enforced and independent of the `byte_limit` advertised in the stream request.

**💡 AI reviewer (inline): confusing size-mismatch message** — Reworded to `"Size mismatch: declared {} bytes, decompressed {} bytes"`.

The `decompress_into` guard against expansion beyond the bounded buffer is still covered by a dedicated test.

Issue #2186, opened 2026-04-21
node/delta: throttle cascade processing to avoid long DAG write-lock holds


chefsale:
## Context

Deferred from #2178 (review comment by meroreviewer).

The cascade refactor in #2178 unified the parent-restore loop and \`try_process_pending\` under a single DAG write-lock scope in \`get_missing_parents\` (\`crates/node/src/delta_store.rs:1044-1160\`). That was necessary to close a TOCTOU gap between the \`pending_before\` snapshot and the post-cascade diff.

The side effect: the write lock is now held through:

- All \`ParentPlan::Add\` entries (each runs \`add_delta\` → WASM \`__calimero_sync_next\` execution).
- Any cascaded children promoted by \`add_delta\`'s internal \`apply_pending\`.
- All cascaded children found by the explicit \`try_process_pending\` call after the loop (each also runs WASM).
- Cloning of the cascaded delta bodies.

For a context where a restored parent unblocks many pending children, this is an unbounded amount of WASM execution under one write lock. Concurrent \`add_delta_with_events\` and \`get_missing_parents\` calls for the *same* \`DeltaStore\` (i.e. same context) serialize behind us.

## Threat / impact model

- Per-\`DeltaStore\` scope: the cost is bounded to a single context. Cross-context sync stays unaffected.
- Requires legitimate gossipsub participation: an attacker has to be a member of the context to push the pending deltas that make the cascade long.
- The pre-refactor design had the same total work; it just released/re-acquired the lock between \`add_delta\` calls. The *total* time a context was blocked was similar; this refactor just concentrates it into one contiguous critical section.
- Not a new attack surface. But it amplifies the worst-case tail latency for any concurrent sync that touches the same context.

## Fix shape

A few options, in order of intrusiveness:

1. **Cap \`try_process_pending\` per invocation**: add a \`max_deltas\` argument to \`DAG::try_process_pending\` (in \`calimero-dag\`). Caller invokes in a loop, releasing/re-acquiring the write lock between batches. Trades lock hold time for multiple lock acquisitions.

2. **Split the write-lock scope deliberately**: release the lock after each \`ParentPlan::Add\` (or every N of them), re-acquire for the next. Would reintroduce the TOCTOU gap that #2178 closed, so this needs its own correctness story — probably a per-DeltaStore \"cascade generation\" counter that pending_before/pending_after both carry.

[...]

Issue #2025, opened 2026-02-17
perf(sync): Batch delta application to reduce lock overhead


xilosada:
## Problem

During delta catchup, each delta acquires and releases the lock separately:

```rust
// Current: crates/node/src/delta_store.rs
for delta in received_deltas {
    // Each call acquires lock internally
    delta_store.add_delta(delta).await?;
}
```

**Impact**:
- N deltas = N lock acquisitions
- Lock overhead dominates for small deltas
- Increased contention during catchup

## Proposed Solution

1. Add batch API to DeltaStore:

```rust
impl DeltaStore {
    /// Apply multiple deltas in a single lock acquisition
    pub async fn add_deltas_batch(
        &self, 
        deltas: Vec<CausalDelta<Vec<Action>>>
    ) -> Result<BatchAddResult> {
        let mut dag = self.dag.write().await;  // Single lock
        
        let mut results = Vec::with_capacity(deltas.len());
        for delta in deltas {
            let result = dag.add_delta(delta, &*self.applier).await;
            results.push(result);
        }
        
        drop(dag);  // Single unlock
        
        // Update DB heads once at end
        self.update_heads_batch(&results).await?;
        
        Ok(BatchAddResult { results })
    }
}
```

2. Use in sync handler for bulk delta reception

## Acceptance Criteria

- [ ] `add_deltas_batch` method implemented
- [ ] Single lock acquisition for N deltas
- [ ] DB heads updated once after batch
- [ ] Sync protocols use batch API
- [ ] Benchmark shows reduced lock time

## Implementation Notes

- Keep individual `add_delta` for gossipsub (single delta at a time)
- Batch size limit (e.g., 100) to prevent holding lock too long
- Consider transaction batching at RocksDB level too

**Priority**: P2 (Medium Impact, Low Effort)

chefsale:
### Status update (re-triaged against current `master`)

Re-checked this against the code as of today. **Still valid, but the framing needs updating** — the problem is real and unaddressed, yet the proposed solution is partly stale.

#### The described problem still exists
- No batch API was ever added — there is still no `add_deltas_batch`.
- The per-delta lock pattern is present in every catchup path:
  - `crates/node/src/sync/delta_request.rs:414` — DAG catchup fetches each missing delta and calls `add_delta(...)` one at a time.
  - `crates/node/src/sync/manager/mod.rs:1858`, `:2121` — `add_delta(...)` inside loops.
[...]

PR #2623, opened 2026-06-02
perf(node/sync): batch delta application during catchup (#2025)

chefsale:
## What

Implements #2025 — a bounded batch-delta API that collapses the per-delta lock + persist overhead in the DAG-catchup path.

`DeltaStore::add_deltas_batch` is a batch counterpart to `add_delta`. Versus calling `add_delta` in a loop it:
- takes the `dag` write lock **once** instead of once per delta, and
- collapses **N** per-delta atomic persists + **N** `dag_heads` writes into **one** atomic `WriteBatch`.

The catchup loop in `sync/delta_request.rs` now buffers verified deltas and flushes them in `DELTA_BATCH_MAX` (100) chunks.

## How / correctness

- **Verification unchanged.** Per-delta signature + governance/membership verification stays in the caller (it runs *before* registration). The batch path only registers and persists — exactly like `add_delta`.
- **Ordering & cascades preserved.** Inputs register in order; one that arrives before its parent goes pending and is unblocked by a later input via the DAG's own `apply_pending`. The final applied set is read back from the *settled* DAG (`is_applied`) rather than inferred from per-call return values, which are stale under within-batch cascades.
- **Envelope fidelity.** Applied inputs are routed as envelope-preserving *primaries* (author / governance-position / signature retained) so this node can serve verifiable catchup for them. Only *pre-existing* pending deltas unblocked by the batch take the existing cascade path that nulls those fields — the same lossy-but-accepted behaviour the single path already has for cascaded children.
- **Atomicity.** The persist helper is widened from a single optional `primary` to a `Vec` of primaries; applied inputs + unblocked pre-existing pendings + new heads commit as one atomic batch. A failed commit carrying applied inputs is fatal (caller must not treat them as durable); cascade-only failures stay best-effort (next sync corrects the heads).
- **Lock-window bound.** `DELTA_BATCH_MAX = 100` caps the worst-case write-lock hold; instrumented via the existing `record_dag_write_lock_hold` (#2186) so contention can be measured from real data.

The single-delta gossip path (`handle_state_delta`) is untouched. `manager/mod.rs` (genesis carve-out + head-pull) is intentionally left on the single `add_delta` path for now (interleaved with per-head admission counting) — a follow-up.

## Testing

- New unit tests (`delta_store_batch_test.rs`): empty-noop, all-pending classification, and **pending-path equivalence** to a loop of `add_delta`.
- Regression: `dag_storage_integration` (14/14) + `sync_protocols` (12/12) green.
- `cargo fmt --check` + `cargo clippy -p calimero-node` clean.

The *applied/cascade* path runs real WASM via `ContextStorageApplier` (needs an installed app), so it isn't unit-testable in-crate — it's structurally identical to the proven single-delta path (same applier loop, same record shape, same atomic helper) and is exercised by the e2e sync-catchup suites.

---

> [!NOTE]
> **Medium Risk**
> Touches DAG persistence, lock hold time, and atomic commit semantics on the sync catchup path; behavior is intended to match the single-delta path but batching increases blast radius if classification or persist grouping is wrong.
> 
> **Overview**
> Adds **bounded batch registration** for pre-verified DAG deltas during sync catchup, replacing a per-delta loop that repeatedly took the DAG write lock and persisted each delta separately.
> 
> **`DeltaStore`** gains `DELTA_BATCH_MAX` (100), `BatchDeltaInput` / `BatchAddResult`, and `add_deltas_batch`: one write-lock scope, in-order registration with within-batch cascades, per-delta apply failures logged as `failed` (not whole-batch abort), and a **single atomic** persist of applied inputs (envelope-preserving primaries), unblocked pre-existing pendings, and updated `dag_heads`. `persist_cascaded_deltas_and_update_heads` now accepts a **`Vec` of primaries** instead of one optional primary.
> 
> **`request_missing_deltas`** buffers verified fetches and flushes via `flush_delta_batch` at chunk boundaries (and before hitting the fetch limit). Gossip / single-delta paths are unchanged.
> 
> New **`delta_store_batch_test`** covers empty batches, all-pending classification, and pending-path equivalence vs repeated `add_delta`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3bc999da10a8f23b300b05a421b10155e8226ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
## Review findings addressed in 7c696050

Thanks for the thorough pass. Triaged and resolved as follows.

### Fixed in code

- **Mid-batch apply error dropping already-applied deltas** (cursor 🟠, meroreviewer 🔴-adjacent): the loop no longer `?`-propagates. A failing `dag.add_delta` is logged and skipped (re-fetched next sync), so deltas that already applied in the same chunk still flow through the atomic persist below — restoring the single-delta path's "each success is durable" guarantee. This was the most important finding.
- **`cascaded_other_ids` double-counting a duplicate input** (github-actions 🟡, meroreviewer): now filtered to exclude batch-input ids, so a delta that is both pre-existing-pending and a re-fetched duplicate is persisted exactly once — as an envelope-preserving primary, never also via the envelope-nulling `applied_bodies` path.
- **"fatal" comment vs. caller's warn-and-continue** (meroreviewer 🟡): comment reworded to match reality — the bail returns `Err`; `flush_delta_batch` deliberately warns and relies on the next sync, identical to the pre-PR per-delta path's error handling.

### Acknowledged trade-offs / pre-existing patterns (no change)

- **DAG write lock held across WASM applies for the batch** (the 🔴 + 🟡s): real in the abstract, but the *catchup caller fetches reverse-BFS (child-first)*, so inputs overwhelmingly go **pending** (no WASM) and the applies concentrate in the cascade — which already runs under a single lock on master today. The "100 sequential independent WASM applies under one lock" worst case effectively cannot arise from this caller; the `DELTA_BATCH_MAX=100` cap plus the existing `record_dag_write_lock_hold` instrumentation bound and surface it. Releasing the lock between applies would reintroduce exactly the per-delta churn this PR removes.
- **`head_root_hashes` populated then pruned to heads / pruned with heads captured inside the lock** (meroreviewer 🟡 ×2): this is byte-for-byte the same pattern `add_delta_internal` already uses (pre-insert at start, `retain` to `get_heads()` after `drop(dag)`). Moving the prune inside the DAG lock would create a new `dag.write()`+`head_root_hashes.write()` lock-ordering not present anywhere else. `head_root_hashes` is a best-effort merge-detection hint corrected by the next sync; keeping parity with the single path is the safer choice.
- **Phase 0 pre-persist: N non-atomic puts / "handle never commits"** (github-actions 🟡, meroreviewer 🟡): `Handle::put` writes through to the store layer immediately (`crates/store/src/handle.rs:86`) — same as the single path, which has no commit call either. Phase 0 only runs for `events: Some`, which **no current caller passes** (catchup uses `events: None`); the records are recoverable `applied:false` rows, not corruption. Left as-is to keep the public API's events support.
- **Applied/cascade path lacks in-crate unit coverage** (github-actions 💡): already disclosed in the PR description and the test file. `ContextStorageApplier::apply` runs real WASM, so the applied path isn't unit-testable in-crate (no node unit test stands up an app); it's structurally identical to the proven single-delta path and is exercised by the e2e sync-catchup suites.

chefsale:
## Follow-up in a3bc999d

- **Failed deltas misclassified as `pending`** (cursor, new): after the skip-on-apply-error change, an errored delta fell into `BatchAddResult::pending` and emitted a spurious `pending` metric. Now tracked in a distinct `failed` bucket with its own `failed` outcome metric; they're re-fetched next sync. Good catch — this was introduced by the previous fix.
- Also folded in: clone the delta bodies **before** taking the write lock (per the earlier clone-inside-lock note) and merged `master`.

The remaining re-flagged threads (WASM-under-lock, `head_root_hashes` pre-insert/prune window, the `events: Some` Phase-0 pre-persist atomicity) are the same acknowledged trade-offs / pre-existing patterns covered in my earlier summary — they match `add_delta_internal`'s existing behaviour or apply only to the unused `events`-carrying path. No code change there; flagging for human reviewer judgment on the lock trade-off specifically.

Issue #2423, opened 2026-05-20
`#[app::private]` containers leak tree-backed collections into the synced merkle tree


chefsale:
## Summary

The `#[app::private]` macro annotation only privatises the outer container's borsh-serialised bytes (stored via `private_storage_write`). Tree-backed Calimero collections (`UnorderedMap`, `UnorderedSet`, `Vector`, etc.) used *inside* a `#[app::private]` struct still create entities in the main synced merkle tree. Result: writes intended to be node-local leak into sync, producing asymmetric merkle hashes — the writing node holds entities the receiving nodes don't.

## Symptom

`scaffolding-e2e` consistently fails on "Wait for sync after secret add" on master after #2421 landed (which closed the unrelated `(created_at, id)` sort divergence). The failure is deterministic across runs, not a flake.

Final state from [run 26167935542](https://github.com/calimero-network/core/actions/runs/26167935542/job/76979600513):

```
context=HKkBhkiccBf97eg5GWSwqjSyE17MBoCnySuE5rAaAx6e:
  e2e-node-1: ASWDPUVuzL4ZVe7jMDroycsWryB6w6HqLTV94RW3tioy   ← the writer
  e2e-node-2: FFpPYxd9T2JLbreEQwz7bNaeCkuJoT3UTws4xRwtxYeM
  e2e-node-3: FFpPYxd9T2JLbreEQwz7bNaeCkuJoT3UTws4xRwtxYeM
```

Two of three nodes converge; the writing node is the outlier.

## Log evidence

From node-1 (the writer) during the stuck sync window:

```
WARN HashComparison sync did not match remote handshake root (#2407).
     context_id=HKkBhk...  local_hash=548f0cb2  remote_hash=b716fa11
     nodes_compared=46  entities_merged=40  entities_pushed=0  nodes_skipped=5
```

Decoded:
- DFS walked 46 nodes
- 40 leaves matched local content exactly (so #2414's `local.hash != remote.hash` bidirectional push didn't fire)
- 5 internal subtrees were equal-and-skipped
- 0 entities pushed back

Yet root hashes still differ. The divergence is in entity *existence*, not entity *content*: node-1 holds entities (the secret-store internals) that nodes 2+3 don't, so node-1's tree includes children that the receivers' trees don't.

## The mechanism

`apps/scaffolding-e2e/src/lib.rs:90`:

```rust
#[app::private]
pub struct PrivateSecrets {
    secrets: UnorderedMap<String, String>,
}
```

[...]

PR #2628, opened 2026-06-02
feat(sdk/macros): compile-error when a multi-writer/sync-semantic collection is used inside #[app::private]

### LINKED ISSUE #2428: feat(sdk/macros): compile-error when a multi-writer/sync-semantic Calimero collection is used inside #[app::private]
## Context

#2425 narrows `#[app::private]` auto-substitution to the storage-primitive collections — `UnorderedMap`, `UnorderedSet`, `Vector` — and deliberately drops the multi-writer / sync-semantic types: `AuthoredMap`, `AuthoredVector`, `Counter`, `GCounter`, `PNCounter`, `ReplicatedGrowableArray`, `SharedStorage`, `UserStorage`, `FrozenStorage`, `LwwRegister`.

Those types still *parse* inside an `#[app::private]` struct today — the macro just leaves them untouched. That means a user who writes:

```rust
#[app::private]
struct Secrets {
    history: AuthoredVector<String>,  // ← falls back to MainStorage by default
}
```

…silently re-introduces the #2423 leak (entries land in the synced Merkle tree) — but the user's `#[app::private]` annotation implies they expected privacy.

## Why a compile-error is the right diagnostic

The misuse is *semantic*, not mechanical: these types model concepts (multi-writer convergence, per-entry authorship, permissioning) that have no meaning in a single-writer namespace. The right answer isn't "make them work in private" — it's "tell the user this combination doesn't make sense, and point at the type they probably meant."

## What this issue asks for

`crates/sdk/macros/src/private.rs` should, while walking each field's type, additionally check for a *known-Calimero-but-deliberately-unsupported* collection name appearing inside `#[app::private]` and emit a `compile_error!` (or push to the `Errors` collector already used in this module). Suggested message shape:

```
error: `AuthoredVector` cannot be used inside `#[app::private]`.

    AuthoredVector tracks per-entry authorship for multi-writer
    conflict resolution. A node-local private namespace has only one
    writer, so authorship is meaningless. Use `Vector` instead — it's
    the storage-primitive that earns its place in `#[app::private]`.
```

Per-type explanations, with the "what to use instead" suggestion. The Authored*/Counter*/RGA/Shared|User|Frozen/LwwRegister family each has its own one-liner.

## Sketch

A second const list paired with explanation strings:

```rust
const PRIVATE_INCOMPATIBLE: &[(&str, &str)] = &[
    (\"AuthoredMap\",       \"tracks per-entry authorship for multi-writer convergence; use UnorderedMap instead.\"),
    (\"AuthoredVector\",    \"tracks per-entry authorship for multi-writer convergence; use Vector instead.\"),
    (\"Counter\",           \"CRDT counter for concurrent increments across nodes; use a plain integer field.\"),
    (\"GCounter\",          \"CRDT counter for concurrent increments across nodes; use a plain integer field.\"),
    (\"PNCounter\",         \"CRDT counter for concurrent increments across nodes; use a plain integer field.\"),
    (\"ReplicatedGrowableArray\", \"collaborative-edit text CRDT; use Vector<u8> or a String field.\"),
    (\"SharedStorage\",     \"models single-signature shared/causal reconciliation; pointless with one writer.\"),
    (\"UserStorage\",       \"models per-user scoping at the sync layer; pointless in a node-local namespace.\"),
    (\"FrozenStorage\",     \"models cross-node immutability; redundant in single-node storage.\"),
    (\"LwwRegister\",       \"last-write-wins register for concurrent writes; just use the value type directly.\"),
];
```

`inject_private_storage` (or a sibling `check_private_compatible`) consults this list when walking each field's type and pushes a `compile_error!` to the macro's `Errors` accumulator. Add a trybuild test under `crates/sdk/tests/macros/` confirming the error messages.

## Refs

- #2425 — narrows the supported list; this issue covers the missing-error case.
- #2423 — original leak that #2425 closes for the supported types.

chefsale:
Closes #2428.

## Problem

#2425 narrowed `#[app::private]` auto-substitution to the storage primitives (`UnorderedMap`/`SortedMap`/`UnorderedSet`/`SortedSet`/`Vector`) and deliberately dropped the multi-writer / sync-semantic family. But those types still *parsed* inside an `#[app::private]` struct — the macro just left them untouched, so their `::new()` stayed pinned to `MainStorage` and their entries silently leaked into the synced Merkle tree (the #2423 leak), even though the `#[app::private]` annotation implies the user expected privacy.

The misuse is *semantic*, not mechanical: `AuthoredVector`, `Counter`, `SharedStorage`, … model per-entry authorship, cross-node merge, per-user scoping, and LWW reconciliation — none of which have meaning in a single-writer node-local namespace. The right answer is a targeted compile error that points at the primitive the user most likely meant.

## What this does

- **`PRIVATE_INCOMPATIBLE`** — a `(name, explanation)` list covering the excluded family, each pairing the type with a one-liner on *why* it's wrong and *what* to use instead.
- **`collect_private_incompatible`** — a read-only twin of `inject_private_storage` that recurses through the same wrappers (generics, tuples, refs, arrays, slices, groups, parens), so offenders nested in `Option<…>`/`Vec<…>`/tuples are caught too. It collects the offending *ident* so the `compile_error!` is spanned right at the culprit type.
- **`check_private_compatible`** — walks every struct/enum/variant field and pushes a spanned error per offender into the existing `Errors` accumulator, before substitution runs.
- New **`ParseError::PrivateIncompatibleCollection`** variant.

### Sample diagnostic

\`\`\`
error: (calimero)> \`AuthoredVector\` cannot be used inside \`#[app::private]\` — it tracks
per-entry authorship for multi-writer convergence; use \`Vector\` instead.

       A node-local private namespace has a single writer, so the multi-writer /
       sync-layer semantics this type models have no meaning here.
\`\`\`

## Tests

- 6 unit tests in `private.rs`: per-type detection, nested detection, multi-offender reporting, primitives-not-flagged, plus two anti-drift checks keeping `PRIVATE_INCOMPATIBLE` in sync with the existing `EXCLUDED_*` test lists and disjoint from `TREE_BACKED_TYPES`.
- New trybuild compile-fail test `error_private_incompatible.rs` (+`.stderr`) asserting the message text and spans. Uses locally-named stand-in types so only the macro diagnostics surface in the captured stderr.

`cargo test -p calimero-sdk-macros` → 23 passed. `cargo fmt --check` clean; clippy clean for the touched files.

---

> [!NOTE]
> **Low Risk**
> Compile-time-only macro validation; no runtime behavior change for valid apps, only clearer failures for previously leaky misuse.
> 
> **Overview**
> **`#[app::private]`** now **fails at compile time** when a field uses sync- or multi-writer Calimero collections (CRDTs, `SharedStorage` / `UserStorage` / `FrozenStorage`, `AuthoredMap` / `AuthoredVector`, etc.) instead of leaving them on default **`MainStorage`** and leaking into the synced tree.
> 
> The macro maintains a **`PRIVATE_INCOMPATIBLE`** list with per-type guidance, walks field types with the same recursion as **`PrivateStorage`** injection (including nested `Option` / tuples), and emits **`ParseError::PrivateIncompatibleCollection`** with spans on the offending ident. The check runs **before** tree-backed substitution.
> 
> **Unit tests** lock detection, nesting, and list consistency with existing **`EXCLUDED_*`** / **`TREE_BACKED_TYPES`** tables; a **trybuild** test asserts diagnostic text and spans.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0842c2ece55c8baf6203f186cd0c21476fa3435. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/sdk/macros/src/errors.rs
+++ b/crates/sdk/macros/src/errors.rs
@@ -113,6 +113,15 @@ pub enum ParseError<'a> {
         type_name: &'static str,
         suggestion: &'static str,
     },
+    #[error(
+        "`{type_name}` cannot be used inside `#[app::private]` — it {explanation}\n\n\
+         A node-local private namespace has a single writer, so the multi-writer / \
+         sync-layer semantics this type models have no meaning here."
+    )]
+    PrivateIncompatibleCollection {
+        type_name: String,
+        explanation: &'static str,
+    },
 }
 
 impl AsRef<Self> for ParseError<'_> {
--- a/crates/sdk/macros/src/errors.rs
+++ b/crates/sdk/macros/src/errors.rs
@@ -119,7 +119,7 @@ pub enum ParseError<'a> {
          sync-layer semantics this type models have no meaning here."
     )]
     PrivateIncompatibleCollection {
-        type_name: String,
+        type_name: &'static str,
         explanation: &'static str,
     },
 }
--- a/crates/sdk/macros/src/private.rs
+++ b/crates/sdk/macros/src/private.rs
@@ -31,9 +31,11 @@ use sha2::{Digest, Sha256};
 /// straight into the outer private blob and need no substitution.
 ///
 /// **Deliberately excluded** (using them inside `#[app::private]`
-/// is a semantic mismatch and currently produces a normal type
-/// error, since their `::new()` stays pinned to `MainStorage`; a
-/// dedicated compile-time diagnostic is tracked in #2428):
+/// is a semantic mismatch — rather than silently leaving their
+/// `::new()` pinned to `MainStorage` and letting a downstream type
+/// error surface, they're listed in `PRIVATE_INCOMPATIBLE` and
+/// rejected with a targeted compile error pointing at the primitive
+/// to use instead):
 ///
 /// - CRDT data-types (`LwwRegister`, `Counter`, `GCounter`,
 ///   `PNCounter`, `ReplicatedGrowableArray`) — CRDTs exist for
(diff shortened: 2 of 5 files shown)


chefsale:
Thanks for the review — addressed all four in `c0842c2e`:

1. **`type_name: String` → `&'static str`** (errors.rs): the variant now matches `ForbiddenStdCollection`/`ForbiddenBarePrimitive`. `collect_private_incompatible` carries the matched `&'static (name, explanation)` const entry alongside the ident (kept for spanning), so we reuse the const string instead of allocating per offender.

2. **Recursing into a matched type's generics** (`AuthoredVector<Counter>`): now returns early once a path segment matches — only the outer type the user actually got wrong is reported. Nested-in-a-*valid*-container cases (e.g. `UnorderedMap<K, AuthoredVector<..>>`) are unaffected, since the valid outer type doesn't match and we still recurse.

3. **Asymmetric anti-drift test**: `excluded_lists_match_private_incompatible` now checks set equality in **both** directions, so a name in `PRIVATE_INCOMPATIBLE` that's in neither `EXCLUDED_*` list fails even when the lengths coincide (no longer relying on length + no-dupes).

4. **`walk_fields` closure**: extracted to a free fn `collect_incompatible_in_fields` — no implicit lifetime capture, easier to read.

Unit tests (23) and the trybuild test still green; the diagnostic stderr is unchanged.

chefsale:
Second pass addressed in `6316fb3b`:

- **`to_string()` allocation**: `collect_private_incompatible` now matches via `Ident: PartialEq<str>` (`find(|(n, _)| last.ident == *n)`) — no `String` per path segment. (`try_inject_on_path` keeps its own `to_string()` for now; left it consistent with the rest of that fn, can fold it in if you'd like.)
- **Duplicate entries in `PRIVATE_INCOMPATIBLE`**: anti-drift test now asserts `PRIVATE_INCOMPATIBLE.len() == HashSet.len()`, so a dup fails loudly instead of hiding behind the set comparisons.
- **Doc-comment split**: the "read-only twin" note is now its own paragraph.

23 unit tests + trybuild still green.

PR #2648, opened 2026-06-03
fix(runtime): stop installing a process-global panic hook (#2637)

### LINKED ISSUE #2637: Process-wide panic hook is installed once and never restored (clobbers later host hooks)
## Summary

The WASM execution macro in `crates/runtime/src/logic/imports.rs:233-258` installs a **process-global** panic hook via `std::panic::set_hook` inside a thread-local `Once` (`HOOKER.call_once`). It is never restored, and because the first execution on any thread permanently replaces the global hook, the behaviour is load-order-dependent and **clobbers any panic hook the host installs afterwards**.

## Details

```rust
HOOKER.with(|hooker| {
    hooker.call_once(|| {
        let prev_hook = std::panic::take_hook();
        std::panic::set_hook(Box::new(move |info| {
            if !HOST_CTX.with(|ctx| ctx.load(Ordering::Relaxed)) {
                return prev_hook(info);
            }
            // ...capture payload into thread-local PAYLOAD...
            prev_hook(info);
        }));
    });
});
```

Issues:
- `set_hook` mutates **global** process state, but is gated by a thread-local `Once` — so the "install once" intent is per-thread while the effect is process-wide. Which thread runs first determines whose `prev_hook` gets captured.
- The hook is never restored (no `take_hook` pairing on teardown). It lives for the process lifetime.
- Any host/component that calls `set_hook` *after* the first WASM execution silently overrides this capture hook; any host that calls it *before* gets captured as `prev_hook` and chained — order-dependent, fragile behaviour.
- It exists only to capture a panic payload (message + location) into a thread-local for the guest-panic path.

## Impact

- Interferes with host process panic handling (logging, crash reporters, custom hooks) in a non-deterministic, load-order-dependent way.
- Hard-to-debug interactions when the runtime is embedded alongside other code that manages the panic hook.

## Suggested fix

Don't mutate the global panic hook. The payload is already recoverable without it: the execution path at `crates/runtime/src/lib.rs:292` catches the unwind and downcasts the payload via `panic_payload::panic_payload_to_string(...)`. Capture the message/location through `catch_unwind` downcasting at that site (extending it to also recover `Location` if needed) and delete the `set_hook` machinery entirely.

## References

- `crates/runtime/src/logic/imports.rs:233-258` (global `set_hook` in a `Once`)
- `crates/runtime/src/lib.rs:292` (existing `catch_unwind` payload extraction — the preferred mechanism)

_Found during security review (finding #30)._

chefsale:
## What

Fixes #2637. The WASM execution macro in `crates/runtime/src/logic/imports.rs` installed a **process-global** panic hook via `std::panic::set_hook`, gated by a thread-local `Once`.

The gate was per-thread but the effect was process-wide, which meant:
- install order was nondeterministic (whichever thread ran WASM first won);
- every worker thread that ran WASM stacked another nested hook;
- the hook was never restored — it lived for the process lifetime;
- it silently clobbered (or transparently chained to) whatever panic hook the host installed afterward — crash reporters, structured loggers, the `cargo test` harness.

It existed only to recover a host-function panic's message + location into a thread-local for the `catch_unwind` path.

## How

Recover the panic message **directly from the `catch_unwind` payload** via the existing `panic_payload::panic_payload_to_string` (the same mechanism `Module::run` already uses at `lib.rs`), and delete the `set_hook` / `HOOKER` / `PAYLOAD` / `HOST_CTX` machinery entirely.

Tradeoff: the unwind payload can't carry a source location, so a host-function panic's `location` now degrades to `Location::Unknown` — matching the existing guest-panic path. In exchange, the runtime no longer mutates global process state.

## Testing

Added `test_wasm_host_panic_recovered_without_global_hook`: a WAT guest calls `storage_read` backed by a `Storage` whose `get()` panics, driving a **real** host-side panic through the per-host-call `catch_unwind`. It asserts:
- `HostError::Panic { context: Host }`,
- the panic message survives (proves payload-downcast recovery works),
- `location == Unknown`.

The location assertion is a deliberate fail-on-old regression guard: I verified that temporarily re-emitting a precise `Location::At { .. }` (simulating the deleted global hook) makes the test fail exactly on that assertion.

- ✅ 93/93 runtime lib tests pass
- ✅ `cargo fmt --check` clean
- ✅ no new clippy warnings in the runtime crate

---

> [!NOTE]
> **Medium Risk**
> Touches panic handling on every WASM host import boundary; behavior change is intentional (no global hook, host panic locations become Unknown) with a targeted integration test.
> 
> **Overview**
> Removes the WASM host-import layer’s **process-global** `std::panic::set_hook` setup (thread-local `Once`, `PAYLOAD`, `HOST_CTX`) so the runtime no longer overwrites or chains against the node’s own panic hooks.
> 
> Host-function panics are still caught per call with `catch_unwind`, but the message now comes from **`panic_payload::panic_payload_to_string`** on the unwind payload—the same approach as `Module::run`—and **`HostError::Panic`** uses **`Location::Unknown`** because the payload has no source location (guest `panic` still supplies location via the guest ABI).
> 
> Adds **`test_wasm_host_panic_recovered_without_global_hook`**: a WAT guest calls `storage_read` against a `Storage` that panics in `get`, asserting **Host** context, message preservation, and **`Unknown`** location as a guard against reintroducing the global hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0b5eee64f5e727a3bd34fd91d35ba460f245995. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -978,6 +978,12 @@ mod wasm_integration_tests {
                     message.contains("storage.get panicked deliberately"),
                     "panic message must be recovered from the unwind payload, got: {message:?}"
                 );
+                // `Unknown` because the unwind payload carries no location. This
+                // guards against silently re-introducing the process-global hook
+                // (which captured a precise location). If a *non-global* location
+                // recovery mechanism is ever added on purpose, update this assertion
+                // to expect the recovered location — that is an intended improvement,
+                // not a regression.
                 assert!(
                     matches!(location, Location::Unknown),
                     "without the global hook the panic location can't be recovered; \
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -891,6 +891,103 @@ mod wasm_integration_tests {
         }
     }
 
+    /// A `Storage` whose `get` panics. Used to drive a panic from *inside* a
+    /// host function so we can exercise the runtime's per-host-call
+    /// `catch_unwind` recovery (the path that replaced the process-global
+    /// `set_hook` machinery in `logic/imports.rs`).
+    struct PanicStorage;
+
+    impl Storage for PanicStorage {
+        fn get(&self, _key: &store::Key) -> Option<store::Value> {
+            panic!("storage.get panicked deliberately");
+        }
+        fn set(&mut self, _key: store::Key, _value: store::Value) -> Option<store::Value> {
+            None
+        }
+        fn remove(&mut self, _key: &store::Key) -> Option<store::Value> {
+            None
+        }
+        fn has(&self, _key: &store::Key) -> bool {
+            false
+        }
+    }
+
+    /// A panic that originates *inside a host function* must be caught by the
+    /// runtime's per-host-call `catch_unwind` and surfaced as
+    /// `HostError::Panic { context: Host, .. }` with its message intact —
+    /// WITHOUT installing a process-global panic hook.
+    ///
+    /// The message is recovered directly from the unwind payload. Asserting
+    /// `location == Unknown` doubles as a regression guard: the deleted
+    /// `set_hook` hook captured a *precise* `Location::At { .. }`; the payload
+    /// can't carry a location, so it must now degrade to `Unknown`. A revival
+    /// of the global-hook approach would make this assertion fail.
+    #[test]
+    fn test_wasm_host_panic_recovered_without_global_hook() {
+        // The guest builds a 16-byte `{ ptr: u64, len: u64 }` buffer descriptor
+        // (the host-guest ABI; see `prepare_guest_buf_descriptor`) pointing at a
+        // 3-byte key "abc", then calls `storage_read`. The host reads the key and
+        // calls `storage.get`, which panics.
+        let wat = r#"
+            (module
+                (import "env" "storage_read" (func $storage_read (param i64 i64) (result i32)))
+                (memory (export "memory") 1)
+                (func (export "read_probe")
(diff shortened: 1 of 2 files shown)


chefsale:
Thanks for the review. Responses to the findings:

- **Inline: `Location::Unknown` assertion fragility** — fair point, addressed in e0b5eee. Added an inline note that an intentional non-global location-recovery mechanism added later should update the assertion rather than be read as a regression.

- **Suggestion: `PanicStorage` duplicated** — not the case; there's exactly one `struct PanicStorage` (lib.rs:898). `grep -c "struct PanicStorage"` returns 1, and the crate compiles with no duplicate-definition error. No change.

- **Nitpick: `panic_payload_to_string` `&'static str` vs `&str`** — the finding itself concludes "No issue here — just confirming correctness." `downcast_ref::<&str>()` matches the `&'static str` a string-literal `panic!` produces, and the `String` arm covers formatted panics. No change.

- **Warning: unsafe `*mut () -> &mut VMLogic` cast inside `catch_unwind`** — pre-existing and unchanged by this PR (as the finding notes). On a host-function panic the behavior is identical to before: the wrapper returns `Err(HostError::Panic)` and `logic.finish(err)` runs, producing an error `Outcome`. This PR only changes *how the message is recovered* (payload downcast instead of a global hook), not the post-panic `VMLogic` lifecycle. The use-after-panic soundness question is worth its own issue but is out of scope here.

PR #2649, opened 2026-06-03
feat(context): log context cache hit rate periodically (#2031)

### LINKED ISSUE #2031: feat(context): Log context cache hit rate periodically
## Problem

No visibility into context cache effectiveness.

## Proposed Solution

```rust
struct ContextCacheStats {
    hits: AtomicU64,
    misses: AtomicU64,
}

impl ContextManager {
    fn log_cache_stats(&self) {
        let hits = self.stats.hits.load(Ordering::Relaxed);
        let misses = self.stats.misses.load(Ordering::Relaxed);
        let total = hits + misses;
        let hit_rate = if total > 0 { hits as f64 / total as f64 } else { 0.0 };
        
        info!(
            hits,
            misses,
            hit_rate = format!("{:.1}%", hit_rate * 100.0),
            cache_size = self.contexts.len(),
            "Context cache statistics"
        );
    }
}

// Log every 5 minutes
ctx.run_interval(Duration::from_secs(300), |act, _| {
    act.log_cache_stats();
});
```

## Acceptance Criteria

- [ ] Cache hits/misses tracked
- [ ] Stats logged every 5 minutes
- [ ] Metric exported to Prometheus

**Estimated Effort**: 1 hour

**Priority**: P2 (Quick Win)

chefsale:
Closes #2031.

## Summary

Adds visibility into in-memory context-cache effectiveness — a follow-up to the cache cap from #2606, instrumenting the same `contexts`/`applications` hot caches.

- **`ContextCacheStats { hits, misses: AtomicU64 }`** — always-on source of truth for the periodic log line, so it works even on nodes without a Prometheus registry (`metrics: None`, e.g. headless/test).
- **Single increment chokepoint** at `get_or_fetch_context` via `record_cache_access(hit)`, reusing the existing `was_cached` flag and mirroring into Prometheus when a registry is present. Probes for non-existent contexts (the early `Ok(None)`) are deliberately **not** counted — they're not a cache-effectiveness signal.
- **`log_cache_stats()`** driven by `ctx.run_interval(300s, …)` in `Actor::started` (mirrors the existing `start_namespace_heartbeat` pattern). Idle nodes log at `debug`; active caches log `info` with hit rate (`{:.1}%`) and both cache sizes.
- **Prometheus** under `context.cache.*`: `context_hits` / `context_misses` counters plus `context_size` / `application_size` gauges. Hit *rate* is derived in PromQL (`rate(hits) / (rate(hits) + rate(misses))`) rather than exported pre-divided.

## Acceptance criteria

- [x] Cache hits/misses tracked
- [x] Stats logged every 5 minutes
- [x] Metric exported to Prometheus

## Notes

- Hit rate is **cumulative (lifetime-average)**, matching the issue's snippet. A per-interval windowed delta in the log line is a possible follow-up; windowed rates are already available from the Prometheus counters.

## Testing

- 7 unit tests pass (2 new: `cache_stats_record_and_snapshot`, `cache_stats_hit_rate_formatting`)
- `cargo fmt --check` clean; clippy adds 0 warnings over baseline
- `calimero-node` + `calimero-server` check green

---

> [!NOTE]
> **Low Risk**
> Read-only metrics and logging on the existing cache-aside path; no change to eviction, locking, or datastore behavior beyond counting accesses.
> 
> **Overview**
> Adds **observability for the in-memory context and application metadata caches** in `ContextManager`, building on the existing cache caps.
> 
> **Hit/miss tracking** uses in-process `AtomicU64` counters (`ContextCacheStats`) updated only from `get_or_fetch_context` via `record_cache_access`: hits when the context is already cached, misses after a successful datastore load. Lookups that return `None` (missing context) are **not** counted. The same events increment Prometheus `context.cache.hits` / `context.cache.misses` when a registry is configured.
> 
> **Periodic reporting** runs every 5 minutes from `Actor::started` (`log_cache_stats`): lifetime hit rate at `info` when there has been traffic, `debug` when idle, plus current context/application cache sizes. Size gauges (`context.cache.size`, `context.cache.application_size`) are refreshed on that interval.
> 
> Unit tests cover counter behavior, hit-rate formatting, and Prometheus text encoding for the new series.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bb8be08d725708a985967c09a743849d51c4417. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/lib.rs
+++ b/crates/context/src/lib.rs
@@ -6,11 +6,12 @@ use calimero_governance_store::{
 };
 use std::collections::{btree_map, BTreeMap, HashMap, HashSet};
 use std::future::Future;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
 use actix::prelude::{ActorResponse, WrapFuture};
-use actix::Actor;
+use actix::{Actor, AsyncContext};
 use calimero_context_client::client::ContextClient;
 use calimero_context_client::local_governance::SignedNamespaceOp;
 use calimero_context_config::types::ContextGroupId;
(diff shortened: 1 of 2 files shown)


chefsale:
Thanks for the review — addressed all four in `7bb8be08`:

1. **Prometheus integration untested** — added a registry-level test (`context_cache_metrics_register_and_encode`) that constructs `Metrics` with a real `Registry`, exercises the same `inc()`/`set()` calls the manager makes, encodes via the text encoder, and asserts the `context_cache_*` counter/gauge values round-trip. I kept it at the `Metrics`/registry layer rather than spinning up a full `ContextManager` (which needs a `Store` + node/context clients) — the wiring in `record_cache_access`/`log_cache_stats` is a thin pass-through to these same primitives.

2. **Gauge staleness (5-min resolution)** — documented in the metric help text that the size gauges are refreshed by the periodic task (~5-min) and may lag faster scrape intervals, with a pointer to the hits/misses rates as the fine-grained signal. I deliberately did not update the gauges on every evict/insert: those are hot paths and the cap is a slow-moving safety valve, so per-event updates aren't worth the overhead for this quick-win.

3. **Discarded `run_interval` SpawnHandle** — added a comment noting it's intentionally dropped (interval runs for the actor's full lifetime, never cancelled), matching `start_namespace_heartbeat`, which also doesn't retain its handle.

4. **Unnecessary `let _ =` on `fetch_add`** — removed (also from the `inc()` call); confirmed neither return is `#[must_use]`, build stays warning-free.

Also renamed the series to drop a redundant prefix (`context_cache_hits_total` instead of `context_cache_context_hits`).

PR #2650, opened 2026-06-03
fix(storage): cascade re-key registration into deeply-nested custom structs (#2581)

### LINKED ISSUE #2581: Deep custom-struct nesting not auto-re-keyed (one-level scope of #2577 fix)
Follow-up to #2577 / #2579.

## Summary

The deterministic re-key fix that makes a custom struct CRDT value converge (#2579) registers re-key thunks for **one level** of custom-struct nesting — the value types of the **root** `#[app::state]` struct's collection fields. A custom struct reachable only *through another custom struct's collection* is not registered, so its nested collections keep per-replica-random ids and concurrent writes are silently last-writer-wins'd (the exact #2577 data loss).

## Example

```rust
#[derive(Mergeable)]
struct Inner { score: Counter }

#[derive(Mergeable)]
struct Outer { inner: UnorderedMap<String, Inner> }   // custom struct holding a collection of custom structs

#[app::state]
struct App { groups: UnorderedMap<String, Outer> }
```

- `App`'s generated `__calimero_register_rekey()` scans `App`'s fields → registers `Outer` (and `String`). ✅
- It does **not** register `Inner` (it's a value type of `Outer`'s map, not of `App`'s field). ❌
- So two replicas concurrently writing the same `Inner.score` lose data.

**Not affected:** any depth of *built-in* collection nesting (e.g. `Map<_, Map<_, Counter>>`). Built-in collections self-register in their constructors, so they always get deterministic ids. Only **custom-struct-inside-custom-struct** is the gap. Today's example apps don't hit it.

## Where

`generate_rekey_register_method` in `crates/sdk/macros/src/state.rs` collects type tokens from the root struct's fields only (one level). The generated `RekeyTarget` impls recurse correctly at *runtime* via `rekey_field_if_supported!`, but a nested custom value type is never **registered**, so `rekey_nested_value` can't dispatch to it.

## Possible approaches

1. **Cascading registration:** have `#[derive(Mergeable)]` also emit a `__calimero_register_rekey()` for the struct that registers *its* collection-field value types, and have the root's registration recurse into them. (Needs a way to call it on nested types — a trait method rather than an inherent one.)
2. **Link-time registration** (e.g. `inventory`/`linkme`) so every `#[derive(Mergeable)]` self-registers at startup — but wasm support is uncertain.
3. **Compile-time validation** (ties into #1881 `CrdtType::Record`): reject / warn on a synced struct whose value graph isn't fully registerable, turning silent runtime loss into a build-time signal.

## Severity

Low frequency (rare pattern, not in current apps) but **silent data loss** when hit, so worth tracking rather than leaving as a code comment. The convergence harness (#2578) can express the failing case as a `.invariant(...)` test once a fix is attempted.

chefsale:
Closes #2581. Follow-up to #2577 / #2579.

## Problem

The deterministic re-key fix (#2579) registered re-key thunks for only **one level** of custom-struct nesting — the value types of the **root** `#[app::state]` struct's collection fields. A custom struct reachable only *through another custom struct's collection* is never registered, so `rekey_nested_value`'s `TypeId` dispatch silently no-ops and its nested collections keep per-replica-random ids → concurrent writes are silently last-writer-wins'd (the exact #2577 data loss, one level deeper).

```rust
#[derive(Mergeable)] struct Inner { score: Counter }
#[derive(Mergeable)] struct Outer { inner: UnorderedMap<String, Inner> }
#[app::state]        struct App   { groups: UnorderedMap<String, Outer> }
```

`App`'s scan registers `Outer` and `String` but **never `Inner`** (it appears only in `Outer`'s definition, not in any `App` field). So two replicas concurrently writing the same `Inner.score` lose data.

Not affected: any depth of *built-in* collection nesting (they self-register in their constructors). Only **custom-struct-inside-custom-struct** is the gap, and no current example app hits it.

## Fix (issue's approach #1: cascading registration)

Runtime re-keying (`rekey_field_if_supported!`) already recurses to any depth; only **registration** was one-level. So each `#[derive(Mergeable)]` struct now registers its own collection-field value types, and registration cascades through the value graph.

- **`crates/storage/src/collections/rekey.rs`**
  - New trait hook `RekeyTarget::register_nested_value_types()` (default no-op).
  - New `register_rekey_cascade<T>()`: registers `T`'s thunk and, **only on its first registration**, recurses into the value types `T` nests. The first-registration guard makes self-/mutually-recursive value graphs (e.g. `Tree { children: Map<_, Tree> }`) terminate.
  - `register_rekey` now reports whether *this* call inserted the thunk, decided atomically via the `Entry` API under the write lock (so a startup race can't double-walk).
  - `register_rekey_if_supported!` routes its real arm through the cascade.
- **`crates/sdk/macros/src/{state,mergeable}.rs`**
  - `#[derive(Mergeable)]` generates `register_nested_value_types()` scanning its own fields.
  - Extracted `rekey_register_calls` / made `collect_type_paths` shared, so the derive and the root `#[app::state]` scan walk types identically.

## Tests

`crates/storage/tests/rekey_record.rs` (gated `--features testing`) gains two cases on the existing two-replica `drive::<T>()` harness:
- `cascade_registers_deeply_nested_custom_value` — derive path; registers **only** the root tokens (`Outer`, `String`), and the deeply-nested counter still converges to `2`, proving the cascade reaches `Inner`.
- `no_cascade_loses_deeply_nested_data_pre_fix` — hand-written impls **without** the `register_nested_value_types` override (the pre-fix shape); the deeply-nested counter is LWW'd to `1`, documenting the bug and proving the cascade is load-bearing.

All green locally: rekey_record 4/4, storage lib 550, sdk-macros 55, `team-metrics-macro` converge 2/2, `cargo fmt --check` clean.

---

> [!NOTE]
> **Medium Risk**
> Touches sync/convergence registration for nested CRDT values; behavior is guarded by new integration tests and first-registration recursion limits, but incorrect registration could still cause silent data loss in edge layouts.
> 
> **Overview**
> Extends **deterministic re-key registration** so custom structs nested inside other custom struct map values are registered, not only types named on the root `#[app::state]` fields. Without this, deeply nested counters could still diverge across replicas (same class of bug as #2577).
> 
> **Runtime:** `RekeyTarget` gains `register_nested_value_types()`; `register_rekey_cascade` registers a type’s thunk and, on first insert only, walks nested value types. `register_rekey` returns whether it inserted and uses an atomic `Entry` insert so concurrent startup cannot double-cascade. `register_rekey_if_supported!` now calls the cascade path.
> 
> **Macros:** `#[derive(Mergeable)]` emits `register_nested_value_types` using the same field type scan as root state; `rekey_register_calls` / `collect_type_paths` are shared with `#[app::state]`.
> 
> **Tests:** Two-replica deep nesting (`Map → Outer → Map → Inner`) proves cascade reaches `Inner` when only root tokens are registered, and that missing the override still LWWs to `1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88583348275a90de26f783276092022bbfaabedb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/sdk/macros/src/mergeable.rs
+++ b/crates/sdk/macros/src/mergeable.rs
@@ -57,6 +57,17 @@ pub fn derive(input: DeriveInput) -> TokenStream {
         _ => quote! {},
     };
 
+    // Registration of THIS struct's nested value types, scanned exactly like the
+    // root `#[app::state]` does its own fields. The root only ever names the type
+    // tokens in its OWN fields, so a custom struct reachable only through another
+    // custom struct's collection is never registered there; emitting the same
+    // scan per Mergeable struct cascades registration through the whole value
+    // graph (see `RekeyTarget::register_nested_value_types`).
+    let rekey_register_body: TokenStream = match &input.data {
+        Data::Struct(s) => crate::state::rekey_register_calls(&s.fields),
+        _ => quote! {},
+    };
+
     quote! {
         impl #impl_generics ::calimero_storage::collections::Mergeable
             for #ident #ty_generics #where_clause
@@ -90,6 +101,16 @@ pub fn derive(input: DeriveInput) -> TokenStream {
                 parent_id: ::calimero_storage::address::Id,
             ) {
                 #rekey_body
+            }
+
+            // Cascade registration into the value types this struct nests, so a
+            // custom struct reachable only through this one's collections is
+            // registered too (not left to be last-writer-wins'd). Same per-field
+            // scan the root state runs; `register_rekey_if_supported!` recurses
+            // and the first-registration guard makes self-referential value
+            // graphs terminate.
+            fn register_nested_value_types() {
+                #rekey_register_body
             }
         }
     }
--- a/crates/sdk/macros/src/state.rs
+++ b/crates/sdk/macros/src/state.rs
@@ -677,17 +677,39 @@ fn generate_rekey_register_method(
         StructOrEnumItem::Enum(_) => return quote! {},
     };
 
+    let calls = rekey_register_calls(fields);
+
+    quote! {
+        impl #impl_generics #ident #ty_generics #where_clause {
+            /// Registers nested CRDT-value re-key thunks. Called at WASM module
+            /// load and by the native test bridge; idempotent. Not for direct use.
+            #[doc(hidden)]
+            pub fn __calimero_register_rekey() {
+                #calls
+            }
+        }
+    }
+}
+
+/// Emit one `register_rekey_if_supported!(T)` per distinct type token reachable
+/// from `fields` (each field type plus its generic args, recursively). Shared by
+/// the root `#[app::state]` scan and `#[derive(Mergeable)]`'s
+/// `register_nested_value_types` override, so a nested custom struct registers
+/// its own collection-value types the same way the root does — turning the
+/// root's one-level scan into a full walk of the reachable value graph.
+///
+/// Dedup is by token string (syntactic, not semantic): the same value type can
+/// appear in several fields (e.g. two `UnorderedMap<String, TeamStats>`), so
+/// collapse identically-written types to one call. Types written differently but
+/// equal (`TeamStats` vs `crate::TeamStats`) still emit two calls — harmless:
+/// they share a `TypeId`, and registration is idempotent (`or_insert`), so the
+/// registry holds one entry either way.
+pub(crate) fn rekey_register_calls(fields: &syn::Fields) -> TokenStream {
     let mut types: Vec<Type> = Vec::new();
     for field in fields.iter() {
         collect_type_paths(&field.ty, &mut types);
     }
 
-    // Dedup by token string (syntactic, not semantic): the same value type can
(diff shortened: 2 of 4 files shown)


chefsale:
Addressed the reviewer doc feedback in 88583348 (doc-only):
- `register_rekey`: comment now states the cascade fires **exactly once per type** (the `Entry`-under-write-lock decision is atomic — one thread observes `Vacant`/returns true, the other `Occupied`/false), rather than the looser "can't double-walk".
- `register_nested_value_types`: documented that the `Self: Sized` bound is deliberate — it's a type-level op called only as `register_rekey_cascade::<T>()` on a concrete type, never through `dyn RekeyTarget`, so cascading through a trait object fails to compile instead of silently no-op'ing. `rekey_relative_to` remains the object-safe, thunk-dispatched method.

Skipped the `_static` rename suggestion — the doc note covers it without churn. No logic change; all three findings were 'correct, document more precisely'.

Issue #2027, opened 2026-02-17
feat(api): Stream sync progress via JSON-RPC for better UX


xilosada:
## Problem

During snapshot sync, context returns `Uninitialized` error with no progress info:

```rust
// crates/context/src/handlers/execute.rs:96-97
if !is_state_op && *context.meta.root_hash == [0; 32] {
    return ActorResponse::reply(Err(ExecuteError::Uninitialized));
}
```

**Impact**:
- Users see opaque "Uninitialized" error
- No visibility into sync progress
- Can't differentiate "syncing" from "broken"

## Proposed Solution

1. Track sync progress in NodeState:

```rust
pub struct SyncProgress {
    pub state: SyncProgressState,
    pub started_at: Instant,
    pub bytes_received: u64,
    pub bytes_total: Option<u64>,
    pub deltas_received: u64,
    pub deltas_pending: u64,
}

pub enum SyncProgressState {
    Idle,
    Connecting,
    ReceivingSnapshot { percent: u8 },
    ApplyingDeltas { applied: u64, total: u64 },
    Finalizing,
}
```

2. Add JSON-RPC endpoint:

```rust
// GET /rpc/context/{context_id}/sync_status
#[derive(Serialize)]
pub struct SyncStatusResponse {
    pub context_id: ContextId,
    pub is_initialized: bool,
    pub sync_state: String,
    pub progress: Option<SyncProgressInfo>,
}
```

3. WebSocket subscription for live updates

## Acceptance Criteria

- [ ] `SyncProgress` tracked in NodeState
- [ ] `/rpc/context/{id}/sync_status` endpoint returns progress
- [ ] WebSocket subscription for live updates
- [ ] Progress updates during snapshot sync
- [ ] ETA calculation based on transfer rate

**Priority**: P3 (Low Impact, Medium Effort)

chefsale:
Status check (as of 2026-05-30, branch state at core@4a36752) — **still relevant, feature unimplemented.**

**Problem code still present.** It moved from `crates/context/src/handlers/execute.rs` to `crates/context/src/handlers/execute/mod.rs:117-119`, but is otherwise unchanged:

```rust
if !is_state_op && *context.meta.root_hash == [0; 32] {
    return ActorResponse::reply(Err(ExecuteError::Uninitialized));
}
```

[...]

PR #2652, opened 2026-06-03
feat(api): expose per-context sync status via JSON-RPC (#2027)

chefsale:
## What

Adds a `sync_status` JSON-RPC method so a client blocked on `Uninitialized` can tell **syncing** vs **waiting-for-peers** vs **stuck**, instead of guessing from one opaque error.

Closes the query-endpoint half of #2027 (Phase 1). WebSocket streaming + fine-grained page percent/ETA are deliberate follow-ups (see below).

## Why

A node that has joined a context but not yet adopted its initial state reports `root_hash == [0; 32]` and rejects execution with `ExecuteError::Uninitialized`:

```rust
// crates/context/src/handlers/execute/mod.rs
if !is_state_op && *context.meta.root_hash == [0; 32] {
    return ActorResponse::reply(Err(ExecuteError::Uninitialized));
}
```

That single error can't distinguish:

| Situation | Today | After |
|---|---|---|
| Snapshot streaming in | `Uninitialized` | `syncing` |
| Joined, no peer yet | `Uninitialized` | `backingOff` + `lastError: "No peers to sync with"` |
| Failed, retry pending | `Uninitialized` | `backingOff` + `retryInSecs` + `failureCount` |
| Initialized | works | `isInitialized: true, idle` |

The sync run-loop already tracks the answer (per-context `SyncState`: in-flight, `failure_count`, `last_error`, backoff) — but it lives on the run-loop's stack and the server layer can't reach it. This PR publishes a best-effort, lock-free snapshot of it and surfaces it over JSON-RPC.

## How

```
SessionTracker (run-loop)  --publishes-->  NodeState.sync_status  (Arc<DashMap>, lock-free)
                                                  │
JSON-RPC sync_status  →  NodeClient.sync_status() ┘  (round-trips via NodeManager actor)
                      →  merges is_initialized (root_hash) from ctx_client
```

- **node-primitives**: `SyncStatusSnapshot` / `SyncPhase` types.
- **NodeState**: a `sync_status` `DashMap` + run-loop publisher handle + out-of-band reader.
- **SessionTracker**: publish a derived snapshot after every `SyncState` transition (dispatch → `syncing`, success → `idle`, failure/wedge → `backingOff`). The two benign-failure early returns in `apply_result` are folded into an if/else chain so the snapshot also publishes on those paths.
- **NodeClient.sync_status** round-trips through `NodeManager` (the snapshot lives on node-crate-private `NodeState` the server can't name) via a new `NodeMessage::GetSyncStatus`, mirroring the existing `GetBlobBytes` oneshot pattern.
- **server**: `SyncStatusRequest`/`SyncStatusResponse` in server-primitives with a coarse, self-describing `sync_state` string; thread `node_client` into the JSON-RPC `ServiceState`; handler merges `is_initialized` from the context root hash.

### Example response

```json
{ "contextId": "Cx7…", "isInitialized": false, "syncState": "backingOff",
  "retryInSecs": 8, "failureCount": 1, "lastError": "No peers to sync with" }
```

## Scope / notes

- **Coarse phase only** (`idle`/`syncing`/`backingOff`). Snapshot-page percent, ETA, and the **WebSocket live subscription** from the issue's acceptance criteria are follow-ups that layer onto this snapshot **without changing the wire shape**.
- **Method-based, not the literal REST path** `GET /rpc/context/{id}/sync_status` from the issue — the server has no REST surface, everything is JSON-RPC. Implemented as a `sync_status` method on the existing `/jsonrpc` endpoint.
- The snapshot is **advisory** (published from the run-loop, read out-of-band, may be ~ms stale) — meant for UX, not control flow.

## Testing

- `cargo fmt --check` (workspace) — clean
- `cargo test -p calimero-node sync::session` — 44 passed (40 existing + 4 new: syncing / idle / backing-off / absent-context)
- Clippy on touched crates — no findings in changed code

---

> [!NOTE]
> **Low Risk**
> Read-only observability path with advisory snapshots; no changes to sync control flow or execution authorization beyond clearer client-facing status.
> 
> **Overview**
> Adds a **`sync_status` JSON-RPC method** so clients blocked on `Uninitialized` can see whether a context is actively syncing, waiting for peers, or backing off after failures—instead of inferring from a single execute error.
> 
> The sync run-loop’s **`SessionTracker` now publishes** coarse `SyncPhase` snapshots (idle / syncing / backing off, plus `failure_count` and `last_error`) into a lock-free **`NodeState.sync_status` map** after every relevant transition, including benign no-peer / not-materialized outcomes (refactored so status is published on those paths too). **`NodeClient.sync_status`** reads via **`NodeMessage::GetSyncStatus`** through `NodeManager`.
> 
> The server handler merges **`is_initialized`** (non-zero root hash) with the advisory snapshot and maps to wire **`SyncState`**, notably turning uninitialized + idle/no snapshot into **`waitingForPeers`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52e004f933df165d683fc5a8b9fe6b9de8f418fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/primitives/src/client.rs
+++ b/crates/node/primitives/src/client.rs
@@ -780,6 +780,30 @@ impl NodeClient {
         &self.sync_client
     }
 
+    /// Read the best-effort sync-status snapshot for `context_id`.
+    ///
+    /// Round-trips through `NodeManager`, which owns the run-loop-published
+    /// snapshot map. `Ok(None)` means the sync run-loop has no record for the
+    /// context — it was created locally, or just joined and has not been
+    /// dispatched yet. The snapshot is advisory (see
+    /// [`SyncStatusSnapshot`](crate::SyncStatusSnapshot)); use it for UX, not
+    /// control flow.
+    pub async fn sync_status(
+        &self,
+        context_id: ContextId,
+    ) -> eyre::Result<Option<crate::SyncStatusSnapshot>> {
+        let (tx, rx) = oneshot::channel();
+        self.node_manager
+            .send(NodeMessage::GetSyncStatus {
+                context_id,
+                outcome: tx,
+            })
+            .await
+            .map_err(|_| eyre::eyre!("node manager mailbox dropped"))?;
+        rx.await
+            .map_err(|_| eyre::eyre!("sync status response channel dropped"))
+    }
+
     pub async fn sync(
         &self,
         context_id: Option<&ContextId>,
--- a/crates/node/primitives/src/lib.rs
+++ b/crates/node/primitives/src/lib.rs
@@ -9,6 +9,8 @@ pub mod join_bundle;
 pub use join_bundle::JoinBundle;
 pub mod messages;
 pub mod sync;
+pub mod sync_status;
+pub use sync_status::{SyncPhase, SyncStatusSnapshot};
 pub mod topic_manager;
 pub use topic_manager::TopicManager;
 
--- a/crates/node/primitives/src/messages.rs
+++ b/crates/node/primitives/src/messages.rs
@@ -53,4 +53,14 @@ pub enum NodeMessage {
     /// claim "FSM observes every monotonic advance regardless of
     /// origin" only held for the receive path until #2237 follow-up.
     ForwardNamespaceOpApplied { namespace_id: [u8; 32] },
+    /// Read the best-effort sync-status snapshot the sync run-loop has
+    /// recorded for a context. Routed through `NodeClient -> NodeManager`
+    /// because the snapshot lives on the node-crate-private `NodeState`,
+    /// which the server layer cannot name directly. `outcome` carries
+    /// `None` when the run-loop has no record for the context (never
+    /// synced — e.g. created locally or just joined).
+    GetSyncStatus {
+        context_id: ContextId,
+        outcome: oneshot::Sender<Option<crate::SyncStatusSnapshot>>,
+    },
 }
--- a/crates/node/primitives/src/sync_status.rs
+++ b/crates/node/primitives/src/sync_status.rs
@@ -11,12 +11,14 @@
 //! lock-free map and read out-of-band, so a reader may observe a value that is
 //! a few hundred milliseconds stale. It is meant for UX ("syncing, please
 //! wait" vs "stuck"), not for control flow.
-
-use serde::{Deserialize, Serialize};
+//!
+//! These types are passed in-process only (run-loop publisher → `NodeManager`
+//! → `NodeClient`); the wire-facing shape lives in `calimero-server-primitives`
+//! (`jsonrpc::SyncState`), which the server handler maps onto. So there are
+//! deliberately no `serde` derives here.
 
 /// A snapshot of where a context is in the sync lifecycle.
-#[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
+#[derive(Clone, Debug)]
 pub struct SyncStatusSnapshot {
     /// Coarse phase the sync manager last recorded for this context.
     pub phase: SyncPhase,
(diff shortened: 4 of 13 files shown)


chefsale:
Thanks for the reviews — pushed `52e004f` addressing the substantive findings.

**Fixed**

- **🔴 No-peers shows `idle` (Bugbot, meroreviewer)** — the real one. A benign `NoPeersAvailable` / `PeerNotMaterialized` outcome calls `on_not_materialized()`, which clears the in-flight marker *without* bumping `failure_count`, so the node phase lands on `Idle`. The handler mapped that straight to `"idle"`, so an uninitialized context with no peer looked settled — exactly the case this endpoint exists to disambiguate. The handler now resolves the phase against `is_initialized`: uninitialized + idle (or never-dispatched) → **`waitingForPeers`**. Extracted as a pure `resolve_sync_state()` with unit tests covering all four arms.
- **Stringly-typed `sync_state` (both reviewers)** — replaced `sync_state: String` + separate `retry_in_secs` with a typed, internally-tagged `SyncState` wire enum, so `retryInSecs` is bound to `BackingOff` and clients can match exhaustively. Added the `WaitingForPeers` variant.
- **Misleading serde on `SyncPhase`/`SyncStatusSnapshot`** — dropped the derives; those types are in-process only (run-loop → `NodeManager` → `NodeClient`), the wire shape lives in server-primitives.
- **Unused `SyncStatusSnapshot::idle()`** — removed.
- **`error!` log-flooding on unknown context** — downgraded to `debug!`, matching the execute path.

**`is_initialized: true` + `syncState: "syncing"`** — keeping this as-is. It's not actually contradictory: `is_initialized` answers "can I execute / is the initial state here", while `sync_state` reports whether sync activity is happening. An initialized context legitimately keeps syncing later deltas, so reporting both independently is correct (documented on the response type). The normalization only applies to the genuinely ambiguous idle case above.

**Consciously not changing**

- **Actor round-trip vs direct `Arc<DashMap>` handle** — `NodeClient` is constructed before `NodeState` in `run.rs`, so injecting the handle means reordering construction; the `NodeManager` hop avoids that and mirrors the existing `GetBlobBytes` pattern. Latency is irrelevant on an observability path. Open to revisiting if you'd prefer the direct handle.
- **`retry_in_secs` staleness** — inherent to the advisory snapshot model (published at failure time, read later); documented as advisory/UX, not control-flow.
- **`last_error` exposure** — intended (it's the "why" behind a stuck sync, e.g. "No peers to sync with") and behind auth. It's the eyre `Display`, not `Debug`, so no backtraces/paths.

chefsale:
Thanks — went through the 4 new 🟡 from the latest pass. None need a change to this PR; here's the reasoning, with two of them resolved upstream in the stack.

**1. `BackingOff retry_in_secs … after on_not_materialized` (failed, then ran out of peers → reports `BackingOff`)** — real inaccuracy in *this* PR, and **fixed in the stacked #2659**. There, the benign no-peers / not-materialised arms publish `WaitingForPeers` explicitly at the source instead of letting it derive as `BackingOff`/`idle`. Since #2652 → #2659 → #2660 merge as one stack, this residual is gone by the time it lands. (Backporting it into #2652 only to undo it in #2659 isn't worth the churn.)

**2. `BackingOff failure_count > 0 but last_error may be None` — internally inconsistent** — this combination **can't occur**. `failure_count` is incremented only by `on_failure`, which always sets `last_error = Some` in the same call (`tracking.rs:98-102`); `on_success` clears both together; `on_not_materialized` touches neither. So the invariant `failure_count > 0 ⟹ last_error = Some` holds, and the "N failures, null error" response the comment describes isn't reachable. The reviewer's own note ("`last_error` is *not cleared* by `on_not_materialized`") actually confirms it stays `Some`.

**3. `retry_in_secs` is the full backoff at publish time and goes stale** — accurate, and **by design** for this advisory snapshot. It's published once on transition with the backoff that *just* started; it does not tick down between transitions. That's the documented contract (the type's doc calls the whole snapshot advisory/UX, and `retry_in_secs` "the manager's best estimate"). Making it a live countdown would mean re-publishing every run-loop tick — which would also re-fire WS events on every decrement — for a value clients can already treat as "≤ this long." Deliberately not doing that.

**4. `Syncing` reported with a non-zero `failure_count` on a non-first re-sync** — intended. `failure_count` describes the last settled outcome; surfacing "currently syncing, has failed N times before" is useful history, not a contradiction. It resets to 0 on the next success.

Net: #1 resolved in #2659, #2 not reachable, #3/#4 intentional. Happy to revisit #3 if we decide a live countdown is worth the re-publish cost.

PR #2659, opened 2026-06-03
feat(api): stream sync status over WebSocket + snapshot progress (#2027)

chefsale:
**Stacked on #2652** (base branch is `feat/2027-sync-status`). Review/merge that first; this diff shows only the Phase 2 delta.

## What

Phase 2 of #2027: makes `sync_status` **push-based** over WebSocket and adds **snapshot-receive progress** — completing the remaining acceptance criteria.

## Changes

- **Unified wire vocabulary.** Moved the `SyncState` phase enum down into `calimero-primitives` so the JSON-RPC response *and* the new WebSocket event share one type (server-primitives re-exports it; the node side uses it directly, dropping the separate `SyncPhase`). No more parallel representations.

- **WebSocket streaming.** New `ContextEventPayload::SyncStatus` event, emitted from the run-loop whenever a context's phase changes. Reuses the existing per-context subscription fan-out — a client already subscribed to a context now receives live phase transitions instead of polling. Emission is change-detected, so the stream stays quiet between real transitions.

  Event shape (under the existing `{type, data}` envelope):
  ```json
  { "type": "SyncStatus", "data": { "syncState": { "state": "syncing" }, "failureCount": 0 } }
  ```

- **`WaitingForPeers` reported at the source.** The benign no-peers / not-materialised outcome clears the in-flight marker without recording a failure, so the run-loop used to derive it as plain `idle`. It now publishes `WaitingForPeers` explicitly, so the RPC *and* the event stream are accurate and consistent. The handler's `is_initialized` normalization (added in #2652) becomes just a safety net for stale snapshots.

- **Snapshot progress.** Emits `ReceivingSnapshot { recordsReceived }` per page-burst from the receive loop, to both the status mirror and the event stream.

## Scope note — why `recordsReceived`, not a percentage

`recordsReceived` is a **monotonic liveness signal, not a percent**. The snapshot stream is cursor-paged and the `SnapshotPage` wire message carries only per-burst `page_count`/`sent_count` — the receiver is never told a grand total. A true percent/ETA would require the **sender to advertise `total_entries`** in the page message (which it already computes and logs), a backward-incompatible wire change I've deliberately left as a follow-up rather than fabricate a number. Tracked as the remaining open item on the issue.

`is_initialized` is intentionally **not** on the event (it's a context-layer fact, not a sync-phase one); clients read it from the RPC or infer it from the first `StateMutation` event.

## Acceptance criteria (issue #2027), after #2652 + this

- [x] sync progress tracked in `NodeState`
- [x] `sync_status` endpoint returns progress
- [x] **WebSocket subscription for live updates** ← this PR
- [x] **progress updates during snapshot sync** (`recordsReceived`) ← this PR
- [ ] ETA / percent — needs sender-advertised total (documented follow-up)

## Testing

- `cargo fmt --check` — clean
- `cargo test -p calimero-node sync::session` — 45 passed (incl. new no-peers → `WaitingForPeers` regression test)
- `cargo test -p calimero-server jsonrpc::sync_status` — 3 passed (normalize unit tests)

---

> [!NOTE]
> **Low Risk**
> Advisory status and event emission only; no changes to sync execution, auth, or persistence beyond mirroring existing tracker state.
> 
> **Overview**
> **Unifies sync phase types** by moving `SyncState` into `calimero-primitives` and replacing the node-only `SyncPhase` / server duplicate enum so JSON-RPC, the advisory snapshot, and WebSocket events share one vocabulary (`SyncStatusSnapshot.state`).
> 
> **Adds live `SyncStatus` WebSocket events** on phase changes from `SessionTracker` (optional `NodeClient`, change-detected) and **snapshot page progress** as `ReceivingSnapshot { records_received }` from the snapshot receive loop, both updating the existing lock-free status mirror.
> 
> **Publishes `WaitingForPeers` at the source** when benign `NoPeersAvailable` / peer-not-materialized outcomes occur, instead of deriving idle; JSON-RPC `sync_status` mostly passes through with a small `normalize` for initialized contexts.
> 
> **Risk:** Low–medium — observability and API surface only; sync control logic unchanged aside from clearer status reporting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 408cbd95a610fcded13a500e188a02f93594c43b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/primitives/src/lib.rs
+++ b/crates/node/primitives/src/lib.rs
@@ -10,7 +10,7 @@ pub use join_bundle::JoinBundle;
 pub mod messages;
 pub mod sync;
 pub mod sync_status;
-pub use sync_status::{SyncPhase, SyncStatusSnapshot};
+pub use sync_status::{SyncState, SyncStatusSnapshot};
 pub mod topic_manager;
 pub use topic_manager::TopicManager;
 
--- a/crates/node/primitives/src/sync_status.rs
+++ b/crates/node/primitives/src/sync_status.rs
@@ -8,51 +8,25 @@
 //! per-context bookkeeping so the distinction is observable.
 //!
 //! The data is advisory: it is published from the sync run-loop on a
-//! lock-free map and read out-of-band, so a reader may observe a value that is
-//! a few hundred milliseconds stale. It is meant for UX ("syncing, please
-//! wait" vs "stuck"), not for control flow.
+//! lock-free map and read out-of-band (or pushed as a WebSocket event), so a
+//! reader may observe a value that is a few hundred milliseconds stale. It is
+//! meant for UX ("syncing, please wait" vs "stuck"), not for control flow.
 //!
-//! These types are passed in-process only (run-loop publisher → `NodeManager`
-//! → `NodeClient`); the wire-facing shape lives in `calimero-server-primitives`
-//! (`jsonrpc::SyncState`), which the server handler maps onto. So there are
-//! deliberately no `serde` derives here.
+//! The phase vocabulary ([`SyncState`]) is the shared wire type from
+//! `calimero-primitives`, so the run-loop, the JSON-RPC response, and the
+//! WebSocket event all speak the same language.
+
+pub use calimero_primitives::sync_status::SyncState;
 
 /// A snapshot of where a context is in the sync lifecycle.
 #[derive(Clone, Debug)]
 pub struct SyncStatusSnapshot {
-    /// Coarse phase the sync manager last recorded for this context.
-    pub phase: SyncPhase,
+    /// Coarse phase the sync run-loop last recorded for this context.
+    pub state: SyncState,
     /// Consecutive failed sync attempts. Resets to 0 on the next success.
-    /// A non-zero value with `phase == BackingOff` is the "stuck" signal.
+    /// A non-zero value with `state == BackingOff` is the "stuck" signal.
     pub failure_count: u32,
     /// The error from the most recent failed attempt, if any. Carries the
-    /// reason behind a `BackingOff` phase (e.g. "No peers to sync with"),
-    /// which is what lets a client distinguish "waiting for a peer" from a
-    /// genuine protocol failure.
+    /// reason behind a `BackingOff` state (e.g. "No peers to sync with").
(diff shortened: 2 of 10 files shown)


chefsale:
Thanks — pushed `408cbd95` for the two real ones, and the other two don't hold up on a closer look.

**Fixed**

- **🔴 / 🟡 `emit_snapshot_progress` clears `failure_count`/`last_error`** (Bugbot Medium + reviewer 💡2) — real. It wrote `0`/`None` unconditionally, so a context with prior failures would flip to "healthy" mid-snapshot until the next run-loop publish. Now it reads the existing snapshot and carries the failure history forward (this path only advances the *phase*); the fields still clear naturally on the run-loop's `on_success`. Done with an owned copy + dropped guard before `insert`, since a same-key `get`+`insert` on a `DashMap` shard would deadlock.

- **💡1 change detection ignores `last_error`** — added it to the predicate. In practice a `last_error` change always travels with a `failure_count` bump (`on_failure` does both) or a state change (`on_success`), so the gap wasn't reachable, but it's a one-liner that bulletproofs the event stream against future transitions.

**Not changing**

- **🟡1 `records_received` goes backwards across bursts** — not a bug; the premise is incorrect. `total_applied` is declared once at function scope (`snapshot.rs:478`) and accumulated across every burst (`total_applied += applied`, `:693`); `request_and_apply_snapshot_pages` is called **once per snapshot** (`:298`) with the cursor/burst loop *inside* it. So it's cumulative over the whole snapshot and monotonic — the reviewer assumed one call per burst, but there's exactly one call per sync.

- **🟡2 `ContextEventPayload` untagged → may mis-deserialize** — misread. `#[serde(untagged)]` is on the outer `NodeEvent` (which has a single `Context` variant). `ContextEventPayload` itself is `#[serde(tag = "type", content = "data")]` (`events.rs:21`) — adjacently tagged, exactly the `{"type":"SyncStatus","data":{…}}` envelope the suggestion recommends. So the discriminant is already explicit on the wire; no ambiguity with `StateMutation`.

(Stack note: rebased the fix into the percent/ETA follow-up #2660 too, so it carries through.)

PR #2660, opened 2026-06-03
feat(sync): snapshot progress percent + ETA (#2027)

chefsale:
**Stacked on #2659** (base branch `feat/2027-sync-status-streaming`). Merge #2652 → #2659 → this. The diff shows only the Phase 3 delta. Closes the last open acceptance criterion on #2027.

## What

Turns the snapshot `recordsReceived` liveness signal (added in #2659) into a real **percentage + ETA**.

```json
{ "type": "SyncStatus", "data": {
  "syncState": { "state": "receivingSnapshot", "recordsReceived": 50, "percent": 25, "etaSecs": 15 },
  "failureCount": 0 } }
```

## Why it wasn't done before

The receiver was never told how big the snapshot is. The `SnapshotPage` wire message carried only per-burst `page_count`/`sent_count`, and the stream is cursor-paged, so there was no honest denominator — which is why #2659 deliberately shipped a raw count instead of a fabricated percent. The sender already computes the grand total during its boundary scan (the id-sorted `entity_ids` set); it just wasn't on the wire.

## How

- **Advertise the total.** Add `total_records: u64` to the `SnapshotPage` wire message — the grand total of entity records across the whole snapshot, stable across bursts. `0` = unknown (empty snapshot, or a peer too old to advertise). `generate_snapshot_pages` now returns the count; the sender populates it.

- **Derive on the receiver.** `percent` = `recordsReceived / total` clamped to `0..=100`; `eta_secs` extrapolates the remaining records from the average rate so far. Both are `Option` on `SyncState::ReceivingSnapshot` and fall back to `None` against a non-advertising peer — degrading gracefully to the raw `recordsReceived` from #2659. Extracted as a pure `snapshot_progress_estimate` with unit tests.

## ⚠️ Wire compatibility

Appending `total_records` to `SnapshotPage` is a **borsh-positional change**, so snapshot sync between a patched and an unpatched node is incompatible. This needs a **coordinated rollout** (or a protocol-version gate) before release — calling it out explicitly so it's a conscious merge decision, not a surprise. Everything else in the stack is additive/back-compatible; this is the one PR with a wire break.

## Denominator choice

The total is the **entity count** (`entity_ids.len()` from the sender's full scan), which matches the receiver's entity-based applied counter. `present_keys.len()` was rejected — it counts Entry+Index keys per entity and would peg progress below 100%.

## Testing

- `cargo fmt --check` — clean
- `cargo test -p calimero-node sync::snapshot::tests::snapshot_progress` — 4 passed (unknown-total, clamp + ETA extrapolation, pre-record/pre-time None, completion → 0)
- `cargo test -p calimero-server jsonrpc::sync_status` — 3 passed

## Acceptance criteria (#2027) — all met after #2652 + #2659 + this

- [x] sync progress tracked in `NodeState`
- [x] `sync_status` endpoint returns progress
- [x] WebSocket subscription for live updates
- [x] progress updates during snapshot sync
- [x] **ETA calculation based on transfer rate** ← this PR

---

> [!NOTE]
> **High Risk**
> SnapshotPage gains a borsh field that breaks wire compatibility between patched and unpatched nodes during snapshot sync; rollout must be coordinated despite the change being observability-only on the receiver.
> 
> **Overview**
> Adds **`total_records`** to the **`SnapshotPage`** borsh wire message so the snapshot sender advertises the full shippable entity count (stable across cursor-paged bursts); **`0`** means unknown for empty snapshots or older peers.
> 
> On the receiver, each applied page burst now computes **`percent`** and **`eta_secs`** from cumulative applied entities, that total, and elapsed time via **`snapshot_progress_estimate`**, and publishes them on **`SyncState::ReceivingSnapshot`** (JSON-RPC / WebSocket) while still emitting raw **`records_received`** when no total is advertised.
> 
> **Wire note:** appending **`total_records`** is a **borsh-positional break** for **`SnapshotPage`** — mixed-version snapshot sync needs a coordinated rollout or protocol gate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8677cbe9a0aa1b93040783ce9af23ea370198c6a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/primitives/src/sync/wire.rs
+++ b/crates/node/primitives/src/sync/wire.rs
@@ -352,6 +352,13 @@ pub enum MessagePayload<'a> {
         page_count: u64,
         /// Pages sent so far.
         sent_count: u64,
+        /// Grand total of entity records across the whole snapshot (stable
+        /// across bursts — the count from the sender's full boundary scan),
+        /// so the receiver can compute a real percent/ETA against its
+        /// cumulative applied count. `0` means "unknown" — an empty snapshot,
+        /// or a peer too old to advertise it; the receiver then reports raw
+        /// progress only.
+        total_records: u64,
     },
 
     /// Snapshot sync error.
--- a/crates/node/primitives/src/sync/wire.rs
+++ b/crates/node/primitives/src/sync/wire.rs
@@ -352,12 +352,13 @@ pub enum MessagePayload<'a> {
         page_count: u64,
         /// Pages sent so far.
         sent_count: u64,
-        /// Grand total of entity records across the whole snapshot (stable
-        /// across bursts — the count from the sender's full boundary scan),
-        /// so the receiver can compute a real percent/ETA against its
-        /// cumulative applied count. `0` means "unknown" — an empty snapshot,
-        /// or a peer too old to advertise it; the receiver then reports raw
-        /// progress only.
+        /// Grand total of shippable `Entity` records across the whole snapshot
+        /// — every entity with both an `Index` and an `Entry`, counted from the
+        /// sender's full boundary scan and stable across bursts. Excludes
+        /// orphans that are never shipped, so it's the exact denominator the
+        /// receiver's cumulative applied count reaches (percent → 100, ETA →
+        /// 0). `0` means "unknown" — an empty snapshot, or a peer too old to
+        /// advertise it; the receiver then reports raw progress only.
         total_records: u64,
     },
 
--- a/crates/node/src/sync/snapshot.rs
+++ b/crates/node/src/sync/snapshot.rs
@@ -149,7 +149,7 @@ impl SyncManager {
         stream: &mut Stream,
     ) -> Result<()> {
         let handle = self.context_client.datastore_handle();
-        let (pages, next_cursor, total_entries, grand_total_entities) = generate_snapshot_pages(
+        let (pages, next_cursor, total_entries) = generate_snapshot_pages(
             &handle,
             context_id,
             start_cursor.as_ref(),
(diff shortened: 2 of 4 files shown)


chefsale:
Both real — fixed in `8677cbe9`.

- **🔴 Progress total counts unshipped entities** (Bugbot Medium) — correct. The advertised total was `entity_ids.len()`, i.e. every index-discovered entity *including orphans* (Index without a matching Entry), which the receiver never applies — so percent/ETA could asymptote below 100% when orphans exist. Fixed by advertising the existing `total_entries` accounting figure instead, which already tallies exactly the shippable entities (`has_index && has_entry`, `snapshot.rs:1197`), is stable across bursts, and excludes orphans. So the denominator now equals what the receiver actually applies → percent reaches 100, ETA reaches 0. This also drops the redundant 4th return value I'd added — net simpler.

- **🟡 ETA float→u64 cast** — `(remaining / rate).ceil() as u64` is actually already safe in current Rust (the `as` cast saturates: NaN → 0, +∞ → u64::MAX — the "UB before 1.45" the comments cite is outdated). But the NaN→0 path could in theory surface a misleading "0s left", so I guarded it explicitly: only emit when the estimate `is_finite()`, and `.min(u64::MAX as f64)` before the cast. `rate` is always finite-positive given the `records_received > 0 && secs > 0.0` guard, so this is belt-and-suspenders, but it quiets the concern cleanly.

Both verified: `cargo test sync::snapshot::tests::snapshot_progress` green, fmt/build clean.

PR #2662, opened 2026-06-03
fix(storage): implement full HLC update rule and fix counter masking

### LINKED ISSUE #2635: LogicalClock::update never bumps the counter on equal timestamp (breaks causal monotonicity)
## Summary

`LogicalClock::update` (`crates/storage/src/logical_clock.rs:328-334`) does not implement the full Hybrid Logical Clock update rule. When a remote timestamp's physical time **equals** `self.last_time`, it leaves `counter` untouched (and it never reads the remote counter at all). The canonical HLC rule requires `counter = max(local_counter, remote_counter) + 1` on the equal case, so without it the next locally-issued timestamp can be **≤ a previously-issued one**, violating causal monotonicity.

## Details

```rust
// Update to max observed time (causality)
if remote_time > self.last_time {
    self.last_time = remote_time;
    self.counter = 0;
}
Ok(())
```

Problems vs. the canonical HLC `update(remote)`:
- **Equal-time case (`remote_time == self.last_time`)**: nothing happens. The rule should set `counter = max(self.counter, remote.counter) + 1` so a subsequent local event sorts strictly after both the local and the just-observed remote event.
- **Greater-time case**: it resets `counter = 0` and ignores `remote.counter`. Canonically the counter should be `remote.counter + 1` (a later local event must outrank the remote event it just observed at that same physical time). Resetting to 0 can produce a local timestamp that ties/precedes the remote one it causally follows.
- The remote counter is never consulted in either branch.

## Impact

Causal monotonicity is not guaranteed across merge/observe. Two events where A causally precedes B can receive equal or inverted HLC timestamps, which breaks LWW tie-breaking and any ordering that relies on the clock (the whole point of an HLC). This is the kind of subtle clock defect that surfaces as non-deterministic convergence / split-brain under concurrency.

## Suggested fix

Implement the full HLC update rule, comparing against both physical time and the remote counter:

```
let new_time = max(self.last_time, remote_time, local_ntp);
self.counter = if new_time == self.last_time && new_time == remote_time {
    max(self.counter, remote_counter) + 1
} else if new_time == self.last_time {
    self.counter + 1
} else if new_time == remote_time {
    remote_counter + 1
} else {
    0
};
self.last_time = new_time;
```

(Adjust to the existing field types; the key points are: incorporate `remote_counter`, and always advance the counter when physical time does not strictly increase.)

## References

- `crates/storage/src/logical_clock.rs:328-334`

_Found during security review (finding #25)._

### LINKED ISSUE #2661: logical_counter() masks the HLC counter with 0xF (4 bits) but it is 16 bits wide
## Summary

`logical_counter` in `crates/storage/src/logical_clock.rs` extracts the HLC logical counter with `& 0xF`, masking off all but the low **4 bits**. Every other site in the module treats the counter as **16 bits wide** — it is embedded via `& 0xFFFF` in `new_timestamp` and the field comment documents "low 16 bits of timestamp". So any counter value greater than 15 is silently truncated when read back.

## Details

```rust
// crates/storage/src/logical_clock.rs
pub fn logical_counter(ts: &HybridTimestamp) -> u32 {
    (ts.0.get_time().as_u64() & 0xF) as u32   // <-- should be 0xFFFF
}
```

vs. the embed in `new_timestamp`:

```rust
let time_with_counter = NTP64((self.last_time & !0xFFFF) | (self.counter as u64 & 0xFFFF));
```

The counter is a `u16`, so under any sustained burst (a stalled physical clock issues monotonically increasing counters 0, 1, 2, …), `logical_counter` returns a value that wraps every 16 increments instead of every 65536, so it no longer reflects the real ordinal of the event within the tick.

## Impact

`logical_counter` is surfaced through `StorageDelta`/delta metadata as `logical_clock()`. Today nothing compares on it, so there is no live data-loss bug, but it is a latent footgun: any future ordering/debugging logic that reads the logical counter would see truncated, non-monotonic values. It is also a maintenance hazard — the masks for the same field disagree across the file.

## Suggested fix

Mask with the full 16-bit counter width, and centralize the bit layout in one named constant so the masks cannot drift apart again. Found alongside #2635 (the `update` rule defect); both touch the same counter-encoding seam and are fixed together.

chefsale:
## What

Comprehensive fix for the Hybrid Logical Clock in `crates/storage/src/logical_clock.rs`, covering two related defects on the same counter-encoding seam.

### `update()` did not implement the HLC update rule (#2635)

The old `update`:

```rust
if remote_time > self.last_time {
    self.last_time = remote_time;
    self.counter = 0;
}
```

- **Equal physical time** → did nothing: the counter was never moved, so the next local event could tie a just-observed remote event.
- **Greater physical time** → reset the counter to `0` and never read `remote_counter`, so the next local event could *precede* the remote event it causally follows.

Either way causal monotonicity was not guaranteed across observe/merge — the kind of subtle clock defect that surfaces as non-deterministic convergence / split-brain under concurrency.

This implements the canonical receive rule: advance to `max(local, remote, wall_clock)` physical time, and whenever physical time does not strictly increase, advance the counter past every event seen at that tick (consuming `remote_counter` on the equal/greater cases).

#### Before vs after

A local clock at physical tick `P` with counter `5` observes a remote event `(P, 9)`, then issues a new event:

| | before | after |
|---|---|---|
| after `update((P,9))` | counter stays `5` | counter → `10` |
| next `new_timestamp()` | emits `(P, 6)` ❌ `< (P,9)` | emits `(P, 11)` ✓ `> (P,9)` |

### `logical_counter()` truncated the counter to 4 bits (#2661)

`logical_counter` masked with `0xF` while every other site uses the full 16-bit width, silently truncating any counter > 15. Surfaced through `StorageDelta::logical_clock()` (no live consumer compares on it today, so latent, but a footgun).

## Also in this change

- **Consistent internal representation**: `last_time` is now quantized to the physical-time bits, so the internal clock and emitted timestamps use one encoding. Previously `new_timestamp` kept full fractional precision in `last_time` while `update` stored counter bits in the same field — two incompatible encodings being compared with `>`.
- **Counter-overflow carry**: a stalled wall clock issuing more than 2^16 events per tick now carries into physical time instead of wrapping the counter back to zero, preserving strict monotonicity.
- **Centralized bit layout** in named constants (`COUNTER_BITS`/`COUNTER_MASK`/`PHYSICAL_MASK`) so the masks can no longer drift apart.
- **Doc fixes**: the module claimed to "wrap the uhlc crate" (it is a self-contained reimplementation) and that the "Last 4 bits = logical counter" (16 bits).

## Tests

Added unit tests for: equal-time update bumps the counter, greater-time update consumes the remote counter, cross-node causality after observe, full-width (16-bit) counter extraction, counter-overflow monotonicity (70k events on a stalled clock), and drift rejection. Full `calimero-storage` suite green (553 lib + converge/crdt_contract/merge-registry/rekey integration tests), `cargo fmt --check` clean, no new clippy findings.

Closes #2635
Closes #2661

---

> [!NOTE]
> **High Risk**
> Changes core timestamp ordering used by CRDTs and sync (`hlc_timestamp` / `update_hlc`); incorrect HLC logic causes subtle convergence bugs, though behavior is now spec-correct with broad tests.
> 
> **Overview**
> Fixes the **Hybrid Logical Clock** in `logical_clock.rs` so CRDT ordering stays strictly causal after observing remote timestamps.
> 
> **`update()`** now follows the full HLC receive rule: physical time becomes `max(local, remote, wall clock)`, and when physical time does not move forward the logical counter advances past both local and remote events (including reading the remote counter). The old path could leave the next local timestamp **≤** a just-observed remote event at the same tick.
> 
> **Counter encoding** is aligned everywhere: `logical_counter()` used a 4-bit mask while the clock used 16 bits; shared `COUNTER_MASK` / `PHYSICAL_MASK` constants replace scattered masks. **`last_time`** is stored as physical-only bits so `new_timestamp` and `update` compare the same representation. **`new_timestamp`** uses `set_after` for counter bumps and **carries into physical time** on 16-bit overflow instead of wrapping. Drift rejection compares **physical** time only. Docs describe the in-tree HLC (not a uhlc wrapper) and 16-bit counter layout. New unit tests cover causality branches, full-width counter readout, overflow monotonicity, and far-future rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aece3a9948392d7eddd7d92462fa80c395f115b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/storage/src/logical_clock.rs
+++ b/crates/storage/src/logical_clock.rs
@@ -278,7 +278,10 @@ impl LogicalClock {
         if let Some(next) = base_counter.checked_add(1) {
             self.counter = next;
         } else {
-            self.last_time = phys.wrapping_add(1 << COUNTER_BITS);
+            // Carry one tick into physical time. Re-mask defensively so the
+            // counter bits stay zero even if a future caller passes an
+            // unquantized `phys` (the OR in `new_timestamp` relies on this).
+            self.last_time = phys.wrapping_add(1 << COUNTER_BITS) & PHYSICAL_MASK;
             self.counter = 0;
         }
     }
@@ -315,6 +318,8 @@ impl LogicalClock {
         }
 
         // Embed the counter in the reserved low bits of the timestamp.
+        // `set_after` always leaves `last_time` quantized (counter bits zero)
+        // and `counter` in range, so this OR never collides.
         let time_with_counter = NTP64(self.last_time | u64::from(self.counter));
 
         // Safety: self.id is initialized to non-zero in `new()` and never changes
@@ -349,11 +354,13 @@ impl LogicalClock {
         let frac = (nanos * (1_u64 << 32)) / 1_000_000_000;
         let local_ntp = (secs << 32) | frac;
 
-        // Drift protection: reject if >5s in future
+        // Drift protection: reject if >5s in future. Compare physical time
+        // only — the remote counter bits are logical ordering, not clock
+        // drift, and would otherwise inflate the comparison at the boundary.
         const DRIFT_TOLERANCE_SECS: u64 = 5;
         let drift_ntp = local_ntp + (DRIFT_TOLERANCE_SECS << 32);
 
-        if remote_time > drift_ntp {
+        if remote_phys > drift_ntp {
             return Err(());
         }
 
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed the MeroReviewer findings in aece3a99:

- **W2 (drift check uses raw `remote_time`)** — valid, fixed. The drift check now compares `remote_phys`, so a large remote counter (up to ~15µs at the 5s boundary) can no longer inflate the value and spuriously reject a timestamp.
- **S1 (untested branches 2 & 4)** — added `test_update_local_clock_ahead_of_remote_still_advances` (local already at max tick) and `test_update_wall_clock_ahead_resets_counter` (wall clock strictly ahead → counter reset).
- **S2 (set_after consistency comment)** — added.

- **W1 (carry not re-quantized → corrupt `last_time`)** — not actually a live bug: `phys` is quantized by every caller (`PHYSICAL_MASK`), and adding `1 << COUNTER_BITS` to a value whose low 16 bits are zero leaves them zero, so `last_time` was already valid for the OR in `new_timestamp`. Applied the suggested `& PHYSICAL_MASK` anyway as cheap hardening so `set_after` no longer depends on the caller honoring that precondition.

All 11 logical_clock tests plus the converge/crdt_contract suites pass; fmt clean.

PR #2664, opened 2026-06-03
feat(dag): bound delta-log growth with DAG compaction (#2026)

### LINKED ISSUE #2026: feat(dag): Implement DAG compaction to bound storage growth
## Problem

DAG history grows unbounded:

```rust
// All deltas persisted forever in ContextDagDelta
// parent_hashes pruned at 10k but deltas remain in DB
```

**Impact**:
- Storage grows linearly with transaction count
- Old deltas never cleaned up
- Sync sends full history even when snapshot would be smaller

## Current State

Checkpoints are partially supported:

```rust
// crates/node/src/delta_store.rs:1139
let checkpoint = CausalDelta::checkpoint(head_id, boundary_root_hash);
```

But no compaction logic exists to:
1. Determine when to checkpoint
2. Prune deltas older than checkpoint
3. Handle sync requests for pruned deltas

## Proposed Solution

1. Add compaction configuration:

```rust
pub struct DagCompactionConfig {
    /// Minimum deltas before compaction eligible
    pub min_deltas_before_compact: usize,  // Default: 10000
    /// Keep recent history for delta sync
    pub retain_recent_count: usize,        // Default: 1000
    /// Compaction interval
    pub check_interval: Duration,          // Default: 1 hour
}
```

2. Implement compaction logic that:
   - Creates checkpoint at boundary
   - Prunes deltas older than checkpoint
   - Handles sync for pruned history (offer snapshot)

## Acceptance Criteria

- [ ] Compaction runs periodically based on config
- [ ] Creates checkpoint delta at compaction boundary
- [ ] Prunes deltas older than checkpoint from DB
- [ ] Retains configurable recent history
- [ ] Sync falls back to snapshot for pruned deltas
- [ ] Compaction metrics exported

## Risks & Mitigations

| Risk | Mitigation |
|------|------------|
| Data loss if checkpoint fails | Transactional checkpoint + prune |
| Slow nodes need old deltas | Keep generous retain_recent_count |
| Checkpoint verification | Verify state hash matches before prune |

**Priority**: P3 (Medium Impact, High Effort)

chefsale:
## What

Implements DAG compaction to bound on-disk delta-log growth. Closes #2026.

Every regular delta was persisted forever in `Column::Delta`, so the DAG log grew linearly with lifetime transaction count — only the in-memory ancestry caps (`MAX_TOPOLOGY_ENTRIES` etc.) were bounded. This adds periodic pruning of history older than a configurable recent window, from both the in-memory DAG and the durable delta column.

## Why it's safe to prune

Deltas are an **optimization, not a convergence requirement**. `select_protocol` never returns `DeltaSync` — a diverged *initialized* node converges via `HashComparison`, which reconciles current state from the Merkle tree **without consulting the delta log**. Delta-catchup is only HC's fallback. So pruned history is never needed for correctness:

- A peer that requests a pruned delta gets **not-found** and aborts delta-catchup.
- Its next sync round selects `HashComparison` (already the default for diverged nodes) and converges from state.

The retained window keeps the cheap single-round delta-catchup fast-path for small reconnect gaps.

## Design decision

Considered (a) prune outright + not-found→HC fallback, (b) frontier checkpoint objects, (c) single head-checkpoint. Chose **(a)**: simplest, fewest moving parts, and leans on the protocol that already handles divergence. (b)/(c) still require the requester-side detection, so they only add an object to maintain.

## Changes

- **`DagCompactionConfig`** (`[dag_compaction]`, **disabled by default**) plumbed `ConfigFile → NodeConfig → merod`. Defaults anchor to existing constants rather than magic numbers:
  - `min_deltas_before_compact = 10000` (= `MAX_TOPOLOGY_ENTRIES`, so on-disk prune and in-memory eviction trip at one threshold)
  - `retain_recent_count = 1000` (≤ `MAX_DELTA_QUERY_LIMIT = 3000`, so a behind peer catches up in one delta round; 10:1 gap gives hysteresis)
  - `check_interval = 1h`
- **`DagStore::prune_to_recent`** + `delta_count` — retain set = BFS-from-heads up to `retain_recent_count`; heads always kept; pending never pruned; **no re-parenting** (traversal already stops at an absent parent).
- **`ContextClient::prune_delta_records`** — atomic `Transaction` batch delete, mirroring `persist_delta_records`.
- **`DeltaStore::compact`** — runs under the DAG write lock (serialises against the responder send path + apply); skips mid-sync (pending > 0).
- **`DagCompactor`** actor — interval-scheduled, mirrors `GarbageCollector`; sweeps live contexts; gated on `enabled && is_valid()`.
- **Requester fallback** — `request_missing_deltas` aborts on a pruned ancestor and defers to HC.
- **Metrics** — `dag_compaction_deltas_pruned_total`, `dag_compaction_pruned_ancestor_fallbacks_total` (a rising fallback counter means `retain_recent_count` is too small for this node's reconnect gaps).

## Tests

- `DagStore::prune_to_recent`: linear-chain prune, never-strands-heads, below-threshold no-op.
- `DagCompactionConfig`: disabled default, validity guard, absent-field fallback, seconds round-trip.
- `cargo fmt --check` clean; `merod` builds.

## Why draft / not yet enabled

Compaction is **off by default** and should stay off until the pruned-ancestor → HashComparison fallback has soaked in e2e/merobox. Never prune in production before that path is proven.

- e2e/merobox soak of the fallback path, then flip `enabled` on.
- `DeltaStore::compact` integration test (needs a full `ContextStorageApplier`; only the pure dag-level prune is unit-tested here).
- Cold (non-loaded) context compaction — currently only live in-memory DAGs are swept.
- Coordinate with the parallel `rotation_log.rs` P6 compaction effort (shared threshold/config story).

---

> [!NOTE]
> **Medium Risk**
> When enabled, pruning changes what peers can serve over delta-sync; correctness relies on HashComparison fallback and skipping compaction during pending catch-up. Default-off limits production exposure, but misconfiguration or premature enablement could widen sync gaps for lagging peers.
> 
> **Overview**
> Adds **optional DAG compaction** (#2026) to cap unbounded growth of persisted context delta rows. A new **`[dag_compaction]`** section (`DagCompactionConfig`, **off by default**) is wired through **`config.toml` → `merod` → `NodeConfig`**, with defaults aligned to existing topology/query limits (e.g. 10k min before compact, 1k retain, 1h interval).
> 
> **In-memory + on-disk pruning:** `DagStore` gains **`delta_count`** and **`prune_to_recent`** (BFS from heads, always keep heads, never prune pending). **`ContextRegistry` / `ContextClient`** add **`prune_delta_records`** for atomic batch deletes in the delta column. **`DeltaStore::compact`** runs under the DAG write lock, skips contexts with pending deltas, prunes memory then calls **`prune_delta_records`**.
> 
> **Runtime:** a **`DagCompactor`** actix actor periodically sweeps **live** in-memory `DeltaStore`s (startup + interval, overlap guard). Node startup only spawns it when **`enabled && is_valid()`**. Prometheus counter **`dag_compaction_deltas_pruned_total`** records reclaimed rows.
> 
> **Sync behavior:** comments/clarifications where peers return **`DeltaNotFound`** — treat as ambiguous (compaction, race, checkpoint) and continue; convergence is expected via **HashComparison**, not delta replay.
> 
> Unit tests cover **`prune_to_recent`** and config serde/validation. Cold contexts with large on-disk logs but small in-memory DAGs are explicitly called out as a follow-up.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd11660cba534f50ffe2851d29491b55ca1d8838. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -17,7 +17,7 @@ use url::Url;
 
 use mero_auth::config::AuthConfig;
 
-pub use calimero_node_primitives::NodeMode;
+pub use calimero_node_primitives::{DagCompactionConfig, NodeMode};
 
 pub const CONFIG_FILE: &str = "config.toml";
 
@@ -68,6 +68,12 @@ pub struct ConfigFile {
     /// TEE-related configuration (KMS, attestation, etc.).
     #[serde(default)]
     pub tee: Option<TeeConfig>,
+
+    /// DAG compaction (`[dag_compaction]`) — bounds on-disk delta-log growth
+    /// by pruning history older than a checkpoint. Disabled by default;
+    /// absent section falls back to [`DagCompactionConfig::default`].
+    #[serde(default)]
+    pub dag_compaction: DagCompactionConfig,
 }
 
 /// Configuration for TEE (Trusted Execution Environment) features.
@@ -494,6 +500,7 @@ impl ConfigFile {
             context,
             runtime: RuntimeConfig::default(),
             tee: None,
+            dag_compaction: DagCompactionConfig::default(),
         }
     }
 
--- a/crates/context/primitives/src/client/mod.rs
+++ b/crates/context/primitives/src/client/mod.rs
@@ -388,6 +388,31 @@ impl ContextRegistry {
         Ok(())
     }
 
+    /// Atomically deletes a batch of DAG delta rows by key (issue #2026
+    /// compaction). Like [`persist_delta_records`](Self::persist_delta_records),
+    /// the whole batch lands in one [`Transaction`] so a crash mid-prune can
+    /// never leave the delta column half-deleted. Pruning is idempotent:
+    /// deleting an already-absent key is a no-op.
+    pub fn prune_delta_records(&self, delta_keys: &[key::ContextDagDelta]) -> eyre::Result<()> {
+        if delta_keys.is_empty() {
+            return Ok(());
+        }
+
+        let mut tx = Transaction::default();
+        for delta_key in delta_keys {
+            tx.delete(delta_key);
+        }
+
+        self.datastore.apply(&tx)?;
+
+        tracing::debug!(
+            delta_count = delta_keys.len(),
+            "Atomically pruned delta records"
+        );
+
+        Ok(())
+    }
+
     /// Updates the ApplicationId for a context.
     ///
     /// # Arguments
@@ -1135,6 +1160,12 @@ impl ContextClient {
         deltas: &[(key::ContextDagDelta, types::ContextDagDelta)],
     ) -> eyre::Result<()> {
         self.registry.persist_delta_records(deltas)
+    }
+
+    /// Atomically deletes a batch of DAG delta rows (issue #2026 compaction).
+    /// Delegates to [`ContextRegistry::prune_delta_records`].
+    pub fn prune_delta_records(&self, delta_keys: &[key::ContextDagDelta]) -> eyre::Result<()> {
+        self.registry.prune_delta_records(delta_keys)
     }
 
     /// Updates the ApplicationId for a context.
--- a/crates/dag/src/lib.rs
+++ b/crates/dag/src/lib.rs
@@ -772,11 +772,11 @@ impl<T: Clone> DagStore<T> {
     /// matching rows from durable storage).
     ///
     /// "Recent" is defined by a breadth-first walk back from the current
-    /// heads: the retained set is the heads plus the first `retain_count`
-    /// deltas reachable from them. Heads are always retained regardless of
-    /// the budget — stranding a head would corrupt parenting of future
-    /// deltas. Everything applied and outside that window is dropped from
-    /// `deltas` and `applied`.
+    /// heads: the retained set is the heads plus the first `~retain_count`
+    /// deltas reachable from them (`retain_count` is a soft floor — see the
+    /// loop body). Heads are always retained regardless of the budget —
+    /// stranding a head would corrupt parenting of future deltas. Everything
+    /// applied and outside that window is dropped from `deltas` and `applied`.
     ///
     /// Retained deltas whose parents fall outside the window keep their
     /// original parent pointers; the pruned parents simply become absent.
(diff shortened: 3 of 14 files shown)


chefsale:
Thanks for the reviews. Addressed in 72a94ff7:

**Fixed**
- **Gossip-path prune fallback** (cursor, medium): the gossip-driven parent fetch in `handlers/state_delta/mod.rs` has its own `request_missing_deltas`; it now aborts on `DeltaNotFound` and records `pruned_ancestor_fallbacks_total`, matching `SyncManager::request_missing_deltas`. This was a real gap — without it, compacted history could strand the gossip catch-up.
- **`is_valid()` rejects zero `check_interval`** — a zero interval would busy-loop `run_interval`. Added a test.
- **Overlapping sweeps** — `DagCompactor` now skips a tick if the previous sweep is still running (`AtomicBool` guard), so sweeps can't double-count metrics or race DB deletes.
- **First sweep on startup** — a node already over threshold no longer waits a full interval.
- **`prune_to_recent` BFS** — dropped the inner early-break so `retain_count` is a soft floor (outer condition is the sole budget gate; the result can only retain *more* recent history, the safe direction), and filtered the genesis root out of the retained seed so it never occupies a budget slot.
- **Docs/logs** — config module no longer says history is "collapsed into a checkpoint" (no checkpoint object is written); `compact()` now documents that `delta_count()` is the in-memory size (cold/large contexts under-prune — the tracked follow-up) and that stale-pending eviction bounds the pending-skip; sweep log now emits `contexts_compacted` alongside `contexts_scanned`.

**Not a bug (no change)**
- *"BFS queue not deduplicated → quadratic on diamond DAGs"*: the queue is already deduplicated — `!retained.insert(parent)` skips the push for any parent already retained, so each node is enqueued at most once.
- *"BFS may retain fewer than retain_count / over-prunes"*: when the old inner break fired, `retained.len()` was exactly `retain_count`, so it retained exactly the budget — not fewer. Removed it anyway for clarity (now a soft floor).

**Deferred (acknowledged in PR description)**
- Cold/non-loaded context compaction (the `delta_count()` in-memory limitation) and `DagCompactor`/`DeltaStore::compact` integration tests. Compaction stays `enabled = false` until the pruned-ancestor → HashComparison fallback soaks in e2e/merobox.

chefsale:
Confirmed real — fixed in cd11660c.

The responder sends `DeltaNotFound` in three cases, only one of which is compaction (`handle_delta_request`):
1. row absent from RocksDB **and** the in-memory `DeltaStore` — pruned *or* never-had;
2. delta present in-memory but **not yet persisted** with author info — a normal post-broadcast race (the code comment even says "initiator will re-fetch after persist settles");
3. row **without an author claim** — snapshot checkpoint / pre-author-tracking row.

Aborting the whole fetch on all three was wrong: cases 2 and 3 are unrelated to compaction and would stop the remaining batch ids and other branches from the same peer, and the `pruned_ancestor_fallbacks_total` counter would have been dominated by persist races (useless for tuning `retain_recent_count`).

It was also unnecessary: the abort was never required for correctness. A genuinely pruned ancestor leaves its descendants pending, and the next sync round converges via **HashComparison** — already the protocol selected for a diverged initialized node — which reconciles state without the delta log.

So I restored the original skip-and-continue in **both** `request_missing_deltas` paths and dropped the ambiguous receiver-side metric. The compactor-side `dag_compaction_deltas_pruned_total` stays — it's measured where the figure is unambiguous.

Issue #2233, opened 2026-04-23
DAG-causal Shared verification (epic)


xilosada:
Split out from item 5 of #2230 (\`SharedStorage\` v2). v2 ships with monotonic-nonce replay protection (matches \`UserStorage\`); this epic replaces that with **DAG-causal** verification, which correctly handles partition races that the monotonic-nonce model can't.

## Why

The original #2197 issue called out four scenarios where monotonic-nonce loses information:
- **Update-vs-rotation race.** Alice rotates Bob out (nonce N → N+1). Bob, partitioned, writes an update signed against the old set with nonce N in his view. Two deltas arrive at a third peer in either order. The correct CRDT answer accepts Bob's update if it happens-before the rotation in the DAG; monotonic-nonce can only choose by timestamp.
- **Concurrent rotations.** Two current writers issue conflicting rotations concurrently. v2 picks LWW by \`writers_nonce\`. Sequentializing via causal ordering is the principled answer.
- **Self-removal mid-flight.** Whether a rotated-out writer's in-flight updates survive depends on whether they're causally before or after the rotation.
- **Bootstrap race.** Two nodes concurrently initialize the same collection with different initial writer sets. First-write-wins by causal order is well-defined; LWW by clock is not. **See bootstrap section below — \"causal order\" is itself underspecified when nodes don't share an ancestor.**

v2 papers over all four with monotonic-nonce + LWW. That's acceptable for v1/v2 but eventually needs the real fix.

## Constraints inherited from v2

- \`Interface::apply_action\` only receives an \`Action\`; it does NOT see DAG-causal parents (those live one layer up on \`CausalDelta\`).
- There's no per-id rotation history — only the latest writer set per entity is in the index.
- The verifier currently uses *currently stored* writers; v2 commit \`51dda840\` extends this to \`(stored ∪ claimed)\` for stamping but the verification check is still against current state.
- v2 ships a \`signer: Option<PublicKey>\` hint in \`SignatureData\` for O(1) verifier lookup. **Compatibility rule for this epic:** when DAG-causal verification runs, the hint MUST be validated against \`writers_at(id, causal_parents)\` — NOT against current stored writers. Validating against current would reject pre-rotation signatures (the exact scenarios this epic is trying to fix). Spelled out explicitly because this is the kind of detail that gets misremembered mid-implementation.

## What does NOT need this work

[...]

PR #2665, opened 2026-06-03
feat(node): complete DAG-causal rotation — self-log own rotations so writers_at is total

chefsale:
Completes the concurrent-rotation correctness that #2655 explicitly deferred — the "P4" piece of the #2233 DAG-causal rotation epic — uniformly for `Shared` and `SharedMember`.

## Problem
A node only logged rotations it **received** (`maybe_append_rotation_log`, on the apply path); it never logged **its own** (local execute doesn't go through that path). So under concurrent rotations each node held an *asymmetric* log — only the other node's rotation — and `writers_at` resolved a different writer set per node → divergent verification / split-brain:

```
{Alice,Bob}.  node-1 rotates →{Alice,Carol} (R1).  node-2 rotates →{Alice,Bob,Dave} (R2).  (concurrent)
node-1 log = [genesis, R2]   (logged only what it received — missing its own R1)
node-2 log = [genesis, R1]   (missing its own R2)
writers_at(after both):  node-1 → {Alice,Bob,Dave}   node-2 → {Alice,Carol}   ✗ diverge
```
And pre-rotation cuts compounded it: with no genesis recorded, `writers_at` returned `None` → callers fell back to a non-causal "latest writers" guess.

## Fix — every node holds the complete `[genesis, …all rotations]` log
- **Self-log own rotations** (`DeltaStore::add_local_applied_delta`): once a local delta is finalized (id/hlc assigned), append a rotation-log entry for each `Shared` Add (genesis bootstrap) / writer-changing Update — mirroring the receive-path hook. Value-writes that don't change writers are skipped. Uses a **direct datastore write** (`save_rotation_log_direct`) because this runs *outside* the WASM `RUNTIME_ENV` where `MainStorage` is valid.
- **Cold-join floor** (`snapshot.rs`): a snapshot-syncing joiner seeds each anchor's rotation log with the settled writers at its boundary (`seed_rotation_log_genesis_direct`) — its causal floor; it never applies deltas predating the boundary.

With a complete log on every node, `writers_at` is **total** (genesis covers any pre-rotation cut; the existing ADR-0001 merge resolves concurrent rotations order-independently), so the merge path always supplies the causal `effective_writers` and **never reaches the non-causal `resolve_anchor_writers` fallback** — now reserved for local execution (where "current writers" is correct) and legacy pre-P4 anchors. No state-hash change (the rotation log is auxiliary, node-local).

```
After the fix, both nodes log own+received → both = [genesis, R1, R2]
writers_at merges deterministically (order-independent) → identical on every node ✓
```

## Tests
- `add_local_applied_delta_self_logs_own_rotations` (node) — genesis + rotation logged, value-write skipped.
- Convergence follows by composition of existing unit tests: each node logs own+received (this test + the receive path), and `writers_at` is order-independent for concurrent rotations (`writers_at_concurrent_siblings_resolved_by_hlc`) + total via the genesis/snapshot floor (`writers_at_falls_back_to_snapshot_when_entries_unreachable`).
- storage 550 + node 267 lib tests green; fmt clean.

## Remaining / follow-up
A dedicated **cross-node concurrent-rotation merobox e2e** is the integration-level coverage (CI-only; not locally runnable). The existing shared-storage + sync-resilience e2e suites also exercise the rotation paths.

---

> [!NOTE]
> **Medium Risk**
> Changes Shared writer-set history and merge-time `effective_writers` resolution (consensus-adjacent), but writes are idempotent, best-effort on failure, and auxiliary to state hashes; snapshot genesis seeding only runs when no log exists.
> 
> **Overview**
> Completes **DAG-causal rotation (P4)** so every node keeps a **symmetric** per-anchor rotation log: received rotations were already recorded on apply; **locally created** deltas were not, which made `writers_at` disagree under concurrent writer-set changes.
> 
> The node layer adds **direct datastore** read/write helpers for rotation logs (outside WASM `RUNTIME_ENV`), **`self_log_rotations_direct`** (mirrors receive-path rules: genesis / writer-set change only, dedup by `delta_id`), and wires it into **`add_local_applied_delta`** and **`load_persisted_deltas`** (restart backstop). **Snapshot sync** seeds a genesis floor via **`seed_rotation_log_genesis_direct`** when an anchor has no log yet. Storage docs are updated to say merge-path verification should use causal **`writers_at`** / `effective_writers`, with non-causal **`resolve_anchor_writers`** left for local execution and legacy logs.
> 
> A new test asserts local deltas self-log genesis and rotations but skip plain value-writes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fff19ce218e2d950a1ccbf71c1005461bfeece1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/delta_store.rs
+++ b/crates/node/src/delta_store.rs
@@ -997,6 +997,82 @@ pub(crate) fn seed_rotation_log_genesis_direct(
     save_rotation_log_direct(context_client, context_id, entity_id, &log)
 }
 
+/// Self-log the `Shared` rotations carried by a locally-originated delta's
+/// `actions` into their entities' rotation logs, via direct datastore writes.
+///
+/// The receive path records rotations in `Interface::maybe_append_rotation_log`
+/// (inside the WASM sync-apply); a locally-created delta never goes through that
+/// path, so without this a node is missing its OWN rotations and `writers_at`
+/// stays asymmetric across nodes under concurrent rotation. Mirrors the
+/// receive-path logic: append only on a real rotation — bootstrap (no prior
+/// writers) or a changed writer set — never on a plain value-write. Idempotent
+/// (dedup on `delta_id`), so it is safe to run on both the live notify
+/// (`add_local_applied_delta`) and crash/restart recovery
+/// (`load_persisted_deltas`); the latter backstops the case where the live
+/// notify was skipped (DeltaStore not yet up) and the delta is later restored
+/// as already-applied.
+///
+/// Best-effort: a per-entity read/write failure is logged and skipped, leaving
+/// that entry for the next restart's restore to re-attempt — it never aborts
+/// the (already-committed) delta.
+fn self_log_rotations_direct(
+    context_client: &ContextClient,
+    context_id: ContextId,
+    delta_id: [u8; 32],
+    delta_hlc: calimero_storage::logical_clock::HybridTimestamp,
+    actions: &[Action],
+) {
+    for action in actions {
+        let (entity_id, metadata) = match action {
+            Action::Add { id, metadata, .. } | Action::Update { id, metadata, .. } => {
+                (*id, metadata)
+            }
+            Action::DeleteRef { .. } | Action::Compare { .. } => continue,
+        };
+        // Only `Shared` anchors own a rotation log; members/others don't.
+        let StorageType::Shared {
+            writers,
+            signature_data,
+        } = &metadata.storage_type
+        else {
+            continue;
+        };
+        let mut log = match load_rotation_log_direct(context_client, context_id, entity_id) {
+            Ok(existing) => existing.unwrap_or_else(RotationLog::empty),
+            Err(e) => {
+                warn!(?e, %context_id, %entity_id,
+                    "self-log: failed to read rotation log — skipping (restore will retry)");
+                continue;
+            }
+        };
+        // Append only on an actual rotation — bootstrap (no prior writers) OR
+        // the writer set changed. Mirrors `maybe_append_rotation_log`.
+        let prior = log
+            .entries
+            .last()
+            .map(|e| &e.new_writers)
+            .or_else(|| log.snapshot.as_ref().map(|s| &s.writers));
+        if !prior.map_or(true, |p| p != writers) {
+            continue;
+        }
+        // Idempotent on delta_id (one rotation per entity per delta).
+        if log.entries.iter().any(|e| e.delta_id == delta_id) {
+            continue;
+        }
+        log.entries.push(RotationLogEntry {
+            delta_id,
+            delta_hlc,
+            signer: signature_data.as_ref().and_then(|s| s.signer),
+            new_writers: writers.clone(),
+            writers_nonce: signature_data.as_ref().map(|s| s.nonce).unwrap_or(0),
+        });
+        if let Err(e) = save_rotation_log_direct(context_client, context_id, entity_id, &log) {
+            warn!(?e, %context_id, %entity_id,
+                "self-log: failed to write rotation log — leaving for restore retry");
+        }
+    }
+}
+
 /// Whether the anchor entity `anchor` has synced to this node, by a direct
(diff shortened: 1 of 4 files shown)


chefsale:
Thanks — responses to the bot review (commit 2fff19ce):

**Cursor Bugbot**
- 🔴 *Applied check skips self-log* — **fixed.** Extracted the self-log into `self_log_rotations_direct` and now also run it (idempotent, dedup on `delta_id`) for each applied delta in `load_persisted_deltas`. So if the live notify was skipped (DeltaStore not yet up) and the delta is later restored as already-applied, restore backstops it. Received deltas already logged by the receive path dedup-skip.
- 🟡 *Ignores stale nonce* — N/A for a **locally-created** delta: its `payload` is exactly the actions `execute` just applied (fresh, monotonic nonces). The receive-path stale-nonce guard exists for out-of-order *remote* delivery, which doesn't apply to a node's own freshly-produced delta.
- 🟡 *Failure still registers DAG* — best-effort by design: the delta is already committed to storage by `execute` and can't be un-applied, so a self-log write failure can't abort it. It now **self-heals** — the restore backstop re-attempts on next startup.

**meroreviewer**
- 🟡 *Skips signature verification* — intentional: the self-log records the node's **own** committed rotations, which `execute`/`save_raw` already authorized (executor ∈ current writers) and `sign_authorized_actions` already signed before the delta was built. Signature verification is the **receive** path's job (untrusted peer input); the self-log is trusted local output.
- 🟡 *seed_rotation_log_genesis_direct read-check-write race* — not reachable: it runs only on the cold-join snapshot apply, which is I5-gated (local context uninitialized) and single-session, so there are no concurrent rotation writers during it.
- 🟡 *prior vs writers comparison* — it's the identical `BTreeSet<PublicKey>` equality the receive path uses (`stored != writers`); no normalization on either side.
- 💡 *Test `bob_sk` unused for signing* — `bob_sk` derives `bob`'s public key (`pubkey_of(&bob_sk)`) for the `{alice,bob}` writer set; the test asserts the **log** records the writer-set change, which `alice` signing is sufficient to exercise.

Resolving accordingly; happy to reopen any if you'd like a different call.

chefsale:
Second-round bot review — responses:

- **Bugbot 🟡 "Backstop skipped when delta already in DAG":** the self-log backstop's recovery trigger is **process startup** — `load_persisted_deltas` rebuilds the DAG from empty, so every persisted delta is re-processed (not `already_in_dag`) and the idempotent self-log runs, healing any prior failed write. The `already_in_dag` early-continue is the deliberate decode-cost optimization (#2238) for *repeat in-process* calls (sync recovery), where the delta and its log are already handled. The precondition is narrow — a *transient single-key* rotation-log write failure while the delta-record write succeeded (a persistent datastore failure would also fail the delta persist → nothing to restore) — it restart-heals, and meanwhile the originator's *local* resolution stays correct (the index reflects current writers; only a cross-cut merge resolution on another node would be affected). I'd rather not regress the decode-skip for this; happy to move the backstop ahead of the skip (decode cost on the recovery path) if you prefer in-process healing.
- **meroreviewer 🟡 "rotation entry appended without verifying signature" (restore path):** restore replays the node's **own persisted, already-committed** history — entries that were signature-verified at receive time or authored+authorized locally (`save_raw` enforces executor ∈ writers) when first applied. Re-deriving the log from them needs no re-verification (trusted local history, not untrusted peer input); on-disk tampering is outside the threat model. The self-log is idempotent, so it only re-confirms entries consistent with applied state.
- **meroreviewer 📝 "bob_sk dead code":** `bob_sk` is used — `bob = pubkey_of(&bob_sk)` derives the second writer's key for the `{alice, bob}` set, and the test builds warning-free. The test asserts the **log** records the writer-set change, which alice-signing alone exercises; `bob` only needs to be in the set.

Resolving; reopen any if you'd prefer a different call.

PR #2666, opened 2026-06-03
feat(dag): enable DAG compaction by default (#2026)

chefsale:
## What

Enables DAG compaction (#2026) by default. Follow-up to #2664, which shipped the machinery disabled.

Flips `DagCompactionConfig::enabled` to `true` in both the `Default` impl and the serde field default, so a node with no `[dag_compaction]` section compacts with conservative defaults. Operators opt out with:

```toml
[dag_compaction]
enabled = false
```

## Why it's safe to default-on

Deltas are an optimization, not a convergence requirement. A peer that requests a pruned delta gets `DeltaNotFound`, and the next sync round reconciles via **HashComparison** — already the protocol selected for a diverged initialized node — which converges current state without consulting the delta log. Nothing prunes within the recent retain window, so cheap incremental delta catch-up still covers normal reconnect gaps.

Thresholds are unchanged and conservative:
- `min_deltas_before_compact = 10000` (= `MAX_TOPOLOGY_ENTRIES`)
- `retain_recent_count = 1000` (≤ `MAX_DELTA_QUERY_LIMIT`)
- `check_interval = 1h`, with the overlap guard + startup sweep from #2664.

## Tests

- `default_is_enabled_with_sane_thresholds`, `absent_fields_fall_back_to_defaults` (now expect enabled), and a new `explicit_disable_is_respected`.
- `cargo fmt --check` clean; node + config build.

---

> [!NOTE]
> **Medium Risk**
> Default-on changes on-disk delta retention for upgraded nodes without config changes; invalid enabled settings now abort startup instead of only logging a warning.
> 
> **Overview**
> Turns on **DAG compaction by default** so nodes without a `[dag_compaction]` section (or with an empty/default block) run periodic pruning with the existing conservative thresholds. Operators opt out with `enabled = false`.
> 
> Docs now explain the default-on rationale (pruned deltas fall back to **HashComparison** sync) and call out that **upgrades** from older builds will start compacting on restart unless they disable it explicitly.
> 
> **Node startup** no longer silently skips compaction when `enabled` is true but thresholds are invalid—it **errors at boot** via `eyre::ensure!` so misconfigured `retain_recent_count` / `check_interval` cannot leave delta growth unbounded without notice.
> 
> Tests expect enabled defaults and cover explicit disable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d61f917687ec2834ea924d18488d7d867937b95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -70,8 +70,9 @@ pub struct ConfigFile {
     pub tee: Option<TeeConfig>,
 
     /// DAG compaction (`[dag_compaction]`) — bounds on-disk delta-log growth
-    /// by pruning history older than a checkpoint. Disabled by default;
-    /// absent section falls back to [`DagCompactionConfig::default`].
+    /// by pruning history older than a recent retain window. Enabled by
+    /// default; an absent section falls back to [`DagCompactionConfig::default`]
+    /// (set `enabled = false` to opt out).
     #[serde(default)]
     pub dag_compaction: DagCompactionConfig,
 }
--- a/crates/node/primitives/src/dag_compaction.rs
+++ b/crates/node/primitives/src/dag_compaction.rs
@@ -43,6 +43,11 @@ pub const DEFAULT_CHECK_INTERVAL_SECS: u64 = 3_600;
 /// selected for a diverged initialized node — which converges current state
 /// without consulting the delta log. Operators can still set `enabled = false`
 /// to opt out.
+///
+/// **Upgrade note:** a node upgrading from a build where compaction was
+/// disabled-by-default, and whose config has no `[dag_compaction]` section,
+/// will begin compacting automatically on the next restart. Set
+/// `enabled = false` to preserve the old behavior.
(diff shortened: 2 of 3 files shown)


chefsale:
Good calls — both addressed in 1d61f917.

- **Silent-skip on invalid-but-enabled config** (warning + suggestion 2): agreed, especially now that it's default-on. Changed `run.rs` from a `warn!` to a hard startup error (`eyre::ensure!`): an `enabled` config with `retain_recent_count >= min_deltas_before_compact` or a zero `check_interval` now fails fast instead of silently leaving growth unbounded. The default config is always valid, so this only trips a deliberate misconfiguration.
- **Migration note** (suggestion 1): added to the config doc comment — a node with no `[dag_compaction]` section begins compacting on the next restart after upgrading from a disabled-default build, and `enabled = false` preserves the old behavior.

PR #2667, opened 2026-06-03
refactor(context): consolidate hot caches into BoundedCache + bound namespace DAG history

### LINKED ISSUE #2638: namespace_dags in-memory DAG cache is never bounded or evicted
## Summary

`ContextManager::namespace_dags` (`crates/context/src/lib.rs:331`) is an unbounded `HashMap<[u8;32], Arc<Mutex<DagStore<SignedNamespaceOp>>>>`. Unlike the `contexts` / `applications` / `modules` caches — which are all size-capped — this map only ever grows: `get_or_create_namespace_dag` (`:541-549`) does `entry().or_insert_with(...)` and nothing ever removes an entry. Each `DagStore.deltas` additionally retains **every** applied governance op for that namespace, so memory grows with both namespace count and op history.

## Details

```rust
fn get_or_create_namespace_dag(&mut self, namespace_id: &[u8; 32])
    -> Arc<tokio::sync::Mutex<DagStore<SignedNamespaceOp>>>
{
    self.namespace_dags
        .entry(*namespace_id)
        .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(DagStore::new([0u8; 32]))))
        .clone()
}
```

- No size cap, no idle/TTL eviction, no GC. The companion caps added for `contexts`/`applications`/`modules` (size-capped with `Arc::strong_count`-gated eviction) were not applied here.
- Inner `DagStore.deltas` accumulates every applied op indefinitely — even for namespaces no longer active — so a single long-lived namespace grows without bound over its op history.
- A node that participates in (or merely observes/admits ops for) many namespaces accumulates one resident DAG per namespace for the process lifetime.

## Impact

Unbounded memory growth proportional to (distinct namespaces ever touched) × (ops per namespace). Slow OOM / DoS on long-running nodes, especially those exposed to many namespaces.

## Suggested fix

- Apply the same size-cap + eviction pattern used for the other caches (evict gated on `Arc::strong_count == 1` so owned guards aren't dropped underneath callers), and/or idle-eviction by last-access.
- Prune already-applied deltas from `DagStore.deltas` once they are durably persisted / no longer needed for DAG-head reconciliation, so retained DAG memory is bounded by the live frontier rather than full history.

## References

- `crates/context/src/lib.rs:331` (unbounded `namespace_dags` field)
- `crates/context/src/lib.rs:541-549` (`get_or_create_namespace_dag` — insert-only)

_Found during security review (finding #34). Companion to the contexts/applications/modules cap work._

chefsale:
## Summary

Closes #2638.

`ContextManager` kept **four** in-memory hot caches with **three** different hand-rolled size-cap/eviction implementations across two files, plus a fourth map (`namespace_dags`) that was **never bounded at all** — it grew for the whole process lifetime, a slow OOM on nodes that touch many namespaces (#2638).

This consolidates all four onto one abstraction and then bounds the one remaining unbounded dimension (per-namespace DAG history).

### Part 1 — cache consolidation
- New `BoundedCache<K, V>` (`crates/context/src/cache.rs`) + an `Evictable` trait. The only axis the caches differed on was whether an entry may be dropped while a caller holds a live handle; `Evictable::is_idle` captures exactly that — default `true` for plain caches (`applications`, `modules`), `Arc::strong_count == 1` for lock-gated ones (`contexts`, `namespace_dags`). All cap / eviction / over-cap-deferral logic now lives in **one** place, tested once.
- Migrated `contexts`, `applications`, `modules`, and `namespace_dags` onto it. The last one is now size-capped (`MAX_CACHED_NAMESPACE_DAGS = 1024`) with the same `strong_count`-gated eviction as `contexts`, closing the #2638 leak. `MAX_CACHED_MODULES` moved out of the execute handler into `lib.rs` next to the other caps.
- The three bespoke helpers (`evict_idle_context_if_full`, `evict_application_if_full`, the inline `pop_first`) are gone.

### Part 2 — bound per-namespace governance DAG history
Even a single long-lived namespace grew its in-memory `DagStore<SignedNamespaceOp>.deltas` without bound. After an applied op advances the frontier, prune applied history back to a recent window via `DagStore::prune_to_recent` (the primitive added in #2664).

**Why this is safe and lossless for peers:** applied governance ops are durably persisted as `NamespaceGovOp` rows, and the backfill responder serves peers **from RocksDB, not this in-memory DAG**. The in-memory DAG is already transient (rebuilt from gossip/backfill after a restart). So pruned ids are **discarded, not deleted from disk** — pruning only bounds the in-memory footprint and merely accelerates what a restart already does. Only pathologically long histories prune (threshold 8192 → retain 4096); the retain window is many backfill batches deep (`MAX_BACKFILL_OPS` is 500), so a re-delivered recent op still parents onto a retained delta instead of re-pending on a pruned ancestor.

Built on top of #2664 (DAG compaction), which merged the `prune_to_recent` / `delta_count` primitive this reuses.

## Testing
- `cargo test -p calimero-context --lib` — 78 pass (incl. 13 `BoundedCache`/cache-wiring tests)
- `cargo test -p calimero-dag --lib` — 55 pass (incl. a new `test_prune_to_recent_never_prunes_pending`)
- `cargo check --workspace` green; `cargo fmt --check` and `clippy` clean on all touched files

---

> [!NOTE]
> **Medium Risk**
> Changes eviction and pruning on governance/context hot paths where lock-gating must stay correct; mitigated by existing Evictable semantics, tests, and authoritative datastore/backfill.
> 
> **Overview**
> Introduces shared **`BoundedCache`** + **`Evictable`** and moves **`ContextManager`** hot maps (`contexts`, `applications`, `modules`, **`namespace_dags`**) onto it, replacing three ad-hoc eviction paths and capping **`namespace_dags`** (1024) with the same lock-aware eviction as contexts.
> 
> After a successful namespace governance apply, trims **in-memory** applied DAG history when applied count exceeds **8192**, keeping **4096** recent deltas; durable ops and peer backfill stay on RocksDB.
> 
> Adds unit tests for cache behavior and confirms **`prune_to_recent`** does not drop pending deltas.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 758826bb631b1e94b94df57cf00633c848b42c22. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/cache.rs
+++ b/crates/context/src/cache.rs
@@ -1,0 +1,306 @@
+//! A single size-capped in-memory cache abstraction shared by every hot cache
+//! `ContextManager` keeps (`contexts`, `applications`, `modules`,
+//! `namespace_dags`).
+//!
+//! Before this module each cache hand-rolled its own cap + eviction: two
+//! free functions (`evict_idle_context_if_full`, `evict_application_if_full`),
+//! an inline `pop_first` in the execute handler, and — for `namespace_dags` —
+//! nothing at all (it grew unbounded; see the `namespace_dags` issue this
+//! consolidation closes). The three live implementations only ever differed on
+//! one axis: whether an entry may be evicted while a caller holds a live handle
+//! to it. That axis is captured by [`Evictable`]; everything else (the cap, the
+//! fetch-before-evict insert discipline, the "all entries live → defer and
+//! stay over-cap" behaviour) lives here once.
+//!
+//! The datastore is always authoritative, so an evicted entry costs at most a
+//! re-fetch on next access — the cap is a safety valve against unbounded growth
+//! on long-running nodes, not a correctness boundary.
+
+use std::collections::{btree_map, BTreeMap};
+use std::fmt::Debug;
+
+/// Whether a cached value may be dropped right now.
+///
+/// The default (`true`) suits a *plain* cache whose values are pure clones of
+/// datastore state — evicting one is always safe. A cache whose entries hand
+/// out a live handle (e.g. an `Arc<Mutex<_>>` guard held across an async
+/// operation) overrides this so the entry is only evictable once no caller
+/// holds that handle; see the `ContextMeta` / namespace-DAG impls in `lib.rs`
+/// for why evicting a live entry would let two operations serialize on
+/// different mutexes and corrupt state.
+pub(crate) trait Evictable {
+    /// `true` when nothing outside the cache references this entry and it is
+    /// safe to evict. Always `true` for plain caches.
+    fn is_idle(&self) -> bool {
+        true
+    }
+}
+
+/// A `BTreeMap`-backed cache bounded to `cap` entries.
+///
+/// Eviction runs only just before inserting a *new* key (re-touching an
+/// existing key never evicts) and removes the lowest-keyed [`Evictable::is_idle`]
+/// entry. Victim selection is by key order, not true LRU — deliberately:
+/// `ContextId`/`ApplicationId` keys are hashes, so "lowest key" is effectively
+/// arbitrary, and a wrong guess only costs a re-fetch. Upgrading to LRU is a
+/// possible follow-up if profiling ever shows churn on a hot entry.
+#[derive(Debug)]
+pub(crate) struct BoundedCache<K, V> {
+    map: BTreeMap<K, V>,
+    cap: usize,
+    /// Stable label for tracing (`contexts`, `applications`, …).
+    name: &'static str,
+}
+
+impl<K: Ord + Clone + Debug, V: Evictable> BoundedCache<K, V> {
+    /// A new empty cache holding at most `cap` entries, tagged `name` in logs.
+    pub(crate) fn new(cap: usize, name: &'static str) -> Self {
+        Self {
+            map: BTreeMap::new(),
+            cap,
+            name,
+        }
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.map.len()
+    }
+
+    pub(crate) fn contains_key(&self, key: &K) -> bool {
+        self.map.contains_key(key)
+    }
+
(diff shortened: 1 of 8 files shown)


chefsale:
Thanks for the review — addressed in 758826bb. Triage of the 6 findings:

**Fixed (real):**
- **Pruning gated on `delta_count()` which includes pending** (the important one): switched the trigger to the *applied* count (`dag.stats().applied_deltas`). A large pending backlog (partition / missing parents) could otherwise keep the total above the threshold and fire `prune_to_recent` on every applied op while pruning nothing prunable — that's also the real mechanism behind the "prunes on every Applied op" finding.
- **Raw `entry()` bypasses the cap** (both reviewers): the cap is now enforced *inside* `entry()` — a new key evicts one idle entry first, an existing key never evicts. Removed the doc-only contract and the now-redundant `evict_if_full()` in `create_context`. Added a test for new-vs-existing entry capping.
- **`get_or_insert_with` double-lookup + extra clone on miss**: now materialises through the entry API in a single lookup with the key moved.

**Won't-fix, with reasoning:**
- **"Pruning fires on every Applied op (O(n))"**: the per-apply check is `applied_deltas` (O(1)); `prune_to_recent` only runs once the applied window climbs past the threshold and then drops it back to the retain window, so prunes are ~`RETAIN..THRESHOLD` applies apart — a built-in hysteresis band. The legitimate failure mode here was the pending-count issue above, now fixed.
- **`insert_new` debug_assert only fires in debug**: this is an intentional precondition contract (all three call sites guard on absence first). The replace-safe `insert()` already exists for the may-exist case; hardening `insert_new` would either reintroduce a double-lookup or change its "new" semantics.

Builds clean; context 78 lib + 14 cache tests green; fmt + clippy clean.

PR #2670, opened 2026-06-03
fix(node): hold context lock across inbound delta apply + dag_heads commit

chefsale:
## Problem

The `shared-storage` e2e ([run 26894380526](https://github.com/calimero-network/core/actions/runs/26894380526/job/79330340913)) hangs at **"Wait for post-rotation value to sync"**: after node-1 rotates the writer set to node-2 and node-2 writes, the nodes never converge (node-1 stays `6twez…`, node-2 `2KgT…`). Node-1 rejects node-2's write at merge with a fatal `InvalidSignature`.

The rejection is *correct* — the write is genuinely unauthorized in DAG-causal terms. A writer-set rotation is **hash-neutral**, so an inbound rotation delta's two effects straddle two separate critical sections:

1. The WASM apply (`ContextStorageApplier::apply`) makes the rotated-in writer authoritative **in storage**, then releases the per-context execution lock when it returns.
2. `add_delta_internal` commits the new `dag_heads` **afterwards**, under the DAG write lock.

A local write that runs in that window reads the **pre-rotation** DAG heads (`get_or_fetch_context` snapshots `dag_heads` *before* the handler takes the lock) and authors its delta off the stale head. Its parents exclude the rotation, so on every peer `writers_at(parents)` resolves the *old* writer set → `InvalidSignature` → permanent split-brain.

Observed on node-2 (sub-ms):

```
.378577  rotation apply EXEC-DONE     → storage updated, node-2 is now a writer
.378714  set_shared EXEC-START         → reads dag_heads = [15,75,234]  (stale!)
.378827  rotation HEAD-COMMIT [213,162] → heads advance — too late
```

`set_shared`'s delta parents off `[15,75,234]` (a sibling of the rotation), not the rotation `[213,162]`.

## Fix

Keep the per-context lock held across **both** steps, with no lock-order inversion (order stays `DAG write lock → context lock`):

- The inbound apply runs with `ContextAtomic::Lock` when `add_delta_internal` arms `retain_apply_lock`, and hands the guard back via `apply_lock_slot`. `add_delta_internal` holds it across the `dag_heads` commit and releases it (RAII) on every exit path. Cascaded buffered-child applies reuse the guard via `ContextAtomic::Held` (the lock isn't re-entrant). The head persist is direct datastore I/O with no executor re-entry → no deadlock.
- `internal_execute` re-reads `dag_heads` from the authoritative store under the held lock before selecting parents.

A local write now blocks until an in-flight inbound apply has committed its heads, then parents off it → `writers_at` resolves the new writer set → converges. **No behavior change when unarmed** (every other `apply()` caller).

## Testing

- `cargo build`, `cargo clippy --lib`, `cargo fmt --check`, `cargo test -p calimero-node --lib delta_store` (23 passed) all green.
- Authoritative regression gate: the `shared-storage` merobox e2e (`apps/kv-store-with-shared-storage`), which deterministically failed on the post-rotation write before this change. A deterministic Rust repro isn't feasible (the race needs the live executor actor + WASM + sub-ms interleaving); the e2e covers it.

## Follow-up (not in this PR)

The workflow's "Wait for rotation to sync" is a no-op on a hash-neutral rotation (root hashes already match). Worth hardening the e2e to confirm node-2 observes itself as a writer before writing, to also cover the rarer "rotation not yet gossiped" timing.

---

> [!NOTE]
> **High Risk**
> Changes concurrency and locking around DAG head selection and writer-set rotation; mistakes could deadlock or worsen split-brain, though unarmed apply paths are unchanged.
> 
> **Overview**
> Closes a **DAG split-brain** race where a **hash-neutral writer-set rotation** could be visible in storage before **`dag_heads`** advanced, so a concurrent local write parented off stale heads and peers rejected it as **`InvalidSignature`**.
> 
> Inbound sync applies now **retain the per-context execution lock** from WASM apply through **`dag_heads` commit** (`ContextAtomic::Lock` / `Held` via `retain_apply_lock` and `apply_lock_slot` in single-delta and batch paths), releasing it before recursive paths that would self-deadlock. **`internal_execute`** re-reads **`dag_heads`** from the store under that lock before choosing parents for locally authored deltas.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e88fbe056152e358b8e4e7e1f19b27d7348689b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/execute/mod.rs
+++ b/crates/context/src/handlers/execute/mod.rs
@@ -1567,6 +1567,25 @@ async fn internal_execute(
                 outcome.artifact = new_artifact;
             }
 
+            // Refresh the DAG heads from the authoritative store before
+            // choosing this write's parents. `context` (hence `dag_heads`) was
+            // snapshotted in `get_or_fetch_context` BEFORE this handler took
+            // the per-context lock, so an inbound delta that committed new
+            // heads while we waited for the lock would be missed here.
+            // Authoring on the stale head forks the DAG: the new delta's
+            // parents exclude an already-applied ancestor — e.g. a writer-set
+            // rotation the executor has locally applied — and every peer then
+            // rejects it (`writers_at(parents)` resolves the pre-rotation set),
+            // a permanent split-brain. The inbound apply now holds this same
+            // lock across its `dag_heads` commit (see
+            // `DeltaStore::add_delta_internal`), so once we hold the guard the
+            // persisted heads are current.
+            if let Ok(Some(meta)) = store.handle().get(&key::ContextMeta::new(context.id)) {
+                if context.dag_heads != meta.dag_heads {
+                    context.dag_heads = meta.dag_heads;
+                }
+            }
+
             // Use current DAG heads as parents, verifying they exist in RocksDB
             let parents = if context.dag_heads.is_empty() {
                 // Genesis case: parent is the zero hash
--- a/crates/node/src/delta_store.rs
+++ b/crates/node/src/delta_store.rs
@@ -14,6 +14,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use calimero_context_client::client::ContextClient;
+use calimero_context_client::{ContextAtomic, ContextAtomicKey};
 use calimero_dag::{
     ApplyError, CausalDelta, DagStore as CoreDagStore, DeltaApplier, PendingStats,
     MAX_DELTA_QUERY_LIMIT,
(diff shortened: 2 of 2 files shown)


chefsale:
Thanks — all four findings addressed in `e88fbe05`.

**🔴 Cursor (high) — context lock held during orphan re-injection → self-deadlock.** Correct and the most important catch. `add_delta_internal` recurses into itself to re-inject buffered orphan members, and the retained guard living to function end meant the recursion's inbound apply (`ContextAtomic::Lock`) blocked on the lock we held. Fixed: the guard is now `drop`ped explicitly right after the `dag_heads` persist and **before** the re-injection loop. It only needs to cover the head commit, which is done by then; the earlier `?`/`bail!` paths still release it via scope exit.

**🟡 meroreviewer — `retain_apply_lock` not armed in `add_deltas_batch`.** Valid coverage gap — the catchup/flush path had the same race. Armed `retain_apply_lock` around the batch apply loop and held the retained guard across the final persist. The batch path has no orphan re-injection, so it holds to the commit without deadlock risk.

**🟡 meroreviewer — slot empty on the WASM-timeout re-await / cascaded child.** Not reachable: `dag.add_delta` processes the primary and cascaded children strictly **sequentially** on one task (`apply_pending` runs only after the current apply returns), so no concurrent `apply()` can observe the empty slot and issue a second `Lock`. Documented this invariant in `apply()`.

**💡 meroreviewer — `Acquire`/`Release` rationale.** Added a comment: same-task under the `dag` write lock; the ordering is defensive clarity since the WASM body runs on a `spawn_blocking` thread inside `execute`.

Build / clippy `--lib` / `fmt --check` / `cargo test -p calimero-node --lib delta_store` (23 passed) all green. The `shared-storage` merobox e2e remains the end-to-end gate.

Issue #2468, opened 2026-05-23
test(e2e): add sync-resilience-lossy-link workflow (needs custom merod+iproute2 image)


chefsale:
Follow-up to #2466.

## What

Add back `sync-resilience-lossy-link.yml` as a fifth workflow under `apps/scaffolding-e2e/workflows/` to exercise the degraded-but-alive channel that real flaky Wi-Fi / congested cellular looks like in production. The workflow was prototyped against PR #2466 but had to be dropped from the initial landing because of an image prerequisite (see Why).

## Test design

Burst-write 20 keys on node-1 with no inter-write wait, then inject 90% packet loss on node-2 for 20s via `inject_network_fault` (calimero-network/merobox#249 step), then assert convergence within 90s after the loss clears.

Three assertions of decreasing strictness:

1. **Diagnostic read (non-fatal)** — read \`burst_20\` on node-2 right after the fault clears. Value is captured in CI logs. We do NOT assert; at 90% loss for 20s the math is too close to either outcome to assert strictly without flake.
2. **Strict convergence** — \`wait_for_sync\` with a 90s timeout must succeed. Primary assertion. A failure here means recovery doesn't actually work, not just that the lossy window had no observable impact.
3. **Strict per-key value check** — read all 20 keys on node-2, assert each contains its expected value. Catches the class of regression where the DAG converges but the per-key state doesn't (CRDT bug dropping events under retransmission).

A draft of the workflow is in commit \`4cbd9985\` of \`feat/relay-recovery-workflows\` (see PR #2466 history); it has full inline rationale in the header comment.

## Why we couldn't land it in #2466

The stock \`ghcr.io/calimero-network/merod:edge\` image does not ship \`tc\` (iproute2):

\`\`\`
✗ Failed to apply netem on sync-resil-node-2: OCI runtime exec failed:
  exec: \"tc\": executable file not found in \$PATH —
  the container image does not ship \`tc\` (iproute2). The stock merod image
  is one of these; build a thin image on top with \`apt-get install -y iproute2\`
  and point \`nodes.image\` at it.
\`\`\`

Merobox's own CI explicitly skips its \`fault-injection-tc\` workflow for the same reason.

## Plan

1. Add a \`ghcr.io/calimero-network/merod-fault:edge\` build step somewhere in core CI (or a separate workflow under \`.github/workflows/build-merod-fault-image.yml\`) that layers iproute2 on top of the stock image.
[...]

PR #2671, opened 2026-06-03
test(e2e): cross-node concurrent writer-rotation convergence (#2668)

### LINKED ISSUE #2668: test: cross-node concurrent-rotation convergence e2e (merobox)
Follow-up to #2665 (DAG-causal rotation completion / #2233 epic).

#2665 made each node self-log its **own** rotations so `writers_at` is total and symmetric across nodes — the fix for concurrent-rotation divergence. It's covered by unit/node tests (self-log + `writers_at` order-independence/totality), but the one angle not exercised locally is a **multi-node, genuinely-concurrent rotation** converging end-to-end.

## Ask
Add a dedicated merobox workflow (2–3 nodes) that:
1. Creates a context with a `SharedStorage` writer set `{A, B}` (and ideally a guarded collection with a few member entries).
2. Has **two nodes rotate the writer set concurrently** (e.g. node-1 → `{A, C}`, node-2 → `{A, B, D}`) without seeing each other first (brief partition / simultaneous execute).
3. Heals the partition and asserts **all nodes converge** to the same resolved writer set and the same root hash, and that member entries written before/around the rotations converge consistently.
4. (Bonus) asserts a rotated-out writer's subsequent write is rejected on every node (retroactive revocation holds under concurrency).

This is belt-and-suspenders integration coverage for the convergence guarantee #2665 establishes by composition. Today it's only indirectly touched by the `shared-storage` + `sync-resilience` suites.

Relates: #2665, #2655, #2588, #2233.

🤖 Filed via Claude Code as a tracked follow-up from #2665.

chefsale:
Closes #2668.

Follow-up to #2665 (DAG-causal rotation completion / #2233 epic). #2665 made each node self-log its **own** rotations so `writers_at` is total and symmetric across nodes — the fix for concurrent-rotation divergence. That's covered by unit/node tests; the one angle not exercised there is a **multi-node, genuinely-concurrent rotation** converging end-to-end over real sync. This adds that integration test.

## What's here

**New merobox workflow** `apps/scaffolding-e2e/workflows/shared-storage-concurrent-rotation-converge.yml` (3 nodes; identities A=node-1, B=node-2, C=node-3):

1. Context starts with SharedStorage writer set `{A, B}` plus a few KV "member entries".
2. node-2 is partitioned off the bridge. While split, **two nodes rotate concurrently** without seeing each other: node-1 → `{A, C}` (drops B), node-2 → `{B, C}` (drops A).
3. Heal. `wait_for_sync` across all three converges `contextStateHash` — **this is the root-hash convergence assertion** — with both rotations merged.
4. node-3 (C) issues a **settling rotation** to the definitive `{A, C}`.
5. Asserts **every node** resolves the same writer set (A, C writers; **B retroactively revoked**), member entries written before/around the rotations converge, and (bonus) the rotated-out writer's subsequent write is **rejected** while a current writer's write converges.

**App method** `shared_rotate_writers` — full writer-set replacement (so a test can *drop* a writer; the existing `shared_add_writer` can only union one in).

**CI** — registered in the `e2e-rust-apps` matrix.

## Design notes

- **Why C is the settler:** C is a member of *both* concurrent sets, so whichever rotation wins the causal HLC tiebreak, `writers_at` resolves to a set containing C on every node — node-3's settling rotation is authorized everywhere.
- **Why a settling rotation is needed for a deterministic cross-node writer-set assertion:** the app-facing `SharedStorage::writers()` reads the rotation log's most-recently-*appended* entry (per-node insertion order, not causal order — full local causal resolution is the deferred P4 item; the causal `writers_at` is the merge-path security boundary). Two concurrent rotations can be appended in different orders on different nodes, so reading `writers()` immediately after the heal could differ across nodes even though the logs hold the same entries. A single causally-latest rotation is last on every node, collapsing that ambiguity. The pre-settle convergence #2665 actually guarantees — that both rotations propagate, authenticate, and reconcile the storage root — is asserted by the post-heal `wait_for_sync`.
- **Why deterministic:** the partition only has to make the two rotations invisible to each other while split; after heal, assertions depend solely on final convergence, not on any race against TCP-buffer drain.

Relates: #2665, #2655, #2588, #2233.

---

> [!NOTE]
> **Low Risk**
> Test-only changes (e2e workflow, scaffolding helper, unit tests); no production sync or auth logic modified in this diff.
> 
> **Overview**
> Adds **end-to-end coverage** for **SharedStorage** when two nodes rotate the writer set **concurrently** and then sync: a new merobox workflow drives a 3-node partition (disconnect/connect), concurrent rotations to `{A,C}` vs `{B,C}`, heal + **`wait_for_sync`**, a **settling rotation** to `{A,C}`, and asserts identical **`shared_is_writer`** on all nodes, KV convergence, **revoked writer rejection**, and final shared value.
> 
> The scaffolding app gains **`shared_rotate_writers`** (full writer-set replacement, so tests can **drop** writers). **CI** registers the workflow in **`e2e-rust-apps`**.
> 
> **Node unit tests** in **`rotation_log_reader.rs`** assert **`writers_at`** is **insertion-order invariant** under concurrent rotations and that a **settling rotation** converges to `{A,C}` with **B revoked** — mirroring the e2e scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4fe2464b657179f4a1a7d980a83fbcc1bc2a8e68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -209,27 +209,9 @@ jobs:
           - workflow: hashcomparison-shared-user-reconcile
             file: workflows/hashcomparison-shared-user-reconcile.yml
             app: scaffolding-e2e
-          # The shared-storage-concurrent-rotation-converge workflow exists at
-          # `apps/scaffolding-e2e/workflows/shared-storage-concurrent-rotation-converge.yml`
-          # but is intentionally NOT registered in the matrix yet — same class
-          # of limitation as member-removed-partition-window-reconcile below.
-          # A genuinely-concurrent writer-set rotation needs node-1 and node-2
-          # to be RPC-reachable (so each can be told to rotate) yet unable to
-          # exchange libp2p traffic (so neither sees the other's rotation), and
-          # no current merobox primitive provides that:
-          #   * disconnect_node (bridge detach) also severs the published-port
-          #     DNAT, so a partitioned node can't be `call`ed at all;
-          #   * pause_container freezes the process but the kernel keeps
-          #     buffering inbound gossip, so on unpause the node drains the
-          #     competing rotation before it can rotate (verified: a DELTA_APPLY
-          #     fires immediately on unpause), turning the two rotations causal;
-          #   * stop/start re-syncs the competing rotation on boot.
-          # This needs a surgical peer-pair partition (iptables/tc drop on the
-          # libp2p swarm port to a specific peer IP, keeping the RPC port) —
-          # the same primitive member-removed-partition-window-reconcile wants.
-          # The concurrent-rotation convergence is covered by the unit/node
-          # tests #2665 added; the workflow is kept as the e2e spec + local
-          # repro for when that primitive lands.
+          - workflow: shared-storage-concurrent-rotation-converge
+            file: workflows/shared-storage-concurrent-rotation-converge.yml
+            app: scaffolding-e2e
           - workflow: group-subgroup-visibility-inheritance
             file: workflows/group-subgroup-visibility-inheritance.yml
             app: scaffolding-e2e
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -209,6 +209,9 @@ jobs:
           - workflow: hashcomparison-shared-user-reconcile
             file: workflows/hashcomparison-shared-user-reconcile.yml
             app: scaffolding-e2e
+          - workflow: shared-storage-concurrent-rotation-converge
+            file: workflows/shared-storage-concurrent-rotation-converge.yml
+            app: scaffolding-e2e
           - workflow: group-subgroup-visibility-inheritance
             file: workflows/group-subgroup-visibility-inheritance.yml
             app: scaffolding-e2e
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -209,9 +209,27 @@ jobs:
           - workflow: hashcomparison-shared-user-reconcile
             file: workflows/hashcomparison-shared-user-reconcile.yml
             app: scaffolding-e2e
-          - workflow: shared-storage-concurrent-rotation-converge
-            file: workflows/shared-storage-concurrent-rotation-converge.yml
-            app: scaffolding-e2e
+          # The shared-storage-concurrent-rotation-converge workflow exists at
+          # `apps/scaffolding-e2e/workflows/shared-storage-concurrent-rotation-converge.yml`
+          # but is intentionally NOT registered in the matrix yet — same class
+          # of limitation as member-removed-partition-window-reconcile below.
+          # A genuinely-concurrent writer-set rotation needs node-1 and node-2
+          # to be RPC-reachable (so each can be told to rotate) yet unable to
+          # exchange libp2p traffic (so neither sees the other's rotation), and
+          # no current merobox primitive provides that:
+          #   * disconnect_node (bridge detach) also severs the published-port
+          #     DNAT, so a partitioned node can't be `call`ed at all;
+          #   * pause_container freezes the process but the kernel keeps
+          #     buffering inbound gossip, so on unpause the node drains the
+          #     competing rotation before it can rotate (verified: a DELTA_APPLY
+          #     fires immediately on unpause), turning the two rotations causal;
+          #   * stop/start re-syncs the competing rotation on boot.
+          # This needs a surgical peer-pair partition (iptables/tc drop on the
+          # libp2p swarm port to a specific peer IP, keeping the RPC port) —
+          # the same primitive member-removed-partition-window-reconcile wants.
+          # The concurrent-rotation convergence is covered by the unit/node
+          # tests #2665 added; the workflow is kept as the e2e spec + local
+          # repro for when that primitive lands.
           - workflow: group-subgroup-visibility-inheritance
             file: workflows/group-subgroup-visibility-inheritance.yml
             app: scaffolding-e2e
--- a/apps/scaffolding-e2e/src/lib.rs
+++ b/apps/scaffolding-e2e/src/lib.rs
@@ -19,7 +19,7 @@
 
 #![allow(clippy::len_without_is_empty)]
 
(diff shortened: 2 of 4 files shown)


chefsale:
Pushed fixes addressing the CI failure and the review comments.

**Root cause of the CI failure** (`Node-2 rotates writer set to {B, C}` → `execute_function failed`): the first revision partitioned node-2 off the Docker bridge and then tried to call its RPC. That can't work — merobox reaches a node's admin RPC over the same Docker network as its gossip (published-port DNAT to the bridge IP), so a bridge-detached node is unreachable for *calls* too; the rotate RPC never arrived (confirmed: node-2's container log has no `shared_rotate_writers` request).

Replaced the bridge partition with a **race-free `docker pause` choreography** that freezes only the nodes holding the competing rotation and only ever issues a `call` against an un-paused node:
1. pause B; node-1 (with live node-3) rotates → `{A, C}`; B never sees it
2. pause A and C (no live node now holds `{A, C}`)
3. unpause B; B is still at the pristine head and its only peers are frozen, so it can't fetch `{A, C}`; B rotates → `{B, C}`
4. unpause everyone; `wait_for_sync` (trigger_sync) drives the exchange

At each rotation, every node holding the competing rotation is frozen, so concurrency is guaranteed with no race window. This mirrors the pause usage already in the committed `sync-resilience-peer-pause` workflows.

**Review comments:**
- _Final value only asserted on node-2/node-3_ (meroreviewer) — ✅ added a `shared_get` + assertion on node-1.
- _During-split writes use plain `set`, not `shared_set`_ (meroreviewer, ai-reviewer) — ✅ added a `shared_set` by node-2 during its split window. B is still a writer in its own `{B, C}` view, so it's a Shared-*value* write concurrent with node-1's branch; the post-heal convergence to node-1's causally-latest `after-settle` now proves B's concurrent value loses LWW and survives nowhere. The plain `set` writes stay too — they intentionally test KV data convergence around the rotation window, not authorization.
- _`m2` write races the `shared_add_writer` rotation_ (ai-reviewer) — non-issue: `m2` uses the unrestricted `set` (plain KV map), which performs no writer-set check, so propagation order of the writer-set rotation is irrelevant to it.

chefsale:
**Update: parked from CI on a merobox primitive gap (same as `member-removed-partition-window-reconcile`).**

Iterating on this surfaced a real limitation: a *deterministic* genuinely-concurrent writer-set rotation needs node-1 and node-2 to be **RPC-reachable** (so each can be told to rotate) yet **unable to exchange libp2p traffic** (so neither sees the other's rotation). No current merobox primitive provides that, and I verified each candidate fails:

- **`disconnect_node` (bridge detach)** — also kills the node's RPC: merobox reaches RPC via the published-port DNAT, which detaches with the bridge. The rotate call never arrived (run 1; node-2's container log had no `shared_rotate_writers` request).
- **`pause_container`** — freezes the process but the **kernel keeps buffering inbound gossip**, so on unpause the node drains the competing rotation *before* it can rotate. Verified: a `DELTA_APPLY` fires on node-2 immediately on resume (19:46:51.81), dropping B, so its own `{B,C}` rotate is then correctly rejected with `Executor is not a current writer` (runs 2–3).
- **stop/start** — re-syncs the competing rotation on boot.

This is the same wall `member-removed-partition-window-reconcile` is parked on; both want a **surgical peer-pair partition** (iptables/tc drop on the libp2p swarm port, keeping the RPC port). So I've parked this workflow the same way: kept as the e2e spec + local repro with a documented header + matrix comment, but **not gating CI**. The concurrent-rotation convergence guarantee #2665 establishes is covered by the unit/node tests it shipped.

Two real findings the iteration produced, both documented inline in the workflow:
1. A cold joiner can only act as a SharedStorage writer if it applied the granting rotation **as a forward-applied delta** — if it converges the post-rotation tree via Snapshot/HashComparison, the rotation never lands in its rotation log and `writers()` (which reads the log ahead of the index) keeps returning the stale genesis set. The baseline now settles joiners first to avoid this.
2. The `shared_rotate_writers` app method (full writer-set replacement) is retained as it's generally useful for SharedStorage testing.

Net change kept in the PR: `shared_rotate_writers` app method + the (unregistered) workflow as spec/repro. Happy to follow up with the surgical-partition primitive in merobox if you'd like it as a separate effort — it would unblock both this and `member-removed-partition-window-reconcile`.

chefsale:
**Update — landed deterministic coverage, re-enabled the e2e on plain-bridge, and filed the infra gap.**

After mapping the partition-primitive space (and confirming the NAT testing issues #2447/#2453/#2467 cover *relay recovery*, not concurrent rotation), here's where this PR landed:

1. **Deterministic node-level test (the real coverage).** `crates/node/src/sync/rotation_log_reader.rs` gains two tests for the #2668 guarantee. The convergence property is that `writers_at` is **invariant to rotation-log insertion order** (two nodes append concurrent rotations in opposite orders but must resolve the same set). `writers_at_is_insertion_order_invariant_under_concurrent_rotations` and `settling_rotation_converges_and_revokes_under_concurrency` assert exactly that + retroactive revocation, using the real resolver — fully deterministic, no Docker. This is the reliable home for the guarantee.

2. **Re-enabled the merobox e2e on a plain bridge.** Driven entirely by the existing `disconnect_node`/`connect_node` fault-injection steps (no NAT, no custom scripts), with one rule: never `call` a disconnected node — isolate the two non-rotators while each rotator acts. Plain bridge on purpose: NAT topology exists to exercise relay recovery (the flaky path), which is irrelevant here and would import flakiness across 3 reconnects. Re-registered in the `e2e-rust-apps` matrix; CI will tell us if it's stable.

3. **Filed the real infra gap: [merobox#278](https://github.com/calimero-network/merobox/issues/278)** — a surgical peer-pair partition primitive (cut libp2p between two nodes, keep RPC up). It would simplify this workflow and also unblock `member-removed-partition-window-reconcile` and `sync-resilience-lossy-link` (#2468) — three parked needs, one primitive.

Also retained: the `shared_rotate_writers` app method, and the baseline now settles joiners before rotating (cold joiners only become usable SharedStorage writers if they apply the granting rotation as a forward-applied delta, not via snapshot/HC catch-up).

Issue #2319, opened 2026-05-10
node, sync: SyncSessionActor mailbox saturates under stuck-session backpressure (post-#2316), wedging fuzzy test


rtb-12:
## Summary

Following the SyncSessionActor isolation introduced in #2316 / a403cfed, a single
slow `HashComparison` sync session can saturate the actor's bounded mailbox.
While the stuck session waits for a peer response (~10 s), every subsequent
interval-tick `try_send` returns `SyncSessionSendError::Full` and is silently
dropped per the #2316 design ("next interval tick will retry"). In practice the
next tick also sees a full mailbox, so the affected node stops dispatching new
sync sessions for the rest of the test.

The KV-Store-with-Handlers fuzzy load test on master is failing intermittently
because of this: the test passes when at most one node wedges this way and
fails when two or more do. There have been **7 failing runs across 4 master
commits** in the last 4 days (May 7–10), all on the same job
`Run KV Store with Handlers Fuzzy Test with Profiling`, all with the same
failing assertion `cross_node_handler_check` (cross-node reads return
`{'output': None}`).

## Evidence — same SHA, two outcomes

`f6c3c999` (latest master at time of writing) ran fuzzy tests three times.
Identical code, three different results:

| Run ID | Trigger | Verdict | cross_node pass rate | Total assertion pass rate |
|---|---|---|---|---|
| [25609447428](https://github.com/calimero-network/core/actions/runs/25609447428) | push      | ✅ PASSED | 119 / 130 = **91.5%** | **99.5%** (just above 99% threshold) |
| [25609762483](https://github.com/calimero-network/core/actions/runs/25609762483) | workflow_run | ❌ FAILED | similar to schedule below | — |
| [25615750020](https://github.com/calimero-network/core/actions/runs/25615750020) | schedule  | ❌ FAILED | 3 / 464 = **0.6%**    | **94.7%** |

In the failing run, only the first 3 cross-node reads succeed; reads 4 → 464
all return `None`. The pass count freezes at 3 and never increases.

## Sync wedge stats — failing run (run 25615750020, SHA f6c3c999)

| Node | `Scheduled sync` | `Sync finished successfully` | `mailbox full` warns | `timeout receiving` errors | Last successful sync |
|---|---|---|---|---|---|
| node-1 | ~5099 | **4956** | 0     | 143 | 01:45:15 (62 min in) |
| node-2 |   238 |   **122** | **4705** | 76  | 00:47:29 ( 4 min in) |
[...]

PR #2675, opened 2026-06-03
feat: durable pull-based group-key recovery (hybrid push + pull) (#2613)

### LINKED ISSUE #2613: bug: KeyDelivery has no durable retry — joiner can be permanently locked out of group decryption
## Description

Supersedes #2135 (file paths and call graph have changed; the core defect remains). On current `master`, key delivery to a joining node is still best-effort with no durable retry, so a joiner can be permanently locked out of group decryption if delivery fails at the wrong moment.

### Current behavior

`maybe_publish_key_delivery` (`crates/node/src/key_delivery.rs`) is fire-and-forget. Every failure path logs a warning and returns with no retry:

- `wrap_for_member` fails → `warn!(?e, "failed to wrap group key for joiner")`
- `sign_and_publish_namespace_op` fails → `warn!(?e, "failed to publish KeyDelivery")`

It is now invoked from two apply paths, giving some cross-member redundancy:

- `crates/node/src/handlers/network_event/namespace.rs:266`
- `crates/node/src/sync/manager/mod.rs:4343` (backfill)

But both are gated on a **newly-applied** `MemberJoined`/`MemberJoinedOpen`, never on `Duplicate`. This guard is deliberate (#2319 — re-publishing a fresh `KeyDelivery` every backfill round grows the namespace governance DAG without bound and never converges). As a side effect, each member takes exactly **one** shot at delivery, at the instant it first applies the join op.

The retry service `NamespaceRetryService` (moved since #2135: now `crates/governance-store/src/namespace/retry.rs`) only retries **decryption** of `NamespaceOp::Group` ops whose group key is already available locally. It never re-publishes a `KeyDelivery`.

There is no joiner-initiated re-request mechanism (no `RequestKeyDelivery`-style op).

### Impact

If every currently-online member fails to publish `KeyDelivery` at the moment it newly-applies the join op (transient network issue, partition, all members briefly offline), the joiner permanently cannot decrypt any group-scoped governance ops. Subsequent syncs see the join op as `Duplicate`, so no member retries. The only recovery is leave-and-rejoin.

### Files

- `crates/node/src/key_delivery.rs` — `maybe_publish_key_delivery` (fire-and-forget)
- `crates/node/src/sync/manager/mod.rs:4343` — one-shot publish gated on newly-applied (#2319)
- `crates/node/src/handlers/network_event/namespace.rs:266` — second publish site
- `crates/governance-store/src/namespace/retry.rs` — retries decryption only, not delivery

### Suggested fix

Add a joiner-initiated `RequestKeyDelivery` namespace op that any group member holding the key can answer with a fresh `KeyDelivery`. This is rate-limitable / idempotent on the request side, so it avoids the #2319 unbounded-DAG-growth problem that forced the passive path to be one-shot, while still giving the joiner a durable recovery path.

chefsale:
Closes #2613.

> **Note:** this PR's approach changed mid-flight. It started as a pure pull-based replacement that hard-removed on-DAG `KeyDelivery`; CI showed that breaks the admin-adds-member flows, so it landed as a **hybrid** (keep the admin push, add a durable pull). This description reflects the final state. e2e-rust-apps is green.

## Problem

Group-key delivery to a joiner was a single fire-and-forget `RootOp::KeyDelivery` op pushed onto the namespace governance DAG **once**, when a member first applied the join. It was gated on the newly-applied join because re-publishing it every backfill round grows the DAG without bound (#2319). Every failure path just logged a warning. So if that one delivery attempt failed at the wrong moment (transient partition, gossip mesh still forming, members briefly offline), the joiner was **permanently locked out** of group decryption — every later sync saw the join as `Duplicate`, so nothing retried. The only recovery was leave + rejoin.

**Root cause:** the durability defect was specifically the **receiver-side, one-shot** delivery (`maybe_publish_key_delivery`). Admin-*initiated* delivery was never the problem.

## Fix — keep the prompt push, add a durable pull

1. **Admin-initiated push (kept, one-shot, on the DAG).** `add_group_members` and `admit_tee_node` still publish a `KeyDelivery` immediately, so a just-added member promptly decrypts its `MemberAdded`/`ContextRegistered` and can use the (sub)group in-band. One publish per add — never the #2319 re-publish source.

2. **Joiner-initiated pull (new, off-DAG, durable — the actual #2613 fix).** A member that lacks a key it needs requests it from a sync peer via new sync-stream messages:
   - `InitPayload::GroupKeyRequest { namespace_id, group_id, requester_public_key }`
   - `MessagePayload::GroupKeyResponse { key_envelope_bytes, responder_identity }`

   Responder authorises by membership + cross-namespace pin and ECDH-wraps the key (`build_group_key_delivery`); the joiner unwraps, stores, seeds the bootstrap admin, and replays buffered ops (`apply_received_group_key`). It fires on every sync round, at end of namespace sync, **and on gossip receipt**, so it retries until the key arrives.

3. **Removed:** the one-shot receiver-side `maybe_publish_key_delivery` — replaced by the pull.

## Properties
- **Fixes #2613** — durable recovery; no permanent lockout.
- **No #2319 regression** — admin pushes are one-shot; the durable layer is off-DAG.
- **No wire break** — `RootOp::KeyDelivery` retained in its original enum slot (no borsh discriminant shift, no coordinated-upgrade requirement).
- **No regression for admin-add / TEE-admit** — those keep the prompt push.

## Before / after

| | Before (master) | After |
|---|---|---|
| Joiner key delivery | one-shot DAG push; miss ⇒ permanent lockout | admin push **+** durable joiner pull |
| Recovery from a missed delivery | none (leave + rejoin) | pull retries until delivered |
| #2319 (DAG growth) | the reason delivery was capped at one shot | n/a — durable retry is off-DAG |
| Admin-add / TEE-admit | prompt push | unchanged (prompt push) |

## Tests
- **e2e:** full `e2e-rust-apps` matrix green (group / subgroup / membership / auto-follow / shared-storage).
- **In-repo, deterministic:** cross-store responder↔joiner key exchange; recovery-trigger decoupling from the governance-pending gate; per-op `key_id` "awaiting" (a node holding the namespace key still awaits a Restricted subgroup's own key); buffered-op replay.

## Open follow-up (not blocking)
The pull's bootstrap-admin seed trusts an unauthenticated `responder_identity` field, whereas the push path uses the signature-authenticated `op.signer`. Narrow surface (pull-only fallback; root-group + no-existing-meta), but a mild weakening vs. master — suggest a follow-up to bind it to the stream-authenticated peer identity.

---

> [!NOTE]
> **High Risk**
> Changes group membership cryptography and key distribution (ECDH wrap, membership gates, bootstrap admin seeding); missed edge cases could leave members unable to decrypt subgroup ops until pull succeeds.
> 
> **Overview**
> Fixes **#2613** by replacing the one-shot, receiver-side `KeyDelivery` push on join (`maybe_publish_key_delivery`, deleted) with a **durable pull** while keeping **admin one-shot DAG pushes** from `add_group_members` and `admit_tee_node`.
> 
> **Pull path:** New sync messages `GroupKeyRequest` / `GroupKeyResponse` on the sync stream. Governance-store adds `namespace_groups_awaiting_key` (per-op `key_id`, so Restricted subgroups still pull even when the namespace key is held), `build_group_key_delivery`, and `apply_received_group_key`. `SyncManager::recover_missing_group_keys` runs after namespace sync, on interval sync, and on gossip apply when ops are buffered undecryptably.
> 
> **Apply path:** `MemberJoined` / `MemberJoinedOpen` no longer enqueue `PendingKeyDelivery`. On-DAG `KeyDelivery` apply delegates to `apply_received_group_key`. Types/docs now describe joiners pulling keys from peers, not peers reacting on every join apply.
> 
> **Tests** cover cross-store wrap/unwrap, awaiting-key logic, and decoupling key recovery from the governance-pending backfill gate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 058db7daaca94e73626e3abaf3c589b3988942e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
## Breakage surface — before/after with real-world scenarios

A scenario-by-scenario map of everything this PR could break, so reviewers can target testing. Grouped by **wire/version**, **delivery flows**, **security**, and **operational**.

---

### 🔴 1. Mixed-version namespace DAG divergence (the headline risk)

The borsh discriminant of `MemberJoinedOpen` shifts when `KeyDelivery` is removed, and old nodes still *write* `KeyDelivery` ops onto the shared namespace governance DAG.

**Scenario:** Rolling upgrade of a 5-node namespace; nodes A,B on new build, C,D,E still old. Admin C invites a new member.

- **Before:** C publishes a `KeyDelivery` op to the namespace DAG. Every node decodes and applies it; DAG heads stay identical across A–E.
- **After (mixed cluster):** C still publishes `KeyDelivery`. C,D,E apply it and advance their DAG head to include it. A,B receive it, **fail to borsh-decode** (variant gone), skip-and-log, and do **not** advance their head. → **A,B vs C,D,E diverge on namespace DAG heads** → governance convergence stalls / state-hash checks mismatch.
- **Risk:** High in any non-atomic rollout.
- **Mitigation:** Coordinated all-at-once upgrade (documented in PR). Reviewers: confirm the deployment story can guarantee this, or we add a transition release that *stops producing* but still *decodes* `KeyDelivery` first.

---

### 🔴 2. Cross-version request/response (new joiner ↔ old peer)

`GroupKeyRequest`/`GroupKeyResponse` are new `InitPayload`/`MessagePayload` variants.

**Scenario:** New-build joiner J in a still-mostly-old namespace; J synced the DAG but lacks the key.

- **Before:** any key-holding member pushes `KeyDelivery`; J picks it up off the DAG regardless of peer versions.
- **After:** J sends `GroupKeyRequest` to its sync peer. An **old** peer can't decode the new Init variant → request fails → J retries next round. J only recovers once it happens to sync with a **new** key-holder.
- **Risk:** Medium, mixed-cluster only. J's recovery is delayed, not broken (retries every round).
- **Mitigation:** coordinated upgrade; the retry loop is the safety net.

---

### 🟠 3. Legacy persisted `KeyDelivery` ops on disk / replay

**Scenario:** A node upgrades and restarts; its RocksDB namespace op-log already contains `KeyDelivery` ops from before.

- **Before:** decode + apply on replay.
- **After:** those ops fail to decode and are **skip-and-logged** (`record_namespace_decode_invalid`), never panic. The key they once carried is already stored locally (off-DAG in `GroupKeyring`), so no key is lost. Deltas are an optimization (HashComparison reconciles without them), so no re-encode migration.
- **Risk:** Low. Watch for log-noise spikes (`signed_iter_*` / decode-invalid metrics) right after upgrade.

---

### 🟢 4. Invited member join (`MemberJoined`) — the #2613 path

**Scenario:** Admin invites Carol; Carol runs `join_group`.

- **Before:** primary delivery = direct `NamespaceJoinResponse` (one-shot key in the join handshake). Fallback if that peer was keyless = wait for a gossip `KeyDelivery` op, **one shot**, permanent lockout if it failed.
- **After:** primary path **unchanged** (still `NamespaceJoinResponse`). Fallback = `recover_missing_group_keys` pulls every sync round until delivered. `join_group` still wakes on the `GroupKeyDelivered` event (kept) and still honors `key_delivery_fallback_wait`.
- **Risk:** Low — this is strictly more robust. Main behavioral change: fallback latency is now sync-cadence-bound rather than apply-instant.

---

### 🟠 5. Admin adds an existing member to a subgroup (`add_group_members`)

**Scenario:** Admin adds Bob (already a namespace member) to subgroup S.

- **Before:** admin proactively pushes a `KeyDelivery` for S, ECDH-wrapped for Bob, right after `MemberAdded`.
- **After:** no push. Bob backfills the **encrypted `MemberAdded` op itself** (a `NamespaceOp::Group` op he can't decrypt) → it buffers → `groups_awaiting_key` returns S → Bob pulls S's key from a sync peer. So the membership op *is* the trigger; the pull fires the same sync round it backfills.
- **Risk:** Medium. Works because the add is itself encrypted group traffic. Edge to test: a **completely silent subgroup with no ops at all** — Bob has nothing buffered to trigger on until the first op appears (he can't write until he has the key, and he won't pull until there's traffic). Verify Bob can bootstrap into a brand-new empty subgroup.

---

### 🟠 6. TEE node admission (`admit_tee_node`) + fleet-join

**Scenario:** Verifier V admits TEE node N via attestation.

- **Before:** V proactively publishes a `KeyDelivery` for the namespace root key, wrapped for N.
- **After:** no push. N's `MemberJoinedViaTeeAttestation` is an encrypted `NamespaceOp::Group` only V can apply; N backfills it, can't decrypt → buffers → `groups_awaiting_key` returns the root group → N pulls from V (a key-holding sync peer). N already triggers `sync_namespace` post-fleet-join, so the pull runs there. Bootstrap-admin seed still fires (root group + no local meta).
- **Risk:** Medium-high — **delicate, HA-critical, and not e2e-validated in this PR.** Requires V to be reachable as N's namespace sync peer when N pulls (same liveness requirement the old push had). Reviewers: TEE HA e2e before merge.

---

### 🟠 7. Open-subgroup self-join (`MemberJoinedOpen`)

**Scenario:** Member self-joins an Open subgroup.

- **Before:** any key-holder pushes `KeyDelivery` on applying the join op.
- **After:** joiner pulls (same mechanism as #4). The dedicated `OpenSubgroupJoinResponse` direct path is unchanged.
- **Risk:** Low-medium; same family as #4/#5.

---

### 🟠 8. Bootstrap-admin trust anchor weakened (security)

**Scenario:** Keyless node bootstraps into a namespace root without an invitation (TEE path) and seeds its namespace admin from whoever delivered the key.

- **Before:** trust anchor = `op.signer` of the `KeyDelivery` op — **cryptographically authenticated** by the op signature.
- **After:** trust anchor = `responder_identity`, a **self-declared field** in `GroupKeyResponse` with no signature binding it to the actual key-wrapper. A peer that holds the real group key (required to produce a valid envelope) could set `responder_identity` to an arbitrary identity and seed a bogus admin on a bootstrapping node.
- **Risk:** Narrow (TEE-bootstrap only; attacker must already be a legitimate key-holder; seed only on root-group + no-existing-meta), but it **is** a weakening of the old authenticated-signer model.
- **Mitigation (follow-up):** bind `responder_identity` to the stream-authenticated peer identity (`their_identity`) instead of trusting the payload field, or sign the response. Flagging for reviewer decision.

---

### 🟡 9. Recovery latency / peer availability

**Scenario:** Joiner's namespace sync peer happens to be another keyless fresh joiner.

- **Before:** any key-holder, anywhere, pushed on apply.
- **After:** the pull asks the **peer it just synced with**; if keyless, empty envelope → retry next round (possibly a different peer). Recovery is bounded by sync cadence × rounds-until-it-hits-a-key-holder.
- **Risk:** Low (eventually converges) but worth a worst-case bound check in sparse/partitioned topologies.

---

### 🟡 10. Observability / dashboards

- **Before:** `op_kind="key_delivery"` metric label; `published KeyDelivery…` logs; DAG carries delivery ops.
- **After:** that op-kind label is gone; new logs `recovered group key via direct delivery`; smaller namespace DAG (a **positive** — eliminates the #2319 unbounded-growth vector entirely).
- **Risk:** Low — alerts/dashboards keyed on `key_delivery` op metrics go silent. Update them.

---

### Net

The functional flows (#4–#7) are covered by the pull because every membership op is itself encrypted group traffic that triggers it. The genuinely load-bearing risks are **#1/#2 (coordinated upgrade)**, **#6 (TEE e2e)**, and **#8 (trust-anchor authentication)** — recommend gating merge on those three.

chefsale:
## Follow-up: fixed a critical flaw — the pull was one-shot, not durable (44a678d2)

Caught a design flaw that nearly reintroduced the exact lockout this PR set out to fix.

**The flaw:** I claimed the joiner "pulls every sync round until it has the key." That sync round doesn't exist. `recover_missing_group_keys` is only reachable via `sync_namespace_from_peer`, which is **edge-triggered** (join / startup / readiness / fleet-join) or gated on `should_backfill_governance` (governance-pending > 0). A member that is caught up on governance but missing only a key has an **empty** pending buffer and no pending edge — so the pull fired **exactly once, at join time**, and never retried. If the responder wasn't ready at that instant (mesh forming, membership not yet applied, peer keyless) → permanent lockout. Same #2613 failure mode, just moved from the push side to the pull side.

It slipped through because the unit tests call the recovery functions directly, bypassing the (missing) trigger wiring — green that hid the gap.

**Fix (44a678d2):**
- `perform_interval_sync` now drives `recover_missing_group_keys` **every interval tick**, gated on "do I lack a key for a group I hold buffered ops for" — **decoupled** from the governance-pending buffer. This is what makes the pull durable.
- `recover_missing_group_keys` now tries the synced peer first, then **all namespace-mesh peers**, stopping at the first that serves each group's key (extracted `request_group_key_from_peer`). A keyless first peer no longer costs a whole interval; recovery converges in one round once any key-holder is reachable. (Also resolves the latency concern from scenario #9.)
- Regression test `key_recovery_trigger::keyless_member_awaits_key_even_with_empty_governance_pending` pins the decoupling so a refactor can't silently re-gate key recovery behind governance-pending.

Tests: node 273 lib + sync_protocols 12 + sync_sim 312 green; fmt/clippy clean.

**Still owed:** a true multi-node "miss-at-join → recover-on-next-interval" e2e (merobox/scaffolding) — the local suites don't exercise the trigger under realistic timing, which is precisely the blind spot that hid the original flaw.

chefsale:
## ⚠️ Design changed during this PR — please read this instead of the earlier comments

This supersedes my two earlier comments (the *"hard-remove"* description and the *"breakage surface / 10 scenarios"* matrix). Those described an approach that **was abandoned**. CI proved a pure hard-remove of on-DAG `KeyDelivery` breaks the admin-adds-member flows, so the PR pivoted to a **hybrid**. The notes below are the accurate, final state (e2e-rust-apps **57/57 green**, branch up to date with master).

## What this PR actually does (final)

`#2613`: group-key delivery to a joiner was a single fire-and-forget `KeyDelivery` op pushed onto the namespace DAG **once** (gated on the newly-applied join, because re-publishing every backfill round grows the DAG without bound — `#2319`). If that one shot failed (transient partition, mesh still forming, members briefly offline) the joiner was **permanently locked out** of group decryption; the only recovery was leave + rejoin.

**Root cause:** the *durability* defect was specifically the **receiver-side, one-shot** delivery (`maybe_publish_key_delivery`). It is **not** a problem with admin-initiated delivery.

### The fix: keep the prompt push, add a durable pull

1. **Admin-initiated push — kept (one-shot, on the DAG).** `add_group_members` and `admit_tee_node` still publish a `KeyDelivery` immediately, so a just-added member can promptly decrypt its `MemberAdded`/`ContextRegistered` and use the (sub)group in-band. One publish per add — this was **never** the `#2319` re-publish source.
2. **Joiner-initiated pull — new (off-DAG, durable).** A member that lacks a key it needs requests it from a sync peer via new sync-stream messages `InitPayload::GroupKeyRequest` → `MessagePayload::GroupKeyResponse` (`build_group_key_delivery` responder, `apply_received_group_key` joiner). It fires on every sync round, at end of namespace sync, **and on gossip receipt** — so it retries until delivered. **This is the actual `#2613` fix.**
3. **Removed — the one-shot receiver-side `maybe_publish_key_delivery`** (the defect), replaced by the pull.

### Corrections to my earlier comments
- ❌ **No wire break.** `RootOp::KeyDelivery` is retained in its original enum slot — the borsh discriminant shift I described earlier does **not** happen. No coordinated-upgrade requirement.
- ❌ The "all three producers removed / lazy-pull" framing is obsolete — the admin producers are **kept**.
- ✅ Still true: `#2613` is fixed by the durable pull; `#2319` is not reintroduced (admin pushes are one-shot; the durable layer is off-DAG).

## Before / after

| | Before (master) | After |
|---|---|---|
| Joiner key delivery | one-shot DAG push; miss ⇒ permanent lockout | admin push **+** durable joiner pull |
| Recovery from a missed delivery | none (leave + rejoin) | pull retries until delivered |
| `#2319` (DAG growth) | the reason delivery was capped at one shot | n/a — durable retry is off-DAG |
| Admin-add / TEE-admit | prompt push | unchanged (prompt push) |

## Tests
- e2e: full `e2e-rust-apps` matrix green (group / subgroup / membership / auto-follow / shared-storage).
- Unit/integration (in-repo, deterministic): cross-store responder↔joiner key exchange; recovery-trigger decoupling from the governance-pending gate; per-op `key_id` "awaiting" (a node holding the namespace key still awaits a Restricted subgroup's own key); buffered-op replay.

## One open follow-up (not blocking, reviewer's call)
The **pull's** bootstrap-admin seed trusts an unauthenticated `responder_identity` field, whereas the **push** path uses the signature-authenticated `op.signer`. Narrow surface (pull-only fallback, root-group + no-existing-meta), but a mild weakening vs. master. Suggest a follow-up issue to bind it to the stream-authenticated peer identity rather than blocking here.

_(Note: the PR title still says "remove on-DAG KeyDelivery" — that's stale; it should read "durable pull-based group-key recovery" or similar.)_

Issue #1546, opened 2025-10-21
feat: add bidirectional communication inside the websockets for query and mutate


chefsale:
Currently we have SSE and WS for events streaming unidirectionally, would make sense to maybe repurpose websockets for a bidirectional communication of mutate/query requests instead.

PR #2689, opened 2026-06-04
feat(meroctl): interactive shell over a persistent WebSocket (execute)

### LINKED ISSUE #2615: feat(meroctl): support execute (query/mutate) over WebSocket
## Follow-up to #1546

#1546 added bidirectional `execute` (query/mutate) support to the node's WebSocket server (`crates/server/src/ws.rs`): clients can now send an `execute` request over the same socket used for event streaming and get the result back. The server side is done; this issue tracks the **meroctl client** side.

## Context

meroctl already speaks WebSocket for the event watch — see `crates/meroctl/src/cli/context/watch.rs` (`tokio_tungstenite::connect_async`). Today `execute`-style calls from meroctl go over JSON-RPC (HTTP). With the server now accepting `execute` over WS, meroctl could issue query/mutate calls on a persistent socket alongside (or instead of) the per-call HTTP round-trip.

## Wire format

The WS `execute` request reuses the JSON-RPC `ExecutionRequest` envelope, tagged as method `execute`:

```json
{ "id": 1, "method": "execute", "params": { "contextId": "…", "method": "…", "argsJson": {}, "substitute": [] } }
```

Response shape mirrors the existing WS responses: `{ "id": 1, "result": { "output": … } }` on success, `{ "id": 1, "error": { … } }` on failure.

## Tasks
- [ ] Add a WS-based execute path to the meroctl client (likely alongside the existing tungstenite usage in `cli/context/watch.rs`, factored into a small reusable helper).
- [ ] Decide whether this is opt-in (e.g. a flag) or replaces the HTTP execute path for relevant commands.
- [ ] Correlate request/response by the `id` field (multiple in-flight calls on one socket).

## Notes
- Auth on WS is **connection-level** (validated once at the HTTP upgrade), unlike HTTP JSON-RPC which re-authenticates every call. Keep that in mind for long-lived sockets and token expiry.
- Full happy-path execute requires a live node; covered by e2e, not unit tests.

chefsale:
## Summary

Closes #2615. Follow-up to #1546, which added bidirectional `execute` (query/mutate) support to the node's WebSocket server. This wires up the **meroctl client** side.

### Why a shell, not a one-shot flag

A *one-shot* call gains nothing from WebSocket over HTTP — it pays the TCP connect **plus** the WS upgrade handshake just to send a single frame, and HTTP additionally re-authenticates per call (WS auth is connection-level, validated once at upgrade). The socket only earns its keep when **reused across many calls**. So instead of a pointless `--ws` flag, this ships an interactive shell.

## What changed

**`WsSession` — `crates/meroctl/src/ws.rs`:**
- `connect()` derives the `ws(s)://…/ws` URL and attaches connection-level auth to the upgrade handshake.
- `WsSession` connects once and runs many `execute` calls, **correlating each reply to its request by a monotonic id** (event pushes carry no `id` and unrelated replies are skipped).
- Responses are adapted into the JSON-RPC `Response` shape so output renders identically regardless of transport.

**`meroctl call -i / --interactive` — `crates/meroctl/src/cli/call.rs`:**
- Opens a REPL over a single persistent socket: read `<method> [args-json]` lines, run each over the same session, print the result.
- Meta-commands: `:context <alias>` (switch context), `:help`, `:quit` (and Ctrl-D to exit). An optional positional `METHOD` runs as the first call. `METHOD` is now optional.
- Handler-level errors keep the shell alive (printed); only a transport/protocol failure ends the session. The one-shot HTTP JSON-RPC path is unchanged.

**Client plumbing — `crates/client/src/`:**
- `Client::ws_url()` and `Client::auth_header()` (+ `ConnectionInfo::auth_header()`) expose the WS URL and connection-level auth that previously lived behind the private `connection` field.

**`context watch`** refactored onto the shared `connect()` helper, which also gives it **auth support it lacked before**.

## Addresses the issue's three tasks
- WS-based execute path, factored into a small reusable helper (`WsSession`) alongside the tungstenite usage in `watch.rs`.
- Decision: WS is exposed via the interactive shell (`-i`); HTTP JSON-RPC stays the default for one-shot calls.
- Replies correlated by `id` across multiple in-flight calls on one socket.

## Testing
- `cargo build` / `cargo fmt --check` / `cargo clippy` clean (no new warnings from the added code).
- 3 unit tests for the WS->JSON-RPC body adaptation (result / handler-error / server-error mapping); existing client tests green.
- Connection-failure path smoke-tested (fails fast and cleanly, no panic).
- Full happy-path execute through the shell requires a live node and is covered by e2e, not unit tests (per the issue).

---

> [!NOTE]
> **Medium Risk**
> Touches connection-level bearer auth on WebSocket upgrade and long-lived sessions (token expiry not auto-refreshed on the socket); scheme handling is hardened to avoid cleartext WS with tokens.
> 
> **Overview**
> Adds **`meroctl call -i`**, an interactive REPL that keeps one WebSocket open and runs many context `execute` calls over it, with per-call timeout, meta-commands (`:context`, `:help`, `:quit`), and optional seed `METHOD`/`--args`. One-shot calls still use HTTP JSON-RPC; `METHOD` is optional only in interactive mode.
> 
> Introduces shared **`crates/meroctl/src/ws.rs`**: `connect()` attaches bearer auth at upgrade; **`WsSession`** sends `execute` frames, correlates replies by monotonic request `id` (skips event pushes and stray frames), adapts WS responses to JSON-RPC shape, and **`keepalive`** answers pings while waiting at the prompt.
> 
> **`Client::ws_url()`** and **`Client::auth_header()`** (plus **`ConnectionInfo::auth_header()`**) centralize `ws(s)://…/ws` derivation with explicit `http`/`https` mapping, query/fragment stripping, and the same proactive auth as HTTP. **`context watch`** is refactored onto shared `connect()`, gaining auth it previously lacked.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2ff6632a70453b7237f912372076024027d24ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -151,6 +151,11 @@ where
         url.set_scheme(scheme)
             .map_err(|()| eyre::eyre!("failed to set WebSocket URL scheme"))?;
         url.set_path("ws");
+        // Drop any query/fragment from the HTTP base so they don't ride along
+        // to the upgrade request (they're meaningless to `/ws` and could carry
+        // sensitive values).
+        url.set_query(None);
+        url.set_fragment(None);
 
         Ok(url)
     }
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -16,7 +16,7 @@ use calimero_server_primitives::admin::{
     CreateContextIdAlias, CreateContextIdentityAlias, DeleteAliasResponse, ListAliasesResponse,
     LookupAliasResponse,
 };
-use eyre::Result;
+use eyre::{bail, Result};
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 use url::Url;
@@ -136,12 +136,17 @@ where
     /// The WebSocket endpoint URL (`ws(s)://…/ws`) for this connection, derived
     /// from the HTTP `api_url` by swapping the scheme and pointing at `/ws`.
     /// Used by the event-stream (`subscribe`) and `execute`-over-WS paths.
+    ///
+    /// The scheme mapping is explicit (`http`→`ws`, `https`→`wss`); an
+    /// unexpected scheme errors rather than silently degrading to cleartext
+    /// `ws://`, which would leak the bearer token attached to the handshake.
     pub fn ws_url(&self) -> Result<Url> {
         let mut url = self.connection.api_url.clone();
 
         let scheme = match url.scheme() {
+            "http" => "ws",
             "https" => "wss",
-            _ => "ws",
+            other => bail!("cannot derive a WebSocket URL from scheme `{other}://`"),
         };
         url.set_scheme(scheme)
             .map_err(|()| eyre::eyre!("failed to set WebSocket URL scheme"))?;
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -133,6 +133,34 @@ where
         &self.connection.api_url
     }
 
+    /// The WebSocket endpoint URL (`ws(s)://…/ws`) for this connection, derived
+    /// from the HTTP `api_url` by swapping the scheme and pointing at `/ws`.
+    /// Used by the event-stream (`subscribe`) and `execute`-over-WS paths.
+    pub fn ws_url(&self) -> Result<Url> {
+        let mut url = self.connection.api_url.clone();
+
+        let scheme = match url.scheme() {
+            "https" => "wss",
+            _ => "ws",
+        };
+        url.set_scheme(scheme)
+            .map_err(|()| eyre::eyre!("failed to set WebSocket URL scheme"))?;
+        url.set_path("ws");
+
+        Ok(url)
+    }
+
+    /// Resolve the bearer auth header for this connection, if the node requires
+    /// auth and credentials are available (triggering proactive authentication
+    /// when no token is stored, just like the HTTP request helpers).
+    ///
+    /// WebSocket auth is connection-level — validated once at the upgrade
+    /// handshake — so callers attach this to the upgrade request rather than to
+    /// every message.
+    pub async fn auth_header(&self) -> Result<Option<String>> {
+        self.connection.auth_header().await
+    }
+
     /// Create context identity alias (legacy method for backward compatibility)
     pub async fn create_context_identity_alias(
         &self,
--- a/crates/client/src/connection.rs
+++ b/crates/client/src/connection.rs
@@ -219,6 +219,17 @@ where
         response.json::<O>().await.map_err(Into::into)
     }
 
+    /// Resolve the bearer auth header for this connection (always treating auth
+    /// as required), if a node name and credentials are available.
+    ///
+    /// Public counterpart to [`Self::ensure_auth_header`] for callers outside
+    /// the HTTP request helpers — e.g. attaching auth to a WebSocket upgrade
+    /// handshake, where auth is validated once at connect rather than per call.
+    /// May trigger a browser-based authentication flow when no token is stored.
+    pub async fn auth_header(&self) -> Result<Option<String>> {
+        self.ensure_auth_header(true).await
+    }
+
     /// Ensure a valid auth header is available and return it.
     ///
     /// Returns `None` immediately if `requires_auth` is false or `node_name` is unset.
--- a/crates/meroctl/src/cli/call.rs
+++ b/crates/meroctl/src/cli/call.rs
@@ -56,6 +56,12 @@ pub struct CallCommand {
         value_delimiter = ','
     )]
     pub substitute: Vec<Alias<PublicKey>>,
+
+    #[arg(
+        long = "ws",
+        help = "Issue the call over a WebSocket connection instead of HTTP JSON-RPC"
+    )]
+    pub ws: bool,
 }
 
 fn serde_value(s: &str) -> serde_json::Result<Value> {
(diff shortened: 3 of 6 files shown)


chefsale:
Thanks both bots — addressed in 7ccf7948.

**Fixed:**
- 🟡 **Idle shell misses pings** (Bugbot) — the prompt loop now races `WsSession::keepalive()` against stdin via `tokio::select!`, answering server pings while waiting for input. `keepalive` is cancel-safe (`StreamExt::next`) and only runs while no request is outstanding, so it can't swallow a real reply. The socket no longer dies after the ~40s idle window.
- 🟡 **Seed call ignores `--args`** (Bugbot + MeroReviewer) — `--args` is now threaded into the seed call, so `call -i <method> --args '{...}'` runs the first call with the provided JSON.
- 💡 **`ws_url()` cleartext fallback** (MeroReviewer) — scheme mapping is now explicit (`http`→`ws`, `https`→`wss`) and errors on anything else, so a misconfigured scheme can't silently downgrade to `ws://` and leak the bearer token.

**Addressed by clarification:**
- 🟡 **Concurrent use could drop replies** (MeroReviewer) — `execute` takes `&mut self`, so the borrow checker already forbids two in-flight calls on one session; a mismatched-id frame is necessarily an event push and is correctly skipped. Added a doc note making this contract explicit. (No background demultiplexer needed for the single-caller shell.)

**Deferred (with rationale):**
- 💡 **No execute timeout** (MeroReviewer) — long-running WASM calls are legitimate, the idle case is now covered by `keepalive`, and Ctrl-C interrupts. A `--timeout` flag would be a reasonable follow-up but I didn't want a fixed timeout to abort valid slow calls.
- 🟡 **TLS pinning / self-signed** (MeroReviewer) — matches the existing `reqwest` HTTP client and the pre-existing `context watch` path; `wss` certs go through the system trust store as before. Not a regression introduced here, so out of scope for this PR.

Docs bot (AGENTS.md/CONTRIBUTING.md): this is an additive CLI flag, no architecture change — skipping.

chefsale:
Second review pass addressed in d35c24b1. Several findings were stale re-flags of fixes already in 7ccf794 (`--args` seed, explicit `ws_url` scheme, idle ping handling) — those stand.

**Fixed:**
- 🟡 **keepalive exits with an eyre trace** / 💡 **keepalive Err on clean close** — the keepalive arm now prints a one-line `Connection closed: …` and breaks the shell cleanly instead of propagating an error trace, so a server-side close reads as a normal end of session.
- 📝 **No test for id-correlation** — extracted `match_text_reply` and added three tests: matching id returns the reply, a reply for another id is skipped, an event push (no id) is skipped. (6 unit tests total now.)

**Not bugs (clarified):**
- 🟡 **Seed transport error exits the shell** — that's intentional and *consistent* with the interactive loop: handler errors are printed by `execute`/`print_response` and the shell continues; only a transport/protocol failure ends the session, for both the seed and subsequent calls.
- 🟡 **Empty method accepted** — `line.trim()` runs before the split, so the first token can never be empty for a non-empty line (and empty lines are already skipped).

**Deferred (with rationale):**
- 🟡 **Bearer token over `ws://` without TLS** — identical to the existing HTTP client, which sends the bearer to `http://` nodes (the normal localhost case). Gating auth on `wss` would break local dev; not a regression introduced here.
- 🟡 **Unbounded skip loop** — event pushes legitimately interleave with a mutation's reply on the same socket, so skipping unmatched frames is expected; a hard skip cap could abort a valid call on a chatty context. A per-call `--timeout` is a reasonable future lever but a fixed one would wrongly kill slow WASM calls.
- 💡 **`ws_url` drops a base-path prefix** — matches the existing HTTP client (`connection.rs` `set_path` likewise replaces the whole path), so meroctl uniformly assumes a root-mounted API; changing only the WS path would diverge. Worth a separate change if path-prefixed deployments need supporting.

chefsale:
Third pass addressed in d2ff663. This round the recurring robustness themes were worth acting on rather than deflecting again:

**Fixed:**
- 🟡 **Unbounded wait for a matching reply** — added a per-call timeout to the shell (`--timeout`, default 120s, `0` disables). A server that never sends the matching reply no longer hangs the prompt; a timeout keeps the session alive (the call may still be running node-side) and prints a note, so it's not fatal.
- 🟡 **Transport error → raw eyre trace, inconsistent with keepalive** (call.rs:155/168/228) — a transport/protocol failure during a call now ends the shell with the same one-line `Connection closed: …` message as the idle path. Handler errors still print and keep the shell alive (unchanged). The seed call follows the same rule.
- 🟡 **Malformed text frame fails the call** (ws.rs:113) — `match_text_reply` now skips any frame that doesn't parse as a `WsResponse` instead of erroring, since a long-lived socket may carry non-RPC frames. Added a test.
- 🟡 **keepalive arm breaks on `Ok(())` → silent exit risk** (call.rs:183) — `keepalive` now returns `Result<Infallible>`, and the select arm matches `Ok(never) => match never {}`, so the type system proves the shell can't exit silently.
- 💡 **`ws_url` keeps query/fragment** — now cleared via `set_query(None)` / `set_fragment(None)`.

8 unit tests now (added unparseable-frame skip). Build / fmt / clippy clean.

**Still deferred:**
- 🟡 **Bearer token over `ws://`** — unchanged stance: identical to the existing HTTP client talking to `http://localhost`; gating auth on `wss` would break local dev and isn't a regression here.

PR #2690, opened 2026-06-04
refactor(network): drop ipify public-IP lookup for AutoNAT/identify + static config

### LINKED ISSUE #2531: Remove third-party ipify call for public-IP discovery; use AutoNAT/identify + optional static config
## Summary

`Discovery::get_public_ip()` (`crates/network/src/discovery.rs:122`) determines the node's public IPv4 by making an unauthenticated HTTP GET to the third-party service `https://api.ipify.org` and parsing the plaintext response. This should be removed in favor of libp2p-native external-address discovery (identify `observed_addr` confirmed by AutoNAT) plus an optional operator-configured static address.

Surfaced by review on #2528 (the file was touched there but the call predates it).

## Current behavior

- `DiscoveryConfig.advertise_address: bool` (default `false`). When `true` and the node has listen addresses, `Discovery::new` calls `get_public_ip().await?`, building `AdvertiseState { ip, ports }` (`discovery.rs:103-106`).
- `AdvertiseState.ip` is then used only as a **filter** in `handlers/stream/swarm/identify.rs`: when a peer's identify reports an `observed_addr` whose IP+port matches the ipify-derived pair, we `add_external_address(observed_addr)`. The address actually advertised to rendezvous is the swarm's *confirmed* external set, not the ipify value directly.
- The identify handler comment (`identify.rs:~102`) already acknowledges the ipify IP "reflects what api.ipify.org returned, not what's actually dialable" and that AutoNAT is the real reachability source.

```rust
async fn get_public_ip() -> EyreResult<Ipv4Addr> {
    let client = reqwest::Client::builder()
        .timeout(Duration::from_secs(3))
        .build()?;
    let ip_addr = client
        .get("https://api.ipify.org")
        .send().await?
        .text().await?
        .parse()?;
    Ok(ip_addr)
}
```

## Why this should go

1. **Startup coupling to a third party.** `get_public_ip().await?` is on the `Discovery::new` constructor path. If ipify is slow, down, firewalled, or blocked (air-gapped / censored networks), it errors after the 3s timeout and `?` propagates → `NetworkManager::new` fails → **the node does not start** when `advertise_address = true`. A third-party outage becomes a local availability failure.
2. **Untrusted input.** The response is an unauthenticated plaintext body from a CDN with no validation. DNS poisoning, a MITM, or a compromised edge can return an arbitrary IP. Impact is bounded today (the value is only a filter, and AutoNAT/identify still gate `add_external_address`), but the worst realistic outcome is denial of self-advertisement — a spoofed expected-IP never matches the real `observed_addr`, so the node never confirms an external address and silently degrades to relay-only.
3. **Phone-home / privacy.** Every `advertise_address = true` node calls a third-party SaaS on startup, disclosing its existence and source IP — undesirable for a privacy-oriented network.
4. **IPv4-only.** Returns `Ipv4Addr`; no IPv6 path.
5. **Redundant with libp2p.** identify `observed_addr` (how peers actually see us) + AutoNAT v2 (confirms dialability) is the correct, already-present source of truth. The ipify value only *filters* those observations, which is both weaker and unnecessary.

Removing it also lets the `calimero-network` crate **drop the `reqwest` dependency** — it's the crate's only `reqwest` user (`crates/network/Cargo.toml:44`).

## Proposed fix

Remove `get_public_ip()` and the ipify call; source the external address from, in order of preference:

1. **Operator-configured external address (new).** Add an optional `external_address: Option<Multiaddr>` (or `Vec<Multiaddr>`) to `DiscoveryConfig`/`SwarmConfig`. When set, seed it directly via `swarm.add_external_address(addr)` at swarm init. Deterministic, no third party, ideal for known-static-IP / hosted deployments.
2. **identify + AutoNAT (libp2p-native, default).** When no address is configured, drop the `AdvertiseState.ip` filter and accept identify `observed_addr` confirmed by AutoNAT v2 before `add_external_address` — which is already the de-facto path the identify comment describes. The ipify filter is the redundant piece to delete.

### Concrete steps
- Delete `Discovery::get_public_ip()` and remove `reqwest` from `crates/network/Cargo.toml`.
- Replace the `AdvertiseState { ip, ports }` construction in `Discovery::new`: either remove `AdvertiseState.ip` entirely (rely on AutoNAT-confirmed observed addrs in `identify.rs`) or populate it from the new configured address.
- Add the `external_address` config field + plumb it into `merod` init defaults (`crates/merod/src/cli/init.rs` already references `advertise_address`).
- Update `identify.rs` Branch 1 to confirm external addresses via AutoNAT-validated observations (and/or the configured address) instead of the ipify-IP match.

### Back-compat
`advertise_address` defaults to `false`, so the common path is unaffected. For nodes that set it, behavior shifts from "ipify-detected IP filter" to "AutoNAT-confirmed observed address (+ optional configured address)" — strictly more trustworthy. Document the config change in the migration notes.

## Acceptance
- No outbound call to `api.ipify.org` (or any external IP-echo service) anywhere in the node.
- `calimero-network` no longer depends on `reqwest`.
- A node with `advertise_address = true` and no internet-reachable IP-echo service still starts and advertises a correct external address once AutoNAT confirms one (or a configured address is set).
- IPv6 external addresses are representable.

chefsale:
Closes #2531.

## What

Removes the third-party `https://api.ipify.org` HTTP call from public-IP discovery (`Discovery::get_public_ip`) and replaces it with the libp2p-native external-address path plus an optional operator-configured static address.

## Why

`get_public_ip()` sat on the `Discovery::new` constructor path. When `advertise_address = true` it:
- **coupled startup to a third party** — ipify down/blocked → `NetworkManager::new` errors → the node doesn't boot;
- **trusted unauthenticated plaintext** from a CDN (DNS-poison / MITM could spoof the IP);
- **phoned home** on every startup (privacy leak for a privacy-oriented network);
- was **IPv4-only**;
- and was **redundant** — the fetched IP was only used as a *filter* on identify's `observed_addr` before manually calling `add_external_address`, which bypasses AutoNAT.

The correct source of truth was already wired: `identify` emits each peer-observed address as a `NewExternalAddrCandidate`, the AutoNAT v2 client probes it, and the swarm only emits `ExternalAddrConfirmed` for dial-back-reachable addresses. The ipify filter was a weaker, untrusted parallel path.

## Changes

- Delete `Discovery::get_public_ip()` and the `AdvertiseState` IP/port filter; `Discovery::new` is no longer `async`/fallible.
- Remove the identify "Branch 1" ipify match + manual `add_external_address(observed_addr)`; rely on AutoNAT-confirmed candidates.
- Add `DiscoveryConfig.external_address: Vec<Multiaddr>` (`#[serde(default)]`, empty) — seeded directly into the swarm via `add_external_address` at init when `advertise_address` is set. Supports IPv4 **and** IPv6.
- Add a `merod init --external-address <MULTIADDR>` flag.
- Drop the now-unused `reqwest` dependency from `calimero-network` (its only user).

## Acceptance (per #2531)

- [x] No outbound call to `api.ipify.org` (or any IP-echo service) anywhere in the node.
- [x] `calimero-network` no longer depends on `reqwest`.
- [x] A node with `advertise_address = true` and no reachable IP-echo service still starts and advertises a correct external address once AutoNAT confirms one (or a configured address is set).
- [x] IPv6 external addresses are representable (`Multiaddr`).

## Back-compat

`advertise_address` defaults to `false` and `external_address` defaults to empty, so the common path and existing config files are unaffected. For nodes that opt in, behavior shifts from "ipify-detected IP filter" to "AutoNAT-confirmed observed address (+ optional configured address)" — strictly more trustworthy. The opportunistic relay reservation still covers the NAT'd case until AutoNAT confirms a direct address.

## Testing

- `cargo build -p calimero-network-primitives -p calimero-network -p merod` ✓
- `cargo build -p calimero-network --tests` ✓
- `cargo fmt --check` ✓ / `cargo clippy` clean (no new warnings)
- `cargo test -p calimero-network --lib` → 67/67 green
- `cargo test -p calimero-network-primitives --lib` green (incl. autonat_v2 `ExternalAddrConfirmed` tests; `test_dynamic_protocol_change` is a pre-existing `libp2p-swarm-test` timing flake — passes on rerun, unrelated to this change)

---

> [!NOTE]
> **Medium Risk**
> Changes how advertised external addresses are chosen (trust and boot reliability), but defaults stay off and AutoNAT is the safer path; misconfigured static externals could advertise wrong dial addresses.
> 
> **Overview**
> Removes **startup-time public IP discovery** via `api.ipify.org` (`reqwest`) and the **`AdvertiseState` / identify filter** that promoted a single peer-observed address after matching that IP. **`Discovery::new`** is synchronous again; nodes no longer fail boot when an IP-echo service is unreachable.
> 
> **External reachability** now follows libp2p: identify supplies **AutoNAT v2 candidates** only (no manual `add_external_address` on observed addr). Operators can set **`discovery.external_address`** (and **`merod init --external-address`**) to seed confirmed externals at startup when **`advertise_address`** is on, with **filtering** for loopback / unspecified / link-local and warnings for private ranges. **`reqwest`** is dropped from **`calimero-network`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9595d33fd8f2f676ab95b35ff01be9a97cbdc640. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1394,7 +1394,6 @@ dependencies = [
  "owo-colors 3.5.0",
  "prometheus-client",
  "rand 0.8.5",
- "reqwest 0.12.28",
  "serde",
  "serde_json",
  "tokio",
--- a/crates/merod/src/cli/init.rs
+++ b/crates/merod/src/cli/init.rs
@@ -108,8 +108,10 @@ pub struct InitCommand {
 
     /// Static external multiaddr(s) to advertise (e.g.
     /// `/ip4/203.0.113.7/tcp/2428`). Seeded directly into the swarm's
-    /// external-address set when `--advertise-address` is set; otherwise
-    /// external addresses are discovered via identify + AutoNAT v2.
+    /// external-address set at startup; requires `--advertise-address`.
+    /// AutoNAT v2 always additionally discovers and confirms reachable
+    /// addresses regardless of this flag. Non-routable values (loopback /
+    /// unspecified / link-local) are ignored.
     #[clap(long = "external-address", value_name = "MULTIADDR")]
     pub external_address: Vec<Multiaddr>,
 
--- a/crates/merod/src/cli/init.rs
+++ b/crates/merod/src/cli/init.rs
@@ -106,6 +106,13 @@ pub struct InitCommand {
     #[clap(overrides_with("no_mdns"))]
     pub advertise_address: bool,
 
+    /// Static external multiaddr(s) to advertise (e.g.
+    /// `/ip4/203.0.113.7/tcp/2428`). Seeded directly into the swarm's
+    /// external-address set when `--advertise-address` is set; otherwise
+    /// external addresses are discovered via identify + AutoNAT v2.
+    #[clap(long = "external-address", value_name = "MULTIADDR")]
+    pub external_address: Vec<Multiaddr>,
+
     #[clap(
         long,
         default_value = "3",
@@ -270,6 +277,7 @@ impl InitCommand {
                 DiscoveryConfig::new(
                     mdns,
                     self.advertise_address,
+                    self.external_address.clone(),
                     RendezvousConfig::new(self.rendezvous_registrations_limit),
                     RelayConfig::new(self.relay_registrations_limit),
                     AutonatConfig::new(
--- a/crates/network/Cargo.toml
+++ b/crates/network/Cargo.toml
@@ -41,7 +41,6 @@ multiaddr.workspace = true
 owo-colors.workspace = true
 prometheus-client.workspace = true
 rand.workspace = true
-reqwest.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 tokio = { workspace = true, features = ["io-util", "macros"] }
--- a/crates/network/primitives/src/config.rs
+++ b/crates/network/primitives/src/config.rs
@@ -167,6 +167,15 @@ pub struct DiscoveryConfig {
 
     pub advertise_address: bool,
 
+    /// Operator-configured external addresses, seeded directly into the
+    /// swarm's confirmed external-address set at init (gated on
+    /// `advertise_address`). This is the deterministic alternative to
+    /// AutoNAT-based discovery for known-static-IP / hosted deployments,
+    /// and supports both IPv4 and IPv6. When empty, external addresses
+    /// are discovered solely via identify + AutoNAT v2 confirmation.
+    #[serde(default)]
+    pub external_address: Vec<Multiaddr>,
+
     pub rendezvous: RendezvousConfig,
 
     pub relay: RelayConfig,
@@ -179,6 +188,7 @@ impl DiscoveryConfig {
     pub const fn new(
         mdns: bool,
         advertise_address: bool,
+        external_address: Vec<Multiaddr>,
         rendezvous: RendezvousConfig,
         relay: RelayConfig,
         autonat: AutonatConfig,
@@ -186,6 +196,7 @@ impl DiscoveryConfig {
         Self {
             mdns,
             advertise_address,
+            external_address,
             rendezvous,
             relay,
             autonat,
@@ -198,6 +209,7 @@ impl Default for DiscoveryConfig {
         Self {
             mdns: true,
             advertise_address: false,
+            external_address: Vec::new(),
             rendezvous: RendezvousConfig::default(),
             relay: RelayConfig::default(),
             autonat: AutonatConfig::default(),
--- a/crates/network/src/discovery.rs
+++ b/crates/network/src/discovery.rs
@@ -1,6 +1,4 @@
-use core::net::Ipv4Addr;
-use core::time::Duration;
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap};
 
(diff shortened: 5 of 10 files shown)


chefsale:
Addressed the reviewer findings in 9595d33f:

**🟡 External addresses seeded without validation** (raised on `lib.rs:97` and `config.rs:174`) — configured addresses do bypass AutoNAT dial-back, so seeding them blindly could advertise garbage to the whole swarm. Added `is_seedable_external_address()`: addresses a remote peer can provably never reach — loopback, unspecified (`0.0.0.0`/`::`), link-local (`169.254/16`, `fe80::/10`) — are **skipped with a warning** rather than poisoning the swarm's external set. Private RFC-1918 / IPv6 ULA (`fc00::/7`) ranges are **kept but warned** (legitimate for same-network/overlay deployments, a common slip otherwise). Covered by two unit tests (`external_address_tests`). I went with skip-and-warn rather than hard-reject so a config typo can't reintroduce the boot-failure class this PR set out to remove.

**💡 `--external-address` help text** — corrected: AutoNAT v2 discovers/confirms reachable addresses regardless of the flag; the flag only controls static pre-seeding (and requires `--advertise-address`).

Skipped the doc-bot nudges (AGENTS.md / CONTRIBUTING.md) — this is an internal IP-discovery change with no architecture-doc surface.

Issue #2673, opened 2026-06-03
SharedStorage: local causal writer-set resolution under concurrent rotations (P4)


chefsale:
Carved out of #2230 (SharedStorage v2), which closed with items 1–4 done. This tracks the single remaining deferred refinement: **local causal resolution of the writer set under concurrent rotations**.

## Status: not a security gap

The merge-time verifier is the security boundary and is fully causal. A received rotation/write is authorized against the causally-correct writer set via `writers_at(parents)` (`SharedStorage` rotation via the rotation log; `SharedMember` via `effective_writers` / `writers_at(anchor_log, delta.parents)` — see `crates/storage/src/interface.rs:1503`). Forged root-state deltas can't influence it. That all landed in #2601 / #2612 / #2655.

## What's deferred

The **local execution gate** — the writer-set used by `insert` / `rotate_writers` *before* the action syncs — is best-effort, not causal. `SharedStorage::current_writers` (`crates/storage/src/collections/shared.rs:399`, documented at `:368`) resolves the writer set from the **most-recently-appended** rotation-log entry, which is this node's *insertion order*, not causal order. Under concurrent rotations arriving interleaved, a local `insert`/`rotate_writers` can accept or reject against a set the merge-time verifier would resolve differently.

The net effect is a local liveness/UX wrinkle (a write may be locally gated against a slightly stale set), not an authorization bypass — the merge check still rejects anything truly unauthorized.

See the in-code notes:
- `crates/storage/src/collections/shared.rs:378-385` ("best-effort under concurrent rotations … full local causal resolution is deferred to … P4")
- `crates/storage/src/interface.rs:227`
- `crates/storage/src/rotation_log.rs:110`

## Scope

- Give the local execution gate access to enough DAG/causal context to resolve the writer set with `writers_at`-equivalent semantics during local execution (currently no DAG context is available there).
- Decide the behavior when local and causal resolution disagree under concurrent rotation (defer the write? optimistic-then-reconcile?).
- Migration story for any wrappers created before the change.
- e2e: two nodes issue concurrent rotations; confirm local gating matches the merge-time outcome (or fails safe/converges) — none of this is locally checkable, needs merobox.

## Out of scope
- DAG-causal verification rework → #2233
- Threshold / multi-sig, cross-context writer sets

**Anchor:** builds on the merged #2601 (`SharedStorage`) and #2655 (`SharedMember` anchor) writer-set machinery.

PR #2696, opened 2026-06-04
test(e2e): drop the settling rotation now that #2673 converges writers() directly

chefsale:
Follow-up to #2673 (merged). Simplifies the concurrent-rotation e2e (`shared-storage-concurrent-rotation-converge.yml`, added in #2671).

## Why

That workflow issued a causally-latest **"settling" rotation** by node-3 after the two concurrent rotations, *solely* to make the app-visible writer set deterministic across nodes. It had to: the old gate read the rotation log's most-recently-**appended** entry (per-node insertion order), so without a settler two nodes could read `{A,C}` vs `{B,C}` from the same log.

**#2673 removed that need** — the local gate now resolves by `max-(HLC, signer)`, which is insertion-order invariant, so every node resolves the *same* set straight from the two concurrent rotations.

## Changes

- **Remove the settling rotation** + its `wait_for_sync` + settle (3 steps; 72 → 69).
- The converged set is `{A,C}` **deterministically**: node-1's rotation runs ~20s after node-2's (window 2 vs window 1), so it carries the later HLC and wins the concurrent tiebreak — the window gap dwarfs any same-host clock skew. The existing `{A,C}` / B-revoked assertions and the revocation bonus hold unchanged.
- **Add a winner-agnostic convergence assertion** — all three nodes agree on every writer's membership (`a_n1==a_n2==a_n3`, same for B and C). This is the *direct* demonstration of the #2673 property the settling rotation used to manufacture: pre-#2673 `a_n*`/`b_n*` would disagree across nodes; the test now asserts they don't.
- Rename the now-misleading `after-settle` value to `by-writer-a-final`; refresh the header docs to explain the #2673 convergence instead of the settling workaround.

Net: the e2e is shorter, faster, and now asserts the convergence guarantee *as the concurrent rotations produce it* rather than after papering over it.

---

> [!NOTE]
> **Low Risk**
> Test-only workflow YAML changes; no production auth, sync, or storage logic is modified.
> 
> **Overview**
> Updates the **shared-storage concurrent rotation** e2e so it matches **#2673**: nodes no longer need a follow-up “settling” rotation to agree on `writers()`.
> 
> The workflow **drops** node-3’s settling `shared_rotate_writers`, its `wait_for_sync`, and the extra settle. **Window 1** now has node-2 rotate **`{A,B} → {A}`** (not `{B,C}`), with **`shared_set` before** that rotation so B can still write during the split. After heal, assertions rely on **#2673’s `max-(HLC, signer)`** resolution instead of manufacturing a single last log entry.
> 
> **New checks**: all three nodes must **agree pairwise** on A/B/C writer flags (the old insertion-order bug), plus a **hermetic** block that every node **keeps A and revokes B** regardless of tiebreak. The positive-control shared value is renamed to **`by-writer-a-final`**. Header and inline comments explain why settling is gone and how the two differing concurrent rotations still stress cross-node convergence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 656f747d0c4f24ed8b49083a3b3dc0f2c8d78e43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/scaffolding-e2e/workflows/shared-storage-concurrent-rotation-converge.yml
+++ b/apps/scaffolding-e2e/workflows/shared-storage-concurrent-rotation-converge.yml
@@ -27,21 +27,23 @@ description: >
   the rotating node is always connected (callable) yet the node it
   rotates on never receives the other rotation:
 
-    1. Disconnect node-1 and node-3. node-2 (still connected)
-       rotates {A,B} -> {B,C}. Neither node-1 nor node-3 sees it.
+    1. Disconnect node-1 and node-3. node-2 (still connected) rotates
+       {A,B} -> {A}, dropping B (itself). Neither node-1 nor node-3
+       sees it.
     2. Disconnect node-2; connect node-1. Now only node-1 is online,
-       and it is still at H (it was offline when {B,C} was created,
-       and {B,C}'s only holder — node-2 — is now offline, so it
-       can't be fetched). node-1 rotates {A,B} -> {A,C}: causally
-       CONCURRENT with {B,C}.
+       and it is still at H (it was offline when node-2 rotated, and
+       the only holder of that rotation — node-2 — is now offline, so
+       it can't be fetched). node-1 rotates {A,B} -> {A,C}: causally
+       CONCURRENT with node-2's {A}.
     3. Reconnect node-2 and node-3. wait_for_sync converges all
        three — the root-hash convergence assertion — with both
        rotations merged.
     4. Assert every node resolves the SAME writer set directly from
        the two concurrent rotations — no settling rotation needed
-       (see the #2673 note below). It resolves to {A,C}: A and C are
-       writers, B is NOT — retroactive revocation of B holds under
-       concurrency, on every node.
+       (see the #2673 note below). Both rotations dropped B and kept
+       A, so the resolved set retains A and revokes B on every node —
+       retroactive revocation of a baseline writer under concurrency,
+       independent of which rotation wins the tiebreak.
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed in `656f747d` — removed the timing dependence entirely rather than relying on it.

The concern was fair in spirit. (Strictly, on GitHub Actions all merobox containers run on one host VM and share the kernel `CLOCK_REALTIME` — there's no inter-container skew, and the choreography forces a ~20s wall-clock gap between the two rotations, so the winner was deterministic. The "clock skew" in the comment is a multi-host concept that doesn't apply here.) But "depends on a 20s gap" reads fragile, so I made it winner-independent:

- The two concurrent rotations are now **node-1 → `{A,C}`** and **node-2 → `{A}`** (was `{B,C}`). **Both drop B and keep A**, so every node retains A and revokes B *regardless of which rotation wins the HLC tiebreak*. The concrete A/B assertions are now hermetic.
- The sets still **differ** (`{A,C}` vs `{A}`), so the resolution is still order-dependent under the old gate and must collapse identically — the #2673 property under test is unchanged. C's membership is the only HLC-dependent bit, and it's covered by the winner-agnostic agreement assertion (`c_n1==c_n2==c_n3`).
- The revocation bonus (node-2/B rejected, node-1/A succeeds) is hermetic too; node-2's during-split `shared_set` was moved before its rotation (which drops B).

So: no hard-coded winner, no reliance on rotation timing, and full convergence + retroactive-revocation coverage retained.

PR #2697, opened 2026-06-04
refactor(node/state_delta): extract crypto + verify submodules from the god-file handler

chefsale:
## What

First increment of splitting the 3326-line `crates/node/src/handlers/state_delta/mod.rs` god-file into focused submodules. **Pure, behavior-preserving code movement** — `mod.rs` changes are +4 wiring lines / −264 moved lines, zero logic changes.

| New module | Contents |
|---|---|
| `crypto.rs` | `lookup_group_key_with_wait` (+ its key-lookup constants) and `decrypt_delta_actions`, with their unit tests. Leaf helpers with **no external callers** — re-imported into `mod.rs` so call sites are unchanged. |
| `verify.rs` | `GroupIdCheck` (+ `Debug`) and `verify_position_group_id_matches_context` — the anti-bypass `group_id` check shared with `sync::manager` / `sync::delta_request`. Re-exported `pub(crate)` so the existing `crate::handlers::state_delta::{...}` import paths and the in-place `group_id_check_tests` still resolve. |

## Why incremental

`state_delta` is consensus-critical sync code. Rather than cram five scattered extractions into one hard-to-review PR, this lands two cohesive, low-risk modules and establishes the pattern. Remaining clusters for follow-up increments: event execution (parse/execute/emit), governance buffering/replay, and delta-store setup.

## Validation

- `cargo check -p calimero-node` clean.
- node lib tests: **274 passed** (relocated crypto tests run under `state_delta::crypto::tests`; verify tests pass in place via the re-export).
- node `sync_sim`: **314 passed**.
- `mod.rs` diff is exactly 4 added lines (the `mod` + `use` wiring) and 264 removed — verified no stray logic changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Pure move/refactor with re-exports preserving call sites; consensus-critical code but no logic changes per the diff.
> 
> **Overview**
> Splits the oversized **`state_delta/mod.rs`** handler into two focused submodules while keeping behavior and public paths unchanged.
> 
[...]

PR #2699, opened 2026-06-04
refactor(node/state_delta): extract events submodule (increment 2)

chefsale:
## What

Increment 2 of splitting `crates/node/src/handlers/state_delta/mod.rs`. **Stacked on #2697** (review/merge that first). Moves the event-execution cluster into `events.rs` — pure, behavior-preserving code movement (`mod.rs`: **+5** wiring lines / **−292** moved).

Moved (all internal helpers; re-imported into `mod.rs` so call sites are unchanged):

| Item | Role |
|---|---|
| `parse_events_payload` | JSON → `ExecutionEvent` decoding |
| `execute_event_handlers_parsed` | run app handlers for a parsed payload |
| `execute_cascaded_events` | handler execution for peer-cascaded parent deltas |
| `emit_state_mutation_event_parsed` | push `StateMutation` to WebSocket clients |
| `CascadeOutcome` | result struct (fields `pub(super)` so `mod.rs` can read them) |

The `parse_events_payload` unit tests stay in `mod.rs`'s test module and resolve through the re-import. No logic changes.

## Validation

- `cargo check -p calimero-node` clean (no new dead-code warnings).
- node lib tests: **274 passed**.
- node `sync_sim`: **314 passed**.
- `mod.rs` diff is exactly 5 added lines (`mod events;` + the `use`) and 292 removed.

## Progress

`mod.rs`: 3326 → ~2774 LOC across this + #2697. Remaining increments: governance buffering/replay, delta-store setup. The two orchestrators (`handle_state_delta`, `apply_authorized_state_delta`) stay in `mod.rs`.

---

> [!NOTE]
> **Medium Risk**
> Changes how handler failures and bad event blobs interact with DAG apply and persistence on sync/replay paths; scope is localized but affects convergence and at-least-once handler semantics.
> 
> **Overview**
> Moves state-delta **event execution** (`parse_events_payload`, handler runners, cascaded handler batching, WebSocket `StateMutation` emit, and `CascadeOutcome`) into new `events.rs`, with `mod.rs` re-importing them and passing `NodeClient` + `ContextClient` instead of a bundled `NodeClients`.
> 
> Alongside the split, behavior shifts on the apply/replay paths: **cascaded** handler runs no longer abort the batch on `execute_event_handlers_parsed` errors (log, keep the events blob for restart replay); **WebSocket** emission is explicitly best-effort (`()` instead of `Result`); and after a delta is **applied**, a present-but-invalid events JSON blob is cleared via `mark_events_executed` so restarts do not loop on the same corrupt payload (aligned with the cascaded-deserialize path).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c7430ac88de1099f9d02c49b72f7d1c4ef513e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Thanks for the review. Addressed in the latest push (rebased on the updated #2697):

**🟡 Corrupted doc markers in `execute_event_handlers_parsed` — fixed.** Good catch. These were pre-existing corruption (`**IMPORTANTMenu…`) carried in verbatim by the move; cleaned up to `**IMPORTANT**:` / `**Commutative**:` / `**No side effects**:` / `**Verdict**:` and added the missing blank `///` before the summary paragraph. Documentation-only.

**🟡 Glob `use super::*` — fixed.** Converted `events.rs` to explicit imports (consistent with the crypto/verify change in #2697), with `ensure_application_available` imported via `super::`.

**💡 Inner `?` in `execute_cascaded_events` violates the documented no-`?` contract — valid, deferring.** Nice catch — this is a genuine latent inconsistency (the inner `execute_event_handlers_parsed` always returns `Ok` today, so it never fires, but the `?` would abort the cascade loop mid-batch if that ever changed). It's pre-existing code moved verbatim, and the fix is a behavior change rather than a move, so I'd prefer to land it as a dedicated follow-up commit (match-and-log, per your suggestion) rather than muddy this refactor. Flagged for follow-up.

**📝 `emit_state_mutation_event_parsed` returns `Result<()>` but always `Ok` — pre-existing, deferring.** Changing the signature ripples to 3 call sites in `mod.rs` (`…?`); out of scope for a pure move. Reasonable cleanup for the same follow-up.

chefsale:
Addressed the review in a stacked follow-up, #2704 (kept separate so #2699 stays a pure code-move):

- 🟡 **Cascade inner `?`** → match-and-log, per your suggestion. A handler error now counts as "not all succeeded" so the events blob is kept for restart replay, honoring the documented log-and-continue contract.
- 💡 **`parse_events_payload` corrupt-blob loop** → real latent bug, good catch. The direct-apply path now clears a present-but-undeserializable events blob (once the delta is applied), consistent with the cascaded path; `events == None` is left untouched.
- 📝 **`emit_state_mutation_event_parsed` → `()`** → done, dropped the `?` at the three call sites.

**Declined:** 🟡 *`applied_current` set before the `app_available` check.* Your own analysis notes this is "actually the correct recovery path" — `handlers_executed_for_current` stays `false`, so the main flow re-runs handlers when the app becomes available. It's pre-existing, semantically-correct control flow in the consensus apply path, so I'd rather not churn it for a signal that behaves correctly. Happy to add a clarifying doc-comment if you'd prefer.

#2704 validation: node lib tests (274) + sync_sim (314) pass.

chefsale:
Addressed the latest review pass in 7c7430a:

- 💡 **`execute_cascaded_events` / `NodeClients` coupling** — good call. Changed the signature to `&NodeClient` + `&ContextClient` (matching `execute_event_handlers_parsed`), updated the six call sites, and removed **both** throwaway `NodeClients` constructions in `replay_buffered_delta` (also drops two client clones — it now borrows directly).
- 📝 **`parse_events_payload` doc comment** — added; documents the `None`-on-missing-or-corrupt return and how callers distinguish the two.
- 🟡 **`applied_current` set before the app-availability check** — kept the logic (it's the correct recovery path: `handlers_executed_for_current` stays `false`, so the caller's `applied && !handlers_already_executed` guard re-attempts handlers once the app is available) and **documented the DAG-state-only semantics + caller contract inline**, per your suggestion that a comment is the right resolution.

No behavior change. Validated: node lib tests (281) + sync_sim (314) pass.

Issue #2687, opened 2026-06-04
Design: calimero-components (Ownable/AccessControl/Pausable) + operation-aware ACLs (OpMask)


chefsale:
**Design doc for the CRDT-aware access-control components library, consolidated from #2557.**

This issue holds the design for `calimero-components` (`Ownable` / `AccessControl` / `Pausable`) and its protocol-level extension path (operation-aware ACLs via `OpMask`, §8), which is the **collection-scope, op-granular slice of #2541** (Research: Unified fine-grained permissions). #2541 stays open as the broader umbrella (governance/role unification, TEE delegation, account principals, path/field scope); this issue is the concrete, mostly-already-built slice that makes components a real security boundary rather than advisory sugar.

The headline property: **enforcement lives at merge, never in the method** — so a malicious peer cannot bypass a guard by hand-crafting a state delta around the wasm (the original concern that motivated this). Most of the enforcement substrate is already merged (#2588, #2230/#2601/#2612, #2655, #2665).

Doc also lives in-tree at `docs/design/calimero-components.md`.

- Supersedes: #2557
- Slice of: #2541
- Depends on: #2544 (field-id registry, P0)
- Related: #2673 (local causal resolution)

---

# calimero-components — CRDT-aware access-control components

Status: **Design** (issue #2557). Targets the merged enforcement substrate:
#2588 (collection guarding), #2230/#2601/#2612 (authenticated rotation),
#2655 (SharedMember anchor / retroactive revocation), #2665 (concurrent-rotation
convergence).

## 1. Motivation

App authors keep hand-rolling the same authorization patterns: an owner key, a
set of admins, a pause switch. Today they write them by hand against
`SharedStorage`/`UserStorage`, and the easy way to write them is **wrong**:

```rust
pub fn delete_entry(&mut self, k: String) -> app::Result<()> {
    require!(env::executor_id() == self.owner, "not owner"); // advisory only
    self.entries.remove(&k);
    Ok(())
}
```

That `require!` is **not a security boundary**. A malicious context member never
calls `delete_entry`; they run a patched node, craft a signed delta containing a
`DeleteRef` against the entry's entity, and gossip it. It reaches honest nodes
through the **sync/merge path**, which never runs the app's wasm. The check is
bypassed.

`calimero-components` exists to make the *correct* pattern the *easy* pattern:
ship `Ownable`, `AccessControl`, and `Pausable` as thin facades over
writer-set-guarded storage, so protected data physically lives inside an entity
[...]

Issue #2544, opened 2026-05-31
[SDK UX] Auto-inject collection field names in #[app::state] (drop new_with_field_name)


chefsale:
## Summary
Have `#[app::state]` inject collection field names automatically, removing the stringly-typed `new_with_field_name("...")` requirement (and the typo footgun).

## Current friction
Every `init()` repeats each field's name as a string literal that must exactly match the struct field. From `apps/nested-crdt-test/src/lib.rs:64-72`:

```rust
NestedCrdtTest {
    counters: UnorderedMap::new_with_field_name("counters"),
    registers: UnorderedMap::new_with_field_name("registers"),
    metadata: UnorderedMap::new_with_field_name("metadata"),
    metrics: Vector::new_with_field_name("metrics"),
    tags: UnorderedMap::new_with_field_name("tags"),
}
```

A typo (`"conters"`) doesn't fail to compile — it silently assigns the wrong deterministic ID.

## Proposed
```rust
NestedCrdtTest::default()
// or: NestedCrdtTest { counters: UnorderedMap::new(), .. }
```

## SDK work / effort
The `#[app::state]` macro already runs `__assign_deterministic_ids()` walking every field — extend that pass to set the field name on top-level collections, so plain `::new()` (and `#[derive(Default)]`) works. Small/medium.

Removes a whole class of silent-misconfiguration bugs.

chefsale:
Implemented and merged to `master` via PR #2572 (commit 29fb99e8, "feat(apps): drop new_with_field_name for plain ::new()").

The `#[app::state]` macro now auto-injects each collection field's name during `__assign_deterministic_ids()` — generating a `reassign_deterministic_id("<field_name>")` call per collection field, with the name derived from the struct field identifier at compile time (crates/sdk/macros/src/state.rs). App authors can now use plain `UnorderedMap::new()`; the string-literal typo footgun is eliminated. All example/migration apps were migrated off `new_with_field_name(...)`.

Issue #2541, opened 2026-05-31
Research: Unified fine-grained permissions model


chefsale:
# Research: Unified fine-grained permissions model

**Type:** Research / design spike (high-level — to be deepened later)
**Area:** Authorization, storage/CRDT, governance
**Related:** #2540 (multi-tenant read-write TEE nodes — the account/authz half overlaps with this)

## Problem

Calimero's authorization today is **coarse and split across three disjoint mechanisms**, with no fine-grained, per-operation, or per-collection control:

- **Context membership roles** — `Admin / Member / ReadOnly / ReadOnlyTee`. Global to the context; gate whether you can mutate at all, not *what*.
- **Governance capabilities** — `MemberCapabilities` bitmask (`MANAGE_MEMBERS`, `CAN_CREATE_SUBGROUP`, …). Group-management only, not data-plane.
- **Entity-level `StorageType`** — `Public / Frozen / User{owner} / Shared{writers}`. Controls *who* writes an entity, never *which operation*.

Concrete symptom: **you cannot express "write but not delete" on a map.** If an identity can write a collection, it can also delete from it. `Frozen` is all-or-nothing immutability; there's no append-only. Worse, the signed `Action` payload commits to data + the writer-ACL set but **not the operation type** — so a signature valid for `Update` is equally valid for `DeleteRef`. The cryptographic authorization is identical for write and delete.

There's also no account principal above contexts (see #2540), so multi-tenant authz has nowhere to attach.

## Governing principle: enforce at merge, not at the API

Calimero is a P2P CRDT system — peers exchange **signed deltas** directly and apply them at sync/merge time, often long after creation. An API-layer check is necessary for UX but **not sufficient**: a malicious/buggy peer can hand-craft a delta that never hit the API. The real authorization boundary must be **at apply/merge time, bound into the signature**.

We already do this for *membership*: `membership_status_at()` verifies the signer was a member at the op's causal cut. **The entire proposal is to generalize that one trusted mechanism from "was a member" to "had capability C for operation O on scope S."** Everything else is detail.

## Proposed model: one capability concept, three scopes, evaluated at the cut

Collapse the three disjoint mechanisms into a single primitive:

> **A permission is `(principal | role) × scope × operations`, stored as CRDT state in the governance DAG, evaluated at the op's causal cut, enforced at merge.**

- **Principals** — identity pubkey, optionally under an **account** (the #2540 account principal becomes just another principal here — the models unify).
[...]

PR #2700, opened 2026-06-04
feat(storage): access-control components (PermissionedStorage/Ownable)

chefsale:
## What

Adds a policy-parameterised access-control layer over the existing writer-set storage primitive, so app authors get ergonomic ownership/role guards without hand-rolling **bypassable** checks. First slice of the design doc tracked in #2687.

## Why

The easy way to gate a method is a `require!(executor == owner)` — which is **not a security boundary**. A malicious peer never calls the method: it hand-crafts a signed delta and gossips it; the delta is applied through the sync/merge path, which never runs the app's WASM. The guard is bypassed.

These components make the *correct* pattern the *easy* one: protected data lives inside a writer-set-guarded entity, so the authoritative check is the **merge-time signature verification** that already rejects a non-writer's delta.

## How — one base, derive the rest

- `Authorizer` trait + `Op` enum — the policy seam (pure `fn(who, op, writers) -> bool`).
- `PermissionedStorage<T, A = WriterSetAcl>` — wraps `SharedStorage<T>` and adds the policy `A`. It **delegates** all storage mechanics to `SharedStorage`; it deliberately does **not** refactor that merged, security-critical type. The default policy makes `PermissionedStorage<T, WriterSetAcl>` the group-writable form.
- `Ownable<T> = PermissionedStorage<T, OwnerAcl>` — single-owner cell with `owner`/`only_owner` and `transfer_ownership` (an authenticated writer-set rotation — the capability `UserStorage` lacks).

The `Authorizer` is documented as **fail-fast UX** that mirrors what merge enforces; the security boundary is unchanged (`interface.rs` signature check against the writer set).

## Two normalizers also taught the new wrappers

- **State macro** (`is_collection_type`): so a wrapped collection gets a deterministic id and converges — without it the inner wrapper keeps a random id and split-brains.
- **wasm-abi normalizer**: surfaces `Ownable`/`PermissionedStorage` as their underlying `SharedStorage` ABI shape — without it the ABI emitter panics `unknown generic type` at build.

## Tests

- 5 storage unit tests (deterministic id, reassign canonicalisation, writer-gate, owner gating, transfer rotates).
- `apps/components-demo` example with 4 `TestHost` tests: owner-only write, **non-owner rejected**, **transfer moves control** (new owner writes, old owner can't), guarded-map round-trip.
- Full storage suite green (577 passed, 0 failed); fmt + clippy clean; wasm builds.

## Deliberately deferred (follow-ups on this base)

- **`AccessControl`** — roles as nested guarded `Map<role, SharedStorage<Set<PublicKey>>>`; per-key anchor ids make it its own PR.
- **`Pausable`** — inline custom-`Mergeable` convergence unproven; advisory-only, low value.
- **`renounce_ownership`** — `SharedStorage.frozen` is genesis-immutable (no setter), so omitted rather than faked.

Design doc included at `docs/design/calimero-components.md`.

Tracking: #2687 · slice of #2541 · depends on the #2544 field-id registry (this PR follows the existing string-match convention the registry will later replace).

---

> [!NOTE]
> **Medium Risk**
> Touches authorization ergonomics and deterministic ID / ABI paths for new state field types; enforcement still delegates to existing merge-time signature checks on writer sets.
> 
> **Overview**
> Introduces **`PermissionedStorage`** and **`Ownable`** in `calimero-storage`: a policy layer (`Authorizer`, `Op`, `WriterSetAcl` / `OwnerAcl`) on top of existing **`SharedStorage`**, with API-side `guard` / `only_owner` / `transfer_ownership` while **merge-time writer-set verification stays the real boundary**.
> 
> The **`#[app::state]`** macro and **wasm-abi** normalizer now treat **`Ownable`** and **`PermissionedStorage`** like guarded collections (deterministic field IDs and `SharedStorage` ABI shape). Adds **`apps/components-demo`** (owner config + group settings) with **`TestHost`** tests, workspace wiring, and **`docs/design/calimero-components.md`** for the broader components roadmap (#2687).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8d4b49ec09f7c00b1146e87ca93e839b44a849d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Thanks for the reviews — pushed `18f88d1` addressing the CI failure and the substantive findings.

**CI fix (the actual `Rust` failure):** `guard()` was missing the `# Errors` doc section the crate's denied `missing_errors_doc` lint requires. Added. (My local clippy false-passed on its incremental cache — re-ran with the changed file invalidated.)

**Addressed:**
- **API/merge guard consistency** (raised on `rotate_writers` and `transfer_ownership`): `rotate_writers` now calls `guard(Op::Admin)?` and `transfer_ownership` calls `only_owner()?` up front, so each method enforces its own contract and the fail-fast gate agrees with the merge-time check rather than relying on the caller or solely on the inner `SharedStorage`.
- **`OwnerAcl`/`WriterSetAcl` duplication:** `OwnerAcl::authorize` now delegates to `WriterSetAcl::authorize` so the membership predicate can't drift.
- **Single-owner invariant:** `owner()` now `debug_assert!`s at most one writer (the blessed constructors are `new_owned_by`/`new_owned_by_caller`).
- **"False sense of security" / threat model:** added a **Security model** section to the module doc — access control holds only for state stored *inside* the wrapper; a plain field is writable by any member regardless of an `only_owner()` guard.
- **Authorizer `&self` vs associated fn:** reconciled the design doc to the implementation (associated function — policies are zero-sized markers, not stateful values).

**Intentionally not changed:**
- Making `PermissionedStorage::<_, OwnerAcl>::new` unavailable: it's a shared inherent method on the base type (Ownable is a type alias), so it can't be hidden per-policy without a newtype. The `debug_assert` + blessed `new_owned_by*` constructors are the guard instead.
- `get_config` returning `Option`: the example is illustrative and `LwwRegister<String>::default()` is the empty string (a defined value, not a panic), so reading before first write returns `""`.

chefsale:
Pushed `f8d4b49` for the round-2 comments. These partly conflicted with round 1 (round 1 asked to add `only_owner()` inside `transfer_ownership`; round 2 flagged the resulting double-check), so rather than ping-pong I adopted one consistent rule and applied it everywhere:

> **Every mutating method applies the policy once via `guard(op)`, then delegates to the inner `SharedStorage`, which performs the authoritative check that peers re-verify at merge.**

- **`insert` now calls `guard(Op::Write)`** — the important one. Without it a custom `Authorizer` that restricts writes more narrowly than membership (the op-mask policies in §8) would be silently bypassed, defeating the seam. ✅
- **`transfer_ownership` drops the redundant `only_owner()`** — it's owner-gated through `rotate_writers` (`guard(Op::Admin)`), which for `OwnerAcl` *is* the owner check. One gate per method, no same-method double-check. ✅
- **`rotate_writers`** keeps its single `guard(Op::Admin)`; comment clarified that the inner check is authoritative (the guard is the policy-aware API gate, not a competing enforcement path — they check different layers, so "drift" isn't a correctness risk). ✅
- **components-demo**: noted `non_owner_cannot_set_config` exercises the API guard only (merge enforcement needs a 2-node e2e per §6.3), and that `get_config` returns the empty-string `LwwRegister` default before first set. ✅

Net effect: the policy gate (`guard`) is now uniformly the single API-surface check across `insert`/`rotate_writers`/`transfer_ownership`, and the inner `SharedStorage` remains the authoritative merge-equivalent boundary. Round-2 CI was green; storage clippy/tests + app tests pass locally.

chefsale:
The 3 red checks are all **sync-resilience e2e** (`network-partition-symmetric`, `peer-restart`) — not required checks, and a pre-existing flake unrelated to this change. The log shows the baseline converged and only the *post-partition relay re-convergence* timed out:

\`\`\`
✓ Assertion #1: contains({{baseline_n2}}, 'pre_partition')      ← normal sync converged
... network partitioned ...
✗ Sync verification failed after 150.31s (249 attempts)
❌ Step 'Wait for sync to catch up post-partition (via relay)' failed
\`\`\`

This PR is purely additive storage code (new `permissioned` module + a macro/ABI whitelist arm + a new example app); it can't touch merod's partition/relay-recovery path, and these scenarios run the `kv-store` app, not the new one. **Required checks (Rust / Lint / Build Binaries) and all other e2e are green.** Re-ran the failed jobs to clear the flake.

Issue #2684, opened 2026-06-04
feat(sdk,runtime): opt-in WASM read/write method intent annotations (parallel-read accelerator)


chefsale:
Part of the per-context-guard epic #2022. **This is an opt-in optimization, not a correctness prerequisite** — see #2685 for how the RwLock work stays safe (and behavior-neutral) without it.

## Goal

Let a WASM module declare, per method, whether it is read-only or mutating, and have the runtime read that intent **before** execution. This is the up-front signal that lets #2685 take a *shared read lock* for read-only calls (enabling parallel reads on one context) instead of always taking the exclusive write lock.

## Why it's needed (and only as an accelerator)

#2685 swaps the per-context lock to an `RwLock`, but an `RwLock` must choose read-vs-write **at acquisition time — before the method runs**. Today the node only learns whether a call mutated state *after* execution, by inspecting `outcome.artifact` (`crates/context/src/handlers/execute/mod.rs:787`). So there is no up-front signal to base the read/write choice on.

With #2685's fail-safe default (unknown intent → write lock), the absence of this signal is *safe* — it just means every call takes the exclusive lock and there is no parallelism. This issue supplies the signal that flips read-declared methods onto the shared read lock. So:

- **Without this issue:** #2685 is correct but produces no observable change (everything defaults to write lock).
- **With this issue:** read-declared methods run in parallel; everything else is unchanged.

It is therefore a pure, incremental optimization — nothing depends on it for correctness.

## Why not method names

The original epic proposed classifying by name prefix (`get_`/`query_`/`view_` → read). Unsafe: method names are app-author conventions, not enforced — a `get_x` can legally mutate. Misclassifying a writer as a reader → it takes a read lock → a concurrent writer runs on the same context → non-deterministic deltas / split-brain. The intent must be a machine-checked, compiled-in declaration.

## Scope

- [ ] SDK macro surface to mark a method read-only (e.g. an attribute on `#[app::logic]` methods, working name `#[app::view]`); **default is mutating** when unannotated (fail-safe).
- [ ] Encode the intent in the module ABI / method metadata.
- [ ] Runtime reads the metadata and exposes per-call read/write intent to the context execute path, where #2685 consumes it.
- [ ] Validation / defense-in-depth: a method declared read-only that nonetheless produces a non-empty mutation artifact is rejected post-execution (catches a lying or stale annotation before it can corrupt state under a read lock).

## Acceptance criteria

[...]

Issue #2022, opened 2026-02-17
perf(context): Use RwLock for context operations to enable parallel reads


xilosada:
## Problem

All context operations (including read-only queries) share a single `Mutex`:

```rust
// crates/context/src/lib.rs:33-36
struct ContextMeta {
    meta: Context,
    lock: Arc<Mutex<ContextId>>,  // Single mutex for ALL operations
}
```

**Impact**:
- Read-only queries wait for writes to complete
- Multiple concurrent reads are serialized unnecessarily
- During sync (which holds lock), all queries are blocked
- Throughput limited to ~1 request at a time per context

## Benchmark Scenario

With 100 concurrent read requests on same context:
- **Current (Mutex)**: ~100 * 10ms = 1000ms total (serialized)
- **With RwLock**: ~10ms total (parallel)

## Proposed Solution

1. Replace `Mutex` with `RwLock`:

```rust
struct ContextMeta {
    meta: Context,
    lock: Arc<RwLock<ContextId>>,
}

impl ContextMeta {
    fn read_lock(&self) -> impl Future<Output = RwLockReadGuard<ContextId>> {
        self.lock.read()
    }
    
    fn write_lock(&self) -> impl Future<Output = RwLockWriteGuard<ContextId>> {
        self.lock.write()
    }
}
```

2. Classify methods as read-only or mutating:

```rust
fn is_read_only_method(method: &str) -> bool {
    method.starts_with("get_") || 
    method.starts_with("query_") || 
    method.starts_with("view_")
}
```

3. Acquire appropriate lock type based on method classification

## Acceptance Criteria

- [ ] `ContextMeta.lock` changed from `Mutex` to `RwLock`
- [ ] Read-only methods acquire read lock (parallel)
- [ ] Mutating methods acquire write lock (exclusive)
- [ ] `__calimero_sync_next` uses write lock
- [ ] Benchmark shows parallel reads working
- [ ] No deadlocks introduced

## Risks & Mitigations

| Risk | Mitigation |
|------|------------|
| Write starvation | Use `RwLock` with write-preferring policy |
| Misclassified method | Default to write lock if uncertain |
| Existing code assumes exclusive | Audit all lock usages |

## Implementation Notes

- `tokio::sync::RwLock` is the appropriate choice
- Consider adding method metadata to WASM modules indicating read/write
- Could be opt-in initially (flag to enable)

**Priority**: P1 (High Impact, Medium Effort)

chefsale:
[...]

PR #2702, opened 2026-06-04
docs(design): per-method read/write intent — decisions resolved, ready for implementation

chefsale:
## What

Design doc for #2684 (epic #2022) — **design only, no code/behavior change.** Adds `docs/design/context-method-read-write-intent.md`.

Covers: how a WASM app declares per-method read/write intent in the ABI so the node can pick a shared read lock vs the exclusive write lock **before** execution, enabling parallel reads on one context. All four design questions now resolved (see §10 of the doc).

## Decisions

| Question | Decision |
|---|---|
| Intent source | Explicit `#[app::view]`, tri-state ABI (`ReadOnly`/`Mutating`/`Unspecified`) — infer-from-`&self` deferred as a later non-breaking flip |
| Read-only set cache | Sibling `BoundedCache<ApplicationId, Arc<ReadOnlySet>>` in `ContextManager` — keeps `calimero-runtime` decoupled from `calimero-wasm-abi` |
| Write rejection | Read-only `ContextStorage` wrapper — write/delete return an error under a read guard; mirrors existing read-only-member machinery shape |
| Emitter reliability (factual) | ✅ `emitter.rs` is a `syn::visit::Visit` source parser that already reads `#[app::state]` and the full method sig; sees `#[app::view]` directly — no proc-macro side-channel needed |

## Key design points

- **Pure opt-in optimization.** Fail-safe default: unknown/unannotated/old-module → write lock → exactly today's behavior. Old apps work unchanged, no recompile.
- **The load-bearing safety guarantee:** the existing post-hoc discard (`execute/mod.rs:1455`) is only safe under the *exclusive* lock. Under a *shared read* lock, a mis-declared writer would corrupt concurrent readers mid-run (execution mutates the shared Merkle index during the run). Therefore read-lock executions are blocked from writing **at the source** via the read-only storage wrapper; the post-hoc check is kept as defense-in-depth only.
- **Lock-site threading** (`mod.rs:108`): resolve the per-app read-only set from the embedded ABI at module-cache time; query before locking; cache miss/absent/unannotated → write guard (fail-safe). This is where `ContextGuard::Read` (deferred from #2698) is finally minted.
- **Atomic batches** default to the write guard up front, sidestepping the read→write upgrade gap.
- **Backward compatible:** old modules (no annotations) → `Unspecified` → write guard → exactly today. ABI `Method` field is `#[serde(skip_serializing_if)]`; old manifests still deserialize.

Part of #2022.

### DIFF
--- a/docs/design/context-method-read-write-intent.md
+++ b/docs/design/context-method-read-write-intent.md
@@ -1,0 +1,197 @@
+# Per-method read/write intent (parallel-read accelerator)
+
+**Status:** design / spike — for team review. Not implemented.
+**Issue:** #2684 · **Epic:** #2022 · **Consumes:** #2685 increment 2
+
+## 1. Problem
+
+The per-context lock is now an `RwLock` (#2698), but every caller still takes the
+exclusive *write* guard, so it behaves exactly like the prior mutex — no
+parallelism yet. To let read-only calls on a single context run concurrently,
+the node must choose **read vs write lock at acquisition time, before the method
+runs**.
+
+Today there is no up-front signal for that choice:
+
+- The node only learns whether a call mutated state **after** execution, by
+  inspecting `outcome.artifact` (`crates/context/src/handlers/execute/mod.rs:787`)
+  / `outcome.root_hash` (`mod.rs:1455`).
+- Method **names** are app-author conventions, not enforced — a `get_x` may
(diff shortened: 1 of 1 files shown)


chefsale:
Closing in favour of implementing directly. Design decisions are recorded in the epic (#2022) and in memory. Implementation PR follows.

PR #2703, opened 2026-06-04
test(e2e): adopt merobox partition_peers for concurrent-rotation (blocked on merobox 0.6.35)

chefsale:
**Draft — do not merge until [calimero-network/merobox#279](https://github.com/calimero-network/merobox/pull/279) is merged and merobox 0.6.35 is published.** Until then the `setup-merobox` floor check fails by design (the APT repo still serves an older "latest"). Staged now so it's ready the moment the release lands.

## What

merobox#278 adds `partition_peers` / `heal_peers` — a surgical libp2p partition that keeps a node **RPC-reachable** while cutting its peer traffic. This adopts it to simplify the concurrent-rotation e2e and bumps the merobox floor.

- **`shared-storage-concurrent-rotation-converge.yml`** — replace the `disconnect_node` window-juggling (isolate the two non-rotators, never call a disconnected node, swap which node is online between the two rotations) with a single partition:
  1. `partition_peers` node-1 from `{node-2, node-3}` — node-1 is libp2p-isolated from *both* (so node-3 can't relay node-2's rotation to node-1), yet all three stay callable.
  2. node-1 rotates → `{A,C}` and node-2 rotates → `{A}` — concurrent, both driven over RPC while partitioned.
  3. `heal_peers`; `wait_for_sync` converges.

  Drops 5 steps and all the connect/disconnect/await-recovery juggling. The hermetic assertions are unchanged (A retained + B revoked regardless of the HLC winner; winner-agnostic cross-node agreement; the node-level sim test remains the deterministic guarantee).

- **`setup-merobox/action.yml`** — `MIN_MEROBOX` 0.6.34 → 0.6.35 (the release shipping `partition_peers`), with the floor rationale refreshed.

## Follow-up
`member-removed-partition-window-reconcile.yml` (parked out of CI) can be re-enabled with `partition_peers` too — left as a separate change.

---

> [!NOTE]
> **High Risk**
> Adds new sync wire enum variants and changes HashComparison session teardown for Shared writer-set semantics; mixed-version clusters degrade gracefully but upgraded peers must coordinate on the new messages.
> 
> **Overview**
> This PR has two main threads: **e2e/CI** adopts merobox **`partition_peers` / `heal_peers`** and **node sync** adds **end-of-HashComparison rotation-log reconciliation** for hash-neutral `Shared` writer rotations.
> 
> **E2E & CI:** `shared-storage-concurrent-rotation-converge` drops the two-window `disconnect_node` / `connect_node` choreography in favor of one libp2p partition of node-1 from `{node-2, node-3}`, driving **both** concurrent rotations over RPC, then `heal_peers` and `wait_for_sync`. Docs/comments now treat HLC winner for C as **non-deterministic** (winner-agnostic assertions unchanged). **`setup-merobox`** raises **`MIN_MEROBOX` to 0.6.35** and documents the `partition_peers` floor (draft until merobox publishes that release).
> 
> **Sync (core#2716 / #2703):** New wire variants **`RotationLogSyncRequest`** / **`RotationLogSyncResponse`** exchange per-`Shared` anchor `RotationLog` blobs. At **end of an HC session**, the initiator runs **`reconcile_rotation_logs_with_peer`** (best-effort before root re-query); responders union incoming logs via **`union_received_rotation_logs`** and reply with **`collect_local_shared_rotation_logs`**. **`SyncManager`** ignores stray top-level `RotationLogSyncRequest`. Unit/sim tests cover divergent writer sets after HC-only catch-up and convergence after log union.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4cf7439105ae3adb1ce6d0b0792ac5422b7c9b7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed both notes in `da967aa2` (docs/clarity — no behavioral change):

1. **8s wait** — the reviewer's model is `disconnect_node`'s "wait for libp2p to mark the connection dead" (which is why it used 20s). `partition_peers` is different: the `DOCKER-USER` DROP rule is effective the instant the step returns and drops **every** packet between the pair, including on the already-established libp2p stream. So once `partition_peers` returns, no rotation can cross the partition regardless of keep-alive state — there's no dead-connection detection to wait for. The wait is pure settle margin for any pre-partition in-flight gossip (and the baseline already converged, so there's none in practice). Documented that and bumped 8s → 12s for conservatism.

2. **Sequential issuance** — correct, and now stated explicitly near the rotations: the orchestrator issues the two rotations sequentially, but that ordering is irrelevant — the partition prevents either node from receiving the other's rotation, so they're DAG-concurrent (each builds on head H) however they interleave. Concurrency comes from the partition, not from simultaneous issuance.

Re the **E2E Rust Apps failure**: that's expected and by design — this PR bumps `MIN_MEROBOX` to 0.6.35, which isn't published yet, so the `setup-merobox` floor check fails. It'll go green once [merobox#279](https://github.com/calimero-network/merobox/pull/279) merges and 0.6.35 ships; the PR is a draft until then.

chefsale:
Thanks — triaged the latest round:

- **"node-1 wins the HLC tiebreak via the ~20s window gap"** (stale-comment finding) — correct and now fixed: with `partition_peers` both rotations are issued back-to-back in one window, so there's no gap and the winner (whether C survives) is non-deterministic. The assertions were already winner-agnostic from #2696 (cross-node agreement + A-retained/B-revoked; no concrete `c==true`), so it was a comment-only correction.
- **8s / 12s / 20s settle wait** (re-posts) — addressed in the prior commit + comment: the `DOCKER-USER` DROP rule is the barrier and is effective the instant `partition_peers` returns (it drops packets on the already-established stream, not just future connections), so no rotation can cross once it's installed — the wait length is non-load-bearing settle margin, not a "wait for dead-connection detection". Bumped to 12s for headroom.
- **Sequential issuance** — addressed: there's now an explicit comment near the rotations that concurrency comes from the partition (neither node can receive the other's rotation), not from simultaneous issuance.

The **E2E failure remains expected** (the `MIN_MEROBOX=0.6.35` floor; merobox 0.6.35 isn't published yet — PR stays draft until merobox#279 ships).

chefsale:
**The "massive" failures are the expected `MIN_MEROBOX` floor check from a stale pre-release run — now resolved.**

Every failing job (the whole app-migration-e2e / e2e-rust-apps / sync-regression suite) failed in ~30s with:
> `merobox 0.6.34 is older than the required 0.6.35`

That run executed at **18:15 yesterday — before merobox 0.6.35 existed**, so `setup-merobox` installed 0.6.34 from APT and the floor I bumped (0.6.34 → 0.6.35) rejected it. This was the by-design "draft until the release ships" state, just fanned out across every merobox-dependent lane (they all share the floor).

merobox **0.6.35 is now published** — PyPI has it, and the APT repo (`calimero-network.github.io/merobox`) now serves `0.6.35-1`. I've **re-triggered the failed e2e lanes**; they'll install 0.6.35, the floor will pass, and the simplified rotation workflow will run for real — against a `partition_peers` primitive that's now end-to-end-proven in merobox's own CI (merobox#279's behavioral Docker workflow is green; it actually caught and got fixed for two real partition bugs).

On the two inline notes (both re-posts):
- **Settle-wait duration** — addressed earlier: the `DOCKER-USER` DROP is the barrier and is effective the moment `partition_peers` returns; the wait is non-load-bearing settle margin (12s). merobox#279's e2e now empirically confirms the partition blocks delivery.
- **"line ~490 says node-1 wins via the 20s gap"** — that stale comment was already removed in `b0022fe8`; the winner is non-deterministic and the assertions are winner-agnostic.

I'll confirm the re-run goes green and then this is ready to come out of draft.

chefsale:
**Important finding: with merobox 0.6.35 the floor passes, the workflow runs end-to-end, and it now red-flags a real bug — #2716.**

Run trace: partition_peers → node-1 rotates `{A,C}` → node-2 rotates `{A}` (genuinely concurrent now) → heal → **`Converge after concurrent rotations` (wait_for_sync) times out after 181s — the cluster does NOT converge.** That is exactly [#2716](https://github.com/calimero-network/core/issues/2716) ("concurrent SharedStorage writer rotations produce non-convergent CRDT merge — HC skips 3 entities permanently"). My diff here is only the workflow + the `MIN_MEROBOX` bump — no sync/rotation code — so this is the core bug, not the test.

Why this didn't show before: the merged `disconnect_node` version of this workflow **passed vacuously**. On the same Docker bridge, `disconnect_node`/L2 doesn't reliably stop container-to-container libp2p (the exact issue merobox#279 had to fix with `br_netfilter`), so the two "concurrent" rotations actually saw each other and resolved **sequentially** → trivially converged → green. A *real* partition (partition_peers + br_netfilter, now proven in merobox#279's own e2e) makes them truly concurrent and immediately surfaces #2716.

So this PR is doing its job: it turns a vacuous test into a deterministic **reproducer for #2716**. It can't go green until #2716 is fixed — I'm leaving it **draft + blocked on #2716** rather than weakening the convergence assertion (which would re-introduce the false pass).

(The `sync-resilience-peer-restart` / `auto-follow-backfill` failures in the same run are unrelated workflows I didn't touch — pre-existing flakes.)

PR #2705, opened 2026-06-04
feat(storage): SharedStorage→PermissionedStorage inversion + AccessControl component

chefsale:
Lands two stacked changes on master together (the child PR #2706 was merged into this branch, so this PR now carries both commits):

## 1. Inversion — `SharedStorage<T> = PermissionedStorage<T, WriterSetAcl>`

Inverts the base/derived relationship from #2700 so the access-control components form one hierarchy over a single mechanism, matching the intended design where `PermissionedStorage<T, A>` is the base.

- Rename the writer-set storage **struct** `SharedStorage` → `WriterSetCell` (the mechanism: value in its own `SharedMember` entity, signed writes verified at merge, rotatable writer set). Logic **byte-identical** — pure rename.
- `CrdtType::SharedStorage` enum variant **preserved** (wire/persisted-state compatibility).
- `PermissionedStorage<T, A>` wraps `WriterSetCell`; add `pub type SharedStorage<T> = PermissionedStorage<T, WriterSetAcl>`.
- **Source-compatible**: every call site uses the 1-arg `SharedStorage<T>` form, so no app/node/test edits.

## 2. `AccessControl` — merge-enforced role registry

Third component from #2687 (after `Ownable`/`PermissionedStorage`).

- Backed by a single writer-set-guarded map whose **writer set is the admin tier** (OpenZeppelin-style single admin tier: any admin manages any role).
- `grant`/`revoke` record a last-writer-wins present-flag per `(role, member)` — no tombstone; `has_role` is the app guard's lookup.
- Admin changes (`grant_admin`/`revoke_admin`) are **authenticated writer-set rotations** — O(1), last admin cannot be removed.
- Enforcement at merge, like the other components: a non-admin's hand-crafted grant delta is rejected because registry entries inherit the admin writer domain. `only_admin`/`only_role` are fail-fast API guards.
- Built on the proven `SharedStorage<UnorderedMap<String, LwwRegister<V>>>` shape — no nested collections. Macro `is_collection_type` + wasm-abi normalizer learn `AccessControl` (the macro doesn't recurse into nested structs).

## Tests

- Full storage lib suite (581) green; wasm-abi green; clippy + fmt clean.
- `components-demo` exercises all three components with `TestHost` tests and builds on wasm with the ABI emitter.
- All `SharedStorage`-using apps build unchanged.

## Deferred (still on #2687)

`Pausable` (advisory) and the `OpMask` protocol extension (§8, gated on #2541).

---

> [!NOTE]
> **Medium Risk**
> Changes core storage types and merge-time authorization primitives; risk is mitigated by the `SharedStorage` alias preserving call sites and new e2e/unit coverage.
> 
> **Overview**
> This PR reshapes the storage access-control stack and adds a third demo component. The writer-set **struct** is renamed to `WriterSetCell`; **`SharedStorage<T>`** is now a type alias for `PermissionedStorage<T, WriterSetAcl>` wrapping that cell, so app call sites keep the one-argument `SharedStorage` API while `PermissionedStorage` / `Ownable` sit on the same hierarchy.
> 
> A new **`AccessControl`** collection provides an admin-tier writer set over a guarded role registry (`grant` / `revoke` / `has_role`, plus admin rotation). The **`components-demo`** app wires it in with unit tests and a **two-node e2e** workflow (convergence, non-admin refusal, merge-time rejection of forged `PermissionedStorage` writes). **CI** builds the demo and runs that workflow.
> 
> **SDK macros** treat `AccessControl` like other guarded wrappers for deterministic IDs and forbid it in `#[app::private]`; **wasm-abi** surfaces `AccessControl` as a guarded map shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a11a72fa91187ea43302337e4c180288d6aec4bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -16,6 +16,7 @@ on:
       - "apps/scaffolding-e2e/**"
       - "apps/xcall-example/**"
       - "apps/kv-store-with-shared-storage/**"
+      - "apps/components-demo/**"
       - ".github/workflows/e2e-rust-apps.yml"
 
   pull_request:
@@ -25,6 +26,7 @@ on:
       - "apps/scaffolding-e2e/**"
       - "apps/xcall-example/**"
       - "apps/kv-store-with-shared-storage/**"
+      - "apps/components-demo/**"
       - ".github/workflows/e2e-rust-apps.yml"
 
   workflow_dispatch:
@@ -112,6 +114,11 @@ jobs:
           cd apps/kv-store-with-shared-storage
           ./build.sh
 
+      - name: Build components-demo app
+        run: |
+          cd apps/components-demo
+          ./build.sh
+
       - name: Upload app artifacts
         uses: actions/upload-artifact@v4
         with:
@@ -126,6 +133,9 @@ jobs:
             apps/kv-store-with-shared-storage/res/*.wasm
             apps/kv-store-with-shared-storage/res/abi.json
             apps/kv-store-with-shared-storage/res/state-schema.json
+            apps/components-demo/res/*.wasm
+            apps/components-demo/res/abi.json
+            apps/components-demo/res/state-schema.json
           retention-days: 1
 
   # ── Test phase: run each workflow in parallel ──
@@ -267,6 +277,9 @@ jobs:
           - workflow: shared-storage
             file: workflows/test_shared_storage.yml
             app: kv-store-with-shared-storage
+          - workflow: components-demo
+            file: workflows/components-demo.yml
+            app: components-demo
           - workflow: kv-store-joiner-cold
             file: workflows/kv-store-joiner-cold.yml
             app: scaffolding-e2e
--- a/apps/components-demo/src/lib.rs
+++ b/apps/components-demo/src/lib.rs
@@ -61,18 +61,22 @@ impl ComponentsDemo {
         Ok(self.roles.has_role(&role, &who)?)
     }
 
-    /// e2e-support method (not for production). Attempt to grant a role,
-    /// returning whether the local admin guard accepted it (`true`) or refused
-    /// it (`false`) instead of trapping — so the adversarial workflow can assert
-    /// a non-admin is refused without failing the RPC call itself. A real app
-    /// should propagate the error (use `grant_role`). The authoritative
-    /// rejection of a *forged grant delta* that bypasses this gate happens at
-    /// merge (same writer-set-guarded mechanism the settings adversarial proves).
+    /// e2e-support method (not for production). Reports whether the caller is
+    /// allowed to grant (i.e. is an admin) as a bool instead of trapping, so the
+    /// adversarial workflow can assert a non-admin is refused without failing the
+    /// RPC call. A real app should just call `grant_role` and propagate the
+    /// error. Authorization is checked explicitly here (rather than swallowing
+    /// the error from `grant`) so a genuine storage error still propagates rather
+    /// than masquerading as a refusal. The authoritative rejection of a *forged
+    /// grant delta* that bypasses this gate happens at merge (the same
+    /// writer-set-guarded mechanism the settings adversarial step proves).
     pub fn try_grant_role(&mut self, role: String, who: PublicKey) -> app::Result<bool> {
-        match self.roles.grant(&role, who) {
-            Ok(()) => Ok(true),
-            Err(_) => Ok(false),
+        let me: PublicKey = env::executor_id().into();
+        if !self.roles.is_admin(&me) {
+            return Ok(false);
         }
+        self.roles.grant(&role, who)?;
+        Ok(true)
     }
 
     /// Owner-only write. The guard fails fast; the boundary is that `config` is
--- a/apps/components-demo/src/lib.rs
+++ b/apps/components-demo/src/lib.rs
@@ -61,6 +61,20 @@ impl ComponentsDemo {
         Ok(self.roles.has_role(&role, &who)?)
     }
 
+    /// e2e-support method (not for production). Attempt to grant a role,
+    /// returning whether the local admin guard accepted it (`true`) or refused
+    /// it (`false`) instead of trapping — so the adversarial workflow can assert
+    /// a non-admin is refused without failing the RPC call itself. A real app
+    /// should propagate the error (use `grant_role`). The authoritative
+    /// rejection of a *forged grant delta* that bypasses this gate happens at
+    /// merge (same writer-set-guarded mechanism the settings adversarial proves).
+    pub fn try_grant_role(&mut self, role: String, who: PublicKey) -> app::Result<bool> {
+        match self.roles.grant(&role, who) {
+            Ok(()) => Ok(true),
+            Err(_) => Ok(false),
+        }
+    }
+
     /// Owner-only write. The guard fails fast; the boundary is that `config` is
     /// a writer-set-guarded entity whose sole writer is the owner.
     pub fn set_config(&mut self, value: String) -> app::Result<()> {
--- a/apps/components-demo/src/lib.rs
+++ b/apps/components-demo/src/lib.rs
@@ -1,18 +1,22 @@
 //! Example app for the access-control components.
 //!
-//! Demonstrates two of the `calimero_storage` components:
+//! Demonstrates three `calimero_storage` components:
 //! - [`Ownable`] — a single-owner config cell with transferable ownership.
 //! - [`PermissionedStorage`] — a group-writable settings map.
+//! - [`AccessControl`] — an admin-managed role registry.
 //!
-//! The guards in the methods (`only_owner`) are fail-fast UX. The real
-//! authorization boundary is at merge: the components store data inside
-//! writer-set-guarded entities, so a peer that hand-crafts a delta around this
-//! WASM still has an unauthorized write rejected when honest nodes apply it.
+//! The guards in the methods (`only_owner`, the admin checks inside
+//! `AccessControl`) are fail-fast UX. The real authorization boundary is at
+//! merge: the components store data inside writer-set-guarded entities, so a
+//! peer that hand-crafts a delta around this WASM still has an unauthorized
+//! write rejected when honest nodes apply it.
 
 use std::collections::BTreeSet;
 
 use calimero_sdk::{app, env, PublicKey};
-use calimero_storage::collections::{LwwRegister, Ownable, PermissionedStorage, UnorderedMap};
+use calimero_storage::collections::{
+    AccessControl, LwwRegister, Ownable, PermissionedStorage, UnorderedMap,
+};
 
 #[app::state]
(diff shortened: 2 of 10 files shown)


chefsale:
Pushed `9488768` addressing the AccessControl review. Good catches — especially the NUL injection.

**Fixed (code):**
- **NUL-byte role-name injection** 🟡 — added `check_role` rejecting any role name containing the separator byte, called in `grant`/`revoke`/`has_role` (`only_role` is covered transitively). A crafted `"editor\0<hex>"` can no longer collide with another `(role, member)` key. Test added.

**Tests added:**
- `only_role` gating (held vs not-held) — the primary app guard, previously untested.
- `grant → revoke → has_role` in one execution reads `false` — confirms the registry reads from storage, not a stale in-memory snapshot.
- NUL-in-role-name rejected.

**Docs (acknowledged-by-design items):**
- `has_role` returns `false` for both *never-granted* and *revoked* (no history).
- `revoke` writes `false` (no tombstone), so the registry doesn't shrink — fine at context scale, called out for unbounded-churn cases.
- Concurrent `grant`/`revoke` resolve last-writer-wins by `LwwRegister` timestamp (no semantic tie-break).
- `revoke_admin`'s empty-set guard is **best-effort** at the API; merge (`rotate_writers` rejecting an empty set) is the backstop for the concurrent race-to-zero.

**Deliberately not changed:**
- **`grant_admin`/`revoke_admin` explicit `only_admin()`** — these delegate to `rotate_writers`, which already applies the single `guard(Op::Admin)`. Adding `only_admin()` would be the redundant double-check that review flagged on #2700. The asymmetry with `grant`/`revoke` is intentional and now documented: the collection-insert path in `grant`/`revoke` is *not* locally gated, so they need the explicit fail-fast; the rotation path is. One gate per method, just located differently.
- **`Ownable::owner()` debug_assert / multi-writer `rotate_writers`** — can't override `rotate_writers` per-policy without a newtype (it's an inherent method on the shared base, `Ownable` is a type alias). The `debug_assert` + blessed `new_owned_by*` / `transfer_ownership` (single-key) constructors remain the guard, as on #2700.

7 AccessControl tests now green; full storage suite green; clippy + fmt clean.

chefsale:
Pushed `f6ef464` — addresses the second review batch **and** adds the missing e2e.

### The big gap: end-to-end merge enforcement

Every reply (mine and the bots') kept saying "merge enforcement needs a 2-node e2e" — but there wasn't one. Now there is: `apps/components-demo/workflows/components-demo.yml`, modelled on the proven `kv-store-with-shared-storage` adversarial workflow, registered in `e2e-rust-apps.yml` (path trigger + build + artifact + matrix entry as `components-demo`). It proves:

1. **Ownable** — node 1 (owner) sets config → converges to node 2.
2. **AccessControl** — node 1 (admin) grants node 2 `editor` → converges; node 2 (editor but **not** admin) is refused when it tries to grant (`try_grant_role` → `false`, the fail-fast guard).
3. **Adversarial (last, the security claim)** — node 2 is **not** a settings writer, so its forged `settings_set` is applied locally but **rejected by node 1 at merge** — node 1 never sees it. `settings_set` has no local guard, so this hits the merge path like a hand-crafted delta. The AccessControl grant registry is the *same* writer-set-guarded map shape, so this proves the same boundary that protects role grants.

### Second review batch (admin idempotence)

`grant_admin`/`revoke_admin` are now **idempotent**: if the admin set wouldn't change (granting an existing admin / revoking a non-admin), they skip the no-op rotation — but still authorize the caller via `only_admin()` on that path, so a non-admin can't probe. Net: exactly **one auth gate per path** (rotation's `guard(Op::Admin)` on the changing path, `only_admin()` on the no-op path), never both — consistent with the one-gate rule from #2700.

The `revoke_admin` empty-set / pre-vs-post-removal interplay (3358356269) is correct as-is: `rotate_writers` authorizes the executor against the current (pre-removal) set, and the empty-set guard runs on the post-removal set — the two checks are about different things (who may act vs. what the result may be). Documented.

7 storage AccessControl tests + 5 app TestHost tests green; app builds on wasm; both YAMLs validated.

chefsale:
Pushed `ccbd4d6` fixing the two reds:

- **`Test (components-demo)`** — real bug in *my workflow*, not the components: the non-admin grant step referenced `{{public_key_new-calimero-node-1}}`, but merobox only exposes the joiner's pubkey (`public_key_new-calimero-node-2`). The unresolved `{{…}}` was passed literally and failed `PublicKey` JSON deserialization (`invalid character '{' at byte 0`). Fixed to use node 2's own key as the grantee — the assertion is that node 2 is a *non-admin*, which holds regardless of grantee.
- **`Test (shared-storage-concurrent-rotation-converge)`** — pre-existing concurrent-rotation flake (a scaffolding-e2e workflow, not touched here). The log shows baseline + all rotations executed fine; only the post-partition `wait_for_sync` timed out (181s) — the timing-sensitive relay-reconverge path. The `SharedStorage`→`WriterSetCell` change is a pure rename + alias (581 storage tests incl. rotation/`writers_at` coverage are green), so the convergence machinery is unchanged. Not a required check; re-running.

chefsale:
Pushed `afe6f03` for the latest round.

**Fixed (code):**
- **`revoke_admin` info leak** 💡 — now calls `only_admin()?` before the last-admin error, so a non-admin gets "not an admin" rather than learning the set is down to its last member.
- **Registry growth / role-name bound** 🟡 — added `MAX_ROLE_NAME_LEN = 128`, enforced in `check_role` (test added). Bounds key size and caps per-entry inflation by a (compromised) admin. The separator check already references `ROLE_MEMBER_SEP`, so it tracks the constant if that's ever changed (addresses the brittleness note too).
- **`try_grant_role` swallowing all errors** 💡/📝 — now checks `is_admin` explicitly and *propagates* a real error from `grant`, so a storage failure no longer masquerades as a "refused" (`false`).

**Investigated and rebutting — concurrent dual-revoke → zero admins** 🟡: this can't happen, and the code shows why. `writers_at` resolves concurrent rotations by **picking one winner by HLC** (`rotation_log_reader::writers_at_concurrent_siblings_resolved_by_hlc`), not by composing both removals. So `revoke_admin(&B)` (→`{A}`) ∥ `revoke_admin(&A)` (→`{B}`) merges to *either* `{A}` or `{B}` — never `{}`. Each rotation's new set is individually non-empty (the local guard), and merge selects one of them; the removals never accumulate. The residual is that one revoke "wins" and the other is silently lost (a liveness/UX wrinkle tracked as #2673), never a safety/zero-admin issue — so no merge-time empty-set guard is required here.

**Not changed (with reason):**
- **`try_grant_role` gated behind `#[cfg(test)]`** — it can't be: the e2e drives the app as a black box over RPC, so a `cfg(test)`/feature-gated method is unreachable to the harness. This is the same reason `kv-store-with-shared-storage` exposes `try_rotate_writers`/`writer_count`. The doc marks it e2e-only; the `is_admin`-first rewrite removes the error-masking footgun.

8 AccessControl tests + 5 app tests green; clippy + fmt clean; app builds on wasm. The `Test (components-demo)` e2e fix from `ccbd4d6` is also in this run.

chefsale:
Pushed `186ceeb` for this round.

**`Rust` check failure** — `client_discovered_under_global_and_per_namespace_keys_additive` in `crates/network/tests/rendezvous_namespace.rs`, a rendezvous/discovery test (the log shows `Reservation(ResourceLimitExceeded)` noise). This PR touches only storage + the demo app + e2e/CI yaml — not `calimero-network` — so it's an unrelated flake; the fresh run re-runs it.

**Doc clarifications applied:**
- `grant` idempotency — documented: idempotent in effect, not in storage (a re-grant refreshes the LWW timestamp).
- `check_role` length bound — noted it's byte length (role names are keys).
- e2e comment — reframed honestly: the settings adversarial is a **direct** merge-enforcement proof for `PermissionedStorage` and a **proxy** proof for `AccessControl` (identical `SharedStorage<UnorderedMap>` guarded-entry shape; the mechanism is also covered directly by storage-layer adversarial unit tests). A forged `grant_role` can't be driven through the workflow precisely because `only_admin()` fail-fasts it before merge.

**Re-raises I've already addressed (no change):**
- *Concurrent dual-revoke → zero admins* — can't happen: `writers_at` resolves concurrent rotations by picking one HLC winner (`writers_at_concurrent_siblings_resolved_by_hlc`), not by composing removals, so the merge is `{A}` or `{B}`, never `{}`. Detailed in my previous reply.
- *`try_grant_role` `#[cfg(test)]`-gating* — not possible; the e2e drives the app black-box over RPC, so a test-gated method is unreachable to the harness (same as `kv-store`'s `try_rotate_writers`). The `is_admin`-first rewrite already removed the error-masking footgun.
- *`is_admin` allocates / `grant` idempotency optimization* — micro-optimizations on an admin-gated, single-threaded path; current behavior is correct and the LWW re-grant semantics are now documented. Not worth added surface.

8 AccessControl tests + 5 app tests green; clippy + fmt clean.

chefsale:
First — `Test (components-demo)` passed ✅ (the e2e merge-enforcement proof is green end-to-end), and the earlier `Rust` red was an unrelated `calimero-network` rendezvous flake that cleared on re-run.

Pushed `a11a72f` for this round.

**Real bug fixed (Cursor Bugbot, Medium) — components missing from the `#[app::private]` denylist.** `AccessControl`, and also `PermissionedStorage` / `Ownable`, could be placed on a `#[app::private]` struct without a compile error (unlike `SharedStorage`). Since they wrap a syncing writer-set-guarded entity, their data would still enter the synced tree — a silent footgun. Added all three to `PRIVATE_INCOMPATIBLE` in `sdk/macros/src/private.rs` with accurate explanations; the denylist test now locks them in (incl. the non-generic `AccessControl`). Good catch — same plumbing class as `is_collection_type` / the wasm-abi normalizer, which I'd updated but had missed this third site.

**Doc nits applied:** role-length error now states the unit (`… (128 bytes)`); `grant_admin` documents last-rotation-wins under concurrent grants (matching `revoke_admin`'s caveat).

**Not changed (re-raises / nits):**
- *`has_role` errors on over-long/NUL names instead of `Ok(false)`* — consistent input validation (both invalid-name cases error the same way); the message is clear. Defensible as-is.
- *Double `writers()` read in `revoke_admin`'s last-admin path* — the `only_admin()` re-read is actually required for correctness: at that point `who` has been removed from the local snapshot, so the snapshot can't tell whether the *executor* is an admin in the self-revoke-last-admin case. Micro-read on an admin-gated single-threaded path; keeping it correct over saving one index read.
- *`try_grant_role` `#[cfg(test)]`* / *concurrent-zero-admins* — addressed in prior replies (black-box-RPC harness; `writers_at` HLC-winner, never composes removals).

Macro denylist tests + 8 AccessControl tests + 5 app tests green; clippy + fmt clean.

PR #2709, opened 2026-06-04
refactor(node/state_delta): extract buffering submodule (increment 3)

chefsale:
## What

Increment 3 of splitting `crates/node/src/handlers/state_delta/mod.rs`. Moves the governance-pending **drain / absorb / recover / fence** cluster (~765 lines) into `buffering.rs` — pure, behavior-preserving code movement (`mod.rs`: **+10** wiring lines / **−766** moved).

Moved:
- `drain_governance_pending`, `drain_all_governance_pending`
- `drain_absorbed`, `drain_absorbed_leaves`, `drain_absorbed_records`, `drain_all_absorbed`
- `recover_absorbed_records`, `recover_absorbed_on_startup`
- `FenceOutcome` + `fence_and_maybe_absorb`

## Wiring

- The external entry points (`drain_all_governance_pending`, `drain_all_absorbed`, `recover_absorbed_on_startup` — called from `sync::manager` and `network_event::namespace`) are re-exported `pub(crate)` so the `crate::handlers::state_delta::*` import paths are unchanged.
- Apply-path helpers (`drain_governance_pending`, `fence_and_maybe_absorb`, `FenceOutcome`) and the test-only helpers (`drain_absorbed_records`, `recover_absorbed_records`, behind `#[cfg(test)]`) are re-imported, so call sites and the in-module test suite are untouched.
- **Explicit imports** in `buffering.rs` (no `use super::*`), per the convention established earlier in this series.

## Validation

- `cargo check -p calimero-node` clean (no new warnings).
- node lib tests: **304 passed**.
- node `sync_sim`: **314 passed**.
- `mod.rs` diff is wiring + one doc-link path qualification only; the 765 lines moved verbatim.

## Progress

`mod.rs`: 3835 → ~3080 LOC. The handler is now organized as `crypto` / `verify` / `events` / `buffering` submodules. Remaining increment: delta-store setup (`init_delta_store` / `choose_owned_identity` / `DeltaStoreSetup`). The two orchestrators (`handle_state_delta`, `apply_authorized_state_delta`) stay in `mod.rs` by design.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Mechanical module split with existing tests passing; no logic or API path changes beyond file organization.
> 
> **Overview**
[...]

PR #2719, opened 2026-06-05
refactor(node/state_delta): extract store_setup submodule (increment 4)

chefsale:
## What

Increment 4 of splitting `crates/node/src/handlers/state_delta/mod.rs`. Moves the delta-store setup cluster into `store_setup.rs` — pure, behavior-preserving code movement (`mod.rs`: **+4** wiring lines / **−128** moved).

Moved:
- `DeltaStoreSetup` (struct; fields `pub(super)` for the apply-path destructure)
- `choose_owned_identity` — also reached by the sibling `buffering` module via `super::` (re-exported through mod.rs's import)
- `init_delta_store`

## Wiring

Explicit imports in `store_setup.rs` (no `use super::*`), per the convention established in this series. `execute_cascaded_events` is reached via `super::` (it lives in the sibling `events` module). Call sites, the sibling `buffering` module, and the in-module test suite are unchanged.

## Validation

- `cargo check -p calimero-node` clean (no new warnings).
- node lib tests: **304 passed**.
- node `sync_sim`: **314 passed**.
- `mod.rs` diff is wiring only; 127 lines moved verbatim.

## Progress

`mod.rs`: 3084 → ~2956 LOC. The handler is now organized as `crypto` / `verify` / `events` / `buffering` / `store_setup` submodules. What deliberately remains in `mod.rs`: the two orchestrators (`handle_state_delta`, `apply_authorized_state_delta`), plus `request_missing_deltas` and `ensure_application_available` (sync/blob concerns).

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 8360d4c1c92b864e6af55db3c54120b381c2e0b5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/handlers/state_delta/mod.rs
+++ b/crates/node/src/handlers/state_delta/mod.rs
@@ -22,6 +22,7 @@ use crate::utils::choose_stream;
 mod buffering;
 mod crypto;
 mod events;
+mod store_setup;
 mod verify;
 
 pub(crate) use buffering::{
@@ -37,6 +38,9 @@ use events::{
     emit_state_mutation_event_parsed, execute_cascaded_events, execute_event_handlers_parsed,
     parse_events_payload, CascadeOutcome,
 };
+// `choose_owned_identity` is also reached by the sibling `buffering` module via
+// `super::` (re-exported through this import).
+use store_setup::{choose_owned_identity, init_delta_store, DeltaStoreSetup};
 pub(crate) use verify::{verify_position_group_id_matches_context, GroupIdCheck};
 
 pub(crate) struct StateDeltaMessage {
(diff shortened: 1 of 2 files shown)


chefsale:
All three findings are pre-existing logic moved verbatim by this PR (it's a pure code-move, +4/−128) — none is introduced by the extraction, and the reviewer notes as much for the identity one. Keeping the move behavior-preserving, so I'm not changing them here; assessment of each:

- 🟡 **Store removed on init failure, cloned ref still usable (L120)** — not a bug today: every caller propagates with `?`, so the partially-initialized store is dropped. The suggestion is defensive future-proofing, not a correctness fix. Out of scope for a move.

- 🟡 **Race between `DeltaStore` creation and `is_new_store` (L55)** — pre-existing, and almost certainly moot in practice: state-delta processing runs on a **per-context arbiter** (issue #2299), so a context's deltas are handled sequentially — two tasks don't concurrently init the same `context_id`'s store. Worth confirming, but it's a pre-existing concurrency question, not a move artifact.

- 🟡 **Non-deterministic identity via `choose_stream` (L44)** — explicitly pre-existing (your note). Only manifests if a node owns *multiple* identities for one context **and** `choose_owned_identity` is invoked more than once per flow with different picks. Real-ish but narrow; belongs in a separate change against the identity-selection logic, not this move.

If any of these is worth a dedicated follow-up (I'd rank the identity-determinism one highest, the race lowest given the arbiter serialization), happy to take it on separately — same as I did for the #2709 absorb-recovery finding in #2717. None blocks this move.

PR #2745, opened 2026-06-05
feat: unified-causal-log P2–P4 — hashed rotation log + parent-edge governance auth (core#2716)

chefsale:
Phases 2–4 of the unified-causal-log direction (`docs/design/unified-causal-log.md`); builds on #2743. Several **breaking flag-day wire/hash changes** (feature branch, no migration). e2e is the gate.

## The arc (what actually shipped)

This branch started as Phase 2 and grew through Phase 4 as each phase superseded the last. The net result deletes far more than it adds.

### Phase 2 — fold ACL into `own_hash` *(interim bridge, now removed)*
Folded the resolved writer/cap map into a `Shared` anchor's `own_hash` so a hash-neutral rotation would move the root. A bridge: it converged the writer set into the root but kept a redundant side store + reconcile. **P3 superseded it and the fold is gone** (`own_hash` is plain `Sha256(data)` for every storage type again).

### Phase 3 — rotation log as a hashed `UnorderedMap` child *(the destination)*
The writer-set rotation log is now an authoritative `UnorderedMap<delta_id, RotationLogEntry>` **child of the Shared anchor**. A rotation is a hashed Merkle child, so divergent writer sets surface as divergent `full_hash` and converge through the normal entity-tree sync path — no fold, no side store, no bespoke reconcile. Fixed the long-standing **concurrent-rotation split-brain** (6 layered root causes: hand-rolled-child non-convergence, collection-vs-side-store resolve mismatch, signing-nonce-before-payload, receiver sig/data decoupling, the #2673 as-of writer resolution, and an unsigned-bootstrap-entry divergence).

**Deleted:** the side store (`Key::RotationLog`), `RotationLogSyncRequest`/`Response` wire variants + the entire reconcile machinery, the Phase-2 fold + `rehash_shared_anchor` + `refold_anchor_for_rotation_child`.

### Phase 4 — `GovernancePosition` → governance parent edge + `acl_view_at`
The cross-DAG reference embedded in state deltas is now a heads-only **`GovernanceParentEdge { governance_dag_heads }`** (was `{ group_id, group_state_hash, governance_dag_heads }`). Receivers authorize via **`acl_view_at(group_from_context, signer, heads)`** — the group is derived from the **context** (canonical context→group mapping), never the signer-supplied `group_id`. That makes the old `group_id`-equality anti-bypass (`GroupIdCheck`) and `group_state_hash` cross-check **structurally unnecessary** (membership at a cut is a deterministic function of the heads' prefix). All six receiver auth sites route through one shared `authorize_delta_at_edge` helper (single source of truth — the per-site copies had drifted into a bypass before).

**Deleted:** `GovernancePosition`, `membership_status_at`, `GroupIdCheck`, `verify_position_group_id_matches_context`.

## Breaking changes (feature-branch flag-day)
- Borsh wire: dropped `RotationLogSyncRequest`/`Response` variants (discriminant shift).
- Signed delta-envelope payload + wire blob: now commit to the slim `GovernanceParentEdge` (heads only), not the 3-field struct.

## Validation
- **Required checks green:** Rust, Lint, Build Binaries.
- **E2E green:** `shared-storage-concurrent-rotation-converge`, `frozen-rga-convergence`, `group-governance-peer-down`, `group-governance-late-joiner`, `auto-follow-asymmetric-context-membership` (concurrent-rotation deterministically green across many reruns).
- **Known pre-existing failure (NOT introduced here):** `group-remove-from-root-revokes-inherited`. Every membership assertion passes (revoked member removed, cross-peer view agrees, inherited writes still flow); only the negative-control read leaks because a removed member's **authorless `Public` write** propagates via auth-blind HashComparison state-merge — the separate HC-merge-of-authorless-plain-entities gap, which needs receiver-side peer-auth in the sync handshake (its own effort). This scenario was already ~50% flaking before this branch.

---

> [!NOTE]
> **High Risk**
> Breaking borsh/serde on governance edges and delta envelopes, centralized authorization and rotation resolution on security-critical receive paths, and flag-day wire changes without migration.
> 
> **Overview**
> This PR completes **core#2716 P2–P4**: shared writer rotations move into a **hashed Merkle collection** on the anchor, and group-context deltas carry a slim **governance parent edge** for apply-time auth.
> 
> **Governance wire & authorization** — `GovernancePosition` becomes **`GovernanceParentEdge`** (only `governance_dag_heads`; `group_id` and `group_state_hash` removed). Outbound edges are built with a single DAG head read. **`acl_view_at(store, group_id, signer, heads)`** replaces `membership_status_at`; the owning group always comes from the **context→group mapping**, not the envelope. **`authorize_delta_at_edge`** unifies gossip receive, governance-pending drain, snapshot replay, and parent-fetch; **`GroupIdCheck`** / **`verify_position_group_id_matches_context`** are removed.
> 
> **Rotation log & sync** — Side-store `Key::RotationLog`, delta-store self-log helpers, and **`RotationLogSyncRequest`/`Response`** wire variants are deleted. The node loads rotation history from the **hashed collection children** and resolves writers via **`writers_at_authenticated`** (signature verify at resolve time). The execute path runs **`self_log_own_rotations`** after building the signed delta and refreshes **`expected_root_hash`** / context root when rotations change.
> 
> **Correctness fixes** — Action signing **stamps `sig_data.nonce` before `payload_for_signing`** (fixes concurrent SharedMember signature mismatches). The context client **EntityIndex** borsh mirror adds **`SharedMember`** and **`BTreeMap` writers + `OpMask`** so root-hash computation decodes Shared entities correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f0691570117061d8f519826fdf75468b7887e73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
## Review response (commit 2f069157)

Triaged every AI-review finding against the current code:

**Fixed**
- 🔴 **Rotation resolution not authenticated** — wired `writers_at_authenticated` (was test-only) into production `resolve_effective_writers_for_delta` for both Shared-anchor and member-anchor resolution. Each rotation-log entry's stored ed25519 signature is now verified at resolve time (`verify_rotation_entry`), and every post-genesis rotation is gated on the prior set's ADMIN holder — a forged/signature-less HC-synced rotation child can no longer shift the resolved writers. *Residual (documented):* the genesis-of-log entry self-authorizes on a valid signature since the pre-genesis set isn't anchored in the log; this is closed properly by the P5 signed-`SetWriters` chain.
- 🟡 **Swallowed errors** in `authorize_delta_at_edge` — the underlying `get_group_for_context` / `acl_view_at` error is now logged before the static reject reason.
- 🟡 **Rotation-log child decode failures** — now `warn!`-logged (partial log could mis-resolve writers) instead of silently skipped.

**Intentional / documented**
- 🔴 **Wire enum tags shifted** (RotationLogSyncRequest/Response removal) — the deliberate flag-day break, no mixed-version coexistence (see PR description §Breaking + design §0.1).
- 🟡 `ensure_rotation_log_parent` zero `merkle_hash` / 0 timestamps — backfilled by `update_hash_for` (the P3 hash-propagation fix); not stale state.

**Stale (code no longer exists)**
- 🔴 "Reconcile collects side store only" — the reconcile machinery + side store were deleted in this PR; the referenced code is gone.

**Deferred (tracked for P5)**
- 🟡 `load_rotation_log_child` drops the `snapshot` field — only matters under rotation-log compaction, which is disabled; revisited with the genesis-floor work in P5.

PR #2775, opened 2026-06-17
feat: unified causal log foundation — one op model, projection, authorization, isolation harness

chefsale:
The foundation for collapsing Calimero's per-context state, access-control, and membership into **one causal op-log** with **one convergence root** and **one authorization check**.

**Fully additive** — new crates only, nothing is wired into the live apply/sync paths yet, so this cannot change runtime behavior. It lands the model, the decisions, and the safety net that the live migration builds on.

## Why

Today a context's state lives in three parallel places — the data DAG, the governance op-logs, and per-object writer-set rotation logs — each with its own apply path, its own convergence signal, and its own authorization check. That split is where whole classes of bugs come from: a writer-set or membership change that doesn't move the data root hash can silently diverge between nodes, and "is this author allowed to write?" is answered by three different folds in three different layers.

This PR introduces a single model where every change — a data write, a writer-set rotation, a membership change, an admin/policy change — is the same `Op`, a scope's state is the deterministic projection of its op-log, its single `scope_root` is the only convergence signal, and authorization is one fold over the op's causal cut.

## What's here

**New crates:**
- **`calimero-op`** — the one `Op` envelope + `OpPayload` across all four planes (data `Put`/`Delete`, access-control `SetWriters`, membership `MemberAdded`/`MemberRemoved`, admin `AdminChanged`/`PolicyUpdated` + the scope tree `SubgroupCreated`/`SubgroupReparented`/`SubgroupDeleted`); `ScopeId`; content-addressed `compute_id`; and `scope_root`, which folds entities **and** authorization (ACL + groups) into one hash — so a hash-neutral writer/membership rotation (same data, different ACL) is a *different* root and can't pass sync as "converged".
- **`calimero-authz`** — the single `authorize(op, AclView)` decision. Causal-honor semantics: an op is authorized against the ACL/membership *as of its own causal parents*, never the receiver's current state (so a write authored before a revocation stays valid regardless of arrival order). `default-write = membership`: a scope member can write ordinary entities; restricted objects require an explicit ACL entry.
- **`calimero-projection`** — `ScopeState`, the deterministic projection (values + ACL + groups + scope tree). Per-slot last-writer-wins keyed on `(hlc, op_id)`, so the fold is order-independent; `acl_view_at(parents)` resolves the authorization view at a causal cut; `root()` / `scope_root_with_entities()` produce the convergence root.
- **`calimero-op-adapter`** (transitional) — encoders mapping each current operation type (`Action`, `RotationLogEntry`, `GroupOp`, `RootOp`) onto `OpPayload`, with in-model vs out-of-model coverage documented per encoder. Deleted once everything runs on the unified op.

**Safety net** — `calimero-projection`'s `testing` feature: a convergence + **scope-isolation** property harness. It proves that members of a scope converge on the same root regardless of op order, and that a non-member never holds or computes a scope's root (existence doesn't leak). It self-checks that it actually detects both a convergence divergence and an isolation leak.

**Equivalence proofs** (the real pre-migration de-risking) — deterministic, no e2e required:
- the projection's **ACL fold** resolves the same writer set as the current `resolve_local`, over the same rotation sequence;
- the projection's **membership fold** matches the current membership resolver (the one driving the live write-authorization check), across 2000 random + all exhaustive short op sequences.

Together these cover both halves of the data-write authorization decision. ~40 unit/property tests across the crates; fmt + clippy clean.

**Design notes** — `docs/design/unified-causal-log-p5-decisions.md` records the ratified design decisions (concurrent-revoke = causal-honor; owner = the `ADMIN` capability holder; key domain ⇔ restricted membership; existence-hiding for restricted subgroups; a coordinated flag-day with no backwards-compat requirement).

## What's NOT here

Wiring `authorize`/`ScopeState` into the live apply path, switching the wire convergence signal to `scope_root`, replacing the three stores with one op-log, deleting the old folds, and unifying the sync protocols are the follow-on work — each its own PR with full e2e, validated against the harness. This PR is just the green, additive foundation they build on.

---

> [!NOTE]
> **Low Risk**
> Additive library and test-only assertions in governance-store; runtime node paths are unchanged until follow-on migration PRs wire the model in.
> 
> **Overview**
> Adds **four new workspace crates** and a design doc; **no live apply/sync wiring** yet.
> 
> **`calimero-op`** — single `Op` / `OpPayload` for data, ACL, membership, and admin/scope-tree changes; `ScopeId`, content-addressed `compute_id`, and `scope_root` that mixes entity, ACL, and governance hashes so ACL-only divergence changes the convergence signal.
> 
> **`calimero-projection`** — `ScopeState` folds an op-log with per-slot LWW on `(hlc, op_id)`; `acl_view_at(parents)` for causal-cut ACL; optional `testing` harness for per-scope convergence and non-member isolation.
> 
> **`calimero-authz`** — one `authorize(op, AclView)` with causal-honor semantics and **default-write = membership** (members get write/delete on unrestricted entities; explicit ACL on restricted objects).
> 
> **`calimero-op-adapter`** — transitional encoders from `Action`, rotation logs, `GroupOp`, and `RootOp` to `OpPayload`, with documented in-model vs out-of-model ops.
> 
> **Tests / de-risking:** fold-equivalence checks that unified ACL projection matches `resolve_local` and membership projection matches the governance-store resolver (random + exhaustive sequences). **`docs/design/unified-causal-log-p5-decisions.md`** records Phase 5/6 decisions (causal-honor, scope keys, flag-day cutover).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c5fe27592571ab7ecf89a1b8e00acc73df26042. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
## Review addressed — `ce52683c`

Thanks for the thorough pass (mostly `meroreviewer`). The findings cluster into ~14 distinct items; here's how each was resolved.

### Fixed (real correctness)
- **`groups_hash` empty-group convergence break** — the genuine bug. `MemberRemoved` left a phantom empty `BTreeMap` when the last member was removed, so "group never created" and "all members removed" hashed differently → divergence between nodes reaching the empty state via different op orders. Fix: `apply` drops the group entry once empty, **and** `governance_hash` skips empty groups (belt-and-suspenders). `member_clock` is retained so a later re-add still has to beat the removal.

### Fixed (hardening / clarity)
- **`compute_id` parent/author boundary** — now length-prefixes the parent list, so no cross-decomposition collision is even theoretically possible.
- **`groups_hash` naming** — renamed `governance_hash` (it covers admin/policy/subgroups too) and documented as the third `scope_root` component.
- **`wins`/`lww_set` duplication** — `lww_set` now delegates to `wins` (single comparison rule).
- **`acl_view_at` silent partial ancestry** — documented the completeness precondition; the live path guarantees it (the DAG buffers an op as `Pending` until all parents are present — `authorize_delta_at_edge`'s `Buffer { needed }` — so `authorize` only runs on a fully-materialized ancestry).
- **shuffle seed (testing.rs)** — mixes 8 bytes of `ScopeId`, not 1.
- **`shadow_data_delta` sentinel group** — **removed `shadow_data_delta`/`ShadowVerdict` entirely.** The single-site live `authorize` shadow it served proved tautological *and* half-blind (the data-write decision is split across the membership gate and the per-object writer gate — no one site sees both). Deterministic fold-equivalence is the real pre-cutover signal: `acl_plane_matches_resolve_local_*` (ACL half) + the membership-fold property test added to `calimero-governance-store` (membership half). Removing it also eliminates the sentinel-group hazard.
- **`calimero-op-adapter` missing from `[workspace.dependencies]`** — added.

### By design (documented, no change)
- **`acl_view_at` includes the parent ops themselves** — correct. `parents` are an op's causal *predecessors*; the cut is inclusive of them (an op is authorized against everything that causally precedes it, including the membership grant that authorizes it). Excluding parents would hide the very grant.
- **`expected_scope_root`/`signature` not in `compute_id` preimage** — intentional and now documented on the fields. `expected_scope_root` is an unsigned *assertion* peers **recompute** (mirrors the existing `CausalDelta::expected_root_hash`); security never depends on it. `signature` signs the id and so can't be folded back into it; callers MUST verify it before trusting an `Op` (projection/authz assume pre-verified ops — documented).
- **`compute_id` `.expect()` on borsh** — `borsh::to_vec(...).expect(...)` is the established convention here (276 occurrences across the repo); to-`Vec` borsh is infallible and a `# Panics` doc is present. Returning `Result` would ripple through every call site for a can't-happen case.
- **`payload_from_group_op` `_ => None` catch-all** — deliberate, so the transitional adapter survives master growing the governance enums; auth-relevant variants are armed explicitly, and the fold-equivalence coverage tests now exist in this PR (increment 2).
- **`SubgroupCreated` same-stamp LWW** — the reviewer's own analysis lands on "intended": a deleted subgroup (`exists=false`) is skipped by `governance_hash`, so stale `parent`/`restricted` on a dead slot can't perturb the root; order-independence is property-tested.

All four scaffold crates green (op 4, projection 6, op-adapter 6, governance-store membership 21), fmt + clippy clean. Resolving the bot threads accordingly — please re-open any you'd like to discuss further.

chefsale:
The remaining review comments are addressed in follow-up #2779 (this PR merged before they landed): `required_mask_for` → `Option<OpMask>` (kills the empty-mask footgun), `DEFAULT_MEMBER_MASK` delete-authority doc, and an explanation that the op-adapter encoders keep a `_ => None` arm because `GroupOp`/`RootOp` are `#[non_exhaustive]` (a compile-time exhaustiveness guard isn't possible from a downstream crate; the membership fold-equivalence test is the safety net). The by-design items (signature verification at the wire boundary, `expected_scope_root` as an unsigned recomputed assertion, `acl_view_at` completeness via DAG buffering, `.expect()` on infallible borsh) are explained inline there too.

PR #2783, opened 2026-06-17
feat(context): unified-op projection fed live from governance + ACL apply streams

chefsale:
Stands up a unified-op `ScopeState` projection in the context crate and feeds it from the **live apply stream** — governance + ACL planes — building on #2775. **Additive**: nothing reads the projection or routes a decision through it yet, so it can't change runtime behavior.

## What's here (`crates/context/src/scope_projection.rs`)
- **`ScopeProjections`** — per-`ScopeId` registry of `ScopeState` projections (the scope is the unit of convergence).
- **Governance plane (live)** — `op_from_signed_namespace_op` converts a cleartext namespace op into the unified `Op` for the scope it affects (joins → group scope, admin/policy → namespace-root scope, structural ops deferred). Author = the signed op's `signer` (deterministic ⇒ deterministic `op_id`/LWW). Wired into `apply_signed_namespace_op`, folded **only on `Applied`**.
- **ACL plane (live, independent)** — after a successful sync-apply, the execute handler reads the **raw** `RotationLogEntry`s the delta recorded for the touched anchors (directly from the datastore, the same Store-backed read the node uses — no WASM env, no runtime change) and folds a `SetWriters` op per rotation via `op_from_rotation_entry`, authored by the rotation's **signer**. Fires only on actual rotations; reads are synchronous before the projection lock; a poisoned lock is ignored, so the shadow can never affect execution.
- **`actions_to_ops`** — data-plane `Action`→`Op` bridge (not fed; entities come from the storage Merkle).

With both auth-relevant planes fed from their **independent** sources, `scope_root` reflects ACL + governance over the storage Merkle root — a hash-neutral writer/membership rotation moves it.

## Notes
- The ACL feed uses raw rotation entries (real signer), the source the projection owns at cutover — not the resolver's merged output. The remaining cutover step is making `SetWriters` a first-class signed op in the unified DAG (vs. read back from the separate rotation-log store).
- Registry is unbounded (governance/ACL feed it; growth tracks live scopes); eviction + persistence come with the cutover.
- `load_rotation_log_direct` is replicated from the node crate; a TODO marks consolidating both onto one Store-backed reader in `calimero-storage`.

7 unit tests; fmt + clippy clean; node + context build. See `docs/design/unified-causal-log-cutover-plan.md`.

---

> [!NOTE]
> **Medium Risk**
> Changes run on governance DAG apply and post-sync-apply success paths (extra store reads and mutex work), but failures are fail-open with warnings only until cutover makes the projection authoritative.
> 
> **Overview**
> Introduces an in-memory **per-`ScopeId` unified-op projection** (`ScopeProjections`) and wires it into existing apply paths as a **shadow substrate**—nothing authorizes or converges on it yet.
> 
> On **namespace governance apply**, cleartext ops are converted to unified `Op`s and ingested only when the DAG reports `Applied`. Membership ops trigger a **log-only** per-member compare against the live `MembershipRepository` when both sides have a direct row (`unified_projection_divergence`).
> 
> On successful **`__calimero_sync_next`** executes, touched Shared anchors and the delta id are captured up front; after apply, **raw** rotation-log entries for that delta are read from the store and folded in as `SetWriters` ops (rotation signer as author), not the resolver’s merged view.
> 
> `ContextManager` holds the registry behind `Arc<Mutex<…>>` so async handlers can ingest without holding `&mut self`. New crate deps support `Op` conversion, projection fold, and ACL views. Unit tests cover scope mapping, rotation conversion, causal `acl_view_at`, and registry isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d437aba79dee8e7b5bc995268218508aec90025. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
## Review addressed — `df63df4b`

Thanks — the threads cluster into a handful of items. Many are also **outdated**: they target the old `node/src/unified_op.rs` (since consolidated into `crates/context/src/scope_projection.rs`) or `shadow_set_writers_op` (since removed — see below).

### Fixed
- **Silently-swallowed store / decode errors** (`load_rotation_log_direct`, incl. Cursor bugbot) — store-read errors and child-decode failures now log a `warn` instead of collapsing to a silent `None`/skip, so a transient I/O fault or corrupt child no longer drops an anchor's rotations from the ACL feed. Matches the node-crate reader.
- **Double-lock / TOCTOU in the governance shadow-compare** — now takes **one** projection lock across ingest + the per-member role read (no window between them).
- **Silently-swallowed `PoisonError`** — both the governance and execute-handler feeds now `warn` on a poisoned lock instead of skipping silently.
- **Dead code** — removed `actions_to_ops`, the batch `ingest`, and `ScopeProjections::scope_root`/`get` (no consumer in this PR). Kept `acl_view_at` + op-log retention (the causal-cut view the flip branch consumes) and `role_of` (the shadow-compare's reader).
- **TODO discoverability** — the `load_rotation_log_direct` consolidation TODO now cites the exact node-crate path.

### Stale (code already changed)
- **`shadow_set_writers_op` zero-key author / `op_id` collision / determinism** — that helper was **removed** when the ACL feed switched to **raw rotation entries**; the feed now uses `op_from_rotation_entry`, authored by the rotation's real **signer**.

### By design / tracked (answered, no change)
- **Zeroed `signature`/`expected_scope_root` on shadow ops** — these projection ops are internal and never forwarded or signature-verified; the real signed-op construction is the cutover's wire concern.
- **`op_id` computed over `delta.parents` vs `parent_op_hashes`** — correct catch, and it's exactly the **F3 op-id-alignment prerequisite** documented on the flip branch (`unified-causal-log-flip-plan.md`): projection op ids/parents must mirror the source DAG so the live decision's *cut* maps. Doesn't affect this PR (the current-state shadow-compare doesn't use the cut).
- **Rotation log sorted by `delta_id`** — the caller filters to this delta's id, so the sort isn't load-bearing here (it mirrors the node reader's shape).
- **Double payload decode on the sync-apply path** — the documented shadow cost; optimized at cutover.
- **DRY: `load_rotation_log_direct` duplicated node↔context** — tracked TODO; consolidates into `calimero-storage` before this becomes authoritative.
- **`acl_view_at`/op-log retention unused in this PR** — additive substrate, consistent with the merged #2775 approach; consumed by the flip branch (the immediate follow-up).

7 projection tests green; fmt + clippy clean. Resolving the threads accordingly — please re-open any you'd like to keep open.

chefsale:
## Review round 2 — `04387f27`

### Fixed
- **Index-decode error silently dropped** — `load_rotation_log_direct` now `warn`s on a rotation-log *index* borsh-decode failure (the `.ok()?` previously swallowed it), completing the error-surfacing alongside the store-read and child-decode paths.
- **Op-log duplicate accretion** — `ingest_op` now skips a duplicate op (same `id`, e.g. a delta replayed on restart). The fold is already idempotent (LWW by `(hlc, op_id)`) and `acl_view_at` dedups by id (BFS `visited`), so duplicates never corrupted the view — but skipping keeps the retained log from growing with dupes.
- **`acl_view_at` perf cliff** — added a `TODO(cutover)` noting the O(n) full-log fold + the unbounded log must become an indexed, bounded/persisted lookup before it's authoritative.

### By design / tracked (no change)
- **Zeroed `signature`/`expected_scope_root`** — `build_op` ids via `Op::compute_id`, whose preimage *deliberately excludes* both fields (see the `calimero-op` docs: `expected_scope_root` is an unsigned assertion peers recompute; the signature signs the id). So zeroing them doesn't affect the id. The real concern — projection ids vs source-DAG ids for the *cut*-based authorize — is the tracked **F3 op-id-alignment** step on the flip branch; it doesn't affect this PR's current-state compare.
- **Unbounded `logs` growth** — acknowledged; governance/ACL ops are low-volume (the namespace-DAG cap is 1024), the dedup above bounds replay growth, and eviction/persistence land at cutover (now flagged on `acl_view_at` itself).
- **DRY consolidation TODO** — tracked in `unified-causal-log-flip-plan.md` / the cutover plan; consolidates into `calimero-storage` before this becomes authoritative.
- **Child-read `None` continue** — a store I/O error already warns (in the `read` closure); an `Ok(None)` for a listed child is the rare write-skew edge, left as a `continue` to avoid double-warning the I/O path.

7 projection tests green; fmt + clippy clean. Resolving accordingly.

chefsale:
## Review round 3 — see latest commit

### Fixed
- **O(n²) dedup** — `ingest_op` now dedups in **O(1)** via a per-scope `seen: HashSet<[u8;32]>` of ids, replacing the O(n) log scan I added last round.
- **Last silent-skip path** — `load_rotation_log_direct` now warns when a child listed in the index is unreadable (store error or absent value), so a partial log is always flagged. All four paths (index read, index decode, child read, child decode) now surface.

### By design / tracked (no change)
- **Unbounded `logs` / cap** — a naive size cap or evict-oldest would break `acl_view_at` for any cut referencing an evicted op (the causal-honor walk needs the ancestry), so bounding can't be a simple cap — it lands with the persistence + indexed-ancestry work at cutover (flagged in the `acl_view_at` TODO).
- **`acl_view_at` O(n) / make `pub(crate)`** — it has no live caller in this PR, so `pub(crate)` would trip the dead-code lint (only the test uses it); keeping it `pub` matches the additive-substrate approach (#2775). The O(n)/unbounded cliff is TODO'd on the method, and the only caller (the flip branch's decision-site) lands behind the e2e gate.
- **`Option` vs `Result` for absent-vs-corrupt** — both cases are warned inside the function; the shadow caller treats both as "skip this anchor's feed for this delta", so a `Result` would ripple to the call site without changing shadow behavior. Revisited when the feed becomes authoritative.

7 tests green; fmt + clippy clean. Resolving.

PR #2790, opened 2026-06-17
feat(context,node): unified causal log cutover — projection shadow at the auth boundary

chefsale:
**Draft — do NOT merge until a divergence-zero e2e run is green.** Opened so CI/e2e runs against the branch; the actual cutover flip (F4) and deletions (F5) are gated.

This is the **cutover** to the unified causal log: making the unified-op projection the authoritative data-write authorization decision and deleting the old per-plane gates. It builds on the merged substrate (#2775, #2779, #2783). Per `docs/design/unified-causal-log-cutover-plan.md`, the security-boundary flip cannot land until the divergence signal below is proven zero on a real merobox/scaffolding run.

## What's here (all additive — nothing decides against the projection yet)
- **F1 — bridge.** One shared `ScopeProjections` `Arc`, fed by `ContextManager` (governance + ACL apply streams) and readable by the node (`ContextManager::with_scope_projections` + `NodeState.scope_projections` + `run.rs` threading).
- **F3 op-id alignment.** Projection ops now mirror the **source DAG**: `build_op` takes the source delta's id/parents (not a fresh `compute_id`), `op_from_signed_namespace_op` threads the governance delta's `id`, `op_from_rotation_entry` uses the rotation's `delta_id`. So a data-write's `governance_dag_heads` cut maps directly onto the projection and `acl_view_at(scope, heads)` resolves the same ancestry the governance DAG would.
- **F2 backfill.** `ScopeProjections::backfill_namespace` rebuilds a namespace's governance scopes by replaying the **persisted** governance DAG (a derived view — no parallel persistence to diverge), once per namespace, idempotent (`ingest_op` dedups by id). So a just-restarted projection isn't empty.
- **F3 decision-site shadow.** `member_at_cut(store, group, author, heads) -> Option<bool>` (backfill + `acl_view_at` + `is_scope_member`). In `handle_state_delta`'s `Authorized` arm — where the live `authorize_delta_at_edge` just authorized the writer — the node asks the projection and, conservatively, logs `unified_projection_divergence` (plane = `membership-cut`) **only** when the projection sees a non-member at a cut the live path authorized (a feed/backfill gap, not the forward-only cases). **Acts on the live decision.**

## The e2e gate (what to run + assert)
On this branch, exercise the governance + rotation scenarios and **assert zero `unified_projection_divergence` markers**. Zero = the projection agrees with the live decision on real traffic across the cluster → safe to do F4.

## What is NOT here (the gated remainder)
- **F4** — make `authorize(op, projection.acl_view_at(parents))` *the* decision (replacing `authorize_delta_at_edge` + `writers_at_authenticated`).
- **F5** — delete the old folds (~3,500 LOC: rotation log, governance DAG folds, `state_hash`, op-adapter), retire on-disk columns. Closes #19 structurally.

Both land only after the marker is proven zero. See the plan doc on this branch.

---

> [!NOTE]
> **High Risk**
> Changes membership authorization at the state-delta boundary and can reject writes live would allow; incorrect projection/backfill or inheritance folding is security-critical until F4b sole authority lands.
> 
> **Overview**
> **Unified causal log cutover (F4a):** after live authorizes a state delta, the node also requires the shared **unified-op projection** to see the writer as a member at the cited `governance_dag_heads` cut; `Some(false)` **rejects** the delta (strict AND with live—cannot grant what live denied). E2E now **hard-fails** on any `unified_projection_divergence` log marker.
> 
> **Projection model fixes:** governance ops fold into one **namespace-scoped** log (matching the single namespace DAG and heads on writes), using source delta **ids/parents** and **`Noop`** nodes so ancestry walks are not truncated; encrypted `GroupOp`s are **decrypt-and-fold** on apply and on **DAG backfill** (`decrypt_group_op`, `collect_namespace_ops` / refresh when heads are missing). **`acl_view_at`** uses **causal generation** for `hlc=0` ops so re-add-after-remove resolves correctly; **`cut_ancestry_complete`** guards authoritative grant paths.
> 
> **Membership in `authz`:** `AclView` gains caps, subgroup tree, and **`is_member_at_cut`** (open-chain inheritance + `CAN_JOIN_OPEN_SUBGROUPS`), with new op authorization arms for visibility/caps/`Noop`.
> 
> **Plumbing:** one `Arc<RwLock<ScopeProjections>>` shared by context (feed) and node (read); `MembershipReject` is split from structural `Reject` for future sole-authority flip. Grant direction (projection overriding live reject) stays deferred—logged as `unified_projection_overauth` only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3176bec91c5b9a367dbfcae10c1e46da113976b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Addressed the review in `62d2f1cd`. Summary of each thread:

**🟡 `member_at_cut` carve-out masks real feed gaps** (×2) — Correct and the most important point. The carve-out OR-ed in `MembershipRepository::is_member`, which returns true for members established by governance ops (they carry direct rows), so a dropped `MemberJoined`/`MemberAdded` was masked and the gate could pass vacuously. Narrowed to exactly the two pieces of base state that are *not* governance ops in this namespace's DAG and so legitimately cannot be folded: genesis admin (`GroupMeta::admin_identity`) and inherited membership (`check_path => Inherited`). Direct/op-established membership now surfaces as a divergence — which is the point of the gate.

**🟡 `backfilled` marked even when head read fails** (×2) — Fixed. `collect_namespace_ops` returns `None` when the governance head is unreadable, and `apply_backfill`/`backfill_namespace` only mark the namespace done on a successful read, so a transient store fault retries instead of permanently skipping.

**🟡 Mutex held across `member_at_cut`/backfill BFS** (×2) — Fixed. Split into a lock-free `collect_namespace_ops` (the RocksDB DAG walk) + a brief `apply_backfill` (ingest). The decision site resolves+gates under a brief lock, walks lock-free, then re-locks only for the cheap ingest and the read — the apply path no longer blocks on the walk.

**🟡 Mutex poison silently suppresses divergence** (×2) — Fixed. The site used `lock().ok()`, returning `None` on poison and silently disabling the shadow. Now goes through `NodeState::lock_scope_projections`, which recovers the guard via `PoisonError::into_inner` (matching `lock_peer_identity_cache`) so a poisoned writer can't blind the gate.

**🟡 BFS ingests in non-topological order** (×1) — Not a bug. `ScopeState::apply` is per-slot LWW (`wins(stamp, clock)` on every arm), so the folded state converges regardless of ingestion order; a remove visited before its causal add can't be beaten by the lower-HLC add. No sort needed. Documented on `apply_backfill`.

Docs-bot (AGENTS/CONTRIBUTING/docs): no action — this PR is the additive shadow substrate, still behind the gate, no architecture surface change for contributors yet.

chefsale:
**What the red CI told us (and what I changed)**

The stricter carve-out from the last round did its job: it stopped masking and surfaced the real remaining gap. In every failing job the **functional workflow step passed** — only the divergence assertion failed, and the diverging authors are regular members, not the genesis admin.

Root cause: the diverging members were added via the **encrypted `GroupOp` plane** (e.g. `create_mesh` / admin-push `MemberAdded`), not a cleartext Root op. Those ops ride the namespace DAG encrypted, so:
- `op_from_signed_namespace_op` returns `None` for `NamespaceOp::Group`, and
- the Store-backed backfill can't decrypt them either.

So the projection's op-fold legitimately can't see admin-added members yet. The previous all-green run was entirely the old broad `is_member` fallback masking this — exactly the vacuous-pass the review flagged.

Folding the encrypted membership plane is a real workstream, not a quick fix (live decrypt feed off the `SignedGroupOp` apply path + a backfill-from-materialized path for cold nodes). I've written it up as **F3.5** in `unified-causal-log-flip-plan.md`.

Until that lands, failing the job on a known-unbuilt plane would keep this do-not-merge branch permanently red while blocking nothing. So the divergence step is now **informational** — it reports the count + planes + a sample as a CI warning but doesn't fail. It flips back to a hard `exit 1` gate as the **last step of F4**, when zero divergence is actually achievable. The marker keeps measuring readiness in the meantime.

chefsale:
Triaged the review (`3176bec9`). Summary by finding:

**Fixed**
- 🟡 **Unbounded backfill BFS** (flagged ×3) — added `MAX_BACKFILL_OPS` (100k); a corrupted/deep DAG now returns a partial walk with a warning instead of OOM-ing. This also bounds the originator re-walk cost.
- 💡 **Stale intra-doc link** `namespace_needing_backfill` → `namespace_to_refresh`.
- 🟡 **Materialized-fallback comment misleading** — clarified the wholly-unfolded gate prevents masking only for a zero-member group; a partially-folded group correctly surfaces divergence for its un-folded members.

**Consciously deferred (architecture, not this PR)**
- 🟡 **Re-walk frequency / TOCTOU / thundering-herd on `namespace_to_refresh`** — these are the refresh path's perf characteristics. The bot itself notes they're "acceptable for the shadow/pre-cutover phase," and the BFS cap now bounds the worst case. The principled fix (a post-publish live feed instead of re-walk) is part of the deferred sole-authority work.
- 🟡 **`apply_backfill` marks backfilled on empty/`None` re-walk** — a refresh edge; harmless for the deny-direction co-authorizer (errs toward member). Folded into the deferred grant-direction work where backfill becomes authoritative.
- 💡 **`cut_diagnostics` double-walks `acl_view_at`** — only fires on a divergence, which is zero on the gated (deny) path, so the doubled work is effectively never hit. Will fold the view-reuse in when the diagnostics path becomes hot.

**Context for the cluster of findings:** most concern the **grant-direction / refresh** machinery. After extensive validation, the sole-authority **grant** direction is **deferred** — faithfully reproducing live's open-subgroup *inheritance* model (which depends on non-op base state: default caps, the `MemberJoinedOpen` inheritance-proof semantics, recursive subgroup state) repeatedly traded an over-grant for an under-grant. This PR lands the **safe deny-direction co-authorizer** (projection adds rejections only, strict AND, can never over-authorize), with the grant path informational and non-gated. The hard gate is the forward `membership-cut` (deny) divergence = zero.

PR #2795, opened 2026-06-18
feat(sdk): compile-time misuse diagnostics (incl. guarded storage) + revive macro lint suite as a CI gate

chefsale:
## Why

The SDK's proc-macros already reject some misuse with clear `(calimero)>` errors, and a trybuild golden-file suite verifies those messages — but the suite was `#[ignore]`d (so it never ran in CI and had rotted to 16/22 failing), and many common misuses still surfaced as cryptic downstream trait-bound errors far from the cause.

This PR (1) makes the suite a live CI gate again and (2) turns a broad set of SDK misuses — including the permissioned/guarded storage surface — into span-precise, self-explaining compile errors, so the compiler teaches correct usage instead of leaking internals.

## 1. Revive the misuse-lint suite as a CI gate

- Un-ignore `crates/sdk/tests/macros.rs::all` so it runs under plain `cargo test` (CI already runs that → no workflow change). Toolchain is pinned to `1.88.0`, so goldens are stable; a re-bless recipe is documented on the test and in `crates/sdk/AGENTS.md`.
- Re-bless the drifted goldens (only rustc cascade/wording noise changed; every `(calimero)>` assertion preserved).
- Add `calimero-storage` as a **dev-dependency** so full `#[app::state]` expansion compiles in fixtures (a dev-dep cycle, which Cargo permits). This also removed the spurious `could not find calimero_storage` error from the compile-fail goldens.
- Modernize the stale pass-fixtures to the current contract while keeping every case: mandatory `#[app::init]`, no state-lifetime params, bare state fields converted to their CRDT form (`LwwRegister<T>`, which derefs to `T` so accessors are unchanged).

The suite now covers **32 cases** (compile-fail diagnostics + valid-usage pass fixtures).

## 2. New compile-time misuse diagnostics

Each is locked in by a `compile_fail` fixture.

**Use-site (`#[diagnostic::on_unimplemented]`, stable):**
- `Mergeable` — a non-CRDT field/collection-value now says *"wrap in `LwwRegister<T>` / use a CRDT collection / `#[derive(Mergeable)]`"* instead of a bare "trait not satisfied".
- `AppEvent`/`AppEventExt` — emitting a non-`#[app::event]` type, **and** `#[app::state(emits = T)]` where `T` isn't an event (via the `AppState::Event: AppEvent` bound), both explain it must be an `#[app::event]` enum.
- `Authorizer` — the policy parameter of `PermissionedStorage<T, A>` (i.e. `SharedStorage`/`Ownable`). A non-`Authorizer` policy now names the built-ins (`WriterSetAcl`/`OwnerAcl`/`ProtocolAuthorizer`).
- `#[diagnostic::do_not_recommend]` on **all 18 internal `Mergeable` impls** (collections + guarded wrappers): kills rustc's irrelevant "the following other types implement `Mergeable`" suggestion list, and makes the `error_non_mergeable_field` golden drift-proof against future `Mergeable` impls on master.

**Proc-macro checks (point at the offending token):**
- `#[app::state]` on an **enum** is rejected. No `Mergeable` impl is generated for an enum, so the unconditional wasm merge-registration hook (`register_crdt_merge::<Self>()`, which requires `Mergeable`) **fails to build for `wasm32`** — the real app target — even though it compiles on the host. Steers to a struct / `LwwRegister<MyEnum>`.
- `#[app::view]` + `&mut self` is rejected — a view runs under a shared read lock, so a mutable receiver is unsound. (Master added the view/xcall/init *conflict* checks; this adds the receiver check.)
- Duplicate `#[app::init]` is flagged at the second initializer.
- Public method names in the reserved `__calimero` namespace are rejected before they collide with generated exports.
- `#[app::event]` on a struct gives a clear message instead of leaking serde-derive's `#[serde(content)] can only be used on enums`.

**Type-ban lint:**
- Interior-mutability / shared-ownership wrappers (`Cell`/`RefCell`/`Mutex`/`RwLock`/`Rc`/`Arc`) are rejected in state — no merge semantics, don't round-trip through borsh.

**Footgun-as-warning:**
- `#[must_use]` on the read-only `ValueRef` from `get()`: discarding it is a no-op read; the message points to `get_mut`/`entry().or_default()`.

### Deliberately omitted (unsound to hard-error / runtime concerns)
- A non-determinism body-scan lint (`SystemTime::now`/`rand` in logic). App-logic mutations are broadcast as **state deltas**, so a `now()` timestamp or generated id computed once on the originating node replicates verbatim — a syntactic scan can't distinguish that legitimate use (e.g. chat timestamps) from convergence-breaking recompute.
- Guarded-storage usage invariants: `SharedStorage<Collection>` is *valid* via `get_mut`; an unguarded plain field next to a guarded one; `AccessControl` NUL/length role-name limits; `Authorizer` purity. These are dataflow/runtime properties a syntactic macro can't soundly decide.

## Verification

- `cargo test -p calimero-sdk --test macros` → `all ... ok` (32 cases)
- `cargo test -p calimero-sdk-macros` → 80 passed
- `cargo test -p calimero-storage --lib` → green
- `cargo build --workspace --all-targets` → clean (no new `must_use` warnings, no enum-state breakage)
- `cargo fmt --check` + `cargo clippy` → clean

---

> [!NOTE]
> **Low Risk**
> Changes are compile-time diagnostics, test harness, and docs only—no runtime node behavior. Dev-dep cycle on calimero-storage is limited to tests.
> 
> **Overview**
> Turns common Calimero SDK mistakes into **span-precise `(calimero)>` compile errors** instead of obscure trait-bound failures, and makes the **trybuild macro suite a real CI gate** (un-`#[ignore]`, `calimero-storage` dev-dep so full `#[app::state]` expansion compiles, pass fixtures updated for mandatory `#[app::init]` and CRDT-shaped state).
> 
> **Proc-macro rejections** now cover enum `#[app::state]`, struct `#[app::event]`, duplicate `#[app::init]`, `__calimero*`-prefixed public methods, `#[app::view]` with `&mut self`, and interior-mutability / shared-ownership wrappers (`Rc`, `RefCell`, etc.) anywhere in state field types.
> 
> **Trait-level messages** via `#[diagnostic::on_unimplemented]` on `Mergeable`, `AppEvent` / `AppEventExt`, and `Authorizer`, plus `#[diagnostic::do_not_recommend]` on internal `Mergeable` impls to drop rustc’s long “other types implement Mergeable” list. **`ValueRef`** from collection `get()` gets `#[must_use]` so discarding a read-only copy is flagged.
> 
> New **compile-fail goldens** lock these in; `AGENTS.md` documents the suite and re-bless workflow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8962aeed119bab5eab787ddf9b1f3154cb0d5984. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1649,6 +1649,7 @@ dependencies = [
  "calimero-prelude",
  "calimero-primitives",
  "calimero-sdk-macros",
+ "calimero-storage",
  "calimero-sys",
  "calimero-wasm-abi",
  "serde",
--- a/crates/sdk/AGENTS.md
+++ b/crates/sdk/AGENTS.md
@@ -508,3 +508,25 @@ cargo build -p kv-store --target wasm32-unknown-unknown --release
 - Methods without `&mut self` are read-only (queries)
 - Methods with `&mut self` can modify state (mutations)
 - WASM apps must target `wasm32-unknown-unknown`
+
+## Misuse Diagnostics
+
+The macros reject common misuse with span-precise `(calimero)>` compile errors
+(don't rely on the cryptic downstream trait-bound error). Rejected at compile
+time: non-CRDT state fields (bare `String`/`Vec`/primitives, std collections,
+`Rc`/`Arc`/`Cell`/`RefCell`/`Mutex`/`RwLock`), `#[app::state]` on an enum,
+state lifetime params, duplicate `#[app::init]`, `__calimero`-prefixed method
+names, `#[app::event]` on a struct, and emitting a non-`#[app::event]` type.
+A discarded `get()` result (`ValueRef`) warns via `#[must_use]`.
+
+These messages are regression-tested by the trybuild suite in
+`tests/macros/` (`tests/macros.rs::all`, run under plain `cargo test`). The
+golden `.stderr` files are toolchain-sensitive; after an intentional
+`rust-toolchain.toml` bump, re-bless with:
+
+```bash
+TRYBUILD=overwrite cargo test -p calimero-sdk --test macros
+```
+
+then verify the diff only shows rustc wording/cascade noise — never a dropped
+`(calimero)>` line.
--- a/crates/sdk/AGENTS.md
+++ b/crates/sdk/AGENTS.md
@@ -514,7 +514,9 @@ cargo build -p kv-store --target wasm32-unknown-unknown --release
 The macros reject common misuse with span-precise `(calimero)>` compile errors
 (don't rely on the cryptic downstream trait-bound error). Rejected at compile
 time: non-CRDT state fields (bare `String`/`Vec`/primitives, std collections,
-`Rc`/`Arc`/`Cell`/`RefCell`/`Mutex`/`RwLock`), `#[app::state]` on an enum,
+and the interior-mutability / shared-ownership wrappers `Rc`/`Arc`/`Weak`/
+`Cell`/`RefCell`/`UnsafeCell`/`OnceCell`/`OnceLock`/`Mutex`/`RwLock`),
+`#[app::state]` on an enum,
 state lifetime params, duplicate `#[app::init]`, `__calimero`-prefixed method
 names, `#[app::event]` on a struct, and emitting a non-`#[app::event]` type.
 A discarded `get()` result (`ValueRef`) warns via `#[must_use]`.
--- a/crates/sdk/Cargo.toml
+++ b/crates/sdk/Cargo.toml
@@ -31,3 +31,8 @@ tracing = ["dep:tracing", "dep:tracing-subscriber"]
 
 [dev-dependencies]
 trybuild.workspace = true
+# The macro-test fixtures exercise full `#[app::state]` expansion, which emits
+# `::calimero_storage` paths (CRDT merge registry). This is a dev-dependency
+# cycle (calimero-storage depends on calimero-sdk) — Cargo permits it because
+# the back-edge is a dev-dependency.
+calimero-storage.workspace = true
--- a/crates/sdk/macros/src/errors.rs
+++ b/crates/sdk/macros/src/errors.rs
@@ -74,6 +74,14 @@ pub enum ParseError<'a> {
     MustSpecifyLifetime,
     #[error("this event must be public")]
     NoPrivateEvent,
+    #[error(
+        "`#[app::event]` can only be applied to an enum — events are serialized as a tagged \
+         union (`{{ \"kind\": ..., \"data\": ... }}`), which has no meaning for a struct.\n\n\
+         Model each event as an enum variant instead:\n\n\
+         #[app::event]\n\
+         pub enum Event {{ /* one variant per event */ }}"
+    )]
+    EventMustBeEnum,
     #[error("please use a simple `pub` directive")]
     NoComplexVisibility,
     #[error("explicit ABIs are not supported")]
--- a/crates/sdk/macros/src/errors.rs
+++ b/crates/sdk/macros/src/errors.rs
@@ -161,6 +161,12 @@ pub enum ParseError<'a> {
     XCallAndInitConflict,
     #[error("`#[app::xcall]` and `#[app::view]` are mutually exclusive — xcall is fire-and-forget, so a read-only target's return value would be discarded")]
     XCallAndViewConflict,
+    #[error(
+        "a `#[app::view]` method is read-only but takes `&mut self` — the node runs it under a \
+         shared read lock, so mutating state through it is unsound. Use `&self`, or drop \
+         `#[app::view]` if the method must mutate."
+    )]
+    ViewCannotMutate,
 }
 
 impl AsRef<Self> for ParseError<'_> {
--- a/crates/sdk/macros/src/errors.rs
+++ b/crates/sdk/macros/src/errors.rs
@@ -122,6 +122,31 @@ pub enum ParseError<'a> {
         suggestion: &'static str,
     },
     #[error(
+        "`{type_name}` is not allowed in replicated state — interior-mutability and \
+         shared-ownership wrappers have no merge semantics and would silently diverge across \
+         replicas (and don't round-trip through borsh).\n\n\
+         Store the plain value instead — wrap it in `LwwRegister<T>` for last-write-wins, or use \
+         an SDK CRDT collection. If you truly need this for node-local scratch, keep it out of \
+         `#[app::state]`."
+    )]
+    ForbiddenInteriorMutability { type_name: &'static str },
+    #[error(
+        "`#[app::state]` cannot be applied to an enum — no `Mergeable` impl is generated for it, \
+         so the type fails to build for `wasm32` (the runtime's merge registration requires \
+         `Mergeable`), and even if it built, there is no canonical way to merge values from \
+         different variants.\n\n\
+         Use a struct of CRDT fields. For a value that really is one-of-N, make it a struct field \
+         wrapped in `LwwRegister<MyEnum>` (last-write-wins)."
+    )]
+    StateMustBeStruct,
+    #[error("duplicate `#[app::init]` — a type may declare at most one initializer")]
+    DuplicateInit,
+    #[error(
+        "this method name is reserved by the SDK's code generation — pick another. Names starting \
+         with `__calimero` (and the generated runtime exports) collide with emitted symbols."
+    )]
+    ReservedMethodName,
+    #[error(
         "`{type_name}` cannot be used inside `#[app::private]` — it {explanation}\n\n\
          A node-local private namespace has a single writer, so the multi-writer / \
          sync-layer semantics this type models have no meaning here."
--- a/crates/sdk/macros/src/event.rs
+++ b/crates/sdk/macros/src/event.rs
@@ -75,7 +75,16 @@ impl<'a> TryFrom<EventImplInput<'a>> for EventImpl<'a> {
         let errors = Errors::new(input.item);
 
         let (vis, ident, generics) = match input.item {
-            StructOrEnumItem::Struct(item) => (&item.vis, &item.ident, &item.generics),
+            StructOrEnumItem::Struct(item) => {
+                // A struct can't carry the `{ kind, data }` tagged-union shape an
+                // event serializes to. Reject it here with a clear SDK message,
+                // otherwise the `#[serde(content = "...")]` we generate below
+                // surfaces serde-derive's confusing "can only be used on enums".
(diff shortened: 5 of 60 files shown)


chefsale:
Rebased onto latest `master` and addressed the review feedback:

- **`elided_lifetime` comment (×5, complex_generics_valid.rs)** — the comment was misleading. Rewrote it to state accurately that, with `&self` plus another reference param, the elided output lifetime binds to `&self` (the receiver), *not* to the param — so the body can't return the param and returns `""` (`'static`, which coerces). Renamed the unused param to `_data`. Note: the suggested "restore `data`" wouldn't compile in a `&self` method (output elides to `&self`); the self-tied elision that borrows a *real field* is covered by `LifetimeReturnState::borrow_data`.
- **event.rs early return** — kept the early return (intentional) and documented why: the downstream checks (visibility/generics/reserved idents) all assume an enum, so re-running them on a struct only adds noise; converting struct→enum is the single fix and re-surfaces anything else next compile.
- **Doc bot (AGENTS.md)** — added a "Misuse Diagnostics" section to `crates/sdk/AGENTS.md` documenting the new diagnostics, the trybuild gate, and the `TRYBUILD=overwrite` re-bless recipe.
- **E2E failure** — not caused by this PR. The failing apps (`scaffolding-e2e`, `xcall-example`) use `#[app::event]` on enums (not structs), no enum-state, no `Rc`/`RefCell`/`__calimero` names, and the host `cargo build --workspace` passes (macro lints fire at expansion regardless of target). Re-running after the rebase.

Re-blessed `error_non_mergeable_field.stderr` after the rebase (master added `AccessControl`/`PermissionedStorage` `Mergeable` impls, which changed rustc's "other types implement" suggestion list — the `(calimero)>` message itself is unchanged).

chefsale:
## Follow-up issues

Improvements intentionally left out of this PR (either unsound for a syntactic proc-macro, or adjacent scope), filed for follow-up:

- #2800 — **Dataflow lint pass (dylint)** for the footguns a macro can't soundly catch: capability-check-then-mutate-unguarded-field, `SharedStorage<Collection>` mutated via `get()`, and non-determinism in the *merge* path. _(highest leverage)_
- #2801 — **`on_unimplemented` for collection keys (`StorageKey`) and JSON-RPC boundary types (`AppArg`/`AppReturn`)** — finishes the use-site error story with the same mechanism used here.
- #2802 — **Macro expansion snapshot tests (`macrotest`)** — guards the *generated code* (this PR guards the error messages + valid-compile).
- #2803 — **Warn on `version = N>0` with no migration path.**
- #2804 — **`#[must_use]` sweep + `#[doc(hidden)]` on internal API leakage.**
- #2805 — **Richer host-boundary panic messages** (method name + expected arg schema).

chefsale:
Addressed the latest review:

- **`error_view_mutates` cascade (`could not find 'view' in 'app'`)** — this surfaced a real latent gap: `#[app::view]` was recognized by `#[app::logic]` but never exported as a marker attribute (unlike `#[app::init]`/`#[app::xcall]`/`#[app::destroy]`), so on any error path — where the macro re-emits the original tokens — it failed to resolve. Fixed by adding a no-op `#[proc_macro_attribute] pub fn view` and re-exporting it from `calimero_sdk::app`. The golden no longer carries the cascade; only the `(calimero)>` `ViewCannotMutate` message remains.
- **Reserved method names** — extracted `RESERVED_METHOD_PREFIXES` const.
- **Interior-mutability docs** — `crates/sdk/AGENTS.md` now lists every banned wrapper (added `Weak`/`UnsafeCell`/`OnceCell`/`OnceLock`). The error message already names the specific offending type.
- **`do_not_recommend`** — added a comment explaining it's intentionally applied to *every* `Mergeable` impl (full suppression of rustc's suggestion list; the trait's `on_unimplemented` note is the authoritative guidance) and that it's diagnostics-only.
- **Duplicate `#[app::init]` (×3 comments)** — documented in `logic.rs` that the check accumulates with per-method errors (no `check()` runs before it) and that `methods` holds only successfully-parsed methods (a malformed init surfaces its own error). Also documented in the fixture why two initializers can't be isolated from a companion error: both must be named `init` (rustc duplicate-definition), and a differently-named second init would instead trip "must be named `init`".

All green: trybuild 32 cases, macros 80 tests, workspace build + fmt + clippy clean.

chefsale:
Addressed the latest review pass:

**Changed**
- **Duplicate-init golden drift-proofing (×4 comments):** renamed the second initializer to `init_with_value`. Naming both `init` pulled rustc's E0592 into the golden (coupling it to rustc wording). Now both diagnostics are SDK-controlled `(calimero)>` messages — `DuplicateInit` + `AppInitMethodNotNamedInit` — so the golden no longer asserts on any rustc text. (Good catch; my earlier "can't isolate" note missed that the rename trades the rustc error for an SDK one.)
- **`event.rs` early return (×3):** clarified in the comment that this is the first match arm, so `errors` is empty here and nothing is dropped. The `errors.finish(...)` form is kept deliberately — it's the same pattern the sibling visibility checks (`NoPrivateEvent`/`NoComplexVisibility`) use in this function, so switching to a bare `SynError` return would make it the odd one out.
- **`error_value_ref_must_use` (×2):** documented that the fixture is compile-only (never run → the `unwrap`s can't panic) and why `#![deny(unused_must_use)]` is required (the lint is warn-by-default; a `compile_fail` test needs it promoted to an error to assert on).

**Reviewed, no change (with reasons)**
- **"Reserved-name check fires on private methods":** it doesn't — the check (`method.rs:498`) runs *after* the private early-return (`return Ok(Self::Private)` at `:441`), so private methods never reach it.
- **"Duplicate-init runs after `errors.check()?` aborts":** there is no `check()` before that loop in `logic.rs` — per-method failures are accumulated via `errors.combine(...)`, not `check()`, so the duplicate-init diagnostic accumulates alongside them. Already documented inline.
- **`on_unimplemented` referencing `LwwRegister<{Self}>`:** `{Self}` is the non-`Mergeable` type, and wrapping it as `LwwRegister<{Self}>` is exactly the correct fix, so the substitution is right.
- **`Weak`/`leak_str`:** the wrapper ban and the leak-on-clone follow the existing `forbidden_std_collection` pattern (same user-type-name caveat, same short-lived proc-macro leak); kept for consistency.

CI: required checks (Rust/Lint/Build) are green; the e2e/sync failures were "Setup merobox" timeouts (exit 124) and pass on re-run.

chefsale:
Cursor Bugbot caught a **real regression** — thank you. Fixed in 7573f470.

**The bug:** after I renamed the second initializer to `init_with_value` (to drop rustc's E0592 from the golden), it failed its *own* `AppInitMethodNotNamedInit` validation and was excluded from the parsed `methods` list — so the duplicate-init check, which iterated `methods`, saw only one init and **`DuplicateInit` silently stopped firing**. The previous golden confirmed it (only `AppInitMethodNotNamedInit` + the `AppStateInit` cascade, no `DuplicateInit`).

**The fix:** detect duplicates by counting `#[app::init]` attributes directly on the impl items (new `has_init_attr` helper), independent of whether each method parses. `DuplicateInit` now fires regardless of the second initializer's name/validity. Removed the now-unused `is_init`/`name` accessors. The golden now genuinely contains the `(calimero)>` `DuplicateInit` message — verified — and all three diagnostics (`DuplicateInit`, `AppInitMethodNotNamedInit`, and the `AppStateInit` `on_unimplemented` note) are SDK-controlled, so it stays drift-proof.

On the two "**golden shows `AppStateInit` cascade**" comments (duplicate-init + reserved-name): that cascade is expected and consistent — when `#[app::logic]` reports any error it re-emits the original tokens without the generated `AppStateInit` impl, so the type's `AppState` bound is unsatisfied. It appears in every error fixture (`error_init_not_named`, `error_view_mutates`, …), and the message is our own `on_unimplemented` text (not rustc's), so it's stable. Not spurious — it's how the macro has always surfaced errors.

The repeated `event.rs` early-return / `Weak` / `leak_str` / `on_unimplemented {Self}` / reserved-on-private / duplicate-after-check comments are re-posts from prior commits; addressed (fixed or reasoned) in the earlier replies above.

chefsale:
Rebased onto latest `master` (includes #2799).

- **#2799 added the same `#[app::view]` marker fix** I had in this branch — resolved the conflict in favour of master's canonical version (and adjusted its doc, since this PR additionally rejects a `&mut self` view receiver at *compile* time, which #2799's doc said only happens at runtime).
- **Caught a latent issue from #2799:** it added a `ViewState` case to `attribute_interactions.rs`, but the trybuild suite is still `#[ignore]`d on master, so that fixture was never actually compiled there — it used a bare `String` state field (rejected by the pre-existing forbidden-types lint) and had no `#[app::init]`. This PR un-ignores the suite, so it now compiles; converted the field to `LwwRegister<String>` + added an initializer, keeping #2799's `#[app::view]` + `#[must_use]` intent. (Nice illustration of exactly why this PR matters: the guard now catches fixtures that silently never compiled.)

All green post-rebase: trybuild 32 cases, macros 80, workspace build + fmt + clippy clean.

chefsale:
Latest review pass (commit 8962aeed):

- **"Duplicate init spans wrong method"** (Cursor) — valid. With a misnamed `#[app::init]` *before* the real `pub fn init`, `skip(1)` marked the valid `init` and left the earlier offender unmarked. Fixed: when more than one `#[app::init]` is present, *all* of them are flagged, so no arbitrary "valid" initializer is singled out. The golden now shows the `(calimero)>` `DuplicateInit` message on both.
- **"Duplicate init lint never fires"** (Cursor, commit `17d1fcab`) — **stale**: that was the `methods`-based check, already replaced in `7573f470` by an impl-level count (`has_init_attr` over the raw impl items), so it fires regardless of whether a misnamed init parses. The current code is at `logic.rs:144`.
- **"Wrapper lint matches type names"** (Cursor) — intentional and consistent with the pre-existing `forbidden_std_collection` lint, which has matched `HashMap`/`Vec`/`String` by final-segment name since before this PR. By-name matching is deliberate: it also catches aliased/re-exported imports (`use std::cell::RefCell as R;`), which a `std::`-path requirement would miss. The (rare) cost is a user type named exactly `Cell`/`Rc`/etc. — the same accepted trade-off the collection lint already makes.

**CI:** the required checks (Rust / Lint / Build Binaries / Build Apps / Build WASM fixtures) are green. The failing jobs are all non-required e2e/migration/sync scenarios that died at the **"Setup merobox" step with exit code 124** — apt/runner setup timeouts, before any test ran (not code). Re-running.

Issue #2801, opened 2026-06-18
SDK diagnostics: on_unimplemented for collection keys (StorageKey) and JSON-RPC boundary types (AppArg/AppReturn)


chefsale:
## Context

Follow-up to #2795. That PR added `#[diagnostic::on_unimplemented]` to `Mergeable`, `AppEvent`/`AppEventExt`, and `Authorizer`. Two more use-site boundaries still produce cryptic errors; both are fixable with the **same proven mechanism**.

## 2.1 Collection key diagnostics (`StorageKey`)
`UnorderedMap<K, V>` / `SortedMap` keys must satisfy `BorshSerialize + BorshDeserialize + AsRef<[u8]>` (and `Ord` for sorted) — see the bounds at `crates/storage/src/collections/unordered_map.rs:74-77` and the `Mergeable` impl at `:260-264`. A `u64`/numeric/custom key that lacks `AsRef<[u8]>` today yields a raw, hard-to-parse `AsRef<[u8]>`-not-satisfied error.

Proposal: introduce an SDK marker trait `StorageKey: BorshSerialize + BorshDeserialize + AsRef<[u8]>` carrying `#[diagnostic::on_unimplemented]`, bound the collections' key parameter on it (or add the diagnostic to a re-exported alias). Message: *"map keys must be byte-encodable — use `String`/`Vec<u8>` or a newtype implementing `AsRef<[u8]>`; numeric keys need an explicit byte encoding."* Mirrors how #2795 made `Mergeable` (the value side) self-explaining.

## 2.2 JSON-RPC boundary diagnostics (`AppArg` / `AppReturn`)
Exported methods deserialize args via a generated `#[derive(serde::Deserialize)]` struct (`crates/sdk/macros/src/logic/method.rs:123-144`) and serialize the return via `.to_json()` (`:184-196`). A non-`Deserialize` arg or non-`Serialize` return surfaces as a serde-internal error far from the method.

Proposal: define SDK trait aliases `AppArg: serde::Deserialize` and `AppReturn: serde::Serialize` with `#[diagnostic::on_unimplemented]`, and bound the generated wrapper's arg/return types on them. Message: *"`{Self}` can't cross the JSON-RPC boundary — derive `serde::Serialize`/`Deserialize` (with `#[serde(crate = \"calimero_sdk::serde\")]`)."*

## Scope / size
[...]

PR #2810, opened 2026-06-19
feat(sdk): capability-trait diagnostics for collection keys + RPC boundary (#2801)

chefsale:
Implements #2801 — the full "capability-trait" design for SDK misuse diagnostics at two boundaries that previously produced cryptic foreign-trait errors. Each new trait is SDK-owned, carries `#[diagnostic::on_unimplemented]`, and is blanket-implemented over the underlying requirement, so it's **exactly as permissive as the bound it replaces** (behavior-preserving). Same pattern as the merged `Mergeable`/`AppEvent`/`Authorizer` work.

## `StorageKey` — collection keys
A key is addressed by its bytes, so it must be `AsRef<[u8]>` (+ borsh + `PartialEq` + `'static`). The `insert` of `UnorderedMap`/`SortedMap`/`UnorderedSet`/`SortedSet` now bounds the key on `StorageKey`:

```
UnorderedMap<u64, LwwRegister<String>>   // (calimero)> `u64` can't be used as a
                                         // collection key — keys must be byte-encodable
```

**Design note:** bound at the `insert` boundary, *not* the struct. A struct-level `K: StorageKey` is the "purest" form but rippled to **~117 errors** across one collection alone (every impl + wrappers) and changed construction semantics (you couldn't `new()` a non-key map) — disproportionate for an identical user-facing diagnostic.

## `AppReturn` — method return types
A return value (and error type) is JSON-encoded for the caller, so it must be `serde::Serialize`. Replaces the `Serialize` bounds in the return encoder (`returns.rs`) so a non-serializable return points at `derive(Serialize)` instead of failing deep inside `WrappedReturn::…to_json()`.

## `AppArg` — method arguments
An *owned* argument is decoded from the caller's JSON, so it must be `DeserializeOwned`; the macro asserts it per owned arg. Borrowed (`&_`) args stay validated by the generated input-struct's `Deserialize` derive (their lifetime ties to the input buffer, which a standalone owned assertion can't express).

## Testing note
`AppArg`/`AppReturn` live in the `#[cfg(target_arch = "wasm32")]` export body, so they're enforced on a **real wasm build** but not by the host-compiled trybuild suite (same class as the enum-state check). Verified by building `custom-key-store` for wasm and by ad-hoc bad-arg/bad-return wasm builds (both surface the `(calimero)>` messages). `StorageKey` fires in user method bodies (host-compiled) and is covered by a new `error_non_byte_key` compile_fail fixture.

## Example app
Adds `apps/custom-key-store` — a worked example of the **custom-key pattern** the `StorageKey` diagnostic guides toward: a `Slug` newtype forwarding `AsRef<[u8]>`, used as an `UnorderedMap` key, with owned-`String` args and serializable returns. The common cases (String keys, owned args, `app::Result`) were already covered by `kv-store` et al.; this fills the custom-key gap. (Dogfooding bonus: the SDK's own borsh-derive diagnostic caught an early version of the example.)

## Verification
- `cargo build --workspace --all-targets` + all apps for `wasm32` → clean
- trybuild suite (`cargo test -p calimero-sdk --test macros`) → `all ... ok`
- `calimero-storage` / `calimero-sdk` / `calimero-sdk-macros` tests → green
- `cargo fmt --check` + `cargo clippy` → clean

---

> [!NOTE]
> **Low Risk**
> Behavior-preserving trait aliases and diagnostic-only macro assertions; collection bounds are equivalent to the previous requirements at insert sites.
> 
> **Overview**
> Adds **SDK-owned capability traits** with `(calimero)>` compile-time diagnostics at three misuse boundaries, without tightening who can compile valid apps (blanket impls over the same bounds as before).
> 
> **`StorageKey`** marks types usable as map/set keys (byte-addressable via `AsRef<[u8]>` plus borsh/`PartialEq`/`'static`). **`UnorderedMap` / `SortedMap` / `UnorderedSet` / `SortedSet`** now bound **`insert`** on `K: StorageKey` instead of raw `AsRef<[u8]>`, so invalid keys like `u64` fail with a guided message at the call site. A new **`error_non_byte_key`** trybuild fixture locks in that diagnostic.
> 
> **`AppReturn`** aliases `Serialize` for method return (and error) JSON encoding in **`returns.rs`**. **`AppArg`** aliases `DeserializeOwned`; **`#[app::logic]`** codegen asserts **`assert_app_arg`** for each **owned** argument so bad args surface at the method (borrowed `&_` args still go through the generated `Deserialize` struct).
> 
> Ships **`apps/custom-key-store`** as a wasm example: a **`Slug`** newtype forwarding **`AsRef<[u8]>`** as an **`UnorderedMap`** key, plus ABI **`build.rs`**. Workspace **`Cargo.toml`** / lockfile register the new app.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5339e27fcb8b870917e7c77e716ff9c47a4c9b83. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2647,6 +2647,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "custom-key-store"
+version = "0.0.0"
+dependencies = [
+ "calimero-sdk",
+ "calimero-storage",
+ "calimero-wasm-abi",
+ "serde_json",
+ "thiserror 1.0.69",
+]
+
+[[package]]
 name = "darling"
 version = "0.20.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
     "./apps/blobs",
     "./apps/collaborative-editor",
     "./apps/kv-store",
+    "./apps/custom-key-store",
     "./apps/migration-harness-example",
     "./apps/sorted-kv-store",
     "./apps/sorted-set-store",
@@ -124,6 +125,7 @@ members = [
 
     "./apps/scaffolding-e2e",
     "./apps/kv-store",
+    "./apps/custom-key-store",
     "./apps/migration-harness-example",
     "./apps/sorted-kv-store",
     "./apps/sorted-set-store",
--- a/apps/custom-key-store/Cargo.toml
+++ b/apps/custom-key-store/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "custom-key-store"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+license.workspace = true
+publish = false
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+thiserror.workspace = true
+calimero-sdk.workspace = true
+calimero-storage.workspace = true
+
+[build-dependencies]
+calimero-wasm-abi.workspace = true
+serde_json.workspace = true
+
+[package.metadata.workspaces]
+independent = true
+
+[dev-dependencies]
+# Enables `register_crdt_merge` + the native mock host so `TestHost` can drive
+# CRDT state under `cargo test`.
+calimero-storage = { workspace = true, features = ["testing"] }
--- a/apps/custom-key-store/build.rs
+++ b/apps/custom-key-store/build.rs
@@ -1,0 +1,38 @@
+use std::fs;
+use std::path::Path;
+
+use calimero_wasm_abi::emitter::emit_manifest;
+
+fn main() {
+    println!("cargo:rerun-if-changed=src/lib.rs");
+
+    // Parse the source code
+    let src_path = Path::new("src/lib.rs");
+    let src_content = fs::read_to_string(src_path).expect("Failed to read src/lib.rs");
+
+    // Generate ABI manifest using the emitter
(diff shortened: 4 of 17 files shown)


chefsale:
Both are advisory and don't need a change here:

- **`build.rs` `.expect()` panics** — `apps/custom-key-store/build.rs` is a **verbatim copy** of `apps/kv-store/build.rs`; **16 app build scripts** in the repo use this exact `.expect(...)` pattern. Converting just this one to a `Result`-returning `main()` would make it the odd one out. If we want clean Cargo diagnostics from build scripts, that's a worthwhile but separate sweep across all app `build.rs` files — out of scope for this diagnostics PR, and I didn't want to diverge the example from the canonical template it's based on.

- **`AppArg` doc-comment in `__private`** — the `__private` module is already `#[doc(hidden)]` (lib.rs:84), so `AppArg`/`AppReturn`/`assert_app_arg` don't appear in public rustdoc. The `///` comments are intentional **maintainer** documentation (why the trait exists, the borrowed-arg caveat); the *user-facing* guidance is the `#[diagnostic::on_unimplemented]` message, which surfaces in the compile error — that's the whole point of the trait. So there's nothing user-visible to trim. Happy to switch them to `//` if you'd prefer they read unambiguously as internal notes.

CI: required checks pending; the trybuild suite, full workspace build, and all wasm app builds pass locally.

Issue #2805, opened 2026-06-18
SDK: richer host-boundary panic messages (method name + expected arg schema)


chefsale:
## Context

Follow-up to #2795. The generated method wrapper (`crates/sdk/macros/src/logic/method.rs`) panics on host-boundary failures with terse messages, e.g. `panic_str("Failed to deserialize input from JSON: {err}")` on bad args and `"Failed to serialize output to JSON: {err}"` on the return. These are where app authors lose the most debugging time, because the message doesn't say **which method** or **what was expected**.

## Proposal
Enrich the host-boundary panics emitted by the macro with actionable context:
- **Method name** in every input/output (de)serialization panic (`"<method>: failed to deserialize arguments: ..."`).
- **Expected argument shape** — the generated input struct already knows the field names/types; include them (e.g. *"expected fields: {key: String, value: String}"*) so a caller sending the wrong JSON sees what was wanted. Pairs with the existing `deny_unknown_fields` (which already rejects extras, #1646).
- Consider distinguishing "no input provided but arguments required" from "input present but malformed" (the macro already branches on these at `method.rs`).

## Why it matters
These are runtime messages (not compile-time), but they're the first thing a developer hits when wiring a client to a method. Better messages here shorten the integrate-and-debug loop and reduce "it just panics" reports.

## Scope / size
Small–medium. Localized to the codegen in `logic/method.rs`; add an e2e or integration assertion that a malformed call surfaces the method name + field hint. Keep messages within the `max_log_size` constraints noted in the WASM tracing work.

Issue #1646, opened 2025-11-11
JSONRPC: forbid external function arguments


sept-en:
# Description

Currently, if we have an app with a method `set(value: String)` and we create a JSON RPC call to it with a JSON containing extra arguments, those arguments will be just ignored. It makes sense to panic early if extra arguments are provided, this may reduce the errors during the app development (that some values won't be used).

Example JSON args that currently won't return an error (even though the method doesn't have `other_arg` parameter:
```
{
   "value": "value1",
   "other_arg": "some_value"
}
```

chefsale:
## Implementation plan

For whoever picks this up — the fix is genuinely small and well-scoped. Here's a head-start.

### 1 · The change (one attribute)

The `#[app::logic]` macro generates a per-method args struct that deserializes the JSON-RPC payload. Right now it has no `deny_unknown_fields`, so serde silently ignores extras.

**File:** `crates/sdk/macros/src/logic/method.rs`, lines **72–79**:

```rust
#[derive(::calimero_sdk::serde::Deserialize)]
#[serde(crate = "::calimero_sdk::serde", deny_unknown_fields)]   // ← add `deny_unknown_fields`
struct #input_ident #input_lifetime {
    #( #args ),*
}
```

That's the whole code change.

### 2 · Resulting behaviour

A JSON-RPC call to `set(value: String)` with payload `{"value":"v1","other_arg":"x"}` will now fail at the existing arm a few lines below (≈ line 89):

```rust
Err(err) => panic_str(&format!("Failed to deserialize input from JSON: {:?}", err))
```

with the serde error `unknown field "other_arg", expected "value"`. The server (`crates/server/src/jsonrpc/execute.rs`) turns the WASM panic into a `FunctionCallError`, so JSON-RPC clients see a clean error response.

### 3 · Tests

Two things to cover:

**a) Macro compile test (trybuild)** — confirm the attribute is well-formed and the macro still produces compiling code. Follow the existing pattern in `crates/sdk/tests/macros/` (e.g. extend `valid_args.rs`, or add a new `args_deny_unknown.rs`).

**b) Runtime test** — confirm an extra field actually rejects. Easiest path is an **e2e merobox workflow** (the project already has many under `apps/scaffolding-e2e/workflows/`):

  - Use the existing `kv-store` / `scaffolding-e2e` app.
  - Add a `call` step that POSTs an `execute` with a payload that includes a known method arg **plus** an extra field.
[...]

PR #2815, opened 2026-06-19
feat(sdk): richer host-boundary panic messages (#2805)

chefsale:
Implements #2805 — richer host-boundary panic messages.

## Problem
The generated method wrapper (`crates/sdk/macros/src/logic/method.rs`) panicked on JSON (de)serialization failures with terse, context-free messages:
```
Failed to deserialize input from JSON: missing field `value`
Failed to serialize output to JSON: ...
```
They name neither **which method** failed nor **what shape** was expected — exactly the info an author needs when wiring a client to a deployed method.

## What
Bake the method name and the expected argument schema (field names + types, which the macro already knows from the generated input struct) into every host-boundary panic:

| case | message |
|---|---|
| missing args | `set: missing arguments — expected a JSON object { key: String, value: String }` |
| malformed args | `set: failed to deserialize arguments (expected { key: String, value: String }): <err>` |
| no-arg method sent fields | `ping: takes no arguments, but the call sent JSON field(s): ["x"]` |
| bad return | `get: failed to serialize the return value to JSON: <err>` |

`method_name` and `arg_schema` are compile-time string literals built at macro-expansion (the signature is small → no `max_log_size` concern).

## Verification
These panics live in the `#[cfg(target_arch = "wasm32")]` export body, so the host-compiled trybuild suite can't assert them. Verified the generated text via `cargo expand -p kv-store --target wasm32-unknown-unknown` — e.g. kv-store's `set` produces exactly:
```rust
panic_str(&format!("{0}: failed to deserialize arguments (expected {1}): {2:?}",
                   "set", "{ key: String, value: String }", err))
```
All apps build for `wasm32`; the trybuild suite, fmt, and clippy are green (the host-compiled fixtures are unaffected — the change is in the wasm export path only).

---

> [!NOTE]
> **Low Risk**
> Diagnostics-only change in generated WASM wrappers; no change to successful call handling or host APIs.
> 
> **Overview**
> **Host-boundary errors from generated app method exports** now name the failing method and, when applicable, the expected JSON argument shape, instead of generic serde messages.
> 
> For methods with arguments, macro expansion builds a compile-time `arg_schema` string (field names + sanitized types) and uses it in **missing input**, **deserialize failure**, and **return serialize failure** panics (e.g. `set: failed to deserialize arguments (expected { key: String, value: String }): …`). For no-arg methods, unknown JSON object fields are reported as `method: takes no arguments, but the call sent unknown field(s): …`.
> 
> Changes are limited to codegen in `method.rs` and affect the `wasm32` export path only; behavior on success is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9036ff4067dd5274818b7d01d27601222e0ea21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/sdk/macros/src/logic/method.rs
+++ b/crates/sdk/macros/src/logic/method.rs
@@ -70,6 +70,21 @@ impl ToTokens for PublicLogicMethod<'_> {
 
         let arg_idents = args.iter().map(|arg| arg.ident).collect::<Vec<_>>();
 
+        // Baked into the host-boundary (de)serialization panics below so a
+        // malformed call reports *which* method failed and *what shape* was
+        // expected — the raw serde error alone names neither, which is where app
+        // authors lose the most time when wiring a client. Both are compile-time
+        // string literals (the signature is small, so no `max_log_size` concern).
+        let method_name = name.to_string();
+        let arg_schema = {
+            let fields = args
+                .iter()
+                .map(|arg| format!("{}: {}", arg.ident, arg.ty.to_token_stream()))
+                .collect::<Vec<_>>()
+                .join(", ");
+            format!("{{ {fields} }}")
+        };
+
         let init_method = modifiers
             .iter()
             .any(|modifier| matches!(modifier, Modifer::Init));
@@ -91,8 +106,8 @@ impl ToTokens for PublicLogicMethod<'_> {
                         {
                             if !fields.is_empty() {
                                 ::calimero_sdk::env::panic_str(&format!(
-                                    "Failed to deserialize input from JSON: method takes no \
-                                     arguments but received unknown field(s): {:?}",
+                                    "{}: takes no arguments, but the call sent JSON field(s): {:?}",
+                                    #method_name,
                                     fields.keys().collect::<::std::vec::Vec<_>>()
                                 ));
                             }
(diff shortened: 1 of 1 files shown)


chefsale:
Fixed the CI failure and the review comments in `d9036ff4`:

**CI — `crdt_conformance::extra_fields_to_zero_arg_method_are_rejected` (my regression).** That test (which runs a real wasm module — so #2805 *is* runtime-tested) asserts the zero-arg rejection message contains "unknown field", the #1646 contract. My reword to "JSON field(s)" dropped that phrase. Restored "unknown field(s)" while keeping the method-name enrichment:
```
entries: takes no arguments, but the call sent unknown field(s): ["bogus"]
```
Verified by forcing a fresh `scaffolding-e2e` wasm build (the conformance harness rebuilds by app-source mtime, so it's blind to an SDK-macro change — had to `rm` the stale wasm) → **21/21 conformance tests pass**.

**Review comments:**
- *`arg_schema` `{ }` for zero-arg methods* — moved its construction into the args-present branch, so it's only built when there are args.
- *`to_token_stream()` whitespace / lifetime noise* — this was worse than cosmetic: a borrowed arg leaked the internal reserved lifetime (`{ key: & 'CALIMERO_INPUT str }`). Now stripped + spacing collapsed → `{ key: & str }`.

Re-running CI.

PR #2825, opened 2026-06-19
chore(lint): consensus-tier cleanups — workspace now warning-free (3/4)

chefsale:
## What & why

Third of the 4-PR series toward gating CI on a clean clippy build (follows #2821, #2823). This PR clears the **consensus-critical crates** — `calimero-runtime`, `calimero-governance-store`, `calimero-context`, `calimero-node` — plus the two items deliberately deferred from PR 2 (`governance-types` wire enums, `node-primitives` `broadcast`/`new`).

**After this PR, `cargo clippy --workspace --all-targets` reports zero clippy/rustc warnings.** (The only remaining cargo output is the pre-existing "profiles for non-root package" manifest notices, which aren't lints — addressed in PR 4.)

No behavior changes — lint cleanups, mechanical regroupings, and documented judgment calls.

### Representative fixes
- **runtime**: ~22 doc-list continuations, ~11 `field_reassign` → struct-init, examples `fs::read`, let-else→`?`, loop-index, `result_large_err`/`run` justified allows.
- **governance-store**: docs, `ResolvedIdentity`/`AbsorbEntries` aliases, the TEE handlers refactored to take the existing `TeeAttestationClaims` struct, gossip-publish allows.
- **context**: `matches!`, `let_underscore_future`, transmute annotation, type aliases; the 5 core create/execute/update handlers get justified `too_many_arguments` allows.
- **node**: docs, mechanical (clamp/let-else/De-Morgan/struct-init/drop-ref), type aliases (`TopologyHandle`, `StaleBlobAttempts`, …), dead-code allows on documented test helpers + linux-only metric gauges, sync/handler `too_many_arguments` allows.
- **deferred**: `large_enum_variant` allow on the gossip/stored wire enums; `too_many_arguments` allow on `NodeClient::broadcast`/`new`.

### On the judgment calls (`#[allow]`)
Every allow is per-item with an inline `reason`. The rule applied throughout the series:
- **Refactor** when args/types have a cohesive grouping with few callers (e.g. `TeeAttestationClaims`, `PackageInfo`, server `ServiceConfigs`, the `StorageCallbacks`/topology type aliases).
- **Justified `#[allow]`** only when args are genuinely orthogonal on a hot consensus path with many call sites, where a param struct adds arg-mapping risk without readability gain (the create/execute/update handlers, gossip-publish, `broadcast`, `NodeClient::new`), or when the lint fights an intentional design (wire-enum boxing on the gossip hot path; `result_large_err` where the Err hands ownership back for fallback).

[...]

PR #2823, opened 2026-06-19
chore(lint): manual cleanups across non-consensus crates (2/4)

chefsale:
## What & why

Second of the 4-PR series toward gating CI on a clean clippy build (follows #2821). This PR drives **every non-consensus crate to zero clippy/build warnings** with hand-fixes that `clippy --fix` couldn't apply. **No behavior changes** — only lint cleanups and mechanical signature regroupings; callers updated in lockstep and the whole workspace builds `--all-targets`.

### Crates taken to zero
storage, storage-macros, sys, wasm-abi, sdk, sdk-macros, store, blobstore, dag, utils-actix, node-primitives (except 2 deferred fns — see below), mero-auth, primitives, server, server-primitives, context-client, context-config, governance-types (except 3 deferred — see below), meroctl, merod, merodb, and the example/migration apps.

### Representative fixes
- **type_complexity** → named `type` aliases (storage env/js/crdt_meta, node-primitives AliasEntry, etc.).
- **large_enum_variant** → boxed the large variant (storage `Entry::Occupied`, sdk-macros `LogicArg`/`LogicMethod`).
- **too_many_arguments** → grouped cohesive params into structs: `ApplicationMeta::new`→`PackageInfo`, server `with_auth`→`ServiceConfigs` / `mount_runtime_services`→`RuntimeServiceDeps`, context-client `create_context`→`CreateContextParams`, meroctl `create_context`→`CreateContextArgs` / `watch_and_reload`→`ReloadPaths`.
- **result_large_err** → boxed large `Err` types; **borrowed_box** → `&dyn Trait` instead of `&Box<dyn Trait>`.
- **assertions_on_constants** → compile-time `const _: () = assert!(…)`.
- Stale `#[expect(…)]` left over from #2821's restriction-lint removal, doc-list continuations, unused imports/vars, `.clamp()`, private-in-public, module-inception, dead test instrumentation, etc.
- One justified per-item `#[allow]`: sdk-macros `let_and_return` (clippy's rewrite extends a borrow of `args` past its drop and fails to compile) — with an explanatory comment.

### Deferred to PR 3 (consensus tier)
These ripple into the consensus crates and/or sit on hot paths, so they belong with that review:
- `governance-types` (3): `large_enum_variant` on the `NamespaceTopicMsg`/`GroupTopicMsg`/`StoredNamespaceEntry` **wire enums** — boxing adds a per-message heap alloc on the gossip path and ripples to match-sites in node/governance-store; wants a deliberate perf call.
[...]

PR #2826, opened 2026-06-19
ci: gate CI on a warning-free workspace (4/4)

chefsale:
## What & why

Final PR of the series (follows #2821, #2823, #2825). The first three drove the whole workspace to **zero clippy/rustc warnings**; this one flips CI so it can't regress.

### The gate
`ci-checks.yml`'s clippy step changes from:
```yaml
cargo clippy -- -A warnings        # allowed every warning, lib-only
```
to:
```yaml
cargo clippy --workspace --all-targets -- -D warnings
cargo clippy -p calimero-storage --all-targets --features testing -- -D warnings
```
Any clippy- or rustc-level warning in any workspace target (libs, bins, tests, examples) now fails CI. The second line covers `calimero-storage`'s feature-gated `testing` harness, which the default-feature pass doesn't reach. Devs can run the same commands locally before pushing.

**Verified locally on this branch:** both commands exit 0 (zero warnings).

### Why CI-only (not `[workspace.lints]`)
The earlier plan considered `[workspace.lints]` for declarative enforcement. With **169 workspace members**, that means `lints.workspace = true` in 169 manifests plus `warnings = "deny"` applied to every local `cargo build` — a lot of churn and real toolchain-version fragility (a newer rustc adding a warning would break local builds). The CI gate gives the same regression protection without that cost; if we later want local enforcement too, it's an additive follow-up.

### Tidy
Dropped the redundant `[profile.app-release]` blocks from `apps/abi_conformance` and `apps/state-schema-conformance`. Cargo ignores `[profile]` on non-root members (the root already defines `app-release`, which the build scripts use via `--profile app-release`), so these only emitted "profiles … will be ignored" notices. Removal is a no-op functionally.

## Series complete
1. #2821 — policy + auto-fix ✅ merged
2. #2823 — non-consensus crates ✅ merged
3. #2825 — consensus tier (workspace warning-free) — open
4. **This PR** — the gate

> Note: stacked on #2825. Best merged after it (or rebased onto master once #2825 lands).

---

> [!NOTE]
> **Low Risk**
> Changes are CI policy, Cargo.toml cleanup, and style/clippy fixes with no intended runtime behavior change.
> 
> **Overview**
> **CI now fails on any clippy/rustc warning** across the full workspace. The Rust job’s clippy step switches from allowing warnings on a default pass to `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`, with a short comment documenting the gate and the `calimero-storage/testing` feature so feature-gated test harness code is linted in the same run.
> 
> **Manifest tidy:** `[profile.app-release]` is removed from `apps/abi_conformance` and `apps/state-schema-conformance` because Cargo ignores profile tables on non-workspace-root packages (root already defines `app-release`), which only produced “profiles will be ignored” notices.
> 
> **Remaining warning fixes** bundled for the new gate: `RootOp` is imported only under `#[cfg(test)]` in `scope_projection.rs`; tee-attestation uses inline `format!` capture and `split('/').next_back()` instead of `.last()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c298ebdbe84881ea403501a97bc566c846bdf368. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -47,15 +47,12 @@ jobs:
         run: cargo fmt --check
 
       # No-warnings gate: every workspace target (libs, bins, tests, examples)
-      # must be clippy- and rustc-warning-free. Run the same command locally
-      # before pushing: `cargo clippy --workspace --all-targets -- -D warnings`.
+      # must be clippy- and rustc-warning-free. `--features calimero-storage/testing`
+      # pulls in that crate's feature-gated `testing` harness so it's linted in the
+      # same single pass. Run the same command locally before pushing.
       - name: Cargo clippy
         if: ${{ !cancelled() }}
-        run: |
-          cargo clippy --workspace --all-targets -- -D warnings
-          # calimero-storage's feature-gated `testing` harness isn't reached by
-          # the default-feature pass above, so lint it explicitly.
-          cargo clippy -p calimero-storage --all-targets --features testing -- -D warnings
+        run: cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings
 
       - name: Cargo test
         if: ${{ !cancelled() }}
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -46,9 +46,16 @@ jobs:
       - name: Cargo format
         run: cargo fmt --check
 
+      # No-warnings gate: every workspace target (libs, bins, tests, examples)
+      # must be clippy- and rustc-warning-free. Run the same command locally
+      # before pushing: `cargo clippy --workspace --all-targets -- -D warnings`.
       - name: Cargo clippy
         if: ${{ !cancelled() }}
-        run: cargo clippy -- -A warnings
+        run: |
+          cargo clippy --workspace --all-targets -- -D warnings
+          # calimero-storage's feature-gated `testing` harness isn't reached by
+          # the default-feature pass above, so lint it explicitly.
+          cargo clippy -p calimero-storage --all-targets --features testing -- -D warnings
 
       - name: Cargo test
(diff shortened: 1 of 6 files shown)


chefsale:
Thanks — good catches. Addressed in 843adba4:

**Orphaned `overflow-checks = true`** (MeroReviewer #1/#2, Cursor Bugbot) — real bug, fixed. When I deleted the redundant `[profile.app-release]` block, its last key (`overflow-checks = true`) was left dangling as a bare top-level key. Removed from both `apps/abi_conformance` and `apps/state-schema-conformance`. Nothing is lost: cargo already ignores `[profile]` on non-root members, and the workspace-root `[profile.app-release]` (which the build scripts actually use via `--profile app-release`) already sets `overflow-checks = true`. `cargo metadata` parses clean.

**Two clippy passes / cache miss** (inline suggestion) — collapsed into a single command:
`cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`
`--features calimero-storage/testing` activates the feature-gated `testing` harness in the same pass, so there's no second cache-missing compile. Verified exit 0 locally.

**`drop(state)` in session.rs** (MeroReviewer #3) — this is in #2825 (already merged), not in this PR's diff (which is only `ci-checks.yml` + the two app manifests). It was reviewed and resolved there: `publish_state` takes `&self` (not `&mut self`) and doesn't touch `self.state`, so the immutable borrow ends at its last use with no overlap — the original `drop` was a no-op on a reference (why `drop_ref` fired).

**Doc bot (AGENTS.md / CONTRIBUTING.md / docs/)** — not applicable; this is a CI/manifest change with no architecture impact.

PR #2827, opened 2026-06-19
fix(context): fold open-subgroup MemberJoinedOpen as inheritance-derived

chefsale:
## What

Folds `RootOp::MemberJoinedOpen` (open-subgroup inheritance self-join) as a `Noop` graph node instead of a direct `MemberAdded`, because in the live model it's an inheritance *proof* (no persistent `GroupMember` row — apply requires `check_path == Inherited`), re-derived from the anchor each time. The projection's inheritance walk (`AclView::is_member_at_cut`) derives the subgroup membership from the foldable anchor membership + subgroup tree + visibility + cap, so it tracks the anchor **both ways** (revoked on root-removal, restored on rejoin).

## Why

The merged #2790 folded `MemberJoinedOpen` as a permanent direct membership, which survives anchor removal → the over-grant the divergence shadow caught in `group-remove-from-root-revokes-inherited`. This is the **walk-logic** half of the grant-direction fidelity work.

## Validation

New `tests/projection_membership_equivalence.rs` — a **deterministic** fold-equivalence harness: applies governance ops to a live store (`MembershipRepository`/`check_path` reference) and folds the same ops into the projection, asserting `member_at_cut` matches live across:
- inherited join → root-removal **revokes** (the over-grant), and
- leave → rejoin **restores** (the under-grant).

Both pass — reducing the e2e over/under-grant saga to unit tests. Confirmed the structural ops (`GroupCreated`, `SubgroupVisibilitySet`) already flow through the foldable namespace feed in production.

## Scope

This validates the **deny-direction co-authorizer** holds with the new fold (the gate must stay zero). Re-enabling the **grant direction** (sole authority) is a follow-up once this is green. Part of the unified-causal-log cutover (P5), follows #2790.

---

> [!NOTE]
> **Medium Risk**
> Changes how the membership co-authorizer interprets open-subgroup joins and partial backfills; incorrect behavior could false-deny or mis-shadow live auth, though grant authority and scope are limited and heavily tested.
> 
> **Overview**
> Fixes **open-subgroup inheritance** in the unified governance projection so it matches live membership instead of over-granting after root removal.
> 
> **`RootOp::MemberJoinedOpen`** now folds as **`OpPayload::Noop`** (DAG node only), not a permanent **`MemberAdded`**. Inherited subgroup access is derived via **`AclView::is_member_at_cut`** from anchor root membership, subgroup tree, visibility, and caps—so it **revokes** when the member is removed from the root and **restores** on rejoin.
> 
> **`member_at_cut`** (deny co-authorizer) returns **`None`** when the cited cut’s ancestry is not fully folded, avoiding **false denies** during governance backfill races (especially for inherited memberships that need a long folded chain).
> 
> Adds **`projection_membership_equivalence.rs`** with deterministic live-vs-projection tests for inherited join → root removal, leave → rejoin, and incomplete-ancestry deferral; updates the unit test for the Noop fold.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10cfe2d6910b8f3b2e542e89e031876412c2e90c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -24,7 +24,7 @@ use calimero_governance_store::{
     CapabilitiesRepository, MembershipRepository, MetaRepository, NamespaceDagService,
     NamespaceOpLogService, NamespaceRepository,
 };
-use calimero_governance_types::{GroupOp, NamespaceOp, SignedNamespaceOp};
+use calimero_governance_types::{GroupOp, NamespaceOp, RootOp, SignedNamespaceOp};
 use calimero_op::{Op, OpPayload, ScopeId};
 use calimero_op_adapter::{payload_from_group_op, payload_from_root_op, set_writers_payload};
 use calimero_primitives::context::{ContextId, GroupMemberRole};
@@ -793,7 +793,6 @@ mod tests {
     use calimero_context_config::types::{
         ContextGroupId, GroupInvitationFromAdmin, SignedGroupOpenInvitation,
     };
-    use calimero_governance_types::RootOp;
     use calimero_op::OpPayload;
     use calimero_primitives::context::GroupMemberRole;
     use calimero_storage::entities::OpMask;
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -218,16 +218,17 @@ pub fn op_from_namespace_op(
     parents: &[[u8; 32]],
 ) -> Op {
     let payload = match &signed.op {
-        // NOTE on `MemberJoinedOpen` (open-subgroup inheritance join): live treats
-        // it as an inheritance PROOF (no persistent direct row; re-derived from the
-        // anchor each time, so revoked on anchor removal). Folding it as a direct
-        // `MemberAdded` makes the projection see the inherited member as present —
-        // which is correct for the DENY direction (never false-deny a live member,
-        // incl. leave/rejoin-via-inheritance) but over-permissive for the GRANT
-        // direction (survives anchor removal). The co-authorizer only ACTS on the
-        // deny direction, so `MemberAdded` is the right fold today; faithfully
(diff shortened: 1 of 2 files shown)


chefsale:
Pulled the master merge and addressed the review (commit `10cfe2d6`):

- **Compile fix** — the merge dropped `RootOp` from the crate-level import while `op_from_namespace_op` still folds `RootOp::MemberJoinedOpen`; restored it (and removed the now-redundant tests-module re-import that `use super::*` already provides) so the lib compiles.
- **Invitation expiry** — `expiration_timestamp` → `u64::MAX` (the prior `365*24*3600` resolves to ~1971 as an absolute wall-clock and would silently break the tests once the apply path enforces expiry).
- **Inherited-join `Some(true)` could pass via the live fallback** — documented that it's the at-cut **fold** that drives it: `MemberJoinedOpen` writes no direct subgroup row, so `member_at_cut`'s `role_of` materialized fallback can't fire here.
- **Asymmetric removal path** — documented why the root removal uses the repo write rather than `apply_signed_namespace_op`: `MemberRemoved` is an encrypted `GroupOp`, and round-tripping it needs a per-group key/ciphertext this harness doesn't model (the joins are cleartext `RootOp`s, hence signed-applied). The equivalence under test is fold-vs-materialized-membership, which `remove_member` reproduces faithfully for `is_member`.

Deferred: switching the hardcoded op IDs to `Op::compute_id` — a broader test refactor; the hand-wired parent chains exercise the ancestry walk correctly for these focused cases. Can do as a follow-up if you'd prefer.

Also added `projection_defers_when_cut_ancestry_incomplete`, pinning the backfill-lag deny-defer fix (the single transient divergence the gate caught). e2e is green with **0 divergence-gate failures**.

PR #2828, opened 2026-06-19
feat(node): projection sole authority — grant over live's reject (gated)

chefsale:
**Stacked on #2827** (base = `feat/projection-inheritance-model`).

## What
Flips the `MembershipReject` arm to **act** on the projection: when `member_at_cut_authoritative` returns `Some(true)`, the write is **authorized**, overriding live's membership-reject. Paired with the `Authorized` arm's deny (already on master), the projection is now the **sole membership decider in both directions**; live supplies only structural classification + the buffer-on-unknown-heads liveness gate.

## Why it's safe now
- Depends on #2827's inheritance fold (`MemberJoinedOpen` → `Noop`, derived by the walk), which makes the at-cut walk match live in both directions.
- `member_at_cut_authoritative` grants ONLY on **complete folded ancestry** + the at-cut inheritance walk (never the materialized fallback), so it can't out-run live.
- Cross-validated against the live reference in `tests/projection_membership_equivalence.rs`: the grant resolver must **not** grant where live rejects (over-grant guard) and must grant on inherited join / rejoin.
- Any disagreement trips the hard divergence gate (`membership-cut-grant`), so a wrong grant fails CI.

## Review note
Merge **after #2827**. This is the sole-authority milestone of the unified-causal-log cutover (P5); it unblocks F5 (delete the live resolver) once the remaining consumers are migrated.

---

> [!NOTE]
> **High Risk**
> Changes who may apply peer state deltas when live denies membership on the main gossip path; a mistaken authoritative grant would accept unauthorized writes, though grants are gated on fully-folded ancestry only and CI equivalence tests plus the divergence gate are meant to catch regressions.
> 
> **Overview**
> **Membership on the primary gossip path:** when live returns `MembershipReject`, the handler no longer always drops the delta. It refreshes the namespace projection (via new shared `refresh_projection_for_cut`), calls `projection_member_at_cut_authoritative`, and **continues to apply** if that returns `Some(true)`—with `unified_projection_divergence` / `membership-cut-grant` logging and `observe_peer_identity` without role. `None` or `Some(false)` still reject (projection concurs with live).
> 
> **Refactor:** projection backfill logic is extracted from `projection_member_at_cut` into `refresh_projection_for_cut`; deny-direction reads and the new grant wrapper both use it so both arms see the same fold.
> 
> **Tests:** `projection_membership_equivalence.rs` asserts `member_at_cut_authoritative` grants on inherited join/rejoin and **never** `Some(true)` when live revokes after root removal (over-grant guard).
> 
> Other `MembershipReject` sites (sync parent-fetch, etc.) stay live-only per comments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a13aa6d78578ed07f2063105b08f214364fe3eb9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/tests/projection_membership_equivalence.rs
+++ b/crates/context/tests/projection_membership_equivalence.rs
@@ -253,6 +253,13 @@ fn projection_matches_live_across_inherited_join_and_root_removal() {
         Some(true),
         "projection must agree the joiner is a member after the inherited join"
     );
+    // The GRANT resolver (sole-authority path) must also see the member: complete
+    // ancestry is folded, so it returns the at-cut verdict.
+    assert_eq!(
+        proj.member_at_cut_authoritative(&store, subgroup, &joiner, &[id2]),
+        Some(true),
+        "authoritative grant resolver agrees the joiner is a member after the join"
+    );
 
     // (3) admin removes the joiner from the NAMESPACE ROOT only (a GroupOp on the
     // root). Live: removes the root row; the subgroup has no direct row, so the
@@ -297,6 +304,13 @@ fn projection_matches_live_across_inherited_join_and_root_removal() {
         Some(false),
         "projection must revoke the inherited subgroup access after root removal \
          (matching live) — granting here is the over-grant"
+    );
+    // The GRANT resolver MUST NOT grant where live rejects — this is the
+    // sole-authority safety property (it can never over-authorize).
+    assert_ne!(
+        proj.member_at_cut_authoritative(&store, subgroup, &joiner, &[id3]),
+        Some(true),
+        "authoritative grant resolver must NOT grant a write live rejected (over-grant)"
     );
 }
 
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed the review feedback (folded into the sole-authority commit):

- **Grant path skipped projection refresh** (Bugbot + 2 reviewers) — the grant arm now calls a new `projection_member_at_cut_authoritative` wrapper that shares `refresh_projection_for_cut` with the deny arm, so a cold/stale fold no longer spuriously abstains and drops a legitimate inherited-join write. `None`/`Some(false)` after refresh stay safe (reject, never over-grant).
- **`observe_peer_identity` absent on the grant path** — now recorded (with no membership annotation, since the authoritative read returns no role) so a peer authorized solely by the projection still feeds anchor-preferred sync peer selection.
- **Other `MembershipReject` sites (sync delta-request, parent-fetch replay)** — kept live-only for now, intentionally; documented the convergence argument inline: they DROP, which is recoverable via gossip re-delivery or hash-heartbeat snapshot sync, so a recovery-path drop can't cause permanent divergence. They migrate once this gossip path is validated.
- **`warn!` noise on grant** — kept `warn!`: inherited-join writes hit the **Authorized** (deny co-author) arm, not this grant arm (live correctly authorizes inherited members), so this path is rare, not per-write — and the `warn!` carries the `unified_projection_divergence` marker the hard gate greps.

e2e re-dispatched; the prior run was green with **0 divergence-gate failures**.

PR #2829, opened 2026-06-19
feat(node): governance-drain re-check resolves via projection (gated)

chefsale:
## What

Third stacked step of the F5 consumer migration (unified causal log, core#2716 Phase 5). Migrates the **governance-pending buffer drain** off the live `acl_view_at` resolver onto the projection's `member_at_cut_authoritative` — the same sole-authority resolver the data-write grant path uses (#2828, the base of this PR).

Stack:
- #2827 — fold open-subgroup `MemberJoinedOpen` as inheritance-derived (Noop)
- #2828 — sole authority: projection grants over live's reject (gated) ← **base**
- **this PR** — governance-drain re-check → projection

## How

On drain, the buffered delta's membership is re-resolved at the cut the author signed at (`pos.governance_dag_heads`). The three projection verdicts map onto the drain's three outcomes:

| `member_at_cut_authoritative` | meaning | outcome |
|---|---|---|
| `Some(true)` | complete cited ancestry folded + at-cut walk confirms membership | re-apply |
| `Some(false)` | folded ancestry says not-a-member | drop |
| `None` | cited ancestry not fully folded yet | re-buffer (old `Unknown`) |

`member_at_cut_authoritative` grants **only** when the complete cited ancestry is folded and the at-cut inheritance walk confirms membership, so it never over-authorizes. `None` faithfully reproduces the old "governance hasn't caught up, retry" case, preserving the drain's liveness (the per-delta attempt counter + hash-heartbeat recovery are unchanged).

## Safety

`acl_view_at` is retained here **only as the divergence cross-check**: any disagreement between projection and live trips the hard `unified_projection_divergence` gate — grant plane when the projection applies a delta live would drop (the dangerous direction), deny plane otherwise. So a wrong verdict fails an e2e scenario **and** the gate. The live resolver is deleted in F5 once every consumer is validated divergence-free.

## Test

- `cargo build`/`clippy`/`fmt` clean.
- Deterministic fold-equivalence (`crates/context/tests/projection_membership_equivalence.rs`) already cross-validates `member_at_cut_authoritative` against the live resolver in both directions across inherited-join + removal + leave/rejoin.
- e2e gate validates zero divergence on the drain path.

---

> [!NOTE]
> **Medium Risk**
> Changes authorization on deferred state-delta apply (grant vs drop); wrong projection verdicts could lose or wrongly apply deltas, though divergence logging and existing drain limits mitigate rollout risk.
> 
> **Overview**
> **Governance-pending drain** now decides whether to re-apply, drop, or re-buffer buffered state deltas using the projection’s **`member_at_cut_authoritative`** at the author’s signed governance cut, instead of **`acl_view_at`** as the live resolver.
> 
> The three projection outcomes map to the same drain behavior as before: **`Some(true)`** → re-apply via `apply_authorized_state_delta`; **`Some(false)`** → drop (metrics still distinguish removed vs never-member when live agrees); **`None`** → re-buffer when cut ancestry isn’t fully folded (replacing the old **`Unknown`** path). Lookup errors from live are no longer a separate drop arm.
> 
> **`acl_view_at`** is kept only as a **divergence cross-check** on decisive projection verdicts (`Some(true)` / `Some(false)`), logging **`unified_projection_divergence`** on grant-plane or deny-plane mismatch. It is **not** called on **`None`**, to avoid false positives while governance is still catching up. Re-buffer attempt limits and drain structure are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed08aa284e3b2a681f0a3e79977ac13c8835c906. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/handlers/state_delta/buffering.rs
+++ b/crates/node/src/handlers/state_delta/buffering.rs
@@ -21,24 +21,25 @@ use super::{
 /// delta's authorization status against current local governance state and
 /// dispatching by outcome.
 ///
-/// Outcomes per drained delta:
-/// * `Member` — the referenced governance heads are now known and the author
-///   is authorized. The buffered delta is reconstructed into a
+/// Authorization is resolved by the projection (`member_at_cut_authoritative`)
+/// at the cut the author signed at; outcomes per drained delta:
+/// * `Some(true)` — the cited governance ancestry is fully folded and the
+///   author is a member at that cut. The buffered delta is reconstructed into a
 ///   [`StateDeltaMessage`] and applied directly via
 ///   [`apply_authorized_state_delta`]. Gossipsub does *not* auto-rebroadcast
 ///   already-delivered messages, so dropping here would lose the delta
 ///   permanently — recovery would only happen via hash-heartbeat divergence
 ///   detection triggering snapshot sync.
-/// * `Removed` / `NeverMember` / `Err` — author is permanently not
-///   authorized at this position; drop with a warn log.
-/// * `Unknown { needed }` — governance still hasn't caught up; push the
-///   delta back into the pending buffer.
+/// * `Some(false)` — folded ancestry says the author is not a member at this
+///   cut; drop with a warn log.
+/// * `None` — the cited ancestry isn't fully folded yet (governance hasn't
+///   caught up); push the delta back into the pending buffer.
 ///
 /// Calls `apply_authorized_state_delta` directly (not `handle_state_delta`)
 /// so the call graph stays linear — no async recursion, no per-recurse
-/// future allocation. The cross-DAG check we just performed via
-/// `membership_status_at` is the same check `handle_state_delta` would have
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed the MeroReviewer catch (commit `8ade43f1`):

- **False-positive divergence on `None`** — the cross-check now runs **only on a decisive projection verdict**, never on `None`. `None` = "cut ancestry not folded yet" (re-buffer), which is a deferral, not a disagreement; comparing it against live would have tripped the gate whenever live already saw `Member` while the projection hadn't caught up. On `Some(true)` a live `Removed`/`NeverMember` trips the grant plane; on `Some(false)` a live `Member` trips the deny plane.
- **Wasted DAG walk** — `acl_view_at` is now computed lazily inside the two decisive arms, so the re-buffer (`None`) path no longer pays for an unused live walk.

PR #26, opened 2024-02-28
integration: networking + storage + runtime + consensus

miraclx:
yay.

### Changes

1. We now have a new `run` subcommand.

2. A new `store` section is now part of the config, you can add this to existing config to get set up:

    ```toml
    [store]
    path = "data"
    ```

#### First, compile the kv-store application

```console
$ ./apps/kv-store/build.sh
   Compiling calimero-sdk v0.1.0 (/git/calimero-is-near/cali2.0-experimental/crates/sdk)
   Compiling kv-store v0.1.0 (/git/calimero-is-near/cali2.0-experimental/apps/kv-store)
    Finished app-release [optimized] target(s) in 2.35s
```

#### Create a data folder for all configs

```console
$ mkdir data
```

#### Spin up a coordinator node

```console
$ cargo run -p calimero-node -- --home data/coordinator init --port 2429
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/calimero-node --home data/coordinator init`
2024-02-28T20:02:57.715257Z  INFO calimero_node::cli::init: Generated identity: PeerId("12D3KooWCiyHe2yeu53qSyRs4g9sTGwgEPjR8iGdi4XG5iv11TgH")
2024-02-28T20:02:57.725088Z  INFO calimero_node::cli::init: Initialized a chat node in "data/coordinator"

$ cargo run -p calimero-node -- --home data/coordinator run apps/kv-store/res/kv_store.wasm --node-type coordinator
```

#### Spin up node 1

```console
$ cargo run -p calimero-node -- --home data/node1 init --port 2428
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/calimero-node --home data/node1 init`
2024-02-28T20:02:57.715257Z  INFO calimero_node::cli::init: Generated identity: PeerId("12D3KooWHJMh2hv9wai6UqPoHf5jED2gNaUbTTx6ZThAUqroCgtF")
2024-02-28T20:02:57.725088Z  INFO calimero_node::cli::init: Initialized a chat node in "data/node1"

$ cargo run -p calimero-node -- --home data/node1 run apps/kv-store/res/kv_store.wasm
```

#### Spin up node 2

```console
$ cargo run -p calimero-node -- --home data/node2 init --port 2430
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/calimero-node --home data/node2 init`
[...]

PR #2841, opened 2026-06-19
feat(context): membership query auth-gates resolve via projection (gated)

chefsale:
## What

F5 consumer migration (unified causal log, core#2716), stacked on #2840. Migrates the four **read-endpoint membership auth-gates** off the live `MembershipRepository::is_member` onto the unified projection:
- `list_all_groups` (filter groups by node membership)
- `list_group_contexts` (gate)
- `get_cascade_status` (gate)
- `get_group_upgrade_status` (gate)

## How

Unlike the node, the context-layer query handlers hold **no cached `ScopeProjections`** — so they build one **fresh from the store per query** (`ScopeProjections::member_now_ephemeral`: resolve namespace → fold its governance DAG → `member_at_cut` at the current heads). Per-query rebuild is fine for these low-frequency read endpoints; the governance history is small, and hot paths still use the node's cached projection.

`member_now_checked` centralizes the decision so each site is a one-line swap: act on the projection's verdict, fall back to live only when the projection can't decide (`None`), and trip `unified_projection_divergence` (plane `membership-query`, caught by the same e2e gate) when the two definitely disagree.

## Scope

Boolean `is_member` gates only. The member-**enumeration** endpoints (`list_group_members`, `get_group_info`, `get_migration_status`) need a projection member-enumeration API (direct + inherited members) and follow as a separate PR.

## Test

- `cargo build`/`clippy --all-targets` (warning-free gate)/`fmt` clean; equivalence harness 3/3 green.
- e2e `membership-query` divergence gate validates equivalence on real traffic. (Production has the full governance history in the namespace DAG, which `member_now_ephemeral` folds; the deterministic harness uses direct ingest for structural ops so it doesn't exercise the ephemeral rebuild — e2e covers that.)

Stacked on #2840 (#26 sync peer selection).

---

> [!NOTE]
> **Medium Risk**
> Changes authorization for group read APIs (membership-sensitive), but projection verdicts are cross-checked against live with fallback when undecidable and e2e divergence monitoring.
> 
> **Overview**
> Moves **four read-endpoint membership gates** from live `MembershipRepository::is_member` to **`ScopeProjections::member_now_checked`**: `list_all_groups`, `list_group_contexts`, `get_cascade_status`, and `get_group_upgrade_status`.
> 
> Context handlers have no cached projection, so **`member_now_ephemeral`** builds one per query (namespace resolve → governance DAG fold → `member_at_cut` at current heads). **`member_now_checked`** uses that verdict when decidable, falls back to live when the projection returns `None`, and logs **`unified_projection_divergence`** (plane `membership-query`) when projection and live disagree.
> 
> **`list_all_groups`** now **skips** a group on membership-check errors (with a warn) instead of failing the entire listing, matching existing behavior for missing node identity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 05c0ce9ad9c4d7911717bf3c5b8f9403ef2df0bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/get_cascade_status.rs
+++ b/crates/context/src/handlers/get_cascade_status.rs
@@ -1,6 +1,6 @@
 use actix::{ActorResponse, Handler, Message};
 use calimero_context_client::group::{CascadeStatusEntry, GetCascadeStatusRequest};
-use calimero_governance_store::{MembershipRepository, NamespaceRepository, UpgradesRepository};
+use calimero_governance_store::{NamespaceRepository, UpgradesRepository};
 use eyre::bail;
 
 use crate::ContextManager;
@@ -47,9 +47,11 @@ impl Handler<GetCascadeStatusRequest> for ContextManager {
             let Some((node_identity, _)) = self.node_namespace_identity(&namespace_id) else {
                 bail!("node has no group identity configured");
             };
-            if !MembershipRepository::new(&self.datastore)
-                .is_member(&namespace_id, &node_identity)?
-            {
+            if !crate::scope_projection::ScopeProjections::member_now_checked(
+                &self.datastore,
+                &namespace_id,
+                &node_identity,
+            )? {
                 bail!("node is not a member of namespace '{namespace_id:?}'");
             }
             collect_cascade_status(&self.datastore, &namespace_id)
--- a/crates/context/src/handlers/get_group_upgrade_status.rs
+++ b/crates/context/src/handlers/get_group_upgrade_status.rs
@@ -1,6 +1,6 @@
 use actix::{ActorResponse, Handler, Message};
 use calimero_context_client::group::GetGroupUpgradeStatusRequest;
-use calimero_governance_store::{MembershipRepository, UpgradesRepository};
+use calimero_governance_store::UpgradesRepository;
 use eyre::bail;
 
 use crate::ContextManager;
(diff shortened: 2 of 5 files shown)


chefsale:
Addressed (commit `05c0ce9a`):

- **Stale cut vs fold (Bugbot, high):** `member_now_ephemeral` now reads the cut heads **after** the fold. The fold covers the head `collect_namespace_ops` observed, and the cut heads are read no earlier — so a concurrent head-advance can't leave the cut naming a pre-revocation tip the fold moved past; `member_at_cut`'s completeness guard returns `None` (defer to live) in that race instead of deciding against a stale cut.
- **Silent `None` on store fault:** `collect_namespace_ops` returning `None` now logs a warn and returns `None` explicitly, rather than silently reading an empty projection.
- **Live error masking a projection verdict:** `member_now_checked` now decides on the projection first and cross-checks live **best-effort** — a transient live-read error no longer denies a verdict the projection already formed. Live is read authoritatively (error propagates) only when the projection returns `None`.
- **`list_all_groups` aborting the whole listing:** now skips a group on a per-group membership error (log + continue), matching the `node_namespace_identity` miss.

Deferred (noted): the per-group ephemeral rebuild in `list_all_groups` (N groups → N folds) — acceptable for this low-frequency admin read; a per-namespace rebuild reuse is a clean follow-up optimization if it shows up in profiles.

e2e was green with **0 divergence-gate failures** before these (read-path-only) changes; re-validating.

PR #2843, opened 2026-06-19
feat(context): shadow member-enumeration endpoints against the projection

chefsale:
## What

Second half of the query-API migration (F5, core#2716), stacked on #2841. Cross-checks the three membership-**enumeration** endpoints against the projection's effective-member set:
- `list_group_members` (full member list)
- `get_group_info` (member count)
- `get_migration_status` (subtree cohort)

## How — shadow, not flip

Enumeration is intricate (open-subgroup parent walk, deny-list, via-admin roles), so this lands as a **SHADOW**: each handler still **returns the live result**, but logs `unified_projection_divergence` (plane `membership-enum`, caught by the e2e gate) on any set/count difference vs the live `list ∪ enumerate_inherited` it already computes. That validates the projection enumeration on real traffic before any flip — risk-free for clients.

Adds:
- `ScopeProjections::member_identities_now_ephemeral` — one fold of a fresh ephemeral projection, then the candidate universe (every direct member of any group in the view + group/root admins) filtered through the same at-cut inheritance walk the boolean reads use.
- `ScopeProjections::shadow_check_member_enum` — the set-comparison + divergence log.

Also migrates `list_group_members`' and `get_group_info`'s boolean `is_member` gates to `member_now_checked`, matching the #2841 gate handlers.

## Test

- `cargo build`/`clippy --all-targets`/`fmt` clean; equivalence harness 3/3 green.
- e2e `membership-enum` gate validates the enumeration on real traffic. If divergence surfaces (deny-list / via-admin role edge cases), it's caught here as a shadow — clients are unaffected — and is the signal to refine before the flip.

Stacked on #2841 (#25 query gates).

---

> [!NOTE]
> **Medium Risk**
> Read-path changes add extra governance DAG folds and alter membership gates when projection succeeds; API responses stay live-only, but wrong projection logic could affect access until fallback or could spam divergence noise in production.
> 
> **Overview**
> Adds **shadow** cross-checks for membership **enumeration** on `list_group_members`, `get_group_info`, and migration cohort building in `get_migration_status`. Handlers still return live `list ∪ enumerate_inherited` results; mismatches against the folded governance projection only emit `unified_projection_divergence` with plane `membership-enum`.
> 
> **`scope_projection`:** New helpers fold each namespace’s governance DAG once per request (`ephemeral_projection` / `ephemeral_fold`) and derive effective member identity sets via `member_identities_in_view` (at-cut inheritance, deny-list asymmetry, namespace-leave cascade, materialized fallback for unfolded groups). `member_now_checked_with` and `shadow_member_enum_with` reuse that fold for gates and set comparison. `member_identities_subtree_ephemeral` evaluates an entire subtree at a **single** cut for migration cohort shadowing.
> 
> **Handlers:** `list_group_members` and `get_group_info` switch the `is_member` gate to the projection path when the fold succeeds; `get_group_info` also dedupes inherited enumeration for count vs shadow. `collect_migration_cohort` compares the live subtree union to the projection cohort without changing the returned cohort.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5196a6130a7741935db0316e4050aeb1e89b1d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/get_group_info.rs
+++ b/crates/context/src/handlers/get_group_info.rs
@@ -51,25 +51,27 @@ impl Handler<GetGroupInfoRequest> for ContextManager {
                     .enumerate_inherited(&group_id)?
                     .len()) as u64;
 
-            // SHADOW: validate the projection's effective-member count against the
-            // live union (logs `membership-enum` divergence; still returns live).
-            if let Some(projected) =
-                crate::scope_projection::ScopeProjections::member_identities_now_ephemeral(
-                    &self.datastore,
-                    &group_id,
-                )
-            {
-                if projected.len() as u64 != member_count {
-                    tracing::warn!(
-                        marker = "unified_projection_divergence",
-                        plane = "membership-enum",
-                        group_id = ?group_id,
-                        projection_count = projected.len(),
-                        live_count = member_count,
-                        "query-enum: projection effective-member count differs from live"
-                    );
-                }
-            }
+            // SHADOW: compare the projection's effective-member SET against the live
+            // union (not just the count — equal counts with different members would
+            // otherwise slip through). Logs `membership-enum` divergence; still
+            // returns the live `member_count`.
+            let live_ids: std::collections::BTreeSet<_> =
+                MembershipRepository::new(&self.datastore)
+                    .list(&group_id, 0, usize::MAX)?
+                    .into_iter()
+                    .map(|(pk, _)| pk)
+                    .chain(
+                        MembershipRepository::new(&self.datastore)
+                            .enumerate_inherited(&group_id)?
+                            .into_iter()
+                            .map(|(pk, _)| pk),
+                    )
+                    .collect();
+            crate::scope_projection::ScopeProjections::shadow_check_member_enum(
+                &self.datastore,
+                &group_id,
+                &live_ids,
+            );
 
             let context_count =
                 MetadataRepository::new(&self.datastore).count_contexts(&group_id)? as u64;
--- a/crates/context/src/handlers/get_group_info.rs
+++ b/crates/context/src/handlers/get_group_info.rs
@@ -25,7 +25,11 @@ impl Handler<GetGroupInfoRequest> for ContextManager {
             let Some((node_identity, _)) = self.node_namespace_identity(&group_id) else {
                 bail!("node has no group identity configured");
             };
-            if !MembershipRepository::new(&self.datastore).is_member(&group_id, &node_identity)? {
+            if !crate::scope_projection::ScopeProjections::member_now_checked(
+                &self.datastore,
+                &group_id,
+                &node_identity,
+            )? {
                 bail!("node is not a member of group '{group_id:?}'");
             }
 
@@ -46,6 +50,26 @@ impl Handler<GetGroupInfoRequest> for ContextManager {
                 + MembershipRepository::new(&self.datastore)
                     .enumerate_inherited(&group_id)?
                     .len()) as u64;
+
+            // SHADOW: validate the projection's effective-member count against the
+            // live union (logs `membership-enum` divergence; still returns live).
+            if let Some(projected) =
+                crate::scope_projection::ScopeProjections::member_identities_now_ephemeral(
+                    &self.datastore,
+                    &group_id,
+                )
+            {
+                if projected.len() as u64 != member_count {
+                    tracing::warn!(
+                        marker = "unified_projection_divergence",
+                        plane = "membership-enum",
+                        group_id = ?group_id,
+                        projection_count = projected.len(),
+                        live_count = member_count,
+                        "query-enum: projection effective-member count differs from live"
+                    );
+                }
+            }
 
             let context_count =
                 MetadataRepository::new(&self.datastore).count_contexts(&group_id)? as u64;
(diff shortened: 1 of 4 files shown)


chefsale:
Addressed (commit `4a0165f4`):

- **Materialized fallback** (Bugbot, high): `member_identities_in_view` now adds a wholly-unfolded group's live direct rows, mirroring `member_at_cut`'s `role_of` fallback — a Restricted subgroup whose members arrived as materialized rows (or via ops this node can't decrypt) no longer reads as a spurious subset of live's `list`.
- **Set, not count** (Bugbot, medium): `get_group_info` now compares the full live identity set via `shadow_check_member_enum` instead of `projected.len()`, so equal-count/different-member drift is caught.
- **Single subtree fold / mixed cuts** (Bugbot, medium): `get_migration_status` uses the new `member_identities_subtree_ephemeral`, folding the namespace **once at one cut** for the whole cohort — fixes the mixed-cut synthetic mismatch under concurrent governance and drops the cost from O(N×DAG) to O(DAG+N).
- **Silent `None` on store fault**: now warns (in the shared `ephemeral_view`), matching `member_now_ephemeral`.
- **Candidate universe**: confirmed complete w.r.t. the fold — `is_member_at_cut` only accepts identities reachable from `view.groups ∪ group_admin ∪ root_admin` (direct, admin, or inherited-from-an-ancestor-group); the only members it can miss are materialized-only ones, now covered by the fallback above.

Refactor: extracted `ephemeral_view` (fold + at-cut view + the warn) and the view-taking `member_identities_in_view`, so the single-group and subtree paths share one implementation.

Deferred (acknowledged): `get_group_info`/`list_group_members` still fold twice per request — once for the boolean gate (`member_now_checked`) and once for the enum shadow. Sharing one fold would mean threading the built view through `member_now_checked`; it's a transient cost on low-frequency read endpoints and goes away when the shadow is removed at the F5 flip. Left as a follow-up rather than reshaping the gate signature here.

chefsale:
Addressed (commit `bb6514b2`) — eliminated the double fold:

`list_group_members` and `get_group_info` now fold the namespace governance DAG **once per request** via a shared `ScopeProjections::ephemeral_projection` (the expensive `collect_namespace_ops` RocksDB walk), and derive **both** the boolean gate (`member_now_checked_with`) and the enum shadow (`shadow_member_enum_with`) from that one fold — no more gate-fold + shadow-fold.

For context on the rest of the round:
- **N-fold cohort / mixed cuts** (Bugbot): already fixed in `4a0165f4` — `get_migration_status` uses `member_identities_subtree_ephemeral`, which folds the namespace **once at one cut** for the whole subtree (those comments were posted against the older `5c5c187e`).
- The standalone `member_now_checked` / `shadow_check_member_enum` stay for the gate-only endpoints (`list_all_groups`, etc.) and single-call sites.

Net: every enumeration handler now does exactly one governance fold per request.

chefsale:
Latest round (commit `42687ba7`):

- **Standalone double-fold helpers**: removed `shadow_check_member_enum` and `member_identities_now_ephemeral` outright — after the shared-fold refactor (`bb6514b2`) they had no callers (the dual handlers use `shadow_member_enum_with` over the request's single `ephemeral_projection`). So that path no longer folds at all.
- **`enumerate_inherited` called twice in `get_group_info`**: now resolved once and reused for the count and the live id set.
- **Materialized fallback skips cascade/deny filters**: documented why this is intentional — those rows are live's current `list`, already cascade- and removal-consistent; re-filtering them through this node's *incomplete* fold (the reason we fell back) would wrongly drop valid members.
- **Root-membership filter for `group == root_group`**: documented that it's a deliberate no-op there — the first `is_member_at_cut(*group, …)` already decides root membership and the root's `MemberLeft` is folded, so there's no un-cascaded stale row.

On the **sampling-gate** suggestion: declining intentionally — the shadow must run on **every** request for the e2e divergence gate to catch drift (a sampled-out request that diverged would slip the gate). It's a transient cross-check that's deleted at the F5 flip, when these endpoints read the projection directly (one fold, no shadow). Per-request cost is now exactly one governance fold on these low-frequency read endpoints.

Stale (already fixed, will clear when Bugbot re-runs off the old `5c5c187e`): the N-fold cohort and mixed-cut comments — `get_migration_status` folds once at one cut via `member_identities_subtree_ephemeral`.

chefsale:
Latest round (commit `a5196a61`) — went through all of them:

- **`ephemeral_projection` / `ephemeral_view` duplicate the fold**: extracted a shared `ephemeral_fold` (collect + backfill + heads); both call it now.
- **Materialized fallback under-includes inherited members**: a wholly-unfolded group now defers to live's **full** `list ∪ enumerate_inherited` (was direct rows only), so it can't miss the inherited side the unfolded structure can't derive.
- **Dangling doc-comment**: `collect_namespace_ops`'s doc had been orphaned onto `namespace_current_heads` (inserted between the doc and its fn earlier) — moved it back to the function it documents.
- **Candidate universe completeness** (raised a few times): documented the proof — `is_member_at_cut` only accepts direct members (in `view.groups.values()`), folded group/subgroup admins (in `view.group_admin.values()`, plus Admin-role holders already counted as direct), or the genesis root admin, so nothing it accepts lies outside the candidate set.
- **`acl_view_at` returns `None` silently**: documented that this is the empty-never-fed-namespace case (nothing to compare → skip, not a fault).

Already addressed / not bugs:
- `deny.is_denied` per candidate: the `DenyListRepository` is constructed **once** (`let deny = …`) and reused across candidates; the per-candidate call is the unavoidable lookup.
- `inherited` "double-consumption" in `get_group_info`: `inherited.len()` borrows, then `into_iter()` moves — valid and correct (count and id-set come from the same single walk, and `list`/`enumerate_inherited` are disjoint so there's no double-count).
- N-fold cohort / mixed cuts: fixed back in `4a0165f4` (`member_identities_subtree_ephemeral` folds once at one cut) — those are stale re-anchors of comments posted against `5c5c187e`.

Issue #2770, opened 2026-06-16
governance-store: OpEvents emitted before op-log entry is persisted (subscriber read-back race)


xilosada:
## Problem

In the group-op apply path, `notify_op_event(...)` fires **inside the per-op handlers** (e.g. `OpEvent::TeeMemberAdmitted` at `crates/governance-store/src/ops/group/member_joined_via_tee_attestation.rs:43`) **before** the caller persists the op-log entry (`crates/governance-store/src/lib.rs:~1340` / `crates/governance-store/src/namespace/governance.rs:~1365`). There is no transaction wrapping the apply — writes are write-through (`crates/store/src/layer.rs:93-95`), and `notify` dispatches to subscribers on other tasks.

Result: a subscriber that reacts to an event and immediately reads the **op-log** (e.g. `tee_admission_record`, `is_tee_admitted_identity`, `read_tee_admission_policy`, `read_op_log_after`) can observe stale state (entry not yet written). Membership rows are written before the event, so `has_direct_member`/`list` are unaffected — only op-log-scanning reads race.

This currently affects `calimero-context::tee_subgroup_admit` (`handle_new_tee_member` reads back the verdict via `tee_admission_record` on `TeeMemberAdmitted`). It is mitigated there by a bounded wake-then-reread (commit e2b12885), but the underlying emit-before-persist ordering is latent for any future op-log-reading subscriber, and is already noted as a deferred "single atomic batch" item at `crates/governance-store/src/local_state.rs:223-226`.

## Proposed durable fix (Option A)

Collect events during apply and flush them **after** the op-log entry persists (ideally once the apply uses a single atomic `StoreBatch`/`Store::apply`). Add a `pending_events: Vec<OpEvent>` to the apply ctx (mirroring the existing `divergence` field), have per-op handlers queue instead of `notify`, and drain after persist. Gate the flush on the op-log-append path so idempotent replays/dedup don't re-emit. All current subscribers (`self_purge`, `auto_follow`, `tee_subgroup_admit`) are idempotent, so emit-on-append-only is the safer, more correct behavior — but verify against the e2e suite.

Blast radius: ~8 files in `calimero-governance-store` (per-op handlers + two flush sites). `GroupKeyDelivered` (join_group wake-then-reread) must be left as-is. No public-API change.

## Context
Found during Phase 1 of transparent per-subgroup TEE admission. The wake-then-reread in `tee_subgroup_admit` makes that feature correct meanwhile; this issue tracks the root-cause fix.

PR #2845, opened 2026-06-20
fix(namespaces): admit root TEE into Restricted subgroups created via REST (mint keys before GroupCreated)

xilosada:
## Problem

A namespace's root-admitted `ReadOnlyTee` member was **not** admitted into Restricted subgroups created via the REST path (`POST /admin/.../namespaces/:id/groups`) — the path the desktop/tauri client and mdma drive. New private channels silently shipped without the TEE replica.

## Root cause

`create_group_in_namespace` minted the subgroup's signing key and group key **after** `sign_apply_and_publish_namespace_op` returned. Since emit-after-persist (#2770), that call drains `OpEvent::SubgroupCreated` *once the root op is persisted* — i.e. before this handler reached the key write. The `tee_subgroup_admit` subscriber reacts to that event by reading the new subgroup's group key (`GroupKeyring::load_current_key`) to decide whether it is the key-holder that should admit the namespace's root TEE members. It raced ahead of the write, observed `None`, logged `tee-subgroup-admit: skip — not the subgroup key-holder`, and skipped.

The actor handler (`crates/context/src/handlers/create_group.rs`) mints the key *before* applying, which is why creating/joining a channel from a local node worked while the REST path did not.

## Fix

Mint the subgroup signing key and group key **before** `sign_apply_and_publish_namespace_op`, mirroring the actor handler, so the key is visible the instant `SubgroupCreated` fires. A failed key write now fails the request instead of being a warn-only skip (a second, pre-existing gap on this handler).

Also adds decision logging at each `handle_new_subgroup` early-return so any future "TEE didn't join subgroup X" is self-diagnosing (one `info!` line per skip/admit reason).

## Verification

Reproduced and fixed on a real node (desktop owner + Phala TEE replica). Before: `skip — not the subgroup key-holder`. After:

```
tee-subgroup-admit: new-subgroup trigger received   subgroup=4b360e54… namespace=8d41cc4e…
tee-subgroup-admit: admitting root TEE into new Restricted subgroup   member=F7yj5…
TEE node admitted via attestation   group_id=4b360e54…
published KeyDelivery for admitted TEE node   group_id=4b360e54…
```

Disable-HA cascade subsequently removed the TEE from this subgroup too, confirming the full lifecycle.

## Scope

- No published-surface change (`calimero-server-primitives` / `calimero-tee-attestation` untouched); no mero-tee rev bump needed.

---

> [!NOTE]
> **Medium Risk**
> Touches namespace subgroup creation and TEE auto-admission timing; behavior change is intentional and aligned with the actor `create_group` handler, with no public API changes.
> 
> **Overview**
> **Fixes root TEE replicas missing from Restricted subgroups** created through the admin REST `create_group_in_namespace` path. The handler now **writes the subgroup signing key and group encryption key before** `sign_apply_and_publish_namespace_op`, so when `SubgroupCreated` fires (post-persist), `tee_subgroup_admit` sees the key-holder and can admit namespace `ReadOnlyTee` members. Previously keys were minted only after publish returned, racing the subscriber and causing a silent skip.
> 
> Key persistence failures on that path are now **request failures** instead of warn-only success responses. **`tee_subgroup_admit`** gains **debug/info/warn** lines at each early-return in `handle_new_subgroup` to make “TEE didn’t join subgroup X” diagnosable from logs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 730bce3a6db050065c08cd136ef59f120de3ea9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/tee_subgroup_admit.rs
+++ b/crates/context/src/tee_subgroup_admit.rs
@@ -239,12 +239,21 @@ async fn handle_new_subgroup(
     // a row, so `is_open_chain_to_namespace` fails safe (treats a not-yet-written
     // subgroup as Restricted → we proceed) rather than wrongly skipping it.
 
+    info!(
+        subgroup = %hex::encode(child_group_id),
+        namespace = %hex::encode(namespace_id),
+        "tee-subgroup-admit: new-subgroup trigger received"
+    );
+
     // Only act for Restricted subgroups — Open subgroups are already readable
     // by a root-admitted TEE node (inherited membership + namespace key).
     match CapabilitiesRepository::new(store).is_open_chain_to_namespace(&child_gid, &namespace_gid)
     {
-        Ok(true) => return, // Open → skip
-        Ok(false) => {}     // Restricted → proceed
+        Ok(true) => {
+            info!(subgroup = %hex::encode(child_group_id), "tee-subgroup-admit: skip — Open subgroup (TEE reads via inheritance)");
+            return;
+        }
+        Ok(false) => {} // Restricted → proceed
         Err(e) => {
             error!(?e, "tee-subgroup-admit: open-chain check failed");
             return;
@@ -253,8 +262,11 @@ async fn handle_new_subgroup(
 
     // Only the key-holder (the creator) can deliver the per-subgroup key.
     match GroupKeyring::new(store, child_gid).load_current_key() {
-        Ok(Some(_)) => {}   // we hold the key → we can admit + deliver
-        Ok(None) => return, // not the key-holder → leave it to the creator / pull
+        Ok(Some(_)) => {} // we hold the key → we can admit + deliver
+        Ok(None) => {
+            info!(subgroup = %hex::encode(child_group_id), "tee-subgroup-admit: skip — not the subgroup key-holder (leave to creator / key pull)");
+            return;
+        }
         Err(e) => {
             error!(?e, "tee-subgroup-admit: load_current_key failed");
             return;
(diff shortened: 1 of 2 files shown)


xilosada:
### Review pass — maintainer-assisted (automated)

I reviewed the two existing commits for correctness and quality:

**`fix(namespaces): mint subgroup keys before applying GroupCreated op`** — ✅ correct.
- `group_id` is freshly random per request, so minting the signing key + group key *before* the op apply carries no orphan/collision risk; on a failed apply the unused keys are unreachable.
- Making a key-store failure fatal (return error) is stricter and better than the old post-apply `warn!`-and-continue, which could leave a "created" group with no signing/group key.
- It correctly mirrors the actor path (`crates/context/src/handlers/create_group.rs`) and closes the race where `tee_subgroup_admit` observed `load_current_key() == None` and wrongly skipped TEE admission for Restricted subgroups created via REST.

**`diag(tee-subgroup-admit): log decision at each early-return`** — quality nit, addressed in a follow-up commit.
- The diagnostic logging logged every early-return at `info!`. "new-subgroup trigger received" fires on *every* subgroup creation regardless of TEE involvement, and the skip decisions duplicate state the sibling `handle_new_tee_member` leaves silent → net info-level log spam in production.
- Pushed `refactor(tee-subgroup-admit): downgrade routine decision traces to debug` (`730bce3a`): routine traces → `debug!` (still available under `RUST_LOG=debug`), keeping the genuine action ("admitting root TEE…") at `info!` and the anomaly ("member with no admission verdict") at `warn!`.

Build is clean (no warnings) and the `tee_subgroup_admit` unit test passes locally.

PR #2850, opened 2026-06-23
feat(context): shadow apply-time governance-op authorization (F5 #28)

chefsale:
## What

Begins migrating the **last** F5 consumer — apply-time governance-op authorization — onto the unified projection, as a **log-only shadow** (the live apply gates stay authoritative). Continues the F5 series (#2829/#2840/#2841/#2843 merged).

## The crate-boundary constraint

The apply gates (`PermissionChecker`, `require_namespace_admin`) live in `calimero-governance-store`, which the projection (`calimero-context`) **depends on** — so governance-store can't reach the projection. The shadow therefore lives at the **context layer** (`apply_signed_namespace_op.rs`), alongside the existing membership shadow, which already holds the op's parents and the projection. No governance-store change.

## How

For a just-applied op (live authorized its signer), `apply_auth_requirement` maps the op variant to the signer-authority the live gate enforced, and the shadow asks whether the projection authorizes the **signer** at the op's **parent cut** — the state *excluding* this op, the correct causal cut to authorize against (the membership shadow uses the op's own id, right for a post-apply read, wrong for auth).

- `Some(false)` (projection would reject what live accepted) → `unified_projection_divergence` plane=`governance-auth`.
- `None` (cited ancestry not fully folded) → skip.
- Reverse direction is unobservable here (a live-rejected op never reaches this fold).

Adds at-cut accessors `is_admin_at_cut` + `is_admin_or_capability_at_cut`, mirroring `member_at_cut_authoritative` (resolve namespace, gate on `cut_ancestry_complete`, read genesis root admin / default cap as immutable base facts, decide from the folded `AclView`).

**Conservative coverage:** the unambiguous admin/cap gates (namespace-root admin ops, `GroupCreated`, `MemberAdded`/`Removed`, role/cap/visibility sets). Skips variants whose authority isn't the signer's group-admin/cap — an admin-signed invitation (`MemberJoined`), an inheritance proof (`MemberJoinedOpen`), a self-authored `MemberLeft`.

## The flip is a separate decision

This shadow validates equivalence; making the projection **authoritative** at the gates requires it inside governance-store (the crate boundary) — either moving the auth gates up to the context layer or dependency-inversion via a trait. That's a follow-up design, informed by this shadow's divergence data; shadow-only may even be the pragmatic endpoint for #28 (the apply gates are the deepest, most convergence-critical code and the live state is itself derived from the same ops).

## Test

- `cargo build`/`clippy --all-targets`/`fmt` clean; equivalence harness 3/3 green.
- e2e `governance-auth` gate validates the apply-auth equivalence on real traffic.

---

> [!NOTE]
> **Low Risk**
> Shadow-only logging on the apply path; live authorization is unchanged. Divergence warnings are observability, not enforcement.
> 
> **Overview**
> Adds a **log-only apply-auth shadow** (F5 #28) next to the existing membership shadow in `apply_signed_namespace_op`: after a governance op is applied, it checks whether the unified projection would have authorized the **signer** at the op’s **parent cut** (`delta_parents`), not the post-apply cut.
> 
> **`calimero-authz`:** New `AclView::is_authorized_admin` mirrors live apply-gate admin authority—group admin, folded root admin (plus genesis root carve-out), or admin of an open-chain ancestor; restricted edges stop the walk. Plain inherited members are not treated as admins.
> 
> **`scope_projection`:** New `is_admin_at_cut`, `is_admin_or_capability_at_cut`, and shared `auth_cut_context` (complete ancestry gate, folded view, genesis root, default cap base).
> 
> **Apply handler:** `apply_auth_requirement` maps applied ops to `Admin` vs `AdminOrCap` requirements for a conservative subset (root admin ops, subgroup create, member add/remove, visibility, role/cap defaults). `Some(false)` logs `unified_projection_divergence` with `plane=governance-auth`. Live gates stay authoritative; no governance-store changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 367d257ecbf7a5b5afd4f07165d63fba4990eff6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/authz/src/lib.rs
+++ b/crates/authz/src/lib.rs
@@ -222,6 +222,49 @@ impl AclView {
         anchor_is_member.unwrap_or(false)
     }
 
+    /// Is `author` authorized as an ADMIN of `group` at the cut — the apply-gate
+    /// admin authority. Mirrors live's `is_authorized_with_capability` admin path:
+    /// a direct group admin (subgroup creator / `Admin`-role holder), the
+    /// namespace ROOT admin (who administers every group — folded, so it tracks
+    /// `AdminChanged`, with the genesis `root` carve-out as the un-folded base), OR
+    /// an admin of an ANCESTOR reached over the open-subgroup chain (an admin of an
+    /// Open parent administers its children). Restricted edges stop the walk.
+    ///
+    /// Admin-only — unlike [`is_member_at_cut`](Self::is_member_at_cut), a plain
+    /// inherited MEMBER is not authorized. Capability holders are checked
+    /// separately by the caller.
+    #[must_use]
+    pub fn is_authorized_admin(
+        &self,
+        group: ContextGroupId,
+        author: &PublicKey,
+        root: Option<(ContextGroupId, PublicKey)>,
+    ) -> bool {
+        let is_admin = |g: ContextGroupId| -> bool {
+            self.is_group_admin(author, g)
+                || self.is_root_admin(author)
+                || root.is_some_and(|(root_g, root_admin)| g == root_g && *author == root_admin)
+        };
+        if is_admin(group) {
+            return true;
+        }
+        let mut current = group;
+        for _ in 0..=MAX_NAMESPACE_DEPTH {
+            let Some(edge) = self.subgroups.get(&ScopeId::from(current.to_bytes())) else {
+                return false;
+            };
+            if edge.restricted {
+                return false;
+            }
+            let parent = ContextGroupId::from(*edge.parent.as_bytes());
+            if is_admin(parent) {
+                return true;
+            }
+            current = parent;
+        }
+        false
+    }
+
     /// `member`'s effective capability bitmask in `group` at the cut: the
     /// explicit per-member override if present, else the group default, else
     /// `0`. Mirrors the live `member_capability` read used by inherited-
--- a/crates/context/src/handlers/apply_signed_namespace_op.rs
+++ b/crates/context/src/handlers/apply_signed_namespace_op.rs
@@ -117,49 +117,6 @@ impl Handler<ApplySignedNamespaceOpRequest> for ContextManager {
                                         &[shadow_op.id],
                                     )
                                 });
-
-                                // APPLY-AUTH shadow (F5 #28, log-only): the op just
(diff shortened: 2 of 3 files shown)


chefsale:
Rebased onto latest master and addressed the review (commit `d120e39e`). Note: the pre-fix e2e was already **green (0 `governance-auth` divergences)** — the scenarios use direct admins and don't hit these edges — so these are latent-divergence hardening rather than live bugs.

- **Missing inherited admin walk / GroupCreated wrong gate**: added `AclView::is_authorized_admin` — an admin-only walk up the open-subgroup chain (mirrors live `is_authorized_with_capability`'s admin path), including the namespace **root** admin who administers every group. The accessors now use it, so an inherited/root admin's op is no longer false-flagged. This also covers `GroupCreated` by a root admin at any parent depth.
- **Root admin = current-state read, not at-cut**: `is_authorized_admin` consults the **folded** `is_root_admin` (tracks `AdminChanged`), with the genesis meta as the un-folded base.
- **TransferOwnership / GroupDeleted mapped wrong**: both **dropped** from the shadow — `TransferOwnership` gates on `meta.owner_identity` and `GroupDeleted` on the subgroup owner / `CAN_DELETE_SUBGROUP`, neither of which is the signer's group admin/cap. Skipping avoids the false flag.
- **`feed_store` borrowed under the write lock**: the auth check resolves at the **parent** cut (independent of the just-ingested op), so it now runs **outside** the write lock under a brief read lock — no store I/O while ingest is blocked.
- **Duplicated setup**: extracted `auth_cut_context` (namespace resolve + ancestry gate + folded view + root tuple + default cap), shared by both accessors.

On the **explicit-zero capability** nit (`folded != 0 ? folded : default`): kept, intentionally — it mirrors `is_member_at_cut`'s `effective_cap` exactly (live `member_capability` = override.or(default)); an explicit all-bits-cleared override is an edge the membership path also treats this way, so matching it keeps the two consistent.

`cargo build`/`clippy --all-targets`/`fmt` clean; authz 10/10, equivalence harness 3/3. Re-validating e2e.

PR #2852, opened 2026-06-23
feat(context): flip identity-set enum consumers onto the projection

chefsale:
## What

F5 enum flip (identity-set subset), continuing after #2851 (#29a). The enum shadow validated that the projection's effective-member **set** matches live's `list ∪ enumerate_inherited` across the e2e `membership-enum` plane. This flips the two consumers that need only the identity set off live onto the projection:

- **`get_group_info` member_count** → the projection set's `len()` (one shared fold; `None` on an empty/unfed namespace falls back to live `count + enumerate_inherited`).
- **`get_migration_status` cohort** → `member_identities_subtree_ephemeral` (folded once at one cut for the whole subtree; `None` falls back to the live union).

Both drop their `membership-enum` shadow comparison — the projection is now the answer, not a cross-check. The live fallbacks retire in #29b.

## Not flipped (on purpose)

**`list_group_members`** returns per-member **roles**, and the shadow validated only the identity **set**, not roles. Flipping it needs a role-bearing enumeration (`member_entries_in_view`) plus its own role-divergence validation pass — a follow-up. It keeps its live + shadow for now.

Adds `member_identities_with` (the set read off an already-built fold).

## Test

- `cargo build`/`clippy --all-targets -- -D warnings`/`fmt` clean; equivalence harness 3/3 green.
- e2e: the two `membership-enum` consumers now act on the projection; `list_group_members`'s `membership-enum` shadow still gates.

---

> [!NOTE]
> **Medium Risk**
> Changes observable counts and migration cohort membership for admin APIs; incomplete-fold guards and live fallback limit wrong answers, but behavior shifts from always-live to projection-first.
> 
> **Overview**
> After the `membership-enum` shadow validated projection vs live, **identity-set enumeration** now uses the governance projection as the primary answer instead of live store unions plus shadow comparison.
> 
> **`get_group_info`** derives `member_count` from `member_identities_with` on the shared ephemeral fold (set length), reusing the same fold as the membership gate. **`collect_migration_cohort`** returns early from `member_identities_subtree_ephemeral` when the projection succeeds, instead of building the live cohort first and only comparing afterward.
> 
> **`scope_projection`** adds `member_identities_with` (authoritative set off an existing fold) and tightens `member_identities_subtree_ephemeral` with `cut_ancestry_complete` before reading the ACL view so partial folds cannot under-count. **`ephemeral_view`** is removed in favor of `ephemeral_fold` + completeness checks.
> 
> When projection returns `None` (unfed namespace, store fault, or incomplete ancestry during governance backfill), both handlers still fall back to live `count`/`list` ∪ `enumerate_inherited`; those paths are slated to retire in #29b. **`list_group_members`** is unchanged—still live + shadow—because it needs roles, not identity sets only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ac16b527d7cf0a83b18e9efbbe747f077459986. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/get_group_info.rs
+++ b/crates/context/src/handlers/get_group_info.rs
@@ -58,9 +58,10 @@ impl Handler<GetGroupInfoRequest> for ContextManager {
             // scans the full group state on every call.
             // Effective member count from the PROJECTION's effective-member set
             // (direct ∪ inherited), validated divergence-free across the e2e
-            // `membership-enum` plane. `None` (empty/unfed namespace or store fault)
-            // falls back to the live `count + enumerate_inherited` union. The live
-            // fallback retires in #29b.
+            // `membership-enum` plane. `None` — an empty/unfed namespace, or a cited
+            // ancestry not fully folded (governance-backfill race) — falls back to
+            // the live `count + enumerate_inherited` union. The live fallback retires
+            // in #29b.
             let member_count = proj_ctx
                 .as_ref()
                 .and_then(|(proj, ns, heads)| {
--- a/crates/context/src/handlers/get_group_info.rs
+++ b/crates/context/src/handlers/get_group_info.rs
@@ -56,28 +56,26 @@ impl Handler<GetGroupInfoRequest> for ContextManager {
             // chain walk bounded by `MAX_NAMESPACE_DEPTH`. This is minor
             // next to `compute_group_state_hash` below, which already
             // scans the full group state on every call.
-            // Resolve the open-subgroup inherited members ONCE — reused for the
-            // count and (below) the shadow's live id set, instead of two parent
-            // walks.
-            let inherited =
-                MembershipRepository::new(&self.datastore).enumerate_inherited(&group_id)?;
-            let member_count = (MembershipRepository::new(&self.datastore).count(&group_id)?
-                + inherited.len()) as u64;
-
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed (commit `5ac16b52`):

- **Enum skips ancestry completeness** (Bugbot, medium): `member_identities_with` and `member_identities_subtree_ephemeral` now gate on `cut_ancestry_complete` and return `None` (→ live fallback) on a partial fold — the same defer-on-incomplete guard the membership gate's `member_at_cut` uses. Without it, a governance-backfill race could return an under-counted set/cohort. Removed the now-dead `ephemeral_view` (the subtree path uses `ephemeral_fold` directly so it can read the heads for the check).
- **`None`-comment accuracy**: corrected to "empty/unfed namespace **or** incomplete fold (backfill race)" — `member_identities_with` returns `None` only for those, not a generic store fault (the in-view deny/cap reads are swallowed to `false`/`0`, not propagated).

Pre-fix e2e was already green (0 `membership-enum` divergences); these are correctness hardening for the backfill-race edge. `clippy --all-targets -- -D warnings`/`fmt` clean, equivalence 3/3.

PR #2853, opened 2026-06-23
feat(context): shadow member-enumeration roles (precursor to list_group_members flip)

chefsale:
## What

`list_group_members` is the last enum consumer on live. Unlike the count/cohort consumers (flipped in #2852), it returns **roles** — which the identity-set shadow never validated. This **validates the projection's enumeration roles** against live, log-only, as the precursor to flipping it.

## How

- `AclView::member_path_at_cut` — the at-cut analogue of live `check_path`, returning a role-bearing `MemberPathAtCut` (`Direct{role}` / `Inherited{anchor, via_admin}` / `None`) by mirroring the `is_member_at_cut` walk.
- `ScopeProjections::member_role_at_cut_with` — derives the enumeration role exactly as live's `list ∪ enumerate_inherited`: a direct member's folded role, `Admin` for inherited-via-admin, else the member's role at the anchor (matching live's #2846 `via_admin ? Admin : role_of(anchor)`). Gated on `cut_ancestry_complete` (abstain → skip).
- `list_group_members` shadows each live member's role vs the projection's at-cut role, logging `unified_projection_divergence` plane=`membership-role` on a mismatch. **Still returns the live list.**

## Why a shadow first

A wrong role would be a client-facing bug, so I'm validating roles on real e2e traffic before flipping (same shadow→prove→flip discipline as the rest of F5). The flip lands once the `membership-role` plane is divergence-free.

## Test

- `cargo build`/`clippy --all-targets -- -D warnings`/`fmt` clean; authz 10/10, equivalence 3/3.
- e2e: the new `membership-role` plane gates role equivalence across the scenarios.

---

> [!NOTE]
> **Medium Risk**
> Touches group membership role resolution and enumeration parity with live; behavior is shadow-only today but wrong logic would surface on the eventual list_group_members flip.
> 
> **Overview**
> Adds **log-only validation** that projected **member roles** match live for `list_group_members`, ahead of serving that API from the causal projection (the existing identity shadow did not cover roles).
> 
> **`calimero-authz`:** Introduces `MemberPathAtCut` and `AclView::member_path_at_cut`, mirroring the `is_member_at_cut` walk but returning direct role, inherited-via-admin, or inherited-from-anchor (with stored-row precedence over genesis admin).
> 
> **`scope_projection`:** Adds `member_roles_for` to resolve enumeration roles in one fold (direct / `Admin` when inherited via admin / anchor role otherwise), abstaining when ancestry is incomplete or the group has no folded direct membership (same materialized-fallback parity as the identity enum).
> 
> **`list_group_members`:** After the membership-enum shadow, compares each live member’s role to the projection map and logs `unified_projection_divergence` with plane `membership-role` on mismatch; the handler **still returns the live list**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bb4720fc63830000d2f2ce20835c23d6ed1e0bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/authz/src/lib.rs
+++ b/crates/authz/src/lib.rs
@@ -31,6 +31,25 @@ const CAN_JOIN_OPEN_SUBGROUPS: u32 = MemberCapabilities::CAN_JOIN_OPEN_SUBGROUPS
 /// `MAX_NAMESPACE_DEPTH`).
 const MAX_NAMESPACE_DEPTH: usize = 16;
 
+/// How `author` reaches membership of a group at a cut — the at-cut analogue of
+/// the live `MembershipPath`. Carries enough to derive the enumeration ROLE
+/// (mirrors live's `list ∪ enumerate_inherited`): a direct member keeps its
+/// folded role; an inherited member is `Admin` when reached via an admin, else
+/// its role at the `anchor` it inherits from.
+#[derive(Clone, Debug, PartialEq)]
+pub enum MemberPathAtCut {
+    /// Not a member of the group at this cut.
+    None,
+    /// A direct member, with its folded role.
+    Direct { role: GroupMemberRole },
+    /// Inherited over the open-subgroup chain from `anchor`; `via_admin` when the
+    /// path was through an admin ancestor (role resolves to `Admin`).
+    Inherited {
+        anchor: ContextGroupId,
+        via_admin: bool,
+    },
+}
+
 /// Why an op was refused. One rejection type for every plane — the caller
 /// doesn't have to know which plane said no.
 #[derive(Clone, Debug, PartialEq, Eq, ThisError)]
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed (3141bcab):

- **Root admin skews inherited roles (Medium) / restricted-subgroup divergence** — `member_path_at_cut` was using the **global** `is_root_admin`. Narrowed its `is_admin` to the *identical* closure `is_member_at_cut` uses (`is_group_admin || genesis-root-admin-of-this-group`, no global predicate). The root admin now reaches a subgroup only over the open chain (the walk), so on a **Restricted** subgroup the role path agrees with the validated membership set (excluded), instead of marking them a `Direct{Admin}` everywhere.
- **Walk continues past a non-qualifying anchor** — confirmed already correct: `anchor_decision.is_none()` only gates *re-setting* the anchor; the loop continues up and an admin ancestor higher in the open chain still returns `Inherited{via_admin:true}`. Structurally identical to `is_member_at_cut`.
- **Per-member re-fold (nit)** — replaced `member_role_at_cut_with` with `member_roles_for`, which folds the view **once** and resolves the whole live-member batch.
- **`Inherited{via_admin:false}` → `Member` fallback** — kept, now documented as a defensive, unreachable-by-construction mirror of live's `role_of(anchor).unwrap_or(Member)` (the arm is only emitted when the anchor row is present).

clippy `-D warnings` / fmt / authz 10/10 / equivalence 3/3 green. The prior run was **0 divergences on the `membership-role` plane**; re-dispatching e2e to confirm the fixes hold.

chefsale:
Second round addressed (no behavior change — clarifications + an assert):

- **Direct-row-before-admin order** — deliberate, now documented. `is_member_at_cut` returns a bool so its `admin || row` order is immaterial; `member_path_at_cut` must pick a *role*, and when an identity is both a stored member and the genesis admin, live's `list` returns the **stored row's** role. So the row is authoritative; the admin carve-out only supplies `Admin` when there's no row.
- **`member_roles_for` returns `None` for non-members** — intended. Member **presence** is validated by the identity shadow directly above (`membership-enum`, 0 divergences). The role shadow only compares **roles** for members both sides agree exist; a member the projection no longer sees is already that plane's divergence, so skipping `None` here isn't a silent gap. Documented at the call site.
- **`zip` truncation** — added `debug_assert_eq!(members.len(), projected_roles.len())`.
- **`.clone()` on `GroupMemberRole`** — it's `#[derive(Clone, ...)]`, **not** `Copy` (that's why `MemberPathAtCut` can't be `Copy` either), so the clone is required.

The narrowed-admin run came back **0 divergences on `membership-role`** (and `membership-enum`). Re-dispatching once more to cover this push.

chefsale:
Third round addressed (3rd push):

- **Medium — unfolded-group role mismatch.** Real gap: `member_roles_for` didn't mirror the identity shadow's **materialized fallback**. When a group's direct membership isn't folded into `view.groups` (a Restricted subgroup arriving as materialized rows, or undecryptable member ops), the fold has no opinion on its roles, so the walk would derive an `Inherited`/`Admin` role that differs from live's stored direct row. Now `member_roles_for` **abstains wholesale** on an unfolded group — the exact `!view.groups.contains_key(group)` gate `member_identities_in_view` already uses. No spurious `membership-role` divergence on lagging/undecryptable groups.
- **`debug_assert_eq` is a no-op in release (both bots).** Removed entirely by switching `member_roles_for` to a `BTreeMap` keyed by member; the shadow now looks each live member up **by key**, so there's no positional `zip` and no length coupling — a wrong-length return is structurally impossible.
- **Walk doesn't continue past a non-qualifying anchor / `Member` fallback unreachable** — these re-anchor the round-1/2 findings: the loop *does* continue (the `is_none()` guard only blocks *re-setting* the anchor, not the walk), structurally identical to `is_member_at_cut`; the `Member` fallback is the documented defensive mirror of live's `role_of(anchor).unwrap_or(Member)`, unreachable because the arm only fires when the anchor row is present.

clippy `-D warnings` / fmt / equivalence 3/3 green. Re-dispatching e2e; prior two runs were 0 divergences on `membership-role` + `membership-enum`.

PR #2854, opened 2026-06-23
feat(context): flip list_group_members onto the projection enumeration

chefsale:
## What

`list_group_members` is the **last membership-enum consumer reading the live resolver**. Flip it onto the projection.

This is the payoff of #2853 (role shadow) + #2852 (identity-set flip): the projection's at-cut effective-member enumeration is validated divergence-free on both the `membership-enum` (identities) and `membership-role` (roles) planes — the role shadow ran **0 divergences across three e2e runs**.

## How

- **`member_entries_with`** — the full effective-member enumeration **with roles**. Identity set is the already-validated `member_identities_with` set; each role is resolved by the same `member_path_at_cut` the role shadow proved equal to live. `None` (defer to live) when the cited ancestry isn't fully folded, or the target group's direct membership isn't folded — the materialized-fallback case live owns.
- **`list_group_members`** builds its member list from `member_entries_with`, falling back to the live `list ∪ enumerate_inherited` union on `None`. The live fallback retires in #29b.
- **Removed dead shadow scaffolding** for this consumer — `member_roles_for` (role shadow) and `shadow_member_enum_with` (identity shadow); this was their last caller. Doc-refs repointed.

## Safety

Behavior-preserving: the shadow proved the projection enumeration (identities + roles) equals live across the e2e scenarios, and the `None` path still defers to live for un-folded cuts. No new wire, no new shadow.

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean; projection equivalence 3/3.
- e2e: behavioral (no shadow markers left); the scenarios exercising `list_group_members` now read the projection.

---

> [!NOTE]
> **Medium Risk**
> Changes authoritative membership listing for a governance API; behavior is intended to match live via prior shadow validation and explicit defer-to-live paths, but incorrect projection enumeration could misreport members or roles until #29b removes the fallback.
> 
> **Overview**
> **`list_group_members`** now builds the effective member list from the ephemeral projection via **`member_entries_with`** (identities + roles at the cut), reusing the same single DAG fold as the membership gate. On **`None`** (unfed namespace, incomplete fold, unfolded direct membership, or fold inconsistency) it still falls back to live **`list ∪ enumerate_inherited`**. Members are **sorted by `PublicKey`** before pagination so projection and live paths share a stable page order.
> 
> **`scope_projection`** adds **`member_entries_with`** and **`member_identities_in_view_with_ctx`** (shared root/cap context for identity set vs role walk). It removes **`shadow_member_enum_with`** and **`member_roles_for`**, which were only used for shadow validation on this handler.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e5e0ae6ae272cb3ee0a774ae77903b1b1278a6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/list_group_members.rs
+++ b/crates/context/src/handlers/list_group_members.rs
@@ -74,15 +74,9 @@ impl Handler<ListGroupMembersRequest> for ContextManager {
                 })
                 .map_or_else(
                     || -> eyre::Result<Vec<_>> {
-                        let mut live = MembershipRepository::new(&self.datastore).list(
-                            &group_id,
-                            0,
-                            usize::MAX,
-                        )?;
-                        live.extend(
-                            MembershipRepository::new(&self.datastore)
-                                .enumerate_inherited(&group_id)?,
-                        );
+                        let membership = MembershipRepository::new(&self.datastore);
+                        let mut live = membership.list(&group_id, 0, usize::MAX)?;
+                        live.extend(membership.enumerate_inherited(&group_id)?);
                         Ok(live)
                     },
                     Ok,
--- a/crates/context/src/handlers/list_group_members.rs
+++ b/crates/context/src/handlers/list_group_members.rs
@@ -24,10 +24,10 @@ impl Handler<ListGroupMembersRequest> for ContextManager {
             let Some((node_identity, _)) = self.node_namespace_identity(&group_id) else {
                 bail!("node has no group identity configured");
             };
-            // Fold the ephemeral projection ONCE for this request: the gate below
-            // and the enum shadow further down both read from it (one RocksDB DAG
-            // walk, not two). `None` (store fault) falls back to live for the gate
-            // and skips the shadow.
+            // Fold the ephemeral projection ONCE for this request: the membership
+            // gate below and the effective-member enumeration further down both read
+            // from it (one RocksDB DAG walk, not two). `None` (store fault) falls
+            // back to live for both.
             let proj_ctx = crate::scope_projection::ScopeProjections::ephemeral_projection(
                 &self.datastore,
                 &group_id,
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed:

- **Partial fold skips live fallback (Medium).** Real limitation, but **not new to this PR**: `member_entries_with` derives its id set from `member_identities_with` → `member_identities_in_view`, the *same* candidate universe the already-merged count (#2852 `get_group_info`) and cohort (`get_migration_status`) consumers use. The all-or-nothing materialized fallback (`!view.groups.contains_key`) is a property of that shared primitive. It's bounded by sync delivering a group's membership **atomically** — fully folded or fully materialized, never half — which is why the `membership-enum` plane held 0 divergences across the shadow runs. The durable close is **folding-completeness in P6 (#17)**, where sync delivers ops not materialized rows. Documented inline so it's not silently load-bearing.
- **`filter_map` silently drops `None`.** Fixed concretely: the `MemberPathAtCut::None` arm now **abandons the projection enumeration and defers to live** (returns `None` → handler uses the complete `list ∪ enumerate_inherited` union), with a `membership-enum` warn — never a truncated list with no signal. If the candidate filter and the role walk ever disagree, the caller gets live's complete answer, not a short one.
- **Two `MembershipRepository` instances in the fallback.** Bound to one local.

clippy `-D warnings` / fmt / equivalence 3/3 green.

chefsale:
Addressed:

- **Double fold (🟡).** Fixed: `member_entries_with` now folds the view **once** via `auth_cut_context` and derives the validated id set from it with the pre-folded `member_identities_in_view`, instead of folding once in `member_identities_with` and again in `auth_cut_context`. One ancestry walk per request; identical result (same view, same `cut_ancestry_complete` gate), and the unfolded-group fallback now short-circuits before the enumeration runs.
- **Per-reason observability of `None` (💡).** The only **abnormal** `None` reason — a validated-id/role-walk inconsistency — already emits a `membership-enum` warn inside `member_entries_with`. The other two (incomplete cited ancestry, unfolded group) are **expected steady-state deferrals** during backfill/transition; warning on them would be noise on the normal path. So the handler intentionally treats "defer to live" uniformly — the one case worth a signal already has one. The whole fallback retires in #29b.

clippy `-D warnings` / fmt / equivalence 3/3 green. Re-dispatching e2e.

chefsale:
Addressed:

- **Inherited-via-non-admin role: `unwrap_or(Member)` on the authoritative path (🟡 ×2, both flag this arm).** Replaced with an explicit bail: if the anchor row is absent, warn (`membership-role`) and `return None` so the caller falls back to live's complete union — consistent with the `MemberPathAtCut::None` arm, never a silently-guessed role. (The lookup still can't actually miss — `member_path_at_cut` only emits this arm when `groups[anchor]` contains the member, read from the same view — but the defensive bail is right for an authoritative read.)
- **`cut_ancestry_complete` gate not obvious (💡 ×2).** The fold-once commit already added the inline comment you asked for: `auth_cut_context` gates `cut_ancestry_complete` (returns `None` on a partial fold) — it's the first call in `member_entries_with`, so the guard is explicit and can't be bypassed.

clippy `-D warnings` / fmt / equivalence 3/3 green. Re-dispatching e2e.

chefsale:
Addressed:

- **BTreeSet vs store order — pagination (🟡).** Real wrinkle: the projection path yields `PublicKey`-sorted ids, the live fallback yields store order, so a request served by the projection and a later page served by the live fallback (e.g. mid-backfill) could skip/repeat members across pages. Now both paths are `sort_by_key(identity)` before `skip/take` — pagination is stable and independent of which path produced the set.
- **`cut_ancestry_complete` gate not obvious in the doc (🟡/💡 ×3).** Added it to the `member_entries_with` doc-comment explicitly: the `None`-on-partial-fold is enforced by `auth_cut_context`'s `cut_ancestry_complete` gate (also the first call in the body).
- **Handler silently falls back on a fold inconsistency (💡 ×2).** Updated the handler comment to spell out all `None` reasons, noting the one abnormal case (fold inconsistency) is already logged with a `unified_projection_divergence` warn inside `member_entries_with`; the others are expected steady-state deferrals where a handler-level warn would be backfill noise.

clippy `-D warnings` / fmt / equivalence 3/3 green. Re-dispatching e2e.

chefsale:
Addressed (no behavior change — a log label + a doc correction):

- **Anchor-absent warn used `plane="membership-role"` (💡).** Correct — a member in the validated identity set whose anchor row is missing from the same view is a membership-GRAPH inconsistency, not a role mismatch. Relabeled to `membership-enum`, matching the sibling `MemberPathAtCut::None` bail right above it.
- **"Identical identity set" contract vs the unfolded-group case (🟡).** Good catch: the equality holds for a FOLDED group, but for a wholly-unfolded group `member_identities_with` injects live rows (its materialized fallback) while `member_entries_with` returns `None`. I went with **option (b)** — documented that the equality is folded-group-only and that the *end result is identical* (on `None` the handler falls back to the live `list ∪ enumerate_inherited` union, yielding those same rows with roles). Not option (a): injecting live rows here would pull a `MembershipRepository::list` read back *into* the projection method, which is exactly what the `None`-defer avoids.

clippy `-D warnings` / fmt / equivalence 3/3 green. No e2e re-dispatch — this commit is a doc comment + a log-plane string, no code-path change.

chefsale:
Addressed:

- **Double-read of `Meta`/`Capabilities` + consistency window (🟡).** Real — `member_entries_with` read `root`/`default_cap_base` in `auth_cut_context` (for the role walk) and `member_identities_in_view` re-read them (for the candidate filter), so a concurrent governance write could make the two disagree. Added `member_identities_in_view_with_ctx(view, …, root, default_cap_base)`; `member_entries_with` now passes the **same** values its role walk uses, so the candidate filter and role resolution share one read. `member_identities_in_view` keeps its signature (computes + delegates), so the count/cohort consumers are untouched.
- **Anchor-absent warn plane: `membership-role` vs `membership-enum` (🟡).** Keeping `membership-enum` — deliberately. The two bail arms (`MemberPathAtCut::None` and anchor-absent) are the *same class* of event: an id in the validated set whose fold can't produce a complete answer, so the **whole enumeration** defers to live. No role value is compared (there's nothing to compare against — the enumeration bails). Labeling both bails `membership-enum` keeps monitoring consistent; a `membership-role` divergence elsewhere means an actual role *value* mismatch, which is a distinct signal worth not conflating.

clippy `-D warnings` / fmt / equivalence 3/3 green. No e2e re-dispatch (one read-dedup + a doc; no behavior change).

PR #2856, opened 2026-06-23
feat(governance): apply-auth at-cut authorizer seam (F5 #28 stage 3a)

chefsale:
## What

Opens the dependency-inversion **seam** for the final F5 consumer — apply-time governance-op authorization (#28). Behavior-neutral: this commit threads the plumbing; the actual flip onto the projection is stage 3b.

## Why a seam

The apply gates (e.g. `NamespaceApplyCtx::require_namespace_admin`) read the **live** membership resolver, which has no causal context — they decide against the receiver's current state, not the op's parents. F5 moves that decision onto the projection at the op's causal cut. But the projection lives in `crates/context`, which *depends on* `governance-store`, so governance-store can't call it directly.

`AtCutAuthorizer` inverts the dependency: **governance-store defines the trait the gates call; `crates/context` implements it over the projection and injects it** at the apply call site. The gate stays put; only the decision *source* is abstracted.

## How (this PR)

- `AtCutAuthorizer` trait + `LiveFallbackAuthorizer` (always `None` → defer to live) in `governance-store::authorizer`.
- `NamespaceGovernance::with_apply_auth` threads the op's parent cut + `&dyn AtCutAuthorizer` into the apply path; `NamespaceApplyCtx` carries them; `require_namespace_admin` consults `is_admin_at_cut` first, live on `None`.
- Production path `apply_signed_namespace_op_at_cut` (called from the context DAG applier, where `delta.parents` lives) passes the op's parents + the `LiveFallbackAuthorizer` — so gates keep using live. `apply_signed_namespace_op` stays a live-default shim; all existing callers unchanged.

## Safety

Byte-identical: `LiveFallbackAuthorizer` always returns `None`, so every gate falls through to the live resolver exactly as before. **All 399 governance-store tests pass unchanged.** No gate decision moves in this PR.

## Next

- **3b**: implement `AtCutAuthorizer` in `crates/context` over `ScopeProjections`, inject it from the DAG applier → the actual flip (live as `None`-fallback), gated on the existing `governance-auth` shadow staying divergence-free.
- **4**: capability gates + circular last-admin invariants (need the parent cut, now threaded).

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean; governance-store 399/399 unchanged; projection equivalence 3/3.

---

> [!NOTE]
> **Medium Risk**
> Touches namespace admin apply gates and the production DAG apply path, but `LiveFallbackAuthorizer` preserves live-only decisions; risk rises when stage 3b swaps in projection-backed auth.
> 
> **Overview**
> Introduces an **at-cut apply authorization seam** (F5 #28) so namespace governance gates can eventually authorize against the unified projection at each op’s causal cut, without `governance-store` depending on `crates/context`.
> 
> Adds **`AtCutAuthorizer`** and **`LiveFallbackAuthorizer`** (always `None` → live membership). **`NamespaceGovernance::with_apply_auth`** threads parent op hashes and a `&dyn AtCutAuthorizer` through apply; **`NamespaceApplyCtx::require_namespace_admin`** tries **`is_admin_at_cut`** first and falls back to **`MembershipRepository::is_admin`** on `None`. **`apply_signed_namespace_op_at_cut`** is the new entry point; **`apply_signed_namespace_op`** delegates to it with empty parents and the live fallback. The namespace DAG applier calls **`apply_signed_namespace_op_at_cut`** with **`delta.parents`** and **`LIVE_FALLBACK_AUTHORIZER`**.
> 
> **Behavior is unchanged in this PR**: the fallback authorizer never supplies a verdict, so gates still use the live resolver. Stage 3b will inject a projection-backed implementation from `crates/context`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8cf6298e13dd36defc3eb546c1f017ab19f34b2b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/governance_dag.rs
+++ b/crates/context/src/governance_dag.rs
@@ -104,9 +104,17 @@ impl NamespaceGovernanceApplier {
 #[async_trait::async_trait]
 impl DeltaApplier<SignedNamespaceOp> for NamespaceGovernanceApplier {
     async fn apply(&self, delta: &CausalDelta<SignedNamespaceOp>) -> Result<(), ApplyError> {
-        let outcome =
-            calimero_governance_store::apply_signed_namespace_op(&self.store, &delta.payload)
-                .map_err(|e| ApplyError::Application(e.to_string()))?;
+        // F5 #28: thread the op's causal cut + the at-cut apply-auth source into the
+        // governance-store apply gates. `LIVE_FALLBACK_AUTHORIZER` always returns
+        // `None`, so the gates keep using the live resolver — this commit only opens
+        // the seam; a projection-backed authorizer replaces it in the flip stage.
+        let outcome = calimero_governance_store::apply_signed_namespace_op_at_cut(
+            &self.store,
+            &delta.payload,
+            &delta.parents,
+            &calimero_governance_store::LIVE_FALLBACK_AUTHORIZER,
+        )
+        .map_err(|e| ApplyError::Application(e.to_string()))?;
         if let Some(report) = outcome.divergence {
             // Last-writer-wins on the outbox. The applier instance
             // is single-flight per actor message turn, so multiple
--- a/crates/governance-store/src/authorizer.rs
+++ b/crates/governance-store/src/authorizer.rs
@@ -1,0 +1,79 @@
+//! At-cut apply authorization seam (F5 #28).
+//!
+//! The apply-time governance gates (e.g. [`NamespaceApplyCtx::require_namespace_admin`])
+//! historically read the LIVE membership resolver, which has no causal context —
+//! they decide against the receiver's current state, not the op's own parents. F5
+//! moves that decision onto the unified projection, resolved at the op's causal
+//! cut. But the projection lives in `crates/context`, which DEPENDS on this crate,
+//! so this crate cannot reach it directly.
+//!
+//! [`AtCutAuthorizer`] inverts that dependency: this crate defines the trait the
+//! apply gates call; `crates/context` implements it backed by the projection and
+//! injects it at the apply call site. The gate stays here; only the decision
+//! source is abstracted. `None` from any method means "the projection has no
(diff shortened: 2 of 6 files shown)


chefsale:
Addressed both:

- **`is_admin_or_capability_at_cut` unused in this PR (💡 ×2).** Removed from the trait + `LiveFallbackAuthorizer`. The seam only needs `is_admin_at_cut` (`require_namespace_admin` is the one gate flipped here). The capability method lands with its caller in the next stage — the `PermissionChecker` capability-check flip — per the no-premature-API convention.
- **Raw `u32` capability param (💡).** Moot now that the method is removed; when it returns in the capability-gate stage I'll match whatever type that gate's caller uses (the codebase's capability bitmasks are `u32` via `MemberCapabilities`, so it'll likely stay `u32` for consistency, but I'll decide at the call site).

clippy `-D warnings` / fmt clean.

chefsale:
Addressed both:

- **`ContextGroupId` by value → `&ContextGroupId` (💡).** Switched `is_admin_at_cut` to take `&ContextGroupId`, matching the apply gates (`MembershipRepository::is_admin`, `require_admin`, …) which pass it by reference. `ContextGroupId` is `Copy`, so it's a call-site-consistency change, not a perf one. Propagated to the stacked flip impl too.
- **Object-safety / future async methods (💡).** Documented on the trait: it's deliberately synchronous and object-safe (the gates drive it via `&dyn AtCutAuthorizer` and the projection resolution is sync). Methods added later must stay sync — an `async` method would force `async_trait` boxing and break the `&dyn` apply-path use. If a future verdict needs async I/O, resolve it before the apply call and pass the result in.

clippy `-D warnings` / fmt clean; builds across context + server + node.

PR #2857, opened 2026-06-23
feat(context): flip apply-auth onto the projection (F5 #28 stage 3b)

chefsale:
## What

Stage 3b of F5 #28 — the apply-auth flip. `require_namespace_admin` now decides admin authority from the projection at the op's causal cut, not the live resolver.

## How

The seam (#2856, merged) routed the gate through `AtCutAuthorizer` (live-fallback no-op). This injects the real one:

- `EphemeralProjectionAuthorizer` (`pub(crate)` in `crates/context`) implements `AtCutAuthorizer::is_admin_at_cut` over the projection — folding an ephemeral projection of the op's namespace per apply, resolving at `delta.parents`.
- `None` defers to live: a store/DAG-head fault (warned) or ancestry not yet folded (quiet, normal mid-backfill).

The capability gate `is_admin_or_capability_at_cut` is NOT in scope — it lands with its `PermissionChecker` caller in stage 4 (no-premature-API).

## Safety

The `governance-auth` shadow (on master) resolved the same `is_admin_at_cut` verdict against live on every passing e2e; 0 divergences proves projection == live for this gate.

---

> [!NOTE]
> **Medium Risk**
> Changes who may apply namespace governance ops by sourcing admin checks from the projection at the op's causal cut, though incomplete folds and store faults still fall back to live per the `AtCutAuthorizer` contract.
> 
> **Overview**
> **F5 #28 stage 3b** wires real at-cut admin checks into namespace governance apply: `NamespaceGovernanceApplier` no longer passes `LIVE_FALLBACK_AUTHORIZER` and instead uses a new **`EphemeralProjectionAuthorizer`**.
> 
> That authorizer implements `AtCutAuthorizer::is_admin_at_cut` by folding an **ephemeral** namespace projection (same `ScopeProjections::ephemeral_projection` path as read-side gates) and resolving admin authority at **`delta.parents`**. **`None`** still means defer to the live resolver—either ancestry not fully folded (quiet) or a store/DAG-head fault when the fold cannot be built (**warn** logged).
> 
> The module is **`pub(crate)`** in `crates/context`; capability-at-cut gates are explicitly out of scope for this change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56306ef2434e196bb59394d40707c8c1b51c3186. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/apply_authorizer.rs
+++ b/crates/context/src/apply_authorizer.rs
@@ -44,15 +44,4 @@ impl AtCutAuthorizer for EphemeralProjectionAuthorizer<'_> {
         let (proj, _ns, _heads) = ScopeProjections::ephemeral_projection(self.store, &group)?;
         proj.is_admin_at_cut(self.store, group, signer, parents)
     }
-
-    fn is_admin_or_capability_at_cut(
-        &self,
-        group: ContextGroupId,
-        signer: &PublicKey,
-        capability: u32,
-        parents: &[[u8; 32]],
-    ) -> Option<bool> {
-        let (proj, _ns, _heads) = ScopeProjections::ephemeral_projection(self.store, &group)?;
-        proj.is_admin_or_capability_at_cut(self.store, group, signer, capability, parents)
-    }
 }
--- a/crates/context/src/apply_authorizer.rs
+++ b/crates/context/src/apply_authorizer.rs
@@ -1,0 +1,58 @@
+//! Projection-backed [`AtCutAuthorizer`] for the apply gates (F5 #28 stage 3b).
+//!
+//! Realizes the dependency inversion opened by the governance-store seam: the
+//! apply gates call the `AtCutAuthorizer` trait (defined in `governance-store`),
+//! and this implementation resolves the decision against the unified projection at
+//! the op's causal cut, falling back to the live resolver (`None`) when the cited
+//! ancestry isn't fully folded.
+//!
+//! It folds an EPHEMERAL projection from the store per call — the same mechanism
+//! the read-side query gates use ([`ScopeProjections::ephemeral_projection`]) — so
+//! it needs no shared projection state threaded into the apply path. The fold is
+//! bounded by `MAX_BACKFILL_OPS`; governance ops are infrequent, and P6 sync
+//! unification removes the per-apply fold.
+
+use calimero_context_config::types::ContextGroupId;
+use calimero_governance_store::AtCutAuthorizer;
+use calimero_primitives::identity::PublicKey;
+use calimero_store::Store;
+
+use crate::scope_projection::ScopeProjections;
+
+/// An [`AtCutAuthorizer`] that resolves each gate against a freshly folded
+/// ephemeral projection of the op's namespace, at the op's causal cut.
+pub struct EphemeralProjectionAuthorizer<'a> {
+    store: &'a Store,
+}
+
+impl<'a> EphemeralProjectionAuthorizer<'a> {
+    pub fn new(store: &'a Store) -> Self {
+        Self { store }
+    }
+}
+
+impl AtCutAuthorizer for EphemeralProjectionAuthorizer<'_> {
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed:

- **Silent fallback on store error (🟡 ×2).** Split the two `None` paths: a failed ephemeral **fold** (store / DAG-head fault) now **warns** before deferring to live, so a transient storage fault doesn't silently downgrade the apply-auth source. The expected "ancestry not yet folded" path (a frequent, normal mid-backfill state) stays quiet — it surfaces as `None` from the projection's `is_admin_at_cut`, not from a fold failure.
- **`pub` → `pub(crate)` (💡/📝 ×3).** `EphemeralProjectionAuthorizer` and its module are now `pub(crate)` — the only consumer is the namespace DAG applier in the same crate.
- **PR description mentioned both methods (💡).** Fixed the description: only `is_admin_at_cut` is implemented here; the capability method lands with its caller in stage 4. (Removed from the trait in the seam per the no-premature-API convention.)
- **Redundant fold per gate call (🟡).** With only `is_admin_at_cut` now (capability deferred), it's **one** fold per admin-gated op — the "folds twice for a combined check" case no longer exists. The per-apply fold cost is bounded by `MAX_BACKFILL_OPS`, governance ops are infrequent, and P6 sync-unification removes it (documented on the module).

clippy `-D warnings` / fmt clean; builds across context + server + node. e2e re-dispatched.

chefsale:
Both are about the ephemeral fold — neither needs a code change, but worth pinning the facts:

- **"Wrong namespace" risk (🟡 #47).** Not possible: `EphemeralProjectionAuthorizer` holds **only `&Store`** — it's stateless w.r.t. namespace, not "constructed for a namespace." `is_admin_at_cut` resolves the namespace **from its `group` argument** each call (`ephemeral_projection(store, group)`), and the gate passes the op's own namespace-root group (`ContextGroupId::from(self.namespace_id)`). So it always folds exactly the namespace of the group being authorized.

- **O(N) fold per apply (💡 #37).** Three mitigations, all already in place:
  1. **Bounded**, not unbounded — `collect_namespace_ops` caps the walk at `MAX_BACKFILL_OPS` (100k) and returns a partial fold past that (`is_admin_at_cut` then defers to live via the incomplete-ancestry `None`).
  2. **Lazy** — the fold only runs when an admin gate actually fires, i.e. inside `require_namespace_admin`, which is reached by `admin_changed` / `group_reparented` / `policy_updated` only. The common high-volume ops (invitation `MemberJoined`, etc.) are signature-gated and never trigger it — so it's per-admin-op, not per-op.
  3. **Documented + scheduled** — the module doc calls out the per-apply fold; the durable fix is reusing the maintained projection in P6 (sync unification), which the ephemeral approach was a deliberate, simpler stepping stone toward (chosen over threading the shared locked projection into the apply path).

So the cost is bounded, infrequent, and deferred — no behavior change here.

chefsale:
Same point as [the earlier thread](https://github.com/calimero-network/core/pull/2857#issuecomment-4780615739) (the bot re-anchored it to the new commit). The one correction worth repeating: it's **not O(n) per applied op** — the fold fires only inside `require_namespace_admin`, which is reached by exactly three handlers (`admin_changed`, `group_reparented`, `policy_updated`). Those are **rare lifecycle events** (changing the namespace admin, reparenting a group, updating policy), not per-member-join or per-data-write. So the bounded (`MAX_BACKFILL_OPS`) O(n) walk runs a handful of times over a namespace's life, not on the hot path.

Bounded + lazy + on rare ops + documented + superseded by the maintained-projection reuse in P6. Happy to file a standalone follow-up issue to track the P6 optimization if you'd prefer it off this PR — otherwise this is a deliberate, accepted tradeoff and not a blocker.

PR #2866, opened 2026-06-23
feat(governance): group-op apply-auth seam + group-auth shadow (F5 #28 stage 4a)

chefsale:
## What

Stage 4a of F5 #28. Stages 3a/3b (merged) flipped the **namespace-root** admin gate onto the projection. The **group-op** apply path (`PermissionChecker` admin + capability gates) still reads the live resolver with no causal context. This opens the seam for it and **shadows** — validate-first, behavior-neutral.

## How

- Thread the op's causal cut + `&dyn AtCutAuthorizer` through the group apply path: `GroupGovernanceApplier` → `apply_local_signed_group_op_at_cut` → `apply_group_op_mutations` → `GroupApplyCtx::new_with_apply_auth` → `PermissionChecker`. The namespace-envelope group-op path (`NamespaceGovernance` decrypt-and-apply) forwards the **seam's existing** parent cut + authorizer, so group ops over the wire shadow at the same cut as their enclosing namespace op.
- `PermissionChecker::is_admin` / `is_authorized_with_capability` compute the live verdict **and** (when an apply-auth authorizer is attached) the projection's at-cut verdict, logging `unified_projection_divergence` plane=`group-auth` on a mismatch. **Still act on live.**
- Reintroduce `is_admin_or_capability_at_cut` on `AtCutAuthorizer` (its caller is the capability shadow) + implement on `EphemeralProjectionAuthorizer`. `apply_local_signed_group_op` stays a live-default shim; the group DAG applier injects the real ephemeral authorizer so divergences actually surface.

## Safety

Byte-identical: the gates still act on live; the projection verdict is log-only. **All 399 governance-store tests pass unchanged.**

## Next

- **4b**: flip the gates (act on projection, live as `None`-fallback) once `group-auth` is divergence-free across e2e.
- **4c**: the circular last-admin invariants (`ensure_not_last_admin_removal`/`_demotion`) at the parent cut — needs a new `has_another_admin_at_cut` accessor.

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean (governance-store + context + server + node); governance-store 399/399 unchanged; equivalence 3/3.
- e2e: validates the new `group-auth` plane (0 markers = projection == live on group gates).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
[...]

PR #2870, opened 2026-06-23
feat(governance): flip group apply-auth gates onto the projection (F5 #28 stage 4b)

chefsale:
## What

Stage 4b of F5 #28 — flip the **group-op apply-auth gates** onto the projection. Stacked on #2866 (4a seam+shadow).

## How

4a threaded the at-cut authorizer through the group apply path and shadowed the `PermissionChecker` admin/capability gates (`group-auth` plane). This flips them:

- `PermissionChecker::is_admin` / `is_authorized_with_capability` decide from the projection at the op's causal cut (`is_admin_at_cut` / `is_admin_or_capability_at_cut`) — admin/capability authority **as of the op's own parents** (causal-honor). `None` (no apply-auth context, or incomplete fold) → live fallback (retires in #29b).
- `shadow_admin` helper removed.

## Scope (deliberately narrow)

Only the group-op **apply path** carries a projection-backed authorizer (via `GroupApplyCtx`), so only it flips. Handler-side pre-checks, cascade pre-scans, and tests keep the default live-fallback authorizer (`None` → live) — the F5 plan's "local-origination pre-checks stay on live" carve-out.

## Gating

> ⚠️ Merge-gates on #2866's `group-auth` plane being **divergence-free** across e2e — that's the proof the projection's group verdicts equal live before this flip acts on them. This PR's own e2e is the behavioral confirmation (group gates now run on the projection; any disagreement would fail a scenario).

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean; governance-store 399/399.
- e2e: behavioral (no shadow markers left on these gates).

> Base is `feat/f5-group-apply-auth-seam` (#2866); rebase to master after it merges.

---

> [!NOTE]
> **High Risk**
> Authorization on the group DAG apply path now follows projection-at-cut decisions; incorrect cut wiring or fold gaps would allow or deny ops differently than live store state until #29b removes the fallback.
> 
> **Overview**
> **F5 #28 stage 4b** moves group-op **apply** authorization from “shadow live, act on live” to **acting on the unified projection at the op’s causal cut** (`parent_op_hashes`), with **live** only when the authorizer returns `None` (no apply-auth context, incomplete fold, or empty cut).
> 
> `PermissionChecker::is_admin` and capability checks consult `is_admin_at_cut` / `is_admin_or_capability_at_cut` first; the `shadow_admin` divergence logging path is removed. Handler pre-checks and tests still use the default live-fallback authorizer.
> 
> `EphemeralProjectionAuthorizer` returns `None` when `parents` is empty so genesis/empty-cut cases are not judged against an empty fold.
> 
> Namespace-wrapped group ops now pass **`signed_group_op.parent_op_hashes`** into apply mutations instead of the enclosing `KeyDelivery` cut, so buffered encrypted-op replays are authorized at each candidate’s own parents.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ac85530be1e67b3d5851039388786bd694a2bc0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/apply_authorizer.rs
+++ b/crates/context/src/apply_authorizer.rs
@@ -81,6 +81,14 @@ impl AtCutAuthorizer for EphemeralProjectionAuthorizer<'_> {
         signer: &PublicKey,
         parents: &[[u8; 32]],
     ) -> Option<bool> {
+        // No cut (empty `parents`) ⇒ no causal context to resolve against; defer to
+        // live rather than judge against an empty/genesis view (which could falsely
+        // reject a capability-holder whose grant isn't in the empty fold). Belt-and-
+        // braces: group ops carry their enclosing namespace op's parents, which are
+        // empty only for the namespace genesis itself.
+        if parents.is_empty() {
+            return None;
+        }
         // `None` here = the cited ancestry isn't fully folded; the gate defers to
         // live (quiet — a normal mid-backfill state, not an error).
         self.folded(group)?
@@ -95,6 +103,10 @@ impl AtCutAuthorizer for EphemeralProjectionAuthorizer<'_> {
         capability: u32,
         parents: &[[u8; 32]],
     ) -> Option<bool> {
+        // Empty cut ⇒ defer to live (see `is_admin_at_cut`).
+        if parents.is_empty() {
+            return None;
+        }
         self.folded(group)?
             .0
             .is_admin_or_capability_at_cut(self.store, *group, signer, capability, parents)
--- a/crates/governance-store/src/lib.rs
+++ b/crates/governance-store/src/lib.rs
@@ -1232,9 +1232,9 @@ fn apply_group_op_mutations(
     Vec<crate::op_events::OpEvent>,
 )> {
     // The op's causal cut + at-cut authorizer ride into the apply gates so the
-    // `PermissionChecker` admin/capability gates can SHADOW the projection against
-    // live (F5 #28 stage 4). The default authorizer (live-fallback) makes the shadow
-    // inert for callers without an apply-auth context.
+    // `PermissionChecker` admin/capability gates resolve against the projection at
+    // the op's cut, with live as the `None`-fallback (F5 #28 stage 4). The default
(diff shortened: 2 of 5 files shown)


chefsale:
Addressed:

- **Medium — retry uses wrong auth cut.** Real bug. `apply_group_op_inner` passed `self.parents`; on the KeyDelivery retry path `retry_encrypted_ops_for_group` re-applies several buffered group candidates within **one** KeyDelivery apply, so `self.parents` was the KeyDelivery cut for every replayed op — the projection gates would judge each at the wrong cut. Now uses `signed_group_op.parent_op_hashes` (the candidate's own enclosing-namespace-op parents, copied in `decrypt_and_apply_group_op`). Equal to `self.parents` on the normal path; per-candidate on retry. (The cache is keyed by group, not cut — the fold is the whole namespace and the cut is applied at resolve time — so it's correct across candidates.)
- **Empty `parents` → projection verdict (🟡 ×2).** Added a guard: empty `parents` (a genesis op, or a `PermissionChecker` built without `with_apply_auth`) make the authorizer return `None` → defer to live, rather than judging against an empty/genesis fold. (The no-`with_apply_auth` case already used `LiveFallbackAuthorizer` → `None`; this also covers a genesis op reaching the real authorizer.)
- **Duplicated ephemeral fold / missing None-comment (💡 ×2) — stale.** These are from before the per-apply fold **cache** (merged in #2866): both methods now go through one `folded()` helper that owns the store-fault warn, so there's no duplication and the two-level `None` distinction lives in `folded()`. The new empty-parents guard is the only per-method preamble.
- **Projection bypasses the inherited-admin walk (🟡)** — intentional: the projection's `is_admin_at_cut` IS the at-cut equivalent of the inherited-admin walk (`is_authorized_admin` walks the open-subgroup chain); the live `is_inherited_admin` only runs on the `None` fallback.

clippy `-D warnings` / fmt clean; 399/399. Re-dispatched e2e.

PR #2871, opened 2026-06-23
feat: last-admin invariants at-cut resolver + shadow (F5 #28 stage 4c)

chefsale:
## What

Stage 4c of F5 #28 — the **circular last-admin invariants**. Stacked on #2870 (4b).

`ensure_not_last_admin_removal` / `ensure_not_last_admin_demotion` are the genuinely *circular* apply gates: they read admin state the op is about to mutate, so they must resolve at the op's **parent cut** (pre-mutation), not post-mutation live state. This adds the at-cut resolver and **shadows** it — validate-first.

## How

- **`ScopeProjections::is_last_admin_at_cut`** — would removing/demoting `member` orphan `group`'s admins at the cut? Mirrors live *exactly*: `member` is an admin via a direct `Admin` row OR the genesis admin; "another admin" counts only another `Admin`-role **row** (the genesis admin alone doesn't satisfy it, matching live's row-based `has_another_admin`). `None` on incomplete ancestry → defer to live.
- **`AtCutAuthorizer::is_last_admin_at_cut`** + `EphemeralProjectionAuthorizer` impl.
- **`MembershipPolicy`** carries the seam's parent cut + authorizer (`with_apply_auth`, threaded from `GroupApplyCtx`). The two invariants compute the live block decision **and** the projection's at-cut verdict, logging `unified_projection_divergence` plane=`last-admin` on a mismatch. **Still act on live.**

## Safety

Byte-identical: live decision is authoritative, projection is log-only. **All 399 governance-store tests pass unchanged.**

## Next

The flip (resolve the invariants on the projection, live as `None`-fallback) follows once `last-admin` is divergence-free across e2e. Then **#29b** deletes the live resolver — the F5 finale.

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean (governance-store + context + server + node); 399/399; equivalence 3/3.
- e2e: validates the new `last-admin` plane.

> Base is `feat/f5-group-apply-auth-flip` (#2870); rebase down the stack as each merges.

---

> [!NOTE]
> **Medium Risk**
> Touches governance apply invariants for admin removal/demotion; behavior stays live-authoritative today, but mismatches in the new shadow path could hide problems before a future flip to projection-first enforcement.
> 
> **Overview**
> Adds **parent-cut** resolution for the circular last-admin rules (block removing or demoting the sole admin), aligned with F5 #28 stage 4c. **`ScopeProjections::is_last_admin_at_cut`** answers whether a member is an admin at the op’s parent cut and has no other **Admin-role row** (genesis admin counts for “is admin” but not for “another admin”), with **`None`** when ancestry isn’t fully folded.
> 
> **`AtCutAuthorizer::is_last_admin_at_cut`** is wired through **`EphemeralProjectionAuthorizer`** and documented on the trait (including the empty-**`parents`** → **`None`** contract). **`MembershipPolicy`** takes the apply cut and authorizer via **`with_apply_auth`** (from **`GroupApplyCtx`**); removal/demotion still **enforce live**, but **`shadow_last_admin`** logs **`unified_projection_divergence`** on plane **`last-admin`** when the projection disagrees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e34e1b4960e0f6877ee64da320b23799d88bb335. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Thanks — these split into "the at-cut resolver" (persists) and "the shadow helper" (removed in the very next PR, #2872 4c-flip):

**`is_last_admin_at_cut` (the resolver that stays):**
- **Genesis-admin asymmetry (🟡 ×2).** Deliberate and *exact*: `member_is_admin` counts the genesis admin (direct `Admin` row OR `group_admin`/root tuple), but `has_other` scans `Admin`-role **rows only** — because that's precisely what live does (`GroupMembershipView::is_admin` includes the genesis admin; `has_another_admin` lists stored rows and ignores it). So e.g. "group has genesis admin G (no row) + one Admin-row A; remove A" → live blocks it (G has no row to count), and the projection blocks it identically. It's a faithful mirror of live's behavior — including that quirk — **validated 0-divergence on the `last-admin` plane** in this PR's e2e. Matching live exactly is the whole point of the shadow; introducing a "smarter" rule here would *create* a divergence.
- **`root_g == group` value-equality (🟡/💡).** Correct as gated: the root tuple only counts the namespace-root genesis admin **for the root group** (`root_g == group`). A *subgroup's* genesis admin is the separate `view.group_admin.get(&group) == Some(member)` branch. Both cases covered; no non-root leakage.

**`shadow_last_admin` (removed in #2872):**
- **`gate: &'static str` → enum (💡), short-circuit before shadow (💡), error handling (🟡), empty-parents skip (🟡 ×2)** — the shadow helper and its `gate` param are **deleted entirely in #2872** (the flip acts on `is_last_admin_at_cut` directly, no shadow). On the specifics: `&&` already short-circuits `has_another_admin` when `!is_admin` (no extra store read); the `?` on the live calls propagates errors before the shadow runs (nothing swallowed); and the empty-parents skip is intentional (genesis ops defer to live — now explicit via the empty-cut guard in the authorizer). Introducing an enum for a 2-call-site helper that's deleted in the next stacked PR would be churn.

clippy `-D warnings` / fmt clean; 399/399; e2e green (0-div on `last-admin`).

chefsale:
Addressed (the branch had master merged in via "Update branch"; my change is on top):

Both comments are about `shadow_last_admin` + the empty-parents edge — and the underlying concern ("don't rely on each `AtCutAuthorizer` impl remembering to guard empty `parents`") is fair. Rather than scatter guards, I made it an explicit **trait contract**: `AtCutAuthorizer` now documents that every method MUST return `None` for empty `parents` (no causal cut → defer to live). `EphemeralProjectionAuthorizer` already honors it; the contract means a future impl can't silently regress it.

On the specifics:
- **Shadow inert without `with_apply_auth` (🟡)** — intended; non-apply constructions (handler pre-checks, cascade, tests) carry `LIVE_FALLBACK_AUTHORIZER` → `None`, so no shadow fires. `shadow_last_admin` itself is removed in the next stacked PR (#2872, the flip).
- **Future impl forgets the empty guard (🟡)** — now covered by the trait contract above.

clippy `-D warnings` / fmt clean. Re-dispatched e2e.

chefsale:
Good catch on the deeper point (the comment's own body lands on it: `projected != live_blocks` does catch both directions). The real observation is that the projection reads the **parent cut** while the live pre-check reads **current materialized state** — so under concurrency they can differ.

That's not a false positive — it's the **causal-honor** property the whole migration is built on: the op is judged as it was authored (its own ancestry), not against state the node happens to have reached via concurrent branches. The DAG order resolves the convergent admin set. e2e held **0 divergences** on `last-admin` precisely because the scenarios don't apply admin-cardinality ops outside an op's ancestry; if they did, the projection diverging from live's non-causal read is the projection being *more* correct.

I documented this on the **persisting** `is_last_admin_at_cut` (the accessor the flip uses), not `shadow_last_admin` (which #2872 deletes) — that's where it'll keep mattering. No behavior change.

clippy `-D warnings` / fmt clean. Re-dispatched e2e.

chefsale:
These refine two earlier threads; the substantive one is the root-admin source, so let me ground it:

**`is_last_admin_at_cut` reads the root admin from the same source live does (🟡 ×2).** `auth_cut_context`'s `root` is `MetaRepository::load(root_group).admin_identity`. Live's last-admin `is_admin` (`GroupMembershipView::is_admin` → `is_group_admin`) checks `meta.admin_identity == identity` (`membership/core.rs:346`) — the **same** `MetaRepository` value. So the root-admin branch of `is_last_admin_at_cut` agrees with live *by construction*; the rows are the folded/causal part. That's exactly why the `last-admin` plane is 0-divergence.
- On the "un-folded vs folded" point: the `auth_cut_context` doc's "un-folded base" describes `is_authorized_admin`'s *preference* (folded root admin over base) for the admin-AUTHORITY gate. But the last-admin invariant must mirror live's `is_admin`, which uses `meta.admin_identity` directly — so using the folded `root_admin` here would *diverge* from live, not improve it. The match with live is the invariant the whole migration preserves; I won't trade a validated equality for a "more folded" read that breaks it.
- `root_g == group` correctly scopes the meta-admin to the root group; subgroup genesis admins are the separate `group_admin` branch (folded). Both covered.

**Shadow logging (🟡 ×2 re-anchored).** `projected != live_blocks` catches both directions (the comment bodies concede this). `None` from the authorizer = defer-to-live = intentional skip. `shadow_last_admin` is deleted in #2872 (the flip acts on `is_last_admin_at_cut` directly), so the helper's logging shape is transient.

Net: faithful live mirror, validated 0-divergence end to end; no behavior change warranted.

PR #2881, opened 2026-06-23
feat(governance): flip MemberJoinedOpen check_path onto the projection (F5 #29b)

chefsale:
## What

Flips `member_joined_open`'s membership-path gate onto the projection. Follows #2878 (the `membership-path` shadow, which validated **0 divergences**).

## How

`member_joined_open::apply` now resolves the path via `NamespaceApplyCtx::membership_path` → the projection's `membership_path_at_cut` at the op's causal cut, with live `check_path` as the `None`-fallback. `shadow_membership_path` becomes `membership_path` (returns the resolved kind rather than just logging).

## Safety

The membership-path equivalence was validated divergence-free on real e2e traffic (#2878); the `None`-fallback keeps live for cold/incomplete folds. Causal-honor: the join is judged against its own ancestry.

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean; 399/399.
- e2e: behavioral (the gate now runs on the projection).

---

> [!NOTE]
> **Medium Risk**
> Changes who may pass open-subgroup join apply checks when projection and live `check_path` disagree; fallback preserves prior behavior when the projection abstains.
> 
> **Overview**
> **MemberJoinedOpen** membership-path authorization now uses the **at-cut projection** (`membership_path_at_cut` at the op’s causal parents) instead of always reading live `check_path` while only shadow-logging mismatches.
> 
> `NamespaceApplyCtx::shadow_membership_path` is replaced by **`projection_membership_path`**, which returns `Option<AtCutMembershipPath>` and no longer compares or warns against an eager live path. In **`member_joined_open::apply`**, the gate takes the projected path when present; on `None` (no apply-auth context, incomplete fold, or default live-only authorizer) it **lazily** falls back to live `check_path` via `membership_path_kind`, so store errors on live reads cannot reject applies the projection would allow. Accept/reject branches now match on **`AtCutMembershipPath`** directly rather than live `MembershipPath`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 197850c8ef79b2f15ff55728396df677fa3f2a04. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/governance-store/src/ops/namespace/context.rs
+++ b/crates/governance-store/src/ops/namespace/context.rs
@@ -96,20 +96,20 @@ impl<'a> NamespaceApplyCtx<'a> {
         Ok(())
     }
 
-    /// The at-cut membership PATH for `member` in `group`, for the `MemberJoinedOpen`
-    /// gate (F5 #29b flip): the projection's `membership_path_at_cut` at the op's
+    /// The PROJECTION's at-cut membership PATH for `member` in `group`, for the
+    /// `MemberJoinedOpen` gate (F5 #29b flip): `membership_path_at_cut` at the op's
     /// causal cut, validated divergence-free on the `membership-path` plane. `None`
-    /// (no apply-auth context — sign/test — or an incomplete fold) falls back to the
-    /// `live_path` the caller computed from `check_path`. The live fallback retires
-    /// once `check_path` is deleted.
-    pub(crate) fn membership_path(
+    /// (no apply-auth context — sign/test — or an incomplete fold) means the caller
+    /// must fall back to live `check_path`. Returning the `Option` (rather than taking
+    /// an eager `live_path`) lets the caller compute live LAZILY — only when the
+    /// projection abstains — so a `check_path` store error can't abort an apply the
+    /// projection would have decided.
+    pub(crate) fn projection_membership_path(
         &self,
         group: &ContextGroupId,
         member: &PublicKey,
-        live_path: crate::authorizer::AtCutMembershipPath,
-    ) -> crate::authorizer::AtCutMembershipPath {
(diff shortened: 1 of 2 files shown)


chefsale:
Good catch — fixed. `ctx.membership_path` took an eager `live_path`, so `check_path` ran (and its `?` could abort the apply on a store error) even when the projection answered `Some` and the live result was thrown away. Changed it to `projection_membership_path` returning the projection's `Option`, and the caller now computes live `check_path` **lazily** — only on `None`:

```rust
let path = match ctx.projection_membership_path(&gid, &member) {
    Some(projected) => projected,
    None => membership_path_kind(&MembershipRepository::new(store).check_path(&gid, &member)?),
};
```

So when the projection decides, live isn't read at all — no wasted walk, and no `check_path` store error can abort an apply the projection would have authorized.

clippy `-D warnings` / fmt clean; 399/399. Re-dispatched e2e.

chefsale:
Both of these (the eager `check_path?` dead-weight and the store-error-on-fallback) are addressed by the lazy-live fix already pushed (197850c8): the gate now reads the projection's `Option` first and only computes live `check_path` when the projection abstains —

```rust
let path = match ctx.projection_membership_path(&gid, &member) {
    Some(projected) => projected,
    None => membership_path_kind(&MembershipRepository::new(store).check_path(&gid, &member)?),
};
```

So when the projection decides, live isn't read at all (no wasted walk, no `check_path` store error reaching the apply). When the projection abstains (`LiveFallbackAuthorizer`, or an incomplete fold), live is the authority and a genuine store error there *should* propagate — that's correct, not masking (the apply can't be authorized if neither source can decide).

PR #2882, opened 2026-06-23
feat(node): shadow the data-write role against the projection (F5 #29b)

chefsale:
## What

The validate-first step for the `verify.rs` data-write migration (the substantial #29b piece). `verify.rs` derives the authorized author's **role** from live `acl_view_at`, and that role writes through to peer-identity observation (sync peer selection + durable cache). The membership verdict is already projection-validated (`membership-cut`/`membership-cut-grant`); the **role** is the last un-validated piece — so shadow it before flipping verify.rs off `acl_view_at`.

## How

- `ScopeProjections::role_at_cut_for_group` — `role_at_cut` with namespace resolution + `cut_ancestry_complete` gating (node-side accessor).
- In the gossip-receive `Authorized` arm, after the existing co-authorizer refresh (`projection_member_at_cut`), read the projection's at-cut role and log `unified_projection_divergence` plane=`data-write-role` on a mismatch. **Reuses the refresh already done** — no extra fold. Log-only.

## Safety

Behavior-neutral: live's `role` still flows to peer observation; the projection role is log-only. The membership decision is unchanged (already projection-authoritative).

## Next

Once `data-write-role` is divergence-free: flip `verify.rs` to source role + membership from the projection, compute `Buffer{needed}` via a direct op-log presence check, retiring `acl_view_at`. Then the deletion PR.

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean.
- e2e: validates the new `data-write-role` plane.

---

> [!NOTE]
> **Low Risk**
> Log-only shadow on an already-gated Authorized path; live membership and role observation are unchanged until a follow-up flip.
> 
> **Overview**
> Adds **`ScopeProjections::role_at_cut_for_group`**, which resolves the **effective** author role at the governance cut (direct, inherited-via-admin, or anchor role) using the same `member_path_at_cut` path as membership-role validation, with namespace resolution and **`cut_ancestry_complete`** gating.
> 
> On the gossip **Authorized** path, after the existing membership co-authorizer refresh, the handler compares that projected role to live’s role from **`verify.rs`** / **`acl_view_at`**. Mismatches emit **`unified_projection_divergence`** with plane **`data-write-role`** only — **no change** to authorization or peer-identity observation (live role still wins).
> 
> This is the validate-first step before flipping verify to source role from the projection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82d24eefc7e4a351babfb12906b6cc562d07eeb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -1396,16 +1396,19 @@ impl ScopeProjections {
             calimero_authz::MemberPathAtCut::Inherited {
                 via_admin: true, ..
             } => Some(GroupMemberRole::Admin),
+            // `member_path_at_cut` only emits this arm when the anchor row is present,
+            // so the lookup resolves; if it somehow doesn't, return `None` (defer to
+            // live / skip the shadow) rather than GUESS `Member` — guessing could emit
+            // a spurious `data-write-role` divergence. Matches `member_entries_with`,
+            // which bails rather than fabricating a role on the same inconsistency.
             calimero_authz::MemberPathAtCut::Inherited {
                 anchor,
                 via_admin: false,
-            } => Some(
-                view.groups
-                    .get(&anchor)
-                    .and_then(|m| m.get(member))
-                    .cloned()
-                    .unwrap_or(GroupMemberRole::Member),
-            ),
+            } => view
+                .groups
+                .get(&anchor)
+                .and_then(|m| m.get(member))
+                .cloned(),
         }
     }
 
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -1372,11 +1372,15 @@ impl ScopeProjections {
             .cloned()
     }
 
-    /// [`role_at_cut`](Self::role_at_cut) resolving the namespace from `group` and
-    /// gating on COMPLETE cited ancestry — the node-side accessor for the data-write
-    /// role (the role written through to peer-identity observation). `None` when the
-    /// namespace is unresolved, the cut isn't fully folded (defer to live), or
-    /// `member` has no folded direct row at the cut.
+    /// The EFFECTIVE role of `member` in `group` at the cut, resolving the namespace
+    /// from `group` and gating on COMPLETE cited ancestry — the node-side accessor for
+    /// the data-write role (written through to peer-identity observation). Mirrors
+    /// live `acl_view_at`'s `Member(role)`, which is the *effective* role: a direct
+    /// member's folded role, `Admin` for an inherited-via-admin member, else the
+    /// member's role at the anchor it inherits from (the same derivation the
+    /// `membership-role` plane validated — NOT just `groups[group]`, which would miss
+    /// inherited/admin paths). `None` when the namespace is unresolved, the cut isn't
+    /// fully folded (defer to live), or `member` isn't a member at the cut.
     #[must_use]
     pub fn role_at_cut_for_group(
         &self,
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed:

- **Shadow uses direct not effective role (Medium).** Correct and important — live's `acl_view_at` `Member(role)` is the EFFECTIVE role (inherited + admin carve-outs), but `role_at_cut_for_group` read only the direct `groups[group]` row, so for inherited/admin-path members they'd disagree (missed mismatches or spurious `data-write-role` divergences). Fixed: it now derives the effective role via `member_path_at_cut` — the same derivation the `membership-role` plane already validated (direct role / `Admin` for inherited-via-admin / role at the inheritance anchor). Now it compares like-for-like.
- **Doc nit re inherited members (📝).** Resolved by the same change — the accessor no longer returns `None` for inherited members; the doc now describes the effective-role derivation.
- **Double read-lock / redundant resolve (🟡).** The effective-role version uses `auth_cut_context` (one resolve + one fold) under a single read-lock acquisition. There's still a separate lock from the earlier `projection_member_at_cut` refresh, but this is the SHADOW path — it's removed entirely when verify.rs flips (the role is then sourced once, where membership already is). Not worth threading a shared lock through a transient shadow.

clippy `-D warnings` / fmt clean. Re-dispatched e2e.

chefsale:
Addressed:

- **Inherited anchor defaults Member (Medium).** Good catch — fixed. The non-admin inherited branch now returns `None` when the anchor row is somehow absent, rather than guessing `GroupMemberRole::Member`. `member_path_at_cut` only emits that arm when the anchor row is present, so it's a fold-inconsistency guard — and for the shadow, returning `None` (defer to live / skip) avoids a spurious `data-write-role` divergence, exactly as `member_entries_with` handles the same case.
- **Second read-lock for the role shadow (💡/🟡, re-anchored).** Acknowledged — it's a second lock after `projection_member_at_cut`'s refresh. This is the SHADOW path; when verify.rs flips, the role is sourced once at the same place membership already is, and this block (plus the extra lock) is removed. Not worth threading a shared lock through a transient shadow.

clippy `-D warnings` / fmt clean. Re-dispatched e2e.

chefsale:
Two follow-ups, both reply-only (the Medium "Inherited anchor defaults Member" and the doc nit are already fixed in 82d24eef / the effective-role change):

- **`None` conflates non-member with incompleteness (💡).** Intentional division of labor. This block runs AFTER the co-authorizer (`projection_member_at_cut`): if the projection said `Some(false)` (not a member), we'd already have denied and returned on the `membership-cut` plane. So at the role-shadow point the membership verdict is `Some(true)` or `None` — and `MemberPathAtCut::None` there means an incomplete fold (skip → defer to live), not a silent "live-says-member-projection-doesn't" gap. Member PRESENCE is the `membership-cut` plane's job; this plane only validates the ROLE for members both sides see — exactly the split the `membership-enum`/`membership-role` planes use.
- **Second read lock (🟡/💡, re-anchored ×3).** It's a second read-lock after the co-authorizer's refresh, but: the projection is already refreshed (no second fold-walk beyond the `acl_view_at` read), it's a *read* lock in the single-threaded `ContextManager` actor (no contention), and this entire block — plus the extra lock — is removed when verify.rs flips (the role is then sourced once, where membership already is). Threading a shared lock through a transient shadow isn't worth it.

e2e green pending; no behavior change.

PR #2885, opened 2026-06-23
feat(node): shadow the pure-projection data-write decision (F5 #29b)

chefsale:
## What

The validate-first step before the `verify.rs` flip — the hardest, convergence-critical part of #29b. To retire `acl_view_at`, the data-write decision collapses from "live baseline + two-direction projection refinement" into a single **pure-projection verdict** (`member_at_cut_authoritative`: `Some(true)`=authorize, `Some(false)`=reject, `None`=buffer). This shadows that collapse before doing it.

Membership + role are already validated (`membership-cut`/`membership-cut-grant`/`data-write-role`); the open question is whether the *single* verdict reproduces the current *net* decision.

## How (log-only, plane `data-write-decision`)

- **Authorized arm**: live authorized + co-authorizer didn't deny ⇒ the pure verdict must be `Some(true)`; log if `Some(false)`/`None`.
- **MembershipReject else-branch**: when `member_at_cut_authoritative` is `None`, the pure path would **buffer** (incomplete fold) where the current path **rejects** — the one deliberate behavioral difference (buffer holds the delta until governance catches up vs dropping it). `Some(false)` is silent (both reject).

## Safety

Behavior-neutral: the current path still decides; the pure verdict is log-only. The known `None`→buffer-vs-reject difference is surfaced explicitly so we can confirm it's the *only* divergence before the flip.

## Next

Flip `verify.rs` to produce the pure verdict (+ `Buffer{needed}` via op-log presence), retiring `acl_view_at`, once `data-write-decision` is divergence-free. Then namespace_sync, then the deletion PR closes F5.

## Test

- `cargo clippy --all-targets -- -D warnings` / `fmt` clean.
- e2e: validates the new `data-write-decision` plane.

---

> [!NOTE]
> **Low Risk**
> Observability-only on the primary gossip path; authorization outcomes are unchanged until the planned verify.rs flip.
> 
> **Overview**
> Adds **log-only** “decision shadow” on the gossip state-delta cross-DAG path in `handle_state_delta`, on plane `data-write-decision`, ahead of retiring `acl_view_at` in `verify.rs`.
> 
> When live **authorizes** and the co-authorizer’s conservative `member_at_cut` is **`None`** (incomplete fold), a `debug!` with marker `data_write_decision_shadow` records that the future pure-projection verdict would **buffer** while today the delta **applies**.
> 
> On **`MembershipReject`**, grant checks now keep the full **`Option<bool>`** from `member_at_cut_authoritative` instead of collapsing to a boolean. If the authoritative read is **`None`**, the same shadow log notes buffer-vs-**reject**—the deliberate post-flip difference when governance ancestry isn’t fully folded.
> 
> These events use a **non-gate** marker so the expected `None` divergence does not trip `unified_projection_divergence`; apply/reject behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd20c45eebab35a4b24c21c0ef12fd0af134ff03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/handlers/state_delta/mod.rs
+++ b/crates/node/src/handlers/state_delta/mod.rs
@@ -1061,28 +1061,24 @@ pub async fn handle_state_delta(
                 }
 
                 // DECISION SHADOW (F5 #29b): once `verify.rs` drops `acl_view_at`, the
-                // data-write decision becomes the PURE projection verdict
-                // (`member_at_cut_authoritative`: Some(true)=authorize, Some(false)=reject,
-                // None=buffer). Here live AUTHORIZED and the co-authorizer above didn't
-                // deny, so the pure path must also AUTHORIZE — i.e. resolve `Some(true)`.
-                // If it's `Some(false)`/`None`, the pure path would reject/buffer a write
-                // the current path applies — log it (plane `data-write-decision`). Live
-                // remains authoritative; the flip lands once this is divergence-free.
-                let authoritative = projection_member_at_cut_authoritative(
-                    &node_state,
-                    datastore,
-                    group,
-                    &author_id,
-                    heads,
-                );
-                if authoritative != Some(true) {
-                    warn!(
-                        marker = "unified_projection_divergence",
+                // data-write decision becomes the PURE projection verdict — the SAME
+                // conservative `member_at_cut` (`projected`) the co-authorizer already
+                // computed (NOT the strict `member_at_cut_authoritative`, which would
+                // reject a materialized member live + this gate authorize). Here
+                // `projected` is `Some(true)` (authorize, matches) or `None`: `None` means
+                // the cited ancestry isn't fully folded yet, so the pure path would BUFFER
+                // (hold until governance catches up) where the current path APPLIES.
+                // (`Some(false)` already returned above.) That `None`→buffer case is the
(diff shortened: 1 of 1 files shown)


chefsale:
(The review summary is "No Issues Found / LGTM"; this is the one inline note.)

`heads` is correct — it's `gp.governance_dag_heads`, the op's signed governance cut, the same cut the co-authorizer and the `data-write-role`/grant arms use. The substance is the redundant **refresh**: `projection_member_at_cut` (co-authorizer) and `projection_member_at_cut_authoritative` (decision shadow) each call `refresh_projection_for_cut` for the same cut.

It's a transient cost: the second refresh runs against an already-refreshed projection (the co-authorizer just folded this exact cut), so it's a near-no-op, in a single-threaded `ContextManager` actor (no contention). And the whole decision-shadow block — including the second refresh — is removed when verify.rs flips: the pure verdict is then sourced ONCE, where membership already resolves. Threading a combined "refresh-once, read both verdicts" helper through a shadow that's about to be deleted isn't worth it.

No behavior change; e2e validates the `data-write-decision` plane.

Issue #2607, opened 2026-06-02
sync: HashComparison reports root_hash_verified=true while permanently divergent after post-restart FrozenValue write (regression of #2407/#2411, peer-restart workflow)


chefsale:
## Summary

`sync-resilience-peer-restart` failed on **master** (commit `35cb322`) — two nodes stay **permanently split-brained** after a single post-restart write, and HashComparison reports `root_hash_verified=true` on every cycle while the roots clearly differ.

- Failing run: https://github.com/calimero-network/core/actions/runs/26806364110/job/79025981606
- Workflow: `sync-resilience-peer-restart` (scaffolding-e2e app)
- Outcome: `Wait for post-restart write to converge` timed out after **120.5s / 200 attempts**.

```
Final state:
  context=Hsy4k9e6c12LHDerJiFnRBKeyajHEvwShR6RgBSHAowY:
    sync-resil-node-1: GJiFLmHYq7VVJAHBRLHLMJ6gmC6K2KoEV7gqpoXhJDz   (post-restart write applied)
    sync-resil-node-2: k1Yyqdaw5ZrBsEKKa9NjmnBMifEgJ6uRCtGjaBnYWfC   (stuck on baseline)
```

This is **not** a connectivity, relay-reservation, or data-loss-on-restart problem (those are ruled out below). It is a HashComparison non-convergence / false-verification defect, and it is a **regression / residual of an already-closed cluster**: #2407, #2411, #2418, #2571/#2575, in the workflow from #2469.

## Timeline (from the uploaded node logs)

1. **Baseline converges** — both nodes reach root `k1Yy` (22 entities). ✅
2. **node-1 restarts** (08:03:16→28) and **comes back with data intact**. Between 08:03:28 and 08:03:42 it advertises `peer_root_hash=k1Yy` and node-2 logs:
   ```
   Protocol selected ... protocol=None reason=root hashes match, already in sync
     local_root=k1Yy remote_root=k1Yy local_entities=22 remote_entities=1
   ```
   → so the restart itself recovered cleanly; **no data loss**.
3. **node-1 post-restart write** (08:03:43.26): root advances `k1Yy → GJiF`. Correct.
4. **node-2 never adopts it.** For the full 120s window node-2 runs HashComparison against node-1 and **every one of 208 cycles** logs:
   ```
   HashComparison sync complete  nodes_compared=23 entities_merged=21
     entities_pushed=0 nodes_skipped=2 root_hash_verified=true
[...]

PR #2900, opened 2026-06-24
feat(sync): scope_root is the authoritative HC convergence verdict (cutover C1a)

chefsale:
**Stacked on #2899 (C0).** Retargets to `master` once C0 lands.

⚠️ **Merge-gated on C0's e2e canary.** C1 turns the C0 shadow into the *actual* convergence verdict, so a C0 false-positive divergence would become a "never converges" stall here. Do **not** merge until C0 shows the hash-neutral-rotation canary fires correctly **and** steady-state scenarios stay divergence-free.

## What

Promotes the C0 shadow from observe-only to the **authoritative verdict**. HashComparison's end-of-session convergence check (`root_hash_verified`, the #2607 re-query path) now decides on `scope_root` when both peers resolve one:

```rust
stats.root_hash_verified = match (local_scope_root, peer_scope_root) {
    (Some(local), Some(peer)) => local == peer,         // scope_root subsumes entities — it alone decides
    _ => local_root_hash == peer_current_root,          // cold projection / non-group → pre-C1 fallback
};
```

`scope_root` folds the governance projection's ACL + membership/admin onto the entity root, so two nodes converge only when **both** the data plane **and** authorization agree — closing the hash-neutral rotation blind spot behind the rotation split-brain family (stale-heads, `clear()`-tombstone-blindness, #2607 verified-but-divergent).

No context drops below its pre-C1 check: when either side can't fold the scope (`None`), it falls back to the exact bare-entity compare from before.

A new `scope_root_governance_divergence` marker distinguishes the case the entity root hides (entities agree, `scope_root` differs ⇒ pure ACL/governance divergence; the entity tree-walk has nothing to reconcile, and the rotation propagates via the ordinary governance sync tick).

## Scope (C1a)

The HashComparison verdict only — the general sync path C0 instrumented. **C1b** follows for LevelWise (advisory handshake-root verdict — needs `scope_root` threaded through the handshake), the snapshot boundary (`snapshot.rs`), and the protocol selector. Sliced out because those carry their own staleness semantics and are independently reviewable. See the C1 section of `docs/design/unified-causal-log-cutover-plan.md`.

## e2e gate

The C1 hash-neutral-rotation canary (maintainer-run): two nodes reach an identical `entities_root`, one rotates its writer set; assert HC **does not** declare converged until the rotation propagates and `acl_hash` agrees. Plus zero false positives on steady-state concurrent-rotation/governance scenarios.

## Test
- `cargo test -p calimero-node --features calimero-storage/testing` — pass
- `cargo clippy -p calimero-node --all-targets --features calimero-storage/testing -- -D warnings` — clean
- `cargo fmt --check` — clean

---

> [!NOTE]
> **High Risk**
> This changes load-bearing sync convergence semantics; a false `scope_root` divergence from C0 would cause HC to never report converged until governance catches up—explicitly merge-gated on C0’s e2e canary.
> 
> **Overview**
> **C1a** promotes C0’s observe-only `scope_root` shadow into the real HashComparison end-of-session convergence check (`root_hash_verified` on the #2607 peer re-query path).
> 
> When **both** peers resolve a `scope_root`, that value alone decides convergence (ACL + governance folded onto the entity root). If either side returns `None` (cold projection / non-group), behavior stays the pre-C1 bare entity-root compare—no weaker check than before.
> 
> The old C0-only path that only **logged** `scope_root_shadow_divergence` when entity roots agreed is removed. A new **`scope_root_governance_divergence`** WARN marks the case entities match but governance/`scope_root` does not (HC has nothing to merge on the entity tree; rotation is expected to land via governance sync).
> 
> `docs/design/unified-causal-log-cutover-plan.md` documents **C1a** (this HC verdict) vs **C1b** (LevelWise, snapshot, protocol selector) as incremental slices, merge-gated on C0’s e2e canary.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0178acc34c1928d7a16e1912980f2e75a0e1f7f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/hash_comparison_protocol.rs
+++ b/crates/node/src/sync/hash_comparison_protocol.rs
@@ -699,37 +699,43 @@ async fn run_initiator_impl<T: SyncTransport>(
     let local_root_hash = with_runtime_env(runtime_env.clone(), || {
         get_local_root_hash_for_context(context_id)
     })?;
-    stats.root_hash_verified = local_root_hash == peer_current_root;
 
-    // C0 scope_root shadow (observe-only): when the entity roots AGREE, the bare
-    // `root_hash` declares convergence — but a hash-neutral GOVERNANCE change
-    // (membership add/remove, admin/policy/subgroup edit) leaves identical entities
-    // under a divergent governance plane, which `root_hash` cannot see (governance
-    // lives in a separate DAG, not the storage Merkle). Fold our governance
-    // projection onto the same entity root and compare against the peer's
-    // `scope_root`; a mismatch here, on agreeing entity roots, is exactly the blind
-    // spot C1 will close. (Writer-set rotations are NOT this case — they move
-    // `entities_root` via the rotation-log storage entity, so they're already
-    // caught by the bare `root_hash`.) Logged, never acted on. Both sides skip when
-    // either can't fold the scope (`None`).
-    if local_root_hash == peer_current_root {
-        if let Some(peer_scope_root) = peer_scope_root {
-            if let Some(local_scope_root) =
-                super::helpers::local_scope_root(store, &context_id, local_root_hash)
-            {
-                if local_scope_root != peer_scope_root {
-                    warn!(
-                        marker = "scope_root_shadow_divergence",
-                        %context_id,
-                        entity_root = %hex::encode(&local_root_hash[..8]),
-                        local_scope_root = %hex::encode(&local_scope_root[..8]),
-                        peer_scope_root = %hex::encode(&peer_scope_root[..8]),
-                        "entity roots agree but scope_root differs — hash-neutral \
-                         ACL/membership divergence the current convergence signal hides"
-                    );
-                }
-            }
-        }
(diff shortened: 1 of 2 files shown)


chefsale:
Both 🟡 findings on the `scope_root_governance_divergence` block addressed in commit 0178acc3 (and mirrored to LevelWise in #2901):

- **Dead `unwrap_or_default`**: right — the `matches!` guard already proved both `Some`, so the default branch was unreachable. Replaced the `matches!` + `unwrap_or_default()` with `if let (Some(local_scope_root), Some(peer_scope_root)) = (local_scope_root, peer_scope_root)` and moved the `!root_hash_verified && entities-agree && local != peer` check inside. No dead branch.
- **Copy-type fragility**: same fix resolves it — binding the values up front means the warn no longer re-reads the `Option<[u8; 32]>`s after the verdict match, so it doesn't rely on the hash staying `Copy`.

Behaviour is identical; just clearer control flow. clippy --all-targets -D warnings clean.

PR #2920, opened 2026-06-24
feat(context): C2.2b read-flip — the unified op-store backs the projection

chefsale:
## What

Re-attempt of the C2.2b read-flip: `ops_for_namespace` reads the unified **op-store** (`load_scope_ops`) as the projection's authoritative governance source, falling back to the governance-DAG walk (`collect_namespace_ops`) only for a cold scope (no persisted ops) or a store read error — a transitional safety net that retires with `collect_namespace_ops` at C5.

## Why it's safe now

The first attempt (reverted in #2916) broke convergence: an encrypted group op applied **before its group key arrived** was frozen into the op-store as a `Noop`, so the op-store dropped late-decrypted membership → the joiner's writes were rejected → sync stalled.

That root cause is fixed by **#2918** (the parent of this PR): a key delivery now re-persists the namespace's ops with current decryption (`repersist_namespace_ops`), so the op-store converges to exactly what `collect_namespace_ops` folds — proven deterministically by `tests/op_store_reconstruction.rs` (op-store reconstruction diverges before the re-persist, converges after).

## Also

Reorders the re-persist to run **before** `drain_governance_pending_after_sync` in `recover_missing_group_keys`. The drain re-checks membership through the projection, which now backs onto the op-store, so it must see the corrected ops rather than the stale `Noop`.

## Gate

This is the behavioral change — **e2e-gated**. The joiner-write scenarios that timed out on the first attempt (`group-3node`, `scaffolding-e2e`, `shared-storage` rotation) must now converge. Unit coverage (op-store ≡ governance-DAG reconstruction) is green via the parent PR's regression test.

Stacked on #2918; rebases onto master when it merges.

---

> [!NOTE]
> **Medium Risk**
> Changes the authoritative source for membership folds and sync gates; mitigated by repersist-on-key-delivery, in-log Noop upgrades, and targeted tests, but e2e convergence remains the real gate.
> 
> **Overview**
> Re-enables **C2.2b**: namespace governance for projections is loaded from the unified **op-store** (`load_scope_ops`), with a fallback to the governance-DAG walk only for empty cold scopes or store read errors.
> 
> **Maintained projection fix:** when the same signed op id was first ingested as `Noop` (encrypted op before the group key) and later arrives decrypted, `ingest_op` **upgrades** the op-log entry in place instead of deduping it away, so `member_at_cut` sees late-decrypted membership.
> 
> **Sync after key recovery:** on direct group-key delivery, the node **re-persists** ops, reloads from the op-store, **backfills** the maintained projection, then runs `drain_governance_pending_after_sync` so membership re-checks see corrected ops—not stale `Noop`s.
> 
> Adds a regression test for the maintained-projection backfill path (the gap the fresh-fold test missed).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbc7725d561c1bae7ddcc7020a02a9fa9121519e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -385,33 +385,38 @@ impl ScopeProjections {
         Ok(proj.scope_root_for(scope, [0u8; 32]))
     }
 
-    /// The governance ops to fold for `namespace_id`.
+    /// The governance ops to fold for `namespace_id` (cutover C2.2b read-flip): the
+    /// unified **op-store** is the projection's authoritative backing, so prefer
+    /// [`load_scope_ops`](crate::unified_op_store::load_scope_ops). Falls back to the
+    /// governance-DAG walk ([`collect_namespace_ops`](Self::collect_namespace_ops))
+    /// only when the op-store has no rows for this scope (a cold namespace not yet
+    /// dual-written) or can't be read — a transitional safety net that retires with
+    /// `collect_namespace_ops` at C5.
     ///
-    /// The C2.2b read-flip (make the unified op-store the projection's authoritative
-    /// backing via [`load_scope_ops`](crate::unified_op_store::load_scope_ops)) is
-    /// DEFERRED: flipping the read onto the op-store broke convergence in the
-    /// maintainer's e2e — group-3node / scaffolding-e2e / shared-storage rotation all
-    /// timed out on the joiner's post-write sync.
-    ///
-    /// ROOT CAUSE (pinned deterministically by `tests/op_store_reconstruction.rs`):
-    /// an encrypted group `MemberAdded` can apply to the governance DAG BEFORE its
-    /// group key arrives (keys lag on a separate delivery path; see the
-    /// buffer-awaiting-key replay in `apply_group_op_inner`). The apply-time
-    /// dual-write then decrypts with no key, so the op is frozen into the op-store as
-    /// a `Noop`. `collect_namespace_ops` RE-DECRYPTS the same op at read time once the
-    /// key lands and recovers the real `MemberAdded`, but the op-store still holds the
-    /// `Noop` — so reading membership from the op-store drops the member, the joiner's
-    /// writes are rejected, and sync never converges. `scope_root` doesn't surface it
-    /// because at shadow time neither side has decrypted yet (both `Noop` → roots
-    /// match); the divergence only appears after the key arrives.
-    ///
-    /// Until the op-store faithfully reconstructs late-decrypted membership (e.g.
-    /// re-persist on the key-arrival replay, or re-decrypt on load) this stays on the
-    /// governance DAG, the proven-correct source. The op-store keeps being
-    /// dual-written.
+    /// The earlier flip attempt broke convergence because an encrypted group op
+    /// applied BEFORE its group key arrived was frozen into the op-store as a `Noop`
+    /// (the apply-time dual-write decrypts with whatever key is present then), so the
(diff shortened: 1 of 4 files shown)


chefsale:
## e2e red — fix was necessary but not sufficient. Marking draft.

group-3node + scaffolding-e2e still time out on the joiner's write. Root cause found, and it's a second wrinkle beyond the decryption-freeze:

**The maintained projection's op-log is sticky.** `ingest_op` always folds `states` but the op-log (`logs`) **dedups by op id** (scope_projection.rs:307), and `member_at_cut` → `acl_view_at` folds the **op-log**. So when an encrypted op is first applied without its key (folded as `Noop`), re-ingesting the corrected same-id op later updates `states` but **not the op-log the auth view reads** — and nothing clears `seen`/`logs` to force a re-fold.

#2918 corrected the op-*store*, but the maintained projection stays stuck on the `Noop`. Pre-flip passed because `collect_namespace_ops` re-decrypts on *every* read, so the projection's first fold already had current decryption; `load_scope_ops` returns the frozen-at-apply snapshot.

My deterministic repro missed this because it used *fresh* projections (op-log folds the corrected op on first touch); production reuses the maintained projection that already folded the `Noop`.

**Next:** (1) extend the repro to use a *maintained* projection (fold Noop → repersist → assert recovery) so it reproduces the e2e failure deterministically; (2) fix = invalidate the scope's `states`/`seen`/`logs`/`backfilled` on key delivery so the next backfill re-folds from the corrected op-store. Holding the flip until that repro is green.

chefsale:
## v2 (`cbc7725d`): second wrinkle fixed, now with a production-faithful repro

The maintained projection's op-log stayed stuck on the `Noop` even after #2918 corrected the op-store. Root cause: `op.id` is the signed-op hash, not the decoded payload's, so the `Noop` and decrypted forms share an id; `ingest_op` dedups the op-log by id and `member_at_cut` folds that log, so the corrected op was dropped as a duplicate.

**Fix:** `ingest_op` upgrades a previously-`Noop` op-log entry in place when a non-`Noop` payload arrives for the same id (never downgrades). `recover_missing_group_keys` re-ingests the corrected ops into the maintained projection after re-persisting the op-store.

**Methodology fix (the important part):** added `maintained_projection_recovers_late_decrypted_membership_on_backfill`, which folds the op as `Noop` in a *maintained* projection first — exactly what production does — then asserts recovery. I verified it **FAILS without the ingest_op fix** (reproduces the e2e timeout) and passes with it. The earlier repro used *fresh* projections, which is why it was green while e2e was red. This one matches production.

Still draft — not marking ready until the e2e actually comes back green. Please re-run e2e when you have a cycle.

chefsale:
Closing: re-sequencing the cutover (decision B). 

Three e2e cycles surfaced three distinct gaps — op-store froze Noop pre-key (#2918), maintained-projection op-log stuck on Noop, and locally-authored governance ops never reaching the op-store — all the same root issue: the op-store is a *partial shadow* of the gov-DAG, and the dual-write keeps missing paths/timing cases.

Rather than chase every dual-write site, we'll make the op-store authoritative only **after the apply is unified onto it** (C3/C4), so it's complete by construction. The read-flip stops being a separate risky step. #2918 stays on master as a harmless correctness improvement (keeps the op-store fresh; nothing reads it yet). The sticky-log `ingest_op` upgrade from this branch is a real latent fix worth extracting standalone later.

Branch `feat/c2.2b-flip-retry` retained for reference.

PR #2935, opened 2026-06-25
refactor(sync): P6.S1 — unify the scope_root convergence verdict

chefsale:
## What

First slice of **P6** (the sync-engine collapse). The cross-plane convergence verdict — `scope_root` is authoritative, falling back to the bare entity-root compare when a scope can't be folded (cold projection / non-group context) — was **open-coded identically** in `hash_comparison_protocol.rs` and `level_sync.rs`.

This extracts it into one `helpers::scope_verdict(local_scope_root, peer_scope_root, local_entity_root, peer_entity_root) -> ScopeVerdict { Converged | GovDiverged | DataDiverged }`. Both protocols now decide against the single function, and the later P6 stages (route gov-plane and data-plane pulls off the verdict) build on it.

## Pure refactor

No behavioral change: both sites compute the same `root_hash_verified` and emit the same `scope_root_governance_divergence` marker as before — the logic is byte-for-byte the prior match, just shared. Unit-tested (4 cases): equal→`Converged`, scope-differs-but-entities-agree→`GovDiverged`, both-differ→`DataDiverged`, cold-projection→entity-root fallback.

Clippy clean. No wire/op-id/protocol change.

---

> [!NOTE]
> **Medium Risk**
> Changes touch all group-op authoring sites and sync convergence decisions; behavior is intended to be equivalent, but any gap in atomic op-store writes would reintroduce the local-author op-store incompleteness this PR removes mirrors for.
> 
> **Overview**
> **P6.S1** centralizes cross-plane sync convergence in `helpers::scope_verdict` → `ScopeVerdict` (`Converged` / `GovDiverged` / `DataDiverged`). **Hash comparison** and **level sync** now call that helper for `root_hash_verified` and the `scope_root_governance_divergence` warning instead of duplicating the same `match` logic.
> 
> **C3 op-store cutover cleanup** removes belt-and-suspenders mirroring now that governance writes land in the unified op-store atomically with the gov-DAG: the context crate drops `sign_apply_and_publish_group_op` and every handler/admin path calls `calimero_governance_store::sign_apply_and_publish` directly. Removed are post-author `ScopeProjections::persist_namespace_head_ops` calls, the receive-path `unified_op_store::persist_op` dual-write in `apply_signed_namespace_op`, and the `persist_namespace_head_ops` API (tests now assert atomic persistence via `store_signed_operation` / idempotent `repersist_namespace_ops`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b701580a08587414760c5db987fe4a4ee323094. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/handlers/add_group_members.rs
+++ b/crates/context/src/handlers/add_group_members.rs
@@ -37,7 +37,7 @@ impl Handler<AddGroupMembersRequest> for ContextManager {
         ActorResponse::r#async(
             async move {
                 for (identity, role) in &members {
-                    let report = crate::sign_apply_and_publish_group_op(
+                    let report = calimero_governance_store::sign_apply_and_publish(
                         &datastore,
                         &node_client,
                         &ack_router,
--- a/crates/context/src/handlers/admit_tee_node.rs
+++ b/crates/context/src/handlers/admit_tee_node.rs
@@ -199,7 +199,7 @@ impl Handler<AdmitTeeNodeRequest> for ContextManager {
                     PrivateKey::from(effective_signing_key.ok_or_else(|| {
                         eyre::eyre!("no signing key available for TEE admission")
                     })?);
-                let report = crate::sign_apply_and_publish_group_op(
+                let report = calimero_governance_store::sign_apply_and_publish(
                     &datastore,
                     &node_client,
                     &ack_router,
--- a/crates/context/src/handlers/apply_signed_namespace_op.rs
+++ b/crates/context/src/handlers/apply_signed_namespace_op.rs
@@ -86,18 +86,6 @@ impl Handler<ApplySignedNamespaceOpRequest> for ContextManager {
                         &delta_parents,
                     );
 
-                    // C2.1b dual-write: persist this governance op to the durable
-                    // unified op-store, keyed by its (namespace) scope. Observe-only —
-                    // nothing reads it yet (the per-plane C2 flips switch reads onto
-                    // it). A failure must never affect the governance apply, so it is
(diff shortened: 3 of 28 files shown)


chefsale:
Addressed the MeroReviewer findings in 6b701580:

- **🟡 GovDiverged re-destructures already-matched Options (consensus 1/1):** the variant now carries the resolved roots — `GovDiverged([u8; 32], [u8; 32])`. `scope_verdict` populates them in the match arm, and both call sites (`hash_comparison_protocol`, `level_sync`) pattern-match the verdict directly (`if let ScopeVerdict::GovDiverged(local, peer) = verdict`). The warning can no longer be silently dropped — the invariant is encoded in the type.
- **💡 Asymmetric-None fallback intent:** added an inline comment on the `_ if local_entity_root == peer_entity_root` arm explaining that a single cold side intentionally falls back to the bare entity-root compare (pre-C1 behaviour) rather than reporting governance divergence, so a future reader doesn't 'fix' it into a false alarm on a partially-warmed node.
- **💡 Parameter naming (`peer_entity_root` vs call-site `peer_current_root`):** left as-is — the parameter name describes its role (the peer's bare entity root) and the doc on `scope_verdict` states this; renaming would only move the cosmetic mismatch to the other side. No logic concern.

fmt + clippy clean; the 4 `scope_verdict` unit tests updated for the new variant payload and green. Pure refactor still — same verdict, same marker.

Issue #2848, opened 2026-06-22
Restricted-subgroup replication lost on (re-)admission: buffered ContextRegistered not re-driven after GroupCreated (post-KeyDelivery retry fails 'group not found for state hash computation')


xilosada:
## Summary

On a node that is admitted to (or re-admitted to) a Restricted subgroup it has not yet fully synced, the buffered encrypted `ContextRegistered` op can permanently fail to apply. The node receives the group key, the post-`KeyDelivery` retry of the buffered op fails with `group not found for state hash computation` (the subgroup's own `GroupCreated`/group-state isn't applied yet), and **nothing re-drives that op once the group state later lands**. The node therefore never learns the context exists → never `join_context`/subscribes → never replicates it. The owner is left publishing to a context with "no peers to sync."

It is **non-deterministic**: of N subgroups admitted in the same batch, some recover (a later namespace-sync replay happens to re-apply their `ContextRegistered` after `GroupCreated`) and some never do. Which ones lose is sync-ordering luck.

## Evidence (real fleet node)

Owner re-enabled HA, re-admitting a `ReadOnlyTee` to a namespace with 4 Restricted subgroups (Private1–4). All four received the group key; all four logged the same retry failure:

```
INFO  namespace::governance: received group key via direct delivery group_id=b924b944… key_id=d9e17e38…
WARN  namespace::governance: failed to retry encrypted op after KeyDelivery group_id=b924b944…
      error=group not found for state hash computation
```

Outcome on the TEE node (`meroctl context ls` + journal), per Restricted subgroup:

| subgroup | context | ContextRegistered | join_context | subscribed | syncs |
|---|---|---|---|---|---|
| Private1 | `4fU8Zv…` | yes | yes | yes | 1015 |
| Private4 | `3Buisw…` | yes | yes | yes | 234 |
| Private2 | `2YccYaid…` | **none** | **none** | **none** | **0** |
| Private3 | `GSQWEsJZ…` | **none** | **none** | **none** | **0** |

Private2/Private3 show *zero* trace of their contexts on the admitted node — it never registered/joined them. The owner's sync driver loops `no peers to sync with this round` for `2YccYaid`/`GSQWEsJZ`. State is durable (still missing hours later), not transient.

## Root cause

- `ContextRegistered` is an encrypted `GroupOp`; on a node missing the group key it is buffered and applied later.
[...]

PR #2939, opened 2026-06-25
fix(governance): C5.S3a — state_hash mismatch warns, never bails (local group apply)

chefsale:
## What

First sub-slice of **C5.S3** (retire op-level `state_hash` as an apply gate). One-site change: the local group-op apply path now **warns** on a `state_hash` mismatch instead of hard-`bail!`ing.

## Why

`apply_local_signed_group_op_at_cut` (`crates/governance-store/src/lib.rs:1290`) is the `DeltaApplier::apply` body for the **group governance DAG**, so concurrent same-subgroup ops from different nodes routinely land here — A and B each sign against their then-current view, and the second to reach C has already drifted. `bail!`ing on that mismatch **rejects a legitimate concurrent sibling the DAG would otherwise merge**, forcing recovery onto anchor-sync reconcile — a real multi-node convergence hazard.

The namespace-wrapped sibling path (`apply_group_op_inner`, `namespace/governance.rs:1452`) **already** downgraded this to warn-and-apply (the PR #2500 caveat), and its comment explicitly calls out the local path as the lone surviving hard-bail. This brings the local path into that same symmetry: now that `scope_root` is the authoritative convergence signal and CRDT merge handles concurrency, the recomputed `state_hash` is **staleness telemetry, not an apply gate**.

Signature verification and the **nonce window** remain the real safety / anti-replay gates; the absent-meta bypass (#2848) is unchanged.

## Scope

- **No wire or schema change.** The `state_hash` field stays on `SignableGroupOp` and in the content hash, so **op ids are stable** — no re-bootstrap. The flag-day field removal + `SCHEMA_VERSION` bump is the separate later **C5.S3b**.
- Replaces the now-obsolete `state_hash_prevents_concurrent_op_divergence` test (which pinned the reject semantics this PR removes) with `state_hash_mismatch_does_not_reject_concurrent_ops`: two admins concurrently add *independent* members against the same genesis state; both nodes converge to the identical member set regardless of apply order (`compute_state_hash` sorts members, so the converged hashes match). Independent adds avoid the authorization-order confound of the old admin-removes-admin scenario.
- fmt + clippy (`-D warnings`) clean; the 3 `state_hash_*` convergence tests pass.

---

> [!NOTE]
> **Medium Risk**
> Changes group governance DAG apply behavior for concurrent ops (previously rejected ops now mutate state), which affects multi-node convergence but keeps signature and nonce gates; scope is localized to the local apply path with no wire format change.
> 
> **Overview**
> **C5.S3a** aligns the **local** signed group-op apply path with the namespace path: a signed `state_hash` that no longer matches the store is **staleness telemetry** (warn + continue), not a hard reject.
> 
> In `apply_local_signed_group_op_at_cut`, mismatches no longer `bail!`; they emit a structured `warn` (including the op variant) and the op still applies. **Anti-replay runs first**: the per-signer nonce window is checked before any `state_hash` logic so replays return early without a misleading “applying anyway” warning. Signature verification and the absent-meta bypass are unchanged; no wire/schema change.
> 
> The integration test is renamed and rewritten: concurrent independent `MemberAdded` ops from two admins against the same genesis state must apply on both nodes in either order, converge on membership and `compute_state_hash`, and replay via nonce dedup without duplicate rows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd5bcd1a0983c1d2845e0f19d525051049113fea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/tests/local_group_governance_convergence.rs
+++ b/crates/context/tests/local_group_governance_convergence.rs
@@ -1467,6 +1467,8 @@ fn state_hash_mismatch_does_not_reject_concurrent_ops() {
         },
     )
     .unwrap();
+    // nonce is per-signer: admin_c's nonce=1 does not collide with admin_a's
+    // nonce=1 — each signer has its own independent nonce window.
     let op_c = SignedGroupOp::sign(
         &admin_c_sk,
         gid_bytes,
@@ -1532,6 +1534,22 @@ fn state_hash_mismatch_does_not_reject_concurrent_ops() {
     assert_eq!(
         final_b, final_c,
         "nodes converge to identical state regardless of apply order"
+    );
+
+    // Replay safety is unaffected by dropping the state_hash gate: the per-signer
+    // NONCE WINDOW (not state_hash) is the anti-replay guard, and it is unchanged.
+    // Re-applying op_a (admin_a, nonce 1) is a no-op — the nonce is already in the
+    // window — so a DAG replay of a seen op cannot double-apply its mutation.
+    assert!(
+        apply_local_signed_group_op(&node_b, &op_a).is_ok(),
+        "replaying op_a is accepted (deduped, not errored)"
+    );
+    let after_replay_b = MetaRepository::new(&node_b)
+        .compute_state_hash(&gid)
+        .unwrap();
+    assert_eq!(
+        final_b, after_replay_b,
+        "replaying a seen op is a no-op — nonce window dedups it, state is unchanged"
     );
 }
 
--- a/crates/context/tests/local_group_governance_convergence.rs
+++ b/crates/context/tests/local_group_governance_convergence.rs
@@ -1540,6 +1540,10 @@ fn state_hash_mismatch_does_not_reject_concurrent_ops() {
     // NONCE WINDOW (not state_hash) is the anti-replay guard, and it is unchanged.
     // Re-applying op_a (admin_a, nonce 1) is a no-op — the nonce is already in the
     // window — so a DAG replay of a seen op cannot double-apply its mutation.
+    let members_before_replay = MembershipRepository::new(&node_b)
+        .list(&gid, 0, usize::MAX)
+        .unwrap()
+        .len();
     assert!(
         apply_local_signed_group_op(&node_b, &op_a).is_ok(),
         "replaying op_a is accepted (deduped, not errored)"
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed in 8615a51a:

**🟡 Replay safety / idempotency contract:** the key point is that **the per-signer nonce window — not `state_hash` — is the anti-replay guard, and this PR doesn't touch it.** When an op is replayed through the DAG, the nonce-window check (`if nonce_window.contains(op.nonce) { return Ok(()) }`, lib.rs:1318) dedups it *before* any mutation runs, so a non-idempotent handler can't double-apply on replay regardless of the `state_hash` decision. The `state_hash` gate never protected against replay — it only rejected *concurrent* (distinct-nonce) ops, which is exactly the multi-node convergence hazard this PR removes (and which the namespace path already removed). What the warn-and-apply widens is the set of *concurrent* ops that apply, and that relies on **commutativity**, which is the cutover's standing CRDT premise (already in force on `apply_group_op_inner`), not idempotency-on-replay.

I made this concrete rather than just asserting it: the test now re-applies a seen op and asserts it's a **deduped no-op** (state hash unchanged), demonstrating the nonce window still blocks double-application. I left the per-arm `INVARIANT` comments out — the real guarantee lives in the nonce window, not in per-handler idempotency, so pinning it at the dedup point is the accurate place.

**📝 Nit (op_c nonce=1):** added the per-signer-nonce comment above `op_c`.

Test green; fmt clean.

chefsale:
Addressed the second review round in fd5bcd1a:

- **🟡 state_hash warn fires before nonce dedup:** reordered — the per-signer **nonce-window dedup now runs first**, so a replayed op returns early and never reaches the staleness check, suppressing the misleading 'applying anyway' warn on pure replays. This matches the ordering `apply_group_op_inner` already documents.
- **🟡 warn lacks the op variant:** added `op = ?op.op` to the mismatch warn so a benign concurrent `MemberAdded` is distinguishable from a stale-hash op on a sensitive variant like `MemberRemoved`.
- **🟡 replay test only checked the hash:** tightened it to also assert the **member-set size is unchanged** across the replay — pins that the dedup returns before `apply_group_op_mutations` so no duplicate member row can escape even via a hash-neutral side-effect.
- **💡 mark removal target:** added an explicit `C5.S3b: remove this entire block once state_hash is dropped from the wire format` comment on the block.

clippy (`-D warnings`) clean; the 3 `state_hash_*` tests green.

PR #2941, opened 2026-06-25
feat(sync): P6.S3 — catch governance divergence pre-sync when entities agree

chefsale:
## What

First slice of **P6.S3** (route the data-plane/selection through the verdict). Closes the last hole in scope_root-driven governance convergence.

`select_protocol` decides on the **entity plane only** (root_hash + dag_heads), so when the entity Merkle already matches it returns `None` — "nothing to sync". But governance lives **outside** the entity Merkle, so a pure governance-plane divergence (a rotation / ACL change) is *exactly* the entities-agree case: `root_hash` equal, `scope_root` differs. The post-sync P6.S2 hook only fires when HC/LevelWise actually run — which a `None` selection skips — so that divergence went uncorrected tick after tick.

## How

- `query_peer_dag_state` now returns the peer's `scope_root` (already on `DagHeadsResponse`, previously discarded as `scope_root: _`).
- At the selection point, when `select_protocol` says `None`, compute the authoritative `scope_verdict`. On a pre-sync `GovDiverged` (entities equal, scope_root differs) → pull the namespace governance DAG from the peer (reusing the P6.S2 `pull_namespace_governance` path, refactored to an inherent method) and skip the pointless entity walk.
- **Cold / non-group projections** resolve `scope_root == None` and fall back to the entity-root compare inside `scope_verdict`, so a partially-warmed node never raises a false governance divergence.

Complements P6.S2: **pre-sync** handles "entities already agree", **post-sync** handles "entities diverge then converge". Logs the same `gov_divergence_pull_triggered` marker, so the existing informational e2e report step covers both.

## Scope / testing
- No wire change (`scope_root` was already on the response); no change to `select_protocol` or the HC/LevelWise/Snapshot decision tree; tombstone / frozen-leaf / authorization paths untouched (they live inside the entity walk, which this only *skips* when there's nothing to walk).
- fmt + clippy (`-D warnings`) clean; 39 manager sync unit tests pass. The pre-sync pull path is exercised by the partition e2e suites (same Stream-fixture gap as the selector keeps it out of unit tests); the verdict itself is unit-tested.

---

> [!NOTE]
> **Medium Risk**
> Changes sync orchestration on the common “already in sync” path with store reads and governance backfill when scope roots disagree; logic is gated on both sides resolving `scope_root` to avoid false positives on cold projections.
> 
> **Overview**
> **P6.S3** closes a gap where periodic sync treated “entity roots match” as fully in sync even when **governance** (`scope_root`) still diverged—because `select_protocol` only looks at entity `root_hash` and DAG heads and returns `SyncProtocol::None`, skipping HashComparison/LevelWise and the post-sync P6.S2 governance pull.
> 
> `query_peer_dag_state` now surfaces the peer’s **`scope_root`** from `DagHeadsResponse` (already on the wire; previously ignored). When protocol selection is **`None`** and the peer sent a resolved `scope_root`, the manager runs **`scope_verdict`**. On **`GovDiverged`** (entities equal, scope roots differ), it **pulls the namespace governance DAG** from that peer via the existing P6.S2 `pull_namespace_governance` path (refactored to an inherent `SyncManager` method) and returns without an entity walk. Peers with **`scope_root == None`** skip the store read on the hot path so cold/non-group nodes don’t false-trigger.
> 
> Logs **`gov_divergence_pull_triggered`** / **`gov_divergence_pull_complete`** for operators and e2e, complementing P6.S2’s post-sync hook when entities actually diverge first.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55ea43b23611ef6b63c3541f1eaa704387eaa4a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/manager/mod.rs
+++ b/crates/node/src/sync/manager/mod.rs
@@ -1233,7 +1233,9 @@ impl SyncManager {
                 .await?;
 
             match peer_state {
-                Some((peer_root_hash, _peer_dag_heads)) if *peer_root_hash != [0; 32] => {
+                Some((peer_root_hash, _peer_dag_heads, _peer_scope_root))
+                    if *peer_root_hash != [0; 32] =>
+                {
                     // Peer has state - use snapshot sync for efficient bootstrap
                     info!(
                         %context_id,
@@ -1423,7 +1425,7 @@ impl SyncManager {
             .query_peer_dag_state(context_id, chosen_peer, our_identity, stream)
             .await?;
 
-        if let Some((peer_root_hash, peer_dag_heads)) = peer_state {
+        if let Some((peer_root_hash, peer_dag_heads, peer_scope_root)) = peer_state {
             // Build handshakes for protocol selection (CIP §2.3)
             // Uses shared functions from calimero_node_primitives::sync::state_machine
             let local_hs = self.build_local_handshake(context);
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed in 55ea43b2:

- **💡 Datastore handle in the hot no-op path:** short-circuited on `peer_scope_root` first — `scope_verdict` only returns `GovDiverged` when *both* sides resolved a scope_root, so when the peer's is `None` (cold / non-group) the verdict can only fall back to the entity-root compare, which already agreed here (`selection == None`). The store handle is now acquired only when the peer actually carries a scope_root.
- **💡 `Ok(None)` indistinguishable from 'already in sync':** added a `gov_divergence_pull_complete` completion marker after the pull so the trigger correlates with its outcome (pairs with `gov_divergence_pull_triggered`).
- **📝 Explicit `SyncManager::` path in the trait forwarder:** kept as-is — the sibling forwarders in the same `impl ProtocolDispatch` block (`request_dag_heads_and_sync`, `fallback_to_snapshot_sync`, `open_stream`) all use the explicit `SyncManager::method(self, …)` form, so this matches the established convention in that block.

clippy (`-D warnings`) clean.

chefsale:
Follow-up: the later-round findings here are addressed in #2944 (P6.S3), which builds directly on this:
- **🟡 gov pull result silently discarded** — `sync_namespace_from_peer` / `pull_namespace_governance` now return the ops-received count, and both `gov_divergence_pull_complete` markers log `ops_pulled` (0 = empty/best-effort-failed pull, no longer indistinguishable from a no-op sync).
- **🟡 ephemeral scope_root fold cost** — deferred deliberately: the ephemeral fold has a `cut_ancestry_complete` guard that the maintained projection's raw `scope_root_for` lacks, so swapping to the maintained read is a separate correctness-sensitive slice (naive swap risks false divergence on a partially-warmed node). Noted as a follow-up in #2944.
- **💡 `resolve_namespace_id` sharing** — trivial; the two callers differ in peer-selection policy, left as-is.

#2944 also closes the broader gap this PR's per-protocol hook had: the post-sync governance pull now fires for **all** data backends (Snapshot/DeltaSync too), not just HC/LevelWise.

PR #2944, opened 2026-06-25
feat(sync): P6.S3 — centralise post-sync governance reconciliation for all backends

chefsale:
## What

Completes **P6.S3**, building on #2941 (P6.S2/S3a, merged). Closes a real gap and centralises the governance-divergence control to a single place.

### The gap
The post-sync governance pull was bolted onto only the two data backends that compute the `scope_root` verdict internally — **HashComparison** and **LevelWise** — via a per-protocol `gov_divergence_detected` stat plus a hook in each `ProtocolSelector::execute` arm. The other backends (**Snapshot, DeltaSync, BloomFilter, SubtreePrefetch**) never checked the verdict. So a node that converged its entities via a snapshot/delta sync while its governance plane still differed would **never** pull the missing rotation/ACL op.

### The fix
Move the post-sync check to **one place in the manager** (`handle_dag_sync`, right after `execute` returns), so it covers **every** backend uniformly: recompute the verdict against the peer's handshake `scope_root` using the **post-sync** local root (entities may have merged); on `GovDiverged`, pull governance.

The selector and the two protocols go back to pure data-transfer + observability:
- removed `gov_divergence_detected` from `HashComparisonStats` / `LevelWiseStats` and the two pull hooks in `ProtocolSelector::execute`;
- removed the now-dead `ProtocolDispatch::pull_namespace_governance` trait method (the manager calls the inherent method directly);
- HC / LevelWise keep computing the verdict for `root_hash_verified` and the `scope_root_governance_divergence` warn (unchanged observability).

This makes the architecture explicit: **one control protocol (the scope_root verdict) with two transfer backends (governance pull + entity data sync)**.

## Folds in an unaddressed #2941 review finding
`sync_namespace_from_peer` / `pull_namespace_governance` now **return the number of governance ops** the backfill delivered, and both `gov_divergence_pull_complete` markers log `ops_pulled`. A pull that came back empty (best-effort failure or already-converged peer) is no longer indistinguishable from a normal no-op sync (the reviewer's "silently discarded result" point).

Two #2941 findings remain explicit follow-ups, noted here rather than rushed:
- **Ephemeral `scope_root` fold cost** — the fold has a `cut_ancestry_complete` guard the maintained projection's raw `scope_root_for` lacks, so swapping to the maintained projection is a separate, correctness-sensitive slice (it would risk false divergence on a partially-warmed node).
- **`resolve_namespace_id` sharing** with `backfill_governance_for_pending_deltas` — trivial; the two paths differ in peer-selection policy.

## Scope / testing
- No wire/op-id change. The HC-internal re-queried-root precision (for `root_hash_verified`) is untouched; the centralised gov check uses the handshake `scope_root` as its reference (the next session re-checks, so a peer changing governance mid-sync self-corrects).
- fmt + clippy (`-D warnings`) clean; 239 node sync unit tests pass. The post-sync pull path is exercised by the partition e2e suites; same `gov_divergence_pull_triggered` / `_complete` markers feed the existing informational e2e report.

---

> [!NOTE]
> **Medium Risk**
> Changes sync convergence behavior for Snapshot/DeltaSync and broadens when governance pulls run; logic is intentionally conservative (idempotent pulls) but touches authorization-adjacent `scope_root` handling in the sync manager hot path.
> 
> **Overview**
> **Governance reconciliation** no longer runs inside HashComparison / LevelWise only. The manager now owns **pre-sync** (`GovDiverged` when protocol is `None`) and **post-sync** (after any successful `execute`) checks that pull namespace governance when `scope_root` still disagrees with the peer—using a direct inequality on folded `scope_root`, not per-protocol stats.
> 
> **Removed** `gov_divergence_detected` from HC/LevelWise stats, the `ProtocolDispatch::pull_namespace_governance` hook, and selector-side pulls after those protocols. HC/LevelWise still compute `scope_verdict` for `root_hash_verified` and `scope_root_governance_divergence` warnings only.
> 
> **Observability:** `sync_namespace_from_peer` / `pull_namespace_governance` return **op counts**; `gov_divergence_pull_complete` logs `ops_pulled` so empty backfills are visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8787be68184999bea33a4730677de885e4add943. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/node/src/sync/hash_comparison_protocol.rs
+++ b/crates/node/src/sync/hash_comparison_protocol.rs
@@ -142,16 +142,6 @@ pub struct HashComparisonStats {
     /// merge did not converge the two peers — see #2407 for the
     /// failure mode this guards against.
     pub root_hash_verified: bool,
-    /// Whether the convergence verdict was `GovDiverged` — the entity
-    /// roots agreed but the authoritative `scope_root` differed, i.e. a
-    /// pure ACL/governance-plane divergence that the entity Merkle walk
-    /// can't reconcile (governance lives outside the storage Merkle).
-    /// The caller (`ProtocolSelector`) reacts by pulling the namespace
-    /// governance DAG from the same peer (P6.S2) — without it, a node
-    /// with no stuck deltas would otherwise just warn and never converge
-    /// the governance plane. Distinct from `root_hash_verified == false`,
-    /// which also covers data-plane divergence.
-    pub gov_divergence_detected: bool,
     /// Root-state byte blobs the DFS encountered on remote leaves
     /// that the host can't merge by itself (separate-address-space
     /// merge registry — see [`crate::sync::helpers::apply_leaf_with_crdt_merge`]).
@@ -729,16 +719,13 @@ async fn run_initiator_impl<T: SyncTransport>(
         peer_current_root,
     );
     stats.root_hash_verified = verdict.converged();
-    stats.gov_divergence_detected =
-        matches!(verdict, super::helpers::ScopeVerdict::GovDiverged(..));
 
     // Surface the case the entity root hides: entities AGREE but `scope_root` differs
     // ⇒ the divergence is purely in the ACL/governance plane. HC's entity tree-walk
     // has nothing to reconcile here (the entities already match) — governance ops
-    // live outside the storage Merkle. The selector reacts to
-    // `gov_divergence_detected` by pulling the namespace governance DAG from this
-    // same peer (P6.S2); distinct marker so a governance-plane divergence is legible
-    // apart from a data-plane one.
+    // live outside the storage Merkle. This is observability only; the corrective
+    // governance pull is centralised post-sync in the manager (P6.S3), so it covers
+    // Snapshot / DeltaSync syncs too, not just HC / LevelWise.
     if let super::helpers::ScopeVerdict::GovDiverged(local_scope_root, peer_scope_root) = verdict {
         warn!(
             marker = "scope_root_governance_divergence",
--- a/crates/node/src/sync/level_sync.rs
+++ b/crates/node/src/sync/level_sync.rs
@@ -145,11 +145,6 @@ pub struct LevelWiseStats {
     pub requests_sent: u64,
     /// Whether final root hash was verified against expected.
     pub root_hash_verified: bool,
-    /// Whether the convergence verdict was `GovDiverged` (entity roots
-    /// agreed, authoritative `scope_root` differed). Same meaning and
-    /// caller reaction as [`crate::sync::hash_comparison_protocol::HashComparisonStats::gov_divergence_detected`]:
-    /// the selector pulls the namespace governance DAG from this peer (P6.S2).
-    pub gov_divergence_detected: bool,
     /// Whether parent_ids were truncated due to `MAX_PARENTS_PER_REQUEST`.
     ///
     /// If true, the sync may be incomplete and a follow-up sync might be needed.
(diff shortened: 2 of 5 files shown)


chefsale:
Addressed both findings in 40098935:

- **🟡 (cursor, Medium) Stale handshake root skips pull:** correct — the post-sync check compared the refreshed local root against the peer's *handshake* entity root, and after HC/LevelWise the peer can advance during the sync, so an entity-root compare could misclassify a real governance divergence as `DataDiverged` and **skip** the pull (a permanent governance stall). Fixed by dropping the GovDiverged-vs-DataDiverged sub-classification at the manager and pulling on a **direct `local_scope_root != peer_scope_root` inequality**. A pull is idempotent (0 ops when governance is already in sync), so this over-pulls at worst during a still-converging data plane — harmless and self-limiting — and never misses a real divergence (the safe direction). HC/LevelWise keep their precise in-protocol verdict for `root_hash_verified` + the warn.
- **🟡 (meroreviewer) Silent stale-root fallback on `get_context` failure:** now matched explicitly — on `Ok(None)` (context vanished) / `Err` (store fault), warn and **skip** the check rather than folding a verdict against a stale pre-sync root; the next tick re-checks against fresh state. A cold/incomplete local fold (`None`) still abstains.

clippy (`-D warnings`) clean.

chefsale:
Latest round:

- **🟡 Post-sync block runs on the `Ok(None)` no-op path** — fixed in 8787be68: gated on `matches!(exec_result, Ok(Some(_)))`, so the post-sync check only runs when a data backend actually executed. The `SyncProtocol::None` case (entities already in sync) is handled by the pre-sync check and no longer triggers a redundant store read/fold each converged tick.

- **💡 Use `scope_verdict` + GovDiverged instead of the direct inequality** — respectfully declining, because this **directly conflicts with the cursor Medium finding** I just fixed in 40098935. `scope_verdict` classifies GovDiverged-vs-DataDiverged by comparing entity roots, and the manager only holds the peer's **handshake** root. After HC/LevelWise the peer can advance, so that compare can misclassify a *real* governance divergence as `DataDiverged` and **skip the pull** → a permanent governance stall. The direct `scope_root` inequality is the deliberate safe direction: it can over-pull at worst during a still-converging data plane (idempotent, 0 ops, self-limiting), but it never misses a real divergence. HC/LevelWise keep `scope_verdict` internally where they DO re-query the peer's current root. I'll keep the pre-sync path on `scope_verdict` (there the handshake root *is* current — no sync has run yet) and the post-sync path on the direct inequality; the asymmetry is intentional and now commented.

- The two earlier comments on commit 44591789 (silent `get_context` fallback + stale handshake root) were already addressed in 40098935 (explicit match + warn/skip on store fault; direct inequality).

clippy clean.

chefsale:
Status on the latest round:

- The three 🟡 warnings (post-sync stale handshake root @1580, `Ok(None)` guard @1601, direct-inequality-vs-`scope_verdict` @1601) are all on the **earlier commits** (44591789 / 40098935) and are already resolved: the stale-root + silent-fallback in 40098935, the `Ok(None)`-guard → `matches!(exec_result, Ok(Some(_)))` in 8787be68, and the direct-inequality is the deliberate choice over `scope_verdict` (declined-with-reason above — using the handshake root in `scope_verdict` is exactly the misclassification cursor flagged).

- 💡 **Pre-sync pull returns `Ok(None)` even on a 0-op pull** (the one finding on HEAD `8787be68`): intentional and I'd keep it. The pull is best-effort with next-tick re-check, and `handle_dag_sync`'s return type is `Option<SyncProtocol>` — a governance backfill isn't a `SyncProtocol`, so there's no faithful way to encode "data in sync, governance pull partial" in it. The observability the reviewer wants already exists: `gov_divergence_pull_complete` logs `ops_pulled`, so a peer that's persistently GovDiverged but can't serve the op shows up as a repeating `ops_pulled=0` for that `context_id` (and the next tick may pick a *different* peer that can serve it). Forcing an `Err` here would wrongly fail an otherwise-successful data sync.

Net: no further code change — all actionable findings are in. clippy clean, 239 sync tests green.

PR #2946, opened 2026-06-25
feat(governance)!: C5.S3b — remove op-level state_hash field + bump schema (flag-day)

chefsale:
## ⚠️ FLAG-DAY / re-bootstrap — no back-compat

Removing `state_hash` changes **every op's content hash (its id)** *and* bumps the wire schema, so **every node must be on this build** and pre-existing governance DAGs are not deserializable. This is the deliberate closing move of the unified-causal-log cutover.

## What

The op-level `state_hash` field stopped being an apply gate in **C5.S3a** (#2939) — `scope_root` is the authoritative cross-plane convergence signal now — leaving it pure dead weight in the signed op bytes. This removes it entirely:

- drop `state_hash` from all four op structs — `SignableGroupOp` / `SignedGroupOp` and `SignableNamespaceOp` / `SignedNamespaceOp` — and from `sign()` / `to_signable()`. Since the field was inside the signed/hashed bytes, every op id changes;
- bump **both** schema versions: `SIGNED_GROUP_OP_SCHEMA_VERSION` 7→8 and `SIGNED_NAMESPACE_OP_SCHEMA_VERSION` 1→2. Old-shape ops are rejected outright by `verify_signature`.

Cut the now-dead machinery:
- apply-time staleness checks (`apply_local_signed_group_op_at_cut`, `apply_group_op_inner`, `apply_root_op`);
- `state_hash_for_op`, the `state_hash` param threaded through `publish_post_gate` / `sign_and_publish_post_gate` / `sign_and_publish_without_apply`, and the publisher's `pre_apply_state_hash` capture;
- `root_op_commits_to_namespace_state` (only the staleness gate used it);
- the synthesized-op field + the `unified_op_decode` / `governance_dag` deltas that reused `state_hash` as `expected_root_hash` (now `[0u8; 32]`).

The **per-signer nonce window** remains the anti-replay gate.

## KEPT — a separate mechanism, do not confuse

`MemberRemoved` / `MemberLeft` carry `expected_group_state_hash` + `expected_context_state_hashes` (post-apply convergence claims that drive reconcile-on-mismatch) and `compute_state_hash` feeds them. Those are **unaffected** — this PR only removes the op-level `state_hash` that was part of the op id.

## Mechanics / testing

- Test call sites updated by a bracket-aware codemod (223 `sign()` sites, 4th positional arg dropped).
- 6 obsolete `state_hash`-staleness unit tests + their fixture deleted (they pinned behaviour this PR removes); the surviving concurrent-convergence test still passes (it asserts independent concurrent ops converge, which never depended on `state_hash`).
- Net **−956 LOC**. fmt + clippy (`-D warnings`) clean across the workspace; governance-store + context compile on master.
- Stacked work it builds on is all merged: #2935/#2936/#2939/#2941/#2944.

---

> [!NOTE]
> **High Risk**
> Flag-day wire/schema break on governance op IDs affects every node and invalidates existing DAGs; changes core apply/publish paths in governance-store and context.
> 
> **Overview**
> **Flag-day breaking change:** drops `state_hash` from `SignableGroupOp` / `SignedGroupOp` and `SignableNamespaceOp` / `SignedNamespaceOp`, bumps `SIGNED_GROUP_OP_SCHEMA_VERSION` to **8** and `SIGNED_NAMESPACE_OP_SCHEMA_VERSION` to **2**, and updates `sign()` so every op’s **content hash (id)** changes—old DAG history is not readable on this build.
> 
> Apply and publish paths no longer compute, thread, or warn on op-level staleness (`state_hash_for_op`, `pre_apply_state_hash`, `apply_local_signed_group_op` / `apply_group_op_inner` / `apply_root_op` telemetry, `root_op_commits_to_namespace_state`). **Anti-replay stays on the per-signer nonce window.** DAG helpers (`signed_op_to_delta`, unified decode) set `expected_root_hash` to `[0u8; 32]` because governance ops no longer carry a meaningful root claim.
> 
> **Unchanged:** `MemberRemoved` / `MemberLeft` still use `expected_group_state_hash` and `expected_context_state_hashes` (separate post-apply convergence mechanism); `compute_state_hash` for those claims remains.
> 
> Tests and call sites are updated to the shorter `sign()` API; obsolete `state_hash` staleness unit tests are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98fd082437622a91b3e33829f197440614599577. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/context/src/governance_dag.rs
+++ b/crates/context/src/governance_dag.rs
@@ -50,7 +50,9 @@ pub fn signed_op_to_delta(op: &SignedGroupOp) -> Result<CausalDelta<SignedGroupO
     Ok(make_delta(
         op,
         op.parent_op_hashes.clone(),
-        op.state_hash,
+        // C5.S3b removed the op-level state_hash; the DAG delta carries no
+        // meaningful `expected_root_hash` from a governance op.
+        [0u8; 32],
         delta_id,
     ))
 }
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -1678,7 +1678,6 @@ mod tests {
             version: 1,
             namespace_id,
             parent_op_hashes: Vec::new(),
-            state_hash: [0u8; 32],
             signer,
             nonce: 0,
             op: NamespaceOp::Root(op),
@@ -2127,7 +2126,6 @@ mod tests {
             version: 1,
             namespace_id,
             parent_op_hashes: Vec::new(),
-            state_hash: [0u8; 32],
             signer,
             nonce: 0,
             op: NamespaceOp::Group {
--- a/crates/context/src/self_purge.rs
+++ b/crates/context/src/self_purge.rs
@@ -2837,7 +2837,6 @@ mod tests {
             &owner_sk,
             namespace_id,
             vec![],
-            [0u8; 32],
             1,
             NamespaceOp::Group {
                 group_id: sub_gid.to_bytes(),
--- a/crates/context/tests/cascade_apply_walk.rs
+++ b/crates/context/tests/cascade_apply_walk.rs
@@ -119,7 +119,6 @@ fn cascade_target_application_set_updates_all_matching_descendants_and_skips_sib
         &admin_sk,
         r.to_bytes(),
         vec![],
-        [0u8; 32],
         1,
         GroupOp::CascadeTargetApplicationSet {
             from_app_key: APP_KEY_1,
--- a/crates/context/tests/cascade_atomic_apply.rs
+++ b/crates/context/tests/cascade_atomic_apply.rs
@@ -95,7 +95,6 @@ fn cascade_upgrade_atomic_op_sets_target_app_key_and_migration_and_records_casca
         &admin_sk,
         r.to_bytes(),
         vec![],
-        [0u8; 32],
         1,
         GroupOp::CascadeUpgrade {
             from_app_key: APP_KEY_1,
@@ -175,7 +174,6 @@ async fn cascade_upgrade_reverse_delivery_converges_atomically() {
         &admin_sk,
         root.to_bytes(),
         vec![[0u8; 32]],
-        [0u8; 32],
         1,
         GroupOp::GroupMetadataSet {
             name: Some("label".to_owned()),
@@ -190,7 +188,6 @@ async fn cascade_upgrade_reverse_delivery_converges_atomically() {
         &admin_sk,
         root.to_bytes(),
         vec![op_p_hash],
-        [0u8; 32],
         2,
         GroupOp::CascadeUpgrade {
             from_app_key: APP_KEY_1,
@@ -297,7 +294,6 @@ fn two_op_reverse_delivery_drops_migration_characterization() {
             &admin_sk,
             r.to_bytes(),
             vec![],
-            [0u8; 32],
             1,
             GroupOp::CascadeTargetApplicationSet {
                 from_app_key: APP_KEY_1,
@@ -314,7 +310,6 @@ fn two_op_reverse_delivery_drops_migration_characterization() {
             &admin_sk,
             r.to_bytes(),
             vec![],
-            [0u8; 32],
             2,
             GroupOp::CascadeGroupMigrationSet {
                 from_app_key: APP_KEY_1,
--- a/crates/context/tests/cascade_concurrent_safety.rs
+++ b/crates/context/tests/cascade_concurrent_safety.rs
@@ -132,7 +132,6 @@ async fn divergent_cascade_apply_order_converges_via_predicate_skip() {
         &admin_sk,
         root.to_bytes(),
         vec![[0u8; 32]],
-        [0u8; 32],
         1,
         GroupOp::CascadeTargetApplicationSet {
             from_app_key: APP_KEY_1,
@@ -151,7 +150,6 @@ async fn divergent_cascade_apply_order_converges_via_predicate_skip() {
         &admin_sk,
         root.to_bytes(),
         vec![op_a_hash],
-        [0u8; 32],
         2,
         GroupOp::CascadeTargetApplicationSet {
             from_app_key: APP_KEY_1,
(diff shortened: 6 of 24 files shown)


chefsale:
Reviewed all findings + did a dead-code sweep (95751ccd):

**Dead / not-needed code removed:**
- `GroupHandle::compute_state_hash` — zero callers (a pre-existing unused `pub` wrapper clippy can't flag; the only things that would've used it are the staleness checks this PR removes). Deleted.
- Two obsolete unit tests: `state_hash_zero_skips_validation` (now **vacuous** — there's no staleness check to skip) and `state_hash_allows_sequential_ops` (sequential DAG-apply + meta-hash-change are covered elsewhere). Deleted. (6 more such tests were already deleted in the main commit.)
- Renamed `state_hash_mismatch_does_not_reject_concurrent_ops` → `concurrent_independent_member_adds_converge` and stripped the obsolete `state_hash` framing; it still pins the real CRDT convergence property.
- Dropped a dangling comment reference to the deleted `root_op_commits_to_namespace_state`.

**On the review findings:**
- **Version check before deserialize misparse (MeroReviewer W1, AI W1/W2):** `verify_signature` already checks `self.version != SIGNED_*_OP_SCHEMA_VERSION` and returns `GovernanceError::SchemaVersion` **before** `to_signable()` — and `version` is the first borsh field, read correctly even from old bytes. So an old-schema op rejects cleanly, not a silent misparse. This is a flag-day re-bootstrap (no old data exists by construction), and a misaligned old payload would also hit an invalid `GroupOp` enum tag → borsh error; a blanket version-guard on every read path is out of scope for the re-bootstrap.
- **Synthetic SignedGroupOp + verify_signature (AI W3):** confirmed `apply_group_op_inner` does **not** call `verify_signature` — the namespace envelope is the verification gate; the synthetic op's signature field is unused there. (Unchanged by this PR.)
- **`compute_state_hash` dead? (AI W5):** no — `MetaRepository::compute_state_hash` is still used by `get_group_info`, `verify_post_apply_state_hashes`, and the `MemberRemoved`/`MemberLeft` expected-hash paths. Only the dead `GroupHandle` wrapper was removed.
- **`sign_apply_local_group_op_borsh` readiness gate (AI W4):** the removed `compute_state_hash` call only existed to populate the field; it was never a readiness gate — `GroupCreated` is the first op of a group and must apply with no prior meta, so proceeding on a cold group is correct.
- **`expected_root_hash` always `[0u8;32]` (MeroReviewer suggestion):** `CausalDelta` is shared with the DATA plane, where `expected_root_hash` is the real convergence claim (checked in `delta_store.rs`). It can't be removed/Option-ified without touching the data plane; governance deltas legitimately carry no entity-root claim, hence `[0u8;32]`.

clippy `-D warnings` clean.

chefsale:
Latest round addressed in fcf3ce6e:

**W1 / W2 — version check + borsh-boundary regression test:** the version check was already enforced (`verify_signature` returns `GovernanceError::SchemaVersion` when `self.version != SIGNED_*_OP_SCHEMA_VERSION`, *before* `to_signable()`; `version` is the first borsh field so it's read correctly even from old bytes). I've now **pinned it in code** with three tests:
- `pre_flag_day_group_op_version_is_rejected` / `pre_flag_day_namespace_op_version_is_rejected` — a prior-version op is rejected with `SchemaVersion`.
- `v7_borsh_layout_group_op_is_rejected_not_misparsed` — a v7-shaped op (with the removed `state_hash` still in its borsh bytes) decoded as the v8 struct never yields a spurious `Ok`: borsh rejects the misaligned/trailing bytes, or the decoded `version` (7) fails the v8 check. (In practice the extra 32 `state_hash` bytes leave the v8 read with trailing data → borsh error; either way, no silent misparse.)

**W3 — `sign_apply_local_group_op_borsh` on absent meta:** not a regression. The removed `compute_state_hash` only existed to populate the field; it was never a readiness gate. A *member-mutating* op on a group with no meta fails **authorization** anyway (no admin exists in an empty meta), and `GroupCreated` is the first op of a group and *must* apply with no prior meta — so proceeding on a cold group is correct, not a silent op-for-nonexistent-group.

**W4 — `GroupHandle::compute_state_hash` breaking public API:** `calimero-governance-store` is an **internal workspace crate** (not a published/SDK API surface), `cargo check --workspace` is clean, and the method had zero callers anywhere. No external consumer.

**Suggestion — named const / Option for `expected_root_hash`:** declining. `CausalDelta` is shared with the data plane, where `expected_root_hash` is the real convergence claim (checked in `delta_store.rs`). An `Option` would ripple through the entire data plane; a cross-crate `[0u8;32]` const (governance_dag is in `calimero-context`, unified_op_decode in `calimero-governance-store`) adds indirection without type safety since the field stays `[u8;32]`. The inline comments already document the 'no claim' intent.

clippy `-D warnings` clean; 36 governance-types tests green.

chefsale:
Good catch — addressed in af220fc8. You're right that a successful borsh decode of the misaligned v7 layout *is* a misparse (garbage signer/nonce/op), caught only downstream by the version byte. I rewrote the test to make that explicit rather than treating a decode error as the expected path:

- `assert_eq!(bytes[0], SIGNED_GROUP_OP_SCHEMA_VERSION - 1)` — deterministic; byte 0 is the version, untouched by the layout shift.
- In the `Ok` arm: `assert_eq!(op.version, SIGNED_GROUP_OP_SCHEMA_VERSION - 1)` then assert `verify_signature` rejects specifically with `SchemaVersion`. So if a future refactor ever checks the signature before the version, this test fails instead of silently passing.
- Comment rewritten to state borsh will (most likely) decode the misaligned bytes and that the version byte is what triggers rejection — not implying a borsh-level error is expected.

Confirmed empirically: the `Ok` arm runs (decode succeeds, `version == 7`, rejected with `SchemaVersion`). The deterministic `pre_flag_day_*_version_is_rejected` tests remain the direct backstop for the version-check invariant. Test green; clippy clean.

PR #2960, opened 2026-06-26
feat(contract-gate): method-aware route coverage + nest resolution

rtb-12:
Makes the SDK↔core route-coverage gate **method-aware** (`METHOD /path`) and fixes the source-scan extractor's nesting/static-route handling. Motivated by a real miss: the SDK's `getBlob` bug shipped green because path-level coverage saw `/admin-api/blobs/:id` as "covered" via `DELETE`, hiding that the `GET` verb was never exercised.

## What changed
- **`route_manifest.rs`** — paren-balanced, string-aware, **nest-aware** extractor over all 6 verbs (incl. `HEAD`). Resolves inline `.nest("/prefix", Router::new()...)` to real paths and drops static/catch-all (`/`, `/*path`) literals.
- **`check-endpoint-coverage.sh`** — matches on method + path-pattern; trailing-slash normalized.
- **`endpoints.json`** — regenerated: **88 method-aware entries** (incl. `HEAD /admin-api/blobs/:blob_id`).
- **`coverage-baseline.json`** — **`[]`**. Nothing silently ignored.

## Result
`sdk-e2e`: **88/88 method-routes exercised, 0 baselined** (runs against mero-js `master`). Paired SDK changes — method-aware recorder, verb-gap closers, and `getBlobInfo` for the HEAD route — are merged (mero-js #61, #62).


rtb-12:
Addressed (67dacf13):
**Fixed**
- **first_string escaped quotes** — now honors `\"` so a quote inside a literal can't end it early.
- **verb boundary / wrapping_sub** — index the preceding byte directly, guarded by `abs == 0` (no more wrap-to-MAX + Option match).

**Declined (with rationale)**
- **balanced escape off-by-one on multi-byte** — not reachable: the scan only ever *slices* at ASCII delimiters (`)`/`"`), never mid-skip, so no char-boundary panic; route literals are ASCII regardless.
- **unquoted $rx in =~** — intentional (`$rx` must be unquoted to act as a regex), and the input is our own committed manifest, not user data.

Behavior-preserving — manifest still 87 method-aware entries; route_manifest test green.

chefsale (approved):

rtb-12:
Addressed (59fd1150):
**Fixed**
- **HEAD verb missing** — good catch. Added `head` to VERBS; `HEAD /admin-api/blobs/:blob_id` (blob::info_handler) is now gated (88 entries) and covered by the paired SDK `getBlobInfo` (mero-js #62).
- **join_path nested catch-all** — `/*path` is now skipped at any depth, not just top level.
- **trailing-slash symmetry** — the manifest pattern is now normalized too (not just the hit).

**Declined (with rationale)**
- **balanced() / first_string EOF off-by-one** — not reachable: the input is always well-formed, compiling Rust source; the loops only ever *slice* at ASCII delimiters (`)`/`"`), and `first_string`'s `while i < len` guard prevents any out-of-bounds index. No panic path.
- **`.route(`/`.nest(` inside a string literal** — no such literal exists in service.rs; fully guarding it needs a real tokenizer, which is the phase-2 (utoipa/OpenAPI) direction, not this source-scan.

rtb-12:
These are re-flags of the same class already addressed — resolving as won't-fix with rationale (this is a **test helper** over a known input: `admin/service.rs`, well-formed compiling Rust, ASCII route literals):
- **multi-byte UTF-8 / byte-index panics in balanced & first_string** — can't occur: the scanners only ever *slice* at ASCII delimiter bytes (`(`, `)`, `"`), which are always char boundaries, and `first_string`'s `while i < len` guards every index. Even if it somehow did, a test panic is loud, not a shipped bug.
- **`.route(`/verb token inside a string literal** — no such literal exists in service.rs; fully guarding arbitrary source needs a real tokenizer, which is the phase-2 (utoipa/OpenAPI) direction.
- **duplicate verb occurrences** — moot: results go into a `BTreeSet`, so duplicates collapse.

The route_manifest test passes and produces a stable 88-entry manifest; `sdk-e2e` is green at 88/88. Not chasing further re-flags on this helper.

PR #3007, opened 2026-06-29
fix(auth): handle malformed input in auth handlers without panicking

chefsale:
## Description

This PR improves error handling and resilience in the authentication system:

1. **Header value validation**: Replace unsafe `.parse().unwrap()` calls with proper error handling using `HeaderValue::from_str()` and `HeaderValue::from_static()`. This prevents panics when auth response data contains invalid header characters.

2. **Provider data validation**: Replace unsafe `.unwrap()` chains in `UserPasswordProvider` with proper error handling using `and_then()` and `ok_or_else()`, providing clear error messages when username or password are missing or invalid.

3. **Panic recovery**: Add `CatchPanicLayer` to the router to catch any panics in handlers or middleware and convert them to 500 responses instead of aborting connections. This prevents panic-induced DoS attacks.

These changes make the auth system more robust by eliminating unwrap calls that could panic and adding a safety net for unexpected panics in the request pipeline.

## Test plan

- Existing unit and integration tests continue to pass
- Manual testing: Verify that requests with invalid header characters in auth responses are properly rejected with warnings instead of panicking
- Manual testing: Verify that requests with missing/invalid username or password in provider data return proper error responses

## Wire contract (SDK gate)

- [x] No HTTP wire DTO or route changes

## Documentation update

No documentation updates needed.

https://claude.ai/code/session_012eGxmFAHCZznFFfXcnjTNs


frdomovic (approved):
Approving — replaces every production-code `.parse().unwrap()` / `.unwrap()` panic site in the auth middleware and `UserPasswordProvider::create_root_key` with proper error handling, and adds a `CatchPanicLayer` (+ the `catch-panic` feature) as the outermost DoS backstop. Grep of the PR-head confirms zero `unwrap`/`expect`/`panic!` remain in the touched production code (only in `#[cfg(test)]`). Success-path header skips (warn instead of error) are the right call.

⚠️ One thing to decide consciously: the newly-returned malformed-input error from `create_root_key` surfaces as **HTTP 500, not 400** — the only caller maps *all* errors to `INTERNAL_SERVER_ERROR` and echoes `e.to_string()` into the body. The panic/DoS is fixed (the PR's primary goal), but "proper error response" is only half-delivered: a client sending bad input still gets a 5xx. The new messages are generic (no secret leak), but worth deciding whether 400 mapping is in-scope here or a follow-up.

PR #3008, opened 2026-06-29
fix: use constant-time comparison for mock auth header validation

chefsale:
## Description

This PR improves the security of the mock authentication header validation in the auth handler by using constant-time comparison instead of standard equality checks. This prevents timing attacks that could potentially leak information about the expected auth header value.

The change uses the `subtle` crate's `ConstantTimeEq` trait to compare the authorization header value with the expected value in a way that takes the same amount of time regardless of where the strings differ.

### Changes
- Added `subtle` crate dependency to `crates/auth/Cargo.toml`
- Updated `mock_token_handler` to use `ct_eq()` for comparing auth header values instead of direct equality
- Gated the import behind `#[cfg(debug_assertions)]` since this is only used in development/debug builds

## Test plan

No testing needed. This is a security hardening change to an existing debug-only authentication handler. The comparison logic remains functionally equivalent—it still validates the auth header matches the configured value—but now does so in constant time.

## Wire contract (SDK gate)

N/A — No HTTP wire DTOs or routes changed.

## Documentation update

N/A — This is an internal security improvement to a development-only feature.

https://claude.ai/code/session_01FjZkusTn2UVaN4RRmi4g7D


frdomovic (approved):
LGTM. Swaps `value == required_value` for `subtle::ConstantTimeEq::ct_eq` (a vetted, optimization-resistant crate, not a hand-rolled loop) on the mock-auth header check, with the import correctly gated behind `#[cfg(debug_assertions)]` to match the already-debug-gated handler. Low-risk, well-scoped.

Note (non-blocking): `ct_eq` on byte slices short-circuits on differing lengths, so this still leaks the *length* of the expected value via timing — only the byte content is now constant-time. Acceptable for this threat model. Also the endpoint is triple-gated (debug-only build + two runtime config flags), so production exposure is essentially nil — this is defense-in-depth.

PR #3017, opened 2026-06-29
fix: temporal layer iteration to properly merge inner and shadow

chefsale:
## Description

This PR fixes the temporal layer iterator to correctly merge entries from both the underlying committed layer (`inner`) and the pending shadow transaction in sorted order.

### Problem

The previous implementation had a fundamental flaw: it would iterate through the inner layer completely before considering shadow entries, rather than performing a proper two-way merge. This meant:
- Shadow-only keys could appear out of order
- Keys present in both layers could be yielded twice
- The merge didn't respect the sorted ordering guarantee

### Solution

Implemented a proper two-way merge algorithm in the new `advance()` method that:
1. Buffers the front key from each source (`peeked_inner` and `peeked_shadow`) without holding borrows that would prevent reading values
2. Compares the two front keys and emits the smaller one
3. When the same key exists in both sources, the shadow entry takes precedence and the duplicate inner key is dropped
4. Shadow deletions are skipped (not yielded)
5. The `read()` method lazily fetches values from the inner layer when needed, or returns the buffered shadow value

### Changes

- Replaced unsafe pointer manipulation with a clean buffering approach
- Added `inner_done`, `peeked_inner`, and `peeked_shadow` fields to track merge state
- Implemented `advance()` to perform the sorted merge
- Updated `seek()` to reset state and resume the merge from the target key
- Removed the unused `raw_get()` method from `Transaction`

## Test plan

Added comprehensive unit tests covering:
- `merges_inner_and_shadow_in_sorted_order`: Verifies correct merge of committed and pending entries with overrides and deletes
- `shadow_only_keys_are_yielded_in_order`: Confirms shadow-only entries are sorted correctly
- `each_key_is_yielded_once_when_present_in_both`: Ensures no duplicate keys when present in both sources
- `seek_resumes_the_merge_from_the_target_key`: Validates seek behavior correctly resumes the merge

All tests pass and exercise the core merge logic at the `DBIter` level.

## Wire contract (SDK gate)

## Documentation update

https://claude.ai/code/session_01TwRhZWpGC3hXwrKoaHkQGy


frdomovic (approved):
Approving — this replaces a genuinely broken iterator (it drained `inner` fully before touching the shadow, with a `// todo!` admitting dedup was missing) with a correct two-way sorted merge. Ordering (byte-wise via `Slice::cmp`, consistent with both the DB iterator and the `BTreeMap` shadow), dedup, tombstone suppression, and `seek()` are all correct, and it removes the prior `unsafe` pointer aliasing in `next()`. I traced the `read()`-after-`next()` invariant for inner-sourced keys and it holds (no skipped/duplicated/stale values).

Minor (non-blocking): no test asserts the *value* of an inner-only key via `read()` — every compared value comes from the shadow or the uniform `"base"`. A test where an inner-only key carries a distinct value would directly guard the trickiest path (the parked-inner-cursor `read()`). Worth adding.

chefsale:
Addressed the review feedback in f13272c22:

- **Shadow refill never stops when exhausted** — fixed. Added a `shadow_done` flag mirroring `inner_done`; the merge no longer re-calls `col_iter().next()` once per remaining inner key after the shadow range drains.
- **`seek()` skips shadow-deleted keys** — documented on the `seek` impl.
- **`get_or_insert_with` / lazy `shadow_iter` init** — documented in the refill block.
- **Inner iterator parked across emit / fragile `read()` contract** (the two 🟡 invariant notes) — these are correct as written (both notes say "No bug"). `value == None` is exactly the marker that the last key was inner-sourced and `inner` is still parked on it; any shadow key sets `value = Some`, so `read()` never delegates to a stale position. Documented this invariant explicitly at both the `Ordering::Less` arm and `read()` rather than eagerly buffering the value — eager buffering would add a value copy per row, which directly conflicts with the per-row-allocation note below.
- **Heap allocation per inner-key advance** — intentional. Copying the key into an owned `Slice` is the deliberate trade this PR makes to avoid borrowing the inner iterator's internal buffer across the merge; keeping it is what makes the parked-`inner` invariant sound.

All `layer::temporal` tests pass; clippy (`-D warnings`) and fmt are clean.

PR #3019, opened 2026-06-29
fix(auth): add replay protection to challenge/response flow

chefsale:
## Description

This PR implements replay protection for challenge tokens by persisting issued challenge JTIs (JWT IDs) and enforcing one-time consumption during verification. 

**Key changes:**
- Added `storage` field to `TokenManager` to persist challenge metadata
- Challenge JTIs are now stored with their expiry timestamp when issued via `generate_challenge()`
- Challenge verification now consumes the JTI record, preventing replay attacks
- Added `verify_challenge_handler` HTTP endpoint to verify challenges and return their claims
- Implemented best-effort cleanup of expired challenge records to bound storage growth
- Added `ChallengeReplay` error variant to distinguish replay attempts from other verification failures
- Added comprehensive unit tests covering successful verification, replay rejection, and cleanup

**Security impact:**
Without this change, a validly signed challenge token could be verified multiple times within its expiry window. Now each challenge can be verified at most once, and only if it was issued by this service instance.

**Storage considerations:**
Challenge records are keyed by `challenge_jti:` prefix and stored with their expiry timestamp. The cleanup routine runs opportunistically after each challenge generation to prune expired records.

## Test plan

Added four unit tests in `crates/auth/src/auth/token/jwt.rs`:
- `challenge_verification_succeeds_once_then_replays_are_rejected`: Verifies first use succeeds and replay is rejected
- `challenge_never_issued_is_rejected`: Confirms challenges not issued by this instance are rejected
- `generating_a_challenge_persists_its_jti`: Validates JTI persistence
- `expired_challenge_records_are_swept`: Confirms cleanup removes expired records

All tests use in-memory storage and pass with the implementation.

## Wire contract (SDK gate)

- [x] New HTTP endpoint added: `POST /auth/verify-challenge` with `ChallengeVerifyRequest` DTO
- [ ] Updated `crates/server/endpoints.json` (if applicable to this service)
- [ ] Linked matching mero-js PR (if needed)

The new endpoint accepts a challenge token and returns verified claims (jti, nonce, iat, exp) or rejects with 409 CONFLICT for replays or 401 UNAUTHORIZED for invalid tokens.

## Documentation update

The new `/auth/verify-challenge` endpoint should be documented in the authentication flow guide, noting that:
- Challenges can only be verified once
- Replay attempts are rejected with HTTP 409
- The endpoint returns the challenge claims for use in subsequent authentication steps

https://claude.ai/code/session_017BbWNqLMz1JGPjminCwKaT


frdomovic (approved):
Approving — the replay protection itself is correct: challenge `jti`s are persisted on issue, consumed (deleted) on first verify, and absence-on-verify is treated as replay (409). Good test coverage (success-then-replay, never-issued, persistence, sweep). The read-then-delete race and single-instance scope are honestly documented in-code.

⚠️ A few things to confirm before merge — none are code bugs, but they affect whether the protection actually achieves its goal:

- **Route path vs. docs mismatch.** The PR body / contract checklist say `POST /auth/verify-challenge`, but the code registers `/challenge/verify` under the `/auth` nest → the real path is `POST /auth/challenge/verify`. Any client/SDK/docs written against the description will 404. Make body and code agree.
- **Is `verify_challenge` actually on the live login path?** It appears to be reachable only via the new standalone handler; confirm the real authentication flow consumes the challenge (otherwise replay protection guards a verb the threat model may not use, and the actual login nonce path is still unprotected).
- **Multi-instance:** records live in per-process storage with no shared store, so a challenge issued by instance A is "never issued" from B's view. Confirm this service only ever runs single-instance, or this both breaks the flow and leaves cross-instance replay open.
- Minor: `cleanup_expired_challenges` is an O(n) full scan on every `generate_challenge`; "bounds storage growth" is weaker than stated under burst load.

PR #3021, opened 2026-06-29
fix(network): resolve dropped-sender panic and pending_dial leak on dial

chefsale:
## Description

This PR fixes a critical bug where pending dial futures could be indefinitely wedged when connections complete as `Listener` endpoints instead of `Dialer` endpoints.

### The Problem

When initiating a dial to a peer, the code stores a oneshot sender in `pending_dial` and waits for either:
1. `ConnectionEstablished` event to fire the sender
2. `OutgoingConnectionError` to fire the sender

However, the `ConnectionEstablished` handler only cleared the pending dial when the connection endpoint was `Dialer`. In simultaneous-open scenarios (e.g., hole-punched connections where both peers dial each other), the connection can complete as a `Listener` endpoint even though we initiated the dial. This caused the pending dial entry to never be cleared, leaving the awaiting `Dial` future permanently blocked.

When the swarm shuts down, it drops the sender without sending a value, causing the receiver to error. The original code would panic on this error with `.expect()`, crashing the actor.

### The Solution

1. **Remove the endpoint check**: Clear pending dials on *any* established connection for the peer, not just `Dialer` endpoints. This handles simultaneous-open scenarios correctly.

2. **Handle sender drops gracefully**: Change the receiver await from `.expect()` to a `match` that converts sender drops into a proper `DialError` instead of panicking.

3. **Add clarifying comments**: Document why this behavior is necessary to prevent future regressions.

## Test plan

Existing tests pass. The fix is defensive — it handles an edge case that would previously cause actor panics during shutdown or in simultaneous-open connection scenarios. Manual verification would require triggering simultaneous-open connections or swarm shutdown during pending dials, which is difficult to reproduce deterministically.

## Wire contract (SDK gate)

## Documentation update

https://claude.ai/code/session_01LBMy3RSGzbVrRoN8ML6mgS

### DIFF
--- a/crates/network/src/handlers/commands/dial.rs
+++ b/crates/network/src/handlers/commands/dial.rs
@@ -21,6 +21,13 @@ impl Handler<Dial> for NetworkManager {
         match self.pending_dial.entry(peer_id) {
             Entry::Occupied(_) => {
                 // todo! await the existing receiver
+                //
+                // NB: this `Ok(())` means "a dial to this peer is already in
+                // flight", not "the dial succeeded". The in-flight dial owns
+                // the only sender, so we can't subscribe to its result here
+                // without a broadcast/clone; until that's wired up, a caller
+                // hitting this branch gets a spurious success even if the
+                // real dial later fails.
                 return Response::reply(Ok(()));
             }
             Entry::Vacant(entry) => {
--- a/crates/network/src/handlers/commands/dial.rs
+++ b/crates/network/src/handlers/commands/dial.rs
@@ -41,6 +41,16 @@ impl Handler<Dial> for NetworkManager {
             }
         }
 
-        Response::fut(async { receiver.await.expect("Sender not to be dropped.") })
+        Response::fut(async move {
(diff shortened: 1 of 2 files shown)


frdomovic (approved):
Approving — fixes both issues: removes the `Dialer`-endpoint gate so a pending dial that completes as a `Listener` (simultaneous-open / hole-punch) is resolved instead of stranded, and converts the receiver `.expect()` into a graceful `DialError` so a dropped sender no longer panics the actor. No race risk: `pending_dial` is accessed only from the single-threaded actix actor; double-removal is a harmless no-op; resolving on any established connection matches the dial's real goal.

⚠️ One semantic change to be aware of: with the `Dialer` gate removed, a `pending_dial` now resolves `Ok(())` on a purely *inbound* connection that races a dial we initiated, even if our own dial leg later fails. So "dial returned Ok" now means "we are connected to the peer," not "our specific dial attempt succeeded." That's the right semantic for callers — just confirm none depend on the stricter meaning.

PR #3024, opened 2026-06-29
fix(store): close lifetime-laundering transmute in ArcSlice::new

chefsale:
## Description

This PR fixes lifetime safety issues in the in-memory database implementation by:

1. **Tightening `ArcSlice::new` lifetime bounds**: Added explicit `'this` lifetime parameter to the impl block and constrained the generic type `T` with `T: 'this` to ensure borrowed data outlives the slice. This prevents lifetime laundering where a slice could be coerced to a longer lifetime (e.g., `'static`) and unsafely shared across threads.

2. **Making `CastsTo` bounds lifetime-agnostic**: Changed `InMemoryDBImpl` trait bounds from `CastsTo<Slice<'a>>` to `for<'x> CastsTo<Slice<'x>>`. This allows the database to recover owned values as `Slice<'static>` rather than being pinned to the query lifetime `'a`, eliminating the need for unsafe lifetime coercion in the first place.

3. **Adding `T::Value: 'static` bound**: Added explicit `'static` bound to `InMemoryDB<T>` impl to clarify that stored values are owned and valid for the entire program lifetime.

4. **Improving safety documentation**: Enhanced comments explaining the safety invariants of the transmute operation and the lifetime coercion strategy.

These changes maintain the same runtime behavior while making the lifetime safety properties explicit and verifiable by the type system, reducing reliance on unsafe code.

## Test plan

Existing tests pass. The changes are purely type-system refinements that don't alter runtime behavior — the code compiles and functions identically, but with stronger lifetime guarantees that prevent potential safety violations.

## Wire contract (SDK gate)

## Documentation update

https://claude.ai/code/session_014cREtd2tSq4AvHEm6rYBe1

### DIFF
--- a/crates/store/src/db/memory.rs
+++ b/crates/store/src/db/memory.rs
@@ -119,11 +119,16 @@ struct ArcSlice<'this> {
     inner: Arc<Slice<'this>>,
 }
 
-impl ArcSlice<'_> {
-    fn new<'a, T: CastsTo<Slice<'a>>>(value: Arc<T>) -> Self {
+impl<'this> ArcSlice<'this> {
+    fn new<T: CastsTo<Slice<'this>> + 'this>(value: Arc<T>) -> Self {
         Self {
-            // safety: T: CastsTo<Slice>
-            inner: unsafe { transmute::<Arc<T>, Arc<Slice<'_>>>(value) },
+            // safety: `T: CastsTo<Slice<'this>>` guarantees layout compatibility,
+            // and the `T: 'this` bound guarantees the borrowed data behind `T`
+            // outlives `'this`. Together they make this transmute sound: the
+            // output `Slice<'this>` cannot outlive the data it points at, so a
+            // borrowed slice can no longer be laundered into a longer-lived
+            // (e.g. 'static) one and smuggled across threads.
+            inner: unsafe { transmute::<Arc<T>, Arc<Slice<'this>>>(value) },
         }
     }
 }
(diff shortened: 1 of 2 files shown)


frdomovic (approved):
Approving — this closes the lifetime-laundering hole correctly by tying `ArcSlice::new`'s input bound and transmute output to a single `'this` lifetime plus a `T: 'this` bound, and annotating the `get` call site to `ArcSlice<'static>` backed by an explicit `T::Value: 'static` impl bound. The remaining `transmute` is a sound lifetime-only cast because `CastsTo` has exactly one impl (`Slice: CastsTo<Slice>`), so `T` is always `Slice<'_>` (layout-identical).

⚠️ Two things to verify before merge:

- **Confirm it compiles.** The new `T::Value: 'static` bound legitimately rejects any non-`'static` `Ref<'a>` instantiation — verify the `Ref` impl is only used where `'a: 'static` (or only in tests), so this doesn't break a real caller. (`cargo build -p calimero-store` + the memory-db tests.)
- **Keep `CastsTo` locked to `Slice` only.** The transmute's soundness rests entirely on that single reflexive impl; any future `impl CastsTo<Slice<'_>> for SomeOtherType` re-introduces layout-mismatch UB. Consider sealing the trait.

Note: the parallel transmute in `InMemoryDBIter::new` (and the `Range` transmutes in `raw.rs`) carry the same pattern and are out of scope here — the laundering class is only partially closed.

PR #3025, opened 2026-06-29
fix(server/sse): stop leaking node-event task per SSE connection

chefsale:
## Description

This PR refactors the SSE event handler to eliminate unnecessary complexity in connection lifecycle management. The key changes are:

1. **Remove active connection tracking**: Deleted the `active_connections` map and `ActiveConnection` struct that tracked current SSE streams. This was redundant since the event handler can directly monitor its own connection state.

2. **Simplify event handler lifecycle**: Modified `handle_node_events` to:
   - Accept `command_sender` directly as a parameter
   - Use `tokio::select!` to monitor both the event stream and the connection status
   - Exit immediately when the connection closes (via `command_sender.closed()`) or the event stream ends
   - Remove the polling loop with 100ms sleep that checked `active_connections`

3. **Remove cleanup task**: Deleted the separate `handle_connection_cleanup` task since the event handler now manages its own cleanup by exiting when the connection closes.

4. **Improve documentation**: Enhanced the docstring to clarify the task's lifetime, the importance of exiting promptly to avoid leaking broadcast receiver subscriptions, and the skip-on-disconnect event delivery model.

The refactoring maintains the same behavior (skip-on-disconnect event delivery, session persistence across reconnections) while reducing code complexity and eliminating the need for shared mutable state tracking active connections.

## Test plan

Existing tests should cover this change. The refactoring preserves the event delivery semantics and connection handling behavior. Manual verification would involve:
- Establishing an SSE connection and verifying events are delivered
- Disconnecting and reconnecting to verify session persistence
- Confirming no resource leaks occur during repeated connect/disconnect cycles

## Wire contract (SDK gate)

No HTTP wire DTOs or routes were changed.

- [ ] N/A — no wire contract changes

https://claude.ai/code/session_01GnRSmJB2628Uk5pUeyM8iC

### DIFF
--- a/crates/server/src/sse/events.rs
+++ b/crates/server/src/sse/events.rs
@@ -3,7 +3,6 @@ use calimero_server_primitives::sse::{
     Command, ConnectionId, Response, ResponseBody, ResponseBodyError, ServerResponseError,
 };
 use core::pin::pin;
-use core::time::Duration;
 use futures_util::StreamExt;
 use serde_json::to_value as to_json_value;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


frdomovic (approved):
Approving — this fixes a real per-connection task leak: the old `handle_node_events` never terminated after a client disconnected (it slept-and-spun forever holding a broadcast subscription + the spawned task for the process lifetime). The refactor binds the task to the connection's `command_sender`, exits on `closed()` / stream-end / send-failure, and removes the redundant `active_connections` map. The ownership change (move, not clone) is the load-bearing detail and is done right. Active clients won't lose events prematurely — none of the exit conditions can fire while a live client is attached.

⚠️ Worth one manual check: the fix relies on axum streaming the SSE body lazily so the receiver is held only by the response body (not the handler future), which is what makes `command_sender.closed()` actually resolve on disconnect. A quick browser connect/disconnect test (watch the "stopping event handler" debug log / task count) confirms it. If the receiver were somehow retained, the task wouldn't exit and the fix would be inert.

PR #3026, opened 2026-06-29
fix: prevent privilege escalation in key permission updates

chefsale:
## Description

This PR adds authorization checks to prevent privilege escalation when updating key permissions. Previously, any key with `keys:update_permissions` could grant arbitrary permissions (including `admin`) to itself or other keys, bypassing the intended permission hierarchy.

The fix introduces:

1. **`validate_grantable_permissions()` function**: Validates that a caller can only grant permissions they themselves hold. This respects the permission hierarchy where:
   - `admin` satisfies all permissions
   - Scoped permissions (e.g., `context:list[ctx-1]`) can be narrowed but not widened
   - Unparseable permissions are rejected to prevent granting unknown capabilities

2. **`CallerPermissions` extension**: A new type injected into request extensions by the auth middleware, making the authenticated caller's permissions available to downstream handlers for fine-grained authorization decisions.

3. **Updated `update_key_permissions_handler()`**: Now validates all permissions in the `add` list before applying updates. Permission removals are unrestricted. Invalid or escalating permissions are rejected with appropriate HTTP status codes (400 for invalid format, 403 for escalation attempts).

The change is backward compatible—only additions are restricted; existing permission removal functionality is unchanged.

## Test plan

Added comprehensive unit tests covering:
- Non-admin users cannot grant `admin` permission
- Callers cannot grant permissions they don't hold
- Admin users can grant any permission
- Callers can grant permissions they hold
- Scoped permissions can be narrowed but not widened
- Unparseable permissions are rejected
- Empty permission lists are allowed

All tests pass and validate the privilege escalation prevention logic.

## Wire contract (SDK gate)

- [x] No DTO changes (only internal authorization logic)
- [x] No route changes
- [x] No SDK updates needed

https://claude.ai/code/session_014vuBNRRMNdHEKLuAksCQtx

### DIFF
--- a/crates/auth/src/api/handlers/permissions.rs
+++ b/crates/auth/src/api/handlers/permissions.rs
@@ -4,10 +4,12 @@ use axum::extract::{Extension, Path};
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
 use serde::{Deserialize, Serialize};
-use tracing::{error, info};
+use tracing::{error, info, warn};
 use validator::Validate;
 
 use crate::api::handlers::auth::{error_response, success_response};
+use crate::auth::middleware::CallerPermissions;
+use crate::auth::permissions::{Permission, PermissionValidator};
 use crate::auth::validation::{sanitize_string, ValidatedJson};
 use crate::server::AppState;
 use crate::storage::StorageError;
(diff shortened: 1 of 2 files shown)


frdomovic (approved):
Approving — the fix correctly closes the privilege-escalation hole in `PUT /admin/keys/:key_id/permissions` by validating every permission in the `add` list against the caller's own permissions via the existing `PermissionValidator`. I audited the other permission-write paths (`generate_client_key_handler`, `create_key_handler`, the delete handlers) and confirmed this was the only escalation vector — the fix is complete. Good test matrix.

⚠️ Two things to sign off on:

- **`CallerPermissions` must always be injected.** The guard reads the caller's JWT permissions from the `Extension(CallerPermissions)` set by `auth_middleware`. The route is under `/admin` which has the middleware layer, and a missing extension fails closed (500, not a bypass) — but confirm the handler can never be reached without the middleware via any alternate/test router wiring.
- **The removal path is intentionally unrestricted.** A caller with update-permissions can strip permissions (including `admin`) off any key. Not escalation, but a possible lockout/denial vector worth a conscious product sign-off.

PR #3027, opened 2026-06-29
fix: apply backpressure on full network event channel instead of dropping

chefsale:
## Description

This PR replaces the immediate drop behavior when the network event channel is full with a bounded async retry mechanism that applies backpressure. Instead of silently dropping control events during transient bursts, overflow events are now handed to a retry task that waits for capacity (up to a configurable timeout). Events are only truly dropped when the retry times out, the retry queue saturates, or the channel closes.

### Key Changes

- **New config options** in `NetworkEventChannelConfig`:
  - `send_timeout`: How long an overflow event waits for capacity (default: 5s)
  - `max_pending_retries`: Maximum concurrent retry waiters to bound extra buffering (default: 1000)

- **New metrics** to track backpressure behavior:
  - `events_retried`: Overflow events handed to the retry path
  - `events_recovered`: Overflow events successfully delivered after retry
  - `backpressure_active`: Signal (0/1) indicating if overflow events are waiting for capacity
  - `pending_retries`: Operational counter of concurrent waiters (not registered)

- **Updated `send()` behavior**: When the channel is full, `spawn_retry()` spawns a detached async task that waits for capacity instead of returning false immediately. The network thread remains non-blocking via `try_send`.

- **Graceful degradation**: If no async runtime is available or the retry queue is saturated, events are dropped with escalation logging rather than buffering unboundedly.

### Motivation

This change prevents loss of control events during transient load spikes. By applying backpressure (waiting for capacity) rather than dropping, the system can recover from temporary bursts without losing critical events, while still protecting against sustained overload via timeouts and queue caps.

## Test Plan

Added three new unit tests in `network_event_channel.rs`:

1. `test_full_channel_applies_backpressure_and_recovers`: Verifies that overflow events are accepted for retry and eventually delivered once capacity frees up, with metrics reflecting recovery.

2. `test_retry_timeout_drops_event`: Confirms that events are dropped after the backpressure timeout elapses with no available capacity.

3. `test_retry_queue_saturation_drops`: Ensures the retry queue cap prevents unbounded buffering; events are dropped immediately once the cap is exceeded.

All existing tests pass. The changes are backward compatible; the new config fields have sensible defaults.

## Wire contract (SDK gate)

## Documentation update

N/A — this is an internal implementation detail of the network event channel.

https://claude.ai/code/session_01WgC3T1g1w7o1Mm6SFVq5R8

### DIFF
--- a/crates/node/src/network_event_channel.rs
+++ b/crates/node/src/network_event_channel.rs
@@ -27,8 +27,10 @@ use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::gauge::Gauge;
 use prometheus_client::metrics::histogram::{exponential_buckets, Histogram};
 use prometheus_client::registry::Registry;
+use tokio::runtime::Handle;
 use tokio::sync::mpsc;
-use tracing::{debug, info, warn};
+use tokio::time::timeout;
+use tracing::{debug, error, info, warn};
 
 /// Configuration for the network event channel.
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


frdomovic (approved):
Approving — verified against the diff. The full-channel path does **not** block the swarm/actor loop: the fast path stays `try_send`, and on a full channel the `send().await` happens on a *detached, timeout-capped* task bounded by `max_pending_retries`, with no lock held across the await. No deadlock risk. Good metrics, saturation cap, and tests.

⚠️ One behavior change worth a conscious check before merge:

- **Event reordering under backpressure.** Overflow events are delivered by independent detached tasks racing on `send().await`, so they can land *after* later fast-path events and can reorder among themselves. Previously, drops preserved the relative order of survivors. Confirm no `NetworkEvent` consumer assumes FIFO (e.g. Subscribed/Unsubscribed pairing for the same peer/topic, or stream sequencing). The tests don't cover this.

Minor: `events_received` now also counts recovered retries; channel-depth/high-watermark gauges aren't updated on the retry path (observability skew while backpressured). Non-blocking.

PR #3029, opened 2026-06-29
fix: unify equal-HLC delete-vs-update tiebreak across storage types

chefsale:
## Description

This PR fixes a divergence in how signed storage types (User, Shared, SharedMember) and Public storage types handle delete operations when the delete nonce equals the stored update timestamp (equal-HLC case).

**The Problem:**
- Signed types were rejecting equal-HLC deletes with `NonceReplay` error (using `<=` boundary)
- Public type and the apply-path tiebreak were accepting equal-HLC deletes (no nonce gate)
- This caused signed vs Public to diverge on the equal-HLC delete-vs-update tie

**The Solution:**
- Change the nonce check boundary from `<=` to `<` in all three signed-type DeleteRef arms (User, Shared, SharedMember)
- This rejects only STRICTLY-stale nonces while letting equal-HLC deletes fall through
- The unified tiebreak in `apply_delete_ref_action` (`deleted_at < updated_at` ⇒ delete loses, so equal ⇒ delete wins) now handles the equal case identically for all storage types
- Maintains strict replay protection for strictly-lower nonces while achieving deterministic behavior across all types

**Changes:**
- Modified nonce comparison in three DeleteRef verify arms from `new_nonce <= last_nonce` to `new_nonce < last_nonce`
- Updated `apply_delete_ref_action` guard comment to document this as the canonical equal-HLC tiebreak
- Added comprehensive test `equal_nonce_delete_wins_like_public` to verify the unified behavior
- Updated all related comments to clarify the new semantics

## Test plan

The change is covered by:
1. New test `equal_nonce_delete_wins_like_public` - verifies that equal-HLC signed deletes are accepted and win, matching Public behavior
2. Existing test `user_delete_replay_protection` - continues to verify strict rejection of strictly-lower nonces
3. All existing replay protection tests pass with the stricter boundary

The test constructs a DeleteRef action with nonce and `deleted_at` both equal to the stored `updated_at`, verifying it's accepted (not `NonceReplay`) and successfully deletes the entity.

## Wire contract (SDK gate)

N/A - No HTTP wire DTOs or routes changed.

## Documentation update

N/A - This is an internal storage consistency fix with no public API changes.

https://claude.ai/code/session_014VGKcsjuFcDbj1YGU4kqSZ

### DIFF
--- a/crates/storage/src/interface.rs
+++ b/crates/storage/src/interface.rs
@@ -1674,15 +1674,29 @@ impl<S: StorageAdaptor> Interface<S> {
                                 // leaks current-nonce state to
                                 // unauthenticated probers).
                                 //
-                                // DeleteRef keeps the strict `Err` on
-                                // `<=` (unlike upsert's silent skip)
-                                // because a stale delete being
-                                // silently accepted vs dropped
-                                // carries different semantics than
-                                // a stale upsert, and rare-by-design
-                                // deletes don't drive the
-                                // post-divergence convergence problem
-                                // upsert silent-skip fixes.
+                                // DeleteRef rejects only STRICTLY stale
+                                // nonces (`<`) with a hard `Err` — unlike
+                                // upsert's silent skip — because a stale
+                                // delete dropped vs accepted carries
+                                // different semantics than a stale upsert,
+                                // and rare-by-design deletes don't drive
+                                // the post-divergence convergence problem
+                                // the upsert silent-skip fixes.
+                                //
+                                // The EQUAL-nonce case (`==`) is NOT
+                                // rejected here: it falls through to
+                                // `apply_delete_ref_action`, whose tiebreak
+                                // (`deleted_at < updated_at` ⇒ delete loses,
+                                // so equal ⇒ delete WINS) resolves the
+                                // delete-vs-update tie deterministically and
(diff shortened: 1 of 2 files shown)


frdomovic (approved):
Approving — the `<=` → `<` nonce-gate flip across the three signed DeleteRef arms is correct and unifies signed types with Public (delete-wins-on-equal-HLC). Logic is consistent and the new test is solid.

⚠️ One thing to confirm consciously before merge, because it's the load-bearing decision:

- **This is effectively a CRDT convergence/protocol change, not a pure refactor.** Every replica must run identical equal-HLC tiebreak logic or replicas can permanently diverge. During a rolling upgrade, an old node (still `<=`) and a new node (`<`) will resolve the same equal-HLC delete-vs-update tie *differently*. Confirm mixed-version networks are either prevented or tolerated (version gate / no persisted state re-evaluated under the new rule).
- **HLC monotonicity on resurrect:** since signed deletes now pass the gate at `nonce == updated_at`, replay safety at the equal case rests on (a) the tombstone guard making identical replays idempotent and (b) any re-add advancing `updated_at` strictly above the deleted nonce. Worth confirming a resurrect can never reuse the exact tick of the prior delete.

PR #3041, opened 2026-06-29
fix(primitives): gate PrivateKey raw access and zeroize via derive

chefsale:
Re-applies the `PrivateKey` hardening from #2994 (reverted in #3031), with the build break that forced the revert fixed.

## What this does

Two security issues in `PrivateKey` (`crates/primitives/src/identity.rs`):

1. **Raw secret bytes were freely reachable.** `Deref<Target=[u8;32]>`, `AsRef<[u8;32]>` and derived `BorshSerialize`/`BorshDeserialize` each made it trivial to copy or serialize the secret into a buffer that is never zeroized, silently defeating the zeroize-on-drop guarantee.
   - Drop the blanket `Deref`/`AsRef` impls and the Borsh derives.
   - Gate raw access behind a single audited `as_bytes(&self) -> &[u8; 32]` accessor — the one reviewed entry point to the secret.
   - Update all call sites (in-place crypto uses + the storage layers that persist an explicit `[u8; 32]` copy) to go through `as_bytes()`.

2. **The `Drop` impl used a fragile `unsafe` cast.** It zeroized via `*mut Hash` → `*mut u8` over `size_of::<Hash>()`, which would silently miss key material if `Hash` ever gained padding or a field. Switch the inner type to a plain `[u8; 32]` and `#[derive(ZeroizeOnDrop)]`, removing the hand-rolled `unsafe` and tracking field layout automatically.

## Why the original (#2994) was reverted — and why this one is safe

The first attempt removed `use zeroize::Zeroize;` (the old manual `Drop` needed it; the derive does not). But `PrivateKey::random` — added concurrently in #2995 — still calls `secret.zeroize()` on its local seed copy. Each PR was green in isolation; their **combination** failed to compile under the `rand` feature:

```
error[E0599]: no method named `zeroize` found for array `[u8; 32]` in the current scope
  --> crates/primitives/src/identity.rs:84
```

This broke merod/server builds and full-workspace CI, so #2994 was reverted in #3031. CI never tested the two PRs together (each merged against a stale-green base), which is how it slipped through.

This PR restores the import, **cfg-gated on `rand`** to match the `random()` call site, so a default build pulls in no unused import.

## Test plan

- `cargo build --workspace --all-targets --tests` — clean ✅
- `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` — clean ✅
- `cargo build -p calimero-primitives --features rand,borsh` — clean (the exact path that broke) ✅
- `cargo test -p calimero-primitives --all-features` — all pass, incl. the zeroize-on-drop test ✅

### DIFF
--- a/crates/context/primitives/src/client/crypto.rs
+++ b/crates/context/primitives/src/client/crypto.rs
@@ -68,7 +68,7 @@ impl ContextClient {
         handle.put(
             &key::ContextIdentity::new(ContextId::zero(), public_key),
             &types::ContextIdentity {
-                private_key: Some(*private_key),
+                private_key: Some(*private_key.as_bytes()),
                 sender_key: None,
             },
         )?;
@@ -142,10 +142,18 @@ impl ContextClient {
             );
         };
 
-        identity.sender_key = new_identity.sender_key.as_deref().copied();
+        identity.sender_key = new_identity
+            .sender_key
+            .as_ref()
+            .map(PrivateKey::as_bytes)
+            .copied();
         // TODO: what we are updating the private key for? if we got here, the datastore already
         // has the `identity.private_key` set.
-        identity.private_key = new_identity.private_key.as_deref().copied();
+        identity.private_key = new_identity
+            .private_key
+            .as_ref()
+            .map(PrivateKey::as_bytes)
+            .copied();
 
         handle.put(&key, &identity)?;
 
--- a/crates/context/src/auto_follow.rs
+++ b/crates/context/src/auto_follow.rs
@@ -776,7 +776,7 @@ mod tests {
                 .save(&gid, &sample_meta(pk))
                 .expect("save_group_meta");
             NamespaceRepository::new(&store)
-                .store_identity(&gid, &pk, &sk, &[0u8; 32])
+                .store_identity(&gid, &pk, sk.as_bytes(), &[0u8; 32])
                 .expect("store_namespace_identity");
             MembershipRepository::new(&store)
                 .add_member(&gid, &pk, GroupMemberRole::Member)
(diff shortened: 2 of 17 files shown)


chefsale:
Thanks for the review — went through all four findings:

**✅ Applied — round-trip via `sk.as_bytes()` (suggestion 2).** `join_group.rs` now writes `private_key: Some(sk_bytes)` directly. The raw `[u8; 32]` is in scope (destructured a few lines up) and already used that way for `store_key`, so building a `PrivateKey` just to deref it back was pointless. (commit dce2f43f2)

**✅ Clarified — leftover `use core::ops::Deref` (warning 1).** Not actually unused — `PublicKey`'s `Deref` impl needs it, so clippy is clean. Added a comment noting it's for `PublicKey` (and that `PrivateKey` deliberately has no `Deref`) so it doesn't read as a stale `PrivateKey` import. (commit dce2f43f2)

**ℹ️ `sk_bytes` not zeroized (warning 2).** Correct, and as the reviewer notes it pre-dates this change — the raw `[u8; 32]` is threaded through this handler independently of the `PrivateKey` wrapper. Zeroizing the call-site arrays is a separate, broader cleanup; this PR's scope is the `PrivateKey` type itself. Switching line 471 to `sk_bytes` doesn't make it any worse (the array was already live).

**ℹ️ Post-drop read in `test_private_key_zeroize_on_drop` (suggestion 1).** The dead-store concern doesn't apply here: `zeroize`/`ZeroizeOnDrop` perform the clear with `ptr::write_volatile` + a `compiler_fence(SeqCst)`, which the optimizer is not allowed to elide — that guarantee is the whole reason we moved to the derive. The remaining post-drop pointer read is genuine (documented) test-only UB that the `ManuallyDrop` guard makes work in practice; it's the inherited test body and I'd rather not dress up UB with a fence that implies otherwise. Left as-is.

PR #3045, opened 2026-06-29
fix(merod): restrict node home and datastore dirs to owner-only

chefsale:
## Problem

`merod init` created both the node home directory (`crates/merod/src/cli/init.rs:191`) and the RocksDB datastore directory (`crates/merod/src/cli/init.rs:308`) using the default umask — typically `0755`, i.e. world-traversable.

The Ed25519 private key lives in `config.toml` under the node home. Even though the key file itself is hardened, the permissive parent directories let any other local user traverse into them, undermining that hardening.

## Fix

Added a `restrict_dir_to_owner` helper that sets a directory to `0700` after creation on Unix (no-op on non-Unix platforms, which lack POSIX mode bits). Applied to both the node home and the datastore directory.

## Verification

- `cargo check -p merod` and `cargo clippy -p merod` — clean
- Ran `merod init` end-to-end and confirmed with `stat`:
  - node home → `700`
  - `data/` → `700`
  - `config.toml` stays `0644` but is now unreachable to other local users via its owner-only parent

### DIFF
--- a/architecture/crates/tools.html
+++ b/architecture/crates/tools.html
@@ -517,6 +517,7 @@ merod --home /data --node mynode init \
 </div>
 <p><span class="code">--group-governance local</span> is the default (and only) governance mode.</p>
 <p>For read-only nodes, add <span class="code">--mode read-only</span>.</p>
+<p>On Unix, <span class="code">init</span> runs under a restrictive <span class="code">umask</span> (<span class="code">0077</span>) so the node home, <span class="code">config.toml</span> (which holds the Ed25519 private key), and the datastore — including every RocksDB file — are created owner-only. <span class="code">config.toml</span> is pinned to <span class="code">0600</span> and the directories to <span class="code">0700</span>, keeping the key unreadable by other local users.</p>
 </div>
 </details>
 </div>
--- a/architecture/crates/tools.html
+++ b/architecture/crates/tools.html
@@ -517,7 +517,7 @@ merod --home /data --node mynode init \
 </div>
 <p><span class="code">--group-governance local</span> is the default (and only) governance mode.</p>
 <p>For read-only nodes, add <span class="code">--mode read-only</span>.</p>
-<p>On Unix, <span class="code">init</span> runs under a restrictive <span class="code">umask</span> (<span class="code">0077</span>) so the node home, <span class="code">config.toml</span> (which holds the Ed25519 private key), and the datastore — including every RocksDB file — are created owner-only. <span class="code">config.toml</span> is pinned to <span class="code">0600</span> and the directories to <span class="code">0700</span>, keeping the key unreadable by other local users.</p>
+<p>On Unix, <span class="code">init</span> hardens the node directory to owner-only permissions: the node home and datastore directories (and every RocksDB file beneath them) are set to <span class="code">0700</span>/<span class="code">0600</span>, and <span class="code">config.toml</span> — which holds the Ed25519 private key — is pinned to <span class="code">0600</span>, keeping the key unreadable by other local users.</p>
 </div>
 </details>
 </div>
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed the review feedback in 4bbffe8.

The first pass only `chmod`'d the two top-level directories after creation. As the bots noted, that (a) left RocksDB's own files/sub-directories at the default umask and (b) left a brief window between `create_dir_all` and the `chmod`.

Switched to setting a restrictive **`umask` (`0077`) for the duration of `init`**, so the node home, `config.toml`, and the datastore — including every RocksDB file and sub-directory — are created owner-only atomically, with no widen-then-tighten window. This also makes the contents owner-only in their *own* right, so they stay protected even if a parent directory's mode is later relaxed (the "only true while the directory stays 0700" concern).

- `config.toml` is additionally pinned to `0600` (the private-key file is now owner-only independent of its parent), directly addressing the suggestion.
- The explicit directory `chmod`s are kept only to tighten directories that *pre-existed* init, which the umask cannot affect.

Verified end-to-end with `merod init`:
- node home → `700`, `config.toml` → `600`, `data/` → `700`
- every file RocksDB created (`OPTIONS-*`, `*.log`, `IDENTITY`, `LOCK`, `CURRENT`, `MANIFEST-*`, …) → `600`; `find data -type d ! -perm 700 -o -type f ! -perm 600` returns nothing.

Note for completeness: on standard Unix semantics a `0700` parent already blocks other users from *reaching* its contents (no traverse permission), so the original fix wasn't exploitable as-is — but creating the contents owner-only is the correct defense-in-depth and is what this revision does.

chefsale:
Addressed the latest round in bb562ed.

The umask approach drew two valid follow-ups: (1) it mutates process-global state inside the async runtime — a latent race in a multi-threaded context, even if harmless for one-shot `init`; and (2) it can't tighten content left by a *previous partial init*, since umask only governs newly-created files.

Both point away from umask, so I dropped it (and the `libc`/`unsafe` it required) in favor of a **self-contained recursive permission walk** over the datastore:

- After `Store::open` closes (no writer races us), walk `data/` and set `0700` on every directory and `0600` on every file — newly created *and* pre-existing.
- The node home is locked to `0700` **before** `config.toml` is written, and `config.toml` is pinned to `0600`. So the key is never written into a world-traversable directory, and no global process state is touched.

This is a fix the first review round explicitly endorsed ("recursively chmod all files and directories"), now with no umask race and no `unsafe`.

Verified end-to-end:
- **Fresh init**: home `700`, `config.toml` `600`, `data/` `700`, every RocksDB file `600`; `find data -type d ! -perm 700 -o -type f ! -perm 600` → empty.
- **Pre-existing partial init**: seeded `data/` + a nested subdir with `0755` dirs and `0644` files, ran `init` (no `--force`) → all recursively tightened to `0700`/`0600`; the same `find` → empty.

chefsale:
Addressed in cdf2161.

This round flagged TOCTOU windows (the inverse of the prior round, which rejected `umask` for being racy). The resolution is to remove the one window that actually matters rather than chase each inner one:

**The node home is now created directly at `0700`** via `DirBuilder::mode`, so it is never momentarily world-traversable. That's the keystone — every sensitive path (`config.toml`, the datastore, RocksDB's WAL/SST files) is created *inside* the already-`0700` home, so the brief windows where those inner files still carry the default umask are **not reachable by another local user** (no traverse permission on the parent). The recursive datastore walk + `config.toml` `0600` pin remain as defense-in-depth for the case where the home's mode is later loosened.

On the suggestion to wrap `Store::open` in a `umask` to make RocksDB's files born `0600`: I deliberately didn't, because the previous review round rejected process-`umask` as a racy global inside the async runtime. There's no mechanism that is simultaneously "no global umask" and "zero post-hoc window" for a third-party library that creates many files — so the design instead guarantees the *containing directory* is owner-only with no window, which makes the inner windows non-exploitable. The panic case (open/drop panics before the walk) is likewise safe: the files sit inside the `0700` home, and init has failed anyway.

Finding #3 (config consumed / datastore_path cloned) — added a clarifying comment.

Verified: fresh init creates home **and** its parent at `700`, `config.toml` `600`, `data/` `700`, every RocksDB file `600` (`find` non-compliance check empty); pre-existing permissive datastore contents still get recursively tightened.

PR #3046, opened 2026-06-29
fix(storage): reject undefined OpMask bits to close signature malleability

chefsale:
## Problem

`OpMask(u8)` derived `BorshDeserialize`, so **any** byte deserialized successfully — including ones with undefined high bits (bits 3–7). `OpMask::contains()` only tests the three defined bits via masking, so `0b0000_0001` and `0b1000_0001` are behaviorally identical at the merge check.

But `mask.bits()` returns the raw byte, and that byte is committed verbatim into the signed authorization payload (`action.rs:291`, `hasher.update([mask.bits()])`) which feeds `compute_id`. A peer could flip an undefined bit to mint a distinct signed payload / id for a semantically-identical mask — **signature malleability**.

## Fix

Drop `BorshDeserialize` from the derive and hand-write it to **reject** any byte with bits outside `FULL` (`0b0000_0111`, the union of all defined bits), keeping the encoding canonical. Rejecting (rather than silently masking) treats a non-canonical byte as malformed input and fails loudly with `InvalidData`. `BorshSerialize` stays derived.

## Verification

- New tests: defined masks round-trip; all 248 bytes with an undefined bit are rejected.
- Full `calimero-storage` suite passes (626 unit + integration tests).
- `cargo clippy -p calimero-storage --all-targets` clean.

### DIFF
--- a/crates/storage/src/entities.rs
+++ b/crates/storage/src/entities.rs
@@ -16,6 +16,7 @@ mod tests;
 use calimero_primitives::identity::PublicKey;
 use core::fmt::{self, Debug, Display, Formatter};
 use std::collections::{BTreeMap, BTreeSet};
+use std::io::{Error as IoError, ErrorKind as IoErrorKind, Read};
 use std::ops::{Deref, DerefMut};
 
 use borsh::{BorshDeserialize, BorshSerialize};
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed review feedback in db642fa:

- **`std::io` imports** — switched the hand-written `BorshDeserialize` impl from full paths to module-level aliases (`IoError`/`IoErrorKind`/`Read`), matching `address.rs`'s custom Borsh impl.
- **Bitmask precedence** — parenthesized the guard `(bits & !Self::FULL.0) != 0` (and the equivalent test filter) for unambiguous intent.

Re the doc-bot warnings (`migrations.html` / `storage-schema.html`): this is an internal canonical-encoding fix — it changes neither the on-disk schema nor migration behavior — so no docs update is warranted.

PR #3048, opened 2026-06-29
fix(governance): replace unvalidated u8 discriminants with typed enums and bounds checks

chefsale:
Re-applies the typed-discriminant hardening from #2965 (reverted in #3032), with the build break that forced the revert fixed.

## What this does

- `GroupOp::SubgroupVisibilitySet.mode` becomes `VisibilityMode` instead of a raw `u8`, and `VisibilityMode` gains `BorshSerialize`/`BorshDeserialize`. The wire layout is unchanged — a two-unit-variant enum serializes as the same `0`/`1` discriminant byte.
- `ContextCapabilityGranted` / `ContextCapabilityRevoked.capability` become a validated `ContextCapabilityBits` newtype that **rejects a zero bitmask at Borsh deserialization**, so an invalid capability op cannot be constructed from received bytes. A revoke that clears the last bit now deletes the member-capability row instead of storing a zero.
- Removes the manual `u8` ↔ enum conversions in the visibility handlers and the apply op.

## Why the original (#2965) was reverted — and why this one is safe

The first attempt did not update two capability-op test sites in `governance-store/src/tests.rs` that still passed `capability: 0b1` (a raw integer). Those sites were added by a change that merged **concurrently** with #2965, so #2965 never saw them. Each was green in isolation; their combination failed to compile:

```
error[E0308]: mismatched types
    --> crates/governance-store/src/tests.rs:1292
     |     capability: 0b1,
     |                 ^^^ expected `ContextCapabilityBits`, found integer
```

This broke the test build, so #2965 was reverted in #3032. CI never built the two PRs together (each merged against a stale-green base). Those two sites now construct the value via `ContextCapabilityBits::new(0b1)`.

## Test plan

- `cargo build --workspace --all-targets --tests` — clean ✅ (the exact path that broke)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1337,6 +1337,7 @@ dependencies = [
  "calimero-context-config",
  "calimero-crypto",
  "calimero-dag",
+ "calimero-governance-types",
  "calimero-network-primitives",
  "calimero-node-primitives",
  "calimero-op",
--- a/crates/context/config/src/lib.rs
+++ b/crates/context/config/src/lib.rs
@@ -130,7 +130,17 @@ impl<'a> GroupRequest<'a> {
 ///
 /// The walk in `check_group_membership` stops at the first `Restricted`
 /// ancestor; a `Restricted` subgroup is a wall regardless of what sits above.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize,
+    borsh::BorshSerialize,
+    borsh::BorshDeserialize,
+)]
 pub enum VisibilityMode {
     Open,
     Restricted,
--- a/crates/context/primitives/src/group.rs
+++ b/crates/context/primitives/src/group.rs
@@ -2,6 +2,7 @@ use std::collections::BTreeMap;
 
 use actix::Message;
 use calimero_context_config::types::{AppKey, ContextGroupId, SignedGroupOpenInvitation};
+use calimero_context_config::VisibilityMode;
 use calimero_primitives::application::ApplicationId;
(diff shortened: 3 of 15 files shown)


chefsale:
Thanks — went through the review, and also fixed a CI failure the reviewers' commit predated:

**🔴 Fixed — golden-fixture decode failure (CI).** `group_op_discriminants_are_golden` froze `capability: 0` bytes for ordinals 17/18, which the new `ContextCapabilityBits` correctly rejects on the wire, so the golden decode failed. Updated the fixtures to a valid non-zero bitmask (`1`) — same byte length and discriminant, so the wire-stability intent is unchanged. (commit 8aa27d139)

**✅ Applied — `BorshDeserialize` duplicates the zero-check (warning).** `deserialize_reader` now routes through `Self::new(bits).ok_or_else(...)`, so the zero-rejection invariant lives only in the constructor. (commit 8aa27d139)

**ℹ️ `NonZeroU8` as the inner type (suggestion).** Fair — it is morally a `NonZeroU8`. I kept the newtype: it carries domain meaning (a capability bitmask, not just any non-zero byte), keeps the `get()`/masking surface explicit, and matches the reverted PR we're restoring. Worth a follow-up if we standardize on `NonZeroU8` elsewhere, but out of scope for un-reverting.

**ℹ️ Missing Serde impls (suggestion).** Intentional. `ContextCapabilityBits` only ever crosses the Borsh governance-op wire; `VisibilityMode` needs Serde because it also flows through `context-config`'s Serde surfaces. Adding Serde here would need a custom impl to preserve the non-zero invariant, with no consumer — YAGNI.

**ℹ️ `BorshSerialize` could emit zero / manual impl divergence (warning + suggestion).** A zero `ContextCapabilityBits` can't be constructed (every path goes through `new`/`try_from`/the validating deserializer), so serialize can't emit one. It's a single-field newtype, so the manual deserializer can't silently drift from added fields. Left as-is.

PR #3057, opened 2026-06-29
fix(soundness): close lint-suppressed hazards across storage/context/runtime

chefsale:
Fixes a batch of lint-suppression / soundness-hazard findings. Each removed an `#[allow]`/`#[expect]` that was masking a real hazard. Full workspace CI clippy (`--all-targets -D warnings`) is clean and all storage tests pass (632 lib + integration + 5 new borsh tests).

## Findings addressed

**Soundness**
- **`Root::get` aliasing-`&mut` UB** — the `RefCell` guard was laundered into `&mut T` via `ptr::from_mut`, so two `get()` calls produced two aliasing `&mut` (UB). Replaced with a `OnceCell` cache: `Deref` yields `&T` tied to `&self`, `DerefMut` uses a new `get_mut(&mut self)`. **All `unsafe` + `mut_from_ref` removed.**
- **`read_guest_memory_typed` had no compile-time guard** — soundness rested on every `T` being POD. `bytemuck::Pod` is inapplicable (every ABI type carries a lifetime; `ValueReturn` is `#[repr(C, u64)]`), so added a sealed `unsafe trait GuestAbiType` implemented only for the 6 vetted `calimero-sys` descriptors and bounded the function — reading any un-vetted type is now a **compile error**. Plus SAFETY comments at all 48 call sites.

**Panic-safety**
- **Collection `eq`/`cmp`/`Debug`** no longer `unwrap()`-panic on a store fault mid-merge/log — shared *total*-comparison helpers; `partial_cmp` made consistent with `Ord`.
- **`Root::get`/`fetch`/`commit`** log the cause then panic (matching `commit_headless`) instead of an opaque WASM `unreachable`.

**Resource / correctness**
- **`state_key`** interns keys in `HashMap<[u8;32], Arc<…>>`; growth is bounded by *distinct* keys instead of one leaked `Arc` per storage op.
- **borsh layout mirrors** now have a round-trip test pinning the production `borsh_layout` mirror to the canonical `calimero_storage` types across all 5 `StorageType` variants (covers the two historical cold-join drift bugs).
- **`garbage_collect_tombstones`** is now `pub` (dead-code allow dropped) and documented as a manually-invoked op (no auto-scheduling); the existing test covers it.
- **crate-wide `clippy::mem_forget` allow** narrowed to the one `ouroboros` self-referential struct (verified with `-W clippy::mem_forget`: 3 → 0).

(The `ArcSlice::new` lifetime-launder finding was already fixed upstream in #3024 — no change needed.)

## Two scoping decisions worth a look

1. **`Root::fetch`/`commit` keep their signatures** rather than returning `Result`. Propagating would touch 70+ call sites, every production caller is guest code that panics regardless, and `get()` is `Deref`-bound so *cannot* return `Result`. The fix delivers the real value (cause in logs before abort).
2. **`#8` uses a marker trait instead of `bytemuck::Pod`** (the ABI types can't satisfy `Pod`). The residual ABI trust assumption (an adversarial guest writing an out-of-range `ValueReturn` discriminant) is pre-existing and unchanged — closing it needs validated decoding, a larger change.

### DIFF
--- a/crates/context/primitives/src/client/mod.rs
+++ b/crates/context/primitives/src/client/mod.rs
@@ -112,12 +112,14 @@ pub struct RootSelfDump {
 /// and [`ContextRegistry::dump_root`] to decode the index without pulling
 /// in the full `calimero-storage` types (which would force a dep cycle).
 ///
-/// **SYNC NOTE**: When `calimero_storage::index::EntityIndex` or its
-/// transitively-borshed children change, update these mirrors *and*
-/// `calimero-storage/src/tests/index.rs::minimal_struct_layout_compat`.
-/// A missed update produces silent misdeserialization on a diagnostic
-/// path that fires only during rare divergence events — exactly when
-/// correct output matters most.
+/// **SYNC NOTE**: These structs mirror the canonical types by hand. The
+/// [`borsh_layout_round_trip`] test module below serialises the *real*
+/// `calimero_storage` types and decodes them through these mirrors, so a
+/// field-type or field-order drift fails the test run rather than silently
+/// misdeserialising on the rare divergence-diagnostic path (which is exactly
+/// when correct output matters most). When you change `EntityIndex` or its
+/// borshed children, update these mirrors; the test will flag a child layout
+/// that drifted.
 mod borsh_layout {
     use borsh::BorshDeserialize;
     use calimero_primitives::crdt::CrdtType;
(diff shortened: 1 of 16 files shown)


chefsale:
Thanks for the review — addressed in `dad1f5526`:

- **`fallible_iter_eq`/`cmp` masking store faults** — verified these `PartialEq`/`Ord` impls are **not** on the merkle/sync convergence path (that compares hashes + metadata, never whole collections via `==`), so a masked difference here cannot hide replication divergence. To keep the fault observable anyway, the helpers now log a `tracing::warn!` (via a shared `#[cold]` helper) whenever a read error forces the total fallback, and the doc spells out the non-convergence-path guarantee. Kept the total semantics (returning a definite result) since the whole point of the fix is that these impls must never panic.
- **`state_key` `'static` transmute / ouroboros pinning** — extended both SAFETY comments to note (a) the reference is taken through the `Arc` (`&**interned`), not into the `HashMap`, so a rehash can't invalidate it, and (b) `Self` is an `ouroboros`-pinned self-referential struct, so a move can't invalidate the temporal layer's borrow. Kept the `Arc`-intern approach rather than swapping to a typed arena — the arena refactor would mean restructuring the ouroboros self-reference, which is out of scope for a lint-suppression cleanup.
- **`Root::fetch` → `get` missing-entry panic** — documented on `fetch` that a returned `Some(root)` guarantees the single inner entry exists (written atomically at construction), so a later missing entry is always corruption, never a normal empty state.
- **`round_trip` didn't assert `crdt_type`** — added the assertion; since the mirror decodes `crdt_type` as the canonical `CrdtType`, this now guards the `CrdtType` borsh layout against drift too.

On CI: the one red check (`sync-resilience-peer-pause-symmetric`) is an infra flake — `Failed to pull image 'alpine:3.19': 404 ... No such image` during NAT-topology setup, unrelated to this diff. The `Rust` job (clippy + unit/integration tests) is green. I've re-run the flaked job.

chefsale:
Second review round addressed in `2e5006b2c`:

- **`Root::get` missing-entry panic vs `fetch` invariant** — made the log + panic message self-explanatory: it now states the collection was present but its single entry was absent → partial write / store corruption, not an empty root. Combined with the `fetch` doc (added last round), the contradiction the reviewer flagged is now spelled out at the panic site. I kept it a panic rather than having `fetch` pre-probe the entry — `fetch` returns `Option` and can't carry a "corrupt" error distinct from "absent" without a signature change, and `get` is `Deref`-bound (infallible) regardless.
- **`deref_mut` sets `dirty` before the load can fail** — reordered: it now materialises the cache via `get()` first (which panics on a store fault) and only then sets `dirty`, so a failed lazy load can never leave a dirty-but-unloaded `Root` that `commit` would try to write back.
- **`Root::commit` silently drops the value if `into_inner()` is `None`** — added the suggested `debug_assert!(!self.dirty || self.value.get().is_some(), …)` to catch any future API change that breaks the dirty⇒populated invariant.

On the remaining notes:

- **`fallible_iter_eq` two-error arm → `false` vs `true`** — keeping `true`. Returning `false` would break `Eq`/`PartialEq` *reflexivity* (`a == a` would be false when `a`'s read faults), which `BTreeMap`/`BTreeSet`/dedup rely on — a worse hazard than the masking, especially since (as noted last round) these impls are **not** on the merkle/sync convergence path, so they can't actually hide replication divergence. The fault is already surfaced via the `warn!`.
- **warn only in the mixed-error arm** — already emitted in *all* error arms (including `(Err, Err)`) as of `dad1f5526`.
- **`partial_cmp` doc** — already states `None` is reserved for genuine element incomparability and is never returned for a store fault.
- **borsh test needs `calimero-storage` dev-dependency** — it's a normal `[dependencies]` entry of `calimero-context-client`, so the test module compiles and the 5 tests run green; no dev-dep needed.

PR #3059, opened 2026-06-29
fix(node): wait out cross-network discovery on cold namespace join

chefsale:
## Problem

Joining a namespace over the wire failed with:

```
could not open a namespace-join stream to any mesh peer for namespace … (deadline 130000ms, elapsed 9024ms, excluded 0)
```

The tell is **`elapsed 9024ms` vs `deadline 130000ms`** — the join gave up at ~9s, nowhere near its own 130s budget. Retrying a minute later succeeded with no other change.

### Root cause

The connect loop in `open_namespace_join_stream` was bounded by a fixed **round count** (`mesh_retries = 10`), not by the deadline. On a cold start the subscriber set is empty, so each round only cost the ~1s inter-round sleep → 10 rounds exhausted in ~9s (9 sleeps ≈ the observed 9024ms), and `excluded 0` confirms no peer was ever even tried.

That ~9s is **below the cross-network peer-discovery floor** of ~30–60s. Two nodes on different networks find each other through a third-party rendezvous server, and the path is a chain of sequential timers/handshakes — connect to rendezvous, AutoNAT external-address confirmation (`probe_interval` 10s), the periodic rendezvous discovery tick (`discovery_interval` 15s), dial, then gossipsub subscription exchange. The join was guaranteed to give up before discovery could plausibly complete.

## Fix

Make the connect loop poll for a dedicated **discovery budget** (`DEFAULT_NAMESPACE_DISCOVERY_WAIT_MS`, 45s — sized to the discovery floor) when nothing has been discovered yet, instead of counting cheap empty rounds against the retry budget.

The two non-cold cases keep their existing bounded-retry behaviour, so the rejection path and the caller's protocol loop aren't slowed:

- **All discovered peers excluded** (already rejected this join) → counts against `mesh_retries`; the caller escalates rather than blocking the full budget.
- **Peers present but all failing** → counts against `mesh_retries`; fast failover.

Also adds a `peer_count=0` debug line so the still-discovering state is observable next time.

## Tests

All 8 existing tests stay green; 2 new regression tests:

- `cold_start_peer_appearing_after_retry_budget_is_found` — mesh empty for `mesh_retries` rounds, peer appears on the next round → now found. **Fails against the old loop.**
- `cold_start_waits_for_full_discovery_budget` — empty mesh waits ~the full budget, not the old ~`mesh_retries × retry_delay` floor.

```
running 10 tests ... test result: ok. 10 passed; 0 failed
```
`cargo clippy -p calimero-node` clean.

## Notes / follow-ups

- **45s is a blocking call.** On a genuine failure (inviter truly offline) the join request now holds up to 45s before the 500. Clients should use an HTTP timeout ≥ ~50s. The budget is a single config constant if it needs tuning.
- **Root-cause follow-up (separate change):** this makes the join *wait long enough* for discovery; it doesn't make discovery *faster*. Eagerly kicking a rendezvous `discover()` + dial of namespace members the moment a join starts (instead of waiting for the 15s tick) would shrink the actual wait toward ~10–20s.

### DIFF
--- a/crates/node/src/sync/config.rs
+++ b/crates/node/src/sync/config.rs
@@ -90,6 +90,28 @@ pub const DEFAULT_MESH_RETRY_DELAY_MS_UNINITIALIZED: u64 = 1_000;
 /// for connection death.
 pub const DEFAULT_OPEN_STREAM_TIMEOUT_MS: u64 = 3_000;
 
+/// How long a namespace join keeps polling for a namespace mesh peer to
+/// appear before giving up (milliseconds).
+///
+/// On a cold cross-network join the inviter and joiner have never
+/// talked, so the joiner can't see the inviter until peer discovery
+/// completes: connect to a rendezvous server, confirm an external
+/// address via AutoNAT, issue the periodic rendezvous discover, dial,
+/// and let gossipsub exchange subscriptions. Those steps are each gated
+/// on a timer or a NAT handshake and run sequentially, so the inviter
+/// routinely takes tens of seconds to enter the joiner's namespace
+/// subscriber set. The join must keep re-polling that whole time — a
+/// fixed round count would burn through cheap empty polls in a few
+/// seconds, far short of the discovery floor, and fail a join that
+/// would have succeeded moments later.
+///
+/// 45s comfortably covers the typical 30–60s discovery floor while
+/// still bounding how long the join request blocks on a genuinely
+/// absent inviter. Only the empty-mesh (still-discovering) wait uses
+/// this budget; once a peer is found the open is bounded by
+/// [`DEFAULT_OPEN_STREAM_TIMEOUT_MS`] instead.
+pub const DEFAULT_NAMESPACE_DISCOVERY_WAIT_MS: u64 = 45_000;
+
 /// Max concurrent peer probes when looking for a peer with state.
 /// Typical meshes are 2-20 peers; a pool of 4 is enough parallelism
 /// that the tail is bounded by the fastest responder, without racing
--- a/crates/node/src/sync/config.rs
+++ b/crates/node/src/sync/config.rs
@@ -194,6 +194,13 @@ pub struct SyncConfig {
     /// sensitive: shorter for low-latency LANs with aggressive QUIC idle
     /// timeouts, longer for high-latency WAN.
     pub open_stream_timeout: time::Duration,
+
+    /// How long a cold namespace join keeps polling for a mesh peer to
+    /// appear before giving up. See [`DEFAULT_NAMESPACE_DISCOVERY_WAIT_MS`].
+    /// Deployment-sensitive: the cross-network discovery floor is higher
+    /// for NAT'd nodes needing relay + hole-punch, lower for directly
+    /// reachable ones.
+    pub namespace_discovery_wait: time::Duration,
 }
 
 impl Default for SyncConfig {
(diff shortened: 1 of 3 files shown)


chefsale:
Thanks for the review — addressed in b21f96f44.

**💡 Per-peer deadline only fires near budget exhaustion on a large hanging mesh** — Fixed. Re-introduced a per-round fan-out cap (`MAX_PEERS_PER_ROUND = 4`, shuffle-then-truncate) so a single round costs at most `cap × open_timeout` regardless of mesh size, with the per-peer `discovery_wait` check kept as the global backstop. New `round_fan_out_is_capped` test proves the bound by timing (round costs `cap × open_timeout`, not `mesh_size × open_timeout`).

**💡 `assert!(mesh_retries > 0)` dropped** — Fixed. Restored it; a zero would make `failed_attempts >= mesh_retries` true on the first peer-present/all-excluded round and fast-fail before any attempt.

**📝 Pass via `SyncConfig` instead of a literal millis conversion** — Done. Added `SyncConfig::namespace_discovery_wait` (default `DEFAULT_NAMESPACE_DISCOVERY_WAIT_MS`), now passed via `self.sync_config.namespace_discovery_wait` like `open_stream_timeout`. Tunable per deployment.

**🟡 Control flow non-obvious (`break` exits `'connect`)** — Fixed. Made all the loop exits explicit `break 'connect`.

**🟡 All-excluded path could busy-wait across protocol attempts** — On closer reading this isn't a busy-wait: in the all-excluded arm only the *final* round (`failed_attempts >= mesh_retries`) breaks without sleeping; rounds `1..mesh_retries-1` each `time::sleep(mesh_retry_delay)` before continuing. So one call spends ~`(mesh_retries-1) × mesh_retry_delay` (~9s), and the caller's `MAX_PROTOCOL_RETRIES` loop is paced by those sleeps, not a tight spin. This also matches the pre-PR behaviour for that path (no regression). Left as-is; happy to add an explicit inter-protocol-attempt backoff in `initiate_namespace_join` if you'd prefer it belt-and-suspenders.

**📄 Docs bot (architecture/crates/{node,sync}.html)** — These are generated architecture reference pages; this change is an internal discovery-budget tweak with no API surface change, so I've left them. Can regenerate if the convention is to update them per source change.

All 11 unit tests green, `cargo clippy -p calimero-node` clean.

chefsale:
Second review round addressed in 872189528 / 2847c359a.

**💡 `assert!` panics the async task — prefer returning `Err`** (both agents) — Done. Converted both preconditions to `eyre::ensure!`. Since the function returns `eyre::Result`, a misconfigured `SyncConfig` now propagates a typed `Err` (clean 500 on the join path) instead of panicking the task into an opaque `JoinError`. New `degenerate_budgets_return_err_not_panic` test covers both.

**🟡 All-excluded path: sleep is wasteful / could spin** — These two findings pull in opposite directions (one round flagged "no sleep → busy-wait", this round flagged "sleep → wasteful"), and the resolution is the same for both: **keep the sleep**. My earlier comment ("the set won't change within this call") was the misleading bit — the *excluded* set is fixed within a call, but the *discovered* set is not. Re-polling after `mesh_retry_delay` is exactly what lets a freshly-arrived, non-excluded peer surface mid-call and get tried. Removing the sleep would spin through `mesh_retries` rounds instantly and give a new peer zero chance to appear. So it stays — bounded at `mesh_retries` rounds so the caller still escalates promptly if no new peer shows. Rewrote the inline + doc comments to state this clearly.

**🟡 Shared sleep/continue tail is hard to reason about** — Added a comment at the shared tail explaining it's the deliberate common poll cadence for both empty cases (cold-start and all-excluded), and why. Kept it shared rather than duplicating into each branch — the cadence and budget-overrun guard are identical for both, so a shared tail is the DRY expression of that.

All 12 unit tests green, `cargo clippy -p calimero-node` clean.

PR #3061, opened 2026-06-29
fix(storage): tombstone whole subtree on delete so GC can reclaim it

chefsale:
## Problem

Deleting an entity only tombstoned the entity itself. `Index::remove_child_from` tombstoned the **direct** child and never descended into its subtree, and the receive path (`apply_delete_ref_action`) tombstoned only the directly-deleted id.

Consequence: a deleted node's descendants kept `deleted_at = None` **forever**. They were left as live orphans under a tombstoned ancestor, and tombstone GC — which only reclaims index rows that have `deleted_at` set — could never collect them. An unbounded index leak, and the reason `garbage_collect_tombstones` "can't reclaim subtrees."

## Fix

Add `Index::tombstone_descendants_of(root, deleted_at)`: a depth-first walk over the index `children` that tombstones every descendant, resolving each with the **same** canonical delete-vs-update tiebreak `apply_delete_ref_action` already uses for a directly-deleted entity — a descendant whose own `updated_at` is strictly newer than the delete nonce survives (the live-update-wins "zombie" case, applied transitively).

Called from **both** delete paths with one shared `deleted_at` (the nonce carried on the `DeleteRef` wire):
- local delete → `Index::remove_child_from`
- remote replay → `apply_delete_ref_action`

This symmetry is what guarantees convergence: the replica only ever receives a single `DeleteRef` (for the subtree root) and reconstructs the descendant tombstones by replaying the identical recursion with the identical `deleted_at`. Internal subtree parent-lists/hashes are intentionally not recomputed — the whole subtree detaches at the root, so none of it feeds a live hash.

### Relocate / re-key interaction (checked carefully)

`remove_child_from` is shared with the collection re-key path (`relocate_child_from`), so relocate now also tombstones the abandoned old subtree. This is **correct** and fixes a latent leak there: re-key moves data to **new deterministic ids** — recursively, which is the whole point of the `RekeyTarget` supertrait ("a `Mergeable` type that nests a collection must re-key it deterministically or it diverges permanently"). The old subtree is genuinely abandoned, and the re-key snapshot is captured **before** relocation, so no data is lost. Confirmed by the existing `reassign_deterministic_id_preserves_entries`, `converge`, and `rekey_record` integration tests, which all stay green.

## Tests

New (`tests/index.rs`, `tests/interface.rs`):
- local delete tombstones a full `root → a → b → c` chain
- `tombstone_descendants_of` tombstones descendants but leaves the root
- a descendant with a strictly-newer `updated_at` survives the subtree delete (tiebreak)
- GC reclaims all three subtree rows after retention (the live root survives)
- **two-store convergence**: originator's local delete and replica's single-`DeleteRef` replay reach identical subtree tombstone state

Full suite: `cargo test -p calimero-storage` (647 lib + integration) green, including re-key/reassign/converge/rekey_record. `cargo clippy -p calimero-storage --all-targets` clean.

## Scope

Subtree tombstoning is the prerequisite that makes tombstone GC *meaningful*. Wiring `garbage_collect_tombstones` into a production maintenance schedule is a separate concern (it remains a manually-invoked operation, by design) and is intentionally **not** in this PR.

### DIFF
--- a/crates/storage/src/index.rs
+++ b/crates/storage/src/index.rs
@@ -906,25 +906,107 @@ impl<S: StorageAdaptor> Index<S> {
 
     /// Removes and deletes a child from a collection.
     ///
-    /// Uses tombstone-based deletion. To move a child to a different parent,
-    /// just add it to the new parent instead.
-    pub(crate) fn remove_child_from(parent_id: Id, child_id: Id) -> Result<(), StorageError> {
+    /// Uses tombstone-based deletion. Deleting a child also tombstones its
+    /// entire subtree (see [`tombstone_descendants_of`](Self::tombstone_descendants_of)):
+    /// a non-recursive delete would leave descendant index rows live, so they
+    /// could never be reclaimed by tombstone GC and would linger as orphans
+    /// under a tombstoned parent.
+    ///
+    /// `deleted_at` is the caller's tombstone nonce — the same value carried on
+    /// the `DeleteRef` wire action — so the child, every descendant, and the
+    /// remote replicas that replay the `DeleteRef` all stamp the subtree with
+    /// one identical timestamp and converge.
+    ///
+    /// To move a child to a different parent, add it to the new parent instead.
+    pub(crate) fn remove_child_from(
+        parent_id: Id,
+        child_id: Id,
+        deleted_at: u64,
+    ) -> Result<(), StorageError> {
         let _mutation_guard = index_mutation_guard();
-        Self::delete_entity_and_create_tombstone(child_id)?;
+        // Tombstone the subtree first (while `child_id`'s `children` are still
+        // readable), then the child itself, then detach it from its parent.
+        Self::tombstone_descendants_of(child_id, deleted_at)?;
+        Self::delete_entity_and_create_tombstone(child_id, deleted_at)?;
         Self::update_parent_after_child_removal(parent_id, child_id)?;
         Self::recalculate_ancestor_hashes_for(parent_id)?;
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


chefsale:
Addressed in the latest commit.

**🟡 Two-pass / double index read + concurrent-add window** — Fixed. Collapsed `tombstone_descendants_of` into a **single** depth-first pass: each node's index is read once, we descend via the `children` read *before* tombstoning the node, and tombstone inline. Removes the separate collect-then-tombstone passes and any window between them. The reentrant guard is still held across the whole walk.

**🟡 Equal-timestamp tiebreak (`deleted_at == updated_at`)** — Documented explicitly in both the doc comment and the inline comment: equal-or-older is tombstoned (delete wins on the tie), matching `apply_delete_ref_action`'s `<` rule. (No behaviour change — it was already `<`.)

**🟡 Guard reentrancy contract** — Documented on `tombstone_descendants_of`: it acquires the reentrant guard internally and is safe to call whether or not a caller already holds it (`remove_child_from` does; `apply_delete_ref_action` doesn't).

**📝/💡 Tests use shared `MainStorage`** — Fixed. Each subtree test now uses its own `MockedStorage<N>` (2101/2102/2103), matching the GC test, so a panic in one can't leak tombstones into another.

**📄 Docs bot (AGENTS.md / CONTRIBUTING.md / architecture/*.html)** — This is an internal storage-layer behaviour change with no API/contract surface, so I've left those; the generated architecture pages are regenerated on merge.

Full `calimero-storage` suite (647) green, clippy clean.

PR #3063, opened 2026-06-29
fix(auth): fail closed when a node-bound token has no request host

chefsale:
## Problem

`validate_node_host` returned `Ok(())` when the request carried **neither** a `Host` nor an `X-Forwarded-Host` header. The entire host-comparison block was wrapped in `if let Some(request_host)`, so a `None` host fell straight through to the trailing `Ok(())`:

```rust
let request_host = headers.get("X-Forwarded-Host").or_else(|| headers.get("host"))...;
if let Some(request_host) = request_host {
    // ...all the binding checks...
}
Ok(())   // <- reached when no host header is present
```

The caller only invokes this for a **node-bound** token (one carrying a `node_url`). So a client could simply omit the host header and bypass node binding entirely — presenting a token minted for node A against node B. A header-stripping proxy/MITM achieves the same.

## Fix

Fail closed: a missing/undeterminable request host is now an error for a node-bound token.

```rust
let Some(request_host) = request_host else {
    return Err("Token is node-bound but the request carries no Host or X-Forwarded-Host header to validate against".to_owned());
};
```

Deliberately **narrow**:
- The internal-auth-service bypass (`host == "auth"` / `"auth:<port>"`) is unchanged — those requests carry a host.
- The `X-Forwarded-Host` path (reverse proxies) is unchanged.
- The token-`node_url` parse is **left to fall through**: `node_url` is not always a URL (it can carry a client name, e.g. `auth.rs` sets it from `client_name`), so an unparseable value is not rejected here. Only the missing-request-host binding-bypass is tightened.

All three call sites (`verify_token_string`, the refresh handler, and the reverse-proxy `validate_handler`) are real HTTP flows where legitimate clients/proxies send a host, so this does not affect valid traffic.

## Tests

- node-bound token with **no host header** → now fails closed
- node-bound token with a matching **`X-Forwarded-Host`** (no `Host`) → still validates
- **non-node-bound** token (no `node_url`) → unaffected by a missing host
- existing matching/mismatched-host and internal-auth-bypass tests stay green

`cargo test -p mero-auth` (69 tests) green, `cargo clippy -p mero-auth` clean.

### DIFF
--- a/crates/auth/src/auth/middleware.rs
+++ b/crates/auth/src/auth/middleware.rs
@@ -930,6 +930,80 @@ mod tests {
             .verify_token_from_headers(&headers_mismatch)
             .await;
         assert!(result.is_err(), "Token should fail with mismatched host");
+
+        // Test with NO host header at all: a node-bound token must fail closed,
+        // otherwise a client could strip Host/X-Forwarded-Host to bypass the
+        // node binding entirely.
+        let mut headers_no_host = HeaderMap::new();
+        headers_no_host.insert(
+            "Authorization",
+            HeaderValue::from_str(&format!("Bearer {access_token}")).unwrap(),
+        );
+
+        let result = token_manager
+            .verify_token_from_headers(&headers_no_host)
+            .await;
+        assert!(
+            result.is_err(),
+            "node-bound token must fail closed when the request carries no host header"
+        );
+
+        // X-Forwarded-Host (set by reverse proxies) is honoured too: a matching
+        // forwarded host with no `Host` header still validates.
+        let mut headers_fwd = HeaderMap::new();
+        headers_fwd.insert(
+            "Authorization",
+            HeaderValue::from_str(&format!("Bearer {access_token}")).unwrap(),
+        );
+        headers_fwd.insert(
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed.

**🟡 Unparseable `node_url` silently passes** — Documented the security contract explicitly at the parse site: only a URL-shaped `node_url` is host-bound; a non-URL `node_url` (e.g. a client name, as `api/handlers/auth.rs` sets) is intentionally *not* bound and validates for any host. The deliberate decision is now auditable rather than reading as an accidental fall-through. (Tightening that case would break `client_name`-as-`node_url` tokens, so it stays out of scope.)

**E2E Rust Apps Failed comment** — That was a transient earlier-run failure; all checks are green now (68 pass, incl. `SDK E2E (paired)` and every `Test (...)` e2e job). The auth fail-closed change only rejects node-bound tokens with *no* host header — legitimate flows (and the e2e harness) send one.

**📄 Docs bot (architecture/crates/auth.html)** — generated page, regenerated on merge; no source-of-truth doc change needed for a fail-closed tweak.

PR #3064, opened 2026-06-29
fix(auth,network,server,context,storage): secret/logging/error-handling hygiene

chefsale:
Addresses a batch of secret-handling, log-injection/PII, and error-context findings across auth, network, server, context, and storage. Full workspace CI clippy (`--all-targets -D warnings`) is clean; auth (68) and context-client (27) test suites pass.

## Findings addressed

**Secrets / High**
- **`VersionedSecret` leaked its cleartext signing secret via derived `Debug`** — replaced with a manual `Debug` that redacts `value` and keeps the non-sensitive metadata.
- **403-vs-401 decided by substring-matching the error message** — added a typed `AuthError::TokenRevoked`. Revoked tokens now map to 403 by *type*; renaming any message can't silently downgrade a revoked token to 401. Updated the emit sites (`jwt.rs`), the middleware, the token-validation handler, and the tests.

**Log injection / PII**
- **Unescaped user-controlled path/method on every auth failure** (CRLF/ANSI injection + flood) — sanitize control chars and switch to structured fields via a shared `utils::sanitize_for_log` (length-bounded).
- **Username logged unescaped** (`user_password`) — same sanitizer + structured field.
- **Gossipsub `{:?}`-dumped the raw message payload** (+ ANSI color) on a hot path — now logs only `message_id`/`source`/`topic`/`payload_len`.
- **Client key ids in a rotation debug log** — trimmed to short prefixes.
- **Guest-log host path wrote to raw stdout** — routed through `tracing` (the wasm path already goes through the host log syscall).
- **`eprintln!` in the certificate HTTP handler** — routed through `tracing::error!`.

**Error context**
- **`AuthError` variants discarded the source** — the error-wrapping variants (`StorageError`, `TokenGenerationFailed`) now carry a real `#[source]` instead of stringifying the cause.
- **Parse/`http::Error` causes dropped at the API boundary** (`map_err(|_| …)`) — hex/length causes are now logged server-side (client message kept generic), and the dropped `http::Error` on the file-response path is logged.
- **~20 distinct execute failures collapsed to one opaque `InternalError`** — added a static `InternalErrorKind` category (`Context`/`Group`/`Encryption`/`Runtime`/`Merge`/`Ipc`/`Other`) carried on the wire variant, so callers can distinguish failure classes. Each site already logs its distinct cause; a couple of duplicate `"failed to execute request"` messages were made specific.

## Scoping notes (proportionate calls, flagged for review)

1. **`AuthError` `#[source]`** was added to the variants that genuinely wrap a foreign error (storage / token-gen), not to the leaf-message variants (`InvalidToken`, `InvalidRequest`, `TokenExpired`, …) which have no nested cause. The wrapping sites use `e.into()` to box the real error (`Box<dyn Error + Send + Sync>`), preserving the typed chain.
2. **`InternalErrorKind`** is a coarse, server-internal category, not a full per-failure error enum. It changes the serde wire shape of `InternalError` (now `{ kind }`); acceptable pre-1.0 and the enum is `#[non_exhaustive]`. No code matches the variant, so nothing breaks.
3. **Log flood / rate-limiting** for the auth-failure path: sanitization removes the injection vector; rate-limiting/sampling is left to the tracing-subscriber layer rather than a bespoke limiter in the middleware (wrong layer for a clean fix).

### DIFF
--- a/crates/auth/src/api/handlers/auth.rs
+++ b/crates/auth/src/api/handlers/auth.rs
@@ -451,7 +451,7 @@ pub async fn validate_handler(
             // Add error type header for better client handling
             if matches!(err, AuthError::TokenExpired) {
                 error_headers.insert("X-Auth-Error", "token_expired".parse().unwrap());
-            } else if err.to_string().contains("revoked") {
+            } else if matches!(err, AuthError::TokenRevoked) {
                 error_headers.insert("X-Auth-Error", "token_revoked".parse().unwrap());
             } else {
                 error_headers.insert("X-Auth-Error", "invalid_token".parse().unwrap());
--- a/crates/auth/src/auth/middleware.rs
+++ b/crates/auth/src/auth/middleware.rs
@@ -9,6 +9,7 @@ use tracing::{debug, info, warn};
 
 use crate::auth::permissions::PermissionValidator;
 use crate::server::AppState;
+use crate::utils::sanitize_for_log;
 use crate::AuthError;
 
 /// Permissions of the authenticated caller.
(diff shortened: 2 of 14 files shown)


chefsale:
Thanks — addressed in `3edba4917`:

- **`sanitize_for_log` double scan** — fixed. It now resumes the same `chars()` iterator after `take(MAX_LEN)` and checks `chars.next().is_some()`, so truncation detection is O(1) instead of re-scanning the input with `chars().nth(MAX_LEN)`.
- **`StorageError` / `TokenGenerationFailed` string shape** — verified, no code change needed. I grepped every `.to_string().contains(...)` / `.contains("…")` against `AuthError` across `crates/auth` and `crates/server`: the only message-substring match is a test on the `InvalidToken` "Key not found" path (`middleware.rs`), not on these two variants. The one production substring matcher (`auth.rs` `…contains("revoked")`) was already converted to a typed `matches!(err, AuthError::TokenRevoked)` in this PR. So nothing relies on the `{0}` string shape of the boxed-source variants.

One related thing I checked while verifying: `crates/server/src/auth.rs::unauthorized_response` maps only `TokenExpired` and intentionally doesn't distinguish revoked (its doc notes revoked keys currently surface as "Key not found" via `get_key`, with the `token_revoked` arm tracked separately). That's a separate server integration, pre-existing, and outside this batch's findings, so I left it as-is rather than expand scope.

PR #3065, opened 2026-06-29
fix(network): add connection limits and cap the startup dial burst

chefsale:
## Problem

The swarm `Behaviour` had **no connection limits**, and `load_peer_cache_and_dial` (`discovery.rs`) fired a `swarm.dial()` at **every** fresh cached peer at once:

```rust
let candidates = self.peer_cache.dial_candidates(now, PEER_CACHE_TTL_SECS); // up to 4096
for candidate in candidates {
    self.swarm.dial(...);   // no throttle, no cap
}
```

The cache holds up to `MAX_PEER_CACHE_ENTRIES` (4096) entries. A full or **poisoned** cache turns startup into a self-inflicted dial storm that can exhaust file descriptors before the node is even useful — and nothing bounded total established connections either. (Pairs with the cache-size cap from #315/#322, which limited the *explosion* but not the unthrottled dials.)

## Fix — two layers

**1. `connection_limits::Behaviour` on the swarm** (`behaviour.rs`), with resource-exhaustion ceilings:

| limit | value | purpose |
|---|---|---|
| `max_pending_outgoing` | 128 | **dial-storm cap** — overflow dials are *denied*, not queued/opened |
| `max_pending_incoming` | 128 | inbound handshake-flood cap |
| `max_established_per_peer` | 8 | headroom for TCP+QUIC+relay→DCUtR, bounds one bad peer |
| `max_established` (total) | 1024 | bounds total open sockets/FDs |

These sit far above any healthy Calimero cluster (2–20 collaborating peers + a few bootstrap/rendezvous/relay links) and only bite under abuse. (`connection_limits` is part of libp2p-swarm core — no new Cargo feature needed.)

**2. Cap the startup cache redial at the source** (`discovery.rs` + `peer_cache.rs`): dial only the `MAX_STARTUP_CACHE_DIALS` (32) **most-recently-seen** peers (most-recent-first — the likeliest still reachable) via a new `PeerAddrCache::startup_dial_set`, logging how many candidates were dropped. Reconnect only needs the handful of peers we last collaborated with; rendezvous rediscovers the rest. The former `dial_candidates` becomes a test-only inspection helper.

## Tests

`startup_dial_set` caps to the most-recently-seen N, orders most-recent-first, and never includes a stale (past-TTL) peer. Full `calimero-network` suite (82) green, `cargo clippy --all-targets` clean, downstream `calimero-node` builds (the derived `BehaviourEvent` gains a no-op `connection_limits` variant — `connection_limits::Behaviour` never emits events).

## Note on configurability

The ceilings are intentionally fixed constants — a safety backstop, not a tuning knob. `SwarmConfig` is `#[non_exhaustive]`, so promoting them to config later is non-breaking if an infra node (bootstrap/relay) ever needs higher limits.

### DIFF
--- a/crates/network/src/behaviour.rs
+++ b/crates/network/src/behaviour.rs
@@ -12,8 +12,8 @@ use libp2p::request_response::{self, ProtocolSupport};
 use libp2p::swarm::behaviour::toggle::Toggle;
 use libp2p::swarm::{NetworkBehaviour, Swarm};
 use libp2p::{
-    dcutr, gossipsub, identify, kad, mdns, noise, ping, relay, rendezvous, tcp, tls, yamux,
-    StreamProtocol, SwarmBuilder,
+    connection_limits, dcutr, gossipsub, identify, kad, mdns, noise, ping, relay, rendezvous, tcp,
+    tls, yamux, StreamProtocol, SwarmBuilder,
 };
 use multiaddr::Protocol;
 use tracing::warn;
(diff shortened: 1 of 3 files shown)


chefsale:
Addressed.

**💡 Connection limits not configurable / track the follow-up** — Added a note at the constants spelling out that they're fixed by design and the exact path to make them tunable for bootstrap/relay nodes (promote to `SwarmConfig`, which is `#[non_exhaustive]`, with these as defaults). I left out an issue-number `TODO(#NNN)` on purpose — this repo's convention is comments that read standalone without ticket footnotes.

**📄 Docs bot (architecture/crates/network.html, wire-protocol.html)** — no wire-protocol change here (connection-count ceilings + a startup dial cap); generated pages regenerate on merge.

PR #3068, opened 2026-06-29
feat(xcall): node-enforced caller policy for #[app::xcall] entry points

chefsale:
## Motivation

Today an `#[app::xcall]` entry point can be invoked by **any** context in the same namespace, and the only authorization is whatever the target method hand-codes with `env::xcall_origin()`. That's **fail-open**: forget the check (or get it wrong) and a different app in the namespace can drive your entry point under the node's identity in your context.

This came out of scoping the "xcall executes as `members.first()`" item. The scoping showed the executor *must* be a member the node owns (a synthetic non-member can't execute and its writes are discarded), so the real weakness isn't the identity — it's that **caller authorization is delegated to every app author** instead of declared and node-enforced. This PR fixes that.

## What this does

A declarative, node-enforced caller policy on xcall entry points:

```rust
#[app::xcall]                  // any context in the same namespace (unchanged default)
#[app::xcall(from_same_app)]   // only contexts running the SAME application id
pub fn pong(&mut self, …) { … }
```

The node enforces `from_same_app` at the existing xcall gate, so it does **not** rely on the target also checking the origin. `env::xcall_origin()` remains for finer, data-dependent checks the declarative policy can't express.

## How

- **ABI** (`wasm-abi`): new `Method.xcall_callers` (`XCallCallers` enum, default `AnyInNamespace`, serde-skipped when default). The emitter parses the `from_same_app` arg from the source attribute. Fully back-compat — manifests without the field deserialize as `AnyInNamespace`, and the existing `xcall_callable` bool is unchanged.
- **Node** (`context`): the per-blob xcall cache maps method → policy. At the gate the node resolves the **caller's** application id (from `xcall_origin`, via `context_client.get_context`) and denies a `SameApp` entry point unless it matches the target's app id. An unresolvable caller is denied — **fail closed**. Enforcement stays entirely in the context crate; no runtime/VMContext threading. The decision is factored into the pure `xcall_caller_denied` helper and unit-tested.
- **Macro** (`sdk-macros`): `#[app::xcall]` accepts `from_same_app` and **rejects unknown args at compile time**, so a policy typo can't silently fall back to the open default.
- **Example**: `xcall-example`'s `pong` now uses `#[app::xcall(from_same_app)]` (ping and pong are the same app), demonstrating the node-enforced boundary; its `env::xcall_origin()` checks become app-specific provenance logic on top.

## Tests

- emitter: `from_same_app` → `SameApp`, bare `#[app::xcall]` → `AnyInNamespace`
- node: `xcall_caller_policy_decision` covers not-an-entry-point (deny), `AnyInNamespace` (allow any incl. unresolved), `SameApp` same-app (allow) / different-app (deny) / unresolved (deny, fail-closed)
- example app tests + full `wasm-abi` (all bins) and `calimero-context` (200) suites green; `calimero-node` builds; clippy clean across touched crates.

## Scope / follow-ups

This is **layer 1**: app-to-app, declarative, in the manifest. A natural **layer 2** is a governance-managed ACL for specific **context-id** pairs (runtime entities can't live in a compile-time manifest) — a separate, larger change. Default stays `AnyInNamespace` for back-compat; flipping to default-deny is a possible future tightening now that the mechanism exists.

### DIFF
--- a/apps/xcall-example/src/lib.rs
+++ b/apps/xcall-example/src/lib.rs
@@ -49,16 +49,19 @@ impl XCallExample {
         self.xcall_to(target_context, &method)
     }
 
-    /// Receive a pong via `xcall`. Marked `#[app::xcall(from_same_app)]` so the
-    /// node only admits callers running this same application — ping and pong
-    /// are the same app — and rejects a same-namespace context running a
-    /// different app before this method even runs.
+    /// Receive a pong via `xcall`. Marked `#[app::xcall]` so other contexts in
+    /// the same namespace may invoke it.
     ///
-    /// The node-enforced policy is the trust boundary; the `env::xcall_origin()`
-    /// checks below are the remaining app-specific logic — rejecting direct
-    /// calls (no origin) and verifying the origin matches the self-reported
-    /// `from_context`.
-    #[app::xcall(from_same_app)]
+    /// A tighter, node-enforced caller policy is available —
+    /// `#[app::xcall(from_same_app)]` restricts callers to contexts running this
+    /// same application. It's deliberately not used here yet: emitting the new
+    /// `xcall_callers` ABI field requires the downstream `abi-codegen` tool to
+    /// understand it first, so the example adopts it once that ships.
+    ///
+    /// `env::xcall_origin()` is set by the node and can't be forged, so this
+    /// method rejects direct calls (no origin) and refuses any call where the
+    /// origin doesn't match the self-reported `from_context`.
+    #[app::xcall]
     pub fn pong(&mut self, from_context: ContextId) -> app::Result<()> {
         let origin = calimero_sdk::env::xcall_origin().map(ContextId::from);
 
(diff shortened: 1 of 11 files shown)


chefsale:
Fixed the **ABI Contract** check failure (the only red check).

It failed because `xcall-example`'s `pong` used `#[app::xcall(from_same_app)]`, which emits the new `xcall_callers` ABI field — and the downstream `@calimero-network/abi-codegen` doesn't parse it yet. That's a genuine cross-repo ordering constraint (the tool must not get ahead of released core), not a bug here.

Resolution: `pong` keeps the plain `#[app::xcall]` (open default) for now, so no app emits `xcall_callers` and the contract check passes. **Everything else stays** — the `xcall_callers` schema field, node-side enforcement, the macro arg + its validation, and the emitter/node unit tests (the `from_same_app` parse + the `xcall_caller_denied` decision are both still covered).

The `xcall-example` adopts `from_same_app` in a small follow-up once abi-codegen ships support (prepared in mero-devtools-js#36), per the designed flow: core ships the field → tooling adopts it against a release → examples use it.

PR #3079, opened 2026-06-30
fix(auth): verify against backup signing secret + harden secret time/cache

xilosada:
## What

Hardens JWT signing-secret handling so secret rotation is safe to enable. Root of the stack #3079 → #3082 → #3083; rotation itself stays disabled here and is enabled in #3083.

Closes security-review finding **14** (verification ignores the backup secret → mass logout on rotation) and the staleness half of **21** (multi-instance cache never refreshed).

## Before → after

| | Before | After |
|---|---|---|
| Token verification | Only the current primary secret is tried; the backup is consulted only when the primary is entirely absent (disaster recovery). The instant a secret rotates, **every outstanding token fails** → fleet-wide logout. | `verify_token` first tries the primary, then falls back to any backup secret still inside its grace window (`SecretManager::get_verify_secrets`, primary first). Expiry/issuer/malformed errors stay terminal — only signature mismatch triggers the fallback. |
| Cross-process rotation | Each process caches secrets in memory forever; a rotation performed by another process is never observed and replicas diverge. | Storage is authoritative: `get_secret` re-reads the stored primary on every call and refreshes the cache on version mismatch, demoting the stale primary to a verify-fallback backup. |
| Clock handling | (superseded) this branch originally added a `Result`-based time helper; master's #3125 shipped the infallible `now_secs()` + saturating arithmetic instead, and this PR now uses that. | Same behavior as master — no panicking clock unwraps. |

## How to test

```
cargo test -p mero-auth
```

Targeted tests (all in `crates/auth/src`):
- `verify_token_succeeds_after_secret_rotation` — mint, rotate, verify still passes via backup (the mass-logout regression).
- `verify_token_fails_once_backup_is_evicted` — two rotations push the original secret out of grace → reject.
- `verify_token_rejects_unknown_secret` — foreign-signed token never verifies.
- `get_verify_secrets_returns_primary_only_initially`, `get_verify_secrets_includes_unexpired_backup_after_rotation` — verify-secret set shape.

Manual: on a running node, mint a token pair, force a rotation (`SecretManager::rotate_secret` or wait for the interval once #3083 enables the task), and confirm the pre-rotation access token still authenticates until the grace window lapses.

## Notes for reviewers
- Merged current master (zeroize-import fix + #3125 time hardening); the earlier CI failures were the pre-existing `calimero-primitives` build break on the stale base, not this change.
- #3082 and #3083 stack on this branch; merge order #3079 → #3082 → #3083.


frdomovic:
bump — still relevant. Closes security-review items **14** (verify against unexpired backup secret) and **21b** (checked epoch helper). Base of the #3082→#3083 stack; must land before rotation is enabled in #3083. Doc-bot advisory (auth.html) is handled by `update-docs` on merge.

PR #3080, opened 2026-06-30
fix(auth): encrypt signing secrets at rest and restrict storage permissions

xilosada:
## What

Encrypts JWT signing secrets at rest and locks down the storage directory. Closes security-review finding **15** (signing secrets stored base64-plaintext in a default-permission directory — any local user could read the HMAC secret and forge admin tokens). Independent of the token-semantics stack (#3079/#3082/#3083).

## Before → after

| | Before | After |
|---|---|---|
| Secret at rest | `URL_SAFE_NO_PAD` base64 inside a JSON blob — i.e. plaintext — in the auth RocksDB. | Sealed with **AES-256-GCM** before hitting disk. The key-encryption-key comes from `MERO_AUTH_SECRET_KEK` (env) or, for the persistent backend, a `0600` sibling key file (`<db-path>.kek`, 32 bytes) auto-generated on first run — resolved via `Storage::at_rest_kek()`. |
| Directory perms | `create_dir_all` under the process umask (~0755) — world-readable. | Storage directory created `0700`, secret/KEK files `0600`. |
| In-memory backend | n/a | No at-rest surface → stores plaintext by design, so two managers sharing one in-memory store interoperate (test fixture parity). |
| Restart behavior | — | The persistent KEK survives restarts; a wrong KEK fails decryption loudly instead of serving a corrupt secret. |

## How to test

```
cargo test -p mero-auth
```

Targeted tests:
- `keyfile_kek_is_persisted_0600_and_stable_across_managers` (unix) — production `with_storage_config` RocksDB path: two independently constructed `SecretManager`s over the same config interop across a simulated restart, and the sibling `.kek` file is exactly 32 bytes with mode `0600`. (Assumes `MERO_AUTH_SECRET_KEK` is unset, as in CI — the env KEK legitimately takes precedence.)
- Sealed round-trip through storage, wrong-KEK rejection, and in-memory two-manager interop tests.

Manual: start a node, then `strings` the auth RocksDB files — no base64 of the raw secret appears; `ls -la` the auth storage dir → `drwx------`, KEK file `-rw-------`.

## Compatibility
Existing plaintext-stored secrets are re-sealed on first write after upgrade; a node whose KEK file is deleted mints new secrets (outstanding tokens invalidate — same effect as a rotation without grace).


chefsale (approved):

chefsale (approved):

frdomovic:
bump — still relevant. Closes security-review item **15** (seal signing secrets at rest with AES-256-GCM + 0700/0600 storage perms). Independent of the token-semantics stack.

PR #3081, opened 2026-06-30
fix(auth): enforce password bounds, key expiry, and exact node-host binding

xilosada:
## What

Four independent hardenings in the user-password provider and key model. Closes security-review findings **4** (unsalted credential hash as key id), **17** (password bounds ignored), **18** (keys never expire), and **20** (prefix-match node binding). Independent of the #3079→#3082→#3083 stack.

## Before → after

| | Before | After |
|---|---|---|
| Password bounds (17) | `min/max_password_length` (defaults 8/128) existed in config but were **never enforced** — length-1 passwords accepted. | `validate_password_length()` (counts Unicode scalars, not bytes) enforced at both entry points: `authenticate_core` (login + bootstrap) and `create_root_key`. Out-of-bounds → clear error. |
| Key id (4) | `hex(SHA256("user_password:user:pass"))` — the **listed key id is an unsalted, GPU-fast offline brute-force target** for the credentials. | PBKDF2-HMAC-SHA256, 100k iterations, 32-byte output, deterministic per-user salt (`calimero:auth:user_password:key-id:v1:<username>`) — defeats cross-user precomputation and stretches per-target attacks. (Salt must be derivable at lookup time, hence per-user not random — documented in code.) |
| Key expiry (18) | `is_valid()` only checked `revoked_at` — root keys valid **forever**. | `KeyMetadata.expires_at: Option<u64>` with `is_expired()` enforced in `is_valid()`; `set_expires_at()`/`with_ttl_secs()` builders. `#[serde(default)]` → legacy persisted keys load as `None` = non-expiring (back-compat). |
| Node binding (20) | `is_valid_for_node` used `starts_with` — a key bound to `node.example.com` **matched `node.example.com.attacker.com`**. | `extract_host()` parses both sides (URL or bare `host[:port]`), lowercases, strips port, requires **equality**; unparseable host or node-bound key with no request host → reject (fail closed). JWT path mirrors it via the pure `validate_node_host_match` helper (also rejects empty/whitespace hosts; case-insensitive, since DNS is). |

## ⚠️ Migration note
Existing user_password key ids do **not** migrate — the id derivation changed, so previously registered users must re-register (bootstrap). Keys without `expires_at` keep working unchanged.

## How to test

```
cargo test -p mero-auth   # 114 lib + 6 integration tests green
```

Targeted tests per finding:
- **17**: `test_password_length_boundaries_inclusive` (min/max pass), `test_password_length_boundaries_exclusive` (min−1/max+1 rejected), `test_password_length_counts_unicode_scalars`.
- **4**: `test_derive_key_id_is_salted_not_plain_sha256`, `test_derive_key_id_differs_by_password` / `_by_username`, `test_derive_key_id_is_deterministic`.
- **18**: `test_expired_key_is_invalid`, `test_key_with_no_expiry_is_valid`, `test_legacy_key_json_without_expiry_loads` (serde round-trip of a pre-fix key).
- **20**: `test_node_binding_rejects_suffix_attack` (the exact `…attacker.com` case), `test_node_binding_absent_request_node_rejected`, `test_validate_node_host_match_absent_host_fails_closed`, plus master's `forwarded_host_spoof_is_rejected_cross_node`.

Manual:
1. 7-char password → "at least 8 characters"; 129-char → "at most 128"; 8/128 → success.
2. Bootstrap a user, dump the key store: root-key id must be 64 hex chars and ≠ `echo -n "user_password:<user>:<pass>" | shasum -a 256`.
3. Patch a key's `expires_at` into the past → auth fails; a pre-upgrade key without the field keeps working.
4. Token from `node.example.com` replayed with `X-Forwarded-Host: node.example.com.attacker.com` → 401; header stripped entirely → 401; correct host any case/port → 200.

## Notes for reviewers
- Merged current master; where master had meanwhile shipped its own fail-closed `validate_node_host` (#3063) the merge keeps master's messages/rationale and this PR's pure testable helper + case-insensitive exact compare. The earlier CI red was the stale-base `calimero-primitives` zeroize break.
- Residual from finding 7 (X-Forwarded-Host is still client-controllable when the auth port is directly reachable) is deliberately out of scope — see the follow-up comment on this PR.
- Overlaps `user_password.rs` with #3221 (bootstrap secret) — whichever lands second rebases trivially.


chefsale (approved):

frdomovic:
bump — still relevant. Closes security-review items **4** (salted-KDF key id), **17** (password bounds), **18** (key expiry), and **7/20** (exact node-host binding, fail-closed). Independent off master. Note: overlaps `user_password.rs` with #3221 (bootstrap-secret) — trivial rebase whichever lands first.

frdomovic:
One residual gap to track from security-review item **7** (out of scope for this PR, follow-up material): this PR hardens the node-binding *comparison* to an exact, fail-closed host match — but the compared host is still derived from client-controllable headers, with `X-Forwarded-Host` preferred over `Host` (`validate_node_host`). A caller that reaches the auth service directly (not via the intended reverse proxy) can present any `X-Forwarded-Host` it likes, so cross-node token replay remains possible in deployments where the auth port is directly reachable. Proper fix is a design change: derive the authoritative host from trusted per-node config, or only honor `X-Forwarded-Host` when the immediate peer is a trusted proxy (else use `Host` only). Filing this as a comment rather than widening the PR.

frdomovic:
## ⚠️ This PR bricked existing nodes — fixed in bb2bf096

Verified against the code: as written, **every node that already has a user was permanently locked out on upgrade**, and that is why 60 CI checks were red (not a harness quirk).

Two independent lockout vectors in `authenticate_core`:

1. **Key-id change had no migration.** `verify_credentials` looked the key up by the new salted-PBKDF2 id, but an upgraded node stores the root key under the **old unsalted-SHA256 id** → lookup misses → `list_keys(Root)` is non-empty → bootstrap branch skipped → `"Invalid username or password"`, forever. There was no recovery path: `/admin/keys` needs a token the operator can no longer obtain, and there is no merod CLI root-key command. Recovery would have meant wiping the auth DB.
2. **Minimum password length was enforced on the login path.** Any pre-existing password shorter than the (previously unenforced) minimum was rejected — including the `dev`/`dev` credentials used by **134 login steps across 57 e2e workflow files**.

### Fixes
- `verify_credentials` now falls back to the legacy id and, on a match, **re-keys the root key onto the salted id in place** (write-then-delete, so a crash can never lose the key). Transparent, one-time, per user.
- The **minimum** now applies only where a credential is **created** (bootstrap / `create_root_key`). The **maximum** is still enforced on the auth path, where it bounds PBKDF2 work per request.
- e2e harnesses bootstrap fresh nodes, so their credentials must satisfy the policy: the 134 login steps, the auth-seam script default and the SDK-e2e harness env now use an 8+ char password.

New tests: `legacy_key_id_is_migrated_in_place_on_login`, `wrong_password_does_not_migrate_or_authenticate`, `short_legacy_password_still_authenticates_after_upgrade`, `bootstrap_still_enforces_the_minimum_length`, `overlong_password_is_rejected_on_the_auth_path`. Suite: 119 passed. Branch is also up to date with master.

PR #3082, opened 2026-06-30
fix(auth): add and enforce token_type, bind refresh to access, derive live perms

xilosada:
> Stacked on #3079 (base branch: `auth-secret-verify`). Review/merge after it.

## What

Fixes token-semantics confusion in the JWT layer. Closes security-review findings **10** (refresh token accepted as an access token), **12** (refresh not bound to the access token's subject), and **19** (tokens carry a stale permission snapshot).

## Before → after

| | Before | After |
|---|---|---|
| Token confusion | `Claims` has no `token_type`; the `TokenType` enum is dead code. A 30-day **refresh token works directly as a Bearer access token** on any endpoint. | Every token carries a required `token_type` claim (`"access"`/`"refresh"`). `verify_token` accepts only access tokens; `verify_refresh_token` only refresh tokens. Pre-existing tokens without the claim are rejected (deliberate clean break — clients re-authenticate once on upgrade). |
| Refresh binding | `refresh_token_handler` requires the access token to be expired but then **discards its claims**; any valid refresh token + any unrelated expired access token mints a fresh pair for the refresh token's subject. | The handler decodes the access token (signature + `Access` type enforced, expiry ignored), applies an explicit must-be-expired policy, and rejects the exchange unless `access.sub == refresh.sub`. |
| Permissions | `verify_token_string` loads the live key but returns the token's **mint-time permission snapshot**; permissions removed from a key keep working until token expiry. | Effective permissions = token claims ∩ live key permissions, so a revoked permission stops working immediately. The claims act only as an upper bound. |

## Review-feedback fixes (meroreviewer)

- **Double decode on refresh**: the handler previously verified the access token twice (once with expiry, once without). It now decodes exactly once via `decode_access_claims_ignore_expiry` and applies its own expiry policy with a leeway constant matching jsonwebtoken's default, removing the redundant work and the theoretical TOCTOU between the two decodes.
- **Misleading name**: `verify_expired_access_claims` → `decode_access_claims_ignore_expiry`, documented explicitly that it does **not** assert the token is expired.
- **Empty secret set**: `decode_with_secrets` now logs an error and fails fast when no verification secrets are available (previously silently swallowed).
- **Per-verification HashSet**: the live-permission intersection uses a linear scan (keys hold a handful of permissions).

## How to test

```
cargo test -p mero-auth
```

Targeted tests:
- `refresh_token_rejected_as_access_token` / `access_token_rejected_in_refresh_slot` / `refresh_token_pair_rejects_access_token` — token-type enforcement in both directions (finding 10).
- `tokens_share_subject_accepts_same_subject` / `tokens_share_subject_rejects_mismatched_subject` — subject binding (finding 12).
- `verify_token_string_rederives_perms_from_live_key` — mint `[admin, context]`, strip `context` from the live key, verify → only `admin` (finding 19).
- `access_token_verifies_as_access` / `refresh_token_verifies_as_refresh` — happy paths.

Manual: `POST /auth/refresh` with `{access_token: <expired A>, refresh_token: <B's refresh>}` → 401 "do not belong to the same subject"; present a refresh token as `Authorization: Bearer` → 401.

## Compatibility

Existing tokens are invalidated on upgrade (missing `token_type`); clients re-authenticate once. This is intentional — grandfathering untyped tokens would keep the confusion window open for 30 days.


frdomovic:
bump — still relevant. Closes security-review items **10** (enforce token_type), **12** (bind refresh↔access subject), **19** (intersect live key perms). Stacked on #3079.

frdomovic:
Addressed all four meroreviewer findings in 6778e20a:

1. **Double decode of the access token** — `refresh_token_handler` now decodes the access token exactly once (`decode_access_claims_ignore_expiry`: signature + `Access` type enforced, expiry skipped) and applies its own explicit must-be-expired policy with `JWT_EXPIRY_LEEWAY_SECS = 60`, matching jsonwebtoken's default `Validation::leeway` so the two notions of expired can't disagree. The redundant `verify_token` pre-check and its TOCTOU window are gone.
2. **Misleading `verify_expired_access_claims` name** — renamed to `decode_access_claims_ignore_expiry` with a doc comment stating explicitly that it does NOT assert the token is expired and that any new caller must add its own expiry policy.
3. **Empty secrets list silently swallowed** — `decode_with_secrets` now fails fast with a `tracing::error!` when no verification secrets are available, so a misconfigured secret manager is visible in production logs.
4. **HashSet per verification** — the live-permission intersection now uses a linear `Vec::contains` scan; keys hold a handful of permissions so the allocation wasn't buying anything.

Also merged the updated base (`auth-secret-verify`, which now carries current master) — the earlier CI failures were the stale-base `calimero-primitives` zeroize build break, not this PR.

PR #3100, opened 2026-07-01
fix(node): single-pass, capped tombstone GC + reclaim tests

chefsale:
## What & why

While confirming that subtree-tombstoning GC (#3061) actually reclaims deleted subtrees, the investigation surfaced that the **real production GC is `crates/node/src/gc.rs`** — the storage-crate `Index::garbage_collect_tombstones` was never wired in and is test-only. `node::gc` had several rough edges this PR fixes, plus it had **zero test coverage**.

### GC efficiency (`node::gc`)
- **Single pass** over `ContextState` covering all contexts — `O(total keys)` — replacing the old full-table rescan *once per context* (`O(contexts × keys)`).
- **Per-run deletion cap** (10k, logged when hit) bounding write amplification and the collected key set; leftovers reclaim next cycle.
- **Off-reactor**: the blocking RocksDB scan runs via `spawn_blocking`, with a sweep-overlap guard and sweep-on-startup (mirrors the `DagCompactor` pattern).
- **Robust index-row detection**: `ContextState` keys are SHA-256 hashed, so the `Index`/`Entry` tag isn't recoverable from the key. A value is treated as a tombstone only if it borsh-decodes as `EntityIndex` **and** re-serializes to the exact same bytes — so GC can never delete entity data that coincidentally decodes.

### Dead-code removal
- Deleted the unused `Index::garbage_collect_tombstones` (the all-keys-into-a-`Vec` scan flagged as an unbounded-memory concern) and its test-only callers. The subtree-tombstoning behavior it exercised is now verified against the **real store** in `node::gc`; the storage-layer test asserts every subtree row is stamped reclaim-ready with the delete's nonce.

### Tests (previously none)
- `reclaims_all_tombstoned_subtree_rows` — subtree reclaim at the real store layer.
- `interrupted_sweep_is_consistent_and_next_pass_completes` — interrupt/resume.
- `backward_clock_skew_never_reclaims` + `forward_clock_reclaims_only_after_retention` — wall-clock retention guard.
- `non_index_values_are_never_reclaimed` — round-trip safety.

## Not addressed
End-to-end (wasm delete → GC reclaim) coverage is out of scope here; this adds unit coverage at the layer `node::gc` operates on.

## Verification
- `cargo build -p calimero-node` ✅
- `cargo test -p calimero-node --lib gc::` → 5/5 ✅
- `cargo test -p calimero-storage --lib` → 646/646 ✅
- `cargo clippy -p calimero-node -p calimero-storage --lib --tests` → clean ✅


chefsale:
Thanks — addressed the review in 1dbba029c. Triage:

**Fixed**
- **Sweep flag leak on cancellation** (`spawn_sweep`): the `sweep_in_progress` reset now lives in an RAII `SweepGuard` held inside the future, so it clears on cancellation/actor-teardown too, not just the normal completion path. (The panic/`JoinError` path already cleared it; this closes the drop-without-completion case.)
- **`HashSet` on the hot path**: replaced with an O(1) transition counter. The column is keyed `context_id ‖ state_key` so entries iterate grouped by context — a `context_id` change marks a new context. It's a log-only metric, so worst case it over-counts; it can't affect reclamation.
- **Canonical-borsh invariant**: added `entity_index_borsh_roundtrips` to lock the property the index-vs-data guard depends on, so a future non-canonical field (e.g. a `HashMap`) fails the test instead of silently weakening the guard. Also documented the assumption inline.
- **iter/get non-atomicity + false-negative caveat**: documented both — every TOCTOU outcome is safe (delete→`None`→skip, update→non-tombstone→skip; round-trip guard is the backstop), and a layout change leaks pre-change tombstones rather than mis-deleting them.

**Declined**
- **`age <= retention` → `age < retention`**: keeping `>`. "Reclaim when strictly older than the retention period" matches the documented contract and the prior `garbage_collect_tombstones` cutoff semantics (`deleted_at < now - retention`). The difference is one nanosecond at day-scale with no practical effect; changing it is a gratuitous behavior tweak.

All 6 gc unit tests pass; clippy + fmt clean.

chefsale:
Round 2 addressed in bafcfe87d.

**Fixed**
- **Guard vs. blocking task race** (real catch): moved `SweepGuard` *into* the `spawn_blocking` closure. A blocking task runs to completion even if its `JoinHandle` is dropped (actor teardown), so having the guard on the outer future could clear `sweep_in_progress` while the detached scan was still deleting — letting a new tick race the in-flight sweep. The flag now tracks the actual scan lifetime.
- **Best-effort deletes**: the deletion loop now counts *actual* removals and continues past a failed delete (logged) instead of aborting the sweep and stranding the remaining reclaimable tombstones. `collected` reflects work done, not intended.
- **Missed-insertion TOCTOU**: documented — harmless. A tombstone inserted past the cursor is skipped this pass, but a just-created tombstone isn't retention-eligible anyway and the next sweep catches it (GC is eventually consistent).
- **Layout-drift signal**: added a `debug!` when a value decodes as a tombstone but fails the borsh round-trip, so silent false-negative leaks are diagnosable.
- **`_handle` drop**: documented that dropping it doesn't cancel the task and the guard owns the flag-release invariant.

**Declined (again)**
- **`age <= retention` → `age < retention`**: keeping `>`. It matches the documented "older than the retention period" and the prior GC cutoff semantics (`deleted_at < now - retention`); the delta is one nanosecond at day-scale retention with no practical effect. Not worth a behavior change.

6/6 gc unit tests pass; clippy + fmt clean. Branch is up to date with master.

chefsale:
Round 3 addressed in 0ffccbc36.

**Fixed**
- **`now_nanos` u128→u64 truncation**: now `u64::try_from(as_nanos()).unwrap_or(0)` instead of an `as` cast that would wrap ~year 2554, keeping the documented broken-clock-skips-reclamation direction explicit.
- **Round-trip INVARIANT comment**: added a load-bearing note at `reserialized == value` pointing at `entity_index_borsh_roundtrips`.
- **Retention boundary**: added an explicit comment that the strict `>` is intentional and conservative (matches the storage-layer cutoff), so it isn't "fixed" to `<`.
- **Shutdown edge**: documented that if the runtime shuts down before the blocking closure runs, the flag stays set — benign, since no further ticks fire on a stopping actor.

**Declined**
- **`Vec::with_capacity(max_deletions_per_run)`**: would reserve ~10k slots on every sweep even though the overwhelming common case reclaims ~0 tombstones — that pessimizes the hot path. `Vec::new()` growing to actual need is the better trade.
- **"Old-nonce TOCTOU" (live-at-scan → tombstoned-before-`get`)**: not a defect. `get` reads the *current* value; if it's a tombstone whose `deleted_at` already exceeds retention, reclaiming it is correct per the retention contract. A lagging peer relies on full resync, not old tombstones — that's exactly why `FULL_RESYNC_THRESHOLD_NANOS > TOMBSTONE_RETENTION_NANOS` is asserted in `storage::constants`.
- **`contexts_scanned` ordering assumption**: already documented as metrics-only (worst case over-counts, never affects reclamation).

6/6 gc tests pass; clippy + fmt clean.

chefsale:
Round 4 addressed in 7000b9c04 (merged with your master-merge, now at a6d385510).

**Fixed**
- **Resurrection TOCTOU (collect→delete window)**: each key is now re-validated against its CURRENT value immediately before deletion. The collect phase snapshots keys earlier in the same scan, so a concurrent resurrection (tombstone → live re-add) in that window could otherwise delete a now-live row. GC now deletes only while the row is still a reclaimable tombstone, shrinking the race to a single read+delete. (`get` returning `None` or a non-tombstone → skipped.)
- **Canonical-borsh constraint at the definition site**: added a doc comment on `EntityIndex` (crates/storage/src/index.rs) warning that new fields must be borsh-canonical, pointing at the `entity_index_borsh_roundtrips` test that locks it.
- **Shutdown/restart flag**: clarified in the comment that a freshly-constructed actor gets a fresh `Arc<AtomicBool>` (run.rs builds a new `GarbageCollector`), so the stuck-flag edge can't survive a restart.

**Declined**
- **Second outer guard for the shutdown edge**: that would reintroduce the exact early-clear race fixed in round 2 (guard dropping while the detached blocking scan is still deleting). The blocking-closure guard is correct for normal operation; the only uncovered case is runtime-shutdown-before-task-start, which is benign per above.

6/6 gc tests pass; clippy + fmt clean.

PR #3108, opened 2026-07-01
fix(merod): reject symlinked home before `init --force` delete (TOCTOU)

chefsale:
## Problem

`merod init --force` deleted the existing node home with `fs::remove_dir_all(&path)` right after the `ConfigFile::exists` / `ConfigFile::load` checks (`crates/merod/src/cli/init.rs`). `remove_dir_all` **follows a symlink at the final path component**, so an attacker who swaps `path` for a symlink between those checks and the removal (a classic TOCTOU race) could redirect the recursive delete onto an arbitrary target directory.

## Fix

Immediately before the removal, re-stat the path with `fs::symlink_metadata` (lstat — does **not** follow the final symlink) and require a real directory. On a symlink or any other non-directory, `bail!` instead of deleting:

```rust
let metadata = fs::symlink_metadata(&path)
    .await
    .wrap_err_with(|| format!("failed to stat existing path {path:?}"))?;
if !metadata.is_dir() {
    bail!(
        "refusing to remove {path:?}: expected a directory, found {}",
        if metadata.is_symlink() { "a symlink" } else { "a non-directory" }
    );
}

fs::remove_dir_all(&path).await?;
```

This shrinks the check-to-use window to a single lstat right before the delete and refuses outright when the target isn't a genuine directory.

## Testing

- `cargo build -p merod` — clean
- `cargo clippy -p merod` — clean

### DIFF
--- a/crates/merod/src/cli/init.rs
+++ b/crates/merod/src/cli/init.rs
@@ -257,12 +257,17 @@ impl InitCommand {
                 return Ok(());
             }
 
-            // Refuse to recurse through a symlink. `remove_dir_all` follows a
-            // symlinked final component, so an attacker who swaps `path` for a
-            // symlink between the checks above and this removal could redirect
-            // the recursive delete onto an arbitrary target directory. Re-stat
-            // the path without following symlinks (lstat) and require a real
-            // directory immediately before deleting.
+            // Refuse to delete anything but a real directory here. `path` is the
+            // node home we're about to wipe; if it has been replaced by a symlink
+            // (or a plain file) since the checks above, fail loudly with a clear
+            // message rather than silently touching an unexpected target. Re-stat
+            // without following symlinks (lstat) right before the removal.
+            //
+            // This lstat only narrows an attacker's race window, it does not
+            // close it. The hard guarantee comes from `remove_dir_all` itself:
+            // it opens the final component with `O_NOFOLLOW`, so a symlink swapped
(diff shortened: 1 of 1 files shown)


chefsale:
Thanks for the review — addressed in `4ad75a540`.

**Finding 1 (missing error context on `remove_dir_all`):** fixed. The call now uses `.wrap_err_with(|| format!("failed to remove existing node home {path:?}"))?`, matching the surrounding fallible calls.

**Finding 2 (residual TOCTOU window):** the remaining window does **not** allow deleting an unintended target, so no fd-based rewrite is needed. `std::fs::remove_dir_all` (which `tokio::fs` calls) opens the final path component with `O_NOFOLLOW` since the CVE-2022-21658 hardening. I verified the actual behavior on rustc 1.88:

```
remove_dir_all(symlink_to_dir) result: Ok(())
real target still exists: true
keep.txt still exists: true
```

A symlink swapped in after the lstat is unlinked in place — its target is never recursed into or deleted. So the hard guarantee already lives in `remove_dir_all`; the lstat we added is an early, explicit, clearly-messaged rejection for the common (accidental) case rather than a silent symlink-unlink. I've rewritten the comment to state this accurately instead of implying the lstat alone closes the race.

**Doc-bot note (`architecture/crates/tools.html`):** this change is internal hardening of `init --force` cleanup with no change to the CLI surface (flags, args, output), so I don't think the tools doc needs an update. Happy to add a line if you'd prefer.

Issue #3111, opened 2026-07-01
Namespace-join open_stream timeout leaves the dial running in the background (uncancelled)


chefsale:
## Summary

When a namespace-join stream open times out, the in-flight libp2p dial/stream-open is **not** cancelled — it keeps running in the background on the `NetworkManager` arbiter until it resolves on its own. This is a low-severity inefficiency (the orphan self-cleans, see below), split out from a batch of async/concurrency hardening fixes so the rest could ship.

## Where

- `crates/node/src/sync/manager/namespace_join.rs:181` — `time::timeout(open_timeout, sync_network.open_stream(*peer))`

The same pattern applies to any caller that wraps `NetworkClient::open_stream` in a timeout.

## Trace / root cause

`NetworkClient::open_stream` (`crates/network/primitives/src/client.rs:154`) sends `NetworkMessage::OpenStream { request, outcome: tx }` to the `NetworkManager` actor, then awaits `rx`. The message is dispatched through `forward_handler` (`crates/utils/actix/src/adapters.rs:127`), which drives the handler's `ResponseFuture` (`crates/network/src/handlers/commands/open_stream.rs`) on the actor **independently of the caller's `rx`**.

So when the caller's `open_stream` future is dropped on timeout:
- Only the caller-side `rx` is dropped.
- The actor's spawned `ResponseFuture` keeps running `stream_control.open_stream(peer_id, CALIMERO_STREAM_PROTOCOL).await` to completion — the dial is **not** cancelled and is bounded only by libp2p's own dial resolution.

## Why it's low severity (not a leak)

When the background dial finally resolves, the opened `Stream` is constructed and then immediately dropped, because the `tx.send(...)` fails (the receiver is gone). Dropping the `Stream` closes it. So the orphan **self-cleans** on completion — it is a lingering background future, not a permanent resource leak.

## Options considered

1. **Handler-side dial timeout** — wrap `stream_control.open_stream(...)` in a `tokio::time::timeout(...)` inside the `OpenStream` handler so a hung dial self-cancels at the network layer. Contained, but: (a) affects **all** `open_stream` callers, and (b) picking a value is largely redundant with libp2p's configured dial timeout and would be a guess.
[...]

PR #3112, opened 2026-07-01
fix(async): hardening pass on task/timeout lifecycles across node, server, sync

chefsale:
## What

Eight independent async-correctness fixes for hazards where a slow or cancelled future stalled a loop, dropped work mid-flight, or leaked tasks. Each is self-contained; grouped here because they came out of one review pass.

| Sev | Fix | File |
|-----|-----|------|
| H | Process each TEE eviction on its own detached task so a slow namespace-root hard-purge can't stall the recv loop and let the broadcast channel lag/drop subsequent events (un-purged key residue with no recovery) | `crates/context/src/self_purge.rs` |
| H | Snapshot `(ContextId, DeltaStore)` and drop the DashMap iterator **before** awaiting `cleanup_stale`/`pending_stats`, so a shard read-guard is no longer held across `.await` (was blocking new-context registration on that shard) | `crates/node/src/manager/startup.rs` |
| M | On sync-session timeout, re-await the pinned session under a bounded grace (`session_timeout / 2`) instead of dropping it mid-protocol (storage/DAG divergence); if the grace also elapses, drop as before so the concurrency permit is released and other sessions aren't starved. Responder + initiator | `crates/node/src/sync_session_bridge.rs` |
| M | Track the node-events/commands/health tasks + per-frame `JoinSet` (reaped each iteration) and abort them all on disconnect so none outlive the socket | `crates/server/src/ws.rs` |
| M | Raise the `get_blob_bytes` mailbox enqueue timeout 10ms → 1s so handoff under NodeManager load no longer false-drops the actor path and reads the blob twice | `crates/node/primitives/src/client/blob.rs` |
| M | Fuse the sync driver arm (no re-poll panic if it resolves) and abort the server/bridge handles on shutdown instead of detaching them | `crates/node/src/run.rs` |
| L | Serve inbound blobs off the NodeManager arbiter (`tokio::spawn` vs `ctx.spawn`) — blob serving can occupy the actor for up to 5 min | `crates/node/src/handlers/stream_opened.rs` |
| L | Drop per-chunk send timeouts in favour of the single outer serve deadline, and `close()` the stream explicitly, so the peer sees a clean end-of-stream instead of a decode error from a half-written frame | `crates/node/src/handlers/blob_protocol.rs` |

## Notes on the sync-session change

The current code deliberately **drops** the session on timeout (see the #2319 comments) to bound how long a concurrency permit is held. Blindly re-awaiting to completion would reintroduce permit starvation on a truly-stuck session. The bounded-grace approach keeps both properties: total wall time stays ≤ 1.5× `session_timeout` (under the 2× watchdog grace), a slow-but-progressing session finishes cleanly, and a genuinely deadlocked one is still dropped so its permit is freed.

## Deferred

A related lower-severity item — the namespace-join `open_stream` timeout leaves the underlying dial running uncancelled in the background (it self-cleans when the dial resolves, so it is not a leak) — is tracked separately in #3111 rather than fixed here, since a proper fix needs cancellation-propagation plumbing disproportionate to the severity.

## Testing

- `cargo check` / `cargo clippy` clean on `calimero-node`, `calimero-server`, `calimero-context`, `calimero-node-primitives`.
- Test targets for `calimero-node` and `calimero-context` compile.

### DIFF
--- a/crates/context/src/self_purge.rs
+++ b/crates/context/src/self_purge.rs
@@ -210,7 +210,22 @@ async fn run(store: Store, node_client: NodeClient) {
         // module docstring for why we don't also react to
         // `OpEvent::MemberRemoved` here.
         if let Some((group_id, member)) = dispatch_target(&event) {
-            handle_member_removed(&store, &node_client, group_id, member).await;
+            // Process each eviction on its own detached task rather than inline.
+            // A namespace-root hard-purge (subtree cascade + per-group
+            // signing-key deletes + gossipsub unsubscribe) can be slow; running
+            // it inline stalls `rx.recv()`, and a slow enough purge lets the
+            // bounded broadcast channel lag and DROP subsequent events. A
+            // dropped `TeeMemberRemoved` leaves un-purged key residue that the
+            // marker-gated startup reconcile sweep cannot recover (see the
+            // `Lagged` arm above). Spawning keeps the recv loop hot so the
+            // channel does not back up. The purge helpers are idempotent and
+            // distinct evictions target distinct groups, so concurrent purges
+            // are safe. `Store` and `NodeClient` are cheap handle clones.
+            let store = store.clone();
+            let node_client = node_client.clone();
+            tokio::spawn(async move {
+                handle_member_removed(&store, &node_client, group_id, member).await;
+            });
         }
     }
 }
--- a/crates/context/src/self_purge.rs
+++ b/crates/context/src/self_purge.rs
@@ -225,9 +225,17 @@ async fn run(store: Store, node_client: NodeClient) {
             // The `JoinHandle` is intentionally dropped: `handle_member_removed`
             // returns `()` (no error to propagate — per-step failures are
             // already surfaced via `record_purge_failure` + logs inside it), and
-            // a panic in the task still fires the process panic hook, so it is
-            // not silently lost. Detaching also isolates a panicking purge from
-            // the recv loop, which is the point of spawning here.
+            // a panic in the task fires the process panic hook (unless a custom
+            // hook swallows it), so it is not silently lost. Detaching also
+            // isolates a panicking purge from the recv loop, which is the point
+            // of spawning here.
+            //
+            // Not joined on shutdown: a namespace-root purge interrupted by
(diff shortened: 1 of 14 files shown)


chefsale:
Rebased onto latest `master` to resolve the conflict.

**WS change dropped.** The original batch included a WebSocket task-abort fix (M2), but #3103 landed the same task-leak hardening first, with a better-fitted design (a graceful `oneshot` shutdown for the node-events task). My hard-abort approach would additionally have interrupted `handle_commands`' graceful WebSocket close handshake, so I dropped it entirely and took master's version of `crates/server/src/ws.rs`. This PR is now 7 fixes (node/context/sync/blob), not 8.

Also folded in the earlier review feedback (spurious grace-window timeout warnings, run.rs completed-handle re-poll, and matching the blob enqueue/response timeouts).

Rebuilt clean against the new master; `Cargo.lock` untouched.

chefsale:
Addressed the latest review round in `24ff99c`:

- **Auth-rejection send still used a per-send timeout** (suggestion) — ✅ removed it, consistent with the data path; that send isn't under the outer `BLOB_SERVE_TIMEOUT` anyway. Dropped the now-unused `CHUNK_SEND_TIMEOUT`.
- **Grace re-await holds the permit across the grace window** (warning) — this is the intended, bounded tradeoff (permit held ≤ 1.5× `session_timeout`, still under the 2× watchdog grace; free for a session that timed out while only *waiting* for a permit). It read like an oversight because the comment didn't spell it out — ✅ documented the tradeoff explicitly. No behavior change (this was the deliberately chosen option over dropping mid-protocol vs. unbounded re-await).
- **self_purge detached tasks not joined on shutdown** (inline) — the durability concern is already backstopped: the pending-self-purge marker is written *before* the cascade and `reconcile_sweep` completes any unfinished purge on next startup (same guarantee as crash-mid-cascade). ✅ noted this in the comment rather than adding a JoinSet drain.
- **`sse/handlers.rs` bind_event_task race** (warning) — out of scope; that file isn't part of this diff (it's untouched master code from the SSE work).

Clippy clean; `cargo fmt` pre-commit passed.

PR #3115, opened 2026-07-01
fix(node): graceful shutdown, real health/ready, and store durability hardening

chefsale:
Implements the eight reported shutdown/durability/observability issues (1 High, 5 Medium, 2 Low) on top of latest `master`.

### [H] Node SIGTERM/SIGINT handling — `crates/node/src/run.rs`
Adds a `shutdown_signal()` future (Ctrl-C + unix SIGTERM, mirroring the auth server) as a new `tokio::select!` arm. On signal the run loop performs a bounded, ordered drain: mark node not-ready → cancel the HTTP server (graceful) and await → drain the network-event bridge → abort+await background tasks → flush the datastore. A `SHUTDOWN_GRACE` (10s) bounds each stage so a wedged subsystem cannot block exit indefinitely.

### [M] RocksDB flush — `crates/store/impl/rocksdb/src/lib.rs`, `crates/store/src/{db,lib}.rs`, encryption
Adds `impl Drop` for `RocksDB` (flush_wal + flush) as the abrupt-path backstop, plus a `flush()` method on the `Database` trait (default no-op) overridden by RocksDB, delegated by `EncryptedDatabase`, and exposed as `Store::flush()` for the deterministic shutdown call.

### [M] Real `/health` + `/ready` — `crates/server/src/…`
Adds a `NodeReadiness` lifecycle signal (Starting/Ready/ShuttingDown) threaded into `AdminState` and driven by `run::start`. `/health` becomes a store-ping liveness check (503 on a wedged store); new `/ready` returns 503 unless the node is Ready **and** the store responds. Node lifecycle is the correct signal for a k8s probe — the governance readiness FSM tiers are per-namespace with no single node-wide tier.

### [M] `session_deadline` validation — `crates/merod/src/cli/validation.rs`
Adds a duration-range check (1s–10min). `session_deadline_ms = 0` now fails at startup instead of wrapping every session's timeout to fire immediately (silent permanent divergence).

### [M] Zeroize encryption key — `crates/store/src/config.rs`, `crates/merod/src/cli/run.rs`
`StoreConfig.encryption_key` is now `Option<Zeroizing<Vec<u8>>>`; merod wraps the fetched KEK in `Zeroizing` so it is wiped from the heap instead of lingering for the process lifetime. (`KeyManager` was already `ZeroizeOnDrop`.)

### [M] Graceful HTTP shutdown — `crates/server/src/lib.rs`
`axum::serve(...).with_graceful_shutdown(...)` driven by a `CancellationToken` fanned out from `run.rs`, so in-flight requests drain rather than the server task being dropped.

### [L] Detached background tasks retained — `crates/node/src/run.rs`
The metrics-tick, peer-identity snapshot/invalidation, and local-delta drainer handles are retained and abort+awaited on shutdown (the drainer was previously block-scoped and detached). RocksDB writes in these tasks are synchronous, so an abort only lands at an `.await` boundary — never mid-write.

### [L] Config validate / readiness flap / gauges
- `merod config` now runs the full `validate_config` on the candidate (not just deserialize), so it cannot persist values that would fail `merod run`.
- `read_local_applied_through` takes a `fallback` and returns the last-known sequence on a transient store `Err` instead of demoting to 0 (no spurious readiness regression).
- The linux-only `process_*` gauges are `#[cfg(target_os = \"linux\")]`-gated so they are not registered as constant-0 series on other platforms.

## Verification
- `cargo check --workspace --all-targets` — clean (0 errors/warnings).
- Unit tests green: `calimero-store`, `calimero-store-rocksdb`, `merod` (55 + integration), `calimero-node` readiness (30).
- `cargo clippy` on the changed crates — clean.
- `Cargo.lock` changes are exactly the three new dependency edges (tokio-util→node, zeroize→store, zeroize→merod); no version churn.

## Known limitation
Shutdown drains HTTP + background tasks and flushes the store, but does not explicitly quiesce the Actix actor system before the flush — the `Drop`-based RocksDB flush is the backstop for any last actor writes. Explicit `System::stop()` + await could be a follow-up if stricter quiescing is wanted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1483,6 +1483,7 @@ dependencies = [
  "tempfile",
  "thiserror 1.0.69",
  "tokio",
+ "tokio-util",
  "tracing",
  "zeroize",
 ]
@@ -1776,6 +1777,7 @@ dependencies = [
  "strum 0.26.3",
  "thiserror 1.0.69",
  "thunderdome",
+ "zeroize",
 ]
 
 [[package]]
@@ -6074,6 +6076,7 @@ dependencies = [
  "tracing-subscriber",
  "url",
  "x509-cert",
+ "zeroize",
 ]
 
 [[package]]
--- a/crates/context/src/handlers/join_group.rs
+++ b/crates/context/src/handlers/join_group.rs
@@ -22,6 +22,16 @@ use calimero_governance_store::op_events::OpEvent;
 
 const NAMESPACE_MESH_GRACE: Duration = Duration::from_secs(2);
 
+/// Upper bound on how long the background re-publisher keeps trying to announce
+/// membership when the joiner's namespace mesh is too thin at join time (e.g. it
+/// joined while another member was down). Generous enough to span a downed
+/// peer restarting and the mesh reforming; bounded so a genuinely partitioned
+/// joiner's task still exits.
+const MEMBER_JOINED_REPUBLISH_WINDOW: Duration = Duration::from_secs(120);
+
+/// Poll cadence while waiting for namespace mesh readiness.
+const MEMBER_JOINED_POLL_INTERVAL: Duration = Duration::from_millis(500);
+
 // Maximum time `join_group` waits on the gossip-fallback path for a
 // `KeyDelivery` op addressed to the joiner now lives on
 // [`ContextManagerConfig::key_delivery_fallback_wait`] so operators can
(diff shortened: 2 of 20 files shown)


chefsale:
Thanks — addressed the reviewer findings in `ea58f42b6`:

- **Server handle not aborted on graceful-shutdown timeout** (real fix): on the `SHUTDOWN_GRACE` timeout the HTTP server (and now the bridge) `JoinHandle` is `abort()`ed and reaped rather than dropped. A dropped `JoinHandle` only *detaches* the task, so it could otherwise keep running and write past the flush/return.
- **SIGTERM "silently ignored" on registration failure**: clarified in comments. If `signal()` registration fails, *no* handler is installed, so the kernel's **default** SIGTERM disposition (terminate) still stops the process — it degrades to abrupt termination (no graceful drain), it is not ignored. Parking the arm with `pending()` is intentional; resolving it would fire a false graceful shutdown with no signal received.
- **Background-task abort vs. in-progress store writes**: verified none of the aborted tasks (`metrics_tick`, `peer_identity_*`, `drainer`) use `spawn_blocking` — their RocksDB writes are synchronous and inline, so a cancellation is only observed at an `.await` boundary after any in-progress write returns. Updated the comment to state this and that a task panic surfaces as a logged non-cancelled `JoinError`.
- **NodeReadiness u8/ordering/masking suggestion**: relaxed the atomics from `SeqCst` to `Release`/`Acquire` (a lone flag needs no total order) and added a `debug_assert!` on an unknown state value so a future state added without updating `label()`/`is_ready()` trips in tests instead of silently mapping to "starting".
- **System-stop exit path bridge drain**: comment now notes the bridge notify/await is best-effort on that path (arbiters may already be gone) — this matches the known-limitation note in the PR description.

`cargo fmt`/`clippy` clean, `calimero-node` + `calimero-server` compile, server unit tests green.

The `E2E Rust Apps` failure (scaffolding-e2e, xcall-example) looks unrelated to these changes (no shutdown/health surface exercised there; `/health` keeps its prior 200 shape and `/ready` is additive) — happy to dig in if it reproduces after a rerun.

chefsale:
## ⚠️ Flagging for maintainer input: this PR reliably trips a governance fork/sync-masking race in `group-governance-peer-down`

Heads up before merge — CI is green on everything except the `group-governance-peer-down` e2e, and after digging into the node logs I don't think this is a flake or a defect in this PR's *logic*. It looks like a **timing perturbation that reliably triggers a pre-existing governance convergence race**. Details below for whoever owns governance internals.

### Not a flake — deterministic on this branch
- `group-governance-peer-down` passes **11/11** on recent `master` commits (incl. this PR's base `78f4ebe65` and my merge base `d94246709`).
- It fails **3/3** reruns on this PR's merge commit `d613e79b2`.
- `git diff d94246709..HEAD` is **only** this PR's intended files — no accidental changes to governance-store / sync / dag.

### What actually fails (from the exported node logs)
- **Context (application) data converges** — `sync::manager` logs `root hashes match, already in sync` (`local_root == remote_root`, 123×).
- **Governance/group state forks at the same sequence:**
  - node-1: `readiness beacon emitted applied_through=4 strong=true`, `get_group_info` OK.
  - node-2: `governance catch-up complete ... ops_received=4 ops_applied=4`, `readiness beacon emitted applied_through=4 strong=true`, tier `Bootstrapping → PeerValidatedReady`.
  - **But the group hashes differ**: node-1 `f5f9c2fc…` vs node-2 `fc15abb7…` — both at `applied_through=4`, both `strong=true`.
- Because both nodes advertise *strong / seq 4*, neither reconciles — the readiness beacon **masks** the fork. This matches the known-open *MemberJoinedAt ack-race + sync-masking* behavior.

### Why I don't think it's this PR's logic
Every runtime path this PR changes is **inert** during the failing window:
- `read_local_applied_through` Err-fallback WARN: **0×** (the changed branch never executes).
- `/health` (store ping) and `/ready`: **0×** requests.
- node-3 is **network-isolated, not SIGTERM'd**, so the new signal/graceful-shutdown/flush code never runs.
- No changes to governance apply, membership, DAG, or beacon emission.

So the fork isn't logically caused by the diff — the most plausible read is that the PR's extra startup task-wiring / the `signal`-feature runtime machinery shifts scheduling just enough to make a **pre-existing** governance apply-order/determinism + sync-masking race reproduce on this branch (master wins it).

### Ask
Since this appears to be a real, latent governance race (the beacon masks a same-sequence fork) rather than a bug in the shutdown/health/store changes here, I'd appreciate a maintainer's call on:
1. Whether to gate/quarantine `group-governance-peer-down` while the sync-masking race is addressed separately, and merge this PR on the green remainder; or
2. Whether the sync-masking (beacon advertises `applied_through`+`strong` without a state-hash check, so a same-seq fork goes unreconciled) should be fixed first.

Happy to CI-bisect exactly which change shifts the timing, or to help on the underlying race, whichever is preferred. Full log excerpts available on request.

chefsale:
## ✅ Root-caused and fixed — the `group-governance-peer-down` failure was a real governance membership race this PR exposed

Update on the e2e that was failing only on this branch. It is **not** a defect in the shutdown-hardening changes — it's a latent `MemberJoinedAt` propagation bug that the PR surfaced by shifting runtime scheduling. All CI is now green (CI, SDK E2E, E2E - Rust Apps, Release, ABI Guard).

### Root cause (proven, not guessed)
The `group-governance-peer-down` workflow stops **node-2**, then **node-3 joins the namespace while node-2 is down**, then node-2 restarts. On join, `join_group` adds node-3 locally and announces `MemberJoinedAt` over gossip. `sign_and_publish_namespace_op` hard-fails its transport gate `assert_transport_ready(mesh, known, n_low)` where `required = min(n_low, known_subscribers)`:

- node-3 can only reach node-1 → **`mesh=1`**
- but the **stopped node-2 still counts as a known subscriber** → **`known=2`**
- so `required=2 > mesh=1` → the announce is **gated out**, logged non-fatal, and dropped (fire-and-forget).

Consequence: node-1 never applies node-3's `MemberJoinedAt` (verified: 0 applications on node-1). node-3 then syncs the authoritative group state (which excludes it) and **reconciles its own local membership away** → permanent `get_group_info: "node is not a member"` → `wait_for_sync` across all three nodes times out. App/context data still converged; only group membership was lost. Master passes the identical scenario only because timing lands so node-3 sees `known=1` (stale node-2 subscription already pruned).

### How the PR exposed it
A bisection probe that replaced the in-loop `shutdown_signal()` future with `pending()` made the e2e pass: registering tokio's signal driver on the main worker runtime adds a wakeup source to its reactor and shifts task-poll ordering, which is enough to change the interleaving of the join vs. subscription-pruning. So the trigger is scheduling, but the bug is the fire-and-forget announce.

### The fix (`crates/context/src/handlers/join_group.rs`)
At announce time, check transport readiness:
- **ready** → publish inline, exactly as before (happy path unchanged);
- **not ready** → hand off to a bounded background task (`MEMBER_JOINED_REPUBLISH_WINDOW = 120s`) that polls until the mesh reforms (e.g. once the down peer restarts) and publishes then — so membership propagates **without blocking the join**.

Safety: publishes at most once per readiness (the op is written to the local log and reaches peers via sync regardless of ack count), so no duplicate-nonce ops; the signing key is copied and rebuilt into a `ZeroizeOnDrop` `PrivateKey` inside the task and dropped when it ends.

This is a concrete reproducer + fix for the known-open `MemberJoinedAt` ack-race — worth a look from whoever owns governance internals. Happy to split this fix into its own PR if you'd prefer to review it separately from the shutdown-hardening batch.

PR #3116, opened 2026-07-01
fix(dag,blob,projection,store): harden DAG/blob/projection internals

chefsale:
Batch of correctness/robustness fixes to the DAG, blob, projection, and store-iterator internals, from a set of reported issues.

## Blob (`crates/store/blobs`)
- **[H] Unbounded/cyclic `Blob::new` recursion** → iterative work-list DFS with a visited set + depth/node caps. A deeply-nested or cyclic (untrusted, synced) meta graph can no longer overflow the stack or loop forever.
- **[M] Chunks served without re-hashing** → leaf chunks are re-hashed against their content-addressed id before serving; a tampered/corrupt on-disk or peer chunk now fails a content-hash check (`IntegrityMismatch`) instead of being served as authentic.
- **[L] Zero-byte blob → `DanglingBlob`** → empty blobs (no stored file) read back as empty.

## DAG (`crates/dag`)
- **[L] `prune_to_recent` → doomed backfill** → track pruned-applied ancestor ids (bounded FIFO); a reference to one counts as an already-satisfied parent, so a later delta on a deliberately-deleted parent applies instead of stranding in pending and backfilling a parent no peer holds.
- **[L] `get_deltas_since` diamond re-emission across pages** → documented (bandwidth waste only; applying twice is idempotent). A real fix needs a wire/API change to carry the returned-id set in the cursor — intentionally deferred.

## Projection (`crates/projection`)
- **[L] `SubgroupReparented`/`SubgroupVisibilitySet` don't assert existence** → both now set `exists`, so a reparent/visibility op folding before its create no longer transiently hides a live subgroup (a later delete still wins by its higher stamp).
- **[M] Maintained projection folds governance at generation 0** → documented non-authoritativeness (the sanctioned option). The maintained/streaming projection is deterministic and convergent (safe as a sync signal) but not causally authoritative for `hlc=0` governance ops; `acl_view_at` remains the authoritative resolver. A recompute-from-log fix was rejected: `scope_root_for` is on the sync hot path over an as-yet-unbounded op-log.

## Store (`crates/store`)
- **[L] `IterEntries` mispairs error-key with value** → fuse on a key decode/iteration error and stop pairing the failed key with a value read from a mismatched/errored position.
- **[L] Default `delete_range` buffers whole range** → delete in bounded batches with a re-seek loop, so non-RocksDB backends stay memory-bounded.

## Deliberately not changed
- **[M] `validate_path_component` panics on bad name** — already fixed on master: it returns `Result` and every blob-path entry point propagates it.
- **[M] `can_apply` vacuously true for empty parents** — the suggested guard was implemented, then reverted: this codebase uses **empty parents as the genesis-op convention** (the unified-op genesis carries `parents: []` and must apply as a root), so rejecting empty parents strands legitimate roots and breaks the context suite. Documented inline in `can_apply` instead.

## Testing
Tests added for each behavioural change. Full workspace builds; green suites: `calimero-dag`, `calimero-projection`, `calimero-blobstore`, `calimero-store`, `calimero-context`, `calimero-authz`, `calimero-op-adapter`, `calimero-governance-store`, `calimero-node` (lib).

### DIFF
--- a/crates/dag/src/lib.rs
+++ b/crates/dag/src/lib.rs
@@ -33,6 +33,13 @@ pub const MAX_DELTA_QUERY_LIMIT: usize = 3000;
 /// in a future sync.
 pub const MAX_PENDING_DELTAS: usize = 10_000;
 
+/// Maximum number of pruned-ancestor ids remembered by [`DagStore::prune_to_recent`]
+/// so a later delta that references a deliberately-dropped parent is treated as
+/// applicable instead of triggering a doomed backfill. Bounded FIFO: once an id
+/// ages out, a reference to it falls back to state-based sync (the peer answers
+/// "not found", which is the cue for HashComparison/Snapshot).
+pub const MAX_PRUNED_TRACKED: usize = 100_000;
+
 /// Type of delta - regular operation or checkpoint (snapshot boundary)
 #[derive(Clone, Debug, PartialEq, Eq, BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
 pub enum DeltaKind {
@@ -270,6 +277,19 @@ pub struct DagStore<T> {
     /// Root delta (genesis)
     root: [u8; 32],
 
+    /// Ids of applied deltas that [`Self::prune_to_recent`] deliberately dropped
+    /// from `deltas`/`applied`. A later delta naming one of these as a parent
+    /// must not be reported as missing — backfilling a parent we intentionally
+    /// deleted just strands the delta (the peer returns not-found). We instead
+    /// treat a pruned ancestor as an already-satisfied parent: it was applied
+    /// before pruning, so its own ancestry was applied too, which is exactly the
+    /// causal guarantee `can_apply` needs. Bounded FIFO via `pruned_order`.
+    pruned: HashSet<[u8; 32]>,
+
+    /// Arrival-order index into `pruned` for O(1) FIFO eviction once the set
+    /// reaches [`MAX_PRUNED_TRACKED`].
+    pruned_order: VecDeque<[u8; 32]>,
+
     /// Maximum number of items returned by query methods to prevent resource exhaustion.
     /// Even if a caller requests more, the DAG will cap the result at this size.
     /// By default, equal to `MAX_DELTA_QUERY_SIZE`.
@@ -298,6 +318,8 @@ impl<T: Clone> DagStore<T> {
             next_pending_seq: 0,
             heads,
             root,
+            pruned: HashSet::new(),
+            pruned_order: VecDeque::new(),
             delta_query_limit: MAX_DELTA_QUERY_LIMIT,
             max_pending: MAX_PENDING_DELTAS,
         }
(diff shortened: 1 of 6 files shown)


chefsale:
Addressed the review in d514913 (blob traversal + `delete_range`). Rundown:

**Valid — fixed:**
- **Cycle check rejected duplicate/shared leaves.** Correct, and it was a real regression: identical file content hashes to the same part id, so a valid blob's `links` can name the same leaf twice. The global `visited` set flagged the second occurrence as `CorruptGraph`, so a repeated-content file failed to read back. Reworked into a streaming cursor DFS where `visited` tracks **internal** nodes only (where cycles actually live — `put` never shares an internal node); duplicate leaves are now served each time, true back-edges still rejected. Added a two-identical-chunk round-trip regression test. This also restores the old walk's memory profile (one cursor per internal node on the path, not every child's meta loaded eagerly).
- **Depth check placed after the leaf branch.** Moved so it guards every internal node before it recurses.
- **`delete_range` re-seeking to `lo` each batch.** Now resumes from the last-deleted key, so each seek stays local to where the previous one stopped.
- **`delete_range` not atomic.** Documented — per-key deletes mean a mid-range error leaves a partial deletion (same as the prior buffer-everything default).

**Noted — no code change:**
- The "batch ends exactly at `hi` → one extra re-seek" case is correctly identified as perf-only; the empty-batch `break` handles it. Not worth special-casing.

The doc-bot flags on `architecture/*.html` are for generated architecture docs; these are internal-hardening changes with no schema/interface impact, so I've left the generated docs alone — happy to regenerate if you'd prefer.

Full workspace builds; blobstore/store/dag/projection/context/node-lib suites green.

chefsale:
Second review pass addressed in b38ffe67a.

**Fixed:**
- **`delete_range` resuming from a just-deleted key.** Valid latent risk — the resume relied on the backend's seek landing *after* an absent (deleted) key, which the `DBIter` contract (`first key >= target`) doesn't actually guarantee for a future non-BTreeMap/non-RocksDB backend (could early-terminate, or loop). Now resumes at the **byte-successor** of the batch's last key (`last ++ [0x00]`), which is strictly greater than every key just deleted, so progress depends only on the documented `>=` semantics. Termination is now backend-independent.
- **`IterEntries` value-slot sentinel.** Documented the contract on the type: on a key error the iterator fuses and yields `(Err(key_error), Err(sentinel))`; callers must inspect the **key result first**, and the value slot is meaningful only when the key is `Ok`. (Note: the idiomatic `k?` before `v?` already surfaces the real key error first; the doc makes that explicit for callers that read the value slot independently.)
- **Blob depth off-by-one.** Switched the guard to `depth >= MAX_BLOB_DEPTH` so the enforced internal-node cap matches the constant exactly.

**Already handled in the prior pass** (these findings re-fired against the first commit's line numbers): the global-visited cycle check → now internal-node-only (duplicate leaves served), depth check moved before recursion, and `delete_range` non-atomicity documented.

Clippy `-D warnings` clean; store + blobstore suites green.

PR #3124, opened 2026-07-01
fix(node): re-drive persisted-but-unapplied deltas on restart (B1)

chefsale:
## Bug

`load_persisted_deltas` treated every persisted row as already-applied and restored each via `restore_applied_delta` (topology only, no execution), consulting the persisted `applied` flag nowhere. A Phase-0 `applied: false` row — written before the atomic heads commit and then orphaned by a crash (or a gossip delta buffered before its parents arrived) — was therefore promoted to *applied* and inserted as a DAG head on restart **without ever running its actions**, advertising committed state that was never written.

Surfaced by the `#[ignore]`d regression test added in #3120 (`phase0_applied_false_row_not_promoted_on_restart`).

## Fix

Partition persisted rows by the `applied` flag:
- **committed rows** (`applied: true`) keep the topology-only `restore_applied_delta` path;
- **`applied: false` rows** are re-driven through `add_delta` after the applied topology is restored (so their parents are present). `add_delta` is the same path a gossip delta takes — it executes the actions (or pends the delta if a parent is still missing) and only marks the delta applied once the apply succeeds. Applied-status once again implies the effects actually ran.

CRDT actions are idempotent by HLC/LWW, so re-driving a delta whose effects partially landed converges rather than double-counting.

Un-ignores the B1 regression test.

## Verification

- `crash_recovery_test`: both tests pass (`phase0_...` now green; `merge_topology_...` unaffected).
- `calimero-node` lib: **425 passed, 0 failed**.
- `dag_persistence` (7) and `concurrent_branches_root_hash` (2) integration tests: green.

Stacked on #3120 (the test PR); retargets to `master` when that merges.

### DIFF
--- a/crates/node/src/crash_recovery_test.rs
+++ b/crates/node/src/crash_recovery_test.rs
@@ -117,10 +117,28 @@ async fn phase0_applied_false_row_not_promoted_on_restart() {
     // Restart: fresh DeltaStore over the same store, then reload from disk.
     // Keep a clone so we can read the persisted row's flag after reload.
     let (delta_store, _tmp, _rx) = delta_store_over(store.clone()).await;
-    let _ = delta_store
+    let result = delta_store
         .load_persisted_deltas()
         .await
         .expect("reload from persisted rows");
+
+    // An applied:false row is re-driven through the apply path, NOT restored as
+    // applied, so it must never surface `pending_handler_events` — those are
+    // harvested only for committed (applied:true) rows. Replaying an uncommitted
+    // row's handlers would fire effects for state that never landed.
+    assert!(
+        result.pending_handler_events.is_empty(),
+        "an applied:false row must not surface phantom handler events on restart"
+    );
+    // The re-drive was ATTEMPTED (the row went to the unapplied re-drive path,
+    // not the applied-topology restore). Its parent is genesis, so `add_delta`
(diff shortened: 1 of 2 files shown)


chefsale:
Rebased onto the updated `test/cat-b-crash` base — the conflict is resolved (PR is `MERGEABLE` again) and the diff is unchanged: still just `crash_recovery_test.rs` and `delta_store.rs`. crash_recovery (2) and delta_store (23) tests pass on the new base.

On the review suggestion (`1e7fe6cc2893`, assert the re-driven delta is in the pending queue): this is already covered, and the literal `count == 1` doesn't hold for this test. The re-drive test runs without a WASM runtime, so `add_delta_internal` attempts the apply immediately (parent is genesis), the apply cannot complete, and the delta is left **neither applied nor pending**. The round-2 commit asserts exactly that plus the core regression guard:

- `pending_stats().await.count == 0` — "a re-driven delta whose apply cannot complete is not left dangling in pending",
- `dag_applied == db_applied` and `!dag_applied` — the DAG must not promote the row to applied while the DB row stays `applied:false` (the pre-fix bug),
- the delta must not become a DAG head.

So the re-drive attempt is asserted (via the applied-state agreement and no-phantom-events checks) without relying on a transient pending count that this no-runtime setup never reaches.

PR #3125, opened 2026-07-01
fix: harden integer casts, clock arithmetic & Eq/Ord contract

chefsale:
## Summary

Correctness fixes for a batch of reported cast/arithmetic/contract issues. Most guard against silent `u64 -> usize`/`u32` truncation on 32-bit (wasm32) targets; the rest fix wall-clock underflow, a shift-width panic, and an `Eq`/`Ord` contract violation.

Two of the originally reported items were already fixed on `master` and are **not** in this PR:
- governance-store pagination `limit == 0` (early-return already present).
- `decode_index_pairs` value-length path (already uses `checked_add`).

## Fixes

| Area | File | Change |
|---|---|---|
| network | `specialized_node_invite.rs` | Bound write-side length to the same `MAX_…MESSAGE_SIZE` the read side enforces + encode via `u32::try_from`, so an oversized payload can't truncate the framing prefix. |
| runtime | `logic.rs` | `usize::try_from` guest-supplied lengths in the read/check/write guest-memory helpers (report named read; check & write had the identical truncation). |
| runtime | `host_functions/storage.rs` | `storage_index_scan` limit uses `usize::try_from(...).unwrap_or(usize::MAX)`, matching the already-saturated offset. |
| blobs | `store/blobs/lib.rs` | `Size::hint()` saturates the u64→usize conversion (mis-sized the links vector on 32-bit); dropped the stale `const`/`cast_possible_truncation` allow. |
| server (SSE) | `sse/session.rs`, `sse/handlers.rs` | `saturating_sub` on persisted-timestamp expiry checks (backward NTP step no longer underflows). |
| auth | `auth/secrets.rs` | `saturating_sub` (rotation check) + `saturating_add` (expiry). |
| merod | `kms_policy.rs` | KMS backoff uses `checked_shl` so retries past the u64 shift width clamp instead of panicking. |
| storage | `entities.rs` | `UpdatedAt` derives `PartialEq` (delegates to inner `u64`) instead of an always-true `eq` that violated the `Eq`/`Ord` contract and made the equal-vs-newer branch in `save_internal` unreachable. |
| storage | `address.rs` | Corrected `Path::depth()` doc to its actual `segments - 1` semantics (code/tests/callers all rely on this; changing the value would break the model). |

### Note on `UpdatedAt`

`Metadata` derives `PartialEq`, so the always-true `eq` was silently making `Metadata` equality ignore the timestamp. Only test code compares `Metadata` by `==`. The full sync_sim convergence suite (314 tests) passes with the change — nodes converge including HLC timestamps.

## Tests

- Added regression tests: `policy_fetch_backoff` saturation past the shift width, and `UpdatedAt` eq/cmp agreement.
- `cargo check` + `cargo clippy` clean on all affected crates.
- Passing: storage (653), sync_sim (314), runtime, auth, blobstore, network-primitives, merod, server.
- `Cargo.lock` untouched.

### DIFF
--- a/crates/auth/src/secrets.rs
+++ b/crates/auth/src/secrets.rs
@@ -99,13 +99,20 @@ impl fmt::Debug for VersionedSecret {
     }
 }
 
+/// Seconds since the UNIX epoch. Degrades to `0` if the system clock is set
+/// before 1970 rather than panicking, so a misconfigured clock can't crash the
+/// auth service. Pairs with the saturating arithmetic on the values it feeds.
+fn now_secs() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_secs()
+}
+
 impl VersionedSecret {
     /// Create a new versioned secret
     pub fn new(secret_type: SecretType, rotation_config: &SecretRotationConfig) -> Self {
-        let now = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_secs();
+        let now = now_secs();
 
         // Generate a secure random secret
         let secret: [u8; 32] = rand::thread_rng().gen();
@@ -124,10 +131,7 @@ impl VersionedSecret {
 
     /// Check if this secret has expired
     pub fn is_expired(&self) -> bool {
-        let now = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_secs();
+        let now = now_secs();
         self.expires_at < now
     }
 }
@@ -279,10 +283,7 @@ impl SecretManager {
             .iter()
             .find(|s| s.secret_type == secret_type && s.is_primary)
         {
-            let now = SystemTime::now()
-                .duration_since(UNIX_EPOCH)
-                .unwrap()
-                .as_secs();
+            let now = now_secs();
             // Saturate so a backward wall-clock step (NTP correction) that puts
             // `now` before `created_at` can't underflow into a spurious early
             // rotation (or panic in debug builds).
--- a/crates/auth/src/secrets.rs
+++ b/crates/auth/src/secrets.rs
@@ -114,7 +114,9 @@ impl VersionedSecret {
             value: URL_SAFE_NO_PAD.encode(secret),
             version: format!("v{now}"),
             created_at: now,
-            expires_at: now + rotation_config.grace_period,
+            // Saturate so a far-future clock plus a large grace period can't
+            // overflow the u64 expiry.
+            expires_at: now.saturating_add(rotation_config.grace_period),
             is_primary: true,
             secret_type,
         }
@@ -281,7 +283,10 @@ impl SecretManager {
                 .duration_since(UNIX_EPOCH)
                 .unwrap()
                 .as_secs();
-            if now - secret.created_at >= self.rotation_config.rotation_interval {
+            // Saturate so a backward wall-clock step (NTP correction) that puts
+            // `now` before `created_at` can't underflow into a spurious early
+            // rotation (or panic in debug builds).
+            if now.saturating_sub(secret.created_at) >= self.rotation_config.rotation_interval {
                 drop(secrets);
                 self.rotate_secret(secret_type).await?;
             }
--- a/crates/merod/src/kms_policy.rs
+++ b/crates/merod/src/kms_policy.rs
@@ -667,8 +667,9 @@ mod tests {
         // saturated backoff instead of panicking on the `1 << exponent`.
         let far = policy_fetch_backoff(1_000);
         assert_eq!(far.as_millis(), u128::from(u64::MAX));
-        // And it must not panic at the exact boundary either.
-        let _ = policy_fetch_backoff(65);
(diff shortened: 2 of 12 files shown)


chefsale:
Addressed the review findings in b88ffd134:

- **`specialized_node_invite.rs` (dead `u32::try_from` error path):** Added a compile-time invariant `const _: () = assert!(MAX_SPECIALIZED_NODE_INVITE_MESSAGE_SIZE <= u32::MAX as u64);` next to the constant, so the write path now bounds against the explicit 1 MiB cap and casts with a plain `buf.len() as u32` (lossless by the asserted invariant). This removes the unreachable error path while keeping the tight cap, and guards the reviewer's "if MAX ever grows past u32::MAX" concern at compile time. The read-side `u32::from_be_bytes(..) as usize` is lossless on all targets (u32 always fits usize), so it's unchanged.
- **`kms_policy` backoff test:** The exponent==64 boundary case now asserts the saturated value (`policy_fetch_backoff(65).as_millis() == u128::from(u64::MAX)`) instead of only checking it doesn't panic.

Re: the doc-bot mapping notes — these are internal correctness/overflow hardening fixes with no change to documented architecture or wire behavior, so no `architecture/*.html` updates are warranted.

chefsale:
Addressed the latest round in ac3040443:

- **Comment clarity (`c82bd0cd3873`):** Reworded the length-prefix comment to attribute each bound to its real source — the *runtime* size check limits `buf.len()` to `MAX_SPECIALIZED_NODE_INVITE_MESSAGE_SIZE`, and the *compile-time* assert limits that constant to `u32::MAX`; together they make the `as u32` cast lossless.
- **`u32::try_from` at the cast site (`29367fbdad68`):** Declining this one. Reintroducing `u32::try_from` here is exactly the dead error path the first review round flagged (`3b7660856eac` / `bcef0dcf5285`) — after the runtime `> MAX` return and the compile-time `MAX <= u32::MAX` assert, the conversion can never fail, so its error arm would be unreachable. The compile-time assert already fails the build if the invariant is ever broken, which is stronger than a runtime `try_from` that would silently keep a now-unreachable branch. Keeping the plain cast + assert.

The read-side `u32::from_be_bytes(..) as usize` is lossless on every target (u32 always fits usize), so it's unchanged.

chefsale:
Addressed the latest review (`review-69b2f756`) in 81429f28d:

- **`UpdatedAt` PartialEq behavioral change (both warnings):** Added `save_internal_selects_branch_by_updated_at` to `tests/crdt.rs`, which drives all three timestamp branches explicitly against a dedicated node — older write (must stale-skip, nonce unchanged), equal-timestamp write (must merge, entity stays present), and strictly-newer write (must win by LWW, nonce advances). The strictly-newer branch is exactly the one that was unreachable under the always-true `eq`, so this locks in the corrected value-based `PartialEq`. Also expanded the `UpdatedAt` doc comment to spell out the consequence for callers (two `Metadata` differing only in `updated_at` now compare unequal; only test convergence assertions compare `Metadata`, so this is intended, not a regression).
- **Backoff practical cap (suggestion):** Good call — clamped `policy_fetch_backoff` to `POLICY_FETCH_MAX_BACKOFF_MS` (60s) via `.min(...)` before building the `Duration`, so a large `attempt` caps at a sane ceiling instead of ~585M years. Updated the test to assert the cap.

storage suite: 656 passing (incl. the new test); sync_sim convergence: 319 passing; clippy clean on both crates.

chefsale:
Addressed `review-777770f8` in d6c9c9c80:

- **`SystemTime::now()...unwrap()/.expect()` panic on a pre-epoch clock (warnings 1 & 2):** Good catch — it was inconsistent to fix the underflow subtraction but leave the timestamp source panicking. Replaced the three `duration_since(UNIX_EPOCH).unwrap()` sites in `auth/secrets.rs` with a shared `now_secs()` helper that degrades to `0` via `unwrap_or_default()`, and changed SSE `now_secs()` from `.expect(...)` to the same. Callers already subtract saturatingly, so a `0` reading just makes a value look maximally stale rather than crashing the service.
- **`attempt as u32` truncation (suggestion):** Switched to `u32::try_from(attempt).unwrap_or(u32::MAX)` for consistency with the PR's cast-hardening theme.
- **Backoff cap-onset clarity (`3853fdcad2fb`):** The test now pins where the 60s `.min(cap)` first engages — attempt 8 (32s, last under cap) vs attempt 9 (clamped) — and keeps the shift-width extremes (attempt 65 = exponent 64, and `usize::MAX`). Comment updated to say the cap engages via `.min` well before any shift concern.
- **Read-side cast comment asymmetry (`dfe55feee9de`):** Added the `// u32 always fits usize` note on both read paths.
- **Metadata SemVer / caller audit (warnings 3 & 4):** Ran the requested audit — the only whole-`Metadata` `==` in the entire tree is `sync_sim/assertions.rs:62` (a test convergence assertion); no production path (dedup/set/HashMap key) relied on the old always-true `eq`. `calimero-storage` is pre-1.0 internal (`v0.0.0`), so no version bump per project convention. The `UpdatedAt` doc already spells out the behavioral change.

auth (43), merod, server (56) suites green; clippy clean incl. `--tests`.

PR #3128, opened 2026-07-01
fix: wasm guest validation, blob refcounting, and hardening batch

chefsale:
Resolves five outstanding review findings. Each was verified still-present on latest `master` before fixing.

## [H] Wasm prepare/validate pass before compile
`Engine::compile` had a `todo!` block instead of validation. Untrusted guests could import host memory or define a `_start` entry point. Now `validate_guest_module` runs after compile and rejects:
- modules that **import** linear memory (guests must define and export their own — the host reads state via `exports.get_memory("memory")`);
- modules that export `_start` (a WASI *command* entry point emitted by `wasm32-wasi` toolchains; Calimero guests are `wasm32-unknown-unknown` libraries invoked by explicit method name).

Uses wasmer's existing `Module::imports()`/`exports()` introspection — **no new dependency**. New `FunctionCallError::ModuleValidationError`.

## [H] Blob delete had no refcount → cross-context data loss
Blobs are content-addressed and deduplicated, so identical content across contexts, or a chunk shared by two root blobs, resolves to one stored entry. The old delete path removed files + metadata unconditionally, destroying content other owners still referenced.

Added `refs` to `BlobMeta`. **Add** increments the root and every chunk; **delete** decrements them and removes files/metadata only at zero. The node-side `delete_blob` now delegates to one refcount-aware, tree-aware call instead of manually expanding root→chunks and deleting eagerly.

> ⚠️ Changes `BlobMeta`'s borsh layout — existing on-disk blobs won't deserialize. Acceptable pre-1.0 (dev data rebuilds).

## [M] `--mock-tee` guard only covered the Phala provider
`has_real_attestation` now destructures `KmsConfig` with no rest pattern, so adding a second KMS provider **fails to compile** until it's folded into the predicate — the mock-deny guard can't silently stop covering a provider.

## [L] `verify_root_hash` never called
Removed the dead public wrapper (+ registry impl). Sync integrity is unaffected: `reconciler.rs` already gates on the peer's **signed** expected root hash. `compute_root_hash` (the actually-used machinery) is retained.

## [L] Host-error policy + unenforced registers invariant
- Replaced the host-error `todo!` with a documented, deliberate policy: host errors surface as ordinary call failures and never panic the node (panicking would be guest-triggerable → DoS; it also fights the `catch_unwind` boundary). On-chain recording is a separate concern.
- Added `VMLimits::validate_invariants()` enforcing `max_registers_capacity >= max_register_size`, called in `Engine::new` so a self-contradictory config fails loud at startup instead of misbehaving per execution.

## Tests
- runtime: reject imported-memory, reject `_start`, accept a conforming guest; registers-invariant valid-by-default + `should_panic` on violation.
- blobstore: duplicate content survives one owner's delete (read back byte-for-byte); a chunk shared across two roots survives a sibling root's deletion.

All touched crates build, pass `cargo clippy` clean, and the new + existing tests pass.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -94,12 +94,17 @@ impl TeeConfig {
     /// must be mutually exclusive on a single node.
     ///
     /// Returns `false` when no KMS provider is configured (e.g. `kms.phala`
-    /// absent), which permits `--mock-tee`. If a second KMS provider is added,
-    /// extend this predicate to cover it.
+    /// absent), which permits `--mock-tee`.
+    ///
+    /// The `KmsConfig` is destructured field-by-field with no `..` rest pattern
+    /// on purpose: adding a new provider field to `KmsConfig` will fail to
+    /// compile here until it is folded into this predicate. That keeps the
+    /// `--mock-tee` deny-guard exhaustive across every provider rather than
+    /// silently ignoring a newly-added one.
     #[must_use]
     pub fn has_real_attestation(&self) -> bool {
-        self.kms
-            .phala
+        let KmsConfig { phala } = &self.kms;
+        phala
             .as_ref()
             .is_some_and(|phala| phala.attestation.enabled && !phala.attestation.accept_mock)
     }
--- a/crates/context/primitives/src/client/mod.rs
+++ b/crates/context/primitives/src/client/mod.rs
@@ -854,44 +854,6 @@ impl ContextRegistry {
         Ok(())
     }
 
-    /// Verifies that the stored root hash matches the actual state.
-    ///
-    /// Computes the root hash from storage and compares with the claimed hash.
-    /// Returns Ok(()) if they match, or an error describing the mismatch.
-    ///
-    /// # Arguments
-    ///
-    /// * `context_id` - The ID of the context to verify.
-    /// * `claimed_hash` - The hash to verify against.
-    ///
-    /// # Returns
-    ///
-    /// A `Result` indicating success if hashes match, or an error if they don't.
-    pub fn verify_root_hash(
-        &self,
-        context_id: &ContextId,
-        claimed_hash: [u8; 32],
-    ) -> eyre::Result<()> {
-        let computed = self.compute_root_hash(context_id)?;
-
-        if computed != claimed_hash {
-            eyre::bail!(
-                "Root hash verification failed for context {}: computed {} != claimed {}",
-                context_id,
-                hex::encode(computed),
-                hex::encode(claimed_hash)
-            );
-        }
-
-        tracing::debug!(
-            %context_id,
-            hash = ?Hash::from(computed),
-            "Root hash verified successfully"
-        );
-
-        Ok(())
-    }
-
     /// Reads ROOT's index entry + `Key::Entry(ROOT)` bytes in a single
     /// pass and returns both the self-dump (own_hash/full_hash/entry
     /// summary) and the children list — diagnostic for #2319.
@@ -1433,15 +1395,6 @@ impl ContextClient {
     /// Forces the root hash for a context to a specific value.
     pub fn force_root_hash(&self, context_id: &ContextId, root_hash: Hash) -> eyre::Result<()> {
         self.registry.force_root_hash(context_id, root_hash)
-    }
-
-    /// Verifies that the stored root hash matches the actual state.
-    pub fn verify_root_hash(
-        &self,
-        context_id: &ContextId,
-        claimed_hash: [u8; 32],
-    ) -> eyre::Result<()> {
-        self.registry.verify_root_hash(context_id, claimed_hash)
     }
 
     /// Diagnostic — dump ROOT's self summary + children list in one read.
--- a/crates/merod/src/cli/run.rs
+++ b/crates/merod/src/cli/run.rs
@@ -54,10 +54,10 @@ impl RunCommand {
         // startup warning below. Do not flip this to an allow-list; this is the
         // agreed dev-only flag behavior.
         //
-        // FIXME: `has_real_attestation` only covers the Phala KMS provider. If a
-        // second KMS provider is ever added, extend that predicate (see its doc
-        // in `calimero_config::TeeConfig`) so this guard keeps refusing `--mock-tee`
-        // on real-attestation nodes.
+        // `has_real_attestation` destructures `KmsConfig` exhaustively (no `..`),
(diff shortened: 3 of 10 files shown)


chefsale:
Thanks for the review — pushed 7164d02a addressing the findings. Disposition of each:

### Fixed
- **`validate_guest_module` doesn't require a `memory` export** — added a positive check: a guest that exports no linear memory named `memory` is now rejected with a clear validation error at compile time instead of a confusing export-not-found at instantiation. New test `guest_without_memory_export_is_rejected`.
- **Non-atomic read-modify-write on refcounts** — elevated the caveat from the `delete` doc to an explicit `INVARIANT:` note on **both** `release_ref` and `persist_ref` (callers must serialize add/delete per blob id; full atomicity means moving onto `Store::apply`). Also clarified that `delete`'s `Ok(true)` does **not** imply physical removal.
- **`saturating_sub` hides `refs == 0`** — now logs a `warn!` when `release_ref` sees a zero-ref row (corruption) before reclaiming it, so the anomaly is observable rather than silently "fixed".

### Not a bug — declining, with reasoning
- **"Chunk refcount decremented even when root's refcount was not zero" (flagged as data loss)** — this is a false positive that stems from assuming the *add* path increments each chunk only once. It does not: `persist_ref` runs during **every** `put_sized`, so re-adding identical content (or a shared chunk) re-increments **each chunk on every add**. Delete is therefore symmetric — decrement root + each chunk once per delete — which is correct.

  Worked example (root `R` links chunk `C`, added twice): after 2 adds `R.refs=2, C.refs=2`; after 1 delete `R.refs=1, C.refs=1` — consistent. The suggested fix (only release chunks when the root hits zero) would **introduce** a leak: two adds → `C.refs=2`, but chunks only released on the final delete → `C` decremented once → stuck at 1 forever. This is exactly what the two passing tests pin: `duplicate_content_is_refcounted_not_aliased` and `chunk_shared_across_roots_survives_sibling_delete` (the latter reads the sibling blob back byte-for-byte after a shared-chunk sibling delete).

- **Return-contract nuances (`Ok(true)` vs physical removal, idempotent delete)** — intentional and now documented. No current caller needs to distinguish "last ref removed" from "ref released" (the node's `delete_blob` releases the owner's reference and bails on a genuinely-absent blob, matching prior behavior). A richer return type would be speculative for a pre-1.0 API with no consumer for the distinction; I've clarified the doc instead.

The docs-bot mapping warnings are noted; the referenced `docs-site` pages are prose architecture docs and don't describe blob refcounting or the runtime validation pass at a level this changes — happy to add a short note if maintainers want it.

chefsale:
Second review round addressed in cff72e5a.

### Fixed
- **`delete` return-contract asymmetry (orphan file returned `Ok(true)`)** — the no-metadata path now returns `Ok(false)` unconditionally (still best-effort sweeping any orphan file), so `Ok(true)` unambiguously means a referenced blob existed and its reference was released. Also dropped the redundant `let _ =` on the file-delete `?`s.
- **`persist_ref` saturating refcount overflow → permanent leak** — switched to `checked_add`, returning an error on overflow. A saturated `u32::MAX` could never decrement back to zero, so surfacing it is correct even though it's not physically reachable.
- **`_start` check was name-only** — now also asserts the export is a function (`ExternType::Function(_)`), matching the memory-export check's name+type form. (Note: the suggested `ExternType::Func` doesn't exist in wasmer 6 — the variant is `Function` — so I used that.)
- **`validate_invariants` doc wording** — now says "Returns `Ok(())` if all invariants hold, or `Err` with a message…".

### Declining, with reasoning
- **"Add a `TODO(#ISSUE)` for the non-atomic refcount"** — the race is already documented as an explicit `INVARIANT` on both `release_ref` and `persist_ref`. This repo's convention is to keep comments self-contained and free of issue-number / roadmap-code footnotes, so I'm not adding a `TODO(#N)`. The path to atomicity (fold into `Store::apply`) is stated in the comment; happy to do that hardening as a follow-up if maintainers want it in-scope.
- **"`release_ref` `saturating_sub` masks a double-free"** — the concurrent double-delete it describes is precisely the documented non-atomic invariant (two callers both read `refs == 1`), which `saturating_sub` vs an explicit match cannot detect either — only real serialization/atomicity can. The `refs == 0` case (the one detectable without a transaction) already emits a `warn!`. So no behavioural change here beyond what's in place.

All touched crates build, pass clippy clean, and tests pass.

chefsale:
Third round addressed in ae4232c0. Several comments on this pass were re-flags on stale line numbers of items already fixed in cff72e5a (the `_start` name+type check, the `validate_invariants` doc wording, the `saturating_sub`/`refs==0` warn) — those stand as fixed.

### Fixed
- **Partial chunk-release leaves refcounts inconsistent (leaked chunks)** — the chunk-release loop in `delete` is now best-effort: a failed read/release of one chunk is logged and skipped instead of propagated via `?`, so it can no longer leave the *remaining* chunks un-decremented. Root/chunk file-delete failures are likewise logged, not fatal (matching the original code's resilience).
- **`Engine::new` panics via `.expect()` (AGENTS.md: avoid expect in library code)** — replaced with a `debug_assert!` on the VMLimits invariant. The invariant is only violable by a code change (not guest input, and operator config doesn't expose these fields), so a debug assertion enforces it in dev/CI/tests without a release-mode panic in a public constructor. The `should_panic` test still exercises it (debug-assertions are on under `cargo test`).

### Declining, with reasoning
- **Non-atomic `release_ref` / chunk-refcount TOCTOU** (multiple duplicate flags) — same documented `INVARIANT` (serialize add/delete per blob id). Real atomicity needs `Store::apply`; the doc states that path. Flagged for a maintainer decision on whether to pull that hardening into this PR vs a follow-up — I'm happy to do either.
- **`persist_ref` overwrites size/hash on duplicate add** — a no-op under content addressing: the blob id *is* the content hash (chunk id = chunk hash; root id = hash of chunk ids), so any two adds resolving to the same id necessarily carry identical size/hash. The "different metadata for the same id" scenario is unreachable without breaking content addressing itself.
- **Validation runs after `wasmer::Module::new`, not before** — introspecting imports/exports inherently requires wasmer to parse the module, and wasmer's own validation *is* the first line of defense (it rejects malformed WASM); the size cap runs before compile. Our checks are a deliberate second policy layer, as the doc-comment states. No pre-parse introspection API exists that wouldn't itself parse the bytes.

Builds, clippy clean, tests pass. `Cargo.lock` untouched.

PR #3139, opened 2026-07-02
fix(client): single-flight token refresh and interactive auth

frdomovic:
## Summary

Coalesces concurrent token refresh / re-authentication in the client library (`crates/client`, consumed by meroctl).

## Before

Refresh and re-auth had no coordination:
- **N concurrent 401s** each independently `POST /auth/refresh` with the **same one-time refresh token**. Once the token rotates (client-type keys), all but the first fail with "key not found" and cascade to full re-authentication.
- **N concurrent token-less requests** each call `authenticate()` in `ensure_auth_header`, which for meroctl opens a **browser window per request** → N popups.

## After

A single `Arc<tokio::sync::Mutex<()>>` is shared across all clones of `ConnectionInfo`, and both paths go through it:

- The 401 path now calls a new `refresh_or_reauth(used_auth_header)` that takes the lock and, **before** refreshing, compares the stored token to the token the failed request used. If another task already refreshed while this one waited, the stored token differs → it **skips** and lets the caller retry with the now-current token (so the one-time refresh token is spent exactly once).
- The proactive (no-token) branch of `ensure_auth_header` takes the same lock and **re-checks storage after acquiring it**, so concurrent first-time requests trigger **one** interactive login instead of N browser prompts.

## Scope

Practical severity is low–medium (meroctl is largely sequential, and root-key refresh isn't strictly one-time), but the coordination removes a real thundering-herd + popup-storm failure mode and is a prerequisite for the interactive/non-interactive `auth_header` split (finding 19).

## Verification

- `cargo test -p calimero-client` → 43 passed.
- `cargo check -p meroctl` passes.

Findings #11 and #19 (popup-storm half) from the security review.

### DIFF
--- a/crates/client/src/connection.rs
+++ b/crates/client/src/connection.rs
@@ -4,10 +4,13 @@
 //! authenticated API requests to Calimero services.
 
 // External crates
+use std::sync::Arc;
+
 use eyre::{bail, eyre, Result};
 use reqwest::Client;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
+use tokio::sync::Mutex;
 use url::Url;
 
 // Local crate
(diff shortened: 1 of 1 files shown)


frdomovic:

PR #3140, opened 2026-07-02
fix(client): percent-encode dynamic id segments in request paths

frdomovic:
## Summary

Prevents `&str`-typed ids from breaking out of their URL path segment in the client library.

## Before

Request paths were built as `format!("admin-api/groups/{group_id}/members/{identity_hex}/role")` and handed to `Url::set_path`, which treats `/` as a segment separator (and applies dot-segment processing). The `&str`-typed ids — `group_id`, `namespace_id`, `identity_hex`, `context_id`, `package`, `application_id` — were interpolated verbatim, so a value containing `/` or `..` could re-route the request to a **different admin endpoint**.

## After

- New `connection::path_segment(value)` percent-encodes everything outside the RFC 3986 unreserved set (`A-Za-z0-9-._~`).
- Each `&str` id is encoded at the top of the group / namespace / packages client methods (via a shadow binding, so the existing `format!` calls are unchanged).
- Ordinary hex/base58 ids and dotted package versions pass through untouched; only reserved characters in hostile input are encoded.

## Not changed (already safe)

- **Typed ids** (`ContextId`, `ApplicationId`) are base58 and cannot contain `/`.
- **Aliases** (`Alias<T>`) are charset-restricted to `[A-Za-z0-9._-]` (see the alias-validation PR), so they can't contain separators either.

## Verification

- `cargo test -p calimero-client` → 45 passed (2 new `path_segment` unit tests; existing group/context tests unaffected since hex ids encode to themselves).

Finding #12 from the security review.

### DIFF
--- a/crates/client/src/client/group.rs
+++ b/crates/client/src/client/group.rs
@@ -63,6 +63,7 @@ where
     S: ClientStorage + Clone + Send + Sync,
 {
     pub async fn get_group_info(&self, group_id: &str) -> Result<GroupInfoApiResponse> {
+        let group_id = crate::connection::path_segment(group_id);
         let response = self
             .connection
             .get(&format!("admin-api/groups/{group_id}"))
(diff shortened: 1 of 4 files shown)


frdomovic:

PR #13, opened 2024-02-12
feat(client-rs): initial project files

frdomovic:
# feat(client-rs): initial project files
 
## Summary:
Client CLI initial project files -> streaming-api functionalities with extended other options from specification [here](https://calimero.atlassian.net/wiki/spaces/CAL/pages/208601117/Client)

Initial Client CLI with skeleton of functionalities.
init  - working same as chat poc example         
join  - tbd from other component side
start-session  - tbd from other component side
add-key-pair  - decide if we need and with what configuration
login  - decide if needed and with what configuration
list-nodes  - tbd - decide if we need it
send-message - leftover from sandi's api - can be removed
read-message  - leftover from sandi's api - can be removed
### Streaming api - all are connected with streaming api
list-remote-apps - handled
list-installed-apps  - handled
install-binary-app  -  tbd from api side
install-remote-app - tbd from api side
uninstall-app  - tbd from api side
subscribe   -  tbd from api side
unsubscribe   - tbd from api side - connect with node
unsubscribe-all   - tbd from api side - connect with node

## Todo
Refactor after streaming api merged in master

Fixes # [DEV-306](https://calimero.atlassian.net/browse/DEV-306)

## Test plan:
-> old

https://github.com/calimero-is-near/cali2.0-experimental/assets/93442516/8070af2f-0ca3-4e35-9565-f83d6dcf9a1c

-> new functions

https://github.com/calimero-is-near/cali2.0-experimental/assets/93442516/6a74005d-58b6-411c-bbf8-3435475f21fd

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have added a test plan that prove my fix is effective or that my feature works
- [ ] I squashed all commits and provided a meaningful commit message
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

PR #3141, opened 2026-07-02
fix(client): cap buffered response bodies to bound memory use

frdomovic:
## Summary

Bounds client memory use when reading HTTP response bodies, so a hostile/compromised node can't OOM the client.

## Before

- Success path: `response.json::<O>()` buffered the entire body with no limit.
- Error paths (main request + token refresh): `response.text()` read the whole body unbounded before extracting a short message.

A malicious node could return (or claim via `Content-Length`) an arbitrarily large body and exhaust client memory.

## After

Two helpers:
- `read_capped_body(resp, max)` — rejects up front if `Content-Length` exceeds `max`, **and** enforces the cap while streaming chunks (so a lying/absent `Content-Length` still can't grow the buffer without bound).
- `read_error_body(resp, max)` — reads up to `max` for message rendering, truncating rather than failing.

Wired into the JSON success path, the main non-2xx error path, and the token-refresh success + error paths. Caps: **16 MiB** JSON, **64 KiB** error bodies. (This also fully closes the residual unbounded read on the error path noted in finding 17.)

## Not changed (follow-up)

`get_binary` / `download_blob` are intentionally left uncapped: blobs are legitimately large, and the right fix is a streaming download-to-writer variant rather than a fixed cap. Tracked as a follow-up.

## Verification

- `cargo test -p calimero-client` → 43 passed (wiremock request tests exercise the new capped read path).

Findings #13 and #17 (residual) from the security review.

### DIFF
--- a/crates/client/src/connection.rs
+++ b/crates/client/src/connection.rs
@@ -234,7 +234,11 @@ where
             })
             .await?;
 
-        response.json::<O>().await.map_err(Into::into)
+        // Cap the JSON body so a hostile/compromised node can't OOM the client
+        // with an unbounded response. Admin-API JSON payloads are small; blob
+        // bytes go through `get_binary`, not here.
+        let bytes = read_capped_body(response, MAX_JSON_BODY).await?;
+        serde_json::from_slice::<O>(&bytes).map_err(Into::into)
     }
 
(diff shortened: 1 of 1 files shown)


frdomovic:

PR #3142, opened 2026-07-02
fix(client): harden HTTP/auth layer against replay, herd, traversal & OOM

chefsale:
## Summary

Batch of hardening fixes for the reported `crates/client` issues (HTTP transport + auth state machine). Each fix is self-contained; several share a helper.

| Reported issue | Fix |
|---|---|
| **[H] 401 replays non-idempotent mutations** | `RequestType::is_idempotent()`; on 401, POST/PATCH re-authenticate the session but are **not replayed** (returns an error asking the caller to retry). GET/PUT/DELETE still retry. Prevents duplicate `create_context`/governance ops when a proxy 401s after the origin already applied the write. |
| **[H] Concurrent refresh thundering herd** | Single-flight `refresh_or_reauth()` behind an `Arc<Mutex<()>>` with a stale-token re-check — N concurrent 401s collapse into one `/auth/refresh` (and one browser prompt) instead of spending the one-time refresh token N times. |
| **[H] User ids interpolated into URL path (traversal)** | Central `resolve_path()` rejects `.`/`..` dot-segments and control chars, covering all interpolation sites at once. |
| **[M] Response bodies buffered with no cap** | Streaming `read_body_capped()` (16 MB JSON / 512 MB binary / 64 KB error) that tallies bytes as they arrive, so a lying `Content-Length` or endless stream can't OOM the client. Applied to `request`, `get_binary`, and blob upload. |
| **[M] Proactive refresh is dead** | `expires_at` is derived from the JWT `exp` claim in `new()`/`with_refresh()` (survives the meroctl storage round-trip); `ensure_auth_header` now refreshes expired / soon-to-expire tokens **before** sending. |
| **[M] Logout persists an empty token** | `JwtToken::is_usable()` treats an empty access token as "no credentials", so a logged-out record is never sent as `Bearer `. |
| **[M] Refresh overwrites token metadata** | Default `update_tokens` loads + `merged_with()` to preserve prior `expires_at`/metadata/refresh token; removed meroctl's redundant blind override. |
| **[M] Error path reads body unbounded** | Routed through the capped reader + `extract_error_message` (same as refresh path). |
| **[M] ws_url discards reverse-proxy base path** | `ws_url` appends `ws` to the existing path (`.../base/ws`) instead of `set_path("ws")`; `request()`/`head()` now split query strings. |
| **[M] auth_header() getter triggers browser** | The interactive flow is now single-flighted so concurrent token-less callers share one prompt; `JwtToken` zeroizes access/refresh secrets on drop. |

## Notable decisions
- **Non-idempotent 401s are not auto-replayed** (per the report). Since expiry is now handled proactively, users rarely hit this; when they do the session is healed and the error tells them to retry. Easy to switch to retry-once if preferred.
- **Zeroize-on-drop** makes moving fields out of a `JwtToken` illegal (E0509) — a couple of meroctl sites switched to `.clone()`.
- **Left alone:** the process-global `SESSION_CACHE` (used by `meroctl/auth.rs`); zeroize scrubs its entries, but removing the singleton is a larger refactor deferred out of this PR.

## Testing
- `cargo test -p calimero-client` → 58 passed (new tests: path guard/traversal, JWT `exp` parsing, token merge, verb-aware retry, query forwarding).
- `Cargo.lock` change is minimal: only the `base64` + `zeroize` dependency edges (both already present transitively).


chefsale:
Addressed the reviewer findings in 31cce1245:

- **`read_body_capped` allocation / overflow (fcd14c215f29, e04a4fb7fca9, d9958d297ba4):** now reserves `Vec::with_capacity(min(Content-Length, cap))` so an inflated `Content-Length` can't trigger a huge up-front allocation; per-chunk tally uses `saturating_add`; doc clarified that the cap is inclusive and that a missing/lying `Content-Length` is still caught per-chunk.
- **Proactive-refresh fallback returned stale local token (44c73eb9ed5c):** after an `expires_soon` refresh, `ensure_auth_header` now returns the freshly-stored token via `current_auth_header` regardless of refresh outcome, falling back to the local token only if storage yields none — so it picks up a concurrent refresh.
- **`update_tokens` TOCTOU (a5693101b2ac):** documented the non-atomic load-merge-save and that the connection layer serializes via its internal lock while external callers must serialize themselves.

Doc-bot nudges (AGENTS.md / CONTRIBUTING.md / architecture HTML): not applicable — this is an internal client transport/auth bugfix with no public architecture or CLI-surface change (the meroctl edit only removes a redundant trait override + adds `.clone()`s required by the new zeroize-on-drop).

`cargo test -p calimero-client` → 58 passed; `cargo clippy -p calimero-client -p meroctl --all-targets -- -D warnings` → clean.

chefsale:
Addressed the round-2 findings in 3be53effb:

- **`resolve_path` replaced base path (patterns-reviewer #1):** it now **appends** request segments to the base URL's existing path via `path_segments_mut()` instead of `set_path`, so a reverse-proxy base path (`https://host/calimero/node1/`) is preserved for admin-api calls — consistent with the `ws_url` fix. Added a test covering base paths with and without a trailing slash.
- **Percent-encoded traversal bypass (patterns-reviewer #2):** each segment is now percent-decoded before the `.`/`..` check, so `%2e%2e` (which a proxy may normalize to `..` at the origin) is rejected up front. Added a test. (As a second layer, `path_segments_mut().push()` re-encodes each segment, so raw `%xx` input can't reach the origin as a decoded dot-segment either.)
- **Swallowed error-body read failures (patterns-reviewer #3):** the `unwrap_or_default()` on both the HTTP error path and the token-refresh error path now emits a `tracing::debug!` on read failure instead of discarding it silently.

`cargo test -p calimero-client` → 60 passed; `cargo clippy -p calimero-client -p meroctl --all-targets -- -D warnings` → clean. Lock delta is just the `percent-encoding` + `tracing` edges (both already in the tree).

chefsale:
Addressed round-3 findings in f135f4308:

- **Proactive refresh swallows errors / may return stale token (570584f89594):** the proactive-refresh failure is now logged at `tracing::debug`. Note the fallback token is *not* stale in the expiry sense — control flow above the `expires_soon` branch already returns early for `is_expired()`, so the token is guaranteed near-expiry-but-still-valid here; added a comment stating this. (I kept it as debug rather than warn since a transient proactive-refresh miss is expected/recoverable — the request still succeeds on the current token.)
- **update_tokens TOCTOU (04569039c4e0):** strengthened the doc to explicitly warn that external/concurrent callers must serialize, and that a backend able to update atomically (CAS/transaction) should override. I did **not** seal the trait / make the method non-overridable — that would prevent exactly the atomic-override the suggestion also asks for.
- **Redundant `.min(max_bytes)` in read_body_capped (fcdf37f6391f, 💡):** kept it (it's independent of the early-return check) with a comment marking it defensive against future refactoring.

`cargo test -p calimero-client` → 60 passed; `cargo clippy -p calimero-client -p meroctl --all-targets -- -D warnings` → clean.

PR #3144, opened 2026-07-02
fix(context): harden xcall, join, upgrade, storage & handler race conditions

chefsale:
## Summary

A batch of correctness/security fixes across the `ContextManager` handlers. Each item is independent; grouped here because they share the same subsystem and review surface.

| Sev | Finding | Fix |
|-----|---------|-----|
| **H** | xcall re-enters the same actor and can deadlock/hang (self-mailbox `execute_with_origin().await` while the source future is in flight; self-xcall also deadlocks on the source execution lock) | Dispatch the whole xcall batch **out-of-band** on a detached `global_runtime().spawn` task. The source execute returns and drops its guard first, so the re-entrant execute runs cleanly. xcalls are already best-effort (their only product is an observability event), so fire-and-forget matches their contract. |
| **M** | xcall authorized on the namespace root → crosses subgroup boundary (any sibling context anywhere in the namespace can call a Restricted-subgroup context and execute as a member) | Gate xcall on the **same owning group** (`get_group_for_context` equality) instead of the shared namespace root. Denial reason renamed `owning group boundary`. No cross-subgroup grant mechanism exists; contexts that must call across subgroups have to share an owning group. |
| **M** | `join_context` fallback re-enumerates all groups every 200ms (full namespace re-enumeration + network sync per unresolved join) | Sync only namespaces the node holds an identity in (`iter_identities`, deduped roots — the only ones the join can succeed in), and back the fallback re-sync off exponentially (200ms → 5s). |
| **M** | `state_key` silently drops keys >32 bytes and zero-pads <32 (data loss for long keys; short keys collide) | Require `key.len() == 32` in both `ContextStorage` and `ContextPrivateStorage`. |
| **M** | Concurrent lazy upgrades both pass validation and emit racing op pairs | Persist a short-lived `InProgress` record **synchronously** before the async work (mirrors the canary path), so a concurrent request fails `validate_upgrade`; a cleanup map deletes it on failure so nothing wedges. |
| **L** | `create_group` existence guard races the async write | Reserve the group id **synchronously** in the handler body; cleanup map frees it on failure. |
| **L** | Ownership-proof fields unbounded / no field validation | Bound & validate `audience`/`subject`/`nonce` (non-empty, ≤256B, no control chars, nonce ≥8B) and document the **mandatory verifier obligations** (nonce single-use, audience match, subject authorization). |
| **L** | `leave_context` tombstones only one local identity | Enumerate and tombstone **all** local `ContextIdentity` rows (new `find_local_signing_identities`), so extra identities can't keep syncing / re-arm auto-follow. |

### Not changed
The reported auto_follow backfill stall (`handle_auto_follow_enabled` awaited inline) is **already fixed on current master** — `run()` dispatches each event onto a `JoinSet` task with an explicit rationale comment. Verified directly; no change needed.

## Tests
- **storage**: exact-32 roundtrip, non-32 rejection, short-key no-longer-collides-with-padded
- **xcall boundary**: same-group allow, cross-subgroup deny, unregistered deny
- **ownership proof**: empty/oversized/control/short-nonce rejection + at-max acceptance; namespace-variant validation
- **leave_context helper**: returns all key-holding identities, excludes keyless / other-context rows
- **upgrade_group**: `validate_upgrade` rejects when an `InProgress` record exists (pins the guard the lazy fix relies on)
- **e2e**: `apps/xcall-example/workflows/xcall.yml` updated to the same-owning-group semantics and the renamed `owning group boundary` denial

`cargo check -p calimero-node` (full downstream), `calimero-context` (216) and `calimero-governance-store` (400) lib tests pass; clippy `-D warnings` clean on touched crates. `Cargo.lock` untouched.

### DIFF
--- a/apps/xcall-example/workflows/xcall.yml
+++ b/apps/xcall-example/workflows/xcall.yml
@@ -1,10 +1,13 @@
-name: XCall Authorization (L1 namespace · L3 entry-point · L2a provenance)
+name: XCall Authorization (L1 owning-group · L3 entry-point · L2a provenance)
 description: >
   End-to-end validation of cross-context-call authorization (closes #2130 / #2137).
 
   A node-local xcall is gated by three independent layers, all exercised here:
-    * L1 — namespace boundary: a context may only xcall a target in its OWN
-      namespace. Cross-namespace calls are denied (fail-closed).
+    * L1 — owning-group boundary: a context may only xcall a target in its OWN
+      owning group. Calls to a target in a different group — a different
+      namespace, or a different subgroup of the same namespace — are denied
+      (fail-closed). Gating on the shared namespace root instead would let any
+      sibling context anywhere in the namespace reach into another subgroup.
     * L3 — entry points: only methods declared `#[app::xcall]` are reachable
       via xcall. The non-entry-point `secret` method is denied.
     * L2a — provenance: `env::xcall_origin()` is set by the node from the
@@ -81,9 +84,11 @@ steps:
       c3_key: memberPublicKey
 
   # ============================================================================
-  # POSITIVE — same-namespace xcall to a declared entry point is ALLOWED.
-  # Exercises L1 (C1 and C2 share namespace N), L3 (pong is `#[app::xcall]`),
-  # and L2a (the node-set origin equals C1, which pong verifies before running).
+  # POSITIVE — same-group xcall to a declared entry point is ALLOWED.
+  # C1 and C2 are both created directly in namespace N's root group, so they
+  # share an owning group. Exercises L1 (same owning group), L3 (pong is
+  # `#[app::xcall]`), and L2a (the node-set origin equals C1, which pong
+  # verifies before running).
   # ============================================================================
   - name: C2 pong counter starts at 0
     type: call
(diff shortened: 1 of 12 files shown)


chefsale:
Addressed the review findings in b0cdd0b53:

- **Reservation carries unverified `app_key` (findings #1, #3)** — the synchronous `create_group` reservation now always uses the verified application-row blob (`row_blob`) for `app_key`, never the caller's still-unverified `requested_app_key`. The async body still overwrites with the final verified value on success, so a concurrent reader can no longer observe an unverified key.
- **Detached xcall task swallows panics (findings #2, #4 + inline)** — the batch is now supervised: a second task awaits the work task's `JoinHandle` and logs an `error!` on abnormal termination (panic/cancel). The source execute still doesn't wait on the xcalls (fire-and-forget preserved), but panics are surfaced instead of vanishing.
- **Blind `member_public_keys[0]` index (findings #5, #6)** — replaced with `.first().copied().expect(...)` making the non-empty invariant explicit and auditable.

All `calimero-context` (216) lib tests pass; clippy `-D warnings` and `cargo fmt --check` clean.

chefsale:
Addressed the latest review in 8b094bcd7:

- **Reservation timestamp silently stamps epoch on clock failure** — `create_group`'s `reservation_now` now logs a `warn!` on the before-UNIX_EPOCH branch instead of silently using `created_at=0`, matching the project's no-silent-failure convention.
- **Misleading `set().is_none()` assertion in the storage rejection test** — dropped it (`None` is also the return for an accepted insert with no prior value); the `get`/`has` reads already prove the key was refused.

clippy `-D warnings` and `cargo fmt --check` clean; `calimero-context` lib tests pass.

PR #3146, opened 2026-07-02
test(xcall): e2e regressions for the deadlock fix and owning-group boundary

chefsale:
## Summary

Follow-up to #3144 (context-handler hardening), adding the end-to-end regression coverage that the merged batch lacked. Extends the existing `xcall.yml` workflow — already in the `e2e-rust-apps` CI matrix and triggered by `apps/xcall-example/**` — with two scenarios:

1. **Self-xcall must complete, not hang** (`C1 -> C1`). The [H] deadlock fix moved xcall dispatch out-of-band; awaiting `execute_with_origin` inline re-entered the actor and deadlocked on the source context's own execution lock. This step exercises exactly that path — if inline dispatch is ever reintroduced, it hangs until `wait_timeout` and the run fails. Asserts C1's pong counter goes 0 → 1.

2. **Cross-subgroup xcall is denied** (`C1` in namespace N's root group → `C4` in subgroup S nested under N). Same namespace, different owning group: the old namespace-root gate allowed this, the owning-group gate ([M] fix) denies it. Asserts C4's counter stays 0 and a second `owning group boundary` denial is logged.

Both use existing merobox step types (`create_group_in_namespace`, `create_context`, `call`, `assert_log_present`); no app-code change. YAML validated.

## Not included (and why)
The two *actor-level* concurrency tests (two same-id `create_group`, two concurrent lazy upgrades) are **not** here: building a `ContextManager` actor in a test requires a real `NodeClient`/`ContextClient` (network + blob + sync stack), and no mock exists — every current integration test deliberately stays at the store/applier level. The upgrade InProgress guard is already unit-tested in #3144 (`validate_upgrade_rejects_when_an_inprogress_record_exists`). Standing up a mock-client actor harness is a separate, larger effort — flagged in the follow-ups memo if we want it.

### DIFF
--- a/apps/xcall-example/workflows/xcall.yml
+++ b/apps/xcall-example/workflows/xcall.yml
@@ -17,9 +17,19 @@ description: >
   Denials are observable: the target counter never moves, and the node logs the
   same "xcall denied: ..." line that drives the `XCall` Denied event.
 
-  Single node: all three contexts are created by node-1, which owns a member of
-  each, so every xcall dispatches locally — the gates are node-local, so one
-  node exercises them fully.
+  Two regression scenarios also run at the end:
+    * Self-xcall (C1 -> C1) must COMPLETE, not hang. xcalls are dispatched
+      out-of-band; awaiting them inline re-entered the actor and deadlocked on
+      the source context's execution lock. A reintroduced inline dispatch hangs
+      this step until wait_timeout.
+    * Cross-subgroup xcall (C1 in N's root group -> C4 in subgroup S under N) is
+      DENIED. This is the owning-group boundary in its own right: same
+      namespace, different owning group — the old namespace-root gate allowed
+      it, the owning-group gate denies it.
+
+  Single node: all contexts are created by node-1, which owns a member of each,
+  so every xcall dispatches locally — the gates are node-local, so one node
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed the review in 3225a75bf — all three were fair points about test fragility:

- **`min_matches: 2` order-dependent** — lowered to `min_matches: 1`. The log is cumulative over the run (the earlier C1→C3 cross-namespace denial also matches), so requiring 2 coupled this check to unrelated steps without adding signal. The `c4_after == 0` counter assertion is and remains the authoritative proof that the cross-subgroup xcall was denied.
- **C1 counter pre-condition assumes no prior pings landed on C1** — added a comment documenting the invariant (C1 was only ever a ping *source*, never a pong *target*, in the prior steps, so its counter is still 0).
- **`c1_after == 1` absolute vs delta** — kept `== 1` on purpose and documented why: with `c1_before` pinned to 0 immediately above, `== 1` proves exactly-once execution — it catches both a hang/no-op (stays 0) *and* an accidental duplicate dispatch (would be 2). `>= 1` would mask the double-dispatch case.

YAML re-validated (45 steps, both owning-group log asserts now `min_matches: 1`).

PR #3149, opened 2026-07-02
fix(client): preserve reverse-proxy base path for token refresh & auth probe

chefsale:
## Summary

Follow-up to #3142. That PR made request and WebSocket URLs preserve a reverse-proxy base path on `api_url`, but left two sibling calls using raw `Url::join` that discard it — so a node behind a proxy mount path (e.g. `https://host/calimero/node1/`) had working admin-api + WS but broken token refresh and auth-mode detection.

| Site | Before | Problem |
|---|---|---|
| `try_refresh_token` (`connection.rs`) | `api_url.join("/auth/refresh")` | Absolute join → refresh POST goes to `https://host/auth/refresh`, dropping the base path (bearer-bearing request to the wrong origin). |
| `detect_auth_mode` (`connection.rs`) | `api_url.join("admin-api/contexts")` | Relative join → drops the last base segment when `api_url` has no trailing slash (`.../node1` → `.../admin-api/contexts`). |

Both now go through the existing base-path-preserving `resolve_path` helper (the same one requests already use), so refresh and probe stay under the mount path. `resolve_path` also uses `path_segments_mut` append, so the result is correct with or without a trailing slash.

## Testing
- New wiremock tests: `token_refresh_preserves_base_path` (asserts the refresh POST hits `<base>/auth/refresh` and interactive auth is never reached) and `detect_auth_mode_preserves_base_path` (probe hits `<base>/admin-api/contexts`).
- `cargo test -p calimero-client` → 62 passed.
- No `Cargo.toml`/`Cargo.lock` change (no new deps).

### DIFF
--- a/crates/client/src/connection.rs
+++ b/crates/client/src/connection.rs
@@ -612,9 +612,13 @@ where
                     Ok(AuthMode::Required)
                 }
             }
-            Err(_) => {
-                // If we can't reach the endpoint, assume no authentication required
-                // This is common for local nodes that don't have the admin API enabled
+            Err(e) => {
+                // If we can't reach the endpoint, assume no authentication
+                // required — common for local nodes without the admin API. This
+                // also fires when a reverse-proxy base path is misconfigured or
+                // the proxy is down, so log the cause at debug to make that
+                // diagnosable rather than silently skipping auth.
+                tracing::debug!("detect_auth_mode: probe request failed, assuming no auth: {e}");
                 Ok(AuthMode::None)
             }
         }
--- a/crates/client/src/connection.rs
+++ b/crates/client/src/connection.rs
@@ -506,7 +506,11 @@ where
     }
 
     async fn try_refresh_token(&self, access_token: &str, refresh_token: &str) -> Result<JwtToken> {
-        let refresh_url = self.api_url.join("/auth/refresh")?;
+        // Resolve via `resolve_path` (not a raw `join`) so a reverse-proxy base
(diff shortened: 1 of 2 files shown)


chefsale:
Addressed the suggestion in cb75d22b1:

- **`detect_auth_mode` masks probe failures (security-reviewer):** the network-error arm now logs the cause at `tracing::debug!` (and the comment notes this also fires on a misconfigured/down reverse proxy) before falling back to `AuthMode::None`. Return value is unchanged, so no behavior change — just diagnosability, consistent with the other swallowed-error logging added in #3142.

`cargo test -p calimero-client` → 62 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

chefsale:
Addressed both findings in 979c4d982:

- **Token-refresh body cap (f8166b762528, 💡):** the refresh response is now read through a dedicated `MAX_TOKEN_BODY_BYTES` (64 KiB) instead of the 16 MiB general JSON cap — a refresh response is just two JWT strings, so the auth path no longer lets a hostile/misconfigured proxy buffer megabytes.
- **Test-comment clarity (01b0c2d3a0fa, 🟡):** reworded the `detect_auth_mode` test comment to spell out that wiremock 404s unmatched routes and `detect_auth_mode` maps a non-401 non-2xx to `AuthMode::Required`, so asserting `None` proves the prefixed path was hit and returned 200.

`cargo test -p calimero-client` → 62 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

PR #3151, opened 2026-07-02
test: harden crypto/op/key/FSM/governance tests; de-flake sleep-based & stub tests

chefsale:
## Summary

Fills the targeted testing gaps and removes wall-clock flakiness across several crates. All new/changed tests are additive; the only non-test production changes are two small, behavior-preserving seams (a `#[cfg(test)]` helper in `dag`, and extraction of pure timing arithmetic into helpers in `node/join_namespace`).

### Governance
- **`governance-store` cross-node determinism** — new `cross_node_state_hash_is_order_independent`: applies commuting `MemberAdded` ops from distinct signers in one order on store A and the reverse on store B, asserts equal `compute_state_hash()` and that it differs from baseline (so it can't pass vacuously). The existing `compute_state_hash_is_deterministic` only proved call-idempotency.
- **`governance-store` capability negative paths** — the six core rejections already existed but fail on the first (`require_manage_members`) guard; added the genuinely-uncovered paths past it: grant `NotInGroup`, grant non-member-grantee, revoke `NotInGroup`.

### Backoff / FSM / keys
- **`node/join_namespace`** — the retry loop mixes `std::Instant` + inline `rand`, so a paused clock can't drive it directly. Extracted the pure backoff/jitter/budget math (`next_backoff`, `jitter_bound_millis`, `attempt_deadline`, `would_exceed_budget`), wired the loop to them, and unit-tested each deterministically.
- **`network` reachability FSM** — transition + action-set tests: Unknown→Reachable, Unknown→Unreachable, no-action-on-non-transition, Reachable→Unreachable on last-address removal.
- **`store/key/context`** — the file had **zero** tests; added round-trip + exact-length-rejection for its key types, plus the missing `GroupContextMemberCap` round-trip in the group key tests.

### De-flaking / stubs / crypto / fuzz
- **`crypto`** — tampered-tag, tampered-ciphertext, and mismatched-nonce AEAD negatives.
- **`op`** — malformed-bytes fuzz negatives (out-of-range discriminant, truncated, trailing garbage, degenerate `Op` buffers).
- **`storage/crdt`** — replaced `thread::sleep` in the three LWW tests with injected `set_updated_at` timestamps (deterministic, instant).
- **`storage/interface`** — implemented the three ignored `todo!()` merkle stubs (own-hash, with-children, with-parents) with an independent recompute cross-check.
- **`dag`** — `cleanup_stale` ages via `std::Instant` (which `tokio::time::pause` can't control), so added a `#[cfg(test)]` backdate seam and dropped the real 100ms sleep.

> Note: the "[H] Borsh discriminant golden vectors" item was already covered on `master` (`governance-types/src/tests.rs` pins every `GroupOp`/`RootOp` discriminant by decoding frozen byte vectors), so no change there.

## Test
Per-crate `cargo test` verified green for all touched crates (crypto, op, store, storage, dag, network, governance-store, node), plus full dag suite (62) and the existing golden-borsh tests.

### DIFF
--- a/crates/crypto/src/lib.rs
+++ b/crates/crypto/src/lib.rs
@@ -172,6 +172,95 @@ mod tests {
     }
 
     #[test]
+    fn test_decrypt_with_tampered_tag() -> eyre::Result<()> {
+        // AES-GCM appends a 16-byte authentication tag after the ciphertext.
+        // Flipping a bit in that tag must make `open_in_place` reject the
+        // message, so `decrypt` returns `None` rather than garbage plaintext.
+        let mut csprng = thread_rng();
+        let signer = PrivateKey::random(&mut csprng);
+        let verifier = PrivateKey::random(&mut csprng);
+        let signer_shared_key = SharedKey::new(&signer, &verifier.public_key())?;
+        let verifier_shared_key = SharedKey::new(&verifier, &signer.public_key())?;
+
+        let payload = b"privacy is important";
+        let nonce = [0u8; NONCE_LEN];
+        let mut encrypted = signer_shared_key
+            .encrypt(payload.to_vec(), nonce)
+            .ok_or_eyre("encryption failed")?;
+
+        // The tag is the trailing bytes of the sealed buffer.
+        let last = encrypted.len() - 1;
+        encrypted[last] ^= 0x01;
+
+        assert!(
+            verifier_shared_key.decrypt(encrypted, nonce).is_none(),
+            "decrypt must reject a tampered authentication tag"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_decrypt_with_tampered_ciphertext() -> eyre::Result<()> {
+        // Mutating the ciphertext body (not the tag) must also fail
+        // authentication — the tag covers the whole ciphertext.
+        let mut csprng = thread_rng();
+        let signer = PrivateKey::random(&mut csprng);
+        let verifier = PrivateKey::random(&mut csprng);
+        let signer_shared_key = SharedKey::new(&signer, &verifier.public_key())?;
+        let verifier_shared_key = SharedKey::new(&verifier, &signer.public_key())?;
+
+        let payload = b"privacy is important";
+        let nonce = [0u8; NONCE_LEN];
+        let mut encrypted = signer_shared_key
+            .encrypt(payload.to_vec(), nonce)
+            .ok_or_eyre("encryption failed")?;
+
+        // Flip the first ciphertext byte (well before the appended tag).
+        encrypted[0] ^= 0x01;
+
+        assert!(
+            verifier_shared_key.decrypt(encrypted, nonce).is_none(),
+            "decrypt must reject tampered ciphertext"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_decrypt_with_mismatched_nonce() -> eyre::Result<()> {
+        // AES-GCM binds the nonce into tag verification. A ciphertext sealed
+        // under nonce A must not open under nonce B, even with the right key.
+        let mut csprng = thread_rng();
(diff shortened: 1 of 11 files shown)


chefsale:
Addressed the review comments in `8a735fd6d`:

- **`dag::backdate_pending_for_test` underflow panic** (3 comments) — removed the `checked_sub(...).expect(...)`. Note `std::time::Instant` has no `saturating_sub` (the suggested call doesn't compile), so it now uses `checked_sub(by).unwrap_or(received_at)`: never panics, and on any realistic host/CI container the boot-relative monotonic clock is always older than the sub-second backdate, so it still ages the entry.
- **`cross_node_state_hash_is_order_independent` vacuous-pass risk** (3 comments) — added explicit `is_member` assertions for all three members on both stores after convergence, so hash equality can't pass if an op silently no-op'd or the hash ignored members.
- **`save__update_merkle_hash`** (2 comments) — the accepted-write assertion now carries a clear message, and the new hash is cross-checked against `recompute_full_hash(...)` to match the with-children / with-parents siblings.

All touched crates re-verified green.

chefsale:
Latest round addressed in `69193ed29`:

- **`dag` backdate seam silently no-ops on fresh hosts** — reworked the seam entirely. `PendingDelta` now has a `#[cfg(test)] age_override: Option<Duration>` that `age()` returns directly, set via `set_pending_age_for_test`. This is unconditional (no `Instant` subtraction, so no dependence on host uptime), cannot panic, and exists only in test builds — no production impact. Fully removes the CI fragility.

- **`set_updated_at` called after the first save** (crdt.rs / interface.rs) — these are false positives: in the current code `set_updated_at(base)` is called *before* the first save in every case (e.g. `lww_newer_update_wins` sets it on line 42, saves on line 45; `save__update_merkle_hash` sets on line 151, saves on line 152). The first save therefore records `base`, and the second write at `base + ONE_SEC_NANOS` is strictly newer, so the ordering is deterministic. No change needed.

The remaining bot comments (12:18–12:21) are re-posts of the earlier round against the pre-fix commit and are already resolved by `8a735fd6d`.

chefsale:
Addressed in `ad7f6eb4d`:

**CI (Rust / clippy `-D warnings`)** — the failure was `doc list item without indentation` in `calimero-storage`: the `recompute_full_hash` doc had a line starting with `/// + children`, which clippy parses as a malformed list bullet. Reworded to `and children`. Ran `cargo clippy --all-targets --features calimero-storage/testing -- -D warnings` across every touched crate locally — clean.

**`dag` `#[cfg(test)]` field** — reworked to avoid the config-dependent struct entirely. `cleanup_stale` now delegates to a `cleanup_stale_since(reference, max_age)` core; production passes `Instant::now()`, tests pass a synthetic reference. `PendingDelta` is now identical across all build configs, the seam is still uptime-independent, and `saturating_duration_since` can't panic. (Note the earlier round's suggestion to subtract from `Instant::now()` would have reintroduced the underflow flagged before — this shares one code path instead.) Also de-flaked the second `test_dag_cleanup_stale` in `dag/src/tests.rs` (the one originally cited) the same way.

**`cross_node_state_hash` admin-signing concern** — not a bug: `is_admin` grants admin authority via an `Admin` *member row*, not only `meta.admin_identity`, so admin1/admin2 can each sign `MemberAdded`. Added a comment explaining this; the `is_member` assertions already prove all three ops were accepted.

**`set_updated_at` "after save"** — false positive as noted last round; `set_updated_at(base)` runs before the first save in every case.

chefsale:
Addressed in `925858c58`: `save__update_merkle_hash` now captures `hash_before` by reloading the persisted entity rather than reading the in-memory element, so the assertion no longer depends on `save` refreshing `element().merkle_hash` — a save that left the in-memory hash zeroed can't make the `assert_ne!` pass vacuously.

PR #3156, opened 2026-07-02
fix(context,runtime): bound local xcall cascade (breadth 8, depth 3)

chefsale:
## Summary

Follow-up to #3143 (the deferred xcall item). xcalls dispatch **locally** — each one sends a fresh `ExecuteRequest` back to the same `ContextManager` actor — and every spawned execution can itself queue more. So a single root call could:

- **recurse without bound** — a cycle `A→B→A→B…` never terminates, or
- **fan out `max_xcalls^depth`** — each level multiplies the number of local executions.

Each of those executions is a full WASM run + storage + delta commit, so this is a real local CPU/IO amplification (and liveness) hazard, not just latency.

## Fix — two composing bounds

1. **Breadth** (runtime): lower the per-execution `max_xcalls` default from **100 → 8**. This cuts the per-level multiplier that drives the exponent (`8³ = 512` vs `100³ = 1,000,000`).
2. **Depth** (context): carry an `xcall_depth` on `ExecuteRequest` — `0` for a direct/RPC call, `parent + 1` for an xcall target — and deny any further xcalls once an execution reaches `MAX_XCALL_DEPTH` (**3**). Denials use the existing `XCallOutcome::Denied` path, so they're observable. Depth is what actually stops cycles (an `A↔B` ping-pong dies at hop 3) and caps chain length.

Worst-case local executions triggered by one root call: `8 + 8² + 8³ = 584` — bounded and reasoned, down from unbounded.

Both are simple knobs: `max_xcalls` is an existing `VMLimits` field; `MAX_XCALL_DEPTH` is a documented const in the dispatch handler (can be promoted to config later if needed).

## Testing

- `cargo build -p calimero-runtime -p calimero-context -p calimero-context-client -p calimero-node` — clean, no warnings
- `cargo clippy` on the touched crates — clean
- `cargo test -p calimero-runtime --lib` — 117 pass (incl. the updated `max_xcalls == 8` default-limits assertion)

**Follow-up:** the xcall dispatch loop is a detached fire-and-forget task with no existing integration-test coverage; a node-level test that wires two contexts into an `A↔B` cycle and asserts `XCallOutcome::Denied` at depth 3 would be a good addition but is a standalone effort.

### DIFF
--- a/crates/context/primitives/src/client/mod.rs
+++ b/crates/context/primitives/src/client/mod.rs
@@ -1460,13 +1460,16 @@ impl ContextClient {
         aliases: Vec<Alias<PublicKey>>,
         atomic: Option<ContextAtomic>,
     ) -> Result<ExecuteResponse, ExecuteError> {
-        self.execute_with_origin(context_id, executor, method, payload, aliases, atomic, None)
-            .await
+        self.execute_with_origin(
+            context_id, executor, method, payload, aliases, atomic, None, 0,
+        )
+        .await
     }
 
     /// Like [`execute`](Self::execute), but tags the run with the source
     /// context that dispatched it via `xcall`, surfaced to the guest as
-    /// `env::xcall_origin()`. Pass `None` for a direct/RPC call.
+    /// `env::xcall_origin()`. Pass `None`/`0` for a direct/RPC call; the xcall
+    /// dispatch loop passes `Some(source)` and `parent_depth + 1`.
     #[allow(clippy::too_many_arguments, reason = "execution context is wide")]
     pub async fn execute_with_origin(
         &self,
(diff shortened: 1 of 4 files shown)


chefsale:
Addressed the review in `5f1b69180`:

**1. Public `xcall_depth` bypass (best catch)** — Fixed. The handler now clamps depth to 0 whenever `xcall_origin.is_none()`, enforcing the invariant *no origin ⟺ depth 0* at the entry point. A direct/RPC `ExecuteRequest` can no longer sidestep the cap by supplying a high `xcall_depth`; only executions our own dispatch loop tagged with `Some(origin)` carry a trusted depth.

**2. `xcall_depth + 1` overflow** — Fixed. Switched to `saturating_add(1)`. It was already unreachable (the `>= MAX` deny gates the `+1`), but this makes the arithmetic sound regardless of future tuning.

**3. Off-by-one / "hops" wording** — Clarified the docs (no logic change; the count was already right). To be precise: an execution *at* `MAX_XCALL_DEPTH` still runs but its xcalls are denied, so executions span depths `0..=MAX_XCALL_DEPTH` = `MAX_XCALL_DEPTH` xcall hops beyond the root, and the **spawned** executions are children at depths `1..=DEPTH` → `B + B² + B³ = 584` at `B=8, DEPTH=3`. That 584 figure in the PR is unchanged and correct.

Build + clippy clean on `calimero-context`.

PR #3157, opened 2026-07-02
refactor(client): remove dead process-global SESSION_CACHE token singleton

chefsale:
## Summary

Follow-up to #3142 (issue #10 tail). Removes the process-global `SESSION_CACHE` — a `Lazy<SessionTokenCache>` static that held JWT access/refresh tokens in plaintext for the whole process lifetime, shared globally and `Clone`-proliferated.

## Why this is safe (behavior-preserving)

Its only consumer was meroctl's `authenticate_with_session_cache`, which used the cache purely as a *"have we already authenticated this URL this process?"* flag:

- The looked-up value was bound to `_cached_tokens` and **discarded** — real tokens always came from `FileTokenStorage`/config.
- `prepare_connection` calls the helper **at most once per process** (the `--node`, `--api`, and default branches are mutually exclusive, and `--node` pre-checks `config.get_connection` first). So the process-global cache is **always empty when probed** — the cache-hit branch never fired, and the flow always took the authenticate-and-persist path.

Removing the never-hit branch leaves the observable behavior unchanged while dropping a plaintext token store that outlived every request.

## Changes

- Delete `SessionTokenCache`, the `SESSION_CACHE` static, and `get_session_cache` from `crates/client` (drop the now-unused `once_cell` dep + `Arc`/`RwLock` imports; `lib.rs` no longer re-exports `get_session_cache`).
- Simplify meroctl's helper and rename it `authenticate_with_session_cache` → `authenticate_and_connect` (name no longer references a cache that's gone); update its 3 call sites.

Net: **-121 / +34** lines.

## Testing
- `cargo test -p calimero-client` → 62 passed.
- `Cargo.lock` delta is a single line (drops the `once_cell` edge from `calimero-client`; the package stays for other crates).

Note: pre-1.0, so the removed `pub` items (`SessionTokenCache`, `get_session_cache`) are dropped outright rather than deprecated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1173,7 +1173,6 @@ dependencies = [
  "chrono",
  "eyre",
  "hex",
- "once_cell",
  "percent-encoding",
  "reqwest 0.12.28",
  "serde",
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -19,7 +19,6 @@ thiserror.workspace = true
 async-trait.workspace = true
 chrono = { workspace = true, features = ["serde"] }
 hex.workspace = true
-once_cell.workspace = true
 webbrowser.workspace = true
 base64.workspace = true
 zeroize.workspace = true
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -27,7 +27,7 @@ pub use client::{Client, ResolveResponse, ResolveResponseValue};
 pub use connection::{AuthMode, ConnectionInfo};
 pub use errors::ClientError;
 pub use eyre::Result;
-pub use storage::{get_session_cache, JwtToken};
+pub use storage::JwtToken;
 pub use traits::{ClientAuthenticator, ClientConfig, ClientStorage};
 // Re-export common types
 pub use url::Url;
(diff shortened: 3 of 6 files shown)


chefsale:
Addressed in 26690b37c.

**`authenticate_and_connect` always re-authenticated (security-reviewer + patterns-reviewer):** good catch — the removed cache had masked this, and both reviewers converged on it. The function now checks `FileTokenStorage` for usable, non-expired tokens for `node_name` before launching the browser flow, and reuses them when present. This:
- makes the helper **idempotent** (safe to call more than once — the concern about future call sites/retry loops),
- avoids a browser popup on repeat `--api`/default invocations (a strict UX improvement over both the old and the first-revision behavior),
- stays safe for a stored-but-revoked (not-yet-expired) token: the connection's own request path recovers via `401 → refresh → re-auth`, so skipping the eager prompt never dead-ends.

Doc-bot nudges (AGENTS.md / CONTRIBUTING.md / architecture HTML): not applicable — this removes dead internal state and tightens a CLI helper; no public architecture or documented CLI surface changed.

`cargo test -p calimero-client` → 62 passed; `cargo clippy -p calimero-client -p meroctl --all-targets -- -D warnings` → clean.

PR #3160, opened 2026-07-02
test(client): e2e coverage for proactive-refresh & single-flight auth

chefsale:
## Summary

Follow-up to #3142 (issue #10 tail — validation). The proactive-expiry refresh and single-flight paths added in the client auth hardening had **no automated coverage** — the existing tests only exercised reactive (401-driven) refresh. These add wiremock-driven tests that drive the real `ConnectionInfo` state machine end-to-end (`ensure_auth_header` → `refresh_or_reauth` → request) against a mock server, using real JWTs with `exp` claims.

A live-node run isn't automatable (the meroctl auth flow needs interactive browser OAuth), so this closes the gap at the integration level instead.

## Tests

| Test | Proves |
|---|---|
| `expired_token_is_refreshed_before_request` | A stored JWT with a **past** `exp` is refreshed *before* the GET is sent (no 401 round-trip). The GET mock only matches the **fresh** bearer, so a stale send would 404 — a passing test proves the refresh was proactive. |
| `token_expiring_soon_is_refreshed_proactively` | Same, for a token inside the proactive-refresh skew window (not yet expired) — exercises the `expires_soon` branch distinct from `is_expired`. |
| `concurrent_expired_requests_refresh_once` | 8 concurrent expired requests produce **exactly one** `/auth/refresh` (`expect(1)`), proving the single-flight lock coalesces them instead of spending a rotating refresh token 8 times. |

Each is regression-catching by construction: the header-filtered GET mock fails the request if the stale bearer is sent, and `expect(1)` on the refresh mock fails if single-flight breaks.

## Testing
- `cargo test -p calimero-client` → 65 passed (was 62).
- `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.
- Test-only; no production code, `Cargo.toml`, or `Cargo.lock` changes.

### DIFF
--- a/crates/client/src/connection.rs
+++ b/crates/client/src/connection.rs
@@ -23,7 +23,7 @@ use crate::traits::{ClientAuthenticator, ClientStorage};
 /// isn't sent with a token that lapses in flight (which would waste a 401
 /// round-trip). Also the window within which a still-valid token is refreshed
 /// proactively rather than reactively.
-const TOKEN_REFRESH_SKEW_SECS: i64 = 30;
+pub(crate) const TOKEN_REFRESH_SKEW_SECS: i64 = 30;
 
 /// Maximum size of a control-plane JSON response body we will buffer. Admin-API
 /// responses are small; a multi-gigabyte body from a hostile or buggy node must
--- a/crates/client/src/lib.rs
+++ b/crates/client/src/lib.rs
@@ -36,7 +36,22 @@ pub use url::Url;
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[cfg(test)]
-mod tests;
+mod test_support {
+    use base64::Engine as _;
+
+    /// Build a minimal JWT (`header.payload.sig`) whose payload carries the
+    /// given `exp` (seconds since the Unix epoch). Only the `exp` claim is read
+    /// by the client, so the header/signature segments are placeholders. Shared
+    /// by the `storage` and `tests` modules to avoid drift.
+    pub(crate) fn jwt_with_exp(exp_unix: i64) -> String {
+        let payload = base64::engine::general_purpose::URL_SAFE_NO_PAD
+            .encode(format!("{{\"exp\":{exp_unix}}}"));
+        format!("aGVhZGVy.{payload}.c2ln")
+    }
+}
 
 #[cfg(test)]
 use tokio_test as _;
+
+#[cfg(test)]
+mod tests;
--- a/crates/client/src/storage.rs
+++ b/crates/client/src/storage.rs
@@ -239,13 +239,7 @@ impl TokenValidation {
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    fn jwt_with_exp(exp: i64) -> String {
-        let payload =
-            base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(format!("{{\"exp\":{exp}}}"));
-        // header and signature segments are irrelevant to `exp` extraction.
-        format!("aGVhZGVy.{payload}.c2ln")
-    }
+    use crate::test_support::jwt_with_exp;
 
     #[test]
     fn decode_exp_from_valid_jwt() {
--- a/crates/client/src/tests.rs
+++ b/crates/client/src/tests.rs
@@ -1392,9 +1392,13 @@ async fn concurrent_expired_requests_refresh_once() {
     let server = MockServer::start().await;
     // Single-flight: 8 concurrent expired requests must produce exactly ONE
     // `/auth/refresh` (a rotating refresh token would be spent 8 times
-    // otherwise) and 8 successful GETs carrying the fresh bearer. The 50 ms
-    // refresh delay holds the lock long enough that all 8 tasks are guaranteed
-    // to contend for it rather than each racing an instant refresh.
+    // otherwise) and 8 successful GETs carrying the fresh bearer.
+    //
+    // Contention is made deterministic rather than timing-dependent: a `Barrier`
+    // releases all 8 tasks into the request path simultaneously, and the 50 ms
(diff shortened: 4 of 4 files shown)


chefsale:
Addressed all three in f3f8a7ceb (the other reviewer marked it LGTM):

- **Fragile skew-window offset (69b5f833a862):** `TOKEN_REFRESH_SKEW_SECS` is now `pub(crate)` and the expiring-soon token uses `now + TOKEN_REFRESH_SKEW_SECS / 2`, so the test can't silently drift off the `expires_soon` branch if the window is retuned.
- **GET mock had no `.expect()` (5bdd5f9975e8):** the guarded GET mock now takes a `get_hits` count — `expect(1)` for the single-request tests and `expect(8)` for the concurrent one — so a short-circuited/missing request is caught on server drop rather than only via a downstream panic.
- **Discarded `MemAuth` counter (06c45e360e8b):** `auth_client_with_token` now returns the calls counter, and each proactive-refresh test asserts `== 0`, proving the refresh path was taken and not the interactive-auth fallback (which would mint a valid token and mask a refresh regression).

`cargo test -p calimero-client` → 65 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

chefsale:
Addressed all four in b5704766b:

- **Single-flight test may not contend (035f071aa8a6):** the best catch — the `/auth/refresh` response now has a 50 ms delay, so the first task holds the single-flight lock long enough that all 8 tasks are guaranteed to queue behind it. Without it, a fast instant refresh could let the first task finish before the others start, passing `expect(1)` for the wrong reason (no contention). Added a `refresh_delay` param; the single-request tests pass `Duration::ZERO`.
- **Confusing failure mode (70710d4aa042):** the concurrent task result is now `.expect("request should succeed with the fresh bearer")` — a stale-bearer/broken-single-flight failure names the cause instead of a bare `unwrap` panic.
- **Duplicate `header` import (6bd2a982747d):** folded into the top-level `use wiremock::matchers::{body_json, header, method, path};`.
- **Duplicated `jwt_with_exp` (8517657a369e):** moved to a crate-level `#[cfg(test)] test_support` module shared by both the `storage` and `tests` modules, so there's a single definition.

`cargo test -p calimero-client` → 65 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

chefsale:
Addressed in f163ce296 (both agents flagged the same point):

- **Single-flight contention was timing-dependent (3698265214bc, d3c4272ddf42):** added an 8-way `tokio::sync::Barrier` so all tasks are released into `ensure_auth_header` simultaneously, and kept the 50 ms refresh delay so the first task holds the single-flight lock across a real await while the other seven — already past the barrier — queue on it. Contention is now structural, not reliant on all 8 spawns landing within the delay window on a loaded runner.

One clarification on the non-determinism concern: the refresh mock's `.expect(1)` can't be *exceeded* even under adverse scheduling — `refresh_or_reauth` re-checks the stored token against the stale one it captured and returns early when they differ, so at most one refresh happens regardless of timing. The real risk was the opposite (a *false green* where no contention occurred), which the barrier now rules out.

`cargo test -p calimero-client` → 65 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

chefsale:
Addressed in fda489fbd:

- **Self-documenting failures (935bcad38b01, 21c3be2fb006):** added a lower-priority stale-bearer GET catch-all with `expect(0)` — a request carrying the stale token now fails structurally (descriptive 400 body) instead of an opaque unmatched-route 404 — and each concurrent task carries its index so the panic names the offending task.

- **Determinism (2274147437f1):** the test runs on tokio's **default current-thread runtime**, where this is deterministic: the first task past the barrier acquires `auth_lock` and then parks on the refresh's network `await` (a real localhost round-trip to the mock server on its own runtime, held open 50 ms) *while still holding the lock*. Cooperative scheduling then polls the other seven, which all reach `lock().await` and block before the refresh can complete — so they exercise the coalescing path, not an already-refreshed early-exit. A single-thread runtime can't let the first task's refresh run to completion without yielding at that await.

  I considered gating the refresh response on a size-9 barrier as suggested, but wiremock's `Respond` is synchronous and can't `await` a barrier, so that isn't expressible. Between the current-thread scheduling guarantee and the structural backstops (`expect(1)` on `/auth/refresh`, `expect(0)` on the stale-bearer GET, `expect(8)` on the fresh GET), a regression in either proactive refresh or single-flight fails this test deterministically.

`cargo test -p calimero-client` → 65 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

chefsale:
Addressed in 2a010336e:

- **Scheduling assumption made structural (78ca5db018eb, 28ca1e4c3ea3):** pinned `#[tokio::test(flavor = "current_thread")]` on the single-flight test, so the cooperative-scheduling guarantee is now compile-time, not a runtime default. On one thread the first task cannot complete its refresh without yielding at the network `await` while holding `auth_lock`, so the other seven provably block on the lock — the exact contention scenario. (The suggested atomic-counter check would be redundant: `Barrier::new(8)` already cannot release until all 8 tasks have arrived, so concurrent entry is guaranteed by construction.)

- **Consistent failure messages (abd63a9b78f8, 💡):** the two proactive single-request tests now use the same `unwrap_or_else(|e| panic!("stale bearer sent (proactive refresh regressed): {e}"))` as the concurrent test, so a catch-all 400 surfaces the root cause instead of a bare unwrap panic.

- **`pub(crate)` on the test helper (c5eca3e94de7, 📝):** keeping `pub(crate)` deliberately — it's accurate (the helper is used across sibling modules `storage::tests` and `tests`) and avoids any `unreachable_pub`-style lint that a bare `pub` in a private module can attract.

`cargo test -p calimero-client` → 65 passed; `cargo clippy -p calimero-client --all-targets -- -D warnings` → clean.

Issue #2474, opened 2026-05-25
Harden TEE replica bootstrap: derive founding admin from DAG genesis (RootOp::NamespaceCreated) instead of first-KeyDelivery TOFU


xilosada:
## Context
PR #2473 (TEE fleet replica bootstrap) added `seed_bootstrap_admin_if_absent` (`crates/context/src/group_store/namespace_governance.rs`), which seeds a bootstrapping `ReadOnlyTee` replica's namespace **founding admin from the first `KeyDelivery` signer** (trust-on-first-use). This is correct in the owner-driven happy path — the owner/verifier is the one who admits the node and publishes the `KeyDelivery` — and is proven working end-to-end (a replica admitted by the owner bootstraps and replicates the owner's data).

## Problem (Medium, surfaced by Cursor Bugbot on #2473)
A `KeyDelivery` carries no signer-authority check on apply (it only verifies `envelope.recipient == our namespace identity`), so **any current namespace member holding the group key can mint one**. A non-owner racing an *unsolicited* `KeyDelivery` to a brand-new replica before the owner's could pin the **wrong local admin**, blocking subsequent valid governance ops from the true owner.

- **Blast radius:** local DoS on that one replica only — the seeded `GroupMeta` is node-local and never gossiped, so there is **no cross-node escalation**.
- **Does not self-heal:** the meta-exists guard turns a later correct `KeyDelivery` into a no-op; recovery needs namespace teardown + re-sync.
- Not exploitable in a single-owner-claims flow; requires an actual malicious co-member.

## Proper fix
Emit a replayable `RootOp::NamespaceCreated { founder }` at namespace creation so a bootstrapping replica derives the founding admin **authoritatively from the synced DAG genesis**, dropping the `KeyDelivery` TOFU entirely. The namespace root founder is currently never recorded as a replayable op (`RootOp` has no genesis variant), so this is a **versioned `RootOp` borsh-discriminant / wire change** that needs a schema-change plan (back-compat for existing namespaces).

Full threat model is documented in the `seed_bootstrap_admin_if_absent` doc comment. Follow-up to #2473.

xilosada:
### Still applies as of master (2026-06-21) — refresh + scope note

Re-verified against current HEAD. All three pillars hold unchanged:

[...]

PR #3162, opened 2026-07-02
fix(create_group): warn on clock error and roll back on mid-write failure

chefsale:
## Summary

Two small pre-existing rough edges in `create_group.rs`'s async body, hardened as a follow-up to #3144. Both are low-severity tidiness fixes, not live bugs.

**1. Silent epoch timestamp.** The async body computed the group's `created_at` with `SystemTime::now().duration_since(UNIX_EPOCH).map(...).unwrap_or(0)` — a pre-1970 clock silently stamps `created_at = 0`. #3144 already made the *reservation's* copy of this log a `warn!`, leaving the two inconsistent. This reuses the already-warned `reservation_now` for the final meta, so there's one timestamp, one warning, and the reservation and its final replacement agree.

**2. Partial-write leak.** The local-row writes (meta → founder member → default caps → group encryption key → name metadata) used bare `?`. A failure partway through (e.g. a transient store fault storing the group key) returned `Err` leaving the earlier rows behind — only the later *genesis-apply* failure path had a rollback, and #3144's outer cleanup map only drops the meta. This factors the genesis rollback into a shared `rollback_local_group_rows` helper and calls it on any mid-write failure too, so a retry with the same id starts from a clean slate. The group key is dropped by the id that was actually stored (`None` when the failure preceded storing it); the namespace identity is still deliberately preserved for deterministic retry (#2474).

## Tests
- `rollback_removes_every_local_group_row` — seeds all six row types, rolls back, asserts each is gone (catches a wrong-key deletion).
- `rollback_is_scoped_to_the_target_group` — a bystander group is untouched.
- `rollback_without_key_id_or_name_is_a_partial_no_op` — the `None` key-id / `has_name = false` guards don't panic and still drop the rows that exist.

Full `calimero-context` lib suite (221) passes; clippy `-D warnings` and `cargo fmt --check` clean.

### DIFF
--- a/crates/context/src/handlers/create_group.rs
+++ b/crates/context/src/handlers/create_group.rs
@@ -3,9 +3,9 @@ use std::sync::Arc;
 use actix::{ActorFutureExt, ActorResponse, Handler, Message, WrapFuture};
 use calimero_context_client::group::{CreateGroupRequest, CreateGroupResponse};
 use calimero_context_client::local_governance::{NamespaceOp, RootOp};
-use calimero_context_config::types::AppKey;
+use calimero_context_config::types::{AppKey, ContextGroupId};
 use calimero_primitives::context::GroupMemberRole;
-use calimero_primitives::identity::PrivateKey;
+use calimero_primitives::identity::{PrivateKey, PublicKey};
 use calimero_store::key::GroupMetaValue;
 use calimero_store::types::ApplicationMeta as ApplicationMetaValue;
 use calimero_store::Store;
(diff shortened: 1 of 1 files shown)


chefsale:
Addressed all three in 8cc104c69:

- **`name.is_some()` vs actually-written (warning)** — real subtle bug, fixed. Added a `name_written` flag set to `true` only after `MetadataRepository::set_group` succeeds, and pass it (instead of `name.is_some()`) to `rollback_local_group_rows` at both the mid-write and genesis call sites. So a failure before the name write — or an invalid name that was skipped — no longer asks the rollback to delete a row that was never written.
- **`.expect()` in non-test code (warning)** — removed. The write closure now returns the group key id on success (`Ok(key_id)`), so the success path takes it straight from the `Ok(..)` match arm; no `.expect()` on the threaded Option.
- **IIFE outer-mutation smell (suggestion)** — kept the outer `group_key_id`/`name_written` mutation, but only because the **error path** rollback genuinely needs to know how far the closure got before failing (which rows to unwind); the success path no longer relies on it. Added a comment explaining exactly that, since returning the values from the closure only covers the success path, not the partial-progress an Err needs.

`calimero-context` lib tests (221) pass; clippy `-D warnings` and `cargo fmt --check` clean.

chefsale:
Addressed in dff506cc1:

- **Bystander test omits capabilities (suggestion)** — done, and went a bit further: the scoped-rollback test now asserts the bystander's default-caps AND name-metadata rows survive too, so every row type the helper deletes is covered against a wrong-group-id delete.

- **Closure mutates outer `group_key_id`/`name_written` (warning)** — I'm keeping this one, and here's why the outer mutation isn't removable. The rollback runs on the **error** path, where the closure returns `Err(..)` with no value, so it can't carry "what was written so far" in its return. Crucially, an error can happen *after* the group key is stored — the name `set_group?` is the last fallible step — so on that path the rollback must know the key was already written (`group_key_id = Some`) to drop it. Returning `Ok((key_id, name_written))` only covers the success path; the partial-write tracking an `Err` needs still has to live in outer state. The `group_key_id = Some(key_id)` set is therefore not redundant with `Ok(key_id)` — it's set at the key-store point precisely so a *later* name-write failure can still roll the key back. A `Vec<WrittenRow>` accumulator would express the same thing with more machinery than a 5-row create warrants; the load-bearing reason is called out in a comment above the closure. Happy to revisit if you'd prefer a scopeguard-style helper.

`calimero-context` lib tests (221) pass; clippy `-D warnings` and fmt clean.

PR #3204, opened 2026-07-06
fix(network): per-overlay rendezvous cookies poisoned global discovery, breaking namespace-invite joins

frdomovic:
## Why this PR exists

Namespace invitations were failing in production between two internet-separated nodes (laptops in Spain and Croatia, testing mero-meet). The desktop app showed only **HTTP 0** — no status, no body. Accepting the same invite eventually worked after several retries, which made it look flaky rather than broken.

## What the logs showed

On the joining node, every `POST /admin-api/namespaces/:id/join` spun for the full discovery deadline and died at `namespace_join.rs:224`:

```
could not open a namespace-join stream to any mesh peer for namespace fb19e206… (deadline 45000ms, elapsed 44095ms, excluded 0)
```

Meanwhile, from ~100 seconds after node startup until shutdown (2h45m, 137 occurrences, every ~15s, zero recoveries):

```
WARN Rendezvous discovery failed rendezvous_node=… namespace=Some(Namespace("/calimero/devnet/global")) error_code=InvalidCookie
```

## Root cause — cookie poisoning

`rendezvous_discover()` sends **two kinds** of discover requests per tick:

1. a **global** discover (`/calimero/devnet/global`) that carries the stored per-peer cookie for incremental results, and
2. **per-overlay** discovers (one per under-connected namespace/group key) that deliberately pass `cookie=None`.

The `Discovered` event handler stored **whichever cookie arrived** into the single per-peer cookie slot. But rendezvous cookies are **bound to the namespace they were issued for**: `libp2p-rendezvous` server rejects a discover whose cookie namespace doesn't match the discover namespace (`CookieNamespaceMismatch` → `ErrorCode::InvalidCookie`, see `libp2p-rendezvous-0.17.1/src/server.rs:492-498`).

So the failure loop was:

1. A per-overlay response lands → its overlay-namespace cookie overwrites the global cookie.
2. The next global discover sends that cookie → server replies `InvalidCookie`.
3. The `DiscoverFailed` handler only logged a warning — the poisoned cookie was **never cleared**, and per-overlay responses kept re-poisoning it.
4. Global discovery stays dead for the lifetime of the process.

Namespace-join is the one flow with **no fallback**: the joiner isn't a member of the namespace yet, so per-overlay discovery, gossipsub topics and the context mesh can't find the inviter — the global discover is the only door (the comment in `rendezvous_discover` says exactly this). With it dead, `join_namespace` burns its whole deadline and 500s.

### Why it looked like it "worked before" (rc.9 / rc.10)

This code is unchanged since #2514 — it is a race, not a version regression. Poisoning only starts once the node has under-connected overlay keys firing per-overlay discovers; fresh test nodes with few namespaces rarely hit it, while long-lived production nodes carrying many namespaces/groups hit it within the first ~2 minutes, every boot. Retries eventually "worked" only when the two nodes got connected through an unrelated door (shared-context sync via the persistent peer cache) — at that point the join stream opened over the existing connection.

### Why the client saw HTTP 0

The admin handler holds the HTTP request for the full 45s discovery deadline, then returns a **bare 500** — `parse_api_error` deliberately masks untyped reports ("Internal server error"), and the desktop client times out before that anyway → HTTP 0 with nothing to debug from.

## The fix

1. **Stop the poisoning** (`handlers/stream/swarm/rendezvous.rs`): only store a cookie when it was issued for the global namespace (`cookie.namespace() == rendezvous_config.namespace`). Per-overlay discovers never use the stored cookie, so dropping their cookies loses nothing — overlay result sets are small and fully re-polled by design.
2. **Self-heal if a cookie is ever rejected anyway** (server restarts also invalidate cookies): on `DiscoverFailed(InvalidCookie)`, clear the stored cookie (`clear_rendezvous_cookie`, new in `discovery/state.rs`) and force an immediate cookie-less re-discover. This cannot loop: a cookie-less discover has no cookie to reject.
3. **Surface the real failure** (`admin/handlers/namespaces/join_namespace.rs`): map the "could not open a namespace-join stream to any mesh peer" report to a typed **504** with an actionable message ("could not reach any member of this namespace… make sure the inviting node is online and retry") instead of the masked generic 500.

## What was tried / considered and rejected

- **Clearing the cookie on failure alone** (fix 2 without fix 1) — insufficient: the very next per-overlay response re-poisons the slot, so global discovery would still fail on most ticks.
- **Storing cookies per-namespace** — unnecessary complexity: only the global discover uses incremental cookies; per-overlay discovers intentionally do a full poll each time.
- **Typed error across the node→server boundary for fix 3** — the report crosses actor/client layers where a downcast isn't reliable; matching the fixed message string from our own repo is the pragmatic option and is scoped to one handler.

## Verification

- New regression test `test_clear_rendezvous_cookie_forgets_rejected_cookie` (store → clear → gone; clearing for an unknown peer is a no-op).
- `cargo test -p calimero-network`: 103 + 9 integration tests pass; `cargo test -p calimero-server`: 49 pass; clippy clean on both crates.
- Field evidence in the production log matches every link of the chain above (timestamps and counts in the PR discussion if needed).

### DIFF
--- a/crates/network/src/discovery/state.rs
+++ b/crates/network/src/discovery/state.rs
@@ -530,6 +530,16 @@ impl DiscoveryState {
             .and_modify(|info| info.update_rendezvous_cookie(cookie.clone()));
     }
 
+    /// Forget the discovery cookie stored for `rendezvous_peer`. Called when
+    /// the server rejects the cookie (`InvalidCookie`) so the next discover
+    /// starts from scratch instead of re-sending the rejected cookie forever.
+    pub(crate) fn clear_rendezvous_cookie(&mut self, rendezvous_peer: &PeerId) {
+        let _ = self
+            .peers
+            .entry(*rendezvous_peer)
+            .and_modify(|info| info.clear_rendezvous_cookie());
+    }
+
     /// Remember that `listener_id` was opened against `relay_peer` as a
     /// relayed listener. The ListenerClosed handler uses this to route
     /// recovery for the quota-denied case where libp2p tears the listener
@@ -947,6 +957,12 @@ impl PeerInfo {
         }
     }
 
+    fn clear_rendezvous_cookie(&mut self) {
+        if let Some(ref mut info) = self.rendezvous {
+            info.clear_cookie();
+        }
+    }
+
     fn update_relay_reservation_status(&mut self, status: RelayReservationStatus) {
         if let Some(ref mut info) = self.relay {
             info.update_reservation_status(status);
@@ -1064,6 +1080,10 @@ impl PeerRendezvousInfo {
     fn update_cookie(&mut self, cookie: Cookie) {
         self.cookie = Some(cookie);
         self.last_discovery_at = Some(Instant::now());
+    }
+
+    fn clear_cookie(&mut self) {
+        self.cookie = None;
     }
 
     pub(crate) const fn registration_status(&self) -> RendezvousRegistrationStatus {
--- a/crates/network/src/discovery/state_tests.rs
+++ b/crates/network/src/discovery/state_tests.rs
@@ -1627,3 +1627,31 @@ fn reachability_reachable_to_unreachable_when_last_address_removed() {
     assert!(gone.disable_autonat_server);
     assert!(gone.has_actions());
 }
+
+#[test]
+fn test_clear_rendezvous_cookie_forgets_rejected_cookie() {
+    let mut state = DiscoveryState::default();
+    let peer_id = PeerId::random();
+    let cookie = Cookie::for_namespace(Namespace::from_static("test"));
+
+    state.update_peer_protocols(&peer_id, &[RENDEZVOUS_PROTOCOL_NAME]);
+    state.update_rendezvous_cookie(&peer_id, &cookie);
+    assert_eq!(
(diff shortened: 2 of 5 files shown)


frdomovic:
**Why CI was green while this bug lived in production** — and what now pins it:

The existing rendezvous integration tests call `behaviour_mut().rendezvous.discover(...)` directly, which **bypasses the `NetworkManager` event handler** where the cookie is stored — so the poisoning path was never exercised. And no test ever replayed a cookie across namespaces, so the server-side rejection contract was invisible too.

Added in 37be88b7:
1. `is_global_discovery_cookie` extracted + unit-tested (global cookie stored; per-overlay / namespace-less cookies rejected) — pins the handler-side rule.
2. `cross_namespace_cookie_is_rejected_and_cookieless_discover_recovers` — end-to-end against a real rendezvous server: per-overlay discover → replay its cookie on a global discover → `InvalidCookie` → cookie-less global discover finds the peer again. If someone ever 'simplifies' cookie handling back to the old behaviour, or a libp2p upgrade changes the cookie-namespace contract, CI fails instead of shipping another silent discovery outage.

chefsale (approved):

Issue #3207, opened 2026-07-06
Open group invitation is multi-claimable by any syncing identity until expiry (no joiner-binding / no single-use nonce)


chefsale:
## Summary

An admin-signed **open group invitation** (`GroupInvitationFromAdmin`) is **not bound to a specific joiner** and has **no single-use / used-nonce marker**. Once the first joiner's `RootOp::MemberJoined` op lands on the namespace DAG, the embedded admin-signed invitation — including its `secret_salt` — is public. Any other identity that syncs the DAG can lift that `signed_invitation`, wrap it in its **own** `MemberJoined` op (signed by itself), and self-join the group. This repeats for any number of distinct identities until the invitation's `expiration_timestamp`.

This is invitation **reuse by distinct self-signing identities**, effectively turning a single open invitation into an unlimited-use invite link within its validity window.

## What is already mitigated (not the concern here)

The naive "replay the same op to inject a third party" attack is closed:

- Outer op signature is verified before apply — `crates/governance-store/src/namespace/governance.rs:187` (`op.verify_signature()`).
- `apply_member_joined` requires **`signer == member`** — `crates/governance-store/src/namespace/membership.rs:136`. So a `MemberJoined` op can only add the identity that signed it (self-join only; no third-party injection).
- Exact-op idempotency (`contains_op(delta_id)`, `governance.rs:213`) + per-member direct-row dedup (`has_direct_member`, `membership.rs:71`) — the same op / same member can't double-apply.
- Deterministic expiry exists via `RootOp::MemberJoinedAt { … joined_at }`; `MemberJoined` with a non-zero `expiration` now requires a signed `joined_at` and rejects `joined_at > expiration` (`membership.rs:47-56`).

These bound the blast radius (self-join only, once per identity, before expiry) but do **not** make the invitation single-invitee or single-use.

## The gap

`GroupInvitationFromAdmin` (`crates/context/config/src/types.rs:831`) carries:

```
inviter_identity, group_id, expiration_timestamp, secret_salt, invited_role
```

There is **no invitee/joiner pubkey** covered by the admin signature, and **no persisted redeemed-invitation nonce**. `secret_salt` only provides pre-reveal MEV/front-run protection; once the invitation is revealed on the public DAG it is spent, and the invitation degrades to "anyone who syncs can self-join until expiry."

### Reproduction (conceptual)

1. Admin issues an open invitation; Alice joins via `MemberJoined { member: Alice, signed_invitation }`.
[...]

PR #3209, opened 2026-07-06
test(storage): probe reachability of the comparisons state-sync path

chefsale:
## What

Adds three WARN-level, greppable probes (`target: \"sync_probe\"`) at the only sites that can exercise the legacy state-based `StorageDelta::Comparisons` / `compare_trees` sync path. **No behavior change** — probes only log.

- **`delta.rs` `commit_root`** — the *sole* producer of a `Comparisons` wire artifact.
- **`root.rs` `apply_comparisons`** — fires whenever any `Comparisons` delta (including the zero-byte no-op that deserializes to `Comparisons(vec![])`) reaches `Root::sync`.
- **`root.rs` `apply_actions` `Action::Compare` arm** — fires if a `Compare` action is ever *received* from peer traffic (the trigger that becomes a Comparisons emission).

## Why

We want to delete the `compare_trees` / `Comparisons` subsystem (it's deletion-unaware and superseded by HashComparison/LevelWise/DAG-CausalActions). Static analysis strongly suggests it's already **unreachable in production**:

- `Action::Compare` is pushed only inside `apply_action`, into `context.actions` (not `comparisons`), and only during `__calimero_sync_next` — a **state-op whose artifact isn't broadcast**.
- Authored/stored DAG deltas are always `CausalActions` built from local writes (Add/Update/DeleteRef), never `Compare`, never empty.
- So peers' `__calimero_sync_next` should always receive `CausalActions` → `apply_actions`, never the `Compare` arm, never `apply_comparisons`.

The one residual uncertainty is whether the node-side state-delta handler ever re-authors/propagates a DAG delta from a sync-apply outcome (which would carry `Compare` actions to peers). Rather than guess, these probes let a real multi-node run answer it with data.

## How to read the result

Grep node logs (guest probes surface via the `env::log` bridge) for:

```
sync_probe: apply_comparisons invoked
sync_probe: received Action::Compare in delta
sync_probe: commit_root emitted StorageDelta::Comparisons artifact
```

Run something that forces Merkle divergence (concurrent merges + deletes across nodes). **If zero `sync_probe` lines appear**, the path is confirmed dead and the follow-up PR removes the whole subsystem (and the now-false convergence comment at `interface.rs:2043-2049`).

- Full removal of `compare_trees` / `compare_affective` / `generate_comparison_data` / `Comparison` / `StorageDelta::Comparisons` / `apply_comparisons`, pending a clean run.
- Related: open-invitation multi-claim gap tracked in #3207.

### DIFF
--- a/crates/storage/src/collections/root.rs
+++ b/crates/storage/src/collections/root.rs
@@ -397,6 +397,17 @@ where
         comparisons: Vec<Comparison>,
         ctx: &crate::interface::ApplyContext,
     ) -> Result<(), StorageError> {
+        // Reachability probe (issue 1): fires whenever a `StorageDelta::Comparisons`
+        // delta reaches `Root::sync` — including the zero-byte no-op artifact that
+        // deserializes to `Comparisons(vec![])` and seeds a root comparison below.
+        // If this never fires across a multi-node sync run, no peer ever delivers a
+        // Comparisons delta and the state-based comparison sync path is dead.
+        tracing::warn!(
+            target: "sync_probe",
+            comparison_count = comparisons.len(),
+            "sync_probe: apply_comparisons invoked (Comparisons sync path reached)"
+        );
+
         if comparisons.is_empty() {
             push_comparison(Comparison {
                 data: <Interface<S>>::find_by_id_raw(Id::root()),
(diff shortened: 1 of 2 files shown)


chefsale:
Superseded by #3215 (the actual removal). This probe branch's purpose — hard-abort instrumentation to confirm, via the multi-node divergence e2e, that the state-based Comparisons sync path is never reached — is complete: no abort fired, so the path is dead and #3215 removes it. Not for merge.

PR #3220, opened 2026-07-13
refactor(storage): remove the now-inert Action::Compare variant

chefsale:
## What

Follow-up to #3216. After the state-based Comparisons sync was removed, **nothing constructs `Action::Compare`** — its only producers (`compare_trees` and the `apply_action` pushes) are gone. This removes the dead variant and the ~20 defensive match arms that only existed to handle it, across `storage`, `context`, `node`, `op-adapter`, `runtime`, and `tools/merodb`.

## Wire compatibility (the one subtle part)

`Action` previously **derived** borsh, giving positional tags `Add=0, Compare=1, DeleteRef=2, Update=3`. Naively deleting `Compare` would renumber `DeleteRef` 2→1 and `Update` 3→2 — silently corrupting **every** persisted and in-flight delta (the live variants), not just the dead one.

So `Action` now has a **hand-rolled `BorshSerialize`/`BorshDeserialize`** that keeps `Add=0, DeleteRef=2, Update=3` and rejects tag 1 — byte-identical encoding for every surviving variant. Same approach #3216 used for `StorageDelta`. A new `action.rs` test pins the tag bytes, roundtrips each variant, and asserts tag 1 is rejected.

## Also

- Fixes now-stale comments that referenced the removed `Action::Compare` push (`store.rs`, `tests/concurrency.rs`, the stale-update test).
- Drops the op-adapter `payload_from_action` `None` arm (kept the `Option` return to avoid a signature change rippling to callers).

## Verification

- `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` — clean
- `cargo test -p calimero-storage --features testing` — **668 passed, 0 failed** (incl. the new wire-tag test)

### DIFF
--- a/crates/context/src/handlers/execute/signing.rs
+++ b/crates/context/src/handlers/execute/signing.rs
@@ -45,7 +45,6 @@ pub(crate) fn sign_authorized_actions(
                 let nonce = *deleted_at;
                 (metadata, nonce)
             }
-            Action::Compare { .. } => continue,
         };
 
         // STAMP THE NONCE BEFORE COMPUTING THE PAYLOAD.
@@ -105,7 +104,6 @@ pub(crate) fn sign_authorized_actions(
         let metadata = match action {
             Action::Add { metadata, .. } | Action::Update { metadata, .. } => metadata,
             Action::DeleteRef { metadata, .. } => metadata,
-            Action::Compare { .. } => continue,
         };
         let sig_data = match &mut metadata.storage_type {
             StorageType::User {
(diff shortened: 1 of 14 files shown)


chefsale:
Addressed the AI-review findings in a9c57ae7f:

- **Fixed (valid):** the `Interface::apply_action` doc comment still referenced the removed `Action::Compare` — dropped the "Generates Compare action for hash verification" line and reworded the `ActionNotAllowed` `# Errors` bullet to its real surface (storage-type access violations, e.g. a `Frozen` delete or an unauthorized `Shared`/`User` write; `ActionNotAllowed` is still returned by apply_action for those, so the bullet was kept rather than deleted).
- Swept the rest of the touched code for other stale `Compare` mentions — none remain (the two `grep` hits are `compare_level_nodes` in HashComparison sync and a generic English "compare", both unrelated).

Non-actionable:
- **Docs-site doc-bot:** path-based heuristic (fires on any `crates/**` change). `docs-site/` has zero references to `Action::Compare` / `compare_trees` / `Comparisons`, so nothing to update.
- **Cursor Bugbot:** paused (spend limit) — no findings produced.

PR #3222, opened 2026-07-13
fix(auth): require admin to create a root key via POST /admin/keys

frdomovic:
## Security finding

**[H] Non-admin can mint root key** (security review item 3) — `crates/auth/src/auth/permissions/validator.rs`, route `crates/auth/src/api/handlers/root_keys.rs`.

`POST /admin/keys` (`create_key_handler`) calls `provider.create_root_key(...)`, which mints a **ROOT key with `["admin"]` permissions**. But the route was mapped to the permission `keys:create`, which is strictly weaker than `admin`. So a key holding only `keys:create` could mint a full admin root key — a privilege escalation. The sibling client-key handler already enforces `admin` inline; this root-key route did not.

## Before

```rust
("/admin/keys", HttpMethod::POST) => vec![Permission::Keys(KeyPermission::Create)],
```

A `keys:create`-only token passes `validate_permissions` for this route and mints an admin root key.

## After

```rust
("/admin/keys", HttpMethod::POST) => vec![Permission::Admin(AdminPermission)],
```

The route now requires `admin`, mirroring the decision already made for `PUT /admin/keys/:id/permissions` (`updating_key_permissions_requires_admin`). `GET /admin/keys` (list) stays a scoped `keys:list` permission. `keys:create` maps to no other route, so nothing legitimate loses access.

## How to test

Automated (added, `cargo test -p mero-auth creating_root_key_requires_admin`):
- `POST /admin/keys` resolves to `[Permission::Admin]`.
- A `keys:create`-only token is **denied**; an `admin` token passes.
- `GET /admin/keys` still resolves to `[Permission::Keys(List)]` (list not admin-gated).

Manual: with a token scoped to only `keys:create`, `POST /admin/keys` must return 403 (`X-Auth-Error: permission_denied`); with an `admin` token it succeeds.

## Follow-up (out of scope here)
Defense-in-depth: additionally cap a newly-minted key's permissions to a subset of the caller's (via the `CallerPermissions` extension + `validate_grantable_permissions`) instead of hardcoding `["admin"]`. Noted for a later change.

Full suite green: 88 lib + 6 integration tests, clippy clean.


chefsale (approved):
Approved, but will need fixing of the desktop app @frdomovic

PR #3234, opened 2026-07-14
fix(node): bail out of sync sessions when the concurrency semaphore is closed

rtb-12:
## Summary
- `sync_session_bridge.rs` had two call sites (`Responder` and `Initiator` session futures) that acquired the `max_concurrent` semaphore with `.acquire_owned().await.ok()`, discarding the `Err` case. If the semaphore is ever closed, that pattern silently proceeds with `_permit = None`, bypassing the concurrency cap entirely instead of failing loudly.
- Nothing today calls `Semaphore::close()`, so this is hardening against a future shutdown-drain path rather than a fix for an observed bug.
- Extracted the acquire into two small helpers, `acquire_responder_permit` and `acquire_initiator_permit`, that explicitly handle the closed case:
  - Responder: logs a `warn` and returns without running the session.
  - Initiator: logs a `warn` and returns a synthetic-failure `Err`, which flows through the existing outcome/`result_tx` pipeline the same way the `ContextBusy` drop path already does — required by the module's #2317 invariant that every accepted job must emit a `SyncSessionResult`, or the context's `last_sync` would stay `None` forever.

## Test plan
- Added `responder_permit_bails_on_closed_semaphore` and `initiator_permit_emits_synthetic_failure_on_closed_semaphore` unit tests exercising the two extracted helpers directly against a closed `tokio::sync::Semaphore` (constructing a full `SyncSessionActor` for this needs a full `SyncManager`, which needs a live `NodeClient`/`ContextClient`/`NetworkClient` — impractical for a unit test, so the acquire-or-bail seam is tested directly instead).
- `cargo fmt` (clean)
- `cargo clippy -p calimero-node --all-targets` (clean)
- `cargo test -p calimero-node` — 322 passed, 0 failed

### DIFF
--- a/crates/node/src/sync_session_bridge.rs
+++ b/crates/node/src/sync_session_bridge.rs
@@ -398,6 +398,54 @@ impl Actor for SyncSessionActor {
     }
 }
 
+/// Acquire a responder concurrency permit, or bail if `concurrency` is
+/// closed. Extracted out of the inline session future so the bail branch
+/// is unit-testable without spinning up a full `SyncSessionActor`.
+///
+/// Nothing today closes `concurrency` (no shutdown-drain exists yet), so
+/// the `Err` arm is currently unreachable — hardening against a future
+/// close() silently lifting the max_concurrent cap via `_permit = None`.
+async fn acquire_responder_permit(
+    concurrency: &Arc<Semaphore>,
+    peer_id: PeerId,
+) -> Option<tokio::sync::OwnedSemaphorePermit> {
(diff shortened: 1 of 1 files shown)


rtb-12:
Closing: nothing today ever calls `Semaphore::close()` on this semaphore, so the `Err` branch this PR hardens is unreachable in the current codebase - there is no live bug to fix. Revisit this only alongside whatever future change actually introduces a shutdown-drain path that closes the semaphore.

PR #3240, opened 2026-07-14
fix(governance): make apply-time authority deterministic; forward secrecy on self-leave

chefsale:
Fixes the two reported issues, plus two related bugs found while tracing them.

## 1. Self-leave did no key rotation

A rotation is minted by whoever **publishes** the op that triggers it. That works for an admin-initiated `MemberRemoved` — the publisher stays in the group and wraps the new key for everyone who remains. It cannot work for `MemberLeft`: the publisher **is** the leaver, who would have to mint the very key they are being cut off from (and would keep it). Peers reject a rotation from a non-admin regardless.

So `MemberLeft` did no rotation at all. A leaver kept the namespace key and every Restricted subgroup key indefinitely; on a namespace-leave they went on reading the root and every Open subgroup under it.

The two halves are now split across nodes. `MemberLeft`'s apply records what is owed (a replicated `GroupPendingKeyRotation` row, written inside the deterministic apply so every node derives the same worklist). A new `GroupOp::GroupKeyRotated`, published by a **remaining admin**, carries the new key and discharges the row. A listener reacts to departures and also drains the persisted worklist on startup — the only thing covering an admin that was offline when the leave applied.

**No election, no quorum.** Any remaining admin may rotate. If several race they mint different keys, but the keyring already converges (highest epoch, ties broken by the larger key id — a total order, identical on every node), and safety survives the race because *every* competing key excludes the leaver. Electing a rotator would add a coordination problem, and a liveness dependency on the elected node, to solve something the crypto already solves.

An Open subgroup on a fully-Open chain correctly owes nothing: it is encrypted with the namespace key, so a per-subgroup rotation there would revoke nothing.

## 2. Cascade authority pre-scan → divergence

`#3190` removed the per-descendant cap scan and gated once at the root. But the root gate it moved to was still nondeterministic, so the reported failure mode survived: `EphemeralProjectionAuthorizer` returns `None` when the op's cited ancestry isn't folded — which its own comment calls *"a normal mid-backfill state"* — and `PermissionChecker` fell back to the **live** resolver on `None`. Live is a *different* cut (this replica's current one), so a replica that had folded a concurrent capability revoke rejected an op its peers applied. The reject path never advances the DAG head, so everything descending from that op stalled on that replica, and its live state never changed back — the retry failed identically forever.

`AtCutAuthorizer::can_resolve_cut` now separates the two reasons the resolver abstains:

- **no cut to resolve against** (a genesis op, or no apply-auth context — emit path, local apply, read side, tests) → live is correct, nothing contradicts it;
- **the cut is real but unfolded here** → live is wrong → `ApplyError::AuthorityUndecidable`.

The DAG already evicts-and-refetches on apply error without advancing the head or burning the nonce, so refusing *is* parking: the op retries once the missing history arrives. `can_resolve_cut` defaults to `true`, so every emit-time and read-side caller is byte-identical.

### A bigger, untouched instance of the same bug

`ctx.settings()` was built **without** `with_apply_auth`, and `GroupSettingsService::permissions()` constructed a fresh **live** checker. So `TargetApplicationSet`, `GroupMigrationSet`, `UpgradePolicySet`, `DefaultCapabilitiesSet`, and `SubgroupVisibilitySet` authorized from live rows, always. The single-group `TargetApplicationSet` had precisely the divergence #3190's comment claims to have fixed — and `DefaultCapabilitiesSet` is a capability op gating on live capabilities. `GroupCreated` read the membership rows directly too. Both now resolve at the cut.

## 3. Keyring split on a non-admin removal (found while tracing)

The publish gate for removal is admin **or** `MANAGE_MEMBERS`; the rotation's receive gate is **strict admin**, checked against the namespace identity that signs the outer op. The publisher minted and stored the new key **unconditionally**, at the top epoch. So a non-admin `MANAGE_MEMBERS` holder removing from a Restricted subgroup would: adopt a key that became its `load_current_key`, attach a rotation every peer silently rejected, and thereafter encrypt every group op under a key **no other node held** — peers buffering them as undecryptable indefinitely. A group-wide liveness break, not a skipped rotation.

Now fail-closed: the rotation is checked publishable before any mutation (before the local apply, since bailing after would leave the removal committed locally but never published), and re-asserted at the mint site. Groups encrypted under the namespace key mint no key on removal, so a `MANAGE_MEMBERS` holder may still remove there.

## Verification

22 new tests. The load-bearing ones drive a fake at-cut authorizer that **deliberately disagrees with the live rows in both directions** — verified they fail with the wiring reverted and pass with it.

- **e2e** (`local_governance_node_e2e.rs`): boots a real node, has a member self-leave a Restricted subgroup, and asserts the production chain (apply → op-event → listener → handler → publisher) records the debt, discharges it, and actually changes the group's current key while retaining the old one. Confirmed it fails with its own diagnostic when the listener spawn is removed, so it is not passing vacuously.
- **park-and-retry**: the same signed op applied twice — once against a projection that cannot resolve its cut (must leave nothing behind, nonce unburned), then against one that can (must apply in full). Parking is only correct if it is recoverable.
- **crypto**: no envelope in a rotation is unwrappable by the departed member, while a remaining member unwraps the real key; two concurrent rotations converge on one key regardless of arrival order.

Full workspace green; `cargo clippy --workspace --all-targets -- -D warnings` clean; `Cargo.lock` untouched.

## Two things to weigh in review

**Wire posture.** `GroupKeyRotated` is appended at the **end** of `GroupOp` so every existing ordinal is preserved, and `SIGNED_GROUP_OP_SCHEMA_VERSION` deliberately does **not** move — it is enforced strictly-equal on decode, so bumping it would make older peers reject *every* op rather than just the new one. Rollout is therefore operator-discipline, same as the cascade variants: deploy everywhere before any node self-leaves a Restricted group. Happy to make it a hard flag day instead.

**Residual window.** Between a leave and its rotation landing, ops are still encrypted under the key the leaver holds. The deny-list stops them *writing*, but a leaver who keeps watching gossip can read that window. It is bounded by rotator liveness and observable via the pending row. I chose **not** to stall publishes while a rotation is pending — the availability cost is real — but that is a config decision worth a second opinion.


chefsale:
## The 25 `app-migration-e2e` failures are pre-existing on master — not from this PR

**Evidence:** #3241 is a **comment-only** change to `crates/governance-store/src/ops/group.rs`, branched from unmodified `62c01a1b3` (current master). It exists solely to trip this suite's path filter. It fails **the same 25 scenarios, identically**:

| | failures |
|---|---|
| #3240 (this PR) | 25 |
| #3241 (clean master, comment-only) | 25 |
| **failing only on this PR** | **0** |

### Why the suite didn't catch this earlier

`app-migration-e2e` is path-filtered. Its last master run was `1a64fa961` (Jul 13). Three commits have landed since, and **none touches a trigger path**, so the suite never ran against them:

- `db8a1211b` fix(crypto): internal AES-GCM nonces + ed25519 verify_strict hardening (#3226)
- `f54192cd0` fix: atomic config writes and stray no_mdns override (#3227)
- `62c01a1b3` fix(crypto): derive AEAD key via HKDF-SHA256 and reject small-order points (#3228)

This PR touches `crates/governance-store/src/ops/**`, so it is simply the first change to trip the filter on top of them.

### What's actually failing

Not governance. Node 2 (the receiver) logs:

```
WARN calimero_context::handlers::execute: lazy upgrade: target blob fetch failed
WARN calimero_context::handlers::execute: ladder hop blocked; proceeding with current application
WARN calimero_node_primitives::client::blob: Failed to query DHT for blob   (x6)
```

It cannot fetch the target application blob from the DHT, so it keeps the old application and never migrates — which is exactly the assertion that fails (`'Executing migration' had 0 hits` on node 2). mDNS discovery and dialing both succeed, so it is not peer discovery. **Zero occurrences of `authority undecidable`** in the node logs, so the park-and-retry change in this PR never fires; the governance ops apply normally.

The ~10 scenarios that *pass* are the ones that don't require a cross-node blob fetch, which fits.

### This PR's own CI

`Rust` (build + `clippy -D warnings` + full workspace test suite, including the new e2e rotation test) — **pass, 16m2s**. `Schema downgrade guard`, `Lint`, `ABI Validation`, `Build Apps`, `Build WASM fixtures` — all pass.

I'd suggest tracking the master blob-fetch regression separately; it blocks cross-node migration on master today and is masked by the path filter.

PR #3262, opened 2026-07-16
chore(server): replace /tee/verify-quote endpoint with a local `merod tee probe` self-check

xilosada:
Closes the unauthenticated `/tee/verify-quote` admin endpoint and moves its verification-assurance role to a **local CLI self-check** instead of an internet-reachable HTTP surface.

**Why not just delete it:** the endpoint had no production runtime consumer, but the mero-tee `node-image-gcp-staging-probe` used it to prove a freshly-built TDX node image actually verifies attestations (accepts a real quote, rejects wrong-nonce/tampered). That coverage matters and isn't reproducible by a non-TDX core unit test.

**This PR:**
- removes the `/tee/verify-quote` HTTP handler + route + `TeeVerifyQuote*` primitives;
- adds `merod tee probe [--json]` — a local self-check that generates a quote on the node's real TDX and runs `verify_attestation` for positive / wrong-nonce / tampered cases, emitting a JSON verdict (mirrors the existing `merod kms probe`). Refuses `--mock-tee` on real-attestation nodes; exits nonzero on the no-TDX path.

The mero-tee staging probe is being updated to invoke `merod tee probe` over SSH (like it already does for `kms probe`), preserving the anti-fake assurance with no network surface. `/tee/attest`, `/tee/info`, `/tee/fleet-join` unchanged.

Verified: check/clippy -p merod clean, fmt (1.88) clean, merod tests pass incl. 3 new probe tests.


xilosada:
Addressed the tampered-quote review finding in `a3fb2237`: `run_probe_checks` now classifies verify errors via an exhaustive `is_quote_rejection_error` match (no wildcard). Only genuine crypto/parse rejections (`QuoteParsingFailed` — incl. the mock header-corruption case, `QuoteConversionFailed`, `QuoteVerificationFailed`, `NonceMismatch`, `ApplicationHashMismatch`) or an `Ok` result with `quote_verified == false` count as tamper-rejected. Infra/transient errors (`CollateralFetchFailed`, etc.) now set `passed: false` with the error surfaced — same handling as `wrong_nonce` — so a PCS/network flake can no longer masquerade as successful tamper detection. Added two classifier unit tests.

PR #3268, opened 2026-07-16
feat(tee-attestation): document the crypto-only verifier contract and add additive policy_valid

xilosada:
## What

Two additive changes to `calimero-tee-attestation`:

1. **Document the verifier contract.** `is_valid()` is crypto/structural validity **only** — it deliberately does not consult `tcb_status`/`advisory_ids`, the measurement registers (`mrtd`/`rtmr0-3`/`mrsigner`), or enforce an app-hash binding. That layering is intentional (callers own policy), but it was undocumented, which is exactly how a caller can end up trusting `is_valid()` alone. The docs now state the contract explicitly and name the enforcement sites that own policy.

2. **Add an opt-in `policy_valid` gate.** New `VerifierPolicy` + `VerificationResult::policy_valid(&VerifierPolicy) -> Result<(), PolicyRejection>`, so future callers get a safe-by-construction option instead of hand-rolling the policy loop (or forgetting it).

## Why

The verifier returns parsed measurements and TCB status for the *caller* to enforce. Today the live paths do enforce them (`admit_tee_node`, merod KMS), but nothing in the API stops a new caller from trusting crypto-only validity. This closes that footgun class at the API level.

## `policy_valid` semantics

Folds in, on top of the existing crypto checks:
- **TCB fail-closed** — `Revoked` rejected unconditionally (case-insensitive); an **empty** `allowed_tcb_statuses` defaults to the secure `{"UpToDate"}`, never fails open.
- **Mock** — a mock quote is accepted only when `accept_mock` is set (an allowlisted `"Mock"` is not a standing bypass token).
- **Measurements** — MRTD is strict (an empty allowlist is a rejection); `rtmr0..3` allowlists are enforced when non-empty, skipped when empty.
- **App-hash** — rejected when unbound.

These deliberately mirror `tcb_status_allowed` from #3023 (`calimero-governance-store::membership::policy_rules`), replicated rather than imported because `calimero-tee-attestation` is a leaf crate — the intentional duplication is recorded in a module doc comment so the two can't silently drift.

`PolicyRejection` is a typed enum (`NotCryptoValid`, `TcbRevoked`, `TcbNotAllowed`, `MockNotAccepted`, `EmptyMrtdAllowlist`, `MeasurementMismatch`, `AppHashNotBound`) with `Display`/`Error`, so callers get actionable errors rather than a bare `bool`.

`policy_valid` is guaranteed **never weaker than `is_valid()`** — clearing `require_app_hash` changes only the reported reason, never the verdict (pinned by a test).

## Additive — no consumer impact

`is_valid()` keeps its name and exact body; `verify_attestation`'s signature and `VerificationResult`'s fields are unchanged. The only public-surface change is additions. Verified by diffing `verify.rs` against master with comment lines filtered out — the result is empty.

This matters because `calimero-tee-attestation` is `publish = true` and consumed downstream, where `is_valid()` is relied on for **crypto-only** semantics with policy layered on top. Tightening it would have been a breaking behavior change; this is not. **No rev bump required.**

## Not included (deliberate)

Existing callers (`admit_tee_node`, merod KMS `enforce_attestation_policy`) keep their own enforcement — they have richer, mode-specific errors and the ReleaseStrict/Config split, so churning them is risk without benefit. `policy_valid` exists to make *future* callers safe. They could delegate later; that's a follow-up, not this PR.

## Verification

- `cargo check -p calimero-tee-attestation` + `--workspace` clean
- `cargo clippy -p calimero-tee-attestation --all-targets` clean (zero warnings)
- `rustup run 1.88.0 cargo fmt --check` exit 0
- `cargo test -p calimero-tee-attestation` — **20 passed**, covering empty-TCB fail-closed, `Revoked` rejected even when allowlisted and even with `accept_mock`, mock gated, MRTD mismatch/empty rejected, RTMR empty-skip, app-hash unbound, `tcb_status: None`, and the happy path.

### DIFF
--- a/crates/tee-attestation/AGENTS.md
+++ b/crates/tee-attestation/AGENTS.md
@@ -14,7 +14,7 @@ Generates and verifies Intel TDX attestation quotes, binding them to a nonce and
 # Build
 cargo build -p calimero-tee-attestation
 
-# Test (crate has no unit tests of its own; exercised via callers)
+# Test (the `policy` module has unit tests; the rest is exercised via callers)
 cargo test -p calimero-tee-attestation
 ```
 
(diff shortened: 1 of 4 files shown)


xilosada:
Fixed in `5d790af2`. The reviewer was right: `QuoteBody` has no `mrsigner` field — the signer register is `mrsignerseam` (alongside `mrseam`/`mrtd`/`mrconfigid`/`mrowner`/`rtmr0..3`). Corrected all three doc references in `verify.rs` (L21/L44/L68). Naming a non-existent field is exactly the confusion this PR exists to prevent.

One correction to the finding: `AGENTS.md` was **not** affected — it has no `mrsigner` reference; the three `verify.rs` doc comments were the only occurrences. `generate.rs`/`policy.rs` already used `mrsignerseam` correctly.

fmt (1.88), `cargo doc` (0 warnings), and `cargo check` all clean.

PR #3293, opened 2026-07-23
test(e2e): cold-cache context auto-resync A/B repro [diagnostic, do not merge]

chefsale:
Diagnostic A/B for the mesh-connectivity bug (members online but no gossip mesh; 3/4 contexts never synced). See the scenario file header for the full hypothesis.

**Experiment:** identical to `sync-resilience-peer-restart-autoresync` (which passes) **minus the pre-restart priming write**, so node-2's peer-identity cache for the context stays cold. After restart, node-1 writes the first delta; node-2 must converge with `trigger_sync: false`.

**Reading the result:**
- **`Test (sync-resilience-cold-cache-autoresync)` FAILS** → isolates the cache-population gap (root cause #2: the durable peer-identity cache is only fed by the state-delta receive path, not governance) → fix = seed the cache from the authenticated governance observation.
- **PASSES** → the cache is repopulated by another path (e.g. join-time sync); the real-node failure is the connectivity/relay black hole (#1), which a fresh merobox NAT setup can't reproduce → needs real-node network logs (`/admin-api/network/status`, relay-reservation / rendezvous / DCUtR log lines).

Not for merge — this is a probe to pick which fix to write.

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -419,6 +419,9 @@ jobs:
           - workflow: sync-resilience-peer-restart-autoresync
             file: workflows/sync-resilience-peer-restart-autoresync.yml
             app: scaffolding-e2e
+          - workflow: sync-resilience-cold-cache-autoresync
+            file: workflows/sync-resilience-cold-cache-autoresync.yml
+            app: scaffolding-e2e
           - workflow: sync-resilience-peer-pause
             file: workflows/sync-resilience-peer-pause.yml
             app: scaffolding-e2e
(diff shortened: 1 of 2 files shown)


chefsale:
Closing: this was a diagnostic probe, and it did its job. The cold-cache variant PASSED (node-2 auto-converged a first delta after restart with a cold peer-identity cache), matching the primed sibling — which rules out the cache-population gap (#2) as the reproducible cause on a fresh merobox NAT setup. Combined with (a) the relay-reservation path no longer being gated on advertise_address in current code and (b) sync-resilience-peer-restart-autoresync already proving relay+restart+gossip-push works over NAT, the real-node 'no mesh' failure is a runtime/environmental connectivity degradation that CI can't reproduce. Next step is capturing /admin-api/network/status + relay/rendezvous/DCUtR logs from the affected node.

Issue #3296, opened 2026-07-23
Node logs "Mesh peer discovery exhausted all retries" with members online — cold contexts have no dial target (peer-identity cache fed only by the state-delta path)


chefsale:
## Summary

A node that is otherwise healthy (WASM calls and RPCs all succeed) logs `Mesh peer discovery exhausted all retries (context mesh + namespace fallback)` (`crates/node/src/sync/peers.rs:214`) for one or more contexts, and state deltas for those contexts reach no other member — even though the members are online. Despite the "mesh" wording, the check does **not** inspect the grafted gossipsub mesh: `sync/peers.rs` asks for `subscribed_peers(topic)`, wired to `gossipsub.all_peers()` (`crates/network/src/handlers/commands/subscribed_peers.rs:23`, commented "The full subscriber set, not the grafted mesh"), i.e. **connected peers that are subscribed to the topic**. Broadcast then uses `flood_publish` (`crates/network/src/behaviour.rs:282`), which fans out to that subscriber set. So the real condition is **"zero connected+subscribed peers for this topic"** — the node never established a connection to the other members on that topic, so both discovery and delta delivery come up empty.

Observed in the field (Slack, 2026-07-23): a node with 4 contexts where **only 1** (`7ccPG…`) had a dialable peer; the other 3 had none, and their topics stayed empty. Environment involved NAT.

## Impact

- **Severity:** reliability / availability (not data-loss or security). The node looks healthy and serves local reads, but is **silently isolated** on the affected context/namespace topics — it neither receives nor pushes deltas there, so those contexts stop converging.
- **Blast radius:** any node that holds a context it **joined via governance but has not yet received a state delta for**, especially behind NAT (no incidental inbound dials to compensate). It is per-topic, so a node can be fine on one context and dark on another (as observed: 1 of 4 healthy).
- **Real-world consequence:** a member appears "in the group" (membership is recorded, lists work) yet never sees others' updates and their updates never propagate — indistinguishable to the user from "sync is broken," with only a WARN to go on.

## Steps to reproduce

Not yet reproduced in CI — it is environment-specific (connectivity / NAT / cold peer cache). CI actually proves the *happy* NAT path works: `sync-resilience-peer-restart-autoresync` restarts a NAT'd node, relies on gossip **push** (`trigger_sync: false`), asserts `ReservationReqAccepted`, and passes. So a faithful repro must isolate the **cold-cache** condition specifically:

[...]

PR #3299, opened 2026-07-24
fix(context): self-heal a "joined, pending key" member via membership-driven key recovery

### LINKED ISSUE #3295: Namespace joiner left "joined, pending key" has no key-delivery reconciler; recovery relies on a manual re-join within the 5s KeyDelivery fallback
## Summary

A node that joins a namespace while the gossip mesh to the inviter is thin, still forming, or recovering from a partition is deliberately left in a **"joined, pending key"** state by the #3292 not‑stranded fix — it records local membership but has no group key yet. The problem is that this pending‑key state has **no self‑healing path**: the group key is only ever fetched *synchronously inside the join RPC*, via (1) a direct `request_namespace_join` stream to a currently‑reachable mesh peer, or (2) a gossip `KeyDelivery` fallback bounded by `ContextManagerConfig::key_delivery_fallback_wait` (**default 5s**, `crates/context/src/lib.rs`). If neither lands in that window, the handler (`crates/context/src/handlers/join_group.rs`, serving the admin `join_namespace` endpoint) returns a typed `KeyDelivery timed out for group ...` error and the node stays keyless **with no background re‑attempt**. Nothing reacts to "a namespace peer became reachable" by re‑fetching the key — none of the four context background listeners (`auto_follow`, `self_purge`, `tee_subgroup_admit`, `rotation_listener`) cover this. Recovery therefore depends on an external actor manually re‑issuing a join that happens to land inside the 5s fallback after connectivity returns.

## Impact

- **Severity:** reliability / availability (not data‑loss, correctness, or security). A joiner ends up with usable *membership* but no *group key*, so it cannot decrypt or participate in the group until it re‑acquires the key.
- **Blast radius:** any node joining under a thin/forming/recovering namespace mesh — cold‑start, partition heal, flaky or NAT'd connectivity. Single‑node symptom, but it's the join path every new member takes.
- **Real‑world consequence:** this is the same family as the original Slack report (2026‑07‑23, "Fran joined fine; Sandi didn't"). The *stranded‑as‑non‑member → HTTP 500* half was fixed (#3292 + #3294, both merged). But a joiner under a slow/recovering mesh can still end up **keyless‑pending with no automatic recovery**: connectivity is fine, membership is recorded, yet the node never gets the key because the one 5s window elapsed and nothing retries.

## Steps to reproduce

Deterministic merobox repro exists: `apps/scaffolding-e2e/workflows/group-join-mesh-not-ready.yml` (the recovery half).

1. Two nodes. node‑1 installs the app, creates a namespace + context, and issues an open invitation.
2. Apply a libp2p‑only partition between node‑1 and node‑2 (`partition_peers`; both stay RPC‑reachable).
3. node‑2 joins while partitioned — the direct request finds no peer, so it records local membership and returns "joined, pending key" (no group key).
4. Heal the partition (`heal_peers`).
5. node‑2 must now obtain the group key.

**Actual:** if libp2p redial + gossipsub GRAFT + the `MemberJoined → targeted KeyDelivery` round‑trip take longer than the 5s fallback after the direct request fails, the join RPC returns `KeyDelivery timed out for group ...` and node‑2 stays keyless. In CI, node‑2 was still unable to reach node‑1 **>60s** after heal.
**Expected:** once connectivity is restored, node‑2 acquires the group key and becomes usable **without** a hand‑issued re‑join.

> Note: the scenario itself was recently hardened (merged via #3294) to *actively drive* the reconnect with `wait_for_sync(trigger_sync)` so CI stops flaking — but that only masks the product gap in the test. The underlying behavior (5s one‑shot fallback, no reconciler) is unchanged.

## Root cause — why this happens

1. **The fallback window is far shorter than a real post‑reconnect round‑trip.** 5s must cover: libp2p redial after heal, gossipsub re‑GRAFT of the `ns/<id>` mesh (several ~1s heartbeats), node‑2's `MemberJoinedAt` reaching node‑1, node‑1 publishing a targeted `KeyDelivery`, and node‑2 receiving it. That routinely exceeds 5s.
2. **The join is one‑shot.** On timeout the handler returns a typed error and does **not** schedule any background re‑attempt (`crates/context/src/handlers/join_group.rs`). The node crate has a `join_namespace_with_retry`, but the admin `join_namespace` endpoint goes through `join_group`, which has no retry.
3. **The direct path only uses peers reachable *at join time*.** `request_namespace_join` serves the key only if a mesh peer holding it is reachable now; after a partition heal there may be no prompt automatic redial, so it finds no peer and falls through to the 5s fallback.
4. **There is no "key‑missing" reconciler.** No listener observes `load_current_key() == None && has local membership` and re‑drives acquisition when a peer later appears.

## Potential solutions

Roughly increasing robustness / effort:

1. **Retune the default** — raise `key_delivery_fallback_wait` from 5s to ~15–20s. Already operator‑tunable via `ContextManagerConfig`; cheap, low‑risk margin for a realistic post‑reconnect round‑trip. Does **not** fix the one‑shot nature — mitigation, not cure.
2. **Bounded retry on the key‑wait / wire in `join_namespace_with_retry`** — give the admin join path backoff‑retry so a single racy window doesn't strand the joiner. Middle ground; still bounded by the RPC lifetime.
3. **Active dial in the direct path** — have `request_namespace_join` dial the inviter / known `ns/<id>` subscribers rather than only using existing mesh peers, so post‑heal recovery doesn't depend on an incidental redial.
4. **Background key reconciler (durable fix)** — a fifth context listener that, for any group with local membership but no stored key, re‑attempts acquisition (direct request + re‑publish `MemberJoinedAt` to re‑trigger a targeted `KeyDelivery`) on a bounded backoff and/or when a namespace peer becomes reachable. Closes the gap regardless of timing and independent of any single RPC's lifetime.

**Recommendation:** ship **(1)** now as an immediate mitigation, and pursue **(4)** as the durable fix (optionally **(3)** to make the direct path robust on its own).

## Criteria for resolving

- [ ] A node that joins under a thin/healing mesh and initially fails to obtain the group key **automatically acquires it** once a namespace peer becomes reachable, with **no** external/manual re‑join.
- [ ] `list_group_*`, context reads, and writes become usable within a bounded time after connectivity is restored.
- [ ] Regression coverage: a merobox scenario (extend `group-join-mesh-not-ready.yml` to assert the joiner gets the key after `heal_peers` **without** a manual re‑join step) and/or a node integration test for the reconciler.
- [ ] `cargo fmt --check`, `cargo clippy -- -A warnings`, and `cargo test` pass.

## References

- #3292 — introduced the "joined, pending key" state (don't strand the joiner). *(merged)*
- #3294 — 403‑not‑500 mapping for non‑member list endpoints + test hardening of the scenario above. *(merged)*
- #2237 — readiness three‑phase contract (context for the mesh‑not‑ready cold‑start work).
- `crates/context/src/handlers/join_group.rs` — the join + key‑wait flow.
- `crates/context/src/lib.rs` — `ContextManagerConfig::key_delivery_fallback_wait` (default 5s).
- `apps/scaffolding-e2e/workflows/group-join-mesh-not-ready.yml` — the reproducing scenario.

chefsale:
Closes #3295.

## Summary

A node that joins a namespace while the mesh to the inviter is thin, still forming, or recovering from a partition records local membership but can miss the group key within the single in-RPC `key_delivery_fallback_wait` window (`join_group.rs`, was **5s**). It is then left "joined, pending key": usable membership, no key, cannot decrypt or participate.

Verifying against current code: the durable pull-based recovery the issue asks for (**solution #4**) *already exists* — `SyncManager::recover_missing_group_keys`, driven by the periodic interval tick, join, startup, readiness, and the namespace-op receive path. **The real remaining gap** is that its worklist is purely **op-driven**: `NamespaceRetryService::groups_awaiting_key` only lists groups the node is *buffering an undecryptable op* for. A member that is keyless with **no buffered op** (the quiescent "joined, pending key" state) never appears, so recovery hits `awaiting.is_empty()` and returns — the node stays keyless, and recovery depends on a manual re-join landing inside the 5s window. This is also the root cause of the `group-join-mesh-not-ready` CI flake seen on the previous merge.

## Changes

- **Mitigation (issue #1)** — raise `key_delivery_fallback_wait` default **5s → 15s** (`ContextManagerConfig`). A realistic post-reconnect round-trip (redial + gossipsub re-GRAFT + `MemberJoinedAt` → targeted `KeyDelivery`) routinely exceeds 5s. Operator-tunable; margin, not cure.
- **Durable fix (issue #4)** — surgical, extends the existing recovery rather than adding a new listener:
  - `NamespaceRetryService::groups_member_but_keyless` (gov-store `retry.rs`, + free fn + re-exports): groups where the node holds a direct membership row for its own namespace identity but no usable key, **independent of any buffered op**.
  - `recover_missing_group_keys` (node) unions those groups into its request set, asking for the group's **current** key (`key_id = None`) alongside the op-driven epoch-specific pairs. The interval tick already drives it, so a quiescent keyless member now re-acquires its key once a peer is reachable — no buffered op, no manual re-join.

## Tests / proof

- **gov-store unit test** `groups_member_but_keyless_reports_then_clears`: reports a member-keyless group with no buffered op, excludes a non-member subgroup (membership filter is load-bearing), clears once the key lands.
- **node integration test** `keyless_member_surfaces_for_recovery_without_any_buffered_op`: with an empty op-driven `groups_awaiting_key`, the membership-driven set still surfaces the keyless group that recovery now requests — the exact gap-closing signal.
- **merobox** `group-join-mesh-not-ready.yml`: asserts passive key recovery (`recovered group key via direct delivery` on the joiner) **before** the manual re-join, so it now proves the joiner self-heals per #3295's criteria. This also removes the flake's root cause.

Local gate: `cargo fmt --check` ✅, `cargo clippy` (gov-store/context/node, all-targets) ✅, gov-store lib (470) ✅, node `sync::manager` (46, incl. new) ✅.

## Notes / residual (by design)

A `Restricted` subgroup where the member holds the namespace key but not the subgroup key is treated as keyed by the membership-driven check — but that case still surfaces through the op-driven set the moment one of its subgroup-key-encrypted ops is buffered, so it is not stranded. Active-dial in the direct path (issue #3) is deferred: the interval retry plus the governance-fed dial targets from #3296 make it unnecessary for now.


chefsale:
Addressed both meroreviewer comments in 8f641f38d:

- **Shared helper (retry.rs:73):** extracted `NamespaceRetryService::holds_key_epoch` (group keyring → namespace keyring fallback); `groups_awaiting_key`, `awaited_group_keys`, and `groups_with_held_key_buffered_ops` now all route through it, so the resolution order lives in one place.
- **Per-loop allocation (retry.rs:106):** in `groups_member_but_keyless` the namespace-key lookup is hoisted out of the loop into a single `has_namespace_key` (with an early return when it's held); the loop now only reads each group's own keyring.

Behavior preserved — 470 gov-store lib tests green.

Also reverted the `group-join-mesh-not-ready.yml` change: its post-`heal_peers` reconnection + forced-sync step is independently flaky (same 120s-timeout signature that flaked on the #3296 merge and passed on rerun), so asserting passive recovery behind it would import that flake. The deterministic regression guard here is the gov-store unit test (`groups_member_but_keyless_reports_then_clears`) and the node integration test (`keyless_member_surfaces_for_recovery_without_any_buffered_op`).

chefsale:
**Update — root-caused the `group-join-mesh-not-ready` failure and dropped the timeout bump (d4c9b979e).**

The failure was a timing interaction with the test, not the durable fix. Under the scenario's partition the join RPC always runs its KeyDelivery fallback to the full timeout (it's `expected_failure`), so raising the wait 5s→15s lengthened the pre-heal partition by exactly 10s (verified in logs: the key-wait ran 05:53:02→05:53:17). That extra window tips node-2's kademlia past retaining a route to node-1, so after `heal_peers` it never redials (mDNS is dead in the CI runner — netlink unavailable), the mesh never reforms, and `wait_for_sync` times out. The scenario went flaky-pass on master → 3/3-fail on the branch on that one change alone.

**Reverted the 5s→15s bump; kept the durable fix.** The membership-driven `recover_missing_group_keys` pull is the actual cure for 'joined, pending key' (re-acquires the key from the interval tick, no buffered op, no manual re-join) — which is exactly what makes widening the one-shot in-RPC window unnecessary. Restoring 5s returned the scenario to master's timing; the full e2e suite (incl. `group-join-mesh-not-ready`) is now green.

Net change of this PR is now: (1) `groups_member_but_keyless` worklist, (2) `recover_missing_group_keys` unions it in, (3) shared `holds_key_epoch` helper, + unit/integration tests. The `key_delivery_fallback_wait` default is unchanged (5s).

PR #3314, opened 2026-07-25
fix(storage): persist root crdt_type on hash updates (JS concurrent-writer convergence)

chefsale:
## What

Follow-up to #3309. Fixes a bug that prevented JS-app concurrent-writer convergence from actually working end-to-end, found during real-node validation.

## The bug

`Index::update_hash_for` updated `own_hash`/`full_hash`/`updated_at` but **never `crdt_type`**. So a root entity's merge-dispatch tag was frozen at whatever it was first stored as (`add_root`).

#3309 stamps the `JsRoot` marker in `persist_root_state` on every JS-app write, so a JS root should route to the guest `__calimero_merge_root_state` on sync. But if the root was **first created opaque** (`crdt_type: None`) — which is exactly what happens on a **joiner** that materialises the root via sync before ever writing locally — that stamp was silently dropped on every subsequent local write. The root stayed opaque, was pushed to peers with the `Opaque` wire tag, was never deferred to the guest merge (`deferred_root_merges=0`), and the two nodes' roots **never converged** (persistent root-hash divergence).

Reproduced on real 2-node networks (merod built from #3309 + JS SDK J1): `user_storage` wrote on both nodes, all entities merged (`entities_merged=37`), but `deferred_root_merges=0` and roots stayed split.

## The fix

Thread `crdt_type` through `update_hash_for` (`None` = leave unchanged; callers only ever pass `Some` to SET a tag, never clear one). Both write paths re-assert the app root's tag:
- `save_internal` (local write) → a local write upgrades an opaque root to `JsRoot`.
- `write_pre_merged_root_state` (sync apply) → a sync-materialised root keeps its wire tag.

Non-root entities pass `None` (unchanged behavior). Non-opted-in JS apps never stamp a tag, so their roots stay opaque exactly as before.

## Tests

- New regression: `js_root_update_reasserts_crdt_type_marker` — a root created opaque, then given a `JsRoot` write, ends up **stored** as `JsRoot` (fails before the fix).
- `cargo fmt --check`, `clippy -p calimero-storage`, full `cargo test -p calimero-storage` (**673 passed**) all green.

## Rollout

Merge → `merod:edge` rebuild → the JS SDK side (calimero-sdk-js J1, PR incoming) can then re-enable the concurrent-writer example workflows. Tracks calimero-network/calimero-sdk-js#83.

### DIFF
--- a/crates/storage/src/index.rs
+++ b/crates/storage/src/index.rs
@@ -1183,6 +1183,7 @@ impl<S: StorageAdaptor> Index<S> {
         id: Id,
         merkle_hash: [u8; 32],
         updated_at: Option<UpdatedAt>,
+        crdt_type: Option<crate::collections::crdt_meta::CrdtType>,
     ) -> Result<[u8; 32], StorageError> {
         // RMW on this entity's index entry — serialize against a concurrent
         // `add_child_to` on the same entry so neither clobbers the other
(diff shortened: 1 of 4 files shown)


chefsale:
Good catch by MeroReviewer — this was a real miss, not a stale diff. The four `update_hash_for` call sites in `crates/storage/src/tests/index.rs` (lines 1114/1145/1206/1229) were updated to the new 4-arg signature locally (which is why my local `cargo test` passed with 673), but I forgot to `git add` that file, so the pushed commit was missing it and would have failed to compile.

Fixed: amended the commit to include `tests/index.rs` (all four calls now pass `None` as the 4th `crdt_type` arg) and confirmed `cargo test -p calimero-storage --no-run` builds clean. Force-pushed.

Re the doc-bot: `AGENTS.md`/`CLAUDE.md` don't need changes — this is a bugfix to existing `update_hash_for` behavior, not an architecture change; the relevant `crdt_type` semantics are already documented in the storage crate.

PR #3317, opened 2026-07-27
fix(blob): resolve namespace identity when signing blob requests

frdomovic:
## Summary

Cross-node blob transfer is broken for **every** namespace-backed context — that is, every context created under the group model, since `context create` requires `--group-id`. It fails at any size; a 1 MiB blob is as unreachable as a 1 GiB one.

`NodeClient::find_owned_identity` (`crates/node/primitives/src/client/blob.rs`) only accepted a `ContextIdentity` row carrying its own `private_key`. Under the group model a member's row is a **keyless marker** — its presence is the membership signal, and the signing key lives with the node's single namespace identity, resolved live. The lookup therefore returned `None` for contexts the node genuinely belongs to.

## Impact

Two independent breakages, both from that one miss:

1. **Announce silently skipped.** `PUT /admin-api/blobs?context_id=…` logs `Skipping announce: node is not a member of the requested context` (`crates/server/src/admin/handlers/blob.rs`) and publishes no provider record.
2. **Every outgoing blob request is unsigned.** `create_blob_auth_for_context` returns `None`, so the requester sends `auth: None`; the serving node's `is_signed_context_member` returns `false` at `crates/node/src/handlers/blob_protocol.rs` and replies `found: false`, surfacing as `404 Blob not found locally or in network`.

User-visible effect: any app sharing member-authenticated blobs cannot deliver bytes to a second node — file sharing, image attachments, and similar. App **bundles** are unaffected (they take the explicit public-blob path in `is_blob_access_authorized`).

## Steps to reproduce

Two nodes sharing a namespace + context, blob uploaded and announced on node1, fetched on node2:

```
node1: Announced blob 5eH5Vd… to network          <- announce OK (from WASM)
node2: Found 1 peers with blob … attempt=1        <- discovery OK
node2: Blob download failed … error=Blob not found on peer
```

node1 never logs `Processing blob request stream` — it bails at the auth gate before looking for a blob it is holding.

Reproduced with both membership flows, so it is not a setup artifact:
- explicit `namespace invite` → `namespace join` → `group join-context`
- merobox `create_mesh`'s flow (identity generated on the joining node first, then namespace auto-join) — what `apps/blobs/workflows/blobs.yml` uses

## Before → after

Same harness, two local nodes, blob uploaded+announced on node1 and fetched on node2, SHA-256 verified, node2 confirmed not to already hold the blob:

| Size | Before | After |
|---|---|---|
| 50 MiB | ✗ refused | ✓ 1 s, hash match |
| 100 MiB | ✗ refused | ✓ 2 s, hash match |
| 250 MiB | ✗ refused | ✓ 6 s, hash match |
| 500 MiB | ✗ refused | ✓ 13 s, hash match |
| 1024 MiB | ✗ refused | ✗ refused — correctly, by the 500 MiB cap: `Blob exceeds maximum allowed size of 524288000 bytes` |

The 1 GiB row is the *intended* limit (`MAX_BLOB_SIZE_BYTES`, `crates/network/src/handlers/commands/request_blob.rs`) and now reports a real reason instead of an auth refusal.

## Approach

`ContextClient::get_identity` already resolved both ownership shapes. Rather than add a second copy of a security-relevant lookup — which is how these two drifted apart in the first place — the resolution moves to `calimero_store::namespace_signer::resolve_owned_namespace_signer`, the lowest crate both callers already depend on. `calimero-context-client` depends on `calimero-node-primitives`, so the latter cannot reach the former; `calimero-store` is below both. The depth bound is a parameter so `calimero-store` needs no `calimero-context-config` dependency; both call sites pass `MAX_NAMESPACE_DEPTH`.

`calimero-context-client`'s private helper now delegates, so there is one implementation and the two paths cannot diverge again.

## Tests

5 new unit tests in `crates/store/src/namespace_signer.rs`: resolution through a multi-level group chain, the context-group-is-root case, `None` when the node holds no namespace identity, `None` for a non-group-backed context, and fail-loud (not resolve-to-ancestor) on a cyclic parent chain.

`cargo fmt --check`, `cargo clippy`, and `cargo test -p calimero-store -p calimero-node-primitives -p calimero-context-client` all pass.

## Notes for reviewers

- **The E2E gap that let this through is real and not closed by this PR.** merobox has no `download_blob` workflow step (only `upload_blob`, `delete_blob`, `delete_blob_on_disk`), so no workflow can assert that a blob's *bytes* reach a second node. `apps/blobs/workflows/blobs.yml`'s only second-node step is `list_files`, which reads CRDT metadata and passes regardless. It also never asserts the announce succeeded. A companion merobox PR adds the step; once released, `blobs.yml` should gain a cross-node fetch assertion (needs a `MIN_MEROBOX` bump).
- **Not covered here:** a joining node did not receive context state in local testing (snapshot generation dropping every record as unrecognized, joiner stuck `Uninitialized`). That is unrelated to this fix and possibly specific to the tested join path — filed separately rather than conflated with this change.
- Reported by the CLI as a misleading `Blob query failed: Timeout` once retries exhaust, even when the real per-attempt error was the size cap. Worth a follow-up so the surfaced error is the last *meaningful* one.


chefsale:
please add e2e merobox tests for this if possible

chefsale (approved):

apps/blobs/workflows/blob-cross-node-sizes.yml
```diff
@@ -0,0 +1,322 @@
...
+  covered, assert the refusal with `expected_failure: true`, do not assert a
+  success.
+
+  FIXTURES: generated, not committed — run ./scripts/gen-blob-fixtures.sh first.
+  They are random bytes on purpose: blobs are content-addressed, so a predictable
+  fixture would already exist on the fetching node and the network path under
+  test would be skipped entirely.
+
+  NOT IN THE CI MATRIX YET. Needs merobox's `download_blob` step
+  (calimero-network/merobox#306, lands in 0.6.44) plus a MIN_MEROBOX bump. See
```
chefsale, line 26: implement this @frdomovic and than add to the matrix before landing

chefsale (changes requested):

apps/blobs/workflows/blob-cross-node-sizes.yml
frdomovic: Dug into this and it can't be implemented here — the blocker isn't the merobox version that comment mentions (that's stale; 0.6.44 is released and `download_blob` works). It's a Kademlia startup race, and the fix belongs in `crates/network`, not in a workflow file.

**What actually happens.** From run 30271797358, with both containers' logs:

```
node 1 starts                          13:54:50.476
node 2 starts                          13:54:52.045   (1.6s later)
node 1 kad dial → node 2               ×3, all FAIL before 13:55:00
dial retries after that                0
grep -c RoutingUpdated (both nodes)    0
```

Node 1 seeds kad at behaviour construction (`behaviour.rs:219`, `for (peer_id, addr) in bootstrap_peers`) and dials immediately — while node 2 is still coming up and not yet listening. Those dials fail. The only re-dial path is `discovery.rs::redial_direct`, which fires on `ConnectionClosed` and so requires having connected once; they never did. libp2p-kad 0.48 does run a periodic bootstrap, but its default interval is **5 minutes** — these workflows finish in ~13 seconds.

So node 2's routing table is empty when it queries:

```
GetRecord Err(NotFound { closest_peers: [] })
QueryStats { requests: 0, success: 0, failure: 0 }
```

`requests: 0` — it didn't fail to find the blob, it had nobody to ask.

**Why nothing else in the suite notices.** Metadata rides gossipsub, blob bytes ride kad. mDNS + gossipsub discover peers independently and don't care about the kad routing table, so `create_mesh` converges, CRDT state replicates, and `blobs.yml` stays green — while the byte path has never been reachable. That's the same gap this PR's description called out as "real and not closed".

**Why it's a separate PR.** Different subsystem (discovery, not authorization), different root cause, and it predates this branch — the announce/signing fix here is complete and independently proven: `blob-announce-membership` is in the matrix and green, asserting the membership bail is gone from the upload path. Holding that behind an unrelated startup race just delays it.

The missing piece is that core never feeds identify-discovered peers into kad, so kad only ever knows bootstrap peers. Opening a PR for that now; once it lands, `blob-cross-node-{transfer,sizes}.yml` go into the matrix with it — activated and green, which is what you asked for. I'd rather add them there, verified, than add them here and turn the suite red.

chefsale (approved):

PR #3319, opened 2026-07-27
refactor(cargo-mero): emit the app ABI in the tool, not in 60 build scripts

### LINKED ISSUE #3310: apps: ~60 byte-identical build.rs copies emit the ABI; extract a helper into calimero-wasm-abi
## Summary

Every WASM app carries a byte-identical ~38-line `build.rs` (read `src/lib.rs`, call `calimero_wasm_abi::emitter::emit_manifest`, write `res/abi.json` + `res/state-schema.json`). There are ~60 copies under `apps/`.

Any change to that logic - error handling, output paths, a `schema_version` bump, a switch to `emit_manifest_from_crate` for multi-file apps - has to be applied ~60 times identically, and a missed copy fails silently (the app just emits a stale or no ABI).

This is the same class of duplication that per-app `build.sh` scripts had before they were replaced by `cargo mero` (#3287): the shared step drifts because nothing owns it centrally.

## Impact

- Maintenance: one behavior change = ~60 identical edits, with silent drift if any is missed.
- Correctness risk: an app whose `build.rs` silently stops emitting produces a wasm with no embedded ABI, which the node then cannot introspect for the xcall gate or migration decisions. `scripts/check-embedded-abi.sh` now catches that in CI, but the guard exists because the emit step is not centrally owned.
- The source-parsing emitter is also cfg-blind (documented elsewhere), so a future fix there would need the same ~60-way edit.

## Steps to reproduce

1. `ls apps/*/build.rs apps/migrations/*/build.rs | wc -l` - about 60 files.
2. `diff apps/kv-store/build.rs apps/private_data/build.rs` - identical.

## Criteria for resolving

- [ ] `calimero-wasm-abi` exposes a build-script helper that does the whole step in one call (e.g. `calimero_wasm_abi::build::emit_abi_artifacts()` writing `res/abi.json` + `res/state-schema.json`, with the `cargo:rerun-if-changed` directive included).
- [ ] Every app's `build.rs` is reduced to that single call (all of them, so one pattern remains in the repo).
- [ ] A test or CI check asserts the helper emits both artifacts for a representative app.
- [ ] `cargo fmt --check`, `cargo clippy -- -A warnings`, and `cargo test` pass.

Raised by review on #3308; deliberately deferred there because converting only the three apps that PR touched would have left two competing patterns in the tree.

rtb-12:
## Summary

Closes #3310.

Every app carried a near-identical `build.rs` whose only job was to leave `res/abi.json` and `res/state-schema.json` on disk for `cargo mero build` to read back and embed.
Nothing owned that step, so it drifted into five variants across `apps/` plus a sixth in the `cargo mero new` template.
A missed copy fails silently: the artifacts are gitignored, so nothing shows in a diff, and the node's upgrade gates fail open when the ABI is absent (`crates/context/src/handlers/upgrade_group.rs`).

`cargo mero build` now emits both artifacts itself from the crate's own sources.
It already held the crate directory and cargo metadata, so the disk round-trip and the `"app build did not emit res/abi.json"` guard both go away.

- `crate_sources` resolves the crate's files by following `mod` declarations out from `src/lib.rs`, tracking each file's module directory and descending through inline `mod { .. }` bodies (a `mod x;` nested in one still loads `<dir>/outer/x.rs`), honoring each declaration's own `#[cfg]` against the resolved feature set as well as `#[cfg(test)]` and `#[path = "..."]`. It is deliberately not a filesystem glob: a file the module tree does not pull in - an optional module gated off, a `#[cfg(test)]` module, a leftover file - is never compiled into the wasm, so its types must not appear in the ABI the node reads. This is also strictly better than the old per-app `build.rs`, which read `src/lib.rs` alone and silently under-described any multi-file app.
- `resolved_features` takes the active feature set from the cargo resolve graph, standing in for the `CARGO_FEATURE_*` vars only a build script receives, and passes it to `emit_manifest_from_crate_with_features`. It errors rather than defaulting to an empty set, so a metadata miss cannot quietly emit an under-described ABI.
- `calimero_wasm_abi::emitter::item_cfg_active` becomes public so the module walk reuses the emitter's cfg evaluation instead of reimplementing it.
- Deletes `build.rs` and the `calimero-wasm-abi` / `serde_json` build-deps from all 60 apps, the 3 `cargo-mero` test fixtures, and the `new` scaffold (`templates/build.rs.tmpl`).
- Deletes `apps/nested-crdt-test`'s build-deps, which had no `build.rs` to serve. It stays outside `cargo mero build` for the reason already documented in `scripts/build-all-apps.sh`.
- `apps/xcall-example`'s manual method/event sort goes with its `build.rs`; the emitter already sorts (`crates/wasm-abi/src/emitter.rs`), and `canonicalize_abi` stays as the compat shim for an `abi.json` an older SDK emitted.
- `scripts/abi-contract-check.sh` builds its corpus through `cargo mero build --profiling` instead of a bare `cargo build`. The check previously relied on the extractor resolving the `res/abi.json` that each app's `build.rs` wrote during that compile; with no build script there is no such file, and every app reported "no ABI" until the run failed its own coverage gate. A missing wasm is now a hard failure rather than a skip, so a tool/script disagreement cannot quietly drop an app from the gate.

Scaffolded apps no longer pin `calimero-wasm-abi`, so the ABI format tracks the tool that builds them rather than a version pinned in every user's manifest.
This also drops one of the copies the "Bumping the SDK version" note in `tools/cargo-mero/README.md` had to keep in lockstep.

## Test plan

Behavior is meant to be identical, so the primary evidence is that the emitted artifacts are unchanged.

- [x] **Byte-identical artifacts.** Built `kv-store` via its `build.rs` to capture a baseline, then deleted the build script and rebuilt through the tool. Same sha256 for both files:
  - `abi.json` `afafae059ace04b9b89e6f8173500158267defe2a06957dc12ed2df70359da00`
  - `state-schema.json` `125985a9480e5d858133a848db05db2bf730a1396b096d6002d89f6fb6a01f92`
- [x] **The multi-file app too.** `abi_conformance` (`pub mod custom_types;`) is the only app that passed an explicit two-file list; its `abi.json` and `state-schema.json` are byte-identical via the module walk.
- [x] `./scripts/build-all-apps.sh` - exit 0
- [x] `./scripts/check-embedded-abi.sh` - `All 19 app wasm(s) carry the calimero_abi_v1 section.`
- [x] `./scripts/verify-abi.sh` - `ABI verify: OK` (extracted ABI still matches the checked-in `abi.expected.json`)
- [x] `./apps/state-schema-conformance/verify-state-schema.sh` - `Build-time and WASM-extracted schemas are identical`
- [x] `apps/xcall-example/res/abi.json` methods and events are still sorted with the manual sort removed, and `xcall_callable` survives on `pong` / `relay`
- [x] `cargo test -p cargo-mero` - 49 passed. New coverage: both artifacts written; `mod` declarations followed; nested `src/foo/bar.rs` followed; `mod` declarations inside inline `mod { .. }` bodies followed; unreferenced, feature-gated, `#[cfg(test)]`, and `#[cfg(test)]`-inline-block files excluded; missing `lib.rs` errors
- [x] `cargo test -p cargo-mero --test pipeline -- --ignored --test-threads 1` - 4/4 passed in 67s, down from 402s: a scaffolded app no longer fetches `calimero-wasm-abi` from crates.io, which was also an intermittent timeout
- [x] `cargo check --workspace` - exit 0
- [x] `cargo fmt --check`, `cargo clippy -p cargo-mero -- -A warnings`
- [x] ABI contract corpus locally: 5 representative apps built through the tool, extracted, `5 accepted, 0 failed`
- [ ] PR CI matrix (e2e-rust-apps, app-migration-e2e, sdk-e2e where triggered)

## Notes for review

`emit_abi` keeps the tolerant `if let Ok(..) = extract_state_schema()` the app build scripts had, so an app whose state root the emitter cannot express still gets a usable `abi.json`.
No in-repo app takes that path. Tightening it into a hard error is a behavior change and deliberately not bundled here.

Rebased onto master now that #3308 has landed - that PR is what makes `cargo mero build` the single build path this change relies on.


rtb-12:
Both review findings are addressed in 2e5b65f6d.

**`crate_sources` ignored mod-level cfg gating (over-inclusion).** Valid, and the more serious of the two, since those types land in the embedded section the upgrade gates read. Took option (a): `crate_sources` now follows `mod` declarations out from `src/lib.rs` instead of globbing the filesystem, evaluating each declaration's own `#[cfg]` against the resolved feature set, and honoring `#[path = "..."]`. `calimero_wasm_abi::emitter::item_cfg_active` is now public so the walk reuses the emitter's cfg evaluation rather than reimplementing it.

One thing the review predicted that the first attempt still got wrong: a file-backed `#[cfg(test)] mod harness;` was *still* collected, because `eval_cfg_meta` only reasons about features and returns `Unknown` for a bare `test` atom, which `item_cfg_active` keeps. A new test caught it. That exact form is now decided locally in `is_cfg_test`, matched narrowly so `#[cfg(not(test))]` stays included.

New coverage in `tools/cargo-mero/src/build.rs`:
- `crate_sources_follows_mod_declarations`
- `crate_sources_follows_nested_modules` (`src/foo.rs` declaring `mod bar;` resolves to `src/foo/bar.rs`)
- `crate_sources_ignores_unreferenced_and_gated_files` (asserts a stray file, a feature-gated module with the feature off, and a `#[cfg(test)]` module are all excluded, then that the gated module *is* included with the feature on)

The misleading test comment the review quoted is gone.

**`resolved_features` swallowed lookup failures.** Valid. It now returns `Result` and errors with the specific missing piece (package, resolve graph, or resolve node) instead of defaulting to an empty set, which would have quietly emitted an under-described ABI.

Re-verified after both changes, since they alter which files reach the emitter: `kv-store` and `abi_conformance` (`pub mod custom_types;`) still produce byte-identical `abi.json` and `state-schema.json` against their `build.rs` baselines, `./scripts/build-all-apps.sh` exits 0, and `check-embedded-abi.sh` reports all 19 app wasms carrying `calimero_abi_v1`. `cargo test -p cargo-mero` 47 passed, `cargo test -p calimero-wasm-abi` green.

rtb-12:
Both new findings addressed in 99ca49c88.

**`crate_sources` did not recurse into inline `mod { ... }` blocks.** Correct, and it is the same silent-under-description class the module walk was introduced to fix, so worth fixing properly rather than documenting. The traversal now tracks each file's module directory explicitly instead of deriving it from the file stem, and descends through inline bodies: a `mod x;` inside `mod outer { .. }` resolves against `<dir>/outer/`. Gating is applied at every level of the descent, so `#[cfg(test)] mod harness { pub mod fixtures; }` pulls in nothing. `#[path = "..."]` is honored on inline modules too, where it redirects the directory children resolve against.

Two tests added for exactly the gap the review named:
- `crate_sources_follows_mods_nested_in_inline_blocks` - `mod outer { mod inner; }` must collect `src/outer/inner.rs`
- `crate_sources_skips_mods_under_a_cfg_test_inline_block` - `#[cfg(test)] mod harness { pub mod fixtures; }` must collect only `lib.rs`

**Wasm path derived by the underscore convention.** The derivation itself is not a guess - it mirrors `build_one`, which computes `crate_name.replace('-', "_")` for both the cargo artifact and the `res/` copy, so the script and the tool cannot disagree unless one of them changes. Resolving it from `targets[].name` would describe cargo's artifact, not the path `cargo mero build` actually writes to, so it would be the less accurate source here.

The part of the finding that does hold is the consequence: a mismatch was being swallowed as a skip, which weakens the gate. That branch is now a hard failure (`FAILED+=("$pkg(wasm-path)")`), since `cargo mero build` already errors when it cannot find its own artifact - so reaching that point means the two disagree, which is a fault, not an app without a cdylib.

Re-verified after the traversal change: `kv-store` and `abi_conformance` still byte-identical to their `build.rs` baselines, `./scripts/build-all-apps.sh` exit 0, `check-embedded-abi.sh` all 19 wasms, `cargo test -p cargo-mero` 49 passed, fmt and clippy clean.

rtb-12:
Three findings addressed in 0adf077ef, and a correction on two threads I resolved too early.

**`cfg(test)` only detected as a bare predicate.** Correct. Rather than teaching `is_cfg_test` to walk `all(..)`/`any(..)`, the cause is one level down: `eval_cfg_meta` returned `Unknown` for a bare `test` atom, and `item_cfg_active` only drops an item on a definite `False`. The ABI describes the shipped wasm, which is never built with `--cfg test`, so `test` is `False` there rather than unknown.

That decides every compound form for free through the existing Kleene logic - `all(test, ..)` is False, `any(test, ..)` rests on its other predicates, `not(test)` is True - and it also covers the inline `#[cfg(test)] mod tests { .. }` blocks every app has, which were previously kept. `is_cfg_test` is deleted, so cfg semantics have a single owner in the emitter instead of a partial copy in the tool.

Verified this changes no emitted ABI: `kv-store` and `abi_conformance` stay byte-identical to their `build.rs` baselines, and the `calimero-wasm-abi` suite is green. New test `crate_sources_decides_cfg_test_in_compound_predicates` covers `all(test, feature)`, `any(test, feature)` and `not(test)` in one case.

**On the two state-schema threads (`extract_state_schema` failure swallowed with no diagnostic, and `emit_abi` tolerating a missing state schema):** these were marked resolved after the bot posted "No longer detected", but the code they flagged was unchanged - the retraction was a false negative and I should have checked the source instead of taking it. The finding was real. `emit_abi` now warns on stderr, naming the crate and the reason, and says explicitly that the node cannot check upgrades against that app. The build still succeeds, since an inexpressible state root should not block a usable `abi.json`, but it is no longer something you discover from an absent file.

`cargo fmt`, `clippy`, `cargo test -p cargo-mero` (50 passed) and `cargo test -p calimero-wasm-abi` all clean.

rtb-12:
Both new findings fixed in 0796cf95c. Both are consequences of this PR making `item_cfg_active` gate whole files rather than items inside one, so they are fair to land here.

**Target cfgs resolved Unknown-keeps-item.** Correct, and the escalation the review describes is exactly right: Unknown was a narrow inaccuracy while the function only filtered items inside an already-included file, and became a file-inclusion decision here.

Took the suggested approach. `wasm32_target_value` maps each `target_*` key to the value it takes on `wasm32-unknown-unknown` (the only target `cargo mero build` produces, `build.rs:19`), so those predicates get a definite verdict the same way `test` does. `unix`/`windows` are False, `wasm` is True. An unrecognized key deliberately stays Unknown - guessing False there would drop items that do belong in the ABI, which is the opposite failure and a worse one.

New test `crate_sources_skips_modules_gated_to_another_target` covers `not(target_arch = "wasm32")`, `target_os = "linux"`, `unix`, and the positive `target_arch = "wasm32"` case in one assertion.

Worth noting I nearly shipped a real bug here: the first version of the new `Meta::NameValue` arm was placed before the existing `feature` arm and shadowed it, which would have silently disabled all feature gating. Caught because `crate_sources_ignores_unreferenced_and_gated_files` exercises features on and off. The arm now carries an explicit `!nv.path.is_ident("feature")` guard so it is order-independent.

**Stale res/state-schema.json surviving a failed extraction.** Valid, and a good catch given `bundle` ships `res/` as-is. It is removed on the failure path now, which is what the warning already implied.

The regression test needed care: `nested-crdt-test` fails at manifest emission, before `extract_state_schema` is reached, so it does not exercise this branch at all. `emit_abi_removes_a_stale_state_schema_when_extraction_fails` uses a source that emits a manifest but has no `#[app::state]`, so extraction is what fails. Verified it is a real gate - with the removal disabled it fails with "a stale state-schema.json must not survive a failed extraction", and passes with it restored.

Verified from a wiped `res/`: `build-all-apps.sh` exit 0, `check-embedded-abi.sh` 17/17, `verify-abi.sh` OK against the golden file, `verify-state-schema.sh` all passed, `cargo test -p cargo-mero` 52 passed, `calimero-wasm-abi` green, fmt and clippy clean. `kv-store` and `abi_conformance` remain byte-identical to their `build.rs` baselines.

chefsale (approved):

PR #3326, opened 2026-07-28
fix(network): seed Kademlia from identify-discovered peers

frdomovic:
Follow-up to the review thread on #3317 (calimero-network/core#3317 (comment)). That PR fixes blob **authorization**; this fixes the **discovery** problem sitting upstream of it, which is why cross-node blob transfer still cannot be exercised end to end.

## The bug

`kad` only ever learned the peers in `discovery.bootstrap.nodes`, seeded once at behaviour construction:

```rust
// crates/network/src/behaviour.rs
for (peer_id, addr) in bootstrap_peers {
    let _ = kad.add_address(&peer_id, addr);
}
```

If those initial dials fail, the routing table stays empty. Nothing re-seeds it:

- the only re-dial path, `discovery::redial_direct`, fires on `ConnectionClosed` — it requires having connected once, which never happened
- libp2p-kad 0.48 does run a periodic bootstrap, but its default interval is **5 minutes**
- kad's automatic bootstrap triggers on a peer *entering* the routing table, so it can't fire either

Meanwhile mDNS + gossipsub discover peers independently and don't consult kad at all. So a fleet gossips and replicates CRDT state perfectly happily while every DHT query returns:

```
GetRecord Err(NotFound { closest_peers: [] })
QueryStats { requests: 0, success: 0, failure: 0 }
```

`requests: 0` — not "couldn't find it", but "had nobody to ask".

## Reproduction

Two nodes starting 1.6s apart is enough to lose the race. From a merobox docker run (core run 30271797358), both containers' logs:

```
node 1 starts                        13:54:50.476
node 2 starts                        13:54:52.045
node 1 kad dial -> node 2            x3, all FAIL before 13:55:00
dial retries afterwards              0
grep -c RoutingUpdated, both nodes   0
```

Node 1 dials the sibling immediately, while node 2 is still coming up and not yet listening. Then nothing for the rest of the run.

Concretely, this is why **blob bytes** cannot cross while **blob metadata** does: metadata rides gossipsub, bytes ride the kad provider lookup. `apps/blobs/workflows/blobs.yml` has always been green because its only second-node assertion is `list_files`.

## The fix

The identify handler already mirrors a peer's advertised `listen_addrs` into our local address book, behind a filter that drops relayed (`/p2p-circuit/`), echoed-back self-observed, and non-dialable (loopback / unspecified / link-local) addresses. Feed those same filtered addresses to `kad.add_address` — the standard libp2p pattern.

The routing table then populates from any connection however it was discovered, and kad's automatic bootstrap throttle fires on the new peer.

**No new trust.** These are exactly the addresses already accepted for direct dials; kad never sees an entry the address book would have rejected. Peer-controlled input stays behind the same filter.

## Verification

- `cargo check -p calimero-network`, `cargo fmt --check`, `cargo clippy -p calimero-network --all-targets -- -D warnings` — clean
- `cargo test -p calimero-network` — 113 tests pass

Not yet proven end to end: the e2e that would demonstrate it (`blob-cross-node-transfer.yml` / `blob-cross-node-sizes.yml`, byte-exact cross-node transfer at 1/8/32/100 MiB with a negative control before each fetch) lives on #3317's branch. **Once #3317 lands, those two go into the e2e matrix in a follow-up and should turn green on the back of this change** — that is the check @chefsale asked for, and the point at which this fix is confirmed rather than reasoned.

Reviewers: worth a sanity check on whether populating kad from identify has any downside for large deployments, where the routing table is meant to be sparse-by-distance. libp2p's own bucket logic bounds it, so I do not think it does, but it is the one judgement call in here.


frdomovic:
## The e2e: `apps/scaffolding-e2e/workflows/kad-routing-from-identify.yml`

**In the matrix, and it can pass on this branch alone** — no dependency on #3317.

It boots two nodes, connects them with `create_mesh` (which is what makes identify fire — that's the precondition under test, not incidental setup), then asserts `RoutingUpdated` is present in **both** nodes' logs. libp2p-kad emits that event exactly when a peer enters the routing table, and the kad handler dumps it verbatim via `debug!("{}: {:?}", "kad", event)`.

The number that makes this a real regression guard: on the failing build, `grep -c RoutingUpdated` over both containers was **0 for an entire run** (core run 30271797358). So this assertion fails on `master` today and passes with this change — it is not a test that would have been green either way.

Design choices worth knowing:

- **Per node, not across the union.** One node populating its table while the other stays empty is still a broken DHT: a lookup from the empty side reaches nobody. merobox's `assert_log_present` aggregates across the named nodes, so asserting both in one step would pass on a half-broken fleet.
- **No DHT query of its own.** The routing table is the thing that was empty; asserting on it directly keeps the test honest about what it proves and fast, with no provider-record propagation to wait out.
- **`log_level: debug` pinned in the workflow.** `RoutingUpdated` is a DEBUG-level dump. That is merobox's default, but pinning it means a change to that default cannot silently turn the assertion into a no-op.
- **The 10s settle is a margin, not a workaround.** The fix populates the table on the identify exchange, so seconds suffice. If this needed minutes it would mean the table was only filling via libp2p's 5-minute periodic bootstrap — i.e. the fix not working — so the short wait is itself part of the signal.

## What this does and does not prove

Proves: a peer discovered over mDNS now reaches the kad routing table, which is the specific defect.

Does not prove: byte-exact cross-node blob transfer. That needs `blob-cross-node-{transfer,sizes}.yml`, which live on #3317's branch; they go into the matrix in a follow-up once it lands, and are the end-to-end confirmation. I would rather land this with the narrow assertion that genuinely covers the change than claim the broader one before it can run.

## Also worth a second opinion

`cargo test -p calimero-network` (113 tests), `fmt --check` and `clippy --all-targets -D warnings` are clean, but none of those cover this path — the identify handler needs a live swarm, so the e2e above is the only real coverage. The one judgement call in the diff is whether feeding kad from identify has a downside for large deployments, where the routing table is meant to be sparse-by-distance. libp2p's bucket logic bounds table size, so I do not believe it does, but it is worth a networking-side review rather than taking my word for it.

chefsale (approved):

PR #3341, opened 2026-07-30
feat(cargo-mero): add --features to build and bundle

rtb-12:
## Summary

`cargo mero` could not build an app with non-default cargo features, so an app that selects a schema variant that way (a v2 migration-target bundle, say) had to stay on a hand-rolled `build-bundle.sh`.

The ABI emitter was already feature-aware: `resolved_features()` reads the activated set out of `cargo_metadata`'s resolve graph, and `crate_sources` / `emit_abi` gate items on it via `item_cfg_active`. The gap was narrower than it looks: nothing ever passed `--features`, so cargo resolved defaults, and the compile and the ABI were *consistently* wrong. Nothing failed loudly.

Changes:

- `--features <FEATURES>` and `--no-default-features` on `build` and `bundle`, spelled the way cargo spells them (comma or space separated, repeatable). They live in one `workspace::FeatureArgs` that both commands flatten in, so the two cannot drift apart.
- Forwarded to `cargo build`, so the wasm is actually compiled with them.
- Forwarded to `cargo metadata`, so `resolved_features()` reports the same set the build used and the emitted ABI describes the same schema as the bytecode.
- Services now compile in a single `cargo build` invocation. Cargo rejects a plain `--features` naming a feature the one selected package lacks, but accepts it when several packages are selected, and a multi-service workspace routinely gates only one service's schema.
- `bundle` passes its features into `build::run_all`, so multi-service bundles honour the flag.

`--all-features` was left out: nothing needs it, and it is trivial to add later.

### Why the two must be resolved together

This is the whole point of the change rather than an implementation detail. If the compile enables `schema_v2` but the ABI is emitted from the default resolve graph, the bundle ships bytecode and an ABI that disagree. That is not a build error. The node resolves migration plans from the embedded `calimero_abi_v1` section alone, so the failure surfaces as a silently wrong migration at upgrade time.

### Known limitation, documented not changed

The emitter cfg-filters **top-level items only** (`file.items.retain(item_cfg_active)`). A `#[cfg]` on a struct field, an enum variant, or a method inside an `#[app::logic]` block is not applied to the ABI. That predates this change and is unaffected by it, but it now matters to anyone reaching for `--features`, so it is called out in the README, the docs site page, and `tools/AGENTS.md`. Express a schema variant as whole gated items.

## Test plan

**Unit** - `build::tests::emit_abi_follows_the_active_feature_set`: a `#[cfg(feature)]`-gated `#[app::state]` reaches the emitted state schema only when its feature is active (`State` vs `StateV2`).

**CLI** - `feature_flags_parse_on_both_build_and_bundle`: `--features a,b`, `--features "a b"`, repeated `--features`, and `--no-default-features`, on both subcommands.

**End to end** - `features_select_the_same_schema_in_the_wasm_and_the_embedded_abi` (in `tests/pipeline.rs`, which CI runs via `--ignored`). The `multi-app` fixture's `svc-a` gained a `schema_v2` variant; `svc-b` deliberately does not declare the feature. The test builds both ways and asserts against the wasm itself, not the emitter:

- reads the `calimero_abi_v1` custom section back off the built wasm and asserts `state_root` is `SvcAV2` with the feature and `SvcA` without it
- asserts the wasm's **export section** contains `revision` (a v2-only method) with the feature and not without, which is what catches the ABI moving while the bytecode does not
- asserts `svc-b` still builds and embeds, proving a feature only one service declares does not break the others

Both halves were verified to be real gates by breaking each in turn:

| break | result |
| --- | --- |
| drop `features.apply_to(&mut cmd)` in `cargo_build` (ABI moves, wasm does not) | fails at the `revision` export assertion |
| drop `CargoOpt::SomeFeatures` in `metadata_for` (wasm moves, ABI does not) | fails `left: "SvcA"` vs `right: "SvcAV2"` |

**Against the real motivating case** - built a multi-service app's docs service both ways with the new flag and diffed the embedded section read back out of each wasm:

```
+  "migrations": [ { "fromVersion": 1, "method": "migrate_v1_to_v2" } ],
-  "state_version": 1,
+  "state_version": 2,
+  { "name": "default_sort_order", "type": { "crdt_type": "lww_register", ... } }
```

The bytecode differs too (583038 vs 588229 bytes), so the wasm and its ABI moved together.

**Gates**

```
cargo fmt --check
cargo test -p cargo-mero                 # 55 passed
cargo check --workspace --all-targets
```

### DIFF
--- a/docs/src/content/docs/build/cargo-mero.mdx
+++ b/docs/src/content/docs/build/cargo-mero.mdx
@@ -137,6 +137,34 @@ meroctl app install --path dist/<package>.mpk ...   # 5. install on a node (mero
 
 </Steps>
 
+## Cargo features
+
+`build` and `bundle` take cargo's feature flags, spelled the way cargo spells them - comma or space separated, repeatable:
+
+```bash
+cargo mero build --features schema_v2
+cargo mero bundle --dev --features "schema_v2 telemetry" --no-default-features
+```
+
+They reach `cargo build` and the `cargo metadata` call the ABI is emitted against, so the embedded `calimero_abi_v1` section always describes the schema the bytecode was compiled with.
+That pairing is the point: a wasm and an ABI that disagree do not fail the build, they produce a wrong migration plan at upgrade time, because the node resolves upgrades from the embedded section alone.
+
+Features gate **top-level items**, which is how one crate carries two schema versions:
+
+```rust
+#[cfg(not(feature = "schema_v2"))]
+#[app::state(version = 1)]
+pub struct State { /* ... */ }
+
+#[cfg(feature = "schema_v2")]
+#[app::state(version = 2)]
+pub struct StateV2 { /* ... */ }
+```
+
+A `#[cfg]` on a struct field, an enum variant, or a method inside an `#[app::logic]` block is *not* applied to the ABI, so express a variant as whole gated items rather than gated members.
+
+In a multi-service workspace all services compile in one `cargo build`, so a feature only one service declares is fine: it applies to that service and is ignored by the rest.
+
 ## Metadata reference
 
 Bundle metadata comes from a `[package.metadata.calimero]` table in the app's `Cargo.toml`.
--- a/tools/AGENTS.md
+++ b/tools/AGENTS.md
@@ -77,6 +77,8 @@ cargo-mero/
 - The `tests/pipeline.rs` end-to-end tests are all `#[ignore]`d because they scaffold and compile fresh crates (slow, and `new_build_test_bundle_ladder` needs network for the git SDK deps). Run them with `cargo test -p cargo-mero -- --ignored`.
 - Bumping the scaffolded SDK version touches two files in lockstep: `DEFAULT_SDK_VERSION` in `src/main.rs` and the version assertions in `src/new.rs` tests.
 - `services` is a workspace-only key: it is rejected under a `[package.metadata.calimero]` table, only accepted under `[workspace.metadata.calimero]`.
+- `--features` is resolved once and used for both the compile and the ABI emit (`workspace::FeatureArgs` feeds `cargo build` and `cargo metadata`). Keep them together: a wasm and an embedded ABI that disagree fail silently, as a wrong migration plan.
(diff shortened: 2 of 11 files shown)


chefsale (approved):

rtb-12:
Thanks - I checked this one carefully because if it were right it would be the exact failure mode the PR exists to prevent. It does not reproduce: the premise about the cargo_metadata API is not correct for the version this crate pins.

**The claim:** "`MetadataCommand::features()` takes a single `CargoOpt` value ... and each call to `.features(...)` overwrites the previously stored option rather than merging with it."

**cargo_metadata 0.20.0 keeps three independent fields**, not one:

```rust
/// Collections of `CargoOpt::SomeFeatures(..)`
features: Vec<String>,
/// Latched `CargoOpt::AllFeatures`
all_features: bool,
/// Latched `CargoOpt::NoDefaultFeatures`
no_default_features: bool,
```

and `features()` extends or latches, never replaces:

```rust
pub fn features(&mut self, features: CargoOpt) -> &mut MetadataCommand {
    match features {
        CargoOpt::SomeFeatures(features) => self.features.extend(features),
        CargoOpt::NoDefaultFeatures => {
            assert!(!self.no_default_features, "Do not supply CargoOpt::NoDefaultFeatures more than once!");
            self.no_default_features = true;
        }
        ...
```

The `assert!` on a repeated `NoDefaultFeatures` is itself evidence the flag is latched rather than overwritten. All three are then emitted as separate arguments when the command is built.

**Verified at runtime, not just by reading.** The suggested regression test was worth adding regardless, so it is now in `422a9d3` as `workspace::tests::metadata_applies_features_and_no_default_features_together`. It builds a throwaway crate with `default = ["baseline"]`, `baseline` and `schema_v2`, calls `metadata_for` with `--features schema_v2` **and** `no_default_features: true`, and asserts both directions of the resolve graph:

- `schema_v2` is present, so `--no-default-features` did not drop `--features`
- neither `default` nor `baseline` is present, so `--features` did not drop `--no-default-features`

It passes. And it is a real gate, not just observed green: deleting the `CargoOpt::SomeFeatures` call makes it fail with

```
--no-default-features must not drop --features, got {}
```

which is precisely the divergence described. So the scenario is now pinned by a test that would catch it if a future cargo_metadata ever did start overwriting.

No production change needed; the finding is a false positive against 0.20.0.

Issue #3347, opened 2026-07-31
join-via-inheritance fails permanently when the only key holder hits a transport error


chefsale:
## Summary

`join_subgroup_inheritance`'s key fetch (`crates/node/src/sync/manager/namespace_sync.rs`, the peer loop ending at the `bail!` around line 974) makes **one pass** over mesh peers and then fails the whole join. It correctly moves on when a peer answers key-less or rejects ("send failed, trying next peer"), but when the pass ends with the only key-holding peer having hit a *transport* error, the call fails permanently instead of retrying.

A key-less answer is authoritative — that peer genuinely does not hold the key. A transport error is not an answer at all. Treating both as terminal means one dropped direct-stream to the sole key holder is indistinguishable from "nobody has the key".

## Impact

Intermittent CI failures in any scenario where exactly one peer holds the subgroup key at the moment a joiner asks — which is the normal shape right after a subgroup is created, before other members have inherited.

Observed on `group-remove-from-root-revokes-inherited`:

```
join_subgroup_inheritance failed on revoke-inherit-node-2
send failed, trying next peer
no mesh peer held the subgroup key for … [2 peer(s): 1 key-less, 1 transport error(s)]
❌ Step 'Node-2 self-joins Open subgroup via inheritance' failed
```

node-3 was legitimately key-less (it had not inherited yet) and node-1 — the creator, and the only key holder — transport-failed. Re-running the identical commit passed, confirming the transience.

Beyond CI, the same shape hits a real joiner whose single key-holding peer is briefly unreachable: the join returns an error that reads as "no peer has the key", when the accurate answer is "ask again".

## Steps to reproduce

1. Create a subgroup and a context in it from node-1, so node-1 is the only holder of the subgroup key.
2. Have node-2 call join-via-inheritance while at least one other key-less peer is in the mesh.
3. Drop or delay the direct stream between node-2 and node-1 (or run under enough CI load for it to happen by itself).
4. Actual: the join fails with `no mesh peer held the subgroup key … 1 transport error(s)`. Expected: a second pass finds the key, since the key holder is reachable moments later.

The per-peer tally already in the error message is what makes this diagnosable — it should be kept.

## Criteria for resolving

- [ ] A pass whose failures are all transport-level is retried (bounded rounds, short backoff) rather than failing the join
- [ ] A pass where every peer authoritatively answers key-less still fails fast — no added latency for the genuine "nobody has it" case
- [ ] The per-peer tally is preserved in the final error
[...]

PR #3348, opened 2026-07-31
test(ci): fail a scenario that reads another node's state without waiting for it

chefsale:
## Summary

Three merobox scenarios failed this way in one afternoon. The cost was never the retry — it was that **each failure named the wrong thing**:

| Symptom | What it actually was |
| --- | --- |
| `output: None` where a schema version was expected | node 2's note had not reached node 1 |
| `Invalid URL: invalid length 9` | the literal string `"undefined"` reaching a base58 parser |
| `no mesh peer held the subgroup key` | the key holder was briefly unreachable ([#3347](https://github.com/calimero-network/core/issues/3347)) |

None of those says "timing", so each bought an investigation into whichever feature the PR happened to touch. Two were on a PR that had not gone near the subsystem that failed.

## The rules

Both taken from an observed failure rather than invented:

- **R1** — a `wait_for_sync` that omits a node which mutated since the last barrier. The `account-device-revoke-lockout` bug a reviewer found: the settle step listed nodes 1 and 2, and the next step had the freshly paired node 3 write with a key it could only have received by syncing.
- **R2** — a step claiming cross-node convergence with a mutation on *another* node since the last barrier. `32-authored-migrate-ux`: three reads asserted "converged" with nothing between them and the writes, and the one reading the other node's last write returned null.

A bare `wait: seconds` deliberately does **not** count as a barrier. Every failure here had one available and it did not help — a fixed sleep asserts nothing about arrival, it just converts a deterministic failure into a load-dependent one.

## Keeping it credible

Most of the work was the false-positive rate. The first draft was useless at **82 findings over 121 scenarios**:

- `json_assert` carries no `node`, so there is no way to know which replica it reads. It compares values earlier `call` steps captured, and those are where the barrier question belongs. Blaming the assert produced one finding per assertion, pointing at the wrong step.
- Read-only `call`s were marking their node dirty, so the *next* read was flagged as racing a write that never happened. Methods are now classified by name, and an **unrecognised method counts as a read** — this check earns its place only if its findings are credible, so it stays silent when it cannot tell.
- A node's own writes are locally visible; only another node's need syncing.

That leaves **9**, all pre-existing.

## Enforcement

Ratcheted against `scripts/e2e-convergence-baseline.txt` — same shape as the endpoint-coverage gate. Only a **new** finding fails. A hard gate would have failed on its first run, and a check that fails on day one gets disabled rather than fixed. A stale baseline entry is a notice, never an error, so fixing a scenario never needs to touch the baseline to go green.

Runs before anything boots docker, so it fails in seconds rather than minutes.

## Proof it works

```
# strip the barrier added to 32-authored-migrate-ux
$ python3 scripts/lint-e2e-convergence.py
3 NEW finding(s):
  ...32-authored-migrate-ux.yml: R2 'Read node 2's note schema_version from node 1 (converged)'
     asserts cross-node convergence, but ['app-migration-ux-2'] mutated after the last barrier
EXIT=1

# restored
$ python3 scripts/lint-e2e-convergence.py
9 baselined (accepted) finding(s) — burndown backlog
no new convergence races found
EXIT=0
```

It names the exact step whose output was null in CI. It stays silent on `group-remove-from-root-revokes-inherited`, whose failure was a core retry gap rather than a missing barrier — so it is not just flagging everything that ever went red.

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -55,6 +55,29 @@ jobs:
     name: Build Image
     uses: ./.github/workflows/build-merod-image.yml
 
+  # ── Scenario convergence lint ──
+  # Static, no nodes, seconds not minutes — so it runs ahead of everything that
+  # boots docker and fails fast on the cheapest class of e2e bug there is.
+  #
+  # A scenario that reads another node's state without a barrier establishing the
+  # state arrived is not merely flaky; it fails with a symptom that names the wrong
+  # thing. Three did in one afternoon: a null where a schema version was expected,
+  # the literal string "undefined" reaching a base58 parser, and a key-less join.
+  # Each cost an investigation into whichever feature the PR happened to touch
+  # before the timing was even suspected.
+  #
+  # Ratcheted against scripts/e2e-convergence-baseline.txt: pre-existing findings
+  # are an accepted backlog and only a NEW one fails.
+  convergence-lint:
+    name: Scenario convergence lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install PyYAML
+        run: pip install --quiet pyyaml
+      - name: Lint merobox scenarios for unbarriered convergence reads
+        run: python3 scripts/lint-e2e-convergence.py
+
   # ── Auth seam: client-token e2e over real HTTP ──
   # Proves a client token minted the way the auth frontend mints it can call
   # the routes the SDK uses (the 0.11.0-rc.9 login-loop outage shape).
--- a/crates/storage/src/interface.rs
+++ b/crates/storage/src/interface.rs
@@ -553,39 +553,45 @@ impl<S: StorageAdaptor> Interface<S> {
         }
     }
 
-    /// Resolve which writer in `writers` produced `sig_data`'s signature over
-    /// `payload`, returning that writer's key on success.
+    /// Resolve which writer produced `sig_data`'s signature over `payload`,
+    /// returning that writer's key on success.
     ///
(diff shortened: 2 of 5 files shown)


chefsale:
**Correction to the merge record: this squash contained more than the PR described.**

The branch was created with `git checkout -b` while standing on `feat/account-keyed-writers` (#3344) instead of from `origin/master`, so it inherited that branch's commit as its parent and the squash merged both:

```
 .github/workflows/e2e-rust-apps.yml   |  23 ++     ← the linter (described)
 scripts/lint-e2e-convergence.py       | 239 ++++   ← the linter (described)
 scripts/e2e-convergence-baseline.txt  |  20 ++     ← the linter (described)
 crates/storage/src/interface.rs       |  56 +--    ← NOT described: resolve_signer
 crates/storage/src/tests/interface.rs |  84 ++     ← NOT described: its test
```

The undisclosed part is stage 1 of #3344: `Interface::resolve_signer` now requires a signed write to name its author instead of falling back to a linear scan over every writer's key. That is a real change to signature verification on the storage apply path, and it was reviewable in the diff but not mentioned in the title or body.

Recording it here rather than quietly, for two reasons: anyone bisecting a storage-verification regression to this commit would otherwise be looking at a commit that claims to be a CI lint, and the merge history should say what it actually did.

**Nothing is broken.** The change was going to land anyway via #3344, it is independently a security fix (a signature verifying under nobody previously cost one `ed25519_verify` *per writer*, on a path any peer can drive with a malformed delta), its regression test came with it, and CI was green on the combination. `cargo test -p calimero-storage` passes on master.

**Consequence for #3344:** stage 1 is now upstream, so a rebase correctly drops it and that PR's remaining content is stages 2–3 only. Retitled accordingly.

PR #3374, opened 2026-08-04
refactor(bundle): derive the manifest from the canonical type

rtb-12:
## What this adds

**`cargo mero bundle` can now write every field the node and the registry actually read.** Before this, the tool emitted a fixed 12-key manifest and nothing else was reachable.

New keys in `[package.metadata.calimero]`:

```toml
[package.metadata.calimero]
package  = "com.acme.my-app"
icon     = "../app/public/icons/icon-512x512.png"  # a path; the tool encodes it
slug     = "com.acme.my-app"                       # -> handlers.slug, defaults to package
license  = "MIT"
tags     = ["social", "productivity"]
github   = "https://github.com/acme/my-app"
docs     = "https://docs.acme.com/my-app"
author   = "Acme"
```

`author`, `license`, `github` and `docs` are rendered on an app's registry page today and were previously unwritable, so every tool-built app published with those slots blank. `handlers.slug` shipped in #3332 with no producer anywhere; this is the first thing that can emit it. `icon` is taken as a **path** and encoded to a `data:image/png;base64,...` URI, so nobody has to paste a 17 KB data URI into a TOML file.

New flags and commands:

| | |
|---|---|
| `--no-icon` | ship without an icon (fine when `frontend` is set: the desktop finds the app's own PWA icon there) |
| `--no-abi` | bundle without an ABI, for fixtures that test the node's refusal path |
| `--bump patch\|minor\|major` | ask the registry for the highest published version and increment it |
| `--print-output-path` | print the path actually written, so CI stops reconstructing filenames |
| `--no-logo` | suppress the terminal icon preview |
| `publish <mpk>` | push a signed bundle to the registry |
| `services` in a package table | a single crate can now back several named services |

`bundle` now prints a summary, including the derived **ApplicationId** (`hash(package, signerId)`). Nothing printed it before, and it is the value that decides whether a release upgrades an app or installs a new one:

```
✓ bundle  dist/com.acme.my-app-1.2.3.mpk  (1.2 MiB)

  Name           My App
  Package        com.acme.my-app
  Version        1.2.3
  Application    7E6eADgrrxr98AUpFUaisPHAvwZofq41xqDD9357ctU9
  Signer         did:key:z6MknF3p5L5FDHJQ7FREUapuX4Wmp4MtF6WrHYaXS2B3eZQd
```

Both are printed in full, since an application id gets pasted into commands and compared between releases.

The Calimero mark is bundled for `icon = "default"`. It is never applied on its own, only when named: the desktop already falls back to the app's own icon at its `frontend` URL, and a generic mark would outrank it.

## Why the tool could not write those fields

It hand-wrote a JSON shape mirroring the node's `BundleManifest`, with no dependency between them and nothing forcing the two to agree. Apps that needed a missing field hand-rolled a manifest in shell instead, computing SHA-256s with `shasum`, which is how a published bundle ends up with a null artifact hash the node refuses.

`crates/node/primitives/src/bundle/` moves to a new crate, `calimero-bundle`, re-exported from `calimero-node-primitives` so no call site changes. The tool could not depend on `node-primitives` directly: that pulls actix, libp2p, reqwest, tokio and RocksDB into a CLI installed with `cargo install`. The extracted module needs only serde, base64, ed25519-dalek and sha2. `mero-sign` uses it too, deleting its own copy of the canonicalization a signature commits to.

`manifest::render` now builds `BundleManifest` with every field named, no `..`, and no `Default` on the type. **Adding a field to the manifest is a compile error in the tool until someone decides what to write there.** That is the property the whole change exists for.

## What it removes

The change is a net deletion.

- `BundleManifest.migrations`: no producer, no consumer. Its only mentions were its declaration, the hash-check loop and a fixture. Real migrations are ABI edges from `#[app::state(version = N, migration = ...)]`, read by `upgrade_group`.
- `canonicalize_abi`: unreachable. `ManifestBuilder::finish` sorts, `abi::extract` calls `validate_manifest` (which rejects unsorted input) before `build_one` runs, and `run_embed` validates again. It landed in the same commit as the check that made it dead. `cargo mero abi embed` bypasses `build.rs` entirely, so no path could deliver an unsorted manifest to it.
- The `--verify` flag on `abi extract`: bound to `_verify` and ignored, while the help advertised it.
- Two dead `BundleManifest`/`BundleArtifact` lookalikes in `server/primitives/admin`, never constructed and reachable from no route.
- `apps/migrations/bundle-wasm.sh` and `apps/scaffolding-e2e/build-multi-service-bundle.sh`, both replaced by tool capabilities. **No script in this repo writes a manifest by hand any more.**
- Forty `-o` arguments in `workflows/app-migration/build-wasms.sh`, made unnecessary by versioned output.

## Breaking changes

1. **`bundle` requires an icon decision.** `icon = "<path>"`, `icon = "default"`, or `--no-icon`. Existing builds fail until one is chosen.
2. **Default output is `dist/<package>-<appVersion>.mpk`**, previously `dist/<package>.mpk`. `--output` still overrides verbatim.
3. **`cargo mero abi extract --verify` is removed.**
4. **`BundleManifest.migrations` is removed.**
5. **The manifest now always declares `handlers.slug`**, defaulting to the package id. An app relying on a name-derived fallback should set `slug` explicitly to keep its existing links.

## Test plan

```
cargo test -p cargo-mero -p calimero-bundle -p mero-sign -p calimero-server-primitives -p calimero-node-primitives
cargo fmt --check
cargo clippy -- -A warnings
bash workflows/app-migration/build-wasms.sh
```

`workflows/app-migration/build-wasms.sh` now bundles every migration fixture through `cargo mero bundle`, including the ABI-less one it previously built by hand, so the app-migration scenarios exercise this path end to end.

`tools/cargo-mero/tests/pipeline.rs::bundle_produces_signed_mpk` is no longer `#[ignore]`d. It builds a real fixture and asserts the produced `.mpk` carries `handlers.slug`, a `data:image/png;base64,` icon, and no nulls anywhere in the manifest. It runs in `cargo-mero-e2e`, which now uses `--include-ignored` so the slower ladder tests still run there too; the workspace job skips it so it does not build a wasm fixture into the shared `CARGO_TARGET_DIR`.

`registry.rs` tests drive `next_version` and `publish` against a one-shot `TcpListener`, asserting the real request: `POST /api/v2/bundles/push` with `Authorization: Bearer`.

## Proof the fix works

**Manifest fields had no producer.**

Reproduction: bundle any app and read the manifest.

Before: `handlers`, `metadata.icon`, `license`, `tags`, `links.github` and `links.docs` are absent, because `ManifestJson` had no such fields.

After: `tools/cargo-mero/tests/manifest_roundtrip.rs::every_field_survives_a_round_trip_through_the_node_type` renders a manifest and deserializes it with `calimero_bundle::BundleManifest`, asserting each field arrives. `bundle_produces_signed_mpk` asserts the same against a real `.mpk`.

**Absent optional fields serialized as `null`.**

Reproduction: bundle an app with no `author` set.

Before: the manifest carried `"author": null`. The registry permits `author` to be absent but rejects it as null (`typeof null !== 'string'`), so the bundle was refused at publish.

After: `omitted_optional_fields_serialize_without_nulls` walks the serialized JSON recursively and fails on a null at any depth, so any future optional field that forgets `skip_serializing_if` fails the suite.

**A relative `icon` path resolved against the wrong directory.**

Reproduction: a workspace member with `icon = "assets/icon.png"`, bundled with `cd apps/<member> && cargo mero bundle --dev`.

Before: `base_dir` returns the workspace root when `--manifest-path` is omitted, so the root `assets/icon.png` was embedded, or the error named a path the user never wrote.

After: a package table resolves against the resolved package's own directory. `package_table_dir_resolves_from_the_resolved_package_not_the_passed_in_dir` covers it, and was confirmed to fail when the fix is reverted.

**`--bump` could sign a version below the published one.**

Reproduction: a registry response that wraps its listing, such as `{"bundles": [...]}`, or one whose entries carry no readable `appVersion`.

Before: both fell through to `Ok("0.1.0")`. A package at `3.4.5` would build and sign `0.1.0`.

After: a non-array top level errors, and a non-empty listing with nothing readable errors naming the strings it could not parse. A genuinely empty array still yields `0.1.0`, which is the distinction that matters. A mixed listing still ignores bad entries and uses the good ones.

## Wire contract (SDK gate)

No live DTO or route changed. The two structs removed from `server/primitives/admin` were never constructed and appear in no wire fixture; `crates/server/endpoints.json` is untouched.

- [x] Wire fixtures unchanged (no live DTO changed)
- [x] `endpoints.json` unchanged (no route changed)
- [x] No paired SDK change needed

## Documentation update

Updated in this PR: `tools/cargo-mero/README.md` (metadata table, flags, breaking changes), `tools/cargo-mero/SIGNING.md` (`ApplicationId` derivation, the shared `MERO_SIGN_KEY` resolution order, and that `metadata.author` gates who may publish a package), `docs/src/content/docs/build/cargo-mero.mdx`, `crates/AGENTS.md`, `tools/AGENTS.md`, the root `AGENTS.md` crate index, and a new `crates/bundle/AGENTS.md`.

Corrected an existing claim in `docs/src/content/docs/build/packaging-signing.mdx`: the `.mpk` carries the wasm and the ABI, not a state schema or migration artifacts.


rtb-12:
Doc bot: addressed the parts that apply.

- Root `AGENTS.md` (`CLAUDE.md` symlinks to it): added `crates/bundle/` to the JIT index table, since it is a new crate and the table is the repo's map. Nothing else in that file went stale, `cargo mero build` is unchanged by this PR.
- `crates/AGENTS.md`, `tools/AGENTS.md`, and a new `crates/bundle/AGENTS.md` were already updated in this PR: the new crate, the `cargo-mero` command and flag surface, and the replacement of the stale "no lib target" gotcha now that `src/lib.rs` exists.
- `CONTRIBUTING.md`: no reference to bundling, `cargo mero`, or the manifest, so nothing to update.
- `docs/`: updated `build/cargo-mero.mdx`, and corrected an existing false claim in `build/packaging-signing.mdx` that the `.mpk` carries a state schema and migration artifacts. It carries the wasm and the ABI.

rtb-12:
Good catch on `next_from_listing`, that was a real hole and the same class as the wrapped-listing one.

Fixed in `b54684a93`. Only an genuinely empty array now means "nothing published"; a listing that has entries but no readable `appVersion` among them errors and names the strings it could not read, rather than defaulting to `0.1.0` and signing below whatever is already out there.

A mixed listing still behaves as before, one bad entry does not discard the good ones. `skips_malformed_versions_instead_of_crashing` still passes, and `errors_when_entries_exist_but_none_carry_a_readable_version` covers the case you found. 14/14 registry tests green.

rtb-12:
On the `build-wasms.sh` warning from review `4854628967`: checked, and the concern is inverted. The `cd "$suite"` is what makes it land in the repo-root `dist/`, and the suggested fix is what would break it.

```rust
fn base_dir(metadata: &cargo_metadata::Metadata, args: &BundleArgs) -> Utf8PathBuf {
    args.manifest_path
        .as_ref()
        .map(|p| workspace::manifest_dir(p))
        .unwrap_or_else(|| metadata.workspace_root.clone())
}
```

The default output directory is anchored to `metadata.workspace_root`, not the process CWD. Every migration suite is a member of the root workspace, so `cargo metadata` run from inside `apps/migrations/<suite>` resolves `workspace_root` to the repo root and the bundle lands at `dist/<package>-<version>.mpk`.

Passing `--manifest-path`, which the suggestion recommends, is the one thing that *would* relocate it: that arm resolves to the manifest's own parent, so each fixture would be scattered into `apps/migrations/<suite>/dist/` and the CI artifact glob would find nothing. That is exactly why the script cds instead.

Verified beyond reading the code: `bash workflows/app-migration/build-wasms.sh` was run end to end from a clean tree and produced every versioned fixture under the top-level `dist/`, and the app-migration scenarios consume those same paths in CI on this branch.

No change made. Happy to add a line to the script noting why it cds rather than passing `--manifest-path`, since it is a genuinely surprising detail and this is the second time it has come up.

Issue #3366, opened 2026-08-03
GET /namespaces/:id/account reports a synthetic account before enrolment, so a grant made from it dies at enrolment


chefsale:
## Summary

`GET /admin-api/namespaces/:namespace_id/account` resolves the node's account through `calimero_governance_store::account_for_group` (`crates/server/src/admin/handlers/namespaces/get_account.rs:60`). That function returns the account from the node's **binding** if one exists, and otherwise falls back to `legacy_account_id(sign_pk)` — the op-adapter's synthetic stand-in (`crates/governance-store/src/node_device.rs:265`, `crates/op-adapter/src/lib.rs:85`).

A binding only exists once the node has enrolled. So the endpoint returns **two different ids over its lifetime** for the same node and namespace:

| | value |
| --- | --- |
| before `account create` | `SHA256("calimero.op-adapter.legacy-account.v1" ‖ sign_pk)` |
| after `account create` | `H(genesis(root_pk, derive_account_nonce(root_secret, namespace)))` |

Different domains, different inputs, unrelated values.

Three layers of documentation assert the opposite — that the value is stable and root-derived:

- `crates/server/src/admin/handlers/namespaces/get_account.rs:19` — *"the account id is derived from this node's root and the namespace, so it exists whether or not `account create` has ever run"*
- `crates/client/src/client/namespace.rs:105` — *"Read-only and always answerable: the account is derived from this node's root, so it exists before any device is enrolled"*
- merobox's `AccountShowStep` docstring, which repeats it

`account_for_group` itself is **not** wrong — it answers "what account do my writes attribute to right now", and pre-enrolment the honest answer is the stand-in, because that is the only value a peer can derive. The defect is that a read API reuses it to answer a different question — "what account does this node own" — and documents the second while returning the first.

## Impact

Everything an operator does with an account id starts from this endpoint (`meroctl account show`, merobox's `account_show`, any UI). The natural sequence *read your account → have an admin grant it → enrol* produces a grant that names a principal the node stops being:

1. node-2 reads its account: `A_synthetic`.
2. Admin grants `A_synthetic` as a writer. Rotation lands, writer set contains it.
3. node-2 enrols. Its writes now attribute to `A_real`.
4. Every guarded write is refused. The writer set still names `A_synthetic`; nothing in the grant, the rotation log, or the refusal mentions that the node's account changed underneath it.

[...]

Issue #3346, opened 2026-07-31
Make accounts first-class: carry AccountId natively and enrol by construction


chefsale:
> Rewritten 2026-08-05. The earlier two-phase split (implicit enrolment first, wire change later) was **wrong** and is gone — #3378 proved the phases are not separable. Comment history is the audit trail.

## Summary

`calimero-op-adapter` mints stand-in accounts for member keys via `legacy_account_id` / `legacy_authorship`, because the governance wire types carry `member: PublicKey` while the unified causal log needs an `AccountId` for every author. The bridge is correct and documented, but it means two representations of "who acted" coexist: a real `AccountId` derived from a verified genesis, and a synthetic one hashed from a signing key.

Every place that must decide *which* it is holding is a place two nodes can decide differently. Not hypothetical — it already produced a `scope_root` split in #3320, fixed by `account_for_author`'s precedence rules, which exist solely because the bridge does.

**Decision (@chefsale): enrolment must be implicit. `POST /namespaces/:id/account` should never be called by an operator.** There is no policy in enrolling — no case where a node that joined a namespace should lack an account there — and every consequence of skipping it is silent misattribution rather than an error. Evidence it is already forgotten in practice: **3 of 57** e2e workflows enrol.

## Why this is one change, not two

The obvious split is "make enrolment implicit now, change the wire later". #3378 implemented exactly that and broke **45 CI jobs** against a master that was 73/73 green the same day.

Enrolling changes **which principal a node presents**: before it, writes attribute to `legacy_account_id(sign_pk)`; after it, to the node's real `AccountId`. So every account-keyed grant or membership recorded *before* enrolment stops matching the moment it enrols. And because a refusal writes nothing while the op keeps its place in the DAG, later ops orphan behind it — so a mismatch on one node surfaces as a timeout somewhere unrelated.

Two suites, two symptoms, one cause:

| suite | symptom |
| --- | --- |
| e2e | `requester lacks permission to change subgroup visibility (CAN_MANAGE_VISIBILITY)` |
| app-migration | `join_subgroup_inheritance failed` — a node inherits membership recorded against its stand-in |

This is the operational footgun already documented for writer sets — *"a writer set seeded before its writer enrolls holds that key's stand-in account, so re-grant after `account create`"* — except it is not specific to writer sets. It applies to **every** account-keyed plane, and enabling enrolment globally turns a documented footgun into a fleet-wide one.

[...]

PR #3378, opened 2026-08-04
feat(context): enrol this node's account without being asked

chefsale:
**Phase 1 of #3346** — implicit enrolment. No wire change, no flag day, no re-bootstrap.

## Why

Enrolling has no policy in it. There is no case where a node that joined a namespace should not have an account there, and every consequence of skipping it is *silent misattribution* rather than an error: without a device binding the node's writes attribute to `legacy_account_id(sign_pk)` — a stand-in derived from a bare key, with none of an account's properties — and a writer grant made to its real account never matches.

A surface with no decision behind it is one whose only function is to be forgotten, and it was: **3 of 57** e2e scenarios enrolled.

## The trigger is key delivery, not the join

`AccountDeviceLinked` is an **encrypted** `GroupOp`, so it needs the namespace scope key — which a joiner does not have at join time. Enrolling on join deadlocks on the very link that would get it the key, which is why `create_account` refuses outright when the keyring is empty.

So this subscribes to `OpEvent::GroupKeyDelivered`, which already fires for exactly that moment and is already consumed by `join_group` as its wake-up signal. A second subscriber, not a new mechanism.

## Plus a startup sweep, and it needs no worklist

The event alone is not enough: a node that received its key while this listener was not running — an older binary, a crash between delivery and enrolment — has no event coming.

The nice property is that the sweep needs **no persisted state**. "I hold a key for this namespace and hold no device" *is* the work item, derivable from what the node already keeps. `rotation_listener` needs a durable `GroupPendingKeyRotation` row for the same reason this does not: what it owes is not otherwise visible.

## Choices worth reviewing

- **Nothing is minted here.** A `DeviceId` is the replica id for that namespace's CRDT state, and `ensure_enrolled` goes to lengths never to mint a second one — a second id strands everything written under the first. This asks the actor, which routes to that one idempotent decision under its lock. That is what makes reacting to a best-effort broadcast safe.
- **Failures are `debug!`, not `warn!`.** There is no caller to return an error to, and the common reason to land there is a race this listener is *designed* to lose (identity or keyring not readable yet, another trigger got there first). All are retried by the next event or sweep. Warning on each would make a healthy node look broken every time it joined something.
- **`holds_any_key`, not "the current key".** Any key means the node can encrypt a `GroupOp`, which is all enrolment needs. Refusing a node whose key has rotated would strand exactly the node most in need of a binding.
- **Groups resolved to namespaces, then deduped.** A key and a device are namespace-scoped while membership is per group, so counting groups would ask the actor to enrol the same node once per subgroup.
- **Modelled on `rotation_listener`** — synchronous subscribe before spawn (closes the missed-event race), lagged handling, own task per event, shutdown-then-spawn rebinding.

## Tests

Five over `pending_namespaces`, the derived condition the sweep rests on:

| test | pins |
| --- | --- |
| `a_namespace_with_a_key_and_no_device_is_pending` | the work item, and that nothing had to be written down for a restart to find it |
| `a_namespace_without_a_key_is_not_pending` | `create_account`'s refusal is unreachable via this path, not merely survivable |
| `an_enrolled_namespace_is_not_pending` | the steady state; re-minting would strand CRDT state |
| `namespaces_are_judged_independently` | one ready namespace does not drag in a keyless sibling, one enrolled does not mask an unenrolled one |
| `subgroups_do_not_produce_duplicate_work` | two subgroups of one namespace share a keyring and a device slot |

239 context tests pass; `cargo fmt --check` and `clippy -D warnings` clean.

## Not in this PR, deliberately

Removing `POST /namespaces/:id/account` and dropping `account_create` from the three e2e scenarios. `ensure_enrolled` is idempotent so the endpoint stays harmless, and retiring an operator surface plus rewriting scenarios is a separate reviewable step. This one is the mechanism.

Also still true and unchanged by this PR: enrolment becomes *automatic*, not *guaranteed*. A node that never receives a key still has no device, and `account_for_group` still falls through to the stand-in. Only Phase 2 (the wire change, bundled with P5/P6) closes that — see #3366.

## Docs

`crates/context/AGENTS.md`: the listener table, the file list, and two invariants — that `enrol_listener` is order-independent among the spawns, and why the trigger is key delivery rather than the join.


chefsale:
**Back to draft: this breaks the suite broadly, and the cause looks structural rather than a trigger bug.**

## The A/B

| | jobs | failing |
| --- | --- | --- |
| master, same day | 73 | **0** |
| this branch | 73 + migration suite | **44** |

So this is mine, not a flake. The failures span scenarios with nothing to do with accounts — `group-capabilities`, `group-metadata`, the whole app-migration suite, `00-single-group-migration-baseline`.

## What the logs say

The good news first: **auto-enrolment now works.** After the periodic-sweep fix, nodes do enrol (`enrolled this node's account automatically`, and `create_account`'s own `enrolled this node's device into a fresh account`). The trigger problem from the previous round is fixed.

The bad news is what follows:

```
15×  WARN sync::manager::namespace_sync: failed to apply namespace governance op from backfill
     namespace_id=907320a0… delta_id=b1762a1f… op_kind=Group nonce=6 err=
 1×  ERROR admin::handlers::context::join_context: Failed to join context via group … 500
 3×  WARN governance_broadcast: governance op applied locally but not confirmed by an authoritative node
```

One `delta_id` fails to apply **15 times** and never succeeds. A namespace governance op that can never apply does not fail in isolation — later ops orphan behind it, which is exactly the shape of "unrelated scenario times out joining a context".

## Why I think this is structural

Phase 1 as specified adds an **encrypted `GroupOp` to the shared namespace DAG, automatically, early in a node's life** — right after key delivery, before much else has settled. If that op cannot be authorized at its causal cut on a peer, it wedges the DAG rather than merely failing.

That is the same class of problem #3346 Phase 2 exists to remove: fold the link into the **cleartext join op**, where it is authorized by construction and needs no scope key. Which raises a question worth deciding before more work here — **whether Phase 1 is safely separable from Phase 2 at all**, or whether "enrol automatically" only becomes safe once the link stops being a separate encrypted op.

I have not confirmed the failing delta *is* the `AccountDeviceLinked` op. `err=` renders empty in that warning (a diagnostic gap of its own), and correlating the delta id to a publisher needs another pass.

## What I would do next, in order

1. Confirm the failing delta is the account link, and get a real error out of that apply path — `err=` being empty is why this took a download and a correlation attempt instead of a read.
2. If confirmed: raise on #3346 that Phase 1's premise needs revisiting, because the trigger was never the hard part.
3. Keep the parts that are independently good — `pending_namespaces` and its five tests, and the finding that no single event means "this node has a key".

Not proposing a quick fix. I got the trigger wrong once already this round, and guessing again on a change with a 44-job blast radius is not the move.

chefsale:
**Mechanism confirmed.** The error was in the log all along — `?err` on an `eyre::Report` renders multi-line, and my earlier single-line grep cut it off. That was my extraction bug, not a logging gap.

```
namespace DAG apply error: Failed to apply delta:
  requester lacks permission to change subgroup visibility (CAN_MANAGE_VISIBILITY) in group …
```

Enrolling changes **which principal the node presents**: before it, writes attribute to `legacy_account_id(sign_pk)`; after it, to the real `AccountId`. So an account-keyed grant made before enrolment stops matching the moment the node enrols. The op is refused, refusals write nothing while the op keeps its place in the DAG, and later ops orphan behind it — a capability mismatch on one node becomes a timeout in a migration scenario.

Also correcting myself twice over: I *did* suspect this and checked for it, grepped for **writer-set** refusals specifically, found none, and concluded the risk had not materialised. It had — one gate over, on capabilities. And my two subsequent mechanism guesses (link op refused; link op parked on `AuthorityUndecidable`) were both wrong: the link ops applied cleanly, 4 applies across 2 nodes, no refusals in the logs.

Full write-up and the options on #3346. Short version: **Phase 1 is not separable from Phase 2** — the split rests on there being no wire change, which is true of the mechanism and false of the effect, and no ordering avoids it while a "before enrolment" window exists. My recommendation there is to fold implicit enrolment into the flag day.

Leaving this draft. What is worth salvaging regardless: `pending_namespaces` with its five tests, and the finding that no single event means "this node has a key".

chefsale:
Closing: **implicit enrolment folds into #3346's flag day.** Decision by @chefsale, on the evidence below.

Phase 1's premise was that enrolling needs no wire change. True of the mechanism, false of the effect: enrolling changes **which principal a node presents** — from `legacy_account_id(sign_pk)` to its real `AccountId` — so every account-keyed grant or membership recorded before it stops matching. Refusals write nothing while the op keeps its place in the DAG, so later ops orphan behind it.

Confirmed in two independent suites, one mechanism:

| suite | symptom |
| --- | --- |
| e2e | `requester lacks permission to change subgroup visibility (CAN_MANAGE_VISIBILITY)` |
| app-migration | `join_subgroup_inheritance failed` — node-2 inherits membership recorded against its stand-in |

45 jobs against a master that was 73/73 green the same day. No trigger change fixes this; Phase 2 removes the window rather than working around it.

## Worth salvaging into that work

The branch stays; the commits are `ac7a7bbc0` (mechanism) and `53a4dc1a7` (periodic sweep).

1. **`pending_namespaces` and its five tests.** The derived work-item — "holds a key, holds no device" — needs no persisted worklist, and the tests cover the cases that matter (key-and-no-device is pending; no key is not; enrolled is not; namespaces judged independently; subgroups of one namespace produce one item, not one per group).
2. **There is no single event meaning "this node now has a key."** `OpEvent::GroupKeyDelivered` is emitted only from `apply_received_group_key_envelope` (an applied `RootOp::KeyDelivery`). A joiner normally gets its key from the direct join response and `recover_missing_group_keys`; a creator mints its own. Measured, not assumed: a three-node scenario that demonstrably had keys logged **zero** deliveries. An event-only trigger enrols nobody — which is how the first version here passed its unit tests and did nothing. This is now recorded in `crates/context/AGENTS.md` on master via no other PR, so it should be carried over explicitly.
3. **The listener shape** (synchronous subscribe before spawn, periodic sweep for correctness, `info!` on success so "did it happen" is answerable from an ordinary log) if enrolment still needs a local trigger after the flag day — it may not, since Phase 2 enrols at join.

## Mistakes worth recording

- I suspected the re-keying risk, grepped for **writer-set** refusals specifically, found none, and concluded it had not materialised. It had — one gate over, on capabilities.
- Two subsequent mechanism guesses (link op refused; link op parked on `AuthorityUndecidable`) were both wrong; the link ops applied cleanly.
- `?err` on an `eyre::Report` renders multi-line. My single-line grep made the error look empty and cost a round.

PR #3399, opened 2026-08-12
feat(storage)!: own an authored entry by account, not by device

chefsale:
Write a note from your phone and you could not edit it from your laptop. `AuthoredMap`/`AuthoredVector` stamped each entry with the inserting **device's** key and gated `update`/`remove` on it, so a second machine was a stranger to its own data. `UserStorage` had it twice over — a per-device slot as well as a per-device stamp — so "this user's data" meant "this machine's".

Authorship is an access-control question, and access control names people. This makes `StorageType::User.owner` an `AccountId` and keys `UserStorage` by one, so slot and gate agree. It is the storage-plane counterpart of #3391, which did the same for governance.

## Why it is bigger than a type swap

`owner` was doing two jobs, and only one survives: it was **also** the ed25519 key the signature verified against (`ed25519_verify(.., owner.digest(), ..)`), and an account is a content hash that verifies nothing.

So authentication and authorization split the way `Shared` already splits them:

- the signature is checked against `signature_data.signer` — now required for `User`, and the permanent record of **which device** wrote the entry, so no information is lost;
- `ApplyContext.signer_account`, the node's resolution of that key **at the action's causal cut**, is checked against `owner`;
- `payload_for_signing` commits to `signer`, which it did not have to when `owner` was itself the verifying key.

## The part that was not obvious

The sync repair paths — HashComparison, snapshot, level-wise — apply through an `ApplyContext::empty()`, because they carry no cut to resolve a signer's account at. Storage cannot ask the ownership question there, and refusing would drop every legitimately repaired `User` entity. (`hashcomparison-shared-user-reconcile` exists precisely to exercise that path.)

Deferring alone would have made this a **downgrade**: `owner` used to be the verifying key, so a repaired leaf was cryptographically bound to its owner with no bindings in hand. Once `owner` is an account, any context member could push another member's authored entry over HC.

So the ownership half moves to the node, which has the bindings and is already running the membership half three lines above. `is_leaf_currently_authorized` now resolves the author's key to its account and compares it against the leaf's `owner`. Unresolvable stays permissive — that is this node not having folded the author's enrolment, not evidence of an impostor.

## What deliberately did NOT move

The per-replica identities CRDT mechanics need — an LWW register's tiebreak, a counter's per-actor slot, an HLC seed — stay on `device_id`, and must: two devices of one person writing concurrently have to stay distinguishable or they share a slot and lose each other's writes. Nothing keys an authored entry's id off its owner, so convergence is untouched.

## Two tests existed to prevent this, and both are inverted

- `two_devices_of_one_account_own_their_entries_separately` refused the phone the laptop's entry, justified as stopping devices from "clobbering each other's per-writer state". That conflated the gate with the state — see above; the state never moved. Now `..._share_ownership_of_its_entries`, with a guard that a **different** account is still refused (without it, "any device may edit" would pass too).
- Step 5 of `shared-storage-account-writers-two-devices` asserted the same over the real wire. The phone now updates the laptop's note and the assertion checks it propagated — which also exercises the new receiver-side resolution on the authored plane, the twin of what step 4 does for writer sets.

Test identities derive the account from a different domain than the key (`create_test_owner` → `account_of_key`), so a green suite cannot be green because the two happened to be equal.

## Compatibility

`StorageType::User.owner` changes meaning with an **identical borsh layout** — both ids are 32 bytes — so nothing fails to decode; a pre-change entry simply reads a device key as an account id and ends up owned by nobody.

Originally written as "ships with the #3346 flag day". That was wrong on two counts: #3346 closed COMPLETED when #3391 merged (it *was* the flag day, and it has happened), and there is no running network — so there is no deployed data to migrate and no mixed-version pair to break.

The only thing to know: **wipe local node data** when picking this branch up, or authored entries written before it will silently have an owner that matches nobody.

`get_user_simple_for` is account-typed, so `e2e.yml` and `hashcomparison-shared-user-reconcile.yml` read `my_account` off the owning node instead of passing a public key.

## Verification

- `cargo test --workspace --no-fail-fast`: **227 suites, 0 failures** (689 in `calimero-storage`)
- `cargo fmt --check` and `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` clean
- Dependency sweep first: the migration scenarios compare `owner_of` to `owner_of` (value-agnostic, unaffected), and the multi-node scenarios use separate nodes = separate accounts, so the distinctness they rely on is preserved.

### DIFF
--- a/apps/kv-store-with-user-and-frozen-storage/src/lib.rs
+++ b/apps/kv-store-with-user-and-frozen-storage/src/lib.rs
@@ -6,7 +6,7 @@ use calimero_sdk::abi::AbiType;
 use calimero_sdk::app;
 use calimero_sdk::borsh::{BorshDeserialize, BorshSerialize};
 use calimero_sdk::serde::Serialize;
-use calimero_sdk::PublicKey;
+use calimero_sdk::{AccountId, PublicKey};
 use calimero_storage::collections::Mergeable;
 use calimero_storage::collections::{FrozenStorage, LwwRegister, UnorderedMap, UserStorage};
 use thiserror::Error;
@@ -17,11 +17,11 @@ pub struct KvStore {
     items: UnorderedMap<String, LwwRegister<String>>,
 
     // Simple user-owned data (e.g., a user's profile name)
-    // Stores: UnorderedMap<PublicKey, LwwRegister<String>>
+    // Stores: UnorderedMap<AccountId, LwwRegister<String>>
     user_items_simple: UserStorage<LwwRegister<String>>,
 
     // Nested user-owned data (e.g., a user's private key-value store)
-    // Stores: UnorderedMap<PublicKey, UnorderedMap<String, LwwRegister<String>>>
+    // Stores: UnorderedMap<AccountId, UnorderedMap<String, LwwRegister<String>>>
     user_items_nested: UserStorage<NestedMap>,
 
     // Content-addressable, immutable data
@@ -152,7 +152,10 @@ impl KvStore {
     }
 
     /// Gets the simple string value for a *specific* user.
-    pub fn get_user_simple_for(&self, user_key: PublicKey) -> app::Result<Option<String>> {
+    ///
+    /// Named by ACCOUNT: user storage is keyed per person, so this reaches the
+    /// one slot every device of theirs writes to.
+    pub fn get_user_simple_for(&self, user_key: AccountId) -> app::Result<Option<String>> {
         app::log!("Getting simple value for specific user {:?}", user_key);
         Ok(self
             .user_items_simple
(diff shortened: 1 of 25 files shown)


chefsale:
⚠️ **Do not merge ahead of the #3346 flag day.**

`StorageType::User.owner` changes meaning with a byte-identical borsh layout — both ids are 32 bytes — so existing data does not fail to decode, it silently reads a device key as an account id and the owner then matches nobody. Landing this on its own would strand every existing authored entry rather than error.

Ready for review; just needs its merge sequenced with the rest of the account-identity break (#3403 has the same constraint).

chefsale:
**Correction to my comment above — disregard the merge hold.**

That comment said to sequence this behind the #3346 flag day. Two things make it wrong:

1. #3346 closed COMPLETED when #3391 merged — it *was* the flag day (schema bump + re-bootstrap), and it already happened. There is nothing pending to ride.
2. There is no running network, so there is no deployed data to strand and no mixed-version pair to break.

No migration and no version gate are needed. The only residue is a **local dev datastore predating this change**: the borsh layout is byte-identical, so such a store opens fine and quietly reads a device key as an account id, leaving authored entries owned by nobody. Wipe local node data when picking this branch up.

Normal review rules apply — no special merge sequencing.

PR #3403, opened 2026-08-13
fix(sync)!: name a namespace joiner by credential before serving it anything

### LINKED ISSUE #3397: join-sync admission skips the deny-list check for a joiner whose account cannot be resolved
## Summary

`NamespaceSync`'s join-sync admission gate (`crates/node/src/sync/manager/namespace_sync.rs`, the `(already_member, joiner_account)` match) cannot run the re-entry / deny-list check when it fails to resolve the requester's signing key to an account, and admits the request instead.

The join request carries a **key and an invitation, not a credential**, so the responder has nothing to name the requesting principal with beyond its own binding rows. Both rows the gate reads — the re-entry block and the consumed-invitation marker — are keyed by account, so an unresolvable requester reads as "no rows, nothing to check".

For a genuine first-time joiner that is the correct answer. For a **denied account presenting a device key this responder holds no binding for**, it is not: the deny row exists, but under a name the responder cannot compute.

## Impact

A previously removed/denied account that holds a live open invitation can open a join-sync stream with a fresh device key and be served the backfill and the wrapped group key, ahead of the apply-time check that does have the credential and does reject them.

Severity is bounded by two things and neither makes it a non-issue:

- It is **not a regression**. Before governance principals were named by account (#3346 slice B / #3391), this gate was keyed by the signing key, so a fresh key walked past it identically. Keying by account is what lets a denied account's *other, already-bound* devices be caught at all — this issue is the remaining tail, not something the account flip introduced.
- The authoritative gate still holds: the joiner's own `MemberJoinedAt` apply carries the credential and rejects them, so they do not become a member.

What leaks in the window is namespace governance state and the current group key — to a party this namespace has already denied.

## Steps to reproduce

1. Namespace with an admin and a member; admin creates an open invitation.
2. Remove the member (writes their deny/re-entry row, keyed by their account).
3. On a fresh node, enrol a **new device** under the same account root, so the account is unchanged but the device key is one no other node has a binding for.
4. Open a namespace join-sync stream from that node with the invitation from step 1.

**Actual**: admitted — `joiner_account` resolves to `None`, the match takes the `(false, None)` arm, and the responder proceeds to wrap the group key and serve backfill. The rejection lands later, at `MemberJoinedAt` apply.

**Expected**: refused at the gate, before any key wrap or backfill, because the account is denied.

## Criteria for resolving

- [ ] The join-sync request carries the joiner's credential (genesis + chain + cert), so the responder can name the account **before** it serves anything — the same material `join_group` already uses for its self-binding.
- [ ] The admission gate resolves the requester through that credential and runs `require_invitation_admits` against the resulting account; a request whose credential does not verify is refused rather than admitted.
- [ ] The `(false, None)` arm no longer exists, or is reachable only for a peer that provably sent no credential (old node), with that leniency explicitly bounded.
- [ ] Regression test: a denied account presenting a newly-enrolled device key is refused at the gate, asserted by the absence of a key wrap / backfill response rather than only by the eventual apply-time rejection.
- [ ] Backward compatibility considered explicitly: an older joiner sends no credential, so this is a wire change that needs a version gate (E2E cannot catch mixed old/new interop — see CLAUDE.md).
- [ ] `cargo fmt` / `clippy` / `test` pass.

Found in review of #3391; the code comment at the `(false, None)` arm points here.

chefsale:
Closes #3397.

A namespace join request named its sender by signing key, while the rows that decide admission — the re-entry block and the consumed-invitation marker — are keyed by account. So whenever the responder held no binding for the presented device, it could not name the requester, and the `(false, None)` arm admitted them.

That is not hypothetical for the case that matters: **a denied account presenting a newly-enrolled device is exactly the request the responder cannot name.** Its deny row exists, under a name the responder cannot compute. It then wrapped the group key and served backfill, ahead of the apply-time check that does reject the join.

## The fix

`NamespaceJoinRequest` now carries the joiner's `JoinAccountCredential` — genesis, root-key chain, and the device cert for the key making the request. The responder verifies it and derives the account **before** it serves anything, then runs `require_invitation_admits` against that account. The unnamed-admit arm is gone; there is no lenient path left.

`verified_joiner_account` checks three things, and the third is the one that is easy to miss:

1. the genesis hashes to the account the certificate claims — so a credential cannot wear an account it does not descend from;
2. the certificate chains to that genesis through the root-key handoffs (`verify_device_cert`, the same check the apply path runs);
3. **the certificate names this request's `joiner_public_key`.** A credential travels in the clear inside a join request, so every peer that has served one holds a copy. Without this tie it is a bearer token — replay it and be admitted as its owner, past their deny-list entry. That would be a worse hole than the one being closed.

Nothing is written during verification. Applying the binding first would let an unauthorized request mutate state on its way to being refused.

Two things fall out of always having an account:

- the joiner pre-registration is now unconditional. It used to be skipped whenever the account could not be named — which was precisely the first-join case the optimisation exists for.
- `join_group` builds the credential *before* the request rather than after, and fails the join if it cannot. The same credential is still published with `MemberJoinedAt`.

## On the version gate

The issue asked for one. There is nothing to gate, and the reasoning changed twice while investigating:

Borsh encodes a variant tag then that variant's fields, so adding a field to `NamespaceJoinRequest` breaks **only that message**. Blob sync, HashComparison, delta requests and snapshots interoperate between versions untouched. An old 3-field request reads a length prefix out of the nonce bytes and errors cleanly rather than mis-parsing.

That failure is also exactly what should happen: an old node cannot supply a credential, so it must not be admitted. A bounded "admit if the peer sent no credential" path would be indistinguishable on the wire from the attack — a denied account simply omits it — so it would preserve the vulnerability behind a flag.

I briefly considered bumping `CALIMERO_STREAM_PROTOCOL`. That would partition *all* stream traffic between versions to gate a break confined to one message type — strictly worse, and not done.

And there is no running network, so none of this is live anyway.

## Verification

- 5 new tests covering each refusal: valid credential resolves to the genesis-derived account; a credential for a **different** key is refused (the replay guard); a genesis that does not derive the claimed account is refused; undecodable bytes are refused rather than falling through; an absent credential is refused, since "could not name them" was the original bug.
- `cargo test --workspace --no-fail-fast` green; `fmt --check` and `clippy -D warnings` clean.

### DIFF
--- a/crates/context/src/handlers/join_group.rs
+++ b/crates/context/src/handlers/join_group.rs
@@ -218,8 +218,33 @@ impl Handler<JoinGroupRequest> for ContextManager {
                 // Previously this was a hard `?`, so a join attempted before the
                 // mesh formed aborted before writing the joiner's membership row
                 // and left the node not-a-member (the reported symptom).
+                // Built BEFORE the request, not after: the responder needs it to
+                // name this joiner's account, and the deny-list gate it feeds
+                // runs before any state is served. The same credential is
+                // published with `MemberJoinedAt` further below.
+                let joiner_credential_bytes = match crate::join_credential::build(
+                    &datastore,
+                    &namespace_id.into(),
+                    &joiner_identity,
+                ) {
+                    Ok(credential) => borsh::to_vec(&*credential)?,
+                    Err(e) => {
+                        // Without one the responder cannot name us, and a
+                        // responder on this version refuses rather than guess.
+                        return Err(e.wrap_err(
+                            "join: could not build the account credential the responder \
+                             needs to authorize this join",
+                        ));
+                    }
+                };
+
                 let join_result = match node_client
-                    .request_namespace_join(namespace_id, invitation_bytes, joiner_identity)
+                    .request_namespace_join(
+                        namespace_id,
+                        invitation_bytes,
+                        joiner_identity,
+                        joiner_credential_bytes,
+                    )
                     .await
                 {
                     Ok(bundle) => bundle,
--- a/crates/node/primitives/src/client.rs
+++ b/crates/node/primitives/src/client.rs
@@ -43,6 +43,9 @@ pub struct NamespaceJoinParams {
     pub namespace_id: [u8; 32],
     pub invitation_bytes: Vec<u8>,
     pub joiner_public_key: PublicKey,
+    /// Borsh-serialized `JoinAccountCredential`, so the responder can name this
+    /// joiner's ACCOUNT before it serves any state. See the wire variant.
+    pub joiner_credential_bytes: Vec<u8>,
 }
 
 /// Parameters for a direct open-subgroup join request (issue #2357).
@@ -122,12 +125,14 @@ impl SyncClient {
         namespace_id: [u8; 32],
         invitation_bytes: Vec<u8>,
         joiner_public_key: PublicKey,
+        joiner_credential_bytes: Vec<u8>,
     ) -> eyre::Result<JoinBundle> {
         let (tx, rx) = oneshot::channel();
         let params = NamespaceJoinParams {
             namespace_id,
             invitation_bytes,
             joiner_public_key,
+            joiner_credential_bytes,
         };
         self.ns_join_tx
             .send((params, tx))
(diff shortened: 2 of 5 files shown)


chefsale:
⚠️ **Do not merge ahead of the #3346 flag day.**

This adds a field to an existing `InitPayload` variant and `borsh::from_slice` is strict, so old and new nodes cannot talk in either direction. Landing it alone would break namespace joins between mixed-version nodes.

Ready for review; just needs its merge sequenced with the rest of the account-identity break (#3399 has the same constraint).

chefsale:
**Correction to my comment above — disregard the merge hold.**

Two corrections:

1. #3346 closed COMPLETED when #3391 merged; it was the flag day and it has already happened. Nothing pending to ride.
2. I overstated the blast radius. Borsh encodes a variant tag then that variant's fields, so adding a field to `NamespaceJoinRequest` breaks **only that message**. Blob sync, HashComparison, delta requests and snapshots all interoperate untouched. An old 3-field request reads a length prefix out of the nonce bytes and errors cleanly rather than mis-parsing.

That failure is also the intended outcome: an old node cannot supply a credential, so it must not be admitted. With no running network it is moot regardless.

No flag day, no protocol-version bump. Normal review rules apply.

Issue #3376, opened 2026-08-04
bug(sync): HashComparison always fails verifying a SharedMember leaf, so entity divergence never reconciles


chefsale:
## Summary

Every HashComparison sync session in the affected scenarios fails with `StorageError::InvalidSignature` — surfaced as `error=Invalid signature for user-owned data` — and falls back to DAG catchup. DAG catchup moves operations; it does not reconcile entities. So the entity divergence that HC exists to repair is never repaired, and two nodes settle into **same DAG heads, different root hash**.

This is the failure behind the intermittent non-convergence currently blamed on #2716. #2716's own defect (the stale-nonce skip dropping a rotation-log append) was genuinely fixed by #2718 + #2727; its signature — a sticky `nodes_skipped=3` HC loop — is absent from these runs. Filed separately so the two are not conflated again.

## Evidence

Two different scenarios, same failure, from the node-log artifacts:

| scenario | HC failures | of which `Invalid signature for user-owned data` | gov-divergence pulls |
| --- | --- | --- | --- |
| [`hashcomparison-shared-user-reconcile`](https://github.com/calimero-network/core/actions/runs/30919307758/job/92027588575) | 186 | 186 | 186 |
| [`shared-storage-concurrent-rotation-converge`](https://github.com/calimero-network/core/actions/runs/30820525092) | 518 | 517 | 518 |

The pull count matching the HC failure count exactly is the giveaway that one causes the other:

```
WARN calimero_node::sync::protocol_selector: HashComparison sync failed,
     falling back to DAG catchup
     error=Invalid signature for user-owned data
```

and the outcome, after ~90–180s of that loop:

```
ERROR calimero_node::handlers::network_event::heartbeat:
  DIVERGENCE DETECTED: Same DAG heads but different root hash (persisted across heartbeats)!
```

with a divergence dump naming **22 distinct entities** whose merkle hashes differ. `nodes_skipped` never sticks at 3 (it appears twice, as 2), so this is not #2716's loop.

Also confirmed with the `ops_pulled` instrumentation from #3371: the corrective governance pull runs to completion every time and transfers **the same 3 ops** on all 186/518 attempts, while both nodes' `scope_root` stay frozen at mirror-image values. The pull is healthy and irrelevant — it has nothing to fix, because the entities are the problem.

## Where it fails

`crates/storage/src/interface.rs:790`, the snapshot-leaf verification's `StorageType::User` arm:

```rust
[...]

PR #3405, opened 2026-08-13
fix(node): name a repair-pushed leaf's signer so HashComparison can repair signed entities

chefsale:
Refs #3376. HashComparison could not repair a signed `Shared`/`SharedMember` entity — the one thing it is invoked for.

## Before

Two nodes diverge, HashComparison finds the differing entities, one pushes the other its copy of a leaf — and the receiver rejects it. Both directions, forever, so the repair never lands and HC falls back to DAG catchup on every attempt.

From one failing run's node logs: **174 rejections, every one identical.**

```
action signature rejected while applying
  id=ae87c628…9a73763
  reason="sharedmember-signer-not-in-writer-set"
  storage_type="SharedMember"   signature="signed"
```

One entity, one reason, `signature="signed"` on every line, split 89 / 85 across the two peers each refusing the other's copy.

**The signature was never the problem.** It surfaces to the caller as `InvalidSignature`, which is why the search kept looking at signatures — but the signature verifies. What failed was naming the **writer**.

A writer set names ACCOUNTS. A signature names a KEY. Storage bridges the two through `ApplyContext.signer_account` — and on the repair path that was always `None`:

```rust
let signer  = sig_data.signer?;
let account = signer_account?;   // ← always None here. Bails.
if !writers.contains_key(&account) { return None; }
```

Not because resolution failed — **because nobody attempted it.** `ApplyContext::empty()` hardcodes `signer_account: None`, and every repair path passes it. So the lookup could not run, and a check that cannot run fails closed. When writer sets became account-keyed, this check did not get stricter; it became unanswerable.

## After

The node answers it, because the node is where the bindings live:

- `repair_signer_account()` resolves the leaf's signer to its account, inside `apply_leaf_with_crdt_merge_gated` — where the store is already in scope for the membership gate a few lines above.
- `apply_leaf_with_crdt_merge_as()` passes it into the `ApplyContext`.
- Storage's existing `resolve_signer` then works unchanged: verify the signature under the named key, then check that account against the writer set.

**`crates/storage` is not touched by this PR.**

## Why resolving it live is sound

Two questions look similar and are not:

| question | changes over time? | how it is answered |
| --- | --- | --- |
| which accounts may write here? | **yes** | already resolved deterministically as of the leaf's own HLC, from the anchor's rotation log (`resolve_local_as_of`) |
| which account does this key speak for? | **no** | resolved live — safe, see below |

A device is bound to exactly one account for its entire life (`BindingRejected::AccountReassignment` — *"A device may not be moved between accounts"*). So two nodes that both know a binding always agree, whatever their fold depth. There is no "different answers" case, only *knows* versus *does not know yet*.

Both halves are therefore deterministic, and the real writer check runs on repairs exactly as it does on deltas.

## Nothing is weakened

- A validly-signed leaf from a **non-writer** is still refused on repair. It is not accepted and reconciled later — it is refused, as on the gossip path.
- An **unfolded binding** still refuses, and the leaf is re-driven on the next repair round. Two peers converge on one verdict rather than one accepting what the other rejects.
- Tampering was always caught and still is: the signature commits to id, data, nonce and owner.

An earlier version of this PR took a different route — deferring the writer question on repair paths, on the mistaken grounds that resolving live would cause divergence. That would have let a group member push a validly-signed value for an entity it holds no write grant on. It is gone; the reasoning that killed it is the table above.

## Second commit

Fixing the repair let `hashcomparison-shared-user-reconcile` clear the reconciliation barrier and run ~30 steps further, which exposed a stale assertion from #3399: `authored_get_owner` returns an account now, but the scenario still compared it to the node's bs58 signing key. It had been broken since #3399 merged and was invisible, because reconciliation died three steps before it. Fixed, along with removing the `get_namespace_identity` step that fed it and now has no reader.

Swept the other scenarios rather than spot-fixing — the app-migration authored cases compare `owner_of` to `owner_of`, so they are value-agnostic and unaffected.

## Verification

`cargo test -p calimero-node --lib` 469 passed · `cargo test -p calimero-storage --features testing` 689 passed · `fmt --check` and `clippy -D warnings` clean.

### DIFF
--- a/apps/scaffolding-e2e/src/lib.rs
+++ b/apps/scaffolding-e2e/src/lib.rs
@@ -1220,7 +1220,10 @@ impl E2eKvStore {
     }
 
     pub fn authored_get_owner(&self, key: String) -> app::Result<Option<String>> {
-        Ok(self.authored_items.owner_of(&key)?.map(|pk| pk.to_string()))
+        Ok(self
+            .authored_items
+            .owner_of(&key)?
+            .map(|account| account.to_string()))
     }
 
     pub fn authored_len(&self) -> app::Result<usize> {
@@ -1331,7 +1334,10 @@ impl E2eKvStore {
     }
 
     pub fn authored_vec_get_owner(&self, index: usize) -> app::Result<Option<String>> {
-        Ok(self.authored_vec.owner_of(index)?.map(|pk| pk.to_string()))
+        Ok(self
+            .authored_vec
+            .owner_of(index)?
+            .map(|account| account.to_string()))
     }
 
     pub fn authored_vec_entries(&self) -> app::Result<Vec<String>> {
--- a/apps/scaffolding-e2e/workflows/hashcomparison-shared-user-reconcile.yml
+++ b/apps/scaffolding-e2e/workflows/hashcomparison-shared-user-reconcile.yml
@@ -312,13 +312,6 @@ steps:
   #     back with downgraded provenance. Post-redesign, name + data
   #     match the LWW winner of the two writes node-1 issued. ---
 
-  - name: Node-1 reports its identity (used to read its User entity)
-    type: get_namespace_identity
-    node: e2e-node-1
-    namespace_id: '{{ns}}'
-    outputs:
-      node1_id: publicKey
-
   - name: Read node-1's User entity from node-2
     type: call
     node: e2e-node-2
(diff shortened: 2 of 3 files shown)


chefsale:
Update: the writer-deferral fix works — the scenario now clears the reconciliation barrier it used to time out on and runs ~30 steps further.

Doing so exposed a second, unrelated breakage that had been hiding behind it: `authored_get_owner` returns an ACCOUNT since #3399, but the scenario still asserted it equalled the node's bs58 signing key. That comparison went stale when #3399 merged and nothing caught it, because reconciliation died three steps earlier and the assertion was never reached.

Fixed here (`40cdb2607") along with removing the now-readerless `get_namespace_identity` step that fed it. Swept the other scenarios rather than spot-fixing — the app-migration authored cases compare `owner_of` to `owner_of`, so they are value-agnostic and unaffected.

Issue #3406, opened 2026-08-13
group-join-mesh-not-ready: a joiner that got no key never converges after heal (namespace-fallback dial is gated on the context topic it cannot join)


chefsale:
## Summary

`apps/scaffolding-e2e/workflows/group-join-mesh-not-ready.yml` fails at **"Wait for governance convergence after heal"** (120s, 190 attempts). It reproduces on master and on unrelated PR branches, so it is not caused by any one change in flight.

Evidence below is from the node-log artifacts of a failing run (`logs-group-join-mesh-not-ready-…`). The scenario has **no `type: call` steps at all**, so no application state is written — this is purely join + governance.

## What the logs show

The scenario deliberately partitions node-2, has it join anyway (expected failure, no key), then heals.

**node-2, while partitioned** — the designed state:
```
06:10:17  WARN join_group: direct namespace-join request found no reachable mesh peer;
                           recording local membership and relying on fallback
06:10:17  WARN join_group: join response contained no group key
06:10:22  KeyDelivery timed out for group ContextGroupId(…)
```

**node-1, for the whole partition window** — 37 discovery failures:
```
20x  Mesh peer discovery exhausted all retries: not connected to any peer (dial/discovery gap)
17x  Mesh peer discovery exhausted all retries: connected to peers but none subscribe to the context topic
 2x  Failed to bootstrap Kademlia err=No known peers
```

**At the heal (06:10:42)** the namespace plane recovers — and then stalls:
```
06:10:42.008  Peer subscribed to namespace topic, emitting out-of-cycle beacon
06:10:42.029  beacon advertises an unknown namespace          ← node-1
06:10:42.030  received namespace governance ops from peer
06:10:44 → 06:12:39  readiness beacon emitted / received, every ~3s, forever
06:10:49, 06:11:19, 06:11:49, 06:12:19  namespace heartbeat: divergence detected
```

So the peers reconnect, exchange beacons continuously, and node-1 even ingests governance ops — but the heartbeat reports divergence on every 30s tick until the barrier gives up.

## Two distinct problems

**1. A bootstrap deadlock in namespace-fallback discovery.**

`crates/node/src/sync/peers.rs:186` intersects namespace-topic peers with **context**-topic subscribers before dialing:

```
namespace fallback found peers but none subscribe to the context topic; not dialing — see #2422
```

[...]

Issue #2422, opened 2026-05-20
Namespace/group join can leave joiners un-auto-followed into contexts — peers reject inbound sync as "unknown context"


chefsale:
## Summary

Joining a namespace propagates the namespace + groups + members metadata correctly, but it does **not** guarantee the joiner is auto-followed into the contexts those groups own. Auto-follow is gated by per-member, per-group `AutoFollowFlags { contexts: true }` (see `crates/context/src/auto_follow.rs`). If that flag isn't on at the time `ContextRegistered` is applied, or the backfill path doesn't run, the joiner ends up as a group member with **no local entry for the context** — even though the UI happily shows them as a group member.

The downstream effect on peers: when the other member of the context tries to sync, the joiner's merod logs `inbound stream for unknown context, closing cleanly`, and the initiator side records a sync failure that eventually hits the 256s exponential-backoff ceiling. From the user's perspective: "I joined the namespace, the UI shows everything is fine, but the data isn't syncing."

## Reproducer (real user report)

Two users — Ronit and Fran — across two physical networks, both on the same `Mero Chat` namespace (`ec3e350a…74f6808c`), application `rXMEUvf7…dPXgVfUm`. Three groups, each containing one context. UI confirms both members see identical namespace/group metadata.

Each user has a different local-state matrix for the 3 contexts:

| Context (group) | Node A (Ronit) | Node B (Fran) |
|---|---|---|
| \`ARqLH3…\` (\`64279fc8…\`) | has state | uninitialized — trying to bootstrap |
| \`EoRockb…\` | uninitialized — trying to bootstrap | has state |
| \`9Gwoey…\` | **unknown context** (not in local store) | has state |

Both nodes are connected (~500 cross-references each in the logs). The failure modes are pure sync-layer:

- **The \`unknown context\` case** — Node A repeatedly logs:
  \`\`\`
  WARN  inbound stream for unknown context, closing cleanly
        context_id = 9GwoeyNFjVbm1F8A1ED3kxCqdSZyWr2VAFNSeEY8jdeA
        their_identity = B8pnz9ob…gJ4iUwTC
  ERROR failed to send OpaqueError for unknown context
        err = \"connection is closed\"
  \`\`\`
  Node B has state for \`9Gwoey…\` and the namespace-mesh-peer fallback picks Node A as a sync target, but Node A's local context store has no entry for \`9Gwoey…\` — auto-follow either wasn't enabled, didn't fire, or was dropped — so the responder rejects.

[...]

PR #3413, opened 2026-08-13
fix(node): let a stranded joiner pull from a beacon it cannot yet verify

chefsale:
Refs #3406. Follow-up to #3411, which did not work — and the logs say why.

## #3411 was a no-op

It widened `beacon_indicates_divergence` so a member holding no state could act on a divergence beacon. On the merge run (`c62b3c00a`) node-2 logged **zero** governance syncs and:

```
26x  ReadinessBeacon failed verification; dropping
```

The beacon never reaches that check. The widened predicate is unreachable on the node that needs it.

## The actual circularity

`verify_readiness_beacon` = valid signature **AND** `signer_is_namespace_member`. Membership is read from rows that a node with no governance state does not have. So a stranded joiner drops **every** beacon — including the ones advertising exactly the state that would end that condition.

**It cannot verify the message that would let it fetch what it needs in order to verify messages.**

A join whose direct request finds no reachable peer lands precisely there: membership recorded, no key, no ops, and then deaf to its only pull trigger.

## The fix, and its precedent

This circularity already has a documented escape hatch in the *other* direction. `beacon_admission_provable` lets an **established member** pull from a **joiner** it cannot recognise, justified as:

> *"This grants nothing… it only tells the caller that pulling governance from this peer is worth the round trip."*

`receiver_is_stranded_in_namespace` is the mirror, for the joiner side, on identical terms.

**"Stranded"** is this node's own namespace **identity** plus an **empty governance head**:

- the *identity* is written by the join before any key or state exists, so it separates "joined but stranded" from "not our namespace" — a stranger stays deaf;
- the *empty head* keeps the hatch shut for anyone who can already verify. An established member has state and keeps going through `verify_readiness_beacon` completely unchanged.

**Nothing is granted.** No membership row, no `ReadinessCache` entry, no trust in the beacon's contents. The signature is still checked, the pull is authenticated, the caller's debounce rate-limits it, and every op fetched still faces the full apply path. A node with no state risks nothing but a round trip.

This **composes with #3411** rather than replacing it: this lets the beacon through, that lets a stateless member act on it. Both are needed, which is why #3411 alone changed nothing.

## Tested deterministically, on purpose

The scenario is **timing-dependent** — it passed on master at `28db645bf` and failed at `c62b3c00a`. A green CI run would not prove a fix and a red one would not disprove it, so the behaviour is pinned in unit tests against a real store instead:

- a member with no governance head **is** stranded (the deadlock case);
- a member that holds state is **not** — verification is not widened for anyone who can already do it;
- a stranger to the namespace is **not** — no identity, still deaf.

`cargo test -p calimero-governance-store` 583 passed · `-p calimero-node --lib` 471 passed · `fmt --check` and `clippy -D warnings` clean.

CI on the scenario is still worth watching, but treat it as one sample, not as the verdict.

### DIFF
--- a/crates/governance-store/src/governance_broadcast.rs
+++ b/crates/governance-store/src/governance_broadcast.rs
@@ -223,6 +223,59 @@ pub fn verify_readiness_beacon(store: &Store, beacon: &SignedReadinessBeacon) ->
     signer_is_namespace_member(store, beacon.namespace_id, &beacon.peer_pubkey)
 }
 
+/// Whether THIS node is a member of `namespace_id` that holds no governance
+/// state yet — stranded, and unable to verify anyone.
+///
+/// The mirror of [`beacon_admission_provable`], and it exists for the mirrored
+/// circularity. That one lets an established member pull from a joiner it
+/// cannot yet recognise. This one lets a joiner pull from an established member
+/// it cannot yet recognise: [`verify_readiness_beacon`] requires the signer to
+/// be a known member, and a node that has applied no governance ops knows no
+/// members — so it drops every beacon, including the ones advertising exactly
+/// the state that would end that condition.
+///
+/// A join whose direct request found no reachable peer lands precisely here: it
+/// records membership, gets no key and no ops, and is then deaf to the only
+/// pull trigger it has.
+///
+/// Membership is read from this node's own namespace IDENTITY, not from the
+/// membership rows — the rows are the governance state it is missing, so
+/// requiring them would restate the circularity. The identity is written by the
+/// join before any of that exists, which is what makes it usable here and what
+/// keeps a stranger to the namespace out.
+///
+/// Grants nothing, exactly as its mirror does not: no membership row, no cache
+/// entry, no trust in the beacon's contents. It says only that pulling
+/// governance from this peer is worth a round trip. The beacon's signature is
+/// still checked by the caller, the pull is authenticated, the caller's
+/// debounce rate-limits it, and every op fetched still faces the full apply
+/// path. A node with no state risks nothing but the round trip.
+#[must_use]
+pub fn receiver_is_stranded_in_namespace(store: &Store, namespace_id: NamespaceId) -> bool {
+    let group_id = ContextGroupId::from(namespace_id.to_bytes());
+
+    // Ours to join at all? Absent identity ⇒ not our namespace ⇒ stay deaf.
+    if !matches!(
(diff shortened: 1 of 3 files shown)


chefsale:
## Do not merge — this regresses 27 app-migration scenarios

Back to draft. The evidence is unambiguous and it is my error:

| | app-migration |
| --- | --- |
| master (last 5 commits) | **all success** |
| #3415 (branched off master, no beacon change) | **pass** |
| this branch | **27 scenarios fail** |

Plus `Test (group-governance-peer-down)` and `Test (group-metadata)`. Builds succeeded, so it is not a cascade.

The failure is:

```
join_subgroup_inheritance failed on app-migration-newmethod-2
```

the documented signature of *"a node inherits membership recorded against its stand-in"* — the same symptom #3378 produced across 45 jobs.

## Why my reasoning was wrong

The guard I removed said, in its own words, that firing a governance sync before the key arrives leaves *"undecryptable opaque skeletons that the post-key join sync then skips as duplicates"*. I argued that hazard was already closed, because `sync_namespace_from_peer` now ends with `recover_missing_group_keys` and `drain_governance_pending_after_sync` (#2613).

**That was not sufficient.** Letting an unverifiable beacon trigger a pull opens the early-governance-pull path in a case the #2613 recovery does not cover: membership folds against a stand-in before the node's real credential is established. The guard was load-bearing for more than the case its own comment described — I treated the comment as an exhaustive statement of its purpose, and it was not.

## What this means for #3406

The stranded-joiner diagnosis still stands: node-2 drops every beacon (`26x ReadinessBeacon failed verification; dropping`) and therefore cannot pull. But **relaxing beacon verification is the wrong lever.** This is the second lever on that deadlock I have tried and had to withdraw — #3411 was a no-op, and this one is actively harmful.

Of the three links, the remaining candidate is the one I have not touched: the namespace fallback declining to dial a peer that is not subscribed to the context topic (`peers.rs:186`, guarding #2422's wasted dials). That is where governance and key delivery are gated on a *context-plane* condition, which looks like the wrong condition rather than a load-bearing guard.

Leaving this closed and re-approaching from that side is better than iterating further here.

chefsale:
Closing rather than leaving as a draft. The approach is wrong, not incomplete — relaxing beacon verification opens the early-governance-pull path that the guard exists to prevent, which is what the 27 app-migration failures were. Leaving it open would invite someone to "finish" it.

The diagnosis it carries is still good and is preserved on #3406, along with a design proposal for the lever that the regression actually points at.

PR #3423, opened 2026-08-14
fix(api): return the caller's ACCOUNT so it can find itself in a member list

### LINKED ISSUE #3402: list_group_members returns members keyed by account but self_identity as a signing key, so clients can no longer identify themselves in the list
`ListGroupMembersResponse` (`crates/context/primitives/src/group.rs:161`) mixes two identity spaces in one response:

```rust
pub struct ListGroupMembersResponse {
    pub members: Vec<GroupMemberEntry>,
    /// The node's own group-level identity (SignerId) so the client knows
    /// which member in the list represents the current node.
    pub self_identity: PublicKey,
}

pub struct GroupMemberEntry {
    /// The member's ACCOUNT — the principal every governance row is keyed by.
    ///
    /// It used to be a signing key. Listing members by key is no longer possible:
    /// the rows name accounts, an account is a one-way hash, and a person can
    /// hold several keys.
    pub identity: AccountId,
    ...
}
```

`self_identity`'s doc comment states its purpose — "so the client knows which member in the list represents the current node" — and that no longer works: it is a `PublicKey`, the entries are `AccountId`s, and the mapping is a one-way hash. `members.iter().any(|m| m.identity == self_identity)` is now always false.

This is not hypothetical for clients. `mero-chat` uses exactly that comparison for:

- `isDirectMember` — am I a member of this subgroup (drives which channels render as joined)
- the role lookup that gates the admin UI and permission checks

Both fail closed, so an **admin reads as a plain user** and a joined channel reads as not-joined.

### Confirmed against a live node

merod `0.11.0-rc.20` still returns signing keys in both fields, so the comparison works today:

```
GET /admin-api/groups/{ns}/members
  members[0].identity = 67dkUZXFveMm2nTCsE4JzKj2Fe5qenYtH7pchXLNSLZH  (Admin)
  selfIdentity        = 67dkUZXFveMm2nTCsE4JzKj2Fe5qenYtH7pchXLNSLZH
```

On `master` the entries are `AccountId`, so the first merod carrying this change silently breaks every client doing the documented thing. (`mero-js` 9.0.1 already documents `identity` as "the member's ACCOUNT (64 hex), as returned by `listGroupMembers`", so the SDK expects the new shape.)

### Suggested fix

Return the node's **account** as `self_identity` (renaming to `self_account` if the wire break is acceptable), or add it alongside so both spaces are available. `list_group_members` already resolves it — the membership gate a few lines up calls `member_account::require(&self.datastore, &group_id, &node_identity)`.

For reference, on an un-enrolled node the account is the derived stand-in and clients can compute it, but only if they know the scheme — verified against the same node:

```
account = SHA256( u64_le(len(domain)) ‖ "calimero.op-adapter.legacy-account.v1" ‖ pubkey )
        = 6deb0c0c…61c8   ← matches GET /namespaces/{ns}/account
```

Clients should not have to reimplement that to answer "which one is me".

chefsale:
Closes #3402. This is fallout from #3391 — my change — and it is client-facing.

## The bug

`ListGroupMembersResponse` mixes two id spaces. `members[].identity` became an `AccountId` when governance principals flipped to accounts; `self_identity` stayed a signing key. So the documented way to answer *"which of these is me"*:

```rust
members.iter().any(|m| m.identity == self_identity)
```

is now **permanently false**. An account is a one-way hash of a genesis; no amount of key-matching recovers it.

**It fails closed and silently, which is the damaging part.** In `mero-chat` that same comparison drives `isDirectMember` and the role lookup behind the admin UI — so an admin renders as a plain user and a joined channel renders as unjoined, with nothing in the response suggesting anything is wrong.

## The fix

`self_account: AccountId` is added **alongside** `self_identity`, not in place of it:

- **`selfAccount`** — the field to match against entries. Documented as such.
- **`selfIdentity`** — kept, and re-documented as *not comparable* to `members[].identity`. A caller that needs the key it signs with has no other source for it; removing it would relocate the breakage rather than end it.

The handler already computed this value via `member_account::require`, but only on the branch that needed it for the membership gate. It is now resolved unconditionally, because it is the answer to the question the response exists to answer.

## On the test

`list_group_members_lets_the_caller_find_itself` pins the wire field and the comparison **together**, since the two halves are only useful as a pair.

It is A/B verified: removing `selfAccount` from the mock response makes it fail; restoring it makes it pass. That check mattered here — a test asserting "`Some(x)` was found among the members" can pass while proving nothing, which is close to how the original bug survived.

## Compatibility

Additive on the wire (`skip_serializing_if = "Option::is_none"`), so an existing client keeps deserializing. It does not *fix* such a client — one still comparing `selfIdentity` against entries stays broken — but that client is broken today, and `mero-js` 9.0.1 already documents `identity` as the account.

## Verification

16 suites green across `calimero-client`, `calimero-context` and `calimero-server`; `clippy -D warnings` and `fmt --check` clean.

### DIFF
--- a/crates/client/src/tests.rs
+++ b/crates/client/src/tests.rs
@@ -180,6 +180,44 @@ async fn list_group_members() {
     assert!(resp.members.is_empty());
 }
 
+/// **A caller can find itself in the listing.**
+///
+/// `members[].identity` is an ACCOUNT and `selfIdentity` is a signing KEY, so
+/// the documented way to answer "which of these is me" —
+/// `members.any(|m| m.identity == self_identity)` — became permanently false
+/// when members started being named by account. Downstream that reads as an
+/// admin rendering as a plain user and a joined channel rendering as unjoined:
+/// both gates fail closed, silently.
+///
+/// `selfAccount` is the answer, and it has to arrive in the same id space the
+/// entries use. This pins the wire field and the comparison together, because
+/// the two halves are only useful as a pair.
+#[tokio::test]
+async fn list_group_members_lets_the_caller_find_itself() {
+    let server = MockServer::start().await;
+    Mock::given(method("GET"))
+        .and(path(format!("/admin-api/groups/{GID}/members")))
+        .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
+            "members": [
+                { "identity": ZERO_HEX_ACCOUNT, "role": "Admin" },
+            ],
+            "selfAccount": ZERO_HEX_ACCOUNT,
+            "selfIdentity": "11111111111111111111111111111111",
+        })))
+        .expect(1)
+        .mount(&server)
+        .await;
+
+    let client = make_client(&Url::parse(&server.uri()).unwrap());
+    let resp = client.list_group_members(GID).await.unwrap();
+
+    let me = resp.self_account.expect("the node reports its own account");
+    assert!(
+        resp.members.iter().any(|m| m.identity == me),
+        "a node must be able to find itself among the members it is listed in"
+    );
+}
+
 #[tokio::test]
 async fn add_group_members() {
     let server = MockServer::start().await;
--- a/crates/context/primitives/src/group.rs
+++ b/crates/context/primitives/src/group.rs
@@ -161,8 +161,17 @@ impl Message for ListGroupMembersRequest {
 #[derive(Clone, Debug)]
 pub struct ListGroupMembersResponse {
     pub members: Vec<GroupMemberEntry>,
-    /// The node's own group-level identity (SignerId) so the client knows
-    /// which member in the list represents the current node.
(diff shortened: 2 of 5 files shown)


chefsale:
Filed #3425 for the underlying problem.

Review feedback on this PR was that adding `selfAccount` to the listing treats the symptom: the real defect is that a collection response carries the caller's identity at all, and that three ids are spread across four overlapping endpoints with two encodings. #3425 removes `selfIdentity` rather than adding a second self-field beside it — so the field this PR adds is one that work deletes.

Holding this open pending a call on sequencing: merge as an interim unblock, or close in favour of #3425. Leaning toward closing — there is no running network to protect, so there is no need for the interim field.

chefsale:
Closing in favour of #3425 — the field this adds is one that work deletes.

Issue #3430, opened 2026-08-14
Offline account root: enable a device by imported certificate, with no root on the node


chefsale:
## Summary

An account root can be exported and imported as a recovery phrase, but it cannot be
kept *offline*: no node can operate under a certificate its root signed elsewhere.
The account/device split is designed so the root signs certificates and nothing
else, and that posture is still unreachable — but for one remaining reason, not the
three this issue was filed with.

### What already works

- `merod account export` / `import` — the root travels as a 24-word BIP-39 phrase.
- `merod account revoke-proof` — sign a device revocation offline, published by any
  node.
- `pair-init` needs no root secret: it takes the root **public** key, mints the
  device id and both keypairs locally, and self-signs.
- **`merod account sign-cert` (shipped).** Takes `--device` / `--sign-pk` /
  `--kem-pk`, or `--generate` to mint and certify in one step, and `--from PATH`
  reads the root from a phrase and *opens no datastore at all* — so it runs on a
  machine with no node. This is the offline signer criterion 1 asked for.
- **A node with no root refuses rather than minting (#3664).** `ensure_account_root`
  is split into `require_account_root` (reads, errors) and `provision_account_root`
  (mints, called only from `merod init`). Nothing mints lazily from any read, and
  `merod init --no-account-root` makes "this node holds no root, by configuration" a
  real state. A root-free node can already *name* its account, because a certified
  device row answers before the root fallback.

### What is missing — one thing, in two halves

**A certificate minted elsewhere cannot be brought onto a node.**

1. **Nothing persists a holder-signed `DeviceCert`.** `try_import_account_root`
   imports a *root*; there is no row for "a cert some other root signed for this
   device", and no command to supply one. Certificates still arrive only via a live
   `pair-complete` (`crates/context/src/handlers/pair_device_complete.rs:89`).
2. **`join_credential::build` signs the certificate it presents.** It calls
   `require_account_root` (`crates/context/src/join_credential.rs:63`) and then
   `DeviceCert::sign`, so on a root-free node it errors. It needs a branch that
   presents the *stored* cert when there is no root.

Until both land, `--no-account-root` is reachable only via pairing: the node gets a
device row from an account whose root lives elsewhere, but cannot build a join
credential on its own.

## Impact

An operator still cannot run the posture the identity model was designed around:
root in cold storage, devices enabled by signed certificates. Since #3664 the root
is no longer silently minted on every node, so the failure is now an honest error
rather than a node quietly speaking as a different account — but the working
[...]

Issue #3428, opened 2026-08-14
One account, one device per node: collapse per-namespace identity


chefsale:
## Summary

A node's identity is per-namespace. Joining five namespaces mints five accounts,
five devices, five signing keys, five KEM keys and five `sender_key`s, all for one
installation. The model is coherent — it buys cross-namespace unlinkability — but
that property is not one we want, and everything expensive about the identity layer
exists to provide it.

This proposes collapsing to **one account and one device per node**, globally.

### What exists today

| thing | scope | how it is derived |
| --- | --- | --- |
| account root secret | node | `PrivateKey::random`, once (`node_device.rs:341`), mnemonic-backed |
| account nonce | **per namespace** | `H("…nonce.v1" ‖ root_secret ‖ namespace_id)[..16]` (`account/src/lib.rs:137`) |
| `AccountId` | **per namespace** | `H("…genesis.v1" ‖ borsh{version, root_pk, nonce})` |
| `DeviceId` | **per namespace** | `H("…device.id.v1" ‖ account ‖ random16)` (`node_device.rs:718`) |
| device signing key | **per namespace** | `PrivateKey::random(OsRng)` (`namespace/core.rs:665`) |
| device KEM key | **per namespace** | `X25519SecretKey::random` at enrolment |
| `sender_key` | **per namespace** | `PrivateKey::random(OsRng)`, alongside the signing key |

The per-namespace account is deliberate and documented: deriving the nonce from the
root **secret** is what keeps one person's accounts in different namespaces from
being correlated (`account/src/lib.rs:126-135`).

### Proposed

```
root secret   random once, mnemonic          the only thing to back up
account       H("…genesis.v1" ‖ {version, root_pk})
device id     H("…device.id.v1" ‖ account ‖ random16)
device keys   generated on the device        sign + KEM
per namespace membership rows only
```

Five things, one derivation, no per-namespace identity.

### What this removes

- `derive_account_nonce` and the `nonce` field of `AccountGenesis`. The nonce's two
  stated purposes — two accounts from one root, and an account id unpredictable from
  a public key — are both unlinkability, and both are being given up by decision.
- Per-namespace signing keys, KEM keys and `sender_key`: all of
  `NamespaceRepository::get_or_create_identity`.
- `NodeDeviceRepository` keyed by namespace collapses to a single row.
- "one node holds N devices", which is the single hardest thing to explain about the
  current model.
- The namespace path parameter on the identity endpoint (see #3425): the surface
[...]

PR #3431, opened 2026-08-14
feat(account)!: one root key is one account, everywhere

chefsale:
Implements #3428. **One root key → one account → N devices. Namespaces hold membership rows and nothing else.**

This started as "slice 1, the account only" and grew to the whole identity collapse, because the pieces turned out to be coupled — see *Why this is one PR* below. The earlier description said devices stayed per-namespace; that was true of the first commit and is no longer true of the PR. Two review findings caught the mismatch, correctly.

## Why

An account was per-namespace: the genesis carried `nonce = KDF(root_secret, namespace_id)`, so one person held a different `AccountId` in each namespace. That bought exactly one property — a member of two namespaces could not tell the two accounts were one person — and the property does not hold, in two independent ways:

**The genesis publishes the invariant.** `AccountDeviceLinked` carries the `AccountGenesis`, which it must, since `H(genesis) == account` is what makes the account self-certifying. Across namespaces only the nonce changed; `root_sign_pk` is the same 32 bytes. A member of both replicates both DAGs and compares two numbers.

**The PeerId links them again.** Every node keeps a persistent `member identity → PeerId` map per group (`peer_identity_cache.rs`) so peer selection can dial real members, and a node has one libp2p PeerId for life.

Restoring the property would need separate root secrets, PeerIds, network paths and timing — two nodes, not a protocol change. So this is a simplification with nothing lost. Full analysis, reversibility, and the roadmap tension with #3430 are on #3428.

## What changed

| | before | after |
| --- | --- | --- |
| account | one per namespace, salted | `H(version ‖ root_pk)`, one per root |
| device | one per namespace | one per node (`NodeDeviceIdentity` singleton) |
| signing / KEM key | one per namespace | one per node (`NodeIdentity` singleton, prefix `0x48`) |
| `sender_key` | one per namespace | **deleted** — see below |
| namespace row | held the keypair | participation marker, no key material |

`ACCOUNT_GENESIS_VERSION` → 2: removing the field already changes the preimage, and a version that moves with it is what makes a v1 and a v2 id from one root provably different rather than two encodings a reader might reconcile.

**The account stays a hash, not the root public key itself.** That keeps the root rotatable, and keeps an account visually distinct from a key — an account rendering like a signing key would revive the id-space confusion behind #3402.

## Why this is one PR

The device and its signing key could not be separated. `create_account.rs` records the namespace identity as the device's `sign_pk`, because "recording anything else would make `sign_pk` a claim no signature ever matches". A global device with per-namespace keys would be a certificate naming a key that only sometimes signs.

## Please review these specifically

**1. Device revocation is now global.** `authorises(account, device)` was implicitly namespace-scoped *only* because the account differed per namespace. A proof minted while pointed at one namespace now verifies anywhere. That reads as correct — a stolen device is stolen everywhere, only the account root can mint the proof, and it names that root's own device — and publication is still per-DAG, so nothing propagates on its own.

**2. Forward secrecy on eviction changes shape.** A node-level agreement secret cannot be dropped for one namespace. The property now rests on deleting the group keyring and the namespace's ops, plus the rotation that removal triggers; what is lost is the defence-in-depth case of ciphertext retained *outside* the keyring. Recovering the strong version means rotating the device's keys rather than deleting them (`device_epoch` exists for it), which needs the account root — tracked on #3430. The reasoning is recorded at the purge site.

**3. `sender_key` is deleted, not made global.** It was a third keypair minted on every namespace join, persisted, redacted in `Debug`, threaded through every identity tuple — and read by nothing: all five production call sites bound it as `_sender` and dropped it. The one path that would have used it, a context with no group, has hard-errored for some time. Group encryption is untouched; `GroupKeyring`, scope keys, delivery and rotation are unchanged and still per group. The local at the encrypt site was *called* `sender_key` while holding the group key — that is now `encryption_key`, as is the same misnomer on `NodeClient::broadcast`, which `encryption.mdx` carried a standing warning about.

## On-disk format: a deliberate flag day

Account ids, device ids, key layouts and every op carrying a genesis all change. There is no migration and none is intended — #3428 records the reasoning, and it rests on there being no running network, which is worth confirming at merge rather than assuming.

Reviewers raised the risk that a node with pre-PR state silently re-identifies. **It fails loudly instead.** The old `NamespaceIdentityValue` is 96 bytes and the new one is 1, and borsh refuses trailing bytes — the same `Not all bytes read` failure the wire goldens in this PR produced — so the first identity lookup on legacy data errors rather than misparsing. A legacy node refuses; it does not quietly become a stranger.

## Bugs this surfaced

- **`get_or_create_identity_bundle` returned early once a key existed.** With the key node-level that fires from the *second* namespace onward, so the participation marker would never have been written for any namespace after the first — the node would have silently stopped syncing every namespace but the one it joined first. Regression test: `joining_a_second_namespace_is_still_recorded`.
- **`identity()`/`identity_record()` were read as "am I in this namespace"** by callers deciding whether to emit a readiness beacon, self-report a migration, or treat a self-purge as done. Same question while keys were per-namespace; not any more. Rather than gate 52 call sites, the repository checks participation before returning the key, so `None` keeps meaning what every caller already assumed.
- **`store_identity` could clobber the node's key.** Now refuses a mismatched one; `replace_identity` covers deliberate re-provisioning. The guard caught two fixtures on its first run.

## A bug class removed

`account_for_context` fell back to scoping the account to the CONTEXT when the context→group row had not landed, so `init` seeded a writer set under one account and every later call presented another — the creator locked out of what it had just created. An account is no longer scoped to anything, so the fallback cannot name a stranger. The test that pinned the divergence said in its own comment that a change making the fallback safe should replace it deliberately; `the_creation_time_resolver_trap_is_gone` is that replacement.

## Known dead field

`accountNonce` is returned as 32 zeros. `calimero-client-py` is a Rust binding that deserializes into this struct, compiled into the released wheel while the field was required, and merobox pins that wheel — so omitting it fails to parse there and surfaces as a bare "account create failed" with no node-side error. Removed once client-py ships without it: calimero-network/calimero-client-py#82.

Note the asymmetry, which is what made this easy to miss: an unknown field in a **request** is ignored, a missing field in a **response** is fatal to a typed client.

## Verification

`cargo fmt --check`, `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`, and `cargo test --workspace` pass. Six wire goldens re-blessed, each validated by requiring the result to decode *and* re-encode identically rather than transcribed. Docs corrected across `identities.mdx`, `glossary.mdx`, `accounts.mdx`, and two AGENTS.md files, which still described identity as per-namespace.


chefsale:
Worked through all eight. Summary of what changed and what I'm answering rather than fixing.

**Fixed**

- `store_identity` clobbering — real, and the best finding here. It now refuses a mismatched key and names both; `replace_identity` covers deliberate re-provisioning. The guard caught **two fixtures on its first run** doing exactly what was predicted: one seeded a second namespace with a second key (no longer a thing the model permits — it reuses the node's key), the other repoints the identity on purpose and now says so.
- `AccountGenesis` row in `crates/account/AGENTS.md` still listing `nonce`.
- `NODE_DEVICE_IDENTITY_PREFIX` doc still describing a per-namespace row.
- `ImportedRoot::released/retained` docs still reading as one-per-namespace.

**Scope vs. description — correct, and the description was the thing that was wrong.** This began as slice 1 and grew to the whole collapse, deliberately: the device and its signing key cannot be separated, because `create_account.rs` records the namespace identity *as* the device's `sign_pk` ("recording anything else would make `sign_pk` a claim no signature ever matches"). A global device with per-namespace keys would be a certificate naming a key that only sometimes signs. The description is rewritten to match, with the signing-key unification and the KEM-secret sharing called out as reviewed decisions alongside the revocation change.

**Migration — deliberate, and the failure is loud rather than silent.** No migration is intended; the flag day and its reasoning are on #3428, and it rests on there being no running network (worth confirming at merge, not assuming). On the specific risk that a pre-PR node silently re-identifies: the old `NamespaceIdentityValue` is 96 bytes, the new one is 1, and borsh refuses trailing bytes — the same `Not all bytes read` failure the wire goldens in this very PR produced. The first identity lookup on legacy data errors; it does not misparse. A legacy node refuses rather than quietly becoming a stranger.

The orphaned-device-row path is real but unreachable in practice for the same reason: every flow that could mint a replacement device resolves the namespace identity first and dies there.

**Forward secrecy** was already flagged for review before this pass and is unchanged: a node-level agreement secret cannot be dropped for one namespace, so the property rests on the keyring delete, the op delete, and the rotation that removal triggers. Restoring the defence-in-depth layer means rotating device keys rather than deleting them — `device_epoch` exists for it, it needs the account root, and it is tracked on #3430.

chefsale:
Second pass. Three of these landed on an older commit and are already addressed; three were new and are fixed; one I want to argue about rather than silently do.

**Already fixed before this batch ran** — `store_identity` clobbering (now refuses a mismatched key, with `replace_identity` for deliberate re-provisioning), and the two scope-vs-description findings (the description was the thing that was wrong and is rewritten; the device/key collapse is deliberate and now called out as a reviewed decision, because `create_account` records the namespace identity *as* the device's `sign_pk` and the two cannot be separated).

**New and fixed.** Three comments that stated the removed property as fact rather than just reading oddly:

- `AccountRoot` claimed per-namespace account ids "stay distinct because the nonce is derived per namespace, which is what lets recovery and unlinkability hold at once". Both halves are now false.
- `pair_device_complete` claimed the genesis is "derived from the root secret and the namespace id". Its conclusion — this node can only certify devices for the account it owns — still holds, for a simpler reason.
- `namespaces_to_sync` described its set as namespaces the node "holds an identity in". Holding an identity is no longer the same question as taking part, which is exactly the distinction the participation row exists to keep.

**The naming point I'd rather do separately.** You are right that `NamespaceIdentity` no longer holds an identity, and a type name that lies is the same failure mode as the `sender_key` misnomer this PR already fixed in two places. But there are **two different types with that name** — the store key (now participation) and `context-primitives`'s `NamespaceIdentity`, which is the API response and genuinely still an identity — and they appear in overlapping files. A 55-site rename across two same-named types is precisely the change that goes subtly wrong, and it is a clarity nit rather than a correctness bug, so I would rather do it as its own compiler-driven pass than tack it onto a PR this size. Renaming the store key to `NamespaceParticipation` also disambiguates the two, which makes it worth doing properly.

**Migration** is answered in the previous comment and unchanged: deliberate flag day, and the failure on legacy data is loud rather than silent — 96-byte values decoding into a 1-byte struct hit borsh's `Not all bytes read`, the same failure the wire goldens in this PR produced.

chefsale:
Third pass. Both new findings are the two consequences I flagged for review in the description, so this is an answer rather than a fix — but the second one deserves more than "already noted".

**Forward secrecy on eviction** (`local_state.rs`). Correct, and deliberate: a node-level agreement secret cannot be dropped for one namespace without evicting the node from every other one at the same moment. What still happens on teardown is the group keyring delete, the namespace's op delete, and the rotation that removal triggers — so the wrapped scope keys this secret could open are gone and nothing it could unwrap stays current. What is genuinely lost is the defence-in-depth case: ciphertext retained *outside* the keyring. Restoring the strong version means rotating the device's keys rather than deleting them, which `DeviceCert::device_epoch` exists for and which needs the account root — tracked on #3430. The reasoning is recorded at the purge site rather than only here.

**Revocation blast radius** (`governance.rs`). Also correct, and this one is a decision rather than a side effect — #3428 records it. Three things bound it:

- Only the account root can mint the proof, and it names that root's own device. A leaked proof therefore does exactly what its author asked, in more places.
- A proof is not a credential and never was: the docs already say it "authorises exactly one revocation of one device", and publishing it needs no authority, which is what makes an offline root workable at all.
- Publication is still per-DAG. Wider *validity* is not wider *reach* — the proof takes effect in a group when someone publishes it there.

What genuinely changes is that "revoke this device from one namespace only" stops being expressible. That is the intended reading: a device belongs to an account, not to a scope, so a revoked device is revoked wherever that account speaks. The alternative — a device that is spent in one namespace and trusted in another — is the state the account model exists to remove.

Also fixed since the last pass: `account-device-revoke-lockout` was failing in its script step on a `None` argument, which was the captured account nonce. The scenarios no longer thread it; the value is dead, and where merobox's step still declares the field required they pass a literal that reads as the stub it is.

chefsale:
Correcting my previous answer on forward secrecy — I had this wrong, and checking it turned up a real gap.

**Rotation is the mechanism, as you said.** `group_keys.rs` excludes the departed by ACCOUNT: *"a caller excluding someone (a rotation dropping the departed) excludes every device they hold rather than only the one that happened to be listed."* Account-scoped, so whether the device is per-namespace or node-level makes no difference to it. Nothing in this PR touches it.

**And the deletion I removed was never forward secrecy.** `delete_namespace_local_state` runs on the node's *own* store — `self_purge` after eviction, or `delete_namespace`. It is a node deleting its own secret, which is no defence against that node: one that wants to keep the secret simply does not purge, and one that would abuse it kept the unwrapped group key too. Calling it a defence-in-depth loss overstated it.

What it actually bought was hygiene: an installation that has left should not sit on key material.

**Which exposed the real gap.** Nothing deleted the device row any more, so a node leaving *every* namespace kept its agreement secret forever — and `NodeDeviceRepository::delete` had no callers left at all, so it was also dead code.

Fixed in 8446db3: the device goes when the **last** participation does. Leaving one namespace still cannot take it, because every namespace that remains opens its scope keys with the same secret. `the_device_survives_leaving_one_namespace_and_goes_with_the_last` covers both halves, and the comment at the purge site now says hygiene rather than forward secrecy.

The revocation blast-radius answer stands unchanged — that one is a decision recorded on #3428, not a mistaken claim.

PR #3441, opened 2026-08-14
chore: drop per-context identity aliases

chefsale:
## Summary

A node now uses one member identity across its contexts. Per-context identity aliases described the older model, where a node could hold several member keypairs per context and callers picked one — so naming identities per context no longer refers to anything real.

Removes:

- the `context identity alias` command tree (`add` / `remove` / `get` / `list`) and `context identity use`
- `call --substitute` and the `{alias}`-in-payload templating behind it (`substitute_aliases_in_payload`)
- `--as` on `context create` (created an alias for the new identity) and on `context update`
- the four `/admin-api/alias/{create,lookup,delete,list}/identity/...` routes
- `ExecutionRequest.substitute` from the JSON-RPC wire, and `ExecuteRequest.aliases` / `ExecuteError::AliasResolutionFailed`
- `Client::create_context_identity_alias`

**Context and application aliases are untouched** — those are node-global and unaffected by the identity model.

## The one part that is not a deletion

`context update --as` fed `authorize_update_application`, the check that gates a context bytecode swap and whole-context state migration. Its comment records why it exists:

> Previously the caller `public_key` was threaded straight through to the migration/finalize path and only the new application's signer *continuity* was checked, never the requester, so any key could drive a migration.

The flag is gone; the gate is not. `meroctl context update` now resolves the node's owned identity for the context and passes that, mirroring how `execute_request` already auto-resolves its executor. `executor_public_key` stays on the wire.

Moving that resolution server-side — so `UpdateContextApplicationRequest` drops the field entirely and the two paths converge — is the cleaner end state, but it is a separate wire change and is deliberately not in this PR.

## Knock-on dead code

Removing `substitute` orphaned three more items, each left with only its own definition and no consumer: `MAX_SUBSTITUTE_ALIASES`, `validate_collection_size`, and `ValidationError::TooManyItems`. They are removed here rather than left behind as a fresh unreachable island.

## Wire change

`ExecutionRequest.substitute` is a JSON-RPC wire field, so `fixtures/wire/jsonrpc/execute.req.json` is regenerated (`UPDATE_FIXTURES=1 cargo test -p calimero-server-primitives --test wire_fixtures`; all 6 fixture tests pass).

No SDK consumed it:

- `mero-js` lists it in `ignoredCoreKeys: ['substitute']` in its wire contract test — **follow-up: that entry should be dropped in a matching mero-js PR**, since the field no longer exists
- `calimero-client-py` only ever passed `vec![]`
- `merobox`, `admin-dashboard`, `tauri-app`, `mero-react`, `swift-sdk`: no references

No merobox scenario, workflow or script referenced identity aliases or `--substitute`, so the e2e suite is unaffected.

## Docs

`docs/protocol/identities.mdx` still documented the superseded model ("A node is not limited to one member key per context… it can hold **several**"). Its executor section is rewritten to describe what the code does: the executor is resolved, never named, and the known caller-vs-executor privilege consequence is stated. The four route rows in `admin-api.mdx` and the alias / `--substitute` entries in `meroctl.mdx` are updated.

## Verification

- `cargo check --workspace --all-targets` ✅
- `cargo fmt --check` ✅
- `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` ✅
- `cargo test -p calimero-server-primitives --test wire_fixtures` ✅ 6/6
- full `cargo test --workspace` was still running locally when this was opened; CI is the gate

Conflicts with #3440 on `cli/context/identity.rs` and `operate/meroctl.mdx` — whichever lands second needs a trivial rebase.


chefsale:
Full local test run has since completed on the rebased branch: **4,759 passed, 0 failed** across 227 test binaries (`cargo test --workspace`, exit 0), alongside `fmt`, `clippy -D warnings` and the 6 wire-fixture tests. Rebased onto current master (post-#3437).

chefsale:
sdk-ref: chore/drop-identity-aliases

**Paired with calimero-network/mero-js#100.**

The first CI run caught something the audit missed: mero-js has all four `*ContextIdentityAlias` methods on `AdminApiClient`, and its e2e wire-contract test calls `createContextIdentityAlias`. Removing the routes 404'd it:

```
HTTPError: HTTP 404 Not Found
url: 'http://localhost:2428/admin-api/alias/create/identity/D81ocJgt2ZPapHGkcG3gQtskDM5k9TrEuCdbtpmVhshG'
❯ AdminApiClient.createContextIdentityAlias src/admin-api/admin-client.ts:626:7
❯ tests/e2e/wire.contract.test.ts:181:5
```

So the routes did have a live external consumer. mero-js#100 removes the four methods, their types, their unit tests and the e2e coverage; the branch is named identically so `resolve-paired-ref.sh` picks it up and the two go green together.

Verified on the mero-js side: `pnpm typecheck` clean, 276 unit tests passing, and `pnpm test:contract` run against **this** branch's fixtures passes 5/5 with `substitute` removed.

Rebased onto current master (post-#3440/#3442/#3443). The `Test (group-join-mesh-not-ready)` failure is pre-existing on master and unrelated — it also failed on #3442, which is a scripts-only PR with no Rust.

PR #3453, opened 2026-08-15
fix(node): pull governance and keys when a namespace peer subscribes

chefsale:
Draft: the change is complete, but the scenario is too flaky for a single green run to mean anything. Holding until repeated-run data is in (see **Validation**).

Addresses #3406 (`group-join-mesh-not-ready`).

## The deadlock

The scenario partitions node-2, joins (degrades by design), heals, and then expects governance to converge **unaided** before the explicit re-join. That is the step that times out at 120s.

Post-heal the transport is **fine** — connection re-established, subscriptions re-learned, GRAFT completed at `08:07:15`, ~2 minutes of budget left. The node stays stuck because every recovery it needs is triggered by something its state denies it:

| Recovery | Trigger | Why it can't fire |
|---|---|---|
| `sync_namespace_from_peer` (governance) | peer's `ReadinessBeacon` advertising an unknown head | `verify_readiness_beacon` requires the signer to be a KNOWN member; knowing members requires the governance state the join never delivered → **26 beacons dropped, all post-heal** |
| `recover_missing_group_keys` (key) | *receiving* a gossiped Group op | needs a populated mesh, which a stranded joiner lacks |
| namespace heartbeat | — | liveness-only by design; observes `we_missing=1 peer_missing=0` 4x, remediates nothing |
| `admission_proof` / `beacon_admission_provable` | a `MemberJoinedAt` that reached publish-with-zero-acks | different scenario — verified **zero** occurrences on both nodes |

Verifying a beacon needs member knowledge; member knowledge arrives via the pull that beacon would have triggered.

## The fix

Add a trigger rather than relax a guard.

A peer subscribing to `ns/<id>` is **our own gossipsub** reporting a reachable peer — not a peer-supplied claim — so acting on it grants nothing that was not already granted. On that event, a node that is a member holding no key pulls governance and keys from that peer.

The `group/` branch of `handle_subscribed` already drives active recovery on subscription (`sync_group` + `broadcast_group_local_state`); the `ns/` branch only emitted a beacon. This brings `ns/` in line.

Gated on `namespace_groups_member_but_keyless` being non-empty:
* **bounds the work** — a busy namespace raises one event per peer, and the beacon path this parallels is debounced where this is not;
* **self-limiting** — the predicate goes false the moment recovery lands, so a healthy node never pulls from here and a recovering one stops on its own.

A member that holds its key but is merely behind is left to the beacon path, which it can verify precisely because it is not stranded.

## Why not the previous approaches

Both prior attempts tried to break the circle by loosening the guard:
* **#3411** (merged, `c62b3c00a`) widened `beacon_indicates_divergence` — a no-op here; the beacon is dropped at verification before the predicate runs.
* **#3413** (drafted, not merged) relaxed beacon verification for a stranded joiner — **broke 27 app-migration scenarios**.

This PR touches neither beacon verification nor join authorization. Relaxing the beacon guard widens what a stranger can assert about membership; adding a trigger does not.

## Validation

**Not yet demonstrated, deliberately.** `group-join-mesh-not-ready` failed **4 of 9 recent master runs (~44%)**, so one green run is not evidence. Two things must happen before this leaves draft:

1. **Re-measure the baseline on post-#3449 master.** That 44% was measured *before* #3449, and this scenario's logs also showed the root-hash rollback that #3449 fixed. Some of its flakiness may already be gone, and attributing that to this PR would be wrong.
2. **Repeated runs on this branch** — roughly 8-10 consecutive passes, reported as counts against the re-measured baseline.

`cargo fmt`, `cargo check`, and `cargo clippy -p calimero-node --all-targets --features calimero-storage/testing -- -D warnings` are clean.

No unit test: the deadlock is a multi-node partition/heal interaction between gossipsub mesh state, beacon verification, and the governance pull. The merobox scenario in this repo is the regression test; the unit suites go green with or without the bug.

---

## Evidence

Raw pass/fail is the wrong metric. The test partitions *before* the join, so **every** run takes the degraded branch; what varies is whether the joiner then wedges. Runs are classified from the joiner's node log:

* `dropped` — `ReadinessBeacon failed verification` count. **26** = wedged for the full 120s window.
* `ran` — `stranded-member recovery:` lines, i.e. this PR's path executing.

### master @ post-#3449 (7 runs)

| run | result | dropped | verdict |
|---|---|---|---|
| 31875113600 | success | 0 | never wedged |
| 31875531290 | failure | 26 | **WEDGED** |
| 31876353885 | failure | 26 | **WEDGED** |
| 31877044678 | failure | 26 | **WEDGED** |
| 31877740476 | success | 0 | never wedged |
| 31878424752 | failure | 26 | **WEDGED** |
| 31879092762 | success | 1 | never wedged |

**4 of 7 wedge. Wedging is terminal — 4 for 4 never recover.**

### this branch, round 2 (7 runs, unconditional logging)

| run | result | dropped | ran | verdict |
|---|---|---|---|---|
| 31880131688 | success | 0 | 0 | never hit the failure mode |
| 31880488666 | success | 1 | **2** | **rescued** after first drop |
| 31881078454 | success | 0 | 0 | never hit the failure mode |
| 31881613527 | success | 0 | 0 | never hit the failure mode |
| 31882177987 | success | 0 | **2** | **rescued** preemptively |
| 31882711694 | success | 0 | **2** | **rescued** preemptively |
| 31883210505 | success | 0 | **2** | **rescued** preemptively |

### Reading

**The recovery fires in 4 of 7 runs — the same rate at which master wedges (4 of 7) — and converges in every one.** That the fire rate matches the wedge rate is the substantive result: the path is engaging in the population of runs that would otherwise have gone terminal, not at random.

Across **15 runs on this branch, zero reached the wedged state**, versus 4 of 7 on master. Fisher exact, one-tailed **p ≈ 0.005**.

Two shapes of rescue are observed, both converging:
* `dropped=1, ran` — the joiner drops its first beacon post-heal, the trigger fires, it converges at **1** drop instead of 26.
* `dropped=0, ran` — the trigger fires on subscription and recovers the node *before* it ever fails to verify a beacon.

### Notes on the evidence

* Round 1 (8 runs, also 0 wedged) is **excluded from the mechanism count**: it ran under a log gated on `ops > 0`, so any run where the recovery fired but pulled zero governance ops logged nothing. That gate is removed in `51afeb934`. Round 1 therefore *understates* the fix's activity, and its quiet runs are not interpretable either way.
* Every run on this branch passed (15/15), but passes alone are weak evidence — roughly half of them never entered the failure mode at all. The load-bearing evidence is the 4 logged rescues and the fire-rate/wedge-rate correspondence.
* No unit test: the deadlock is a multi-node partition/heal interaction between gossipsub mesh state, beacon verification and the governance pull. The merobox scenario is the regression test; the unit suites go green with or without the bug.


chefsale:
## Correction: this fix is incomplete — evidence above is overstated

`Test (group-join-mesh-not-ready)` **failed on this branch** after the rebase ([run 31883720288](https://github.com/calimero-network/core/actions/runs/31883720288)), the first failure in 16 runs. It is not a flake, and it exposes a real hole.

```
dropped = 26   trigger = 0   → HIT+STUCK
```

The joiner wedged **and the recovery never ran**. Timeline from its log:

```
12:14:25.30  ns/ Subscribed event   <- the ONLY one, 56s BEFORE the join degraded
12:15:21.62  degraded join (member recorded, no key)
12:15:46.63  reconnect after heal   <- NO new Subscribed event
12:15:46.64  first dropped beacon -> 26 -> wedged
```

The `Subscribed` event fired *before* this node became member-but-keyless, so the gate correctly found nothing and returned. After the heal the peer reconnected **without gossipsub re-emitting `Subscribed`**, so the trigger never got a second chance.

### What this means for the evidence

Whether the fix engages depends on whether the reconnect happens to produce a fresh subscription event. It did in the runs where the recovery was observed firing; it did not here. **The 0/15 result above is therefore partly luck about which trigger fired, not 15 demonstrations that the deadlock is closed.** The mechanism observations remain valid — when the trigger fires, it does recover — but the trigger itself is not reliable enough to close #3406 on its own.

### The better trigger

The failing run hands it over: **26 beacons arrived from a reachable peer and were discarded.** Every dropped beacon is direct evidence that a namespace peer is up and talking to us — which is exactly the fact the stranded node needs and currently throws away.

Using it trusts nothing new. The beacon is still refused; we only take "an unverifiable beacon arrived from peer X" as a cue to pull from X through the authenticated path, where the governance ops must still validate on apply.

Note for anyone picking this up: this must **not** be routed through `spawn_beacon_divergence_sync`. Its `local_has_state` guard rejects precisely the stranded node (no governance state ⇒ empty DAG head), and relaxing that guard is what broke 27 app-migration scenarios in #3413. The new trigger needs its own narrowly-gated pull, as the `ns/` one has.

**Staying open for review.** What is here is a real improvement and makes nothing worse: when the trigger fires it demonstrably recovers, and it never fires on a healthy node. It is simply not sufficient on its own to close #3406. The dropped-beacon trigger ships as a separate PR on top.

chefsale:
Closing in favour of #3462, which is merged and covers the same recovery from a better signal.

Both PRs do the same thing once triggered — pull governance and keys for a member that holds no key. They differ only in what wakes them:

* **This PR:** a peer subscribing to `ns/<id>`. A one-shot edge, and as the [correction above](https://github.com/calimero-network/core/pull/3453#issuecomment-5302196309) documents, it is not guaranteed to fire when needed: in the run that failed here the subscription event arrived 56s *before* the node became member-but-keyless, and the post-heal reconnect produced no second one.
* **#3462:** an unverifiable beacon arriving. Beacons repeat every ~5s from every Ready peer, so the trigger recurs for as long as the node is stranded rather than depending on catching a single edge.

#3462 has fired and converged **4 times on master** post-merge, and master's `group-join-mesh-not-ready` failure rate went from 4-of-7 to 1-of-6.

The only coverage this PR adds over #3462 is the window where a peer subscribes while we are stranded but has not beaconed yet — worth a few seconds, against the cost of a second code path pulling governance. Not a good trade, so: one trigger to reason about.

The `unconditional recovery logging` commit here is worth keeping in mind if this area is revisited — gating a recovery log on `ops > 0` made "ran and pulled nothing" indistinguishable from "never ran" and cost a probe round of interpretable data. #3462 logs unconditionally for that reason.

#3406 remains open: a residual failure survives both fixes, where the admin never registers the joiner's `ns/` subscription and publishes nothing (`NoPeersSubscribedToTopic` × 38) while the joiner sees a healthy 1-peer mesh. No message flows, so no trigger of any kind can fire. Investigating that separately.

PR #3475, opened 2026-08-16
chore(server): drop the node-minted JWT and challenge wire types

chefsale:
## Summary

`JwtTokenRequest`, `JwtRefreshRequest`, `NodeChallenge` and `NodeChallengeMessage` date from when the node minted its own JWTs and signed login challenges. Authentication moved out to the `mero-auth` service, which has its own DTOs on its own router; the node server only *verifies* bearer tokens (`crates/server/src/auth.rs`) and never issues them.

**−110 lines, single file.**

## The tightest island in the audit

All **eleven** references are inside the file that defines them:

```
$ grep -rn 'JwtTokenRequest|JwtRefreshRequest|NodeChallenge' . --exclude-dir=.git --exclude-dir=target
crates/server/primitives/src/admin/mod.rs:602   pub struct JwtTokenRequest
crates/server/primitives/src/admin/mod.rs:607   impl JwtTokenRequest
crates/server/primitives/src/admin/mod.rs:620   pub struct JwtRefreshRequest
crates/server/primitives/src/admin/mod.rs:627   pub struct NodeChallenge
crates/server/primitives/src/admin/mod.rs:629       pub message: NodeChallengeMessage
crates/server/primitives/src/admin/mod.rs:633   impl NodeChallenge
crates/server/primitives/src/admin/mod.rs:635       pub const fn new(...)
crates/server/primitives/src/admin/mod.rs:646   pub struct NodeChallengeMessage
crates/server/primitives/src/admin/mod.rs:652   impl NodeChallengeMessage
crates/server/primitives/src/admin/mod.rs:1229  impl Validate for JwtTokenRequest
crates/server/primitives/src/admin/mod.rs:1252  impl Validate for JwtRefreshRequest
```

Four struct definitions, three `::new` impls, one field reference (`NodeChallenge` flattens `NodeChallengeMessage`), and two `Validate` impls. Nothing outside that file — not a route, not a handler, not a test, not even an `AGENTS.md` line. Most islands in this sweep had at least a doc echo; this one has nothing.

**Anchors proven absent:** no `token`/`refresh`/`challenge`/`login`/`jwt` path among the entries in `endpoints.json`, no matching `.route(` in `crates/server/src`, and `crates/auth` — the only thing that could plausibly want them — does not reference them at all.

## What made it look alive

The two `Validate` impls contain real logic, not stubs: a 128-byte cap and an empty check on `executor_public_key`, and a 4096-byte cap on `refresh_token`. A reader skimming the file sees working validation. But nothing constructs either request, so the impls and the types they guard are equally unreachable — the "trait impls as the whole point" trap.

## What looks similar but is alive

- **`calimero_client::JwtToken`** (`crates/client/src/storage.rs`) is a different type in a different crate, and is what `calimero-client-py` wraps. Untouched.
- **`crates/auth`** is the live authentication surface: `api/handlers/{auth,client_keys,root_keys}.rs` and `auth/token/jwt.rs`, with their own DTOs on their own router, plus a contract test. None of it imports from `calimero_server_primitives::admin`.
- **`crates/server/src/auth.rs`** is alive — the `AuthGuardService` verify side, mounted at four points. It consumes bearer tokens; it does not mint them.
- **`TeeAttestRequest`'s `Validate` impl sits between the two removed ones** in the same block and is routed at `POST /admin-api/tee/attest`. The dead impls are interleaved with a live one, which is why "the whole `Validate` block is stale" would have been the wrong conclusion.

## Verification

- `cargo check --workspace --all-targets` ✅
- `cargo fmt --check` ✅
- `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` ✅ exit 0
- `cargo test -p calimero-server-primitives --test wire_fixtures` ✅ 6/6 — unchanged, since no field on a live route was touched

Found with the `unreachable-subsystems` skill (#3438).

### DIFF
--- a/crates/server/primitives/src/admin/mod.rs
+++ b/crates/server/primitives/src/admin/mod.rs
@@ -596,70 +596,6 @@ impl GetPeersCountResponse {
     }
 }
 
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
-#[non_exhaustive]
-pub struct JwtTokenRequest {
-    pub context_id: ContextId,
-    pub executor_public_key: String,
-}
-
-impl JwtTokenRequest {
-    #[must_use]
-    pub const fn new(context_id: ContextId, executor_public_key: String) -> Self {
-        Self {
-            context_id,
-            executor_public_key,
-        }
-    }
-}
-
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
-#[non_exhaustive]
-pub struct JwtRefreshRequest {
-    pub refresh_token: String,
-}
-
-#[derive(Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
-#[non_exhaustive]
-pub struct NodeChallenge {
-    #[serde(flatten)]
-    pub message: NodeChallengeMessage,
-    pub node_signature: String,
-}
-
-impl NodeChallenge {
-    #[must_use]
-    pub const fn new(message: NodeChallengeMessage, node_signature: String) -> Self {
-        Self {
-            message,
-            node_signature,
-        }
-    }
-}
-
-#[derive(Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
-#[non_exhaustive]
-pub struct NodeChallengeMessage {
-    pub nonce: String,
-    pub context_id: Option<ContextId>,
-    pub timestamp: i64,
-}
-
-impl NodeChallengeMessage {
-    #[must_use]
-    pub const fn new(nonce: String, context_id: Option<ContextId>, timestamp: i64) -> Self {
-        Self {
-            nonce,
-            context_id,
-            timestamp,
-        }
-    }
-}
-
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SyncContextResponse {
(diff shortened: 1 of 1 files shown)


chefsale:
The four scenario failures were not from this PR — pushed fixes for them here since they block every e2e run, not just this one.

**merobox 0.6.54 shipped this morning (10:03) and now raises on a capture it cannot bind** instead of warning and continuing. This PR's run at 11:14 was the first to hit it; master's last e2e was yesterday at 20:16, so master is not yet known-red but would fail identically.

Each of the four was already broken — the failure simply had nowhere to surface:

| scenario | was | now |
| --- | --- | --- |
| `group-membership` | `group_info: data` | `groupId` |
| `group-join-via-inheritance` | `node1_key` from `create_namespace` | from `create_context` |
| `group-open-self-join-key-delivery` | `node1_key` from `create_namespace` | from `create_context` |
| `simple-sync` | `node1_key`/`node2_key` | removed — unused, one names a nonexistent field |

Two details worth recording:

- **`data` is not a field.** The envelope is unwrapped before the capture runs, so `get_group_info` offers `appKey, contextCount, defaultCapabilities, groupId, groupStateHash, memberCount, subgroupVisibility, targetApplicationId, upgradePolicy` directly.
- **`create_namespace` returns only `namespaceId`.** The two scenarios reading `memberPublicKey` from it never bound `node1_key`, then passed it as `executor_public_key` and asserted `is_set` on it — which *held*, because an unbound placeholder stays the literal string `{{node1_key}}` and that is non-empty. So the assertion designed to catch this was itself passing on the bug. The key comes from creating the context, which is where the capture now sits.

I also swept every scenario for the same two patterns; the only other `: data` hit is a method **argument** in `simple-store.yml` (`args: {key: test, value: data}`), not a capture.

The DTO removal this PR is actually about is unaffected: `check`, `clippy -D warnings`, `fmt` and the wire fixtures all pass.

chefsale:
Rebased onto master and dropped my scenario fixes — #3476 landed the same four independently while this was in review, and its version is better: `group-membership` got a real assertion (`{{group_info_id}} == {{subgroup_id}}`) where mine only kept `is_set`, which is the weak check that let the original bug hide.

This PR is now exactly what it says: **−110 lines, one file**, removing the four node-minted JWT/challenge DTOs and their two `Validate` impls. No scenario changes.

`check`, `fmt`, `clippy -D warnings` and the scenario-coverage gate all pass locally.

Issue #3488, opened 2026-08-16
test: autonat_v2::tests::test_dynamic_protocol_change is timing-flaky and reddens unrelated PRs


chefsale:
## Summary

`autonat_v2::tests::test_dynamic_protocol_change` in `crates/network/primitives/src/autonat_v2/mod.rs` fails intermittently under load with

```
thread 'autonat_v2::tests::test_dynamic_protocol_change' panicked at
  libp2p-swarm-test-0.6.0/src/lib.rs:176:13:
expected 2 items from first swarm but got 3
```

The assertion counts swarm events, and a third event arriving within the window makes it fail. It passes reliably in isolation and fails under parallel load, so it is a timing assumption rather than a logic defect.

## Impact

- **Severity: CI reliability.** No product defect.
- **Blast radius:** any PR that runs `cargo test`, regardless of what it changes.
- **Real-world consequence:** it fails on changes that contain no Rust at all, so the red job points at an innocent PR and costs a re-run plus the time to establish that it is unrelated. Observed twice in one session:
  1. locally, during a full `cargo test --workspace` under parallel load — passed immediately when re-run alone
  2. on #3463, a PR containing only a Python script and CI YAML — passed on re-run

Each occurrence needs someone to open the log and confirm it is spurious, which is exactly the cost a flaky test imposes.

## Steps to reproduce

1. `cargo test --workspace` on a machine with other work in flight (several cargo jobs, or a loaded CI runner).
2. Observe the failure above in `calimero-network-primitives`.
3. `cargo test -p calimero-network-primitives autonat_v2::tests::test_dynamic_protocol_change` — passes.

Actual: fails under load, passes in isolation. Expected: deterministic either way.

## Criteria for resolving

- [ ] The test asserts on the events it cares about rather than on an exact count, so an extra swarm event within the window does not fail it — or it drives the swarms deterministically instead of relying on timing
- [ ] It passes with the rest of the suite running concurrently, not only in isolation (e.g. repeated `cargo test --workspace` runs on a loaded machine)
- [ ] If it cannot be made deterministic, it is marked `#[ignore]` with a stated reason and the coverage it provides is named, matching the convention used elsewhere in the repo
- [ ] `cargo fmt --check`, `cargo clippy`, and `cargo test` pass

---

Not a finding of the `unreachable-subsystems` audit (#3438) — encountered repeatedly while running its verification and filed so the cost is recorded rather than absorbed silently.

PR #3505, opened 2026-08-17
refactor(op-adapter)!: delete the stand-in account

### LINKED ISSUE #3414: Delete the legacy stand-in account bridge (#3346 step 5, never landed)
## Summary

#3346 step 5 — *"delete `legacy_account_id`, `legacy_authorship`, `accounts_to_member_keys`, `member_key_for_account`, and `account_for_author`'s precedence logic"* — never landed. #3346 auto-closed when #3391 merged the wire flip, but the deletion that flip was justified by is still outstanding.

Two of the five are gone (`accounts_to_member_keys`, `member_key_for_account`, both 0 refs). The rest remain.

## Why it matters

The epic's own argument: two representations of "who acted" coexisting **is** the bug. It already produced a `scope_root` split in #3320, and `account_for_author`'s precedence rules exist *only* because the bridge does. `legacy_account_id`'s doc says it plainly:

> Living in the adapter — the crate that is deleted at cutover — is what keeps it from outliving the transition.

It has outlived the transition.

## Scoping: the raw counts are misleading

Raw: `legacy_account_id` 56, `legacy_authorship` 15, `account_for_author` 13, `writer_account` 7. Splitting on the `cfg(test)` boundary in each file gives a very different picture.

**Production is four call sites:**

| site | what it is |
| --- | --- |
| `context/src/scope_projection.rs:141` | `account_for_author`'s key-derived fallback |
| `context/src/scope_projection.rs:187` | `legacy_authorship(author)` in `build_op` |
| `op-adapter/src/lib.rs:86` | `writer_account`'s fallback |
| `op-adapter/src/lib.rs:96` | `legacy_authorship`'s body |

Everything else — roughly 86 references — is **test fixtures that deliberately construct stand-ins**. `unified_op_store.rs` (cfg at 91, refs at 158/167/178) and `unified_applier.rs` (cfg at 94, refs at 156/165/176) are entirely test code, as is `scope_projection.rs` past 2029 and `op-adapter` past 546.

## Is the fallback still reachable? Measured, not assumed

Instrumented `account_for_author`'s fallback and ran the fold-equivalence suites (`projection_membership_equivalence`, `per_device_authorization`, `local_group_governance_convergence`):

- **13 hits total**
- **9 with `view_knows=false`** — the derived id names nobody. Harmless: authorization correctly answers "not authorized".
- **4 with `view_knows=true`** — all from one test, `a_folded_join_device_does_not_hide_an_inherited_admin`, which **seeds the condition itself**:

```rust
.save(&ns, &meta(calimero_op_adapter::legacy_account_id(&admin)))
```

`view_knows_author` matches because `root.1 == genesis_admin` equals the derived id — the fixture built exactly that. Production writes real accounts at every `admin_identity` site (`admin_account`, `founder`, or the placeholder).

**Conclusion: production no longer names a real principal by a bare key.** The fallback fires only for principals nobody knows.

## Also found: dead code in `account_for_author`

```rust
let own = legacy_account_id(key);
if view_knows_author(view, &own, root) { return own; }
own
```

Both branches return `own`; the `view_knows_author` call has no effect. Almost certainly a leftover from when the binding/derived precedence was reversed. Its doc comment is stale too — it cites `SubgroupCreated.admin` as a source of key-derived ids, but that field is `AccountId` now and the encoder explicitly passes the carried account.

## Proposed slices

- **A — remove the dead branch** in `account_for_author` and refresh its stale doc. No behaviour change; confirms the baseline.
- **B — convert the fold-equivalence fixtures** to real accounts, in file-sized chunks. This is the actual cost and the actual risk: these tests are the proof that the unified projection matches the live resolvers, so each must keep proving the same property with real accounts rather than derived ones.
- **C — delete the four production sites and the four functions.** A formality once B lands, and `writer_account`'s documented footgun (*"a writer set seeded before its writer enrolls holds that key's stand-in"*) disappears with it.

## Method note for whoever picks this up

Two instrumentation attempts returned a confident `0` while being structurally blind: `#[cfg(test)]` statics (integration tests do not compile the lib with `cfg(test)`) and `tracing::warn!` (no subscriber installed in these tests). Only `eprintln!` observed the 13 hits. Verify the instrument fires before trusting a zero here.

chefsale:
Closes #3414. `legacy_account_id`, `writer_account` and `legacy_authorship` are
gone, and `src/legacy.rs` with them.

The stand-in hashed a signing key into an account-shaped id, from before accounts
existed. Every row is keyed by a real account now, so it named a principal
nowhere — and because it was account-*shaped* and different per key, "we could
not attribute this op" was indistinguishable from "this is somebody" at every
site that saw it. That is what let it break three separate downstream checks
silently, each found and fixed in turn: a join whose device could not be
cross-checked (#3494), a key rotation dropped by the account-plane fold, and a
writer set refused to its own owner (#3500).

Two things had kept it alive.

## The writer plane was asking the wrong table

`account_for_group` read `AccountBindingRepository`, which is populated by
**applied ops** — so it knew a key's account only once that key's device link had
landed, which for the node currently creating a context has not happened. That
window is why the fallback looked load-bearing, and it is what I previously
believed made this flag-day work.

It isn't. This node's own `NodeDevice` row is minted locally by `ensure_enrolled`
from its own account root and needs no ops at all, so the account is always
available — the code was simply asking the propagated table instead of its own.
`account_for_context` already documented the invariant, *"there is no state in
which this node has no account to name"*, and it now actually holds.

So the plane splits cleanly:

| whose key | resolved from |
| --- | --- |
| this node's own | the local device row — always known |
| somebody else's | the folded binding, or nothing |

## An unattributable op still needs an authorship

It gets `Authorship::unattributed`, naming `UNATTRIBUTED_ACCOUNT` — a value no
genesis can produce.

**A sentinel and not an `Option`, on purpose.** Every gate on this plane asks
whether the author equals some real principal, so a value that matches nothing
**fails closed everywhere by construction**: membership, admin, ownership and the
account-plane handoff check all answer "no" without any of them needing a case
for it. An `Option` would put that decision in a dozen places and depend on each
getting it right — and the fold cannot resolve the author itself (reading a
binding mid-fold splits the root by arrival order), so the field has to stay
carried regardless.

`device_key` is still recorded there, because it is a fact: that key signed the
op and `Op::verify` checks the signature against it. Only the principal it speaks
for is unknown.

Needed one small addition: a `const fn from_raw` on the id macro in
`calimero-primitives`, since `From<[u8; 32]>` cannot be `const` and a sentinel
has no other way to be a `const` item.

## Tests converted, not deleted

What they assert changes, so they were rewritten rather than dropped:

- **A revoked device resolves to nobody**, not to a stable stand-in. That retires
  the footgun the old test pinned: a writer set naming a device's stand-in kept
  admitting the key *after* revocation, because the stand-in outlived the
  binding. There is no such account to name now.
- The op-adapter test that showed a *derived* account is a member of nothing now
  shows the **sentinel** is — which is the property the fail-closed argument
  rests on.

## Compatibility

Deliberately not preserved, per the call on this: pre-credential data whose joins
carried no credential resolves to `unattributed` and is inert.

## Verification

`cargo fmt --check`, `clippy --workspace --all-targets -D warnings`, the
`mock-attestation` clippy and test runs, and `cargo test --workspace` all pass.

`--exclude cargo-mero` on the workspace run: that crate holds
`bundle_produces_signed_mpk`, which needs a wasm32 fixture and an isolated
target dir, and which CI's own main Rust job skips for that reason
(`--skip bundle_produces_signed_mpk`). The dedicated `cargo-mero-e2e` job still
runs it, so this matches CI rather than hiding anything.

One unrelated flake showed up and passed on retry:
`autonat_v2::tests::test_dynamic_protocol_change`, which is #3488 — network code
this change does not touch.


chefsale:
### Fixes the two red E2E jobs: a paired device could not name its account

`shared-storage-account-writers-two-devices` and `account-device-revoke-lockout` both died at `crates/governance-store/src/node_device.rs:680`:

```
this node already holds device 624345ca… for account 9705d753… in ContextGroupId(…),
and it is linked — its replica state is held under that id. Moving a machine between
accounts means revoking the existing device first
```

That account is Alice's — the one node-3 had just been paired onto.

**Cause.** This PR's `account_for_group` resolved the account by *enrolling* (`ensure_enrolled`), which mints this node's device from its **own** account root. Right for a node that speaks for itself, wrong for a paired one: a paired row names an account rooted at another node's key, and enrolment is entitled to replace such a row — that is how a mistyped `pair-init` is recovered from. As a resolver, that entitlement is destruction in two ways:

- a **linked** pairing is refused outright, so every call on a correctly paired device fails (`my_account` → `InternalError`, both red jobs);
- an **unlinked** pairing — the window between `pair-init` and the certificate applying — is silently re-minted instead, taking the KEM secret the in-flight certificate is addressed to, so pairing can never complete. Nothing in CI catches that half.

The file's own tests already name this bug class (`a_paired_device_is_reusable_and_is_not_re_minted_over`: *"the pull called `ensure_enrolled`, which releases a row belonging to another account and mints a replacement"*). The new resolver repeated it.

**Fix.** Read the held row as `reusable_device` — the primitive that exists for callers needing to be *addressable* rather than to be *themselves* — and fall back to the account root when nothing usable is held. The fallback names the same account `ensure_enrolled` would in both remaining cases (no row, revoked row) without releasing the row to do it: one `NodeDeviceIdentity` row serves every namespace while a revocation is per-namespace, so an enrolling resolver would let a read in the namespace that revoked this device spend the device it still holds in the others. Minting stays with the enrolment path, which knows what it is enrolling into. Creation-time behaviour this PR was built for is unchanged: no row → own-root account.

**Reproduction** — three tests in `node_device.rs`, the first two red before the fix, the first with the byte-identical error above:

| test | before | after |
| --- | --- | --- |
| `a_paired_device_resolves_the_account_it_speaks_for` | FAILED: `already holds device … and it is linked` | ok |
| `resolving_the_account_never_re_mints_over_the_held_device` | FAILED: wrong account, device + KEM secret replaced | ok |
| `a_revoked_device_stops_naming_the_account_it_spoke_for` | ok | ok (pins the fix from over-correcting) |

The first is the in-process form of the scenario's own `json_equal({{phone_account}}, {{alice_account}})` assertion.

`cargo fmt --check`, `clippy --all-targets -D warnings`, `cargo test -p calimero-governance-store` (566) and `-p calimero-context` (292) green locally; the two scenarios are the real proof and run on this push.

Unrelated leftover, flagged not fixed: `account_for_context`'s doc still says the account is scoped to the namespace ("accounts in two namespaces stay uncorrelatable"), which this PR's own `the_creation_time_resolver_trap_is_gone` contradicts ("an account is no longer scoped to anything").

PR #3512, opened 2026-08-17
fix(node): stop the actors before the runtime they depend on goes away

### LINKED ISSUE #3503: merod busy-loops on a dead netlink socket after shutdown (76% of all e2e log bytes)
## Summary

When a node finishes shutting down, `libp2p_mdns`'s interface watcher busy-loops on a dead netlink socket until the process is killed. `Swarm::poll` spins at roughly **280,000 iterations per second** for the remainder of the node's life, logging three lines per iteration.

The Tokio runtime shuts down, which takes the netlink socket that if-watch holds. The watcher's stream returns an error, is polled again, returns the same error, forever. The trigger is visible 95 µs after the last shutdown line:

```
08:46:24.958932  INFO calimero_node::network_event_processor: Graceful shutdown complete
08:46:24.959865  INFO calimero_node::run: Node shutdown complete
08:46:24.960048  INFO calimero_node::dag_compactor: DAG compaction actor stopped
08:46:24.960053 ERROR Swarm::poll: netlink_proto::framed: failed to read from netlink socket:
                      Custom { kind: Other, error: "A Tokio 1.x context was found, but it is being shutdown" }
08:46:24.960067  WARN Swarm::poll: netlink_proto::connection: netlink socket stream shut down
08:46:24.960075 ERROR Swarm::poll: libp2p_mdns::behaviour: if watch returned an error:
                      rtnetlink socket closed. Connection has been terminated.
                      … repeating until SIGKILL
```

mDNS is on by default (`mdns: true`, `crates/network/primitives/src/config.rs`), so every node is exposed. Note that mDNS discovery contributes nothing to a containerised fleet, which finds peers over bootstrap, kad and rendezvous.

## Impact

**A CPU core spins for the whole shutdown window on every node stop.** merobox's teardown gives containers a 5 s drain before SIGKILL, so in practice that is ~5 s of a pegged core per node, per scenario, plus whatever a real deployment's stop window allows. It also delays nothing visibly, which is why it went unnoticed.

**It is 76% of all e2e log bytes.** Measured over 20 node logs from 6 scenarios in run 32011251220 (598 MB total):

| line | count | MB | % of all bytes |
|---|---:|---:|---:|
| `netlink_proto::framed` — failed to read from netlink socket | 746,693 | 206.1 | 34.5% |
| `netlink_proto::connection` — netlink socket stream shut down | 746,692 | 138.9 | 23.2% |
| `libp2p_mdns::behaviour` — if watch returned an error | 462,713 | 109.2 | 18.3% |
| **combined** | **1,956,098** | **454.2** | **76.0%** |

One node makes the shape plain: `rust-xcall-1` lived 22.5 s and wrote 1,396,557 lines, of which **1,388,139 — 99.4% — are these three**. Its per-second counts over the final window are 261,979 / 281,683 / 278,601 / 285,809 / 270,565.

Downstream: CI artifacts of 325 MB for a single scenario, and any real shutdown problem is buried under a million lines. Turning the log level down would hide the symptom and leave the spin in place, so a `RUST_LOG` filter is explicitly **not** the fix.

## Steps to reproduce

1. Run any merobox scenario with two or more nodes, e.g. `cd apps/xcall-example && merobox bootstrap run workflows/xcall-example-scenario.yml --e2e-mode`.
2. Let it finish and tear down.
3. `grep -c netlink data/container-logs/*.log`

**Actual:** hundreds of thousands of matches per node, all timestamped after `Node shutdown complete`, at ~280k lines/second.
**Expected:** none. Once shutdown has begun the swarm should stop being polled, and an if-watch error should not be retried in a tight loop.

Reproduces on every scenario in the corpus; the count scales with how long the container survives after shutdown completes.

## Criteria for resolving

- [ ] After `Node shutdown complete`, no further `Swarm::poll` output is produced — the swarm task is not polled once shutdown has begun.
- [ ] A terminal if-watch error is not retried in a tight loop; either the mDNS behaviour is disabled or the watcher stops after a terminal error.
- [ ] mDNS is not enabled where it cannot help. Either default `mdns` to `false`, or leave the default and have merobox/e2e configs disable it — the decision to make explicitly is whether local-network discovery is a default worth its cost.
- [ ] A regression check: a scenario teardown produces no more than a handful of `netlink`/`if watch` lines, assertable with the same `grep -c` above.
- [ ] `cargo fmt --check`, `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`, and `cargo test` pass.

---

Found while auditing log volume, alongside calimero-network/core#3501 (library log noise) and calimero-network/core#3502 (user data in logs).

chefsale:
Fixes #3503.

## Root cause

**Shutdown never told the Actix system to stop.** It came down anyway — but by accident, and far too late.

Dropping `arbiter_pool` closes the arbiter-handle channel, and the task holding that sender calls `System::current().stop()` when the send fails. That drop happens when `run()` **returns** — after the drain, after the flush, after `Node shutdown complete`. So the arbiters tear their actors down against a global runtime that is already going away.

Any actor holding an IO resource registered on that runtime then sees it fail. The swarm's mdns interface watcher is one of them:

```
08:46:24.959865  INFO calimero_node::run: Node shutdown complete
08:46:24.960053 ERROR Swarm::poll: netlink_proto::framed: failed to read from netlink socket:
                      "A Tokio 1.x context was found, but it is being shutdown"
08:46:24.960075 ERROR Swarm::poll: libp2p_mdns::behaviour: if watch returned an error: rtnetlink socket closed
                      … ~280,000 times a second for 5 s, until SIGKILL
```

if-watch reports the error, the behaviour treats it as retryable, and `Swarm::poll` spins — which both pegs a core and stops the arbiter from ever finishing, so the process has to be killed from outside. That spin is **454 MB of the 598 MB corpus, 76% of all e2e log bytes.**

## The fix

Stop the system explicitly during the drain, while the runtime that owns the actors' IO is still healthy — there is then no failing resource to spin on.

Placed **before the datastore flush**, because actors write to the datastore exactly as the background tasks below them do, and **after the server and bridge drains**, which both talk to actors. `ArbiterPool` grows a `stop()`, holding a `System` handle taken on the system's own thread — `System::current()` resolves nowhere else, and every thread it does resolve on is owned by the system. It follows the file's existing `*_done` convention so a finished `JoinHandle` is never re-awaited.

## Before → after

The ordering change is observable **without netlink**, which is what makes it testable off Linux. Running a real node and sending it SIGTERM:

**Before** — `Node shutdown complete` at line 75, then six actor-shutdown lines *after* it:

```
75: INFO calimero_node::run: Node shutdown complete
    INFO calimero_node::sync_session_bridge: SyncSession actor stopped
    INFO calimero_node::gc: Garbage collection actor stopped
    INFO calimero_node::dag_compactor: DAG compaction actor stopped
    INFO calimero_node::state_delta_bridge: StateDelta actor stopped
```

**After** — the actors come down first and that line is genuinely last:

```
    INFO calimero_node::sync_session_bridge: SyncSession actor stopped
    INFO calimero_node::state_delta_bridge: StateDelta actor stopped
    INFO calimero_node::dag_compactor: DAG compaction actor stopped
81: INFO calimero_node::run: Node shutdown complete
```

**That trailing window is the one the netlink socket dies in.** Both runs exit in ~0.04 s with code 0 on macOS, where there is no netlink and so nothing to spin on — which is exactly why the Linux confirmation has to come from CI rather than from a local run. I have deliberately not claimed the spin is gone on the strength of a macOS test; the check is `grep -c netlink` over this PR's own e2e log artifacts, and the container exit code that merobox now records (137 vs 143).

## Regression test

`stop_ends_the_system_while_the_pool_is_still_alive` — calls `stop()` with the pool still in scope and asserts the system task finishes on its own, i.e. without the drop that used to be the only thing ending it.

```
cargo test -p calimero-node --features calimero-storage/testing   → 11 test binaries, 0 failures
cargo fmt --check                                                 → clean
cargo clippy -p calimero-node --all-targets \
  --features calimero-storage/testing -- -D warnings              → clean
```

## Out of scope

The second lever in #3503: **mdns is still `true` by default.** Whether local-network discovery earns that default in a fleet that finds peers over bootstrap/kad/rendezvous is a separate question from this ordering bug, and turning it off would mask this rather than fix it. Worth deciding on its own.


chefsale:
## Verified on Linux, from this PR's own e2e artifacts

The macOS run in the description could only prove the ordering change, not the spin — there is no netlink there. Here is the Linux confirmation, `xcall-example-scenario` node 1, which was the worst-affected scenario:

| | before (run 32011251220) | after (this PR) |
|---|---:|---:|
| lines | 1,396,557 | **600** |
| bytes | ~325 MB | **177 KB** |
| netlink / if-watch lines | 1,388,139 (99.4%) | **0** |
| artifact size | 12.1 MB compressed | **21 KB** |

Zero netlink or `if watch returned an error` lines in every scenario I checked — `xcall-example-scenario`, `shared-storage-concurrent-rotation-converge`, `auto-follow-flag-flipped-late`, `account-device-revoke-lockout`.

And the ordering change holds on the real platform — `Node shutdown complete` is genuinely last, with the actor teardown ahead of it:

```
14:01:44.304311  INFO calimero_node::state_delta_bridge: StateDelta actor summary …
14:01:44.304326  INFO calimero_node::state_delta_bridge: StateDelta actor stopped
14:01:44.304505  INFO calimero_node::gc: Garbage collection actor stopped
14:01:44.306005  INFO calimero_node::run: Node shutdown complete
```

That is the window the netlink socket used to die in, now closed.

PR #3515, opened 2026-08-17
test(e2e): ask the node who it is, not the namespace

chefsale:
The unlock for deleting `GET /admin-api/namespaces/:id/identity`.

48 scenario steps read this node's identity through `get_namespace_identity`, a step that takes a namespace and cannot use it. Core's own handler for the route is explicit about why:

> **Superseded by `GET /admin-api/identity`.** Every field it returns is node-level — one root key is one account everywhere, and a node signs with one key — so the namespace in the path cannot change the answer. What it still does is gate on participation… **Kept because 44 merobox scenarios drive it** through the `get_namespace_identity` step. It retires once merobox has one for the node-level route.

merobox has had that step for a while (`node_identity`, requiring client-py ≥ 0.6.27). These scenarios were the only thing keeping the route alive.

## The migration

Per step: `type: get_namespace_identity` → `type: node_identity`, drop `namespace_id`, and rename the captured field where the node-level step names it differently.

| captured field | steps | change |
| --- | --- | --- |
| `publicKey` | 38 | none — same name |
| `account` | 9 | → `accountId` |
| `namespaceId` | 0 | — |

45 files, 48 steps. Before: 48 legacy / 8 node-level. After: 0 / 49.

## Two things the rename made visible

- **Eleven steps were still named for a namespace identity** ("Get node-1 namespace identity"). They now name the field they actually take — an account or a signing key. Those are different principals, and a scenario that captures one while asserting the other is exactly the bug this naming hid.
- **`35-two-namespaces-coexist` captured an admin key twice**, once per namespace, from the same node. Under a node-level read those are the same call, so the second is gone with a comment recording why it existed.

## One deliberate behaviour loss, stated rather than discovered

The old route **gated on participation**: 404 for a namespace this node had never joined. `node_identity` does not. Nothing relied on it — all 48 usages are getters feeding `outputs`, and none sits under `expect_failure`, which I checked rather than assumed — but these steps no longer double as an implicit "did the join land" check. A scenario that wants that should assert it directly.

## Verification (no Docker here, so: schema + lints, then CI)

| check | result |
| --- | --- |
| all 45 changed scenarios through merobox's `WorkflowConfig` pydantic model | ✓ 45 valid, 0 invalid |
| `scripts/lint-e2e-convergence.py` | ✓ 129 scanned, no new convergence races |
| `scripts/check-scenario-coverage.py` | ✓ 128/129 registered, 1 baselined (unchanged) |

Validation goes through the pydantic model deliberately — `merobox bootstrap validate` keeps a separate hand-maintained step list and reports false unknowns, so it is not the check `bootstrap run` performs. The E2E matrix on this PR is the real proof.

## What this unblocks

```
this PR: 48 steps → node_identity
   ↓
core: delete GET /namespaces/:id/identity + NamespaceIdentityApiResponse
   ↓
merobox: drop the get_namespace_identity step
   ↓
client-py: drop the get_namespace_identity binding
```

Stacks with #3511 in one file (`account-device-revoke-lockout.yml`) — whichever lands second gets rebased. Related: calimero-network/mero-js#101 adds `getNodeIdentity()` and deprecates the per-namespace call on the SDK side.


chefsale:
Superseded by #3522. #3518 merged into this PR's branch rather than into master, and both branches were then deleted, so this PR's head ref no longer exists and it cannot merge — GitHub's compare confirms master never received it (`master...8ae3a3fc → diverged, behind_by: 1`). All three commits were recovered and rebased onto master in #3522, which carries the scenario migration, the paired-node `GET /identity` fix, and the route deletion together.

PR #3528, opened 2026-08-18
fix(api)!: drop governanceOp from both join responses

### LINKED ISSUE #3485: api: governanceOp is always empty on both join responses, and two SDKs type it non-optional
## Summary

`POST /admin-api/groups/join` and `POST /admin-api/namespaces/:namespace_id/join` both return `governanceOp` on every successful join, and its value is always the empty string. `crates/context/src/handlers/join_group.rs:745` is the only construction site of `JoinGroupResponse` and hardcodes `governance_op_bytes: vec![]`; the two handlers then `hex::encode` that into `governanceOp: ""`.

The field is a remnant of a superseded model in which a joiner handed a signed `JoinWithInvitationClaim` back to its caller for an orchestrator to relay to the inviting node. Direct request-response join replaced that, and the relay endpoint was never built — its client method was removed in #3450.

## Impact

- **Severity: correctness of a published API surface**, low blast radius but permanent until addressed.
- **Blast radius:** every caller of either join endpoint, on every deployment.
- **Real-world consequence:** an integrator reads the type, sees a required field, and writes code against a value that is always `""` — with no error to signal that nothing is coming. Both shipped SDKs type it as non-optional:
  - `mero-js/src/admin-api/admin-types.ts:480,891` — `governanceOp: string`
  - `swift-sdk/Sources/MeroKit/Admin/AdminTypes.swift:620` — `public let governanceOp: String`

  and `mero-js`'s unit tests mock it as `'op-hex'`, i.e. as if it carried a value.

## Steps to reproduce

1. Create a namespace on one node and invite a second.
2. Join from the second node: `POST /admin-api/namespaces/{id}/join`.
3. Read `data.governanceOp` in the response.

Actual: `""`, on every successful join. Expected: either a serialized signed op, or no such field.

## Criteria for resolving

Two defensible outcomes; the decision belongs to whoever owns the orchestrator/HA invite flow.

- [ ] Decide whether a relay model returns. Nothing in the tree suggests it does — `join_credential.rs` documents three cleartext join paths, none of which relays
- [ ] **If not:** remove `governance_op_bytes` from `JoinGroupResponse` and `governanceOp` from both response DTOs, with coordinated `mero-js` and `swift-sdk` releases dropping the field from their types
- [ ] **If so:** populate it, and the field stops being a lie
- [ ] Wire fixtures regenerated (`UPDATE_FIXTURES=1 cargo test -p calimero-server-primitives --test wire_fixtures`) if the shape changes
- [ ] `cargo fmt --check`, `cargo clippy`, and `cargo test` pass

---

Found during the `unreachable-subsystems` audit (#3438). #3450 removed the client method and the DTOs behind the relay, and corrected the field's doc comment — which had instructed callers to relay to an endpoint that never existed — but deliberately left the wire field, since removing it is an SDK-visible break rather than a local cleanup.

chefsale:
Closes #3485

## What

`POST /admin-api/groups/join` and `POST /admin-api/namespaces/:namespace_id/join` returned `governanceOp` on every successful join, always `""`. `JoinGroupResponse::governance_op_bytes` had a single construction site that hardcoded `vec![]`; both handlers hex-encoded that into the wire field.

It is the last piece of a superseded model in which a joiner handed a signed `JoinWithInvitationClaim` back for an orchestrator to relay to the inviting node. Direct request-response join replaced that — the inviting node applies the op itself over the stream — and the relay endpoint was never built. #3450 removed its client method and DTOs but deliberately left the wire field, since dropping it is an SDK-visible break.

## Decision: remove, not populate

The issue leaves two branches open. Nothing in the tree suggests the relay returns: `join_credential.rs` documents three cleartext join paths, none of which relays. Populating would mean inventing a producer for a value with no consumer.

Nothing reads it — I grepped every repo in the workspace (core, mero-js, swift-sdk, client-py, merobox, mdma, mero-tee, tauri-app, mero-mcp, mero-react). Every occurrence outside the type declarations is a **test mock inventing a plausible value** for a field that is always empty: `"deadbeef"` here in `crates/client/src/tests.rs`, `'op-hex'` in mero-js, `'MemberAdded'` in mero-react. That is the failure mode the issue predicts for integrators, already reproduced by our own fixtures. `meroctl`'s `Report` never printed it.

Callers that need the principal a join produced read `memberAccount`, which both handlers already return.

## Changes

- Remove `governance_op_bytes` from `JoinGroupResponse`, its initializer, `governance_op` from the DTO, and the two `hex::encode` call sites.
- Drop the field from the `calimero-client` mock.
- **Add a wire fixture** for the join response. The issue's checkbox was a no-op as written — `fixtures/wire/groups/` held only `reparent.*`, so `UPDATE_FIXTURES=1` regenerated nothing. Now the next change to this shape fails a Docker-free test naming the drift instead of reaching the SDKs unannounced. The fixture's key and account are deliberately unequal and rendered in different alphabets, so a field crossed between the two spaces diverges rather than round-tripping.
- CHANGELOG entry under `[Unreleased] → Removed`, marked breaking.

## Coordination

Both shipped SDKs typed the field non-optional and are also missing `memberAccount`; matching PRs fix both:

- calimero-network/mero-js#103
- calimero-network/swift-sdk#20

**Merge those first.** Swift is the reason: a synthesized `Codable` with a non-optional `governanceOp` fails the whole join response with `keyNotFound` once the server stops sending it, not just that one value. Both SDKs ignore unknown JSON keys, so they land cleanly against nodes that still send it. Same asymmetry inside core: the DTO has no `deny_unknown_fields`, so a new client reads an old server fine, while an old pinned `calimero-client` against a new node errors on the missing field — a clean break on the rc line, consistent with #3450.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets` on the touched crates, `cargo check --workspace --all-targets`, and `cargo test -p calimero-client -p calimero-server-primitives -p calimero-context` all green.

### DIFF
--- a/.github/actions/setup-merobox/action.yml
+++ b/.github/actions/setup-merobox/action.yml
@@ -46,9 +46,25 @@ runs:
         #            the typed binding with "node identity read failed" — before
         #            any HTTP call, which is why the node logs showed no request
         #            at all. merobox#337 stopped the two pins drifting.
-        # When this next moves: if the bundled client-py no longer requires
-        # upgradePolicy, delete UPGRADE_POLICY_COMPAT (crates/server/primitives).
-        MIN_MEROBOX="0.6.55"
+        #   0.6.56 — the bundled `calimero-client-py` is >= 0.6.29, which is the
+        #            first release compiled against a core that made
+        #            `governanceOp` optional on the join responses. Below this the
+        #            bundled client rejects a join from a node that has retired
+        #            the field, with a serde `missing field` error before any
+        #            scenario logic runs — every scenario that joins a namespace
+        #            fails at the join step (#3485). 0.6.56 also drops the
+        #            `requester` step field with the DTOs that lost it (#3492).
+        # UPGRADE_POLICY_COMPAT: the previous note here said to delete it once the
+        # bundled client-py stopped requiring `upgradePolicy`. It has — nothing in
+        # client-py master sends or reads the field — but deleting the shim is a
+        # response-shape change of exactly the class #3485 turned out to be:
+        # `mero-js` types `upgradePolicy: string` and `swift-sdk` declares it
+        # non-optional on a synthesized `Codable`, so removing it needs the same
+        # SDK-then-client-py-then-merobox sequence rather than a rider on this
+        # bump. Core's own struct has carried `#[serde(default)]` on the field
+        # since before rc.21, so a client compiled from core survives its
+        # removal; the two hand-written SDKs are what need the coordination.
+        MIN_MEROBOX="0.6.56"
 
         case "$(uname -m)" in
           x86_64) MEROBOX_ARCH="linux-x86-64" ;;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@
   on a published crate and posted to `admin-api/groups/claim-invitation`, a
   route with no handler anywhere, so any caller got a 404. Direct
   request-response join replaced the relay it belonged to ([#3450])
+- **`governanceOp`** from both join responses (`POST admin-api/groups/join`,
+  `POST admin-api/namespaces/:namespace_id/join`), along with
+  `JoinGroupResponse::governance_op_bytes`. **Breaking** for
+  `calimero-server-primitives` and for any client that decodes the response
+  into a type with a required field of that name — but the node only ever sent
+  `""` there, the last piece of the relay #3450 removed, so no working caller
+  can depend on its value. Read `memberAccount` for the principal the join
+  produced. Matching SDK releases drop the field from `mero-js` and
+  `swift-sdk` ([#3485])
 
 ### Changed
 
@@ -789,4 +798,5 @@ Integrations:
 [#3440]: https://github.com/calimero-network/core/pull/3440
 [#3450]: https://github.com/calimero-network/core/pull/3450
 [#3485]: https://github.com/calimero-network/core/issues/3485
+[#3528]: https://github.com/calimero-network/core/pull/3528
 [#3530]: https://github.com/calimero-network/core/pull/3530
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,14 @@
   into a type with a required field of that name — but the node only ever sent
   `""` there, the last piece of the relay #3450 removed, so no working caller
   can depend on its value. Read `memberAccount` for the principal the join
-  produced. Matching SDK releases drop the field from `mero-js` and
-  `swift-sdk` ([#3485])
+  produced.
+
+  Every pinned decoder was updated first, because one of them runs our own
+  E2E: `mero-js` 11.2.0 and `swift-sdk` dropped the field, `calimero-client-py`
+  0.6.29 rebuilt against the tolerance added in [#3530], and merobox 0.6.56
+  bundles that client — so `MIN_MEROBOX` moves to 0.6.56 here, the floor at
+  which a join against a node without the field still decodes ([#3485],
+  [#3528])
 
 ### Changed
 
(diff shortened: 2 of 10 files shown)


chefsale:
## Blocked — converting to draft

The E2E suite went red across essentially every scenario that joins a namespace:

```
Join namespace failed on invite-join-regression-2:
  Client error: missing field `governanceOp` at line 1 column 233
```

That is `serde`, so the decoder is Rust, not one of the two hand-written SDKs this PR coordinated with. I had the SDK ordering right and the Rust one wrong.

**The chain I missed:** `calimero-client-py` pins `calimero-client` / `calimero-server-primitives` to `{ git = "…/core", branch = "master" }` and publishes a **prebuilt wheel**; merobox bundles that wheel; core's CI downloads a **released merobox binary** (`MIN_MEROBOX=0.6.55`). So the struct that refuses the response was compiled from an older master and ships inside the very tool that runs our E2E. No amount of SDK sequencing touches it.

It is also circular as things stand: client-py builds from master, so a tolerant wheel cannot exist until the tolerance is on master, and this removal cannot reach master while CI runs the intolerant wheel.

**#3530 breaks the cycle** — `#[serde(default)]` on the field, no wire change, mergeable on its own. Sequence:

1. #3530 → master
2. `calimero-client-py` wheel from master + merobox release carrying it; bump `MIN_MEROBOX` in `.github/actions/setup-merobox`
3. this PR, rebased, lands green

Nothing is wrong with the diff here — the Wire Contract gate and SDK E2E (paired) both pass, and the two SDK PRs (calimero-network/mero-js#103, calimero-network/swift-sdk#20) are green and mergeable now. It is purely waiting on the release chain.

chefsale:
Unblocked and ready. Rebased onto master and out of draft.

The chain that had to land first is complete:

| link | what | state |
| --- | --- | --- |
| #3530 | `#[serde(default)]` on `governanceOp` — broke the circular dependency | merged |
| calimero-network/mero-js#103 | `governanceOp` → `memberAccount` | merged |
| calimero-network/swift-sdk#20 | same, and the `keyNotFound` hazard this PR would otherwise trip | merged |
| calimero-network/calimero-client-py#93 | 41-commit core catch-up → **0.6.29 on PyPI** | merged |
| calimero-network/merobox#341 | drops the `requester` step field → **v0.6.56 released** | merged |

## What changed in this PR since the first attempt

**Second commit: `MIN_MEROBOX` 0.6.55 → 0.6.56.** `setup-merobox` downloads `releases/latest` and then asserts it is at least `MIN_MEROBOX`, so this does not select the binary — it refuses one too old for this change. 0.6.56 is the first release bundling client-py 0.6.29. It rides here rather than landing separately because the floor exists only to protect this removal; separating them leaves a window where the removal could merge without it.

**The shim's test flipped rather than being deleted.** #3530 added `join_response_deserializes_without_governance_op`, which cannot compile once the field is gone. It is now `join_response_ignores_a_governance_op_from_an_older_node`: with the field removed, the risk is no longer a client that misses it but one that *meets* it, since a node predating this removal still sends it. Same test budget, the property that now matters.

**`UPGRADE_POLICY_COMPAT` stays, deliberately.** The note in `setup-merobox` said to delete it once the bundled client-py stopped requiring `upgradePolicy` — and it has; nothing in client-py sends or reads the field. But removing it drops a field from `GroupInfo` responses, which is exactly the class of change this PR turned out to be: `mero-js` types `upgradePolicy: string` and `swift-sdk` declares it non-optional on a synthesized `Codable`. It needs its own SDK → client-py → merobox sequence. The reasoning is recorded where the old note was, so the next person does not re-derive it.

## Verification

Local: `cargo fmt --check`, `cargo test -p calimero-server-primitives` (15 + 9), `UPDATE_FIXTURES=1` produces no diff. Before pushing I checked the action's actual download path rather than assuming — `releases/latest` resolves to `v0.6.56`, and both `linux-x86-64` and `linux-aarch64` binaries plus `.sha256` return 200.

The run to watch is the E2E: every namespace-join scenario failed at the join step on the first attempt, with `Client error: missing field \`governanceOp\``. Those passing is the proof the chain worked.

Issue #3126, opened 2026-07-01
DAG: unify the genesis-op parent convention (empty parents vs `[0; 32]`)


chefsale:
## Summary

The DAG carries **two conflicting conventions for a genesis (root) delta's parents**, and the disagreement leaves a small correctness/robustness hole that could not be closed in #3116.

- The DAG's own model treats the explicit zero hash `[0; 32]` as the genesis parent — see the `Genesis (zero hash) is always considered applied` branch in `DagStore::can_apply` (`crates/dag/src/lib.rs`).
- The unified-op layer creates genesis ops with an **empty** parent list (`parents: vec![]`) and relies on `all()` over an empty iterator being vacuously true so the op applies immediately as a root. Deltas mirror the op's parents verbatim (`CausalDelta::new(op.id(), op.parents.clone(), …)` in `crates/context/src/unified_applier.rs` and `crates/context/src/unified_op_store.rs`; genesis-parent derivation flows through `crates/context/src/governance_dag.rs::make_delta`).

## Why it matters

Because empty parents are a *legitimate* genesis signal in the unified-op layer, `can_apply` cannot distinguish a genuine genesis from a **malformed/hostile empty-parent delta**. That is exactly the reported finding #3116 could not fix:

> `[M] can_apply is vacuously true for empty parents` — a non-genesis delta with `parents: []` applies immediately as a disconnected head, bypassing the missing-parent machinery and polluting head computation.

In #3116 the obvious guard (reject empty-parent deltas) was implemented and then **reverted**, because it stranded legitimate genesis ops and broke the `calimero-context` suite. The current code documents the conflict inline in `can_apply` (see the `NOTE: an empty parent list is NOT rejected here` comment) rather than fixing it.

Net effect today: a corrupt or hostile peer can send an empty-parent, non-genesis delta and it applies as a disconnected head. Convergence still holds (it's deterministic), but the missing-parent/backfill path is bypassed and head computation is polluted.

## Proposed resolution

Standardize genesis ops on the DAG's existing sentinel: have the unified-op / governance write path stamp a genesis op with `parents: vec![[0; 32]]` instead of `parents: vec![]`. Then:

1. `can_apply` can treat an **empty** parent list as malformed and pend/reject it, closing the disconnected-head hole.
2. The `[0; 32]` genesis branch already in `can_apply` handles real roots, so genesis ops still apply immediately.
3. One convention across the DAG and the op layers.

[...]

Issue #2716, opened 2026-06-05
bug(storage): concurrent SharedStorage writer rotations produce non-convergent CRDT merge (HC skips 3 entities permanently)


chefsale:
## Reproduction

Test: `shared-storage-concurrent-rotation-converge` — fails intermittently when two nodes issue concurrent `shared_rotate_writers` calls while a third node is reconnecting.

Run 26979795218 (job 79617674965), logs artifact 7423437115.

## What happens

**Setup:**
- 3 nodes, all connected
- Two nodes issue concurrent `shared_rotate_writers` while the mesh is temporarily partitioned
- node-2 (reconnecting) calls `shared_rotate_writers` with 1 writer: `{"writers":["B8vqGQ6G5aiRgfapQv7awXvmaWnV4k7PiR2VAd1y3KfL"]}` at 21:23:55
- node-1 (other partition) calls `shared_rotate_writers` with 2 writers: `{"writers":["B8vqGQ6G5aiRgfapQv7awXvmaWnV4k7PiR2VAd1y3KfL","Eduac6WDaBvx7uKX9DkRUuG7EFZoMAZ8YDhWRTgvtkPQ"]}` at 21:24:16

**After reconnect:**
- node-1's delta arrives at node-2 as a concurrent branch (`parent_count=2`, `Concurrent branch detected - applying with CRDT merge semantics`)
- CRDT merge runs `__calimero_sync_next` at 21:24:25
- Merge produces hash `9fHFxAi6D5ZeeudFLyxnyG4srQEaeMP9ZSt9ximgj9dH` (22 entities) — nodes 1+3 converged on `8i45SJ46McsfjavV1W8uzYgrVXbmzbf6y3SpNCXJayNm` (2 entities as seen from node-2's POV)
- HC sync runs immediately: `nodes_compared=23, entities_merged=20, entities_pushed=0, nodes_skipped=3, root_hash_verified=true`
- **This repeats 200+ times over 180 seconds — node-2 never converges.**

## Root cause

HC finds **3 entities it cannot reconcile** and silently skips them (`nodes_skipped=3, deferred_root_merges=0`) **every single cycle**. These 3 entities are SharedStorage entries whose writer-set is determined by the divergent rotation log:

- node-2's rotation log: 1 granted writer (`B8vqGQ6...`)
- nodes 1+3's rotation log: 2 granted writers (`B8vqGQ6...` + `Eduac6WD...`)

[...]

PR #3547, opened 2026-08-18
fix(sync): authenticate state-delta parents by re-deriving the content address

### LINKED ISSUE #3540: State-delta `parent_ids` are unauthenticated: signed `delta_id` is never re-derived on receive
## Summary

A state delta's `parent_ids` are **not covered by any signature and never re-derived on receive**, so any peer that can publish on a context's gossipsub topic can take a legitimate member's delta, rewrite its parents, and have it accepted. Combined with the vacuous empty-parents branch in `DagStore::can_apply` (#3126), rewriting `parent_ids` to `[]` inserts the delta as a **disconnected DAG head**, bypassing the missing-parent/backfill machinery.

Everything else on the state-delta envelope is protected. This is the one malleable field that matters.

## What is and isn't bound today

| Field | Protection |
|---|---|
| `actions`, `root_hash`, `events` | AEAD-sealed inside `artifact` (`crates/node/src/handlers/state_delta/crypto.rs:129`) — needs the group key |
| `context_id`, `delta_id`, `author_id`, `governance_position` | Ed25519 envelope signature (`crates/node/primitives/src/sync/delta_auth.rs:64-71`) |
| **`parent_ids`** | **nothing** |
| `hlc` | nothing (and excluded from `compute_id` by design) |

`CausalDelta::compute_id` (`crates/storage/src/delta.rs:69`) *does* hash `parents` and `actions`, and `delta_id` *is* signature-bound — so the binding exists in principle. It just isn't enforced: `compute_id` has exactly two call sites, both on the **produce** side (`crates/context/src/handlers/execute/mod.rs:2275`, `crates/context/src/handlers/create_context.rs:468`). No receive path recomputes it and compares. `parent_ids` is destructured straight off the wire message (`crates/node/src/handlers/network_event.rs:93` → `crates/node/src/handlers/state_delta/mod.rs:156`) and used as-is.

## Who can exploit it

Gossipsub runs `MessageAuthenticity::Signed` (`crates/network/src/behaviour.rs:259`), so a relay cannot mutate a message in flight. The attack is **republish**, not tamper-in-transit: any node in the context's mesh re-publishes a modified copy under its own libp2p key, reusing the victim's `delta_id` + `delta_signature` verbatim.

The attacker needs **neither** the group key **nor** `author_id`'s identity key. Notably the existing gates do not catch it, because every one of them keys off `author_id` — the honest author — and not the publisher:

- `verify_delta_signature` (`state_delta/mod.rs:198`) passes: the payload it reconstructs contains no parents.
- The ReadOnly gate (`state_delta/mod.rs:263`) looks up `is_read_only_for_context(context_id, author_id)` → the honest author is a full member → passes.
- `membership_status_at(author, pos)` likewise resolves the honest author.

So a **ReadOnly or revoked member** — precisely the identity the ReadOnly gate exists to stop from writing — can replay any current member's delta with mutated parents. The same gap applies on the DAG-catchup receive path (`state_delta/mod.rs:1526`) and the buffered-replay path (`state_delta/mod.rs:1914`), which call the same signature check; a malicious catchup responder can serve mutated parents too.

## Impact

Convergence itself is not broken (application stays deterministic), but:

- **Disconnected heads.** `parent_ids: []` hits the vacuously-true branch of `can_apply` (`crates/dag/src/lib.rs:682`, and the `NOTE` at `:683` documenting exactly this) — the delta applies immediately as a root, skipping missing-parent detection and polluting head computation.
- **Causal-order subversion.** Parents are also the causal cut that at-cut authorization resolves against. Re-parenting a delta moves the cut it is authorized at.
- **Cheap and unattributable.** One republished gossip message; the forged delta is attributed to the honest `author_id`.

## Proposed fix

Recompute the content address on ingest and compare it against the signed `delta_id`:

```rust
// after decrypting actions, before the DAG insert
let recomputed = CausalDelta::compute_id(&parent_ids, &actions, &hlc);
if recomputed != delta_id {
    warn!(%context_id, %author_id, ?delta_id, "rejecting state delta — id does not address its content");
    return Ok(());
}
```

Properties:

- **No wire-format change, no convention change, no back-compat shim.** `compute_id` already hashes parents; honest producers already satisfy it. This is independent of #3126 and can land first.
- Closes empty-`parent_ids` forgery on the state path outright, which removes the remote-attacker pressure from #3126 and lets that convention cleanup be scheduled on its own merits.
- Must run **after** decryption (it needs `actions`) and **before** the DAG insert. Apply at all three receive sites: gossip (`state_delta/mod.rs`), DAG-catchup (`:1526`), buffered replay (`:1914`).

### Known limits of this fix — state them, don't oversell

- `compute_id` deliberately excludes `hlc` and does not cover `expected_root_hash`, so both stay malleable. `hlc` malleability affects LWW ordering; worth a follow-up decision (extend the preimage, or bind them into the envelope payload instead), not a blocker here.
- `compute_id` also excludes action metadata timestamps and ancestors for determinism reasons — the check inherits that coarseness. Those fields are inside the AEAD envelope, so they are group-key-gated regardless.

## Alternative considered

Add `parent_ids` to `DeltaSignaturePayload`. Rejected as the primary fix: it changes the signed payload (a wire break, and it would need `hlc` too to be coherent), whereas the recompute check reuses a binding that already exists and is already signed. Extending the envelope payload is the right vehicle for the `hlc` / `expected_root_hash` follow-up, where no existing binding can be reused.

## Acceptance criteria

- [ ] `compute_id` recomputed and compared against `delta_id` on all three state-delta receive paths, before the DAG insert.
- [ ] Regression test: a delta with `parent_ids` rewritten to `[]`, carrying a valid envelope signature, is rejected (not applied as a head).
- [ ] Regression test: `parent_ids` rewritten to a *different non-empty* set is also rejected — the check is about content-addressing, not just the empty case.
- [ ] Confirm the rejection path is a `warn!` + drop consistent with the neighbouring gates, not a `bail!` that could stall the DAG.
- [ ] Follow-up issue filed for `hlc` / `expected_root_hash` binding.

## References

- #3126 — the empty-parents `can_apply` hole this makes reachable; that issue is the convention cleanup, this one is the authentication gap.
- Anchors: `crates/node/primitives/src/sync/delta_auth.rs`, `crates/storage/src/delta.rs:69`, `crates/node/src/handlers/state_delta/mod.rs`, `crates/node/src/handlers/network_event.rs:93`, `crates/dag/src/lib.rs:682`.

chefsale:
Closes #3540. Removes the remote-attacker pressure from #3126 (which stays open as the genesis-convention cleanup).

## The gap

A state delta's `parent_ids` were covered by nothing:

| Field | Protection before this PR |
|---|---|
| `actions`, `root_hash`, `events` | AEAD-sealed inside `artifact` — needs the group key |
| `context_id`, `delta_id`, `author_id`, `governance_position` | Ed25519 envelope signature |
| **`parent_ids`** | **nothing** |
| `hlc` | nothing (and excluded from `compute_id` by design) |

The binding already existed in principle — `CausalDelta::compute_id` hashes `parents`, and `delta_id` *is* signature-bound — it just was never enforced. `compute_id` had exactly two call sites, both on the produce side.

Gossipsub runs `MessageAuthenticity::Signed`, so this was never tamper-in-transit. The attack is **republish**: any node in the context's mesh re-publishes a member's delta under its own libp2p key, reusing the victim's `delta_id` and `delta_signature` verbatim, with `parent_ids` rewritten. It needs neither the group key nor the author's identity key.

Every existing gate misses it because they all key off `author_id` — the honest author — not the publisher:

- `verify_delta_signature` passes: the payload it reconstructs contains no parents.
- `is_read_only_for_context(context_id, author_id)` resolves the honest author → full member → passes.
- `membership_status_at(author, pos)` likewise.

So a **ReadOnly or revoked member** — precisely the identity the ReadOnly gate exists to stop from writing — could re-parent a current member's write. Stripping `parent_ids` to empty is the worst case: it hits the vacuously-true branch of `DagStore::can_apply` and applies immediately as a disconnected head, bypassing missing-parent detection and polluting head computation.

## The fix

`CausalDelta::content_address_matches`, added next to `compute_id`, called on every receive path:

- gossip apply (`handlers/state_delta/mod.rs`) — after decrypt, before the DAG insert
- buffered replay — buffering is a delay, not a trust boundary
- both parent-fetch walks (`handlers/state_delta/mod.rs`, `sync/delta_request.rs`)
- DAG head pull (`sync/manager/mod.rs`)

Its parameter list mirrors `compute_id` exactly, **including the currently-unhashed `hlc`**, so the two cannot silently drift if the preimage ever changes.

## Second defect, same root cause

Three of the four catchup paths never checked that a responder returned the delta that was **asked for**. They built the DAG delta from the responder-supplied id while verifying the signature against either that claimed id or the requested one — so a responder holding one legitimately signed delta could answer a request for another and have it persisted under the wrong slot.

`sync/manager/mod.rs` already had this guard, and its comment claimed parity with *"the parent-fetch path's sanity check"* — a check that path never had. This PR makes the claim true.

Both structural checks sit **before** the genesis carve-out deliberately. That branch skips every author-keyed check by design (genesis genuinely predates any governance op, so there is no author to check), and `is_genesis_author_sentinel` only tests whether the author pubkey is all-zeros — a value the *responder* chooses to send. So today a malicious catchup responder can serve wholly arbitrary genesis content: parents, actions, hlc, root hash, even a different id. The carve-out is right in intent but substituted *nothing* for the checks it dropped; these two now give it its first integrity check. Since the requested id comes from the `parents` of an already-accepted delta, the pair chains content-addressing off a reference the node already trusts — no author required.

## Why this is safe for honest traffic

The way this could go wrong is false-positiving on honest producers, so I traced the write path rather than assuming:

- `sign_authorized_actions` mutates `actions` and re-serializes the artifact **before** `compute_id` runs, so the hashed actions are the transmitted actions.
- The rotation-log self-log leg (#2716 leg 4) recomputes `expected_root_hash` **after** `compute_id` — but `expected_root_hash` is outside the preimage. Pinned by `content_address_survives_post_id_root_hash_mutation`, since a check that folded the root hash in would reject every writer-set rotation.
- The broadcast reuses `the_delta.id` rather than recomputing.

No wire-format change, no back-compat shim, no convention change.

## Known limit, deliberately not fixed here

`compute_id` excludes the HLC for determinism, so `hlc` stays malleable. `content_address_does_not_cover_hlc` pins that as a **known limit rather than desired behaviour**, so the test failing is the signal to revisit. Binding `hlc` and `expected_root_hash` needs a wire-format decision and is tracked as a follow-up on #3540; folding it in here would have stalled the part that ships cleanly.

## Tests

New:
- 6 unit tests on the helper (`crates/storage/src/delta.rs`): honest delta and `[0;32]` genesis accepted; emptied parents, re-parented, added-parent, swapped-actions rejected; the two invariant-pinning tests above.
- 3 gate-level tests in `byzantine_delta_auth.rs` modeling the republish attack against a **real** `calimero-storage` Merkle tree — genuine signature, genuine `delta_id`, honest `author_id`, only `parent_ids` tampered → rejected, root hash provably unchanged. Plus an honest-delta control that must advance the root, so the negatives can't pass on a broken fixture.

Existing, all green: `calimero-storage --lib` 693, `calimero-node --lib` 495, `sync_sim` 325, `calimero-context --lib` 234, plus `dag_causal_shared_e2e`, `dag_persistence`, `dag_storage_integration`, `concurrent_branches_root_hash`, `sync_protocols`. `cargo fmt --all --check` and `clippy --all-targets` clean.

**Not covered:** no merobox e2e run against real nodes. The honest-path argument above is from tracing the write path plus in-process coverage; a multi-node scenario would be the stronger check, particularly for any cascaded-delta path that re-persists actions.

### DIFF
--- a/crates/node/src/handlers/state_delta/mod.rs
+++ b/crates/node/src/handlers/state_delta/mod.rs
@@ -411,6 +411,35 @@ pub(crate) async fn apply_authorized_state_delta(
         events,
     } = decrypt_delta_actions(artifact, nonce, group_key)?;
 
+    // Content-address check. The envelope signature above binds `delta_id`, but
+    // `parent_ids` and `hlc` ride the wire OUTSIDE it, and `actions` arrive
+    // sealed under a key every member holds. Re-deriving the id from the
+    // content it is supposed to address is what extends the signature's
+    // coverage to `parent_ids`: without it, anyone able to publish on the
+    // context topic can republish a member's delta with `parent_ids` rewritten
+    // and every downstream gate still passes, because they all key off
+    // `author_id` (the honest author) rather than the publisher. An empty
+    // `parent_ids` is the worst case — it applies immediately as a
+    // disconnected head, bypassing missing-parent detection entirely.
+    //
+    // Runs after decryption because it needs `actions`, and before the DAG
+    // insert below.
+    if !calimero_storage::delta::CausalDelta::content_address_matches(
+        &delta_id,
+        &parent_ids,
+        &actions,
+        &hlc,
+    ) {
+        warn!(
+            %context_id,
+            %author_id,
+            delta_id = ?delta_id,
+            parent_count = parent_ids.len(),
+            "Rejecting state delta — id does not content-address its parents/actions"
+        );
+        return Ok(());
+    }
+
     let delta = calimero_dag::CausalDelta {
         id: delta_id,
         parents: parent_ids,
@@ -1442,6 +1471,48 @@ async fn request_missing_deltas(
                         "Received missing parent delta"
                     );
 
+                    // Did we get the delta we asked for? Parity with the
+                    // head-pull path's identical guard in
+                    // `sync/manager/mod.rs`, whose comment already claims
+                    // parity with "the parent-fetch path's sanity check" —
+                    // a check this path never actually had. The envelope
+                    // signature below binds `storage_delta.id`, not
+                    // `missing_id`, so without this a responder holding a
+                    // legitimately signed delta can answer a request for
+                    // one id with a different delta and have it persisted
+                    // under the wrong slot.
+                    //
+                    // Placed BEFORE the genesis carve-out deliberately: that
+                    // branch skips every author-keyed check, so these two
+                    // structural checks are the ONLY thing standing between a
+                    // responder and arbitrary genesis content.
+                    if storage_delta.id != missing_id {
+                        warn!(
+                            %context_id,
+                            requested = ?missing_id,
+                            received = ?storage_delta.id,
+                            "parent-fetch: peer returned a different delta id than                              requested, dropping"
+                        );
+                        continue;
+                    }
+
+                    // And does that id content-address what arrived with it?
+                    // `missing_id` is not attacker-supplied — it came from the
+                    // `parents` of a delta this node already accepted — so
+                    // chaining these two checks authenticates the content
+                    // without needing an author, which is exactly what the
+                    // genesis branch lacks.
+                    if !storage_delta.id_matches_content() {
+                        warn!(
+                            %context_id,
+                            delta_id = ?missing_id,
+                            author = %response_author,
+                            parent_count = storage_delta.parents.len(),
+                            "parent-fetch: delta id does not content-address its                              parents/actions, dropping"
+                        );
+                        continue;
+                    }
+
                     // Genesis carve-out: the responder serves the
                     // genesis delta with the all-zeros sentinel
                     // `author_id` because the wire requires an author
@@ -2110,6 +2181,27 @@ pub async fn replay_buffered_delta(input: ReplayBufferedDeltaInput) -> Result<bo
         actions,
         events,
     } = decrypt_delta_actions(buffered.payload, buffered.nonce, group_key)?;
+
+    // Content-address check — parity with the gossip apply path. A buffered
+    // delta was received over the wire and carried through snapshot-sync
+    // buffering verbatim, `parent_ids` included, so it needs the same
+    // re-derivation the live path now runs; buffering is a delay, not a
+    // trust boundary that launders the fields the signature never covered.
+    if !calimero_storage::delta::CausalDelta::content_address_matches(
+        &buffered.id,
+        &buffered.parents,
+        &actions,
+        &buffered.hlc,
+    ) {
+        warn!(
+            %context_id,
+            delta_id = ?buffered.id,
+            author = %buffered.author_id,
+            parent_count = buffered.parents.len(),
+            "Rejecting buffered state delta — id does not content-address its parents/actions"
+        );
+        return Ok(false);
+    }
 
     let delta = calimero_dag::CausalDelta {
         id: buffered.id,
--- a/crates/node/src/sync/delta_request.rs
+++ b/crates/node/src/sync/delta_request.rs
@@ -98,6 +98,40 @@ fn verify_fetched_parent(
 ) -> VerifiedParent {
     use calimero_context_config::types::GovernanceParentEdge;
 
+    // Did we get the delta we asked for? `verify_delta_signature` below is
+    // called with `delta_id` (the id we REQUESTED) while the caller persists
+    // `fetched.delta.id` (the id the responder CLAIMS) — so without this guard
+    // the signature is verified against one delta and a different one lands in
+    // the DAG. Parity with the head-pull path's guard in `sync/manager/mod.rs`.
+    //
+    // Both structural checks sit BEFORE the genesis carve-out deliberately:
+    // that branch skips every author-keyed check, so these are the only thing
+    // standing between a responder and arbitrary genesis content.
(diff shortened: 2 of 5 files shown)


chefsale:
Follow-up filed as #3552 — binds `hlc` and `expected_root_hash`, the fields this PR deliberately scoped out. Notes there: `hlc` **cannot** go into the `compute_id` preimage (its exclusion is what makes delta ids deterministic), so the envelope payload is the only coherent place to bind it; and the catchup/parent-fetch paths are structurally weaker than gossip, since they serve a plaintext `CausalDelta` while gossip seals actions and root hash under the group key.

Also recorded there: the `expected_root_hash` docstring in `crates/storage/src/delta.rs` ("receiving nodes MUST use this hash rather than their computed hash") is stale and contradicted by `delta_store.rs` ("CRITICAL: We must store the computed hash, NOT delta.expected_root_hash!"). That downgrades the field's malleability from root-poisoning to `merged_deltas` bookkeeping perturbation.

PR #3554, opened 2026-08-18
refactor(api)!: drop upgradePolicy from the namespace and group responses

chefsale:
The follow-up #3485 left open. **Blocked on** calimero-network/mero-js#104 and calimero-network/swift-sdk#22 — merge those first.

## Why now

`UPGRADE_POLICY_COMPAT` fills `upgradePolicy` with the constant `"LazyOnAccess"` on four endpoints. The concept went server-side in #3393; the key survived for clients that declared the field required. Its own doc named the exit condition:

> Delete once merobox ships a client-py built off post-merge master and `MIN_MEROBOX` (`.github/actions/setup-merobox`) names it.

Both halves now hold: merobox 0.6.56 bundles `calimero-client-py` 0.6.29, built off master, and #3528 moved `MIN_MEROBOX` to 0.6.56.

## Endpoints

`GET admin-api/namespaces`, `.../namespaces/:id`, `.../namespaces/for-application/:id`, `admin-api/groups/:id`.

## Requests are untouched

A released node still declares the field **required** on `POST admin-api/namespaces` and `POST admin-api/groups` and rejects a body without it. Both SDKs keep sending it. That shim is gated on which node releases are supported, not on which clients exist, so it has a longer life than this one and is deliberately out of scope.

## Ordering, and why it is shorter than #3485's

The two hand-written SDKs go first, because they typed the field required — swift-sdk especially, where a synthesized `Codable` fails the *whole* response with `keyNotFound`.

There is no compiled-client gate this time: core's struct has carried `#[serde(default)]` on the field since before rc.21, so `calimero-client-py` has always tolerated its absence, and merobox reads namespace data as plain dicts. That is the difference from `governanceOp`, where the client-py struct required the field and took a four-repo sequence to unblock.

## Scope check

The only other `upgrade_policy` left in the tree is a store-layer test pinning that a record written before the *stored* format dropped the tag fails loudly rather than decoding into garbage — a different concern, kept.

Two related findings, handled elsewhere:

- **swift-sdk had a dead `updateGroupSettings`** PATCHing `/admin-api/groups/{id}`; core has no PATCH routes at all. Removed in swift-sdk#22 along with `UpdateGroupSettingsRequest`, whose every field (`upgradePolicy`, `requester`) had already been deleted from core.
- **merobox needed nothing** — it dropped the whole upgrade-policy surface in merobox#318.

## Verification

`cargo fmt --check`, `cargo check --workspace --all-targets`, `cargo test -p calimero-server-primitives -p calimero-server` all green. `UPDATE_FIXTURES=1` leaves the wire fixtures unchanged — no committed fixture covered these endpoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@
 
 ### Removed
 
+- **`upgradePolicy`** from the namespace and group-info responses (`GET
+  admin-api/namespaces`, `.../namespaces/:id`, `.../namespaces/for-application/:id`,
+  `admin-api/groups/:id`), along with the `UPGRADE_POLICY_COMPAT` constant that
+  filled it. The concept went server-side in #3393; the key survived only for
+  clients that declared it required, and its own doc named the condition for
+  removal — a merobox bundling a client-py built off post-merge master, with
+  `MIN_MEROBOX` naming it. merobox 0.6.56 and client-py 0.6.29 satisfy both.
+  It always held `"LazyOnAccess"`, so no caller can depend on its value.
+  Requests are unaffected: released nodes still require the field on
+  `POST admin-api/namespaces` and `POST admin-api/groups`, and both SDKs keep
+  sending it ([#3485])
+
 - **`meroctl context identity grant` / `revoke`** and the per-context
   capability request/response types behind them. The commands were shipped and
   advertised in `context --help`, but had no route, client method or handler and
--- a/crates/server/primitives/src/admin/mod.rs
+++ b/crates/server/primitives/src/admin/mod.rs
@@ -14,11 +14,6 @@ use url::Url;
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 pub struct Empty;
 
-/// Constant compat key: the released calimero-client-py merobox bundles rejects
-/// group/namespace responses that omit it. Delete once merobox ships a client-py
-/// built off post-merge master and MIN_MEROBOX (.github/actions/setup-merobox) names it.
-pub const UPGRADE_POLICY_COMPAT: &str = "LazyOnAccess";
-
 // -------------------------------------------- Application API --------------------------------------------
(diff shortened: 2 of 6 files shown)


chefsale:
Unblocked — rebased onto master and out of draft. Both prerequisites are in:

| link | what | state |
| --- | --- | --- |
| calimero-network/mero-js#104 | `upgradePolicy` → optional on `Namespace` / `GroupInfo` | merged |
| calimero-network/swift-sdk#22 | same, plus the dead `updateGroupSettings` | merged |
| calimero-network/swift-sdk#23 | 18 request types stop declaring the long-dead `requester` | merged (stacked under #22) |

## A correction worth recording

The SDK PRs do **not** remove the field — they make it optional, and that reversal came from mero-js's live wire-contract suite:

```
node sent 'upgradePolicy' but SDK type Namespace does not declare it (listNamespaces)
```

That suite runs against a **released** merod and fails in both directions. So this is the mirror of the `governanceOp` ordering: there the SDK had to drop the field *before* core stopped sending it, or a decoder requiring it broke with `keyNotFound`; here the SDK must keep *declaring* it until no supported release still sends it. Same shape of change, opposite constraint — what differs is which side declares the field required.

Optional satisfies both ends: it decodes against a released node that sends `"LazyOnAccess"` and against a node that has this PR. The SDKs can drop it outright once the release floor moves; both carry a comment saying so.

## Scope confirmed while sequencing

- **merobox needed nothing** — it dropped the whole upgrade-policy surface in calimero-network/merobox#318. My earlier claim that it had a broken `--upgrade-policy` flag was read off a stale local branch.
- **No compiled-client gate**, unlike #3485: core's struct has carried `#[serde(default)]` on the field since before rc.21, so `calimero-client-py` has always tolerated its absence and merobox reads namespaces as plain dicts.

`cargo fmt --check` and `cargo check --workspace --all-targets` clean after the rebase.

PR #3573, opened 2026-08-20
fix(api): let a group add name a member by account

rtb-12:
## Summary

`POST /admin-api/groups/:group_id/members` was the only verb on this resource typed to a signing key. The other eight name an `AccountId`, and the `GET` listing returns one, so a caller that lists a group's members and tries to add one of them to a subgroup gets:

```
HTTP 400  members[0].identity: buffer provided to decode base58 encoded string into was too small
```

Nothing in the API converts between the two id spaces, so any app with nested membership - rooms, channels, teams, shared folders - cannot populate a sub-scope from its own roster.

The key typing was justified by an add whose subject holds no account on this node yet, but `member_account::require` refuses exactly that case, so the endpoint could never serve it.

### Changes

- `MemberIdentity` (`calimero-primitives`): an account or a key, parsed from whichever the string is. The two cannot collide - an account renders as 64 hex characters and bs58 over 32 bytes reaches at most 44.
- `GroupMemberApiInput.identity` and `AddGroupMembersRequest` carry it; the apply resolves a key through `member_account::require` as before and takes an account as given.
- Key delivery follows the id it was given: a key names its one holder, an account is addressed through its live devices - the same device-first rule the scope-key fan-out uses, so a revoked device is not handed the key back. Without this the membership row would be correct while every group op stayed undecryptable until the joiner-side pull ran.
- `meroctl group members add` accepts either rendering.

Existing bs58 callers are unaffected.

### Not covered here

`calimero-client-py` validates this field as a bs58 public key client-side, so merobox cannot yet express an account-typed add. A merobox scenario for it needs that binding widened first.

## Test plan

Reproduced end to end against two local `merod` nodes - namespace, subgroup, the account read back out of `GET /groups/:namespace_id/members`, then the same request against the same subgroup on each binary:

| | request | result |
| --- | --- | --- |
| before | `{"members":[{"identity":"<account>","role":"Member"}]}` | `400 … buffer provided to decode base58 encoded string into was too small` |
| after | identical | `200`, member present in `GET .../members` |

Delivery was verified functionally, not just by status code: node-2 received the subgroup key (no `envelope addresses another device; ignoring`), joined a context in that subgroup, and read back state node-1 wrote. Adding by bs58 key still returns `200`; a malformed id still returns `400`.

Automated:

- `crates/context/src/handlers/add_group_members.rs` - three cases over `key_deliveries`: an account fans out to its live devices as `EnvelopeRecipient::Device`, an account with no binding delivers nothing, a key is addressed directly as `EnvelopeRecipient::Member`. Each was observed failing under mutation before being kept.
- `crates/primitives` - both encodings parse, round-trip through serde, and are shown not to collide.
- `crates/server/primitives` wire fixture carrying both id spaces in one request, so a field that stops accepting either reddens the wire-contract gate.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`, `cargo test --workspace` all pass locally.


chefsale (approved):

crates/context/src/handlers/add_group_members.rs
```diff
@@ -33,14 +39,16 @@ impl Handler<AddGroupMembersRequest> for ContextManager {
...
-                    // joined the namespace earlier, so asking the subgroup would
-                    // refuse every legitimate add.
-                    let member_account =
-                        crate::member_account::require(&datastore, &group_id, identity)?;
+                    // Resolution runs at the NAMESPACE: a direct add targets a
+                    // subgroup and names somebody who joined the namespace
+                    // earlier, so asking the subgroup would refuse every
+                    // legitimate add.
+                    let member_account = match identity {
+                        MemberIdentity::Account(account) => *account,
```
xilosada, line 55: **Design confirmation (non-blocking).** Both reviewer profiles independently flagged this asymmetry, so it's worth a sentence from you.

The `Key` arm below goes through `member_account::require`, which fails closed when the key is bound to no account — `account_bindings.rs` documents that contract explicitly ("None must fail closed ... Callers must refuse rather than fall back to a key-derived stand-in"). The `Account` arm takes `*account` verbatim: no existence, binding, or endorsement check. `member_added.rs::apply` doesn't check either — it validates permissions and writes the member row.

Net effect: a garbage or never-enrolled 64-hex account can be written as a permanent group member. The caller gets `Ok`, the only signal is the `warn!` further down, and that member can never receive the group key because there's no device to wrap for.

Your `an_account_with_no_device_gets_nothing` test says this is deliberate, and it probably has to be — that's the point of naming an account that hasn't converged locally yet. If so, could the doc comment on `GroupMemberApiInput::identity` say so as plainly as the `Key` refusal path does? As written it reads as an oversight against a fail-closed invariant the codebase documents elsewhere.

crates/context/src/handlers/add_group_members.rs
```diff
@@ -113,3 +124,145 @@ impl Handler<AddGroupMembersRequest> for ContextManager {
...
+                        sk,
+                        binding.device,
+                        &X25519PublicKey::from(binding.kem_pk),
+                        &group_id.to_bytes(),
+                        group_key,
+                    )?,
+                    binding.sign_pk,
+                ))
+            })
+            .collect(),
```
xilosada, line 181: **Robustness (non-blocking).** This `.collect()` into `EyreResult<Vec<_>>` combined with the `?` inside the closure makes the whole fan-out all-or-nothing: one device whose `wrap_for_device` fails (stale or malformed `kem_pk`, key-agreement error) aborts `key_deliveries` entirely, and the caller's `Err(e) => warn!(...)` arm then skips *every* device of that member — including the ones that would have wrapped fine.

So a member with several live devices can end up with zero deliveries because of one bad binding, silently degrading all of them to the joiner-side pull instead of just the broken one. A `filter_map` that logs per-device failures would contain the blast radius to the device that actually failed.

crates/context/src/handlers/add_group_members.rs
```diff
@@ -113,3 +124,145 @@ impl Handler<AddGroupMembersRequest> for ContextManager {
...
+    group_id: &ContextGroupId,
+    group_key: &[u8; 32],
+) -> EyreResult<Vec<(KeyEnvelope, PublicKey)>> {
+    match identity {
+        MemberIdentity::Key(key) => Ok(vec![(
+            GroupKeyring::wrap_for_member(sk, key, &group_id.to_bytes(), group_key)?,
+            *key,
+        )]),
+        MemberIdentity::Account(_) => AccountBindingRepository::new(datastore)
+            .devices_of(ns_id, member_account)?
```
xilosada: **Efficiency (non-blocking).** `key_deliveries` is called once per entry of `members`, and `devices_of` does a full `live_bindings(group)` scan then filters by account. For a batch add that's a full rescan of the binding column per member — m×a — which is the exact pattern the sibling `live_devices_by_account` was added to replace (its doc comment calls this out by name). `AddGroupMembersApiRequest.members` is only checked non-empty, so there's no bound on the batch.

Hoisting a single `live_devices_by_account(ns_id)` call outside the per-member loop and looking each member up in the resulting map would collapse it to one scan.

xilosada (approved):
Approving. Ran a two-profile automated review (security + logic) over `origin/master...HEAD`; after confidence filtering the consolidated verdict was **no findings** — nothing here blocks the merge.

Verified clean:
- `MemberIdentity::FromStr` is unambiguous — a 64-hex `AccountId` can never collide with bs58 (max 44 chars), and the dedicated test covers it.
- `devices_of`/`live_bindings` correctly excludes revoked and epoch-superseded devices before wrapping, and `kem_pk` is sourced from the folded binding row rather than the wire, matching `wrap_for_device`'s documented requirement.
- Fan-out wiring is right: `wrap_for_member` vs `wrap_for_device`, `EnvelopeRecipient::Member` vs `::Device`, and `required_signers = [recipient]`.

Three non-blocking notes inline — one design-confirmation, two robustness. Happy for them to land as follow-ups.

crates/context/src/handlers/add_group_members.rs
rtb-12: Confirmed deliberate, and now said plainly on the field rather than left to inference.

An account this node has not converged on is exactly what a key could never name, so requiring a local binding here would put back the case the key typing claimed to serve and could not. The cost is a mistyped account writing a membership row nobody holds, removable the same way it was added, and the add is admin-gated. The doc on `GroupMemberApiInput::identity` now states the asymmetry and why it is not an oversight.

crates/context/src/handlers/add_group_members.rs
rtb-12: Fixed - `key_deliveries` now returns a plain `Vec` and drops a failed recipient with a warning instead of aborting the batch.

Worth recording what it takes to reach that branch: `SharedKey::new` refuses a non-decompressible or small-order point, so the identity arm is reachable and is now covered by `a_recipient_that_cannot_be_wrapped_for_is_dropped_not_propagated` (verified failing under mutation). The device arm agrees over X25519, which accepts any 32 bytes, so I could not construct a failing `wrap_for_device` - that path is shaped correctly but untested, and I would rather say so than imply coverage.

crates/context/src/handlers/add_group_members.rs
rtb-12: Fixed - one `live_devices_by_account(ns_id)` before the loop, looked up per member. `ns_id` was resolved per member too and is now hoisted with it.

Neither is affected by the ops the loop publishes: `MemberAdded` writes no binding and does not rotate the key, so the map is stable across the batch.

PR #3574, opened 2026-08-20
feat(admin-api)!: an add may name a member by account

chefsale:
Reported by a user building mero-chat DMs on `0.11.0-rc.24`. A DM is a Restricted
subgroup, and adding the other person to it was not expressible. Their read of
the code was correct.

## Explain it like I'm five

A group is a clubhouse. There are two ways to name a person:

- **their name** (`AccountId`) — who they are
- **their door key** (`PublicKey`) — one key they happen to carry, and they may carry several

**Before.** The add door asked for **the key**. Its sign said: *"maybe this person
is a stranger with no name here yet — but you're holding their key, so use the
key."* Then, one step inside, the code looked the key up in the name book and
**refused if the name wasn't already written there**. So the stranger the sign
promised to help could never actually get in. Everyone who got in already had a
name.

Meanwhile the person doing the adding only ever gets shown **names** — the member
listing returns accounts, and nothing anywhere hands out the keys behind one. So
the door demanded the one thing it never gives you, in order to serve a case it
rejects. Passing the name got a 500.

And when the door did let someone in, it mailed them the group's secret **to the
one key that was typed** — no check that the key still belonged to them, and
nothing sent to their other devices.

**After.** The door takes **either** the name or a key. They can't be mixed up,
because a name is always 64 characters long and a key is always shorter, so
neither can be read as the other. Give it a name and the secret is mailed to
**every device that person currently has**, to the address each device published
itself, and to none that were taken away from them. If they have no devices right
now, nothing is mailed — rather than mailing it to a device that was taken away.

## What was actually wrong

Two comments in `rc.24` pulled opposite ways, and the handler won:

- `GroupMemberApiInput` said the field is a key *"because an add is the one call
  whose subject may not have an account here yet."*
- `add_group_members.rs` called `member_account::require`, which errors unless the
  key is already bound to an account (`account_bindings.rs`).

Both cannot be true. The doc's rationale described a path the handler had closed,
and `member_account::require`'s own docs say as much: *"an unresolvable key is a
genuine anomaly here."* So the key was protecting nothing that the account does
not protect better — while making the ordinary case (add a known namespace member
to a subgroup) impossible, since `GET .../members` renders accounts and no
endpoint discloses the keys behind one.

The key had one job left, and it was the wrong one. After resolving the account,
the add called `wrap_for_member` — the last identity-addressed key delivery in
the tree. `group_keys.rs` documents that exact thing as the hazard the device rows
exist to prevent:

> A member whose every device has been revoked or superseded receives **nothing**
> … any fallback to identity addressing would hand the fresh key straight back to
> it.

So the add wrapped to whatever Ed25519 key the caller typed, under no revocation
check, and only to that one — meaning a multi-device member got delivery to one
device and the rest waited on the joiner-side pull.

## What this changes

**`identity` takes a `MemberPrincipal`** — `Account(AccountId)` (64 hex) or
`Key(PublicKey)` (base58). Disambiguation is by construction, not by luck:
`Hash::from_str` refuses anything longer than the widest base58 32 bytes can
produce (45 chars), and an account is always 64, so neither encoding parses as the
other principal. Key-named adds are unchanged on the wire.

**Account-named adds fan out over devices.** `live_devices_by_account` +
`wrap_for_device`, sealed to each device's certified X25519 key read from the
folded binding — the row that supplies the key is the row that says the device is
still authorized, so a revoked device's key is not available to wrap with. An
account with no live device is delivered nothing rather than falling back to
identity addressing; the member row still lands and the key follows by pull. One
binding scan per call, not per member (`devices_of`'s docs warn about exactly
that loop).

The addressee decision is a pure function (`plan_deliveries`) so it is testable
without standing up an actor.

**Deleted the stale paragraph** that cost the reporter an afternoon.

### Also fixed

Four `meroctl` commands hex-encoded a *signing key* into an `:account` path
segment — `group members set-role`, `set-caps`, `group metadata member get` /
`set`. Valid hex, so `parse_account` accepted it, naming a principal that exists
nowhere. `crates/server/src/admin/handlers/groups.rs` has a test asserting a bs58
key must **not** parse as an account, for precisely this reason; the CLI defeated
it by re-encoding the bytes. **Breaking** for anyone scripting those four: pass
the account. No previously-working input regresses — the old one could not reach a
real member.

## Testing

12 new tests. The four that carry the argument:

- a key-named member is addressed as the key the caller gave
- an account-named member reaches every device that person holds
- only the named person's devices are addressed
- an account with no live device is delivered **nothing** — not an
  identity-addressed fallback

Plus principal parse / round-trip / mutual-exclusion, API-level acceptance of the
exact body the reporter sends, refusal of an unrecognized encoding naming both
encodings, and a client transport test asserting the account reaches the wire as
hex.

Full suites for the five affected crates pass, `cargo check --workspace
--all-targets`, `cargo fmt --check` and clippy are clean.

## What is NOT proven yet

There is no end-to-end run behind this, and the blocker is worth stating plainly:
merobox needs no change (its `add_group_members` step passes `members_json`
through verbatim), but `calimero-client-py` (`src/client.rs:2289`) does

```rust
let identity = identity_str.parse::<identity::PublicKey>().expect("invalid identity");
```

An account string does not 500 there — it **panics across the FFI boundary**. So
the account path cannot be exercised from a scenario until client-py is rebuilt
against this, via the usual core → client-py → merobox chain. That `expect` is a
defect in its own right (any caller typo panics instead of raising `ValueError`)
and belongs in its own client-py PR rather than folded in here.

Separately, and untouched: the same reporter says invitation + `POST /groups/join`
against a subgroup also 500s. That needs its own issue and repro.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 ## [Unreleased]
 
+### Added
+
+- **`POST admin-api/groups/:group_id/members` accepts an account.** The
+  `identity` field now takes either an account (64 hex characters, exactly what
+  `GET .../members` returns) or a signing key (base58) — the encoding says
+  which, and no string is valid as both, so a mix-up is still a parse error
+  rather than a different principal. Key-named adds are unchanged on the wire.
+
+  Adding an existing namespace member to a Restricted subgroup was previously
+  not expressible: the caller knows that person from a member listing, which
+  renders accounts and never discloses the keys behind one, while the endpoint
+  demanded a key and then refused any key not already bound here. The field's
+  own documentation claimed the key existed to serve subjects with no account
+  yet, which the handler had made unreachable.
+
+  Naming an account also delivers the group key to **every live device** that
+  person holds, sealed to each device's certified X25519 key, under the same
+  revocation rules the rotation fan-out enforces. A key names one device, so a
+  key-named add still delivers to that one and leaves any others to pull.
+
 ### Removed
 
 - **`upgradePolicy`** from the namespace and group-info responses (`GET
@@ -41,6 +61,18 @@
   bundles that client — so `MIN_MEROBOX` moves to 0.6.56 here, the floor at
   which a join against a node without the field still decodes ([#3485],
   [#3528])
+
+### Fixed
+
+- **`meroctl group members set-role` / `set-caps` and `meroctl group metadata
+  member get` / `set` took a signing key where the route names an account.**
+  Each hex-encoded the key's raw bytes into the `:account` path segment, which
+  parses cleanly and names a principal that exists nowhere, so the commands
+  addressed no member and reported success or a confusing miss. They now take
+  the account, as the routes and the member listing do. **Breaking** for anyone
+  scripting those four commands: pass the 64-hex account rather than the base58
+  key. There is no input that worked before and stops working — the previous one
+  could not reach a real member.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@
   Naming an account also delivers the group key to **every live device** that
   person holds, sealed to each device's certified X25519 key, under the same
   revocation rules the rotation fan-out enforces. A key names one device, so a
-  key-named add still delivers to that one and leaves any others to pull.
+  key-named add still delivers to that one and leaves any others to pull
+  ([#3574])
 
 ### Removed
 
@@ -72,7 +73,7 @@
   the account, as the routes and the member listing do. **Breaking** for anyone
   scripting those four commands: pass the 64-hex account rather than the base58
   key. There is no input that worked before and stops working — the previous one
-  could not reach a real member.
+  could not reach a real member ([#3574])
 
 ### Changed
 
@@ -850,3 +851,4 @@ Integrations:
 [#3485]: https://github.com/calimero-network/core/issues/3485
 [#3528]: https://github.com/calimero-network/core/pull/3528
 [#3530]: https://github.com/calimero-network/core/pull/3530
+[#3574]: https://github.com/calimero-network/core/pull/3574
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 ### Added
 
+- **`GET admin-api/groups/:group_id/devices`** — the device bindings of the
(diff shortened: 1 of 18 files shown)


chefsale:
Reopened, with one commit added on top of the two that were already here.

`03f4d51fc` — **`GET admin-api/groups/:group_id/devices`**. It was sitting uncommitted in a worktree; the CHANGELOG entry and the admin-api docs table were already written against this PR number, which is why this reopens rather than opening a new one.

It closes the gap the first commit leaves behind. `POST .../members` accepts a key only because key→account resolution had no home in the API — the endpoint was doing double duty as the sole converter. `?member=` narrows in either encoding, so an account lists the devices that speak for it and a signing key lists the device presenting it. With that, the key form on the add is a back-compat shim rather than load-bearing, and can go once client-py and the merobox scenarios name accounts.

No KEM key is exposed, and reads are gated on the node belonging to the group — the same gate the member listing uses.

Branch merges cleanly with master (44 commits behind, no conflicts). `cargo fmt --check` and `cargo check --workspace` pass on the branch; a compile check of the merged state is running locally. Note the original blocker is unchanged: the e2e is still gated on client-py's `.expect("invalid identity")` panic.

chefsale:
Correcting my previous comment — I reopened this on a wrong premise, and it should go back to closed.

**Both commits here are superseded.**

- `3cae2af2b` (an add may name a member by account) landed on master as **#3573** (`808dbcdbc`), with the account-or-key type named `MemberIdentity` rather than this branch's `MemberPrincipal`. Same feature, different name.
- `03f4d51fc` (the devices listing) is largely covered by master's `GET /groups/:group_id/member-devices` → `list_member_devices`, which already returns `signingKey` per device grouped by account. So key→account resolution is already reachable from the API.

The merge confirms it rather than contradicting it: 10 conflicted files, including `crates/primitives/src/identity.rs`, because master and this branch each added their own spelling of the same type.

The one real delta left is `?member=` narrowing — master's `ListMemberDevicesQuery` carries only `offset`/`limit`, so today a caller resolving a key lists and filters client-side. That is a small filter on an endpoint that already exists, and it belongs in its own PR against `member-devices` rather than in a branch whose other half is a duplicate of #3573.

Closing again. `03f4d51fc` is pushed on `addbyaccount` so the work isn't lost.

PR #3589, opened 2026-08-21
refactor(projection): one ancestry walk, and a verdict that names why a cut is unusable

chefsale:
The first of the semantic findings from reading `calimero-dag` and `calimero-projection` end to end — the class of defect no detector finds.

## The problem

`calimero-projection` walked a cut's causal ancestry in **three** places — the same `by_id` map, the same `visited` set, the same `VecDeque`:

| | on a missing ancestor |
| --- | --- |
| `acl_view_at:719` | **silently skips it** and folds a truncated view |
| `cut_ancestry_complete:796` | returns `false` |
| `cut_ancestry_decoded:834` | returns `false` |

The duplication is not the interesting part. `acl_view_at` silently folding a truncated ancestry is right for a legitimately out-of-slice cross-scope edge and wrong for a same-scope gap — so `cut_ancestry_complete` exists *only* to tell a caller whether the view it just built can be trusted. Its own doc says so.

**Nothing made that second call mandatory.** A caller that skipped it received a plausible, possibly-stale membership view with no signal. The two gating sites in `calimero-context` did make it — by walking the same ancestry **twice per at-cut read**, on the authorization path:

```rust
if !proj.cut_ancestry_complete(&scope, &heads) { return None; }
let view = proj.acl_view_at(&scope, &heads)?;
```

## The change

`ScopeState::cut_ancestry` is now the only walk in the crate, returning a `CutAncestry` that carries the ops **and** what stopped them from being a complete, readable history. `acl_view_at_complete_cut` answers both questions in one call that cannot hand back a view without the verdict — which is what removes the duplicated traversal.

## The verdict deliberately does not short-circuit

This is the design detail worth reviewing. The old presence check traversed *through* an undecoded (`Noop`) ancestor looking for a genuine gap; the readable check stopped at the first one.

An enum verdict returning whichever problem the queue reached first would **report a cut as complete because it never looked past the placeholder.** So `CutAncestry` records `first_missing` and `first_opaque` independently, and there's a test building exactly that shape:

```
an_opaque_op_in_front_of_a_gap_is_both_undecoded_and_incomplete
```

Separating them also fixes the second half of the finding: `false` used to mean either *"an ancestor is absent, sync must fetch it"* or *"an ancestor is present but undecryptable, and will resolve itself when the key arrives"*. The docs explained these mean different things; the caller could not tell them apart. Same conflation split in **#3571** and **#3586**, in a third place.

## No new lax/precise pair

`acl_view_at_complete_cut` deliberately does **not** sit beside a default that returns the view regardless. That is how `dag::add_delta` ended up with 94 callers against `add_delta_with_outcome`'s one — the lossy default is what people reach for. `acl_view_at` remains only for readers that legitimately fold a partial ancestry, and the doc comment says so.

## Behaviour-preserving, measured not asserted

`ancestry_oracle` keeps all three original walks **transcribed verbatim** and sweeps them against the new one over every small DAG — depths 1–4 × every undecoded mask × every dropped mask — asserting the folded op order, the presence verdict and the readable verdict all match. The oracles share no code with the crate on purpose, and the module says not to refactor them; an oracle that shares code with what it checks checks nothing.

## Cost, stated plainly

**Production code +26 lines**, plus a **157-line oracle module** of deliberately duplicated test code. This is not a line-count win. It is three walks becoming one, two traversals becoming one on the auth path, and two causes no longer collapsed into a bare `false`.

## Found and deliberately NOT changed

Neither belongs in a refactor meant to preserve behaviour, but both are worth recording:

- **`auth_cut_context_or_cause` is a third double-walk site.** Collapsing it strands `UndecidableCause::ViewUnavailable`, which I verified is constructed at exactly one place and **unreachable there** — the completeness check already returns `false` for an unfed scope, so the `ok_or` can never fire. That variant is effectively dead and its metric label can never be emitted. Fixing it means deleting a variant or changing a metric, so it wants its own decision.
- **Five other `acl_view_at` callers don't gate on completeness** while their siblings do: `device_account_at_cut`, `role_at_cut`, and three predicates. Whether each should is a per-call-site semantic question — and it's evidence the hazard above has already been exercised. No behaviour changed here.

## Verification

- `cargo test --workspace --features calimero-storage/testing` — **4968 passed, 0 failed**, including `projection_membership_equivalence` (10), which is what would catch a walk divergence
- `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` — clean
- `cargo fmt --check` — clean
- Exactly one ancestry walk remains in the crate (grep-verified)

### DIFF
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -453,6 +453,37 @@ impl ScopeProjections {
         Some(ScopeState::acl_view_at(log, parents))
     }
 
+    /// The at-cut ACL view, but **only** when the cut's whole causal ancestry is
+    /// folded — `None` for an unfed scope or a truncated one.
+    ///
+    /// Every authoritative reader wants exactly this pair of facts, and used to get
+    /// them as `cut_ancestry_complete(..)` followed by `acl_view_at(..)`: two calls
+    /// that walked the same ancestry twice per read, with nothing making the first
+    /// one mandatory. A caller that skipped it received a view folded over a
+    /// truncated history — plausible, possibly stale, and silent about it. One call
+    /// that cannot return the view without the verdict removes both the second walk
+    /// and the chance to forget.
+    ///
+    /// Deliberately *not* offered beside a laxer default that returns the view
+    /// regardless: an imprecise default with a precise opt-in is how the
+    /// distinction gets lost, since the default is what callers reach for.
+    /// `acl_view_at` remains for the readers that legitimately fold a partial
+    /// ancestry — a cross-scope edge out of the slice — and those are not
+    /// authoritative-grant paths.
+    #[must_use]
+    pub fn acl_view_at_complete_cut(
+        &self,
+        scope: &ScopeId,
+        parents: &[[u8; 32]],
+    ) -> Option<calimero_authz::AclView> {
+        let log = self.logs.get(scope)?;
+        let walked = ScopeState::cut_ancestry(log, parents);
+        if !walked.is_complete() {
+            return None;
+        }
+        Some(ScopeState::acl_view_from_ancestry(&walked))
+    }
+
     /// The convergence `scope_root` for `scope`, folding this scope's ACL +
     /// governance (membership / admin / policy / subgroups) onto the
     /// **externally-supplied storage Merkle `entities_root`**. `None` when the
@@ -804,10 +835,8 @@ impl ScopeProjections {
         let scope = ScopeId::from(namespace_id);
         // Defer to live on a partial fold (governance-backfill race) rather than
         // returning an under-counted cohort — same guard as the membership gate.
-        if !proj.cut_ancestry_complete(&scope, &heads) {
-            return None;
-        }
-        let view = proj.acl_view_at(&scope, &heads)?;
+        // One walk for both questions — see `acl_view_at_complete_cut`.
+        let view = proj.acl_view_at_complete_cut(&scope, &heads)?;
         let mut out = std::collections::BTreeSet::new();
         for group in groups {
             out.extend(Self::member_accounts_in_view(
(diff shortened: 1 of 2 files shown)


chefsale:
Rebased onto `master` (`407f3e35a`). The conflict was substantive rather than textual, so worth recording what changed.

## #3587 added a fourth copy of the walk

While this was in review, **#3587 landed `cut_covers`** — a fresh copy of the same BFS this PR exists to unify. That is the argument for the PR making itself: a fourth copy arrived within days, written by someone with no reason to know the other three existed.

So it is folded in rather than merged around. `cut_covers` needed one thing the others did not — the set of ids the walk **reached**, absent ones included, since a cited id is an ancestor whether or not its record is present. That set is the walk's own dedup set, previously built and dropped, so `CutAncestry` now carries it at no cost.

Its original is transcribed as a **fourth oracle**, and the sweep checks the coverage verdict too — over every frontier the chain can pose, one naming an id nowhere in it, and the empty frontier. `the_one_walk_matches_all_three_it_replaced` → `..._all_four_...`.

## The e2e failure on the previous head was #3587's bug, not this PR's

The pre-rebase run failed `Test (group-private-add-then-write)` on the projection-divergence gate:

```
plane="membership"  projected=Some(Member)  live=Some(Admin)
```

That is **the identical failure #3587 fixed** — same scenario, same marker shape, and it was failing on `master` before that fix landed. This branch simply predated it. The rebase incorporates the fix, and folding `cut_covers` into the shared walk preserves its semantics while putting them under an oracle.

I checked this rather than assuming: #3587's own description quotes the same marker from the same scenario.

## Corrected numbers

Three figures in the commit message were stale once master's baseline moved, and are fixed rather than shipped:

| | was | now |
| --- | --- | --- |
| production code delta | +26 | **+21** |
| oracle module | 157 lines | **186 lines** |
| walks unified | three | **four** |

## Re-verified after the rebase

- exactly one production ancestry walk remains in the crate (the other three matches are the oracle transcriptions)
- `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` — clean
- `calimero-projection` 12 + 28 tests, `calimero-context` 240 — pass
- full workspace suite re-running locally as a cross-check

PR #3591, opened 2026-08-21
ci: ratchet near-duplicate functions so cleanups stop being undone

chefsale:
Duplication here doesn't arrive by decision. It arrives because a new collection type, op variant or plane gets added by copying the nearest existing one, by an author with no way to know how many copies already exist.

Two measured cases from the last few days:

- `js_collections.rs` had reached 5,817 lines with **a third of its function bodies redundant** across 26 clone families (#3576 removed 908 lines of it).
- A **fourth** copy of the projection ancestry walk landed in #3587 while #3589 — the PR unifying the first three — was still in review.

So every cleanup is undone by attrition unless something holds the line. `cargo machete` (#3586) now does that for dependencies, where 35 had accumulated, three of them documented as "key deps" in their own crate's AGENTS.md. Nothing did it for duplication.

## A ratchet, not a threshold

The workspace has **330 accepted pairs across 25 crates**. Demanding zero would be absurd; demanding that a crate not grow a *new* one is enforceable today. A cleanup that lands gets locked in by refreshing the baseline — and the gate tells you when pairs have disappeared, so wins don't silently un-bank.

```
crates/runtime  134      crates/node      20
crates/auth      45      tools/merodb     17
crates/context   32      crates/storage   21
```

## It reports pairs, never verdicts

The match is syntactic, so it **cannot** tell a genuine copy from two operations that merely look alike — it groups `crdt_map_get` with `crdt_authored_map_owner_of`, which ask different questions of different planes.

So a new pair fails the build with both names printed, rather than being auto-rejected on a score. Fold it, or run `--update` and say in the PR why the two are honest neighbours. Either answer is fine; the point is that somebody looked.

Pairs are keyed by `(crate, name, name)` and never by line number, so an unrelated edit above a function doesn't churn the baseline. Renaming a function does show up — a one-line baseline edit with a visible cause.

Follows the pattern the repo already uses twice (`check-scenario-coverage.py`, `lint-e2e-convergence.py`, each with a committed baseline) and runs in **~83 s**, in line with those.

## The detector had a bug only a mutation test could find

Worth reading, because it nearly shipped.

The extractor located bodiless declarations by searching for the first `;` after a function name. But `[u8; 32]` contains one — and that is how every hash and key in this codebase is spelled. So:

```rust
fn get_missing_parents(&self, query_limit: usize) -> Vec<[u8; 32]> {
//                                                        ^ read as end of signature
```

read as a declaration, and the function was skipped. **7 of 24 in `dag/src/lib.rs`; more than half in some files.**

I found it by planting a deliberate near-copy of `DagStore::get_missing_parents` under a new name and observing that the gate **did not flag it**. A gate verified only by passing on a clean tree would have shipped blind and gated nothing — the same failure mode as the orphaned scenarios `check-scenario-coverage.py` exists to catch.

Signature termination is now found at bracket depth zero. Two consequences I'd rather state than bury:

- The **"576 clone pairs" figure in my earlier survey was an undercount.** The individual findings stand — those came from reading code — but the aggregate was wrong.
- The first baseline, generated with the buggy scanner, read **318** against the corrected **330**.

## Verified by mutation, not by passing

| | |
| --- | --- |
| clean tree | exit 0 — `no new near-duplicate pairs (330 baselined across 25 crates)` |
| planted near-copy | exit 1 — flags **exactly** `crates/dag: get_absent_parents_copy <-> get_missing_parents`, nothing else |

Plant reverted; `git diff -- '*.rs'` is empty. **No Rust changes in this PR.**

## Usage

```bash
python3 scripts/check-duplication.py             # the gate
python3 scripts/check-duplication.py --update    # re-baseline after a cleanup
python3 scripts/check-duplication.py --list crates/storage
```


chefsale:
Rebased onto `master` (now including #3589), and **the gate immediately earned its keep on its own PR** — worth recording, because it changed a design decision.

## It noticed #3589's win

```
::notice::2 baselined pair(s) are gone — nice:
  crates/projection: cut_ancestry_complete <-> cut_ancestry_decoded
  crates/projection: cut_ancestry_complete <-> cut_covers
  Lock the win in with `scripts/check-duplication.py --update`
```

Exactly the duplication #3589 removed. That is the ratchet doing the thing it exists for: a cleanup that lands gets banked so it cannot quietly come back.

## And it flagged my own oracles, which forced a decision

```
::error::2 new near-duplicate function pair(s)
  crates/projection: reference_complete <-> reference_covers
  crates/projection: reference_complete <-> reference_decoded
```

Those are #3589's **oracle transcriptions** — pre-refactor walks kept verbatim precisely so they share no code with what they check. Syntactically they are duplicates; semantically they are the technique that makes the refactor verifiable.

I could have baselined them with a reason, which is the flow the gate advertises. Measuring first was better: **1,894 of 4,180 extracted bodies (45%) are inline `#[cfg(test)]` code.**

So test code is now excluded, and it is a decision rather than an oversight:

- Inline `#[cfg(test)]` modules are stripped alongside the `/tests/` trees already skipped — scanning one and not the other was an inconsistency, not a choice.
- A test legitimately repeats structure, one input varied at a time. Flagging that buries the production copies the gate is for.
- This workspace duplicates in tests *deliberately* (`inheritance_climb.rs`, `ancestry_oracle`). A gate that fought the oracle pattern would be arguing against the thing making these refactors checkable.

## Result: a smaller, more meaningful baseline

| | before | after |
| --- | --- | --- |
| pairs | 330 across 25 crates | **148 across 14** |
| runtime | ~83 s | **~37 s** |

And the distribution is now informative rather than noisy:

```
101  crates/runtime          6  crates/context/primitives
 12  crates/storage          6  crates/node
  9  crates/governance-store 3  crates/client
```

**101 of 148 are `crates/runtime`** — the read-path half of `js_collections.rs` that #3576 deliberately left, because the detector groups `crdt_map_get` with `crdt_authored_map_owner_of` and folding by similarity there would merge different operations. The gate's largest number pointing straight at known, argued-for work is the intended outcome.

Mutation check re-run against the new baseline: clean tree exits 0; a planted near-copy of `DagStore::get_missing_parents` exits 1 naming exactly that one pair. Plant reverted, no Rust changes in this PR.

chefsale:
Closing — not worth the friction it would impose. Recording the reasoning so anyone who finds this later doesn't have to redo the analysis.

## What was actually verified

The central claim held up under test. Baseline generated at `af04b4858` (the commit before #3587), then run at #3587:

```
::error::1 new near-duplicate function pair(s)
  crates/projection: cut_ancestry_complete <-> cut_covers
```

So it *would* have caught the fourth ancestry walk. It also noticed #3589's cleanup and prompted banking it, and its distribution was informative — 101 of 148 pairs in `crates/runtime`, which is exactly the `js_collections.rs` read-path tail #3576 deliberately left.

## Why that isn't enough

**It freezes the status quo rather than improving it.** 165 baselined pairs means it accepts nearly all existing duplication; the entire value is about future growth, and the evidence for that is **one** instance.

**The escape hatch is one command.** `--update` silences everything wordlessly. This repo's existing gate — `scenario-coverage-baseline.json` — stores `{path: reason}` and *requires* a reason, precisely so an exclusion has to be argued rather than accumulated. A bare list of 165 pairs has no such pressure, so the realistic outcome is a red CI, an `--update`, and a moved-on developer. That is how the "key deps" documented in three crates' AGENTS.md went stale without anyone noticing.

**And the cost is borne by everyone.** Renames churn the baseline, false positives are inherent (the match is syntactic — it groups `crdt_map_get` with `crdt_authored_map_owner_of`, different questions about different planes), and it adds ~37 s to every PR.

By contrast the `cargo machete` gate in #3586 had far stronger evidence behind it — 35 accumulated unused dependencies, three of them documented as "key deps" — and no judgement call per hit.

## Worth salvaging from this

Two things this work produced that live on regardless:

- **A bug in the clone detector used for the wider audit.** It located bodiless declarations by searching for the first `;` after a function name, but `[u8; 32]` contains one — and that is how every hash and key in this codebase is spelled. So `fn f(..) -> Vec<[u8; 32]> {` read as a declaration and the function was skipped: 7 of 24 in `dag/src/lib.rs`, more than half in some files. **The "576 clone pairs" figure quoted in the audit is an undercount.** Found only by planting a deliberate duplicate and noticing the gate did not fire.
- **The same-name-across-files refinement.** Skipping every same-name pair hid the biggest duplication pattern here — `entry` in both map files, `index_marker_current` across `sorted_map`/`sorted_set`, `watch_app_and_update_context` in two meroctl subcommands. Anyone re-running that analysis should compare same-name pairs across files and skip them only within one file.

The script is on the closed branch `chore/duplication-ratchet` if it is ever wanted as a one-off analysis tool rather than a gate.

PR #3599, opened 2026-08-21
feat(sdk): expose the calling account to the guest as env::caller_account()

rtb-12:
## Summary

An application's `env::account_id()` returns the account of the node that executed a call, never the account of whoever made it. The caller's account is already resolved server-side to run the membership gate, then discarded; the execute handler independently derives a different account from this node's own device binding.

An app gating per-member permissions therefore tests this node's identity rather than the requester's:

```rust
// Reads as "is this node the owner?", not "is this caller the owner?"
if AccountId::from(env::account_id()) != self.owner {
    app::bail!("only the owner may reset");
}
```

**This is latent, not live.** No shipped auth path produces a second principal — `user_password` is the only provider and stores a username where a public key would go, so every call today resolves to `NodeOwner`. The defect matters because multi-principal is planned: the moment a node serves more than one account, every already-deployed app with a check like the above starts granting the node owner's rights to everyone it serves, with no code change and no error.

## What changes

| Host function | Returns | Purpose |
| --- | --- | --- |
| `env::device_id()` | node's replica id | CRDT mechanics. Unchanged. |
| `env::account_id()` | account this write is attributed to | Storage stamps, owner gates. Unchanged. |
| `env::caller_account()` | account the caller is authorized as | Permission checks. New. |

The executor and the account stay as they were, and neither can stand in for the caller: the node holds no private key but its own (`get_identity` requires `private_key: Some(_)`), and a delta is attributed by its signer, so an op cannot be labelled with an account this node holds no binding for.

`caller_account()` returns `Option`, absent whenever nobody in particular asked — an xcall hop, or a caller authorized by a route that names no account. Propagating the outer caller across an xcall would let any context in the namespace act under a principal that never addressed it (`#[app::xcall]` defaults to `AnyInNamespace`); falling back to the node's account would reproduce the original defect. Absence is therefore distinguishable from every real account and means "do not authorize".

Every call today is `NodeOwner`, which resolves to `None`, so no deployed app changes behaviour.

Also removes the `TODO` in `execute.rs` proposing the caller's key be passed as the executor. That cannot work: the node would need that key's private half, so such a call fails with `Unauthorized` on first use.

## Rollout constraint

This adds a WASM host import. **Nodes must ship before any app built against the new SDK** — an app importing `caller_account` cannot instantiate on a node that does not offer it (`LinkError { UnknownImport }`). New nodes run old apps fine; the incompatibility is one-directional. Merobox cannot catch this, since every node in a run is the same build against fresh state.

## Test plan

Reproduction — the two resolutions disagree for the same call:

```
caller key             : 5TeWSsjg2gbxCyWVniXeCmwM7UtHTCK7svzJr5xYJzHf
authorized as          : a71c3dd073939d81f972525e9788c6b958818e33f30674ef43b4184eef40aa50   <- resolved, then dropped
node's own account     : ae8b82beb817b8bb8cf5beb2e41f04f5ac504e1d83c96258a64ec616fcef282f
guest env::account_id(): ae8b82beb817b8bb8cf5beb2e41f04f5ac504e1d83c96258a64ec616fcef282f
```

| Test | Covers |
| --- | --- |
| `server::execute::tests` (2) | The gate surfaces the account it authorized instead of dropping it; a node owner surfaces none rather than borrowing one. Driven over a real `ContextClient` and store. |
| `governance-store` | The authorized account and the account this node runs as are different values, and both are real. |
| `runtime` host functions (2) | A present caller is written and reports `1`; an absent one reports `0` and leaves the register untouched. |
| `runtime` integration (WAT) | A real guest declaring `(import "env" "caller_account" ...)` links and reads the caller. The only test that exercises the import table. |
| `apps/scaffolding-e2e` | `my_caller()` beside `my_account()`, so merobox runs the path through a genuinely SDK-built app. |

Every gate was confirmed failing before being believed:

| Sabotage | Result |
| --- | --- |
| gate drops the resolved account (**exactly what master does**) | server gate test failed |
| `NodeOwner` presents an account instead of abstaining | server owner test failed |
| host fn falls back to the node's account on absence | absence test failed |
| host fn returns the node's account (the original bug) | both host tests failed, WAT guest trapped |
| import removed from the host table | `LinkError { Import("env", "caller_account", UnknownImport) }` |

### What the tests do not cover

- **No end-to-end merobox scenario for the bypass.** One cannot be written: `ExecutionRequest` carries no caller field, so a scenario's identity comes entirely from its auth token, and `login` (username/password) is the only auth verb available. Every merobox call is `NodeOwner`. The `executor_public_key` field in scenario YAML selects which *node* to send to, not which principal — it never reaches the wire request.
- **The 4 hops between the gate and the runtime** (`ContextClient` → actor → `internal_execute` → `run_with_origin`) are straight parameter passing, type-checked but not directly asserted.

Verification, in a clean target directory: `calimero-runtime` 154, `calimero-governance-store` 586, `calimero-context` 240, `calimero-server` 105, plus `sys` / `sdk` / `context-client` — all passing. `cargo fmt --check` clean. `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` clean. `cargo mero build` on `apps/scaffolding-e2e` succeeds, with the import present in the emitted wasm and `my_caller` in the embedded ABI.

### DIFF
--- a/apps/scaffolding-e2e/src/lib.rs
+++ b/apps/scaffolding-e2e/src/lib.rs
@@ -1265,6 +1265,17 @@ impl E2eKvStore {
         Ok(AccountId::from(env::account_id()).to_string())
     }
 
+    /// The account that ASKED for this call, hex-encoded, or `None` when nobody
+    /// in particular did.
+    ///
+    /// Exists for the e2e: the distinction from `my_account` is invisible from
+    /// outside the guest, so a scenario needs both to prove the node stopped
+    /// answering "who asked" with its own identity. A direct node-owner call has
+    /// no separate caller, so `None` here is the expected result today.
+    pub fn my_caller(&self) -> app::Result<Option<String>> {
+        Ok(env::caller_account().map(|caller| AccountId::from(caller).to_string()))
+    }
+
     /// Add a writer, by ACCOUNT — the writer set grants people, so one grant
     /// covers every device they hold.
     pub fn shared_add_writer(&mut self, writer: AccountId) -> app::Result<()> {
--- a/crates/context/primitives/src/client/mod.rs
+++ b/crates/context/primitives/src/client/mod.rs
@@ -4,6 +4,7 @@ use std::sync::Arc;
 
 use async_stream::try_stream;
 use borsh::BorshDeserialize;
+use calimero_account::AccountId;
 use calimero_context_config::types::{ContextGroupId, InvitationFromMember, SignedOpenInvitation};
 use calimero_node_primitives::client::NodeClient;
 use calimero_primitives::application::ApplicationId;
@@ -1603,6 +1604,11 @@ impl ContextClient {
     /// # Returns
     ///
     /// A `Result` containing the `ExecuteResponse` on success, or an `ExecuteError` on failure.
+    ///
+    /// Attributes **no caller**: the guest reads `env::caller_account()` as absent. Every caller
+    /// here is system-initiated (delta apply, sync, cascade, event dispatch). A transport that
+    /// authorized a real principal must call [`execute_with_origin`](Self::execute_with_origin)
+    /// and pass it, or the account it checked is silently dropped.
     pub async fn execute(
         &self,
         context_id: &ContextId,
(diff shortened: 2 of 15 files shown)


rtb-12 (commented):
## 🤖 AI Code Reviewer

**Reviewed by 2 agents** | Review time: 0.0s

---

### ✅ No Issues Found

---

<sub>🤖 Generated by [AI Code Reviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-300a8234`</sub>

PR #3604, opened 2026-08-21
ci(e2e): make the projection gate report its own coverage

chefsale:
Third and last piece of the divergence-gate thread (#3587 closed the false
positives, #3593 the drain coverage). This one closes the way the gate could stop
checking anything without going red.

## The problem

The gate asserts that no `unified_projection_divergence` marker appears. A
scenario that made no comparison at all satisfies that exactly as well as one
that made hundreds — the two are the same silence in the logs.

#3587 made that reachable rather than theoretical: comparisons off live's
frontier are now correctly suppressed, and suppression is invisible to the gate.
If comparisons stopped concluding for some systemic reason — a change in apply
ordering, a restart pattern, buffering under load — the gate would go quiet
across all 88 scenarios at once and CI would stay green while checking nothing.
That is a gate that fails OPEN, which is the one property a HARD gate must not
have.

## The fix

**`feat(context)`** — every membership op a node folds logs
`unified_projection_compare` with what the comparison did: `agree`, `diverged`,
`skipped_stale_cut`, `skipped_undecoded`, `skipped_unfed`, or `no_live_row` (the
expected model difference where the live plane holds no direct row for an
inherited or open-join member). The divergence marker is unchanged.

The classification is a function over an explicit `ComparePremise` rather than a
match buried in the apply path, because the property that matters is negative: a
comparison may call itself CONCLUDED only when it was fed, its ancestry decoded,
live held a row, and the cut reached live's frontier.

**`ci(e2e)`** — the gate reads those results, prints the histogram on every run,
and fails a scenario that attempted comparisons and concluded none. A scenario
with no membership ops is untouched (both counts zero, nothing to check).

## Tests

`only_a_complete_premise_yields_a_conclusive_comparison` walks all 32 premises
and asserts the equivalence in both directions, plus that every label is one the
gate knows how to count — a new label returned here and not taught to the gate
would restore the same silence. Negative control: let an off-frontier agreement
count as `agree` and the test fails naming that exact premise.

The workflow shell was run against captured logs rather than reasoned about: the
real failing run from the original report, plus fixtures for healthy, blind,
empty and divergent, checking **exit codes** and not just messages — blind and
divergent exit 1, healthy and empty exit 0. The fixtures carry real ANSI escapes,
since that is what defeated the previous version of these greps.

## What to watch in this PR's e2e

The new hard failure is calibrated on reasoning — the admin node folds at the
frontier with a live row, so every membership scenario should conclude at least
one comparison. This run over all 88 scenarios is the experiment. If some
scenario legitimately refuses every comparison, it will say so here, and the
answer is to make that case observe-only with the reason recorded rather than to
widen the gate silently.

---

> [!NOTE]
> **Medium Risk**
> Touches the governance apply shadow path and a HARD e2e gate: misclassified compare results or a too-strict coverage floor can fail the whole suite or hide real divergence. Does not change authorization or live membership writes.
> 
> **Overview**
> Stops the unified-op projection gate from going green when it compared nothing. Absence of `unified_projection_divergence` used to look the same as “no comparison happened.”
> 
> Every folded membership op now emits `unified_projection_compare` at INFO with an explicit result (`agree`, `diverged`, `skipped_*`, `no_live_row`). Only a complete premise (fed, decoded, live row, at frontier) counts as conclusive; the divergence warn is unchanged.
> 
> Per-scenario CI prints a result histogram and notices thin coverage. A new `projection-coverage` job fails if the whole suite concluded zero comparisons (or logs are missing after the matrix ran). An exhaustive unit test pins that only a complete premise yields `agree`/`diverged`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5591d43b7352ea81765db4990d9451863223e4a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -1025,6 +1025,10 @@ jobs:
       # sits here instead: across every scenario, the gate has to have concluded
       # at least one comparison, or it was not gating.
       - name: Download scenario logs
+        id: logs
+        # Tolerated so the assertion below runs and can say precisely WHY there
+        # are no logs; a bare failure here would be indistinguishable from a
+        # suite that legitimately produced none.
         uses: actions/download-artifact@v8
         with:
           pattern: logs-*
@@ -1034,12 +1038,20 @@ jobs:
 
       - name: Assert the suite concluded comparisons
         run: |
-          # No logs at all means the matrix never ran (build failure, cancelled
-          # run). That is the `test` job's business to report, not this one's —
-          # failing here too would only add noise to an already-red run.
+          # Missing logs is benign in exactly one case: the matrix never ran, so
+          # there was nothing to check and the `test` job already says why. A
+          # download failure, or a matrix that ran and left nothing behind, is
+          # the fail-open case this job exists to prevent — it must not pass as
+          # "nothing to inspect".
           if [ ! -d all-logs ] || [ -z "$(find all-logs -type f -print -quit)" ]; then
-              echo "::notice::no scenario logs to inspect — the matrix did not run"
-              exit 0
+              case "${{ needs.test.result }}" in
+                skipped|cancelled)
+                  echo "::notice::the scenario matrix did not run (${{ needs.test.result }}) — nothing to inspect"
+                  exit 0
+                  ;;
+              esac
+              echo "::error::no scenario logs to inspect though the matrix ran (test=${{ needs.test.result }}, artifact download=${{ steps.logs.outcome }}) — the coverage floor could not be checked"
+              exit 1
           fi
           strip() { sed -E 's/\x1b\[[0-9;]*m//g'; }
           compares=$(grep -rh "unified_projection_compare" all-logs/ 2>/dev/null | strip || true)
@@ -1058,16 +1070,16 @@ jobs:
 
   comment:
     name: Report Failed Workflow
-    needs: [test, auth-seam]
+    needs: [test, auth-seam, projection-coverage]
     runs-on: ubuntu-latest
     if: always() && github.event_name == 'pull_request'
     steps:
       - name: Check if main job failed or was cancelled
         id: check_status
         run: |
-          if [ "${{ needs.test.result }}" != "success" ] || [ "${{ needs.auth-seam.result }}" != "success" ]; then
+          if [ "${{ needs.test.result }}" != "success" ] || [ "${{ needs.auth-seam.result }}" != "success" ] || [ "${{ needs.projection-coverage.result }}" != "success" ]; then
             echo "job_failed=true" >> $GITHUB_OUTPUT
-            if [ "${{ needs.test.result }}" == "cancelled" ] || [ "${{ needs.auth-seam.result }}" == "cancelled" ]; then
+            if [ "${{ needs.test.result }}" == "cancelled" ] || [ "${{ needs.auth-seam.result }}" == "cancelled" ] || [ "${{ needs.projection-coverage.result }}" == "cancelled" ]; then
               echo "was_cancelled=true" >> $GITHUB_OUTPUT
             else
               echo "was_cancelled=false" >> $GITHUB_OUTPUT
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -850,14 +850,16 @@ jobs:
               exit 1
           fi
 
-          # A scenario with no membership ops at all is fine (both counts zero);
-          # what fails is having attempted comparisons and concluded none of
-          # them. `no_live_row` counts as neither — it is the expected model
-          # difference for inherited members, not a refusal.
+          # Observe-only per scenario, and deliberately so. A scenario can hold
+          # a single membership op that lands on a node whose ancestry is not yet
+          # decrypted — one refusal, no conclusions — which is thin coverage, not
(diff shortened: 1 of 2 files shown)


chefsale:
## Calibration result — the per-scenario assertion was too sharp

The first full run (once the merobox `capability` breakage was out of the way via
#3603) failed 9 of 88 scenarios on the new assertion. Not a bug in the change: a
genuinely thin scenario. `group-subgroup` produces exactly ONE membership
comparison for the whole run —

```
membership comparisons:
      1 result="skipped_undecoded"
```

— an admin-push add folded on a node whose cited ancestry was not yet decrypted.
One refusal, no conclusions. Calling that "the gate checked nothing" is true of
the scenario and useless as a signal.

So the floor moved to where the failure mode actually lives (270b341d): a
suite-wide assertion in a new `projection-coverage` job that fails only if the
gate concluded no comparison **anywhere** in the run. No quiet scenario can trip
it, and the systemic case it exists for — the marker path breaking, comparisons
ceasing to conclude across the board — still fails the run. Per scenario the
histogram prints and an all-refused scenario emits a notice, so thin coverage
stays visible without being fatal.

## What the histograms turned up, which is worth more than the gate

Counting comparisons showed the gate's reach is much narrower than it looked.
The compare only fires for `OpPayload::MemberAdded` and `MemberRemoved`:

```rust
let membership = match &shadow_op.payload {
    OpPayload::MemberAdded { group, member, .. }
    | OpPayload::MemberRemoved { group, member } => Some((*group, *member)),
    _ => None,
};
```

Every invitation join folds to `MemberJoinedWithDevice` (membership *and* the
device credential as one fact), and that arm is not matched — so **no invitation
join has ever been compared**, on any scenario, since the shadow was added. The
gate has been checking admin-push adds and removals only, which is why
per-scenario counts are in the single digits.

Widening the compare to the join payload belongs in its own PR, not this one: it
is a behaviour change on the plane most likely to surface real disagreements, and
mixing it in here would make it impossible to tell which change caused which
failure. Filed as the follow-up this instrument was built to find.

chefsale:
## Green — and here is what the gate has actually been checking

151 checks, 0 failures. The 9 scenarios that failed the sharp version now pass
with a notice, and `projection-coverage` passes on real data. The suite-wide
histogram, first measurement ever taken of this gate:

```
suite-wide membership comparisons:
     15 result="agree"
     52 result="skipped_undecoded"
```

Nothing else: no `diverged`, no `skipped_stale_cut`, no `skipped_unfed`, no
`no_live_row`.

Three things follow, and none of them were visible before this PR:

**The gate concludes 15 comparisons across all 88 scenarios.** That is the whole
evidentiary basis of a HARD gate that gates the unified-causal-log cutover. It is
not zero — the assertion in this PR is meaningful — but "the projection reproduces
the live decision across every membership plane" is a much larger claim than 15
checks support.

**78% of attempts are refused for decrypt lag.** Every refusal is
`skipped_undecoded`: the cited ancestry contains an op this node cannot decrypt
yet, so the folded membership is provisional and the comparison abstains. That is
pre-existing and unrelated to the frontier gate — worth its own investigation,
since 52-of-67 abstentions is a lot of governance history sitting undecryptable at
apply time.

**`skipped_stale_cut` is zero.** The suppression #3587 introduced — the thing that
made this fail-open hole reachable — is not hiding anything in practice, now that
#3593 folds drained deltas and the originator feeds its own DAG. Good to know
rather than assume.

Combined with the join-payload gap in the comment above (`MemberJoinedWithDevice`
is never compared, so no invitation join has ever been checked), the follow-up
worth doing is widening what the shadow compares. That is where the coverage is,
and it is now measurable instead of assumed — re-run this job after that change
and the 15 should move.

chefsale:
## Bugbot review — two fixed as stated, one fixed on a different premise (5591d43b)

**"Coverage markers invisible at default log level" (High) — premise incorrect, risk real, fixed.**
e2e nodes do not default to `calimero_=info`. merobox's bootstrap config defaults
`log_level` to `"debug"` (`merobox/commands/bootstrap/config.py:1946`), and the run
that went green on this branch counted 67 of these lines suite-wide — 15 `agree`,
52 `skipped_undecoded`. So the gate is not blind today.

The risk underneath it is real, though: 2 of 86 scenarios override `log_level`
(one to `"info,calimero_storage=debug,…"`), which silences the instrument in that
scenario without silencing anything else — and a gate whose evidence depends on
per-scenario log config is back to trusting silence. The marker is now emitted at
INFO. At 67 lines across an entire suite that costs nothing, and it makes the
coverage signal independent of how a scenario configures logging.

**"Coverage gate fails open" (Medium) — correct, fixed.**
`continue-on-error` plus "no logs → exit 0" also covered a failed download and a
matrix that ran and left nothing, which is exactly the fail-open case this job
exists to close. Missing logs now pass only when `test` was `skipped` or
`cancelled`; anything else fails and names both the matrix result and the download
outcome. `continue-on-error` stays on the download step deliberately, so the
assertion runs and can say WHY the logs are missing rather than dying
indistinguishably.

**"Failure comment ignores coverage" (Low) — correct, fixed.**
`comment` now takes `projection-coverage` in `needs` and in its failure check.

Each branch was exercised against fixtures rather than reasoned about:
matrix-ran-but-no-logs exits 1, matrix-skipped exits 0, logs-with-a-conclusion
exits 0.

PR #3613, opened 2026-08-21
fix(context): compare an invitation join like any other membership op

chefsale:
The follow-up #3604 was built to find. Its coverage instrument measured the
divergence gate concluding **15 comparisons across an entire 88-scenario e2e
suite** — and showed why: the membership compare matched only `MemberAdded` and
`MemberRemoved`.

Every invitation join folds to `MemberJoinedWithDevice` (membership and the
joiner's device credential as one indivisible fact), which fell into the
catch-all arm. So **no invitation join has ever been compared**, on any scenario,
for as long as the shadow has existed. The gate has been checking admin-push adds
and removals — the least common way anyone becomes a member.

## Why a join is exactly as comparable

| | live plane | projection |
| --- | --- | --- |
| writes the row via | `MembershipRepository::add_member` | `fold_member_added` |
| role from | `role_from_invited_role(inv.invited_role)` | `GroupMemberRole::from_invited_role(…invited_role)` |

Same derivation from the same invitation on both sides. There was a directly
comparable answer sitting there the whole time and nothing looked at it.

`MemberJoinedOpen` stays out, and that is a judgement rather than the same
oversight: an open-subgroup self-join is a PROOF of inheritance, live writes no
direct row for it, and the projection models it as inherited too — there is
nothing to compare. The doc comment says so at the call site.

## Shape

`membership_touched` gives the extraction a name and a test, so which payloads
are comparable stops being a match arm that can be forgotten. It also returns the
op kind, which the compare marker carries and both histograms now break down:

```
suite-wide membership comparisons:
   3 result="agree"
   2 result="skipped_undecoded"
by op kind:
   2 op_kind="add"
   3 op_kind="join"
conclusive by op kind:
   1 op_kind="add"
   2 op_kind="join"
```

A total that hides which planes ever conclude is how this stayed invisible, so
the gate now reports per kind.

## The experiment

This PR's e2e run is the measurement. Master's baseline is 15 `agree` / 52
`skipped_undecoded`, and every scenario joins a namespace, so the conclusive count
should rise sharply and `op_kind="join"` should appear.

Two outcomes are both informative and neither is a reason to hold the PR:

* the count rises and everything agrees — the gate now checks the plane it always
  claimed to;
* some comparisons come back `diverged` — then the widening has found real
  disagreements on the join plane, which is what a gate is for. I would rather
  learn that from a red run here than keep not asking.

Verified locally: 244 context lib tests plus every integration suite, clippy and
fmt clean, and the histogram shell exercised against fixtures with real ANSI
escapes.

---

> [!NOTE]
> **Medium Risk**
> Log-only shadow compare and CI coverage, not live apply. Widening the gate to joins can surface real divergences (or more skips) and fail e2e that previously passed on silence.
> 
> **Overview**
> The unified-op membership shadow now treats invitation joins (`MemberJoinedWithDevice`) as comparable, same as admin-push adds and removals. Previously those joins fell through the catch-all, so the e2e coverage gate never checked the most common way members appear.
> 
> `membership_touched` names the comparable payloads and tags each compare with `op_kind`. CI histograms now break down results by kind so a total cannot hide an untested plane. Open-subgroup self-joins stay out: they write no direct live row.
> 
> Ancestry refusals are split: missing history is `skipped_incomplete`, unreadable ops remain `skipped_undecoded`, via a single `cut_ancestry_state` walk. Tests pin the join arm and the new premise.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74b44b983998c88b801160d897bd1711707065f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -834,6 +834,10 @@ jobs:
           echo "membership comparisons:"
           if [ -n "${compares}" ]; then
               echo "${compares}" | grep -oE 'result="[a-z_]+"' | sort | uniq -c
+              # By op kind too: adds, removes and joins are separate planes of
+              # coverage, and a gate that only ever concludes on one of them is
+              # thinner than its total suggests.
+              echo "${compares}" | grep -oE 'op_kind="[a-z]+"' | sort | uniq -c
           else
               echo "  (none — this scenario folded no membership op)"
           fi
@@ -1058,6 +1062,11 @@ jobs:
           echo "suite-wide membership comparisons:"
           if [ -n "${compares}" ]; then
               echo "${compares}" | grep -oE 'result="[a-z_]+"' | sort | uniq -c
+              echo "by op kind:"
+              echo "${compares}" | grep -oE 'op_kind="[a-z]+"' | sort | uniq -c
+              echo "conclusive by op kind:"
+              echo "${compares}" | grep -E 'result="(agree|diverged)"' \
+                | grep -oE 'op_kind="[a-z]+"' | sort | uniq -c || true
           else
               echo "  (none)"
           fi
--- a/crates/context/src/handlers/apply_signed_namespace_op.rs
+++ b/crates/context/src/handlers/apply_signed_namespace_op.rs
@@ -228,9 +228,15 @@ fn membership_touched(
 struct ComparePremise {
     /// The op reached the projection (the lock was not poisoned).
     fed: bool,
-    /// Every op in the cut's ancestry is decrypted, so the folded membership is
-    /// final rather than provisional.
-    decoded: bool,
+    /// Every op in the cut's ancestry is PRESENT in the log. A gap means this
+    /// node has not folded some ancestor yet — sync is behind — and the fold is
+    /// over a truncated history.
+    ancestry_complete: bool,
+    /// Every op in the cut's ancestry is also readable — none is an opaque
+    /// placeholder for a group op this node holds no key for. Implies
+    /// `ancestry_complete`; kept separate because the remedies differ (a fetch
+    /// versus a key).
+    ancestry_decoded: bool,
     /// The live plane holds a DIRECT row for this member. Inherited and
     /// open-join members have none, and the projection modelling them as direct
     /// is an expected difference between the planes, not a disagreement.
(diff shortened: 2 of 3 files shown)


chefsale:
## Result: the gate now concludes 8× more, and every new conclusion agrees

Green — 151 checks, no scenario failures. Against master's baseline:

| | master | this PR |
| --- | --- | --- |
| attempts | 67 | 393 |
| **conclusive** | **15** | **120** |
| `agree` | 15 | 120 |
| `diverged` | 0 | 0 |
| `skipped_stale_cut` | 0 | 7 |
| `skipped_undecoded` | 52 | 266 |

By op kind, which is the part that matters:

```
by op kind:            conclusive by op kind:
  42 op_kind="add"       15 op_kind="add"
 326 op_kind="join"     105 op_kind="join"
  25 op_kind="remove"     — none
```

**Joins agree, all 105 of them.** The projection's fold of an invitation join
matches the live row every time, which is what the shared `from_invited_role`
predicted and now nobody has to take on faith. The gate checks the plane it always
claimed to.

Two things this measurement exposes, neither of them caused by this PR, both now
answerable because the numbers exist:

**Removals conclude zero of 25 attempts.** Every one abstains before it gets as
far as reading a live row — `no_live_row` is 0 suite-wide, so they are not
"member already gone", they are refused earlier. Removal is the operation with
the most security weight on this plane and the gate has never checked one.

**`skipped_stale_cut` is 7, where master had 0.** Those are comparisons whose cut
does not reach live's frontier — the exact false-positive class #3587 suppressed.
On the join plane they are real and non-zero, so before #3587 this PR would have
produced 7 spurious divergence markers and been unmergeable for reasons that were
not bugs.

The 266 undecoded refusals are worth splitting next: `cut_ancestry_decoded`
answers "present AND readable" as one boolean, and `CutAncestry` already
distinguishes `first_missing` (sync is behind — fixable) from `first_opaque`
(no key for that group — inherent). Reporting them as one label makes 266
abstentions unattributable, which is the same mistake this PR is fixing one layer
up.

chefsale:
## The 267 refusals are all opaque ops, not gaps — `skipped_incomplete` is zero

Green again (0 failures) with the labels split:

```
suite-wide membership comparisons:
    121 result="agree"
      6 result="skipped_stale_cut"
    267 result="skipped_undecoded"
```

No `skipped_incomplete` anywhere in the suite. Two conclusions, and the second is
the interesting one:

**The feed is complete.** Not one comparison in 394 hit a missing ancestor. Every
op the cut cites is folded — which is a direct check on #3587's originator feed
and #3593's drained-cascade fold: if either were leaking, gaps would show up here
as `skipped_incomplete`, and they do not.

**So the abstentions are inherent, not a lag — and that reframes the fix.** Every
one of the 267 is an ancestor that IS folded and cannot be read: a group op
encrypted under a key this node does not hold. A node outside a subgroup can never
decrypt that subgroup's ops, so it stays permanently unable to conclude. No amount
of syncing changes that.

What is over-strict is the SCOPE of the requirement. `cut_ancestry_decoded` asks
whether the whole namespace ancestry is readable, while decryptability is
per-group — and an op in subgroup H cannot change who is a member of subgroup G.
Refusing to answer about G because H is opaque throws away a conclusion the node
could soundly reach. That is most of the 267.

Making it per-group needs one thing the fold currently discards: an opaque op
folds to a bare `Noop`, losing the cleartext `group_id` its envelope carried, so
the walk cannot tell which group an unreadable ancestor belongs to. Retain that
and the requirement can narrow from "the whole namespace is readable" to "every
unreadable ancestor belongs to a group that cannot affect this membership".
Separate PR — it changes the fold's node shape.

(±1 differences against the previous comment's numbers — 121 vs 120 agree, 6 vs 7
stale-cut — are run-to-run jitter in a distributed suite, not an effect of this
commit.)

chefsale:
Where the 267 goes: filed as #3615. The cause is not per-group opacity and not sync lag — `OpPayload::Noop` means both "readable but models nothing" (a `KeyDelivery`, a metadata set, a context registration) and "could not decrypt", and `cut_ancestry` marks either as opaque. Probed directly: a three-op chain whose middle op is a readable `Noop` reports `complete=true, decoded=false`. Since every scenario publishes a KeyDelivery and registers a context, nearly every later cut in a namespace abstains — on every node, including one holding every key.

The fix needs the distinction to survive the persisted op-store (which stores the folded `Op`, so a stored `Noop` is ambiguous), which rules out the cheap side-table version and points at an appended `OpPayload::Opaque { group }`. That touches a Borsh-pinned enum and authz's exhaustive match, so it is its own change rather than a third commit here. This PR stands as the two commits it is.

Nothing in this PR depends on that fix; it only explains what the histogram is showing.

PR #3616, opened 2026-08-21
fix(projection): tell "could not read it" apart from "models nothing"

### LINKED ISSUE #3615: Noop conflates "readable but models nothing" with "cannot decrypt", so 68% of projection comparisons abstain
## Summary

`OpPayload::Noop` means two unrelated things, and the projection's "is this cut
decoded" check cannot tell them apart:

1. **readable, models nothing** — a `KeyDelivery`, a `GroupMetadataSet`, a
   `ContextRegistered`. The node read the op fine; the projection simply models
   nothing about it. `payload_from_root_op` / `payload_from_group_op` return
   `None` and the caller folds `Noop`.
2. **unreadable** — an encrypted group op whose key this node does not hold.

Only (2) makes a fold provisional. `ScopeState::cut_ancestry` marks *any* `Noop`
as `opaque`, so `is_decoded()` is false whenever the ancestry contains a
perfectly readable op of kind (1) — which is nearly every namespace, on every
node, including one holding every key.

## Evidence

Probe against the real predicate (three-op chain, middle op a readable `Noop`):

```
cut_ancestry_complete → true      (no gaps)
cut_ancestry_decoded  → false     ← a readable op that models nothing
```

And in production, from the coverage histogram #3604 added, measured over a full
88-scenario e2e suite (#3613):

```
121 result="agree"
  6 result="skipped_stale_cut"
267 result="skipped_undecoded"     ← 68% of all comparison attempts
  0 result="skipped_incomplete"    ← so it is NOT sync lag: no gaps anywhere
```

Every scenario publishes a `KeyDelivery` on the first add and registers a
context, so after those land, every later membership comparison in that namespace
abstains — the divergence gate concludes 121 of 394 attempts, and the remaining
267 are mostly this, not missing keys.

## Why this matters beyond the gate

The same predicate gates the projection's own at-cut answers. A cut whose
ancestry contains any unmodelled-but-readable op is treated as provisional, so
anything requiring a decoded ancestry defers to live indefinitely — for a
projection the cutover intends to make authoritative, that is the property that
has to hold, and today it almost never does.

## The fix, and why the obvious cheap version does not work

The distinction has to be recoverable wherever ops are read, and there are two
sources: the governance-DAG walk (`collect_namespace_ops`, which re-decrypts) and
the persisted unified op-store (`load_scope_ops`, which returns the folded `Op`
as stored). A side table of "ids this node could not read" covers the first and
not the second: the op-store persists `Noop` for both cases, so the information
is not recoverable there. Marking every stored `Noop` conservatively unreadable
puts us back where we started, since the op-store is the preferred source.

So it belongs in the payload:

```rust
// appended — the enum is Borsh-pinned and append-only by contract
Opaque { group: ContextGroupId },
```

* `op_from_namespace_op_with_binding`: an encrypted `NamespaceOp::Group` that
  failed to decrypt folds to `Opaque { group }` instead of `Noop`.
* `ScopeState::cut_ancestry`: `opaque` is set by `Opaque`, not by `Noop`.
* folding `Opaque` stays a no-op, so no state or hash changes for readable ops.
* `calimero-authz`'s exhaustive match gains an arm — an op this node cannot read
  authorizes nothing, which is a rule worth stating explicitly rather than
  inheriting from a catch-all.
* the group id it carries is what a later narrowing needs anyway: an unreadable op
  in subgroup H cannot change membership in G, so the requirement can eventually
  become "no unreadable ancestor in a group that could affect this membership"
  rather than "none anywhere".

Note this does not introduce node-dependent payloads — it disambiguates an
existing one. A node that cannot decrypt already folds `Noop` where a node that
can folds `MemberAdded`, so readability already shows up in the fold; today it
does so ambiguously.

**Alternative considered:** decide readability at query time by re-checking key
availability per `Noop` ancestor. More accurate (a key that arrived after the
fold counts immediately, with no re-ingest) but it puts store I/O inside a read
that currently runs under the projection lock, which the surrounding code is
explicit about avoiding.

## Expected result

The `skipped_undecoded` count should fall sharply and the conclusive count rise
correspondingly — the histogram in `projection-coverage` measures this directly,
so the change verifies itself against the 121/267 baseline recorded in #3613.

chefsale:
Fixes #3615.

`OpPayload::Noop` meant two unrelated things:

1. **read fine, models nothing** — a `KeyDelivery`, a `GroupMetadataSet`, a
   `ContextRegistered`. `payload_from_*_op` returns `None` and the caller folds
   `Noop`.
2. **could not decrypt** — an encrypted group op whose key this node lacks.

Only (2) makes a fold provisional, but `cut_ancestry` marked either as opaque. So
every cut behind the first unmodelled op in a namespace was treated as unreadable
— on every node, including one holding every key. Every scenario publishes a
`KeyDelivery` on its first add, so that is nearly every cut, and the coverage
histogram from #3604 measured it: **267 of 394 comparisons abstained** for want of
nothing at all, with `skipped_incomplete` at zero (so it was never sync lag
either).

## The change

`Opaque { group }`, appended, for the case that actually warrants abstention.
`cut_ancestry` marks only that.

* **No state or hash change.** Folding `Opaque` is a no-op, and `governance_hash`
  is over folded state — an op that writes nothing contributes nothing either way.
* **No id change.** A governance op's id is its signed content hash, supplied at
  construction, not derived from the payload. That is also what keeps the
  late-decrypt upgrade replacing the right log entry when the key arrives.
* **Appended, never inserted.** The enum is Borsh-pinned because a data-plane op's
  id *is* a hash over its payload; `Opaque` takes tag 18 and every existing
  variant keeps its byte. The pinned-tag test enforces it.
* **An authz arm, deliberately.** The enum is exhaustively matched there on
  purpose so a new variant cannot be swept into a catch-all. A hole authorizes
  nothing — and inventing a *refusal* would be wrong in the other direction: the
  op may well be authorized, this node simply cannot see what it says.
* **The group id comes free** from the envelope, which states it in cleartext even
  when the body cannot be read. Nothing consumes it yet; it is what a later
  narrowing needs, since an unreadable op in a sibling subgroup cannot change who
  is a member of this one.

## Why not a side table

Recorded because it is the obvious cheaper design and it does not work.
Readability would have to be recoverable from **both** sources of ops: the
governance-DAG walk (`collect_namespace_ops`, which re-decrypts) and the persisted
unified op-store (`load_scope_ops`). The op-store keeps the folded `Op`, so a
stored `Noop` is ambiguous there — and the op-store is the *preferred* source. A
side table covers the walk and silently misses the common path.

## Tests

* `a_readable_op_the_projection_models_nothing_about_leaves_the_cut_readable` —
  the production path: a real `KeyDelivery` folds `Noop` and leaves the cut
  readable; the same cut behind a genuinely undecryptable group op does not.
  Negative control: restore the old `Noop`-means-opaque behaviour and this fails
  on exactly that assertion.
* `cut_ancestry_decoded_rejects_an_unreadable_op_but_not_an_unmodelled_one` —
  both directions at the predicate.
* `undecryptable_group_op_folds_as_an_opaque_graph_node` — the producer, and that
  the placeholder still occupies its DAG node with its real parents.
* The ancestry oracle now generates **three** payload kinds instead of two
  (modelled / `Noop` / `Opaque`) across every small chain and drop mask, and its
  transcribed reference tracks the corrected meaning — so the exhaustive walk
  covers the distinction rather than assuming it.
* `op_payload_discriminants_are_pinned` gains `Opaque` at 18.

Workspace `cargo check --all-targets`, clippy and fmt clean; op / projection /
authz / op-adapter / governance-store / context suites pass.

## Expected effect

`skipped_undecoded` should collapse and the conclusive count rise
correspondingly. #3613's `projection-coverage` job prints the histogram, so this
verifies itself against the recorded 121 conclusive / 267 undecoded baseline —
whichever of the two lands second gets the comparison for free.

---

> [!NOTE]
> **Medium Risk**
> Touches payload encoding, ancestry-decoded checks, and op-store load reclassification that feed at-cut completeness and divergence abstention. Folding still writes no state and Borsh tags are append-only, but a misclassified hole vs Noop would change whether a cut is treated as conclusive.
> 
> **Overview**
> Stops treating every fold-to-nothing op as unreadable. `Noop` now means the op was read and the projection models nothing about it; a new appended `OpPayload::Opaque { group }` is the hole for an encrypted group op this node cannot decrypt.
> 
> `cut_ancestry` marks only `Opaque` as undecoded, so a cut behind a `KeyDelivery` (or similar) is conclusive even on nodes that hold every key. Folding `Opaque` is still a no-op; authz allows it without inventing a refusal. Late-decrypt log upgrades track `opaque_log_pos` instead of `Noop`.
> 
> On op-store load, stored `Noop`s from older builds are re-derived from the signed op so a never-decryptable row cannot pass as readable. Borsh discriminant 18 is pinned; existing tags are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de7ae673f181fa2953dde031228d9037d54d642a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/authz/src/authorize.rs
+++ b/crates/authz/src/authorize.rs
@@ -204,6 +204,13 @@ pub fn authorize(op: &Op, acl_at_cut: &AclView) -> Result<(), Rejected> {
         }
         // A graph-only node mutates nothing, so there is nothing to authorize.
         OpPayload::Noop => Ok(()),
+        // Neither does a hole. An op this node could not decrypt writes no state
+        // here, so there is no rule to apply — and inventing a refusal would be
+        // wrong in the other direction: the op may well be authorized, this node
+        // simply cannot see what it says. Authorization of its real payload
+        // happens on the nodes that can read it, and on this one once a key
+        // arrives and the op re-folds.
+        OpPayload::Opaque { .. } => Ok(()),
 
         // ---- account plane ----
         OpPayload::DeviceLinked {
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -321,12 +321,16 @@ pub struct ScopeProjections {
     /// Op ids already retained per scope — gives `ingest_op` O(1) dedup of a
     /// replayed delta instead of an O(n) scan of the log.
     seen: HashMap<ScopeId, HashSet<[u8; 32]>>,
-    /// Index in `logs` of each op id currently folded as `Noop`, per scope. The
-    /// late-decrypt upgrade (a non-`Noop` payload re-ingested for a `Noop`-folded id)
-    /// uses it to replace the entry in O(1) instead of scanning the whole log, and the
-    /// common no-upgrade re-ingest skips the scan entirely (id not present here).
-    /// Entries are removed once upgraded, so it only ever holds still-undecrypted ops.
-    noop_log_pos: HashMap<ScopeId, HashMap<[u8; 32], usize>>,
+    /// Index in `logs` of each op id currently folded as `Opaque`, per scope — an
+    /// op this node could not decrypt. The late-decrypt upgrade (a readable payload
+    /// re-ingested for an id folded as a hole) uses it to replace the entry in O(1)
+    /// instead of scanning the whole log, and the common no-upgrade re-ingest skips
+    /// the scan entirely (id not present here). Entries are removed once upgraded, so
+    /// it only ever holds ops still waiting for a key.
+    ///
+    /// `Noop` is deliberately NOT tracked: an op the projection models nothing about
+    /// was read correctly and has nothing to upgrade to.
+    opaque_log_pos: HashMap<ScopeId, HashMap<[u8; 32], usize>>,
     /// Namespaces already replayed from persisted state, so `backfill_namespace`
     /// walks each governance DAG at most once (the live feed maintains it after).
     backfilled: HashSet<[u8; 32]>,
@@ -361,17 +365,17 @@ impl ScopeProjections {
         // O(1) dedup: `insert` is true only for a not-yet-seen id.
         if self.seen.entry(op.scope).or_default().insert(op.id()) {
             let log = self.logs.entry(op.scope).or_default();
-            if matches!(op.payload, OpPayload::Noop) {
+            if matches!(op.payload, OpPayload::Opaque { .. }) {
                 // Remember where this still-undecrypted op sits so a later decrypted
                 // re-ingest can upgrade it in O(1) (see below).
                 let _ = self
-                    .noop_log_pos
+                    .opaque_log_pos
                     .entry(op.scope)
                     .or_default()
                     .insert(op.id(), log.len());
             }
             log.push(op.clone());
-        } else if !matches!(op.payload, OpPayload::Noop) {
+        } else if !matches!(op.payload, OpPayload::Opaque { .. }) {
             // Already seen, but `op.id()` is the SIGNED op's content hash, not the decoded
             // payload's — so an encrypted op first folded as `Noop` (applied before its
             // group key arrived) shares an id with its later, decrypted form. The op-log
@@ -383,7 +387,7 @@ impl ScopeProjections {
             // common re-ingest (already a real payload) is an O(1) miss, and an actual
             // upgrade is an O(1) indexed replace — no log scan either way.
             if let Some(idx) = self
-                .noop_log_pos
+                .opaque_log_pos
                 .get_mut(&op.scope)
                 .and_then(|m| m.remove(&op.id()))
             {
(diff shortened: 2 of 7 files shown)


chefsale:
## Bugbot's finding is correct and now fixed (de7ae673)

`ops_for_namespace` prefers the op-store, which persists the **folded** op — so a row written by an older build holds `Noop` for a group op that build could not decrypt. Once `cut_ancestry` marks only `Opaque`, that row reads as fully readable: a provisional fold presented as conclusive, which is precisely how a divergence marker gets raised for an op nobody could see. And `repersist_namespace_ops` does not cover it — it rewrites rows when a key ARRIVES, so an op this node will never decrypt keeps its stale `Noop` forever.

The feed now re-derives every stored `Noop` from the signed op behind it: consult the op-log, re-attempt the decrypt, take whatever payload that yields — a real one if the key is here now, `Noop` if the op is readable and models nothing, `Opaque` if it cannot be read. Rows already holding a real payload or an `Opaque` are untouched, so steady state costs nothing and the work shrinks as rows get rewritten.

`a_legacy_noop_row_for_an_unreadable_op_is_re_derived_as_a_hole` pins it: a legacy `Noop` row for an op encrypted under a key the node never holds comes back `Opaque`, and the cut through it does not read as decoded. Negative control: disable the re-derivation and it fails on exactly that assertion.

Worth recording why the other three readers of `load_scope_ops` needed nothing — I checked each rather than assuming:

| reader | why it is unaffected |
| --- | --- |
| `scope_root_from_op_store` | folds the ops to a root, and `Noop` / `Opaque` both fold to nothing — identical root either way |
| `check_op_store_completeness` | reads op IDs only |
| `group_key_pull` id check | reads op IDs only |

Also worth noting for the record: this made me double-check the id claim in the PR body. `Op::new` **computes** the id from content including the payload — but governance bridge ops use `Op::from_parts`, which takes the delta id verbatim. So the placeholder and its later decrypted form do share an id, which is what makes the in-place upgrade land on the right entry. The test now uses `from_parts` for that reason, with a comment saying so.

PR #3621, opened 2026-08-22
fix(context): conclude a removal from the absence it was supposed to produce

chefsale:
The gap the coverage instrument surfaced last: **25 removal comparisons attempted
in a full e2e suite, 0 concluded** — on the operation with the most security
weight of the three.

## Why they never concluded

A removal DELETES the live row. So `role_of` returns `None`, which is the
*correct* answer — and the compare required `has_live_row` before it would
conclude anything, so every removal fell out as `no_live_row`: "nothing to
compare". #3616 made this visible by getting removals as far as the live lookup
at all (24 of 25 now land there); before that they abstained earlier as
`skipped_undecoded` and the shape was invisible.

But both planes saying "not a member" **is** an agreement. It is written as blank
space instead of a role, and `agrees` already expresses it — for a removal it
reads `projected == None`.

## The change

What makes a comparison possible becomes:

```rust
let comparable = has_live_row || absence_is_the_answer;
```

The verdict machinery is untouched. A projection that kept a member live removed,
or dropped one live still holds, now **diverges** instead of passing as
nothing-to-compare.

**Only for removals**, and that restriction is the substance rather than caution.
For an add or a join, an absent live row is the expected model difference — an
inherited or open-join member the live plane stores no direct row for while the
projection models one — and reading agreement into that would compare nothing
against nothing. `no_live_row` keeps naming exactly those.

The op kind becomes an enum (`MembershipOpKind`), because it now decides control
flow rather than labelling a log line. `removes()` is the question, asked where
the answer means something, instead of a string compared against `"remove"`.

## Tests

`only_a_complete_premise_yields_a_conclusive_comparison` grows to 128 cases and
pins both halves:

* conclusive **iff** the whole premise holds, including the new disjunction;
* a removal with no live row concludes from `agrees` rather than reporting
  `no_live_row`.

Negative control: drop the disjunction and it fails on exactly that second
assertion.

`a_join_is_as_comparable_as_an_add` now also pins which kinds treat absence as an
answer — `Remove` does, `Add` and `Join` do not.

245 context lib tests plus every integration suite, clippy and fmt clean.

## Expected effect

`op_kind="remove"` should appear under **conclusive by op kind** for the first
time, and `no_live_row` should fall by roughly the removal count. Against
master's baseline (334 conclusive / 24 `no_live_row` / 25 removals attempted,
none conclusive), the arithmetic is visible in `projection-coverage` on this run.

If any removal comes back `diverged`, that is a real finding on the plane that
governs revocation — and I would rather learn it from a red run here than keep not
asking.

### DIFF
--- a/crates/context/src/handlers/apply_signed_namespace_op.rs
+++ b/crates/context/src/handlers/apply_signed_namespace_op.rs
@@ -209,17 +209,61 @@ fn membership_touched(
 ) -> Option<(
     calimero_context_config::types::ContextGroupId,
     calimero_account::AccountId,
-    &'static str,
+    MembershipOpKind,
 )> {
     match payload {
-        calimero_op::OpPayload::MemberAdded { group, member, .. } => Some((*group, *member, "add")),
+        calimero_op::OpPayload::MemberAdded { group, member, .. } => {
+            Some((*group, *member, MembershipOpKind::Add))
+        }
         calimero_op::OpPayload::MemberJoinedWithDevice { group, member, .. } => {
-            Some((*group, *member, "join"))
+            Some((*group, *member, MembershipOpKind::Join))
         }
         calimero_op::OpPayload::MemberRemoved { group, member } => {
-            Some((*group, *member, "remove"))
+            Some((*group, *member, MembershipOpKind::Remove))
         }
         _ => None,
+    }
+}
+
+/// Which way a folded op moved membership. An enum rather than a label because
+/// the answer decides how the comparison reads an ABSENT live row — see
+/// [`MembershipOpKind::removes`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum MembershipOpKind {
+    /// An admin-push add.
+    Add,
+    /// An invitation join.
+    Join,
+    /// A removal or a self-leave.
+    Remove,
+}
+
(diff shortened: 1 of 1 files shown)


chefsale:
## Green (151 checks), and removals are checked for the first time

```
362 result="agree"
 10 result="skipped_stale_cut"
 27 result="skipped_undecoded"

by op kind:              conclusive by op kind:
  42 op_kind="add"         41 op_kind="add"
 332 op_kind="join"       297 op_kind="join"
  25 op_kind="remove"      24 op_kind="remove"
```

Against master's baseline the arithmetic lands exactly where it should:

| | master | this PR |
| --- | --- | --- |
| conclusive | 336 | **362** |
| `no_live_row` | 25 | **0** |
| removals concluded | 0 of 25 | **24 of 25** |
| `diverged` | 0 | **0** |

`no_live_row` went to zero, which is the tell that those 25 were all removals and
nothing else — the inherited-member case the label exists for did not occur in this
suite at all, so every one of them was a removal being discarded.

**All 24 agree.** The projection drops a removed member exactly where the live
plane deletes the row, on the operation that governs revocation, and that is now
measured rather than assumed. It also means the earlier "removals are never
checked" finding was a coverage gap and not a lurking bug — worth knowing which of
the two it was.

PR #3622, opened 2026-08-22
fix(projection): ask about one group's rows, not the whole namespace's readability

chefsale:
The last of the measured abstentions. After #3616 and #3621, the histogram reports
**362 conclusive of 399**, and the remaining 27 are all `skipped_undecoded` — a
direct-row answer refused because *some* op in the cut's ancestry is unreadable,
wherever it belongs.

That is stricter than the question needs. A node outside one subgroup cannot
decrypt that subgroup's ops by construction, so it could never conclude anything
about **any** group in the namespace — including groups whose every op it reads
perfectly.

## Why narrowing is sound

An `Opaque` op is an encrypted op of exactly ONE group. Its envelope names that
group in cleartext, which is how the placeholder came to carry it (#3616). And the
only payloads that write `groups[g][member]` — `MemberAdded`,
`MemberJoinedWithDevice`, `MemberRemoved` — name `g`.

So an unreadable op belonging to a sibling subgroup **cannot** have moved this
group's direct rows. Refusing on its account discards a conclusion the node can
soundly reach.

`first_opaque_in(group)` and `cut_ancestry_state_for_group(...)` ask the narrower
question, and the membership compare uses them.

## What deliberately does NOT change

The apply-auth compare keeps the namespace-wide question. It walks the inheritance
chain — anchor membership, the `CAN_JOIN_OPEN_SUBGROUPS` bit, a visibility wall —
so it genuinely depends on other groups' ops, and narrowing it would be unsound.
Both doc comments state which question they ask and why, so the next reader does
not "simplify" them into one.

## Tests

* `an_unreadable_sibling_group_does_not_block_this_group` (projection) — same walk,
  same log: `first_opaque_in(mine)` is `None` while `first_opaque_in(sibling)` names
  the hole, and `is_decoded()` still reports it for the inheritance-walking callers.
* `a_hole_in_one_subgroup_does_not_blind_the_projection_to_another` (context) — the
  same thing through `ScopeProjections`, built from real `op_from_namespace_op`
  folds: namespace-wide `(true, false)`, asked about `mine` `(true, true)`, asked
  about `sibling` `(true, false)`.

**Coverage caveat, stated rather than glossed:** these pin the API in both
directions, but nothing unit-tests the *handler's choice* of which question to ask
— I checked, and reverting that one line leaves every unit test green. The
`projection-coverage` histogram is what catches it, which is exactly the coverage
instrument this series built, so the wiring is verified by the e2e run below rather
than by a Rust test.

246 context lib tests, projection suites, clippy and fmt clean.

## Expected effect

`skipped_undecoded` should fall from 27 toward zero, and conclusive should rise to
match. If it does not move, the remaining abstentions are same-group holes — a node
genuinely blind to the group it is being asked about, which is the one case that
*should* abstain, and the histogram will have told us that too.

### DIFF
--- a/crates/context/src/handlers/apply_signed_namespace_op.rs
+++ b/crates/context/src/handlers/apply_signed_namespace_op.rs
@@ -470,8 +470,20 @@ fn shadow_fold_and_compare(
                 // conditions with different remedies — a gap needs a fetch, an
                 // opaque op needs a key — and reporting them as one refusal is
                 // what made 52 abstentions in a suite unattributable.
-                let (complete, decoded) = membership.map_or((false, false), |_| {
-                    projections.cut_ancestry_state(&shadow_op.scope, &[shadow_op.id()])
+                // Asked about THIS group, not the whole namespace. The compared
+                // value is a direct member row, which only this group's own ops
+                // write — so an unreadable op belonging to a sibling subgroup
+                // cannot have changed it, and treating it as fatal is what left a
+                // node blind to one subgroup unable to conclude anything about any
+                // group in the namespace. The apply-auth compare below keeps the
+                // namespace-wide question, because it walks the inheritance chain
+                // and so genuinely depends on other groups' ops.
+                let (complete, decoded) = membership.map_or((false, false), |(group, _, _)| {
+                    projections.cut_ancestry_state_for_group(
+                        &shadow_op.scope,
+                        &[shadow_op.id()],
+                        group,
+                    )
                 });
                 (true, role, complete, decoded, at_frontier)
             }
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -717,6 +717,38 @@ impl ScopeProjections {
         })
     }
 
+    /// [`cut_ancestry_state`](Self::cut_ancestry_state), asked about ONE group's
+    /// direct member rows rather than the whole namespace.
+    ///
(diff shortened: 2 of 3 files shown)


chefsale:
## Green, and the narrowing freed about half — with the remainder accounted for

Against master (this branch is based on master, so it does not carry #3621 —
removals show 0 conclusive and `no_live_row` 25 here for that reason, not a
regression):

| | master | this PR | Δ |
| --- | --- | --- | --- |
| conclusive | 336 | **345** | +9 |
| `skipped_undecoded` | 26 | **14** | **−12** |
| `skipped_stale_cut` | 12 | 15 | +3 |
| attempts | 399 | 399 | — |

Every freed abstention is accounted for: of the 12, **9 became conclusive** and
**3 moved to `skipped_stale_cut`** — readable now, but their cut does not reach
live's frontier, which is a different and legitimate refusal. Nothing disappeared
into an unexplained bucket, which is the property worth checking on a change that
makes a gate less strict.

I predicted this would push `skipped_undecoded` "toward zero" and it did not: it
halved. The remaining 14 are same-group holes — a node genuinely blind to the group
it is being asked about, which is the one case that *should* abstain. So that is
the floor, and it is the sound one rather than an artefact.

Stacking, for whoever merges: this and #3621 are independent and touch different
lines of the same function, so either order works; the second will want a trivial
rebase. Together they should read ~369 conclusive with `no_live_row` at 0 and
`skipped_undecoded` at ~14.

PR #3626, opened 2026-08-22
fix(context): abstain at a cut this node cannot read, instead of deciding over it

chefsale:
**This one changes authorization behaviour, so the reasoning is here in full rather
than in a linked issue.** It came out of asking what the coverage work from #3587→
#3622 was actually protecting, and finding that the answer is not a shadow.

## The projection is already authoritative, in two places

`crates/governance-store/src/permission_checker.rs`:

```rust
if let Some(verdict) = self.authorizer.is_admin_at_cut(&self.group_id, identity, self.parents) {
    return Ok(verdict);          // Some(false) refuses — no live fallback
}
```

and the projection-backed `EphemeralProjectionAuthorizer` is what production passes
(`governance_dag.rs:115`, `group_key_pull.rs:245`), not `LiveFallbackAuthorizer`.
On the data plane, `state_delta/buffering.rs:163` admits deltas via
`member_at_cut_authoritative`.

## The gap

Both abstain when the cited ancestry is **incomplete** and neither checks whether
it is **readable**:

* `member_at_cut_authoritative` → `if !self.cut_ancestry_complete(..) { return None }`
* `auth_cut_context_or_cause` (behind every apply gate) → `if !walked.is_complete() { .. }`

An `Opaque` op folds to nothing. So a node missing a key folds a view with a hole
where that op's effect belongs, and answers from it.

| unreadable op in the cut | the fold says | consequence |
| --- | --- | --- |
| `MemberAdded{Admin}` / `MemberRoleSet{Admin}` | not an admin | **refuses** a legitimate op — that node stops applying and its governance DAG stalls |
| `MemberRemoved` of an admin | still an admin | **accepts** an op every other node refuses |

Forward-secrecy rotation makes this reachable rather than theoretical: a member
admitted at a later key epoch never holds the earlier ones, so ops in its own
group's history are unreadable to it.

And the second row is **structurally invisible** to the apply-auth shadow, by its
own admission — *"the reverse (projection accepts what live rejected) is
unobservable here — a live-rejected op never reaches this fold."* No amount of
shadow evidence would have surfaced it, which is why this is a code-reading finding
rather than a histogram one.

## The change

`auth_cut_context_or_cause` refuses on `first_opaque()`, and
`member_at_cut_authoritative` does the same from the same single walk. Abstaining
defers to live, which is the answer every node still agrees on.

* **Namespace-wide, deliberately.** Every gate reached through that context walks
  the inheritance chain — admin via an anchor, a capability bit, a visibility wall
  — so it genuinely depends on other groups' ops. The per-group narrowing from
  #3622 is sound only for a direct-row read, and both doc comments now say which
  is which.
* **Its own cause,** `AncestryUnreadable`, not folded into `AncestryGap`: a gap
  needs history, this needs a key, and the metric exists to tell refusals apart.
  Transient, since a key delivery re-folds the op.

## The behaviour change, stated plainly

`a_pulled_key_unwedges_another_members_open_subgroup_join` documented the old
behaviour. Its scenario is a visibility flip this node cannot decrypt, then an open
self-join citing it. The node used to answer **"no membership path"** — from a fold
missing the flip, contradicting any node that holds the key — and now **parks until
the key arrives**.

The rest of that test is unchanged and still passes: the pulled key re-folds the
flip and the join applies. So a divergent rejection became a retry, and the
recovery path that test exists for is intact.

## Cost

More deferrals to live, in the safe direction: live is today's ground truth and
what every node still agrees on. It slightly reduces how often the projection
answers authoritatively, which is a fair price for not answering wrongly — and the
`at_cut_undecidable{cause="ancestry_unreadable"}` counter now says how often that
happens instead of leaving it to inference.

246 context lib tests, governance-store (585) and projection suites, clippy and fmt
clean. Negative control: drop the `first_opaque()` check and the new grant-path test
fails on exactly that assertion.

### DIFF
--- a/crates/context/src/group_key_pull.rs
+++ b/crates/context/src/group_key_pull.rs
@@ -246,10 +246,19 @@ mod tests {
             )
         };
 
+        // The wedge is an ABSTENTION, not a rejection, and the difference matters.
+        // The flip sits in this cut unreadable, so the node does not know whether
+        // the subgroup is Open; a node holding the key does and would accept this
+        // join. Answering "no membership path" from a fold that is missing the
+        // flip is a verdict about a history this node cannot see, and it is the
+        // verdict the other node contradicts. Parking until the key arrives is
+        // what converges — and the rest of this test is precisely that arrival.
         let wedged = apply(&store).expect_err("the subgroup still reads Restricted");
+        let wedged = format!("{wedged:#}");
         assert!(
-            format!("{wedged:#}").contains("no membership path"),
-            "the wedge must be the membership-path rejection, got: {wedged:#}"
+            wedged.contains("has not folded the ancestry"),
+            "the wedge must be an abstention that retries, not a decision taken \
+             over an unreadable ancestor, got: {wedged}"
         );
 
         // `load_scope_ops` returns the rows in any order, so compare as sets.
--- a/crates/context/src/scope_projection.rs
+++ b/crates/context/src/scope_projection.rs
@@ -1873,7 +1873,15 @@ impl ScopeProjections {
         // op was absent from the log, so the inherited walk still saw the member
         // in the root). If the ancestry isn't whole, abstain (`None`) and defer to
         // live's reject rather than grant on a truncated, possibly-stale view.
-        if !self.cut_ancestry_complete(&scope, heads) {
+        //
+        // An UNREADABLE ancestor costs exactly the same over-grant by a different
+        // route: an op this node holds no key for folds to nothing, so a removal it
+        // cannot decrypt leaves the member folded as present — the same stale view,
+        // reached without any op going missing. This is a grant path, so it abstains
+        // for the same reason it abstains on a gap. The walk is shared, so both
+        // answers come from one pass.
+        let walked = ScopeState::cut_ancestry(self.logs.get(&scope)?, heads);
+        if !walked.is_complete() || walked.first_opaque().is_some() {
             return None;
         }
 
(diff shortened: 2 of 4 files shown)


chefsale:
## Correction to the PR body: "the answer every node still agrees on" is wrong

I wrote that abstaining defers to live, "which is the answer every node still agrees on." That is not true, and the difference matters for how this change should be read.

Live is **equally blind**. `apply_signed_op` applies an inner group op only `if let Some(group_key) = resolved_key`; with no key it buffers for retry and the effect is absent from the live rows. So a node that cannot decrypt an op is missing its effect on **both** planes, and live's verdict differs between nodes with different key epochs exactly as the projection's does.

What this PR actually buys, stated accurately:

1. **It stops the projection producing a second, differently-timed verdict.** The projection answers at the op's cut, live answers about now; both blind to the same op, they can still disagree with each other on the same node — and the gate treats the projection's as authoritative. Abstaining removes that.
2. **It makes the blindness countable** — `at_cut_undecidable{cause="ancestry_unreadable"}` — instead of folding it invisibly into a verdict.
3. **It preserves today's semantics** rather than layering a differently-blind decision on top of them.

What it does **not** buy, and cannot: agreement between nodes with different key epochs. That divergence is in live too, and it is inherent to authorizing over encrypted history.

Which leaves a design question above this PR, and I think it is the one worth answering before F4 goes further: **is authorization-over-encrypted-history expected to be node-dependent?** With forward secrecy, a member admitted at a later key epoch never holds the earlier ones, so ops in its own group's history are permanently unreadable to it — and every cut transitively cites them. Either
* an authorizing node is required to hold every epoch of the group it authorizes for (an admission-time invariant somebody has to state and enforce), or
* the live plane can never be retired for late-admitted nodes, and F4 means "the projection answers where it can" rather than "the projection is the authority".

I cannot settle that from the code. It does not block this PR — the change is strictly an improvement on either answer — but it should not stay implicit.

chefsale:
## Hold this — I got the key model wrong, and as written this PR is too blunt

Two corrections, the second of which needs a code change before merge.

**1. Historical epochs ARE recoverable.** I claimed a member admitted at a later key epoch never holds the earlier ones. Not so — the pull responder gates on CURRENT membership and then serves whatever `key_id` the requester names:

> Serve the EXACT key the requester named when it named one (a buffered op stranded on a specific — possibly rotated-out — epoch)

And unreadable ops are buffered, buffering drives the pull, and `repersist_namespace_ops` re-folds them after. So forward secrecy here protects against *removed* members (they fail `is_member`), not against newly-added ones, and `ancestry_unreadable` really is transient for a member — it clears once the pull lands. The design question I raised above is largely already answered in the code, and answered well.

**2. Which means this PR's namespace-wide abstention is too blunt.** There is one case where the hole is permanent: a node that is not a member of some subgroup in the namespace cannot pull that subgroup's key — correctly refused. Those ops stay holes forever.

`auth_cut_context_or_cause` asks `first_opaque()` over the WHOLE namespace ancestry. So on such a node it would abstain forever, for **every group in that namespace**, including groups whose every op it reads. That is most nodes. Safe (it defers to live) but it would quietly switch the authoritative path off rather than make it correct — the opposite of what F4 needs.

**The fix is #3622's argument extended to the chain:** abstain only for a hole in a group whose ops could affect the question — the target group or its ancestors. Inheritance flows parent→child, so a sibling's or a descendant's hole cannot change admin-or-capability for the target. Subgroup edges come from root ops, which are never encrypted, so the chain is always knowable even when group ops are not.

Pushing that shortly. The soundness property is unchanged; what changes is that a node stops being penalised for holes that cannot affect the answer.

chefsale:
Closing. All three e2e failures are this change: `cause=AncestryUnreadable` → "authority undecidable for signer" → the namespace op fails to apply, and in `dm-subgroup-privacy` that breaks a legitimate `join_context` (node-3 is a member there, no `expected_failure`). Same cause in `shared-storage-account-writers-two-devices` and `account-device-revoke-lockout`.

The premise in the PR body — that abstaining is the safe direction — does not hold on these paths: `None` from the authorizer becomes an apply error, not a deferral to live. So as written the change trades an unsound decision for a stall, which is worse.

The FINDING stands and I am not dropping it: `auth_cut_context_or_cause` and `member_at_cut_authoritative` decide over an ancestry containing ops the node cannot read, and two nodes with different keys can reach different verdicts. What needs work is the remedy. Filing that with the full evidence — the code paths, these three failures, and the key-pull recovery mechanism that makes the window transient for members — and reopening only with something that has been shown not to stall a legitimate join.

PR #3627, opened 2026-08-22
fix(context): abstain at a cut whose relevant history this node cannot read

chefsale:
Replaces #3626, which I closed after e2e rejected it. Same finding, correct scope,
and the scoping is the whole design — so the story of how it was learned is below,
because it is the reviewable part.

## The finding

The at-cut authority paths abstain on an **incomplete** ancestry and never on an
**unreadable** one:

* `member_at_cut_authoritative` → `if !cut_ancestry_complete(..) { return None }`
* `auth_cut_context_or_cause` (behind every apply gate) → `if !walked.is_complete() { .. }`

An `Opaque` op folds to nothing, so a node missing a key folds a view with a hole
where that op's effect belongs — and answers from it. Authoritatively:
`PermissionChecker` hands the projection's `Some(false)` straight back as a refusal
with no live fallback.

| unreadable op in the cut | the fold says | consequence |
| --- | --- | --- |
| `MemberAdded{Admin}` / `MemberRoleSet{Admin}` | not an admin | refuses a legitimate op |
| `MemberRemoved` of an admin | still an admin | **accepts one every other node refuses** |

The second is unobservable to the apply-auth shadow by its own admission — *"a
live-rejected op never reaches this fold"* — so no amount of shadow evidence would
have surfaced it. This came from reading the authoritative paths.

## The fix, and why it is scoped rather than namespace-wide

Both paths now refuse when an unreadable ancestor belongs to a group whose ops can
**move the question** — the target group or an ancestor of it. Every input to an
admin-or-capability answer at `group` is an op naming `group` or an ancestor:
roles and caps name their group, a visibility wall names its subgroup, the root
admin arrives on readable root ops. A sibling subgroup's hole cannot change any of
them.

**#3626 asked the namespace-wide question and e2e refused it inside the hour**, in
three scenarios. `dm-subgroup-privacy` is the shape: node-3 is legitimately not a
member of the other DM subgroup, can therefore *never* obtain that key, and so
parked **every** authority question in the namespace forever — including questions
about its own group, whose every op it reads. Its legitimate `join_context` failed
with `cause=AncestryUnreadable`. A refusal that can never clear is not the
conservative choice; it is a stall.

I had reasoned my way to that risk and then talked myself out of it, on the
assumption that the hole was in the target group. It was not. The logs settled it:
the abstaining group was node-3's own (`4bada981…`, whose key it had received via
direct delivery), so the hole was the sibling's.

## One subtlety worth reviewing closely

`AclView::group_and_ancestors` climbs **through** a `restricted` edge, unlike
`open_ancestors`, which stops there because inheritance stops there. This walk
cannot: whether an edge is restricted is itself decided by a
`SubgroupVisibilitySet` op **of that subgroup**, which may be precisely the op that
cannot be read. Stopping at an edge whose state is unknown would shorten the set
and let a relevant hole pass as irrelevant. The tree's *shape*, unlike its
visibility, comes from root ops and is always readable.

## Tests

`the_grant_path_defers_when_an_ancestor_is_unreadable` pins the scoping **from both
sides**, which is what #3626 lacked:

* an unreadable op **of the target group** → abstain (`None`);
* an unreadable op of a **sibling** → still answers `Some(true)`.

Negative controls both ways: remove the check and the first assertion fails; widen
it back to namespace-wide and the second fails. The second is the one that would
have caught #3626 before CI did.

`a_pulled_key_unwedges_another_members_open_subgroup_join` keeps its updated
assertion from #3626: with a hole in the target group the node parks instead of
answering "no membership path" over a fold missing the visibility flip, and the
rest of that test is unchanged — the pulled key re-folds the flip and the join
applies.

## Cause label

`AncestryUnreadable`, its own cause rather than folded into `AncestryGap`: a gap
needs history, this needs a key, and the metric exists to tell refusals apart.
Transient for a member — the pull responder serves any epoch by `key_id` to a
current member, and `repersist_namespace_ops` re-folds the op after.

246 context lib tests, 585 governance-store, projection and authz suites, clippy
and fmt clean. e2e is the check that matters and it is running here.

### DIFF
--- a/crates/authz/src/inheritance.rs
+++ b/crates/authz/src/inheritance.rs
@@ -98,6 +98,35 @@ impl AclView {
         .take(MAX_NAMESPACE_DEPTH + 1)
     }
 
+    /// `group` and every ancestor above it, nearest first — the set of groups
+    /// whose ops can move an authority question asked about `group`.
+    ///
+    /// Deliberately climbs THROUGH a `restricted` edge, unlike
+    /// [`Self::open_ancestors`]. That walk stops there because inheritance stops
+    /// there; this one cannot, because whether an edge is restricted is itself
+    /// decided by a `SubgroupVisibilitySet` op of that subgroup — which may be
+    /// exactly the op a reader cannot decrypt. Stopping at an edge whose state is
+    /// unknown would silently shorten the set and let an unreadable ancestor pass
+    /// as irrelevant.
+    ///
+    /// Same `MAX_NAMESPACE_DEPTH` bound as the inheritance climb, so a cyclic or
+    /// adversarial tree cannot spin.
+    pub fn group_and_ancestors(
+        &self,
+        group: ContextGroupId,
+    ) -> impl Iterator<Item = ContextGroupId> + '_ {
+        let mut current = Some(group);
+        core::iter::once(group).chain(
+            core::iter::from_fn(move || {
+                let edge = self.subgroups.get(&ScopeId::from(current?.to_bytes()))?;
+                let parent = ContextGroupId::from(*edge.parent.as_bytes());
+                current = Some(parent);
+                Some(parent)
+            })
+            .take(MAX_NAMESPACE_DEPTH + 1),
+        )
+    }
+
     /// Does `author` hold a direct membership row in `group`?
     fn has_direct_row(&self, group: ContextGroupId, author: &AccountId) -> bool {
         self.groups
--- a/crates/context/src/group_key_pull.rs
+++ b/crates/context/src/group_key_pull.rs
@@ -246,10 +246,19 @@ mod tests {
             )
         };
 
+        // The wedge is an ABSTENTION, not a rejection, and the difference matters.
+        // The flip sits in this cut unreadable, so the node does not know whether
+        // the subgroup is Open; a node holding the key does and would accept this
+        // join. Answering "no membership path" from a fold that is missing the
+        // flip is a verdict about a history this node cannot see, and it is the
+        // verdict the other node contradicts. Parking until the key arrives is
+        // what converges — and the rest of this test is precisely that arrival.
         let wedged = apply(&store).expect_err("the subgroup still reads Restricted");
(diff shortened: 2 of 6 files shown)


chefsale:
## The two remaining failures were mine, from #3616 — fixed in `0a5883c3`

`dm-subgroup-privacy` passes now (the scoping did its job). The other two —
`account-device-revoke-lockout` and `shared-storage-account-writers-two-devices`,
both turning on a second device folding its account's history — failed with a
different symptom worth reading carefully:

```
failed to retry ... error=authority undecidable for signer
```

**The retry itself was abstaining.** Chain: `EphemeralProjectionAuthorizer` →
`ephemeral_fold` → `ops_for_namespace` → the op-store → `reclassify_stored_holes`,
which #3616 wrote to re-derive stored `Noop` rows and **skip stored `Opaque`
ones** as already-decided.

That is backwards. `Opaque` means *"I could not read this when I folded it"* — a
cache of a past failure — and the entire design has the key arriving afterwards:
a member pulls the epoch by `key_id`, the op becomes readable. Treating the row as
settled meant the retry that fires the moment the key lands asked the fold, was
told "unreadable", and abstained — so the op the key was pulled **for** could never
apply.

Harmless before this PR: a stale `Opaque` only cost the shadow a comparison it
would have skipped anyway. It could only surface once something started gating on
readability, which is this PR. Both placeholder kinds are re-attempted now.

Also worth recording, because it nearly sent me chasing a ghost: mid-investigation
`cargo check` reported `group_and_ancestors` and `AncestryUnreadable` as
non-existent while both were plainly in the source. Stale artifacts from the shared
`CARGO_TARGET_DIR` being reused across git worktrees — `touch` on the sources
cleared it.

Rebased onto current master. One correction on process: an earlier force-push of
mine discarded a `Merge branch 'master'` commit someone had added to this branch. Its
parents were my commit and the same master I was already based on, so nothing but
the empty merge was lost — but the mistake was fetching immediately before
`--force-with-lease`, which defeats the lease. Using an explicit expected SHA from
here.

Issue #2418, opened 2026-05-20
HashComparison: parent created_at divergence on same-value leaves stays divergent (residual #2407)


chefsale:
## Summary

Residual sub-case from #2407 (now closed). HashComparison can't converge two peers when they hold the **same entity at the same value bytes but with different `created_at` timestamps**, even after the bidirectional-leaf-push fix in #2414.

The most-observed case in CI is the `Root<T>` opaque-leaf entity (`Id::new([118; 32])`) in `kv-store-with-handlers` workloads — each peer writes its own copy when it first mutates the context state, so `created_at` differs per peer, but the value bytes are identical.

## Mechanism (diagnosed during #2414 investigation)

- `Metadata` is `#[borsh(skip)]` on the storage `Element`, so the leaf's merkle hash does NOT depend on HLC or `created_at`. Two peers holding the same value at different timestamps therefore compute the **same leaf merkle hash**.
- `compare_tree_nodes` returns `TreeCompareResult::Equal` for that leaf → DFS skips it → `nodes_skipped++`, no recursion.
- #2414's bidirectional leaf push has entry condition `local.hash != remote.hash`, which is not met for this case — so no push happens for the divergent entity.
- But the **parent**'s `full_hash` factors children by `(created_at, id)` ordering (see `ChildInfo::Ord`). Different per-peer `created_at` → different child ordering → different parent hash → different root hash.

Confirmed in CI logs from #2414 ([failed Sync regression run](https://github.com/calimero-network/core/actions/runs/26154576465/job/76930521599)): `nodes_compared=5 entities_merged=2 entities_pushed=0 nodes_skipped=3` repeats every interval-sync tick, `Flushed bidirectional leaf reconciliation pushes` debug line **never fires** for the stuck Round 2 loop.

## Why this is hard

Each candidate fix has ripple effects on existing invariants:

1. **Include `created_at`/HLC in the leaf merkle hash.** Divergence becomes visible to `compare_tree_nodes`, #2414's Part-2 push fires.
   - Cost: changes the `#[borsh(skip)]` on `Metadata` (or adds a separate non-skipped field), affecting every sync path that already trusts the existing merkle invariants. Snapshot, delta replay, and the on-disk index format are all touched.

2. **HashComparison unconditionally re-pushes leaves at the same entity_id in `common_children`** regardless of merkle-hash equality.
[...]

PR #3633, opened 2026-08-24
perf(storage): bound a parent's per-child write cost with a hash trie

### LINKED ISSUE #3602: Writes stop past a few thousand ROOT children: every nested collection registers as a child of ROOT
## Summary

**This issue previously blamed the wrong parent. Corrected below.**

Writes stop working once ROOT accumulates a few thousand children. The trigger is not
a large collection — it is an entity type that **contains nested collections**, because
every nested collection registers itself as a child of ROOT.

`Collection::new` (`crates/storage/src/collections.rs:340`):

```rust
if id.is_root() {
    let _ignored = <Interface<S>>::save(&mut this).expect("save");
} else {
    let _ = <Interface<S>>::add_child_to(*ROOT_ID, &mut this).expect("add child");
}
```

Re-keying does not relieve it: `reassign_deterministic_id_under` ends by re-adding the
collection to ROOT (`collections.rs:542`).

`Index::add_child_to` (`index.rs:429`) is a read-modify-write of the parent's index
record, which holds the entire children list as one blob, plus the merkle-hash cascade.
So a record carrying K nested collections costs K × O(ROOT children) per write, while
itself growing ROOT by K. That is quadratic with a K² constant.

A chat `Message` carried four (`UnorderedSet` mentions, three `Vector`s). At ~400
messages ROOT held a few thousand children and `send_message` exhausted the gas limit:
`execution exhausted its gas limit of 1000000000 points`.

## Evidence

Full factorial on the same contract, same rig, one node. "Container" is where messages
live; "fields" is whether `Message` carries nested collections or plain `Vec`s.

| container        | message fields      | result                  |
|------------------|---------------------|-------------------------|
| `AuthoredVector` | nested collections  | **dies at 394**         |
| `Journal`*       | nested collections  | **dies at 399**         |
| `AuthoredVector` | plain `Vec`s        | 900 sent OK, ~120 ms    |
| `Journal`*       | plain `Vec`s        | 1200 sent OK, ~100 ms   |

\* `Journal` is an experimental append-only collection whose entries are not listed in
the parent's children list, so its container cost is O(1) by construction.

The container factor moves the failure point by 5 (394 → 399). The nested-collection
factor decides it entirely. **Replacing the container fixes nothing; removing the
per-record nested collections fixes it with the original container.**

Latency before failure climbed 138 → 227 ms (n=100 → 375). After removing the nested
collections it is flat: ~100 ms at n=100 and at n=1200.

## What this means for the original diagnosis

The earlier text claimed the failing operations "share exactly one property: they write
into the large collection", and that `send_message`/`edit_message`/`delete_message` fail
while `save_draft`/`mark_as_read`/`update_reaction` succeed.

That correlation was real but the causal reading was wrong. The three failing methods are
also the three that touch `Message`, the only record type carrying nested collections. The
three succeeding ones write records with no nested collections. The discriminator is the
nested collections, not the size of the collection written into.

`add_child_to` being O(n) in the parent's child count is still accurate and still worth
fixing — it is the underlying mechanism. What was wrong is *which* parent overflows in
practice, and therefore where a fix has to land.

## Suggested direction

1. **A nested collection should be a child of its logical parent, not ROOT.** ROOT
   becomes a single flat list of every collection ever constructed anywhere in the tree,
   which is both a scaling problem and a structurally misleading Merkle shape.
2. **`add_child_to` should not be O(n).** Even with (1), any parent with many children
   pays it. The children list being one serialized blob is the root of it.
3. **Guidance for app authors, today:** a record stored many times over should not carry
   nested collections. Fields written once and never mutated (mentions, attachments)
   belong in plain `Vec`s, which serialize inline and cost ROOT nothing.

## Reproduction

Any contract with a record type carrying nested collections, written in a loop. Append
until writes fail. To confirm the parent: run the same loop with those fields changed to
plain `Vec`s — it will not fail.

## Measurement caveat, worth stating

My first round of numbers on this were invalid and I want that on the record so nobody
re-derives them. `logic/build-bundle.sh` only *packages* a wasm; `cargo mero build` is
what compiles it. Having skipped the latter, three separate "fixes" were measured against
one stale binary and produced identical failure points, which read as a meaningful result
and was not. Anyone benchmarking a contract change should verify the node is executing
the new build — a method that exists only in the new build, called once, is enough.

## Notes

- ~394 is this rig's number with these record sizes. The order of magnitude is the
  finding; the exact threshold varies with record size and how many nested collections
  each record carries.
- A flat `AuthoredVector` at 900 entries still writes fine (~120 ms, mild upward slope),
  so a single growing collection is far from the ceiling. It is O(n) and will eventually
  wall, just much later than a nested-collection record does.
- `deleted_children` on `EntityIndex` suggests tombstones accumulate in the same record,
  so a parent may keep paying for removed children. Not verified.

xilosada:
Closes #3602.

## The bug

A collection's append cost grew with the collection's size until a single write exhausted the gas limit, at which point the collection became **permanently unwritable** — the cost only goes up. On the mero-chat contract that wall was **394 messages**.

A parent's children lived inline in its `EntityIndex` as `Vec<ChildInfo>` (~84 bytes each), so linking one child read, rewrote and re-hashed *all* of them. At the ~1,600 ROOT children a chat context reached, each message sent rewrote ~135 KB four times over.

The hash chain itself was a correct recursive Merkle construction. What was missing was any index structure between a parent and its children: fan-out unbounded, so updates fold every sibling, diffs scan every sibling, and an inclusion proof needs all of them. The logarithmic properties a Merkle tree is used for require bounded fan-out, which nothing provided.

## Why 715 green tests never saw it

Gas is charged per executed wasm operator and nothing else. A storage read costs the ~4 operators of the caller regardless of value size, and there is **no read counter and no read limit in the VM**. So an O(n) read pattern never registers as "reads are expensive" — it registers as gas burnt borsh-decoding whatever those reads returned, which is invisible until a call dies.

Every storage test asserts what the code *computes*. None asserted what it *costs*. That gap is why this shipped, and why three more O(n) regressions were introduced while fixing the first — each behind a call that reads as free at the call site.

## The fix, in the order the costs were found

| # | Hidden cost | Effect |
|---|---|---|
| 1 | Inline sibling list — one link rewrites every sibling | wall **394 → 1,187** |
| 2 | `Collection::len` enumerated the whole child set, and the contract derives a message id from `len()` on every write | 1,187 → 1,179 (**nothing**) |
| 3 | `CollectionMut::insert` called `children_cache()`, reading *every* existing child to record one new id | this was the binding cost |
| 4 | A **log line** asked the trie to `enumerate()` rather than count, on the ancestor recompute of every write | n=599: **7,682 → 184** reads |

Fix 2 is worth reading as a cautionary result: making `len()` O(1) moved the wall by 8 messages. Three earlier attempts at write-side costs moved the ceiling without flattening the curve, because writes were 76 throughout — **bounded all along**. The cost was always reads.

### `ChildTrie`

A sparse hex trie keyed by child id, `DEPTH = 4` nibbles deep, with bucketed leaves. A link touches `DEPTH+1` rows whatever the parent holds:

```
n=    10:   467 bytes/link, 5 rows   (old inline blob: ~840)
n= 1,000: 1,259 bytes/link, 5 rows   (old:         ~84,000)
n=10,000: 1,820 bytes/link, 5 rows   (old:        ~840,000)
n=50,000: 2,018 bytes/link, 5 rows   (old:      ~4,200,000)
```

Rows are constant and bytes plateau once interior nodes fill their 16 slots (10k→50k is 1.109×). Nodes store only occupied slots, so a parent with three children costs three small rows — at n=10 the trie is already cheaper than the blob, so small parents are not taxed to make large ones work.

Keyed by **child id, not append order**: the root must be a function of the child *set*, or replicas that learn children in different orders never converge.

## Measured result

Against the real mero-chat contract, unmodified (`max_gas` 1e9), via `crates/runtime/tests/chat_wall.rs`:

| | n=100 | n=60,000 |
|---|---|---|
| `send_message` gas | 15.6M | 18.3M |
| storage reads | 290 | 290 |
| latency | 3.7ms | 4.5ms |

**No write wall below 60,000**, where the probe stops. 600× the data for 1.17× the gas, and the read count does not move at all.

(Reads count misses as well as hits. That matters here: a link probes `DEPTH+1` spine rows that mostly do not exist yet, so a hits-only counter would be blind to the dominant cost — see the telemetry note below.)

## Compatibility — decided: flag day, no migration

Two hard breaks, and the call has been made: **land as a flag day with a version bump. Nodes start clean; no data migration.**

**1. `EntityIndex` borsh layout changed.** `children: Option<Vec<ChildInfo>>` was field 3 and is removed; children now live in their own `Key::ChildTrie` keyspace. Index rows written by an older build do not decode under the current layout.

**2. Every parent hash changes.** A parent now commits to its children's **ids** as well as their hashes, since a child's id is its position in the trie. The old fold hashed only `merkle_hash`, so two different child sets with matching hashes were indistinguishable — strictly stronger, and it moves every hash including the root. Old and new builds compute different roots for identical data.

Nothing in production runs a child-trie build, so there is no mixed-version fleet to stage and no state that must survive. That makes the migration question moot rather than deferred — and if it ever stops being moot, the pass is already written: `rebuild_child_tries_after_snapshot` pointed at local state instead of freshly-installed state reads index rows, finds every parented one, and links it.

What still matters with no migration is that a stale row cannot be **misread**. It can't be: the `borsh_layout` mirror now covers every field, so `from_slice` rejects a row that does not match this layout exactly, rather than a prefix mirror silently accepting one and handing back a hash read out of the `children` field. Pinned by `an_old_layout_index_row_is_rejected_rather_than_misread`. A startup error is the intended outcome for a stale datastore; a plausible wrong root hash never is.

Per this repo's guidance, merobox E2E cannot verify either property — every node in a run is the same build against fresh state. Which, for a flag day, is exactly the configuration that ships.

## Also in this PR

**A snapshot fix that the trie change made necessary** (`f81e92de`). Snapshot installs an entity by writing its Entry and Index rows verbatim. While children lived *inside* the index row, that shipped the parent→child links for free. Moving them to their own keyspace broke that silently: snapshot discovery finds entities by deserialising `EntityIndex`, and a trie row never will.

It never healed. Hash comparison re-applied entities that were already byte-identical, which goes through the update path rather than `add_child_to`, so no link was ever created — a stable fixpoint, re-syncing every ~10s forever. Measured before: a joiner reported **0 of 25** messages and logged 50 "did not converge" warnings. After: **25 of 25**, matching roots, 0 warnings.

Worth recording, because it is why this shipped: the check that should have caught it cannot. `compute_root_hash` reads the *stored* `full_hash` out of the root's index row rather than recomputing it, so snapshot's "verify integrity by computing the actual root hash from storage (I7)" compared the sender's shipped value against the sender's claim. It logged "root hash verified successfully" against a node with an empty trie. Making I7 recompute would have caught this immediately and guards the whole class. (`master` has since hardened I7 to refuse to publish on mismatch; this branch rebases onto that and keeps it, placing the trie rebuild *before* the verify.)

**Read counters in the VM** (`ed13a2a8`). `storage_reads` / `storage_read_bytes` beside the existing write counters, carried out on `Outcome`. Reads are telemetry, not a budget — there is nothing to charge them against — but read *count* is the leading indicator and *bytes* the better cost predictor, since guest decode work scales with bytes.

`tests/cost_is_flat.rs` pins both on a fixture whose work is fixed by construction:

```
10 vs 10,000 stored entries : gas=1097 reads=64 read_bytes=4160 — identical
64 B vs 16 KiB values       : gas=1097 both, read_bytes 4,160 vs 1,048,640
```

The second is the one worth pinning: reading a megabyte costs the same gas as reading four kilobytes, so a redesign toward fewer, larger rows looks free in gas while costing real wall-clock and real decode. Anyone reading gas alone would draw the wrong conclusion.

## Deliberate behaviour changes

- **Enumeration keeps `ChildInfo`'s `(created_at, id)` order**, applied on the way out of the id-keyed trie. That order is load-bearing: `Vector::get(idx)` walks children in it, so returning id order would have silently reordered every `Vector` while all hashes still matched. Pinned by `sorted_insert_path_preserves_hash_semantics`.
- **`ContextRegistry::dump_root` can no longer enumerate children from the raw row** and now warns loudly rather than returning a silently empty list that would read as "no children" during divergence triage.
- **`calculate_full_hash_for_children` is gone.** Its #2418 guard — parent hash must not depend on `created_at` — is ported to the trie, where the property now holds twice over (buckets are id-sorted *and* the fold covers only `id + merkle_hash`, so no `Metadata` reaches the root). Every other trie test uses `Metadata::default()`, so folding metadata into `TrieBucket::hash` would have passed all of them; now it would not.

## Testing

| | |
|---|---|
| storage, node lib, context-client, primitives, merodb, runtime | **1555 passed, 0 failed** |
| `calimero-node --test sync_sim` | **325 passed, 0 failed** |
| `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` | clean |
| `cargo fmt --check` | clean |

New cost guards, which are the regression protection this class of bug actually needed — correctness tests cannot see it:

- `tests/write_cost_profile.rs` — reads, writes and distinct keys per insert must not grow with n
- `tests/read_cost_profile.rs` — per-insert read profile
- `tests/root_fanout_shape.rs` — parent index row flat at 10/50/200 children (16,927 B → 122 B at 200)
- `tests/trie_scaling.rs` — 1K/10K/100K/1M, `#[ignore]` (~25s release)
- `tests/cost_is_flat.rs` — gas and reads flat against store size and value size
- `tests/chat_wall.rs` — the real contract, `#[ignore]` (needs a guest from mero-chat-pwa)

## Known gaps

- **Fixed here, was [#3637](https://github.com/calimero-network/core/issues/3637):** `AuthoredVector::push` returned `len - 1`, which was correct only while `insert` appended to the child cache — the O(n) read this branch removes. Enumeration now comes from the trie in `(created_at, id)` order and `created_at` does not advance within a call, so entries pushed together tied and the random id decided position. `push` now returns the entry's `Id`, with `update_by_id` / `get_by_id` / `owner_of_id` alongside; the positional API stays for pagination. The deeper reason for id-addressing rather than a corrected index: enumeration is over the child *set*, so a remote insert renumbers everything after it — and `update` is ownership-gated by the address it is given, so a stale index writes to a different entry. **Breaking**: `push` returns `Id`, and the JS bridge writes 32 id bytes where it wrote an 8-byte index.
- **`deleted_children` has the same unbounded-row cost this fixes for `children` — [#3635](https://github.com/calimero-network/core/issues/3635).**

- **Bucket occupancy is what eventually bites.** Past `16^DEPTH` children a bucket is read, rewritten and folded whole — the inline-list problem returning in miniature. `trie_scaling` is the evidence for how far `DEPTH=4` carries.
- **Positional access is still O(n) in id reads.** `Collection::nth` materialises the child-id list from the trie on first use per call. No payloads, no nested collections, so it is far cheaper than a full scan — but it is not flat. Removing it needs an order-preserving index in the trie, which is a larger change and not obviously worth it yet.
- The app-side half of the chat wall (`get_messages` scanning the whole collection to return one page) is fixed separately in `mero-chat-pwa`, not here.

## What two review passes changed

Recorded because the findings say more about the risk surface than the diff does. Both passes are answered in the PR conversation.

Fixed since the first draft:

- **The replayed delete diverged from the local one.** Dropping an entity's trie was wired into the local delete path but not into `apply_delete_ref_action`, which is the path every *peer* takes replaying a `DeleteRef`. Same logical state, different hashes across replicas, silently. Both now go through one helper.
- **`load_rotation_log_direct` read the rotation log back empty on every delta apply.** That function exists specifically to avoid `MainStorage` — it runs before `RUNTIME_ENV` is installed — but the children walk went through it anyway, hit the process-local fallback store, missed, and had the error swallowed by `unwrap_or_default`. The writer set collapsed with nothing reporting anything.
- **Buffered snapshot entities were never linked.** `persist_buffered_snapshot_entity` runs from `drain_absorbed_leaves` long after the rebuild pass, so a late-draining entity stayed present-but-unenumerable permanently.
- **A rebuild failure was warned and then published anyway.** Verification reads the *shipped* root hash, so an unlinked node verifies, matches its peers, and never looks like it needs repair. It now fails the sync.
- **An old index row could be silently misread.** The layout mirror was a prefix, so it accepted a row written under any layout and handed back a hash read out of the `children` field. It now covers every field and rejects.
- **Read telemetry counted hits only** — blind to precisely the absent-row probes the trie's cost gates exist to watch.
- Plus: ghost children surviving a collection's deletion, the fold rule existing in three copies, `is_leaf` walking a whole subtree per level on the sync path, and two doc comments that asserted the opposite of the code.

Tests added for the parts that were structurally unguarded: an oracle pinning `insert_with` against `insert` (the two implementations of one spine walk, whose divergence *is* the snapshot bug), a regression test for the rebuild, the ghost-children case, the `created_at`-invariance guard ported from the code it outlived, and a count-vs-enumeration invariant. The ones guarding a specific defect were verified by mutation — a test that cannot fail is worse than no test, and one of these did not fail until it was fixed.

## Why `ChildTrie` is its own type

Worth stating, because a separate type can read as gratuitous when the only
consumer is `Index`. Inside the storage crate it mostly is — those call sites
could go through an `Index::child_count`-style surface. What justifies the
public API is the three callers that reach the store **directly**, with no
`StorageAdaptor` and no `RUNTIME_ENV` installed: snapshot install
(`insert_with`), the ACL shadow feed in `scope_projection`, and the raw-DB
divergence dump (`children_with`). Those paths cannot use `MainStorage` — it
routes through a thread-local that is only present inside
`context_client.execute()` — so they need the addressing and the walk without
the adaptor. That is the type's reason to exist.

## On the id-addressed API's asymmetry

`update_by_id` **errors** when the id is not an entry of this vector; `get_by_id`
and `owner_of_id` return **`None`**. That reads as inconsistent and is
deliberate. For a read, `None` is already the vocabulary for absence, and
erroring would make "not mine" indistinguishable from "storage broke". For a
write, silently doing nothing is the worse failure — a caller that asked for a
change and got no error is entitled to believe it happened.

The scoping exists because the id-addressed surface takes a caller-supplied
address and resolves it through a global `find_by_id`, while the positional API
can only ever name a child of itself. The owner check does not cover the
difference: it establishes that the caller owns the target, not that the target
is theirs to address from here.

## A note on `d483648c0`

That merge of master into this branch came from someone else working on the same
branch, not from me. It is not mine to undo. The commit after it was replayed on
top rather than rebased over it, and the full workspace was re-run on the merged
base before pushing.

## Reviewing this

Commits are ordered to be read in sequence — each measurement commit precedes the fix it motivates, so the "before" number is in the tree. The two worth the most attention are `e2aa55dc` (trie becomes the source of truth; where the hash and layout breaks live) and `f81e92de` (the snapshot fix).

https://claude.ai/code/session_01RdUocbW2V3oRQ9fE5h3v5S

### DIFF
--- a/crates/node/src/sync/level_sync.rs
+++ b/crates/node/src/sync/level_sync.rs
@@ -986,7 +986,10 @@ fn get_local_hashes_at_level(
             // Level 0: get direct children of root
             {
                 let _ = &root_index;
-                for child in Index::<MainStorage>::get_children_of(Id::root()).unwrap_or_default() {
+                // `root_id`, not `Id::root()`: the latter derives the context
+                // from the `RUNTIME_ENV` thread-local, and a divergence there
+                // yields silently empty level-0 hashes rather than an error.
+                for child in Index::<MainStorage>::get_children_of(root_id).unwrap_or_default() {
                     let child_id = *child.id().as_bytes();
                     if let Some(child_hash) = Index::<MainStorage>::get_hashes_for(child.id())
                         .ok()
--- a/crates/node/src/sync/level_sync.rs
+++ b/crates/node/src/sync/level_sync.rs
@@ -67,6 +67,7 @@ use calimero_node_primitives::sync::{
 use calimero_primitives::context::ContextId;
 use calimero_primitives::identity::PublicKey;
 use calimero_storage::address::Id;
+use calimero_storage::child_trie::ChildTrie;
 use calimero_storage::env::with_runtime_env;
 use calimero_storage::index::Index;
 use calimero_storage::interface::Interface;
@@ -971,25 +972,33 @@ fn get_local_hashes_at_level(
 
     let root_id = Id::new(*context_id.as_ref());
 
-    // Get the root index to access children
-    let root_index = match Index::<MainStorage>::get_index(root_id) {
-        Ok(Some(idx)) => idx,
+    // Existence only. The children come from the trie below, so the decoded row
+    // itself is not wanted — binding it and then reaching for
+    // `get_children_of` is what made this a doubled read.
+    match Index::<MainStorage>::get_index(root_id) {
+        Ok(Some(_)) => {}
         Ok(None) => return Ok(hashes), // Empty tree
         Err(e) => {
             warn!(%context_id, error = %e, "Failed to get root index");
             return Ok(hashes);
         }
-    };
+    }
 
     match parent_ids {
         None => {
             // Level 0: get direct children of root
             {
-                let _ = &root_index;
                 // `root_id`, not `Id::root()`: the latter derives the context
                 // from the `RUNTIME_ENV` thread-local, and a divergence there
                 // yields silently empty level-0 hashes rather than an error.
-                for child in Index::<MainStorage>::get_children_of(root_id).unwrap_or_default() {
+                //
+                // The trie directly, not `get_children_of`: that re-reads and
+                // re-decodes the index row this function just proved decodable
+                // — a doubled read per parent per level, on a protocol chosen
+                // for wide trees — and its only failure here is that re-read,
+                // so `unwrap_or_default` would turn a corrupt row into "no
+                // children" instead of surfacing it.
+                for child in ChildTrie::<MainStorage>::new(root_id).children() {
                     let child_id = *child.id().as_bytes();
                     if let Some(child_hash) = Index::<MainStorage>::get_hashes_for(child.id())
                         .ok()
@@ -1004,12 +1013,11 @@ fn get_local_hashes_at_level(
             // Deeper levels: get children of specified parents
             for parent_id in parents {
                 let parent_storage_id = Id::new(*parent_id);
-                if let Ok(Some(parent_index)) = Index::<MainStorage>::get_index(parent_storage_id) {
+                if let Ok(Some(_)) = Index::<MainStorage>::get_index(parent_storage_id) {
                     {
-                        let _ = &parent_index;
-                        for child in Index::<MainStorage>::get_children_of(parent_storage_id)
-                            .unwrap_or_default()
-                        {
+                        // Same as level 0: existence is already established, so
+                        // walk the trie rather than re-reading the row.
+                        for child in ChildTrie::<MainStorage>::new(parent_storage_id).children() {
                             let child_id = *child.id().as_bytes();
                             if let Some(child_hash) =
                                 Index::<MainStorage>::get_hashes_for(child.id())
--- a/crates/node/src/sync/level_sync.rs
+++ b/crates/node/src/sync/level_sync.rs
@@ -1098,9 +1098,12 @@ fn get_nodes_at_level(
             };
 
             let child_hash = child_index.full_hash();
-            let is_leaf = Index::<MainStorage>::get_children_of(child.id())
-                .unwrap_or_default()
-                .is_empty();
+            // `has_children`, not `get_children_of(..).is_empty()`: the latter
+            // walks every node and bucket row of that child's trie, allocates
+            // every ChildInfo and sorts them — per child, per level, on the sync
+            // hot path. For the ~1,600-child collection this work exists to fix,
+            // that is the linear read put back, just on the read side.
+            let is_leaf = !Index::<MainStorage>::has_children(child.id()).unwrap_or(false);
 
             // Determine parent_id for this node (None for level 0)
             let parent_id_bytes = if level == 0 {
--- a/crates/node/src/sync/level_sync.rs
+++ b/crates/node/src/sync/level_sync.rs
@@ -984,8 +984,9 @@ fn get_local_hashes_at_level(
     match parent_ids {
         None => {
             // Level 0: get direct children of root
-            if let Some(children) = root_index.children() {
-                for child in children {
+            {
+                let _ = &root_index;
+                for child in Index::<MainStorage>::get_children_of(Id::root()).unwrap_or_default() {
                     let child_id = *child.id().as_bytes();
                     if let Some(child_hash) = Index::<MainStorage>::get_hashes_for(child.id())
                         .ok()
@@ -1001,8 +1002,11 @@ fn get_local_hashes_at_level(
             for parent_id in parents {
                 let parent_storage_id = Id::new(*parent_id);
                 if let Ok(Some(parent_index)) = Index::<MainStorage>::get_index(parent_storage_id) {
-                    if let Some(children) = parent_index.children() {
-                        for child in children {
+                    {
+                        let _ = &parent_index;
+                        for child in Index::<MainStorage>::get_children_of(parent_storage_id)
+                            .unwrap_or_default()
+                        {
                             let child_id = *child.id().as_bytes();
                             if let Some(child_hash) =
                                 Index::<MainStorage>::get_hashes_for(child.id())
@@ -1074,9 +1078,10 @@ fn get_nodes_at_level(
             crate::sync::hash_comparison_protocol::collect_deleted_children_wire(&parent_index),
         );
 
-        let Some(children) = parent_index.children() else {
+        let children = Index::<MainStorage>::get_children_of(parent_id).unwrap_or_default();
+        if children.is_empty() {
             continue;
-        };
+        }
 
         for child in children {
             let child_storage_id = child.id();
@@ -1090,8 +1095,9 @@ fn get_nodes_at_level(
             };
 
             let child_hash = child_index.full_hash();
-            let is_leaf = child_index.children().is_none()
-                || child_index.children().map(|c| c.is_empty()).unwrap_or(true);
+            let is_leaf = Index::<MainStorage>::get_children_of(child.id())
+                .unwrap_or_default()
+                .is_empty();
 
             // Determine parent_id for this node (None for level 0)
             let parent_id_bytes = if level == 0 {
--- a/crates/storage/src/collections/vector.rs
+++ b/crates/storage/src/collections/vector.rs
@@ -584,39 +584,6 @@ mod tests {
     use crate::collections::{Root, Vector};
     use crate::store::MainStorage;
 
-    /// A `Vector` must hand back what was pushed, in the order it was pushed.
-    ///
-    /// `get(0)` returning the third push is not a subtle ordering nicety — it
-    /// is the type failing to be a vector. This reproduces the case that
-    /// matters: several pushes inside ONE call.
-    ///
-    /// `created_at` is the execution timestamp, so it is identical for every
-    /// push in a call, and the trie enumerates `(created_at, id)` — leaving a
-    /// random id to decide. `with_merge_mode` pins the timestamp to 0, which
-    /// is what a migration does and what the unit-test clock hides by
-    /// advancing between calls.
-    #[test]
-    fn positional_reads_follow_insertion_order_within_one_call() {
-        crate::env::reset_for_testing();
-        let mut v: Vector<String> = Vector::new();
-
-        crate::env::with_merge_mode(|| {
-            for key in ["k1", "k2", "k3"] {
-                v.push(key.to_owned()).expect("push");
-            }
-        });
-
-        let read: Vec<String> = (0..3)
-            .map(|i| v.get(i).expect("get").expect("present").into_inner())
-            .collect();
-
-        assert_eq!(
-            read,
-            vec!["k1".to_owned(), "k2".to_owned(), "k3".to_owned()],
-            "positional reads did not follow insertion order",
-        );
-    }
-
     #[test]
     fn test_vector_push() {
         let mut vector = Root::new(Vector::<_, MainStorage>::new);
--- a/crates/storage/src/collections/vector.rs
+++ b/crates/storage/src/collections/vector.rs
@@ -584,6 +584,71 @@ mod tests {
     use crate::collections::{Root, Vector};
     use crate::store::MainStorage;
 
+    /// `Vector::get(0)` can return the third push. **Known defect, not fixed.**
+    ///
+    /// Pushing k1, k2, k3 inside ONE call reads back as k1, k3, k2. `ChildInfo`
+    /// sorts by `(created_at, id)`; `created_at` is the host's execution
+    /// timestamp, identical for every write in a call (and a flat 0 under merge
+    /// mode), so every element ties and the random id decides.
+    ///
+    /// This branch is what exposed it. `Collection::insert` no longer
+    /// materialises the child cache to append the new id — an O(n) read the
+    /// child trie exists to remove — and its comment says leaving the cache
+    /// unpopulated is safe because "the trie is authoritative". True of
+    /// membership, not of ORDER: order was previously carried by the
+    /// insertion-ordered cache, and the trie cannot reconstruct what the sort
+    /// key does not distinguish. `master` passes this test for that reason.
+    ///
+    /// Two fixes were tried and are recorded here so they are not tried again:
+    ///
+    /// 1. **Make the write stamp strictly increasing.** Breaks a deliberate
+    ///    merge invariant — `an_entity_constructed_during_merge_loses_to_any_delete`
+    ///    requires merge-mode writes to stamp 0 so nothing can be older than
+    ///    them, while `interface.rs` and `merge.rs` read `created_at ==
+    ///    updated_at` as "never updated". Together those force ties in merge
+    ///    mode, so timestamps cannot carry order there.
+    ///
+    /// 2. **Order the ids** (a monotonic prefix). The trie buckets on the first
+    ///    4 nibbles of the id, so a monotonic prefix puts every entry of a
+    ///    collection in one bucket and turns the bounded lookup back into a
+    ///    scan — the cost this branch exists to remove.
+    ///
+    /// What remains is a per-child ordinal carried as the tiebreak,
+    /// `(created_at, ordinal, id)`. The ordinal cannot be recomputed by the
+    /// receiver — a node applying a synced `Add` would derive it from ITS child
+    /// count, which need not match the writer's — so it has to be assigned by
+    /// the writer and travel in the action payload. That makes it a sync
+    /// wire-format change, and it belongs to whoever owns the merge semantics
+    /// rather than to this branch.
+    ///
+    /// Failure is probabilistic: three random ids sort into insertion order
+    /// about one time in six, so this reproduces ~5 runs in 6 and CI scenario
+    /// `09-scenario-crdt-native` fails intermittently rather than always.
+    /// `master` passes it 10/10, which is what identifies the regression as
+    /// this branch's rather than pre-existing.
+    #[test]
+    #[ignore = "known: insertion order is lost for writes made in one call; see comment"]
+    fn positional_reads_follow_insertion_order_within_one_call() {
+        crate::env::reset_for_testing();
+        let mut v: Vector<String> = Vector::new();
+
+        crate::env::with_merge_mode(|| {
+            for key in ["k1", "k2", "k3"] {
+                v.push(key.to_owned()).expect("push");
+            }
+        });
+
+        let read: Vec<String> = (0..3)
+            .map(|i| v.get(i).expect("get").expect("present").into_inner())
+            .collect();
+
+        assert_eq!(
+            read,
+            vec!["k1".to_owned(), "k2".to_owned(), "k3".to_owned()],
+            "positional reads did not follow insertion order",
+        );
+    }
+
     #[test]
     fn test_vector_push() {
         let mut vector = Root::new(Vector::<_, MainStorage>::new);
--- a/crates/storage/src/collections/vector.rs
+++ b/crates/storage/src/collections/vector.rs
@@ -1,5 +1,6 @@
 //! This module provides functionality for the vector data structure.
 
+use crate::address::Id;
 use core::borrow::Borrow;
 use core::fmt;
 use std::mem;
@@ -165,21 +166,29 @@ where
     /// Push a value with an explicit `StorageType` on the new entry's element.
     ///
     /// Used by `AuthoredVector` to stamp each push with the executor as owner.
-    /// Returns the index of the newly inserted entry.
+    /// Returns the id of the newly inserted entry.
     pub(crate) fn push_with_storage_type(
         &mut self,
         value: V,
         storage_type: crate::entities::StorageType,
-    ) -> Result<usize, StoreError> {
-        let _ignored = self
+    ) -> Result<Id, StoreError> {
+        let (id, _item) = self
             .inner
             .insert_with_storage_type(None, value, storage_type)?;
-        let len = self.inner.len()?;
-        debug_assert!(
-            len >= 1,
-            "Vector::push_with_storage_type: len must be >= 1 after a successful push",
-        );
-        Ok(len - 1)
+        // The id, not `len - 1`.
+        //
+        // `len - 1` meant "the new entry is last", which held only while
+        // `insert` materialised the child cache and appended there. It no
+        // longer does — that populate was an O(n) read per insert — so
+        // enumeration comes from the child trie in `(created_at, id)` order,
+        // and `created_at` does not advance within a call. Entries pushed in
+        // one call therefore tie and the random id decides position, so
+        // `len - 1` could address a different entry (core#3637).
+        //
+        // An id is what the caller actually meant: it names the entry rather
+        // than describing where it currently sits, so it stays correct across a
+        // remote insert, which no positional answer can.
+        Ok(id)
     }
 
     /// Returns the storage id of the entry at `index`, or `None` if out of bounds.
@@ -272,6 +281,43 @@ where
         let old = mem::replace(&mut *entry, value);
 
         Ok(Some(old))
+    }
+
+    /// Replace the value stored under `id`, whatever position it occupies.
+    ///
+    /// The positional [`update`](Self::update) resolves `index` to an id and
+    /// then does exactly this. Callers that already hold the id should not go
+    /// back through a position: enumeration is `(created_at, id)` ordered over
+    /// a set other replicas insert into, so an index is only valid for as long
+    /// as no one else has inserted.
+    ///
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system.
+    pub(crate) fn update_by_id(&mut self, id: Id, mut value: V) -> Result<Option<V>, StoreError>
+    where
+        V: 'static,
+    {
+        // Same re-key as the positional path: two nodes concurrently replacing
+        // the same element with a freshly-built nested CRDT would otherwise
+        // mint divergent random internal ids.
+        super::rekey::rekey_nested_value(&mut value, id);
+
+        let Some(mut entry) = self.inner.get_mut(id)? else {
+            return Ok(None);
+        };
+
+        let old = mem::replace(&mut *entry, value);
+        Ok(Some(old))
+    }
+
+    /// Read the value stored under `id`, whatever position it occupies.
+    ///
+    /// # Errors
+    ///
+    /// If an error occurs when interacting with the storage system.
+    pub(crate) fn get_by_id(&self, id: Id) -> Result<Option<ValueRef<V>>, StoreError> {
+        Ok(self.inner.get(id)?.map(ValueRef::new))
     }
 
     /// Get an iterator over the items in the vector.
--- a/crates/storage/src/collections/vector.rs
+++ b/crates/storage/src/collections/vector.rs
@@ -584,50 +584,18 @@ mod tests {
     use crate::collections::{Root, Vector};
     use crate::store::MainStorage;
 
-    /// `Vector::get(0)` can return the third push. **Known defect, not fixed.**
+    /// A `Vector` hands back what was pushed, in the order it was pushed.
     ///
-    /// Pushing k1, k2, k3 inside ONE call reads back as k1, k3, k2. `ChildInfo`
-    /// sorts by `(created_at, id)`; `created_at` is the host's execution
-    /// timestamp, identical for every write in a call (and a flat 0 under merge
-    /// mode), so every element ties and the random id decides.
+    /// This reproduces the case that broke: several pushes inside ONE call.
+    /// `ChildInfo` sorts by `(created_at, order, id)`, and `created_at` cannot
+    /// separate them — it is the host's execution timestamp, identical for
+    /// every write in a call and pinned to 0 under merge mode. Before `order`
+    /// existed they tied and the random id decided, so this failed about five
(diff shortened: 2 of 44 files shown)


chefsale:
Read through this properly. The measurement discipline is the best part and I want it noted before the nits: keeping the "before" number in the tree commit-by-commit, and publishing that fix #2 moved the wall by *8 messages*, is what actually proves the diagnosis. Negative results are the hard thing to publish.

Ranked notes below. #1 I'd want before merge; #6 too, since it's cheap.

### 1. `insert_with` is a second implementation of the spine walk, with no oracle

`child_trie.rs:380-398` re-derives by hand what `refresh_spine` (`:222-234`) already does — same nibble walk, same `node.set`, same `count` delta. There is no test for `insert_with` or `children_with`; the test module covers only the `StorageAdaptor` path.

That's the highest-risk code in the PR. If the two walks ever disagree, a snapshot receiver reconstructs a *different root* from the sender — the exact failure class the snapshot half of this PR exists to fix, coming back through the fix for it. Per our convention for duplicated logic, the test writes itself: build the same child set through `insert` and through `insert_with` over a `HashMap`-backed read/write pair, assert byte-identical rows and identical roots. Ten lines, and the duplication stops being a latent fork.

### 2. `deleted_children` is the same pathology, in the same row, untouched

`index.rs:354` — still `Vec<Id>`, inline in `EntityIndex`, never pruned. Linear `retain` on every add (`:490-492`), linear `contains` + push on every removal (`:1097-1098`), and `hash_comparison.rs:207` resolves every entry to a signed `EntityDeletion` at wire-build time, so it's O(deletions) reads on the sync path too.

A context with churn walks back into the wall this PR just removed, through the same row. Doesn't belong in this PR, but it should be a filed issue linked from it — otherwise the fix reads as complete when it's half of one. The trie keyspace is the obvious home for tombstones too.

### 3. `count` went from derived to maintained, and it feeds contract-visible ids

Not folding `count` into the hash (`:108-111`) is the right call and well argued. The consequence is that drift is undetectable — nothing forks, `len()` just quietly lies — and `refresh_spine` uses `saturating_add_signed`, so a bug clamps silently instead of panicking.

Follow it through: `Collection::len` reads that count (`collections.rs:729`), and the motivating contract derives a message id from `len()` on every write. A count that drifts low produces **duplicate message ids** — silent data loss, no hash mismatch, no warning. `the_count_tracks_inserts_and_removals` covers the `insert` side only; nothing covers `insert_with`, which is where a snapshot-built trie gets its counts. A `debug_assert` reconciling against enumeration, or at minimum extending that test across `insert_with`, would close it.

### 4. The fold rule exists in three copies

`full_hash_from_trie` (`index.rs:541`) plus hand-inlined `Sha256(own_hash ‖ root)` in `add_child_to` (`:493-500`) and `update_parent_after_child_removal` (`:1103-1107`). The inlining is justified — those callers hold the root and shouldn't pay a row read to get it back — but the *rule* (skip-if-`EMPTY`, field order, no metadata) is triplicated, and drift in one copy forks replicas. A pure `full_hash_from_root(own_hash, root)` that all three call costs nothing.

### 5. Old rows misdecode silently — make them fail loudly instead

The PR is straight about removing `children` from field 3 meaning existing rows read children bytes as `full_hash`. But the response was to re-pin `minimal_struct_layout_compat` and the `borsh_layout` decoder to the new shape. Silent misdecode into a hash field is the worst available failure mode: a plausible, wrong hash with no error anywhere. If a version tag or a length sanity check can make an old row *reject*, that's worth more than settling the migration question, because it turns an undiagnosable divergence into a startup error.

### 6. `dump_root` is dark in exactly the PR that needs it

`context/primitives/src/client/mod.rs:990`. The warning is well written and warning-over-empty-list is the right instinct. But this is the tool for telling "matching children + mismatched own_hash" from subtree divergence, and this is the PR that moves every hash in the system and needs a coordinated upgrade. Worst possible moment for it to be out of service. `children_with` already exists and is already driven from a raw reader 200 lines away in `scope_projection.rs:258` — this looks like ~15 lines. I'd rather have it before merge than as follow-up.

### 7. Smaller

- `ChildTrie::clear()` (`:445`) has **no callers**, and enumerates then removes one child at a time — n reads plus n full spine refreshes. If anything ever calls it, that's the inline-blob cost with extra steps. Delete it or make it a keyspace-prefix drop.
- `rebuild_child_tries_after_snapshot` has no test, only the manual 25/25 measurement. It fixes a bug correctly described as a stable fixpoint that never heals, which is the strongest possible case for a `sync_sim` regression test.
- That rebuild is also a full-store scan plus `DEPTH+1` uncached point reads per link, re-reading the same spine rows for every sibling. One-time on install so probably fine, but grouping links by parent cuts it by most of 5x.
- `collections.rs:729` reaches `ChildTrie` directly from inside the storage crate where an `Index::child_count` would do (same for `index.rs:877` and `:1171`). Keeping the intra-crate surface at `Index` leaves `ChildTrie`'s public API justified purely by the three raw-store callers — which, for what it's worth, is a good answer to "why isn't this just inside `Index`" and is worth stating in the PR body, because right now the separate type reads as gratuitous when it isn't.

### On the compat question

Asking open-endedly which way to go will make this sit. Propose the answer you want. Also worth noticing: the migration the PR says doesn't exist is mostly already written — a one-time pass rebuilding tries from local index rows is `rebuild_child_tries_after_snapshot` pointed at local state instead of freshly-installed state. If no production state must survive, say that explicitly and land it as a flag day with a version bump. If some must, the pass is a few hours on top of code already in this diff. Either way it's a decision this PR can propose rather than defer.

xilosada:
Thanks — this was a genuinely useful read, and #1 in particular was the right thing to lead with. All seven addressed; pushed as four commits (`8c623e12`, `5437c230`, `c0d1aaf2`, `b1cabe27`).

Running the project reviewer over the same range in parallel turned up four more of the same shape as your #1 — call sites converted to the trie by reaching for `Index::<MainStorage>::get_children_of` where that type cannot work. Those are in `c0d1aaf2` and listed at the bottom, since one of them is worse than anything in either review.

### 1. `insert_with` oracle — done

`insert_with_writes_exactly_what_insert_writes`: same parent, same child set, one through each path, then byte-identical rows plus identical roots, counts and enumerations.

I mutation-tested it rather than trusting it green — dropping the count delta and skipping a spine level each fail it, the second on the root-hash assertion specifically. Ten lines was about right.

### 2. `deleted_children` — filed as #3635

Agreed it doesn't belong here and agreed the fix reads as complete without it. Filed with the reproduction shape and the cost-guard criterion, linked from the PR.

### 3. `count` drift — partly closed, honestly

The oracle test in #1 asserts counts agree between `insert` and `insert_with`, which is the "at minimum" you asked for and closes the snapshot-built-trie gap specifically. I did **not** add the `debug_assert` reconciling against enumeration. Your duplicate-message-id analysis is the strongest argument in the review and I'd rather not hand-wave it: the assert wants to be cheap enough for debug builds on a hot path, and I didn't want to guess at that under review pressure. Flagging it as still open rather than claiming it closed.

### 4. Fold rule — one definition

`full_hash_from_root(own_hash, root)`; the two inlined copies call it, `full_hash_from_trie` is now a two-liner over it. Cost nothing, as you said.

### 5. Old rows fail loudly — done, and it was worse than described

You were right that this is worth more than settling the migration question. Two things:

The mirror was a **prefix** — it stopped at `own_hash`. A prefix mirror accepts a row written under *any* later layout without complaint, which is precisely how the misread was possible. It now covers every field, so `from_slice` rejects a row that doesn't match this layout exactly, with a test pinning that an old-layout row is rejected rather than misread.

I also deleted the manual offset parse outright. It existed because index rows could be tens of kilobytes with children inline; they're ~122 bytes at 200 children now, so it was buying nothing and carrying the hazard. That removed `compute_root_hash_via_borsh` with it — the fallback for when the walk couldn't skip past `children`, which this PR had already orphaned.

Worth noting the correction: my first attempt made `from_slice` strict *without* completing the mirror, which broke three snapshot tests. The prefix was load-bearing and I hadn't checked.

### 6. `dump_root` — done, ~15 lines as predicted

Walks the trie with the same raw reader, and warns explicitly if a row read fails, since a short list would otherwise read as evidence during exactly the triage this exists for.

### 7. Smaller

- `clear()` — replaced by `drop_all`, which walks the node structure instead of removing children one at a time. It also now has a caller, which turned out to matter: **deleting an entity left its trie behind**. Trie rows are their own keyspace, so dropping `Entry` and tombstoning `Index` never touched them and GC can't either. Collection ids are deterministic, so a deleted-then-recreated field inherited the old children — ids whose data is gone, folded into the parent hash as a ghost root. That's a regression the trie introduced; fixed with a test.
- `rebuild_child_tries_after_snapshot` — regression test added, mutation-verified (making the rebuild a no-op fails it).
- Grouping links by parent — **not done**. Still a real ~5x on a one-time path; left as noted.
- `collections.rs` reaching `ChildTrie` directly — **not done**, but I took the second half of your point and put the justification in the PR body: the public API is earned by the three raw-store callers, not by the intra-crate ones.

### From the automated pass

- **`delta_store::load_rotation_log_direct`** — same shape as your #1, and the worst thing either review found. That function exists *specifically* to avoid `MainStorage` (it runs before `RUNTIME_ENV` is installed), yet the children walk went straight through it. With no env, `MainStorage` falls back to the process-local mock store, the lookup missed, and `.unwrap_or_default()` swallowed it — the rotation log read back **empty on every delta apply**, collapsing the writer set. The `scope_projection.rs` sibling was converted correctly; this one wasn't, and the orphaned `let _ = &index;` was the tell in both.
- **`persist_buffered_snapshot_entity`** wrote rows with no trie link, and runs from `drain_absorbed_leaves` long after the rebuild pass — so a late-draining entity was unenumerable forever. Both paths now share one link helper.
- **A rebuild failure was only warned**, then the sync published anyway: `verified_snapshot_root` reads the *shipped* `full_hash`, so an unlinked node verifies, matches its peers, and never looks like it needs repair. The warning promised a resync that could not happen. Now fails the sync with the marker set.
- **`handshake.rs` / `level_sync.rs`** used env-derived `Id::root()` where the correct id was already in hand.

Still open from that pass, both noted not fixed: trie rows are excluded from the sync-path double-read cleanup, and the rebuild scans the whole State column filtering by context in Rust.

### On the compat question

Fair, and taken. PR body now proposes **flag day + version bump, no data migration**, on the grounds that nothing in production runs a child-trie build — and, as you spotted, the migration is mostly written: `rebuild_child_tries_after_snapshot` pointed at local state is the pass. Happy to be overruled, but it's a proposal now rather than a question.

Verification after all of the above: 1267 tests across `calimero-storage` and `calimero-node --lib`, 0 failed; `cargo check --workspace --all-targets` clean; the only remaining warnings are the pre-existing layout-mirror fixtures whose fields are never read by design.

xilosada:
Following up on the one item I left explicitly open.

### 3. `count` drift — now closed (`e051e9703`)

I flagged this as still open rather than claiming it closed, on the grounds that I didn't want to guess at the cost under review pressure. Having looked properly, the concern is live: mero-chat computes `message_counter = self.messages.len() + 1` and hashes it into the message id, so a count that reads low mints duplicate ids with nothing anywhere to catch it. Your analysis was right and it survives the id derivation having since changed to sha256 — the counter is still an input.

**I did not implement the literal ask, and want to be straight about why.** Reconciling against a full enumeration makes every write O(n) in debug builds, on the path whose whole purpose is not being O(n) — a landmine of its own shape. What landed checks the *local* invariant instead: a node's count is the sum of what sits directly beneath it, at most 16 rows. Every node holding that is equivalent to the global property, and it fails at the level that introduced the drift rather than somewhere above it.

**The gap it actually closes is append-only**, which I only found by mutating rather than assuming. My first three attempts at a discriminating mutation were all caught by 120 existing tests — drift trips the "a slotless node cannot hold children" assertion as soon as a node empties. So the exposure is workloads that never remove. A chat channel never removes. Verified: an over-count injected at one level **passes** an insert-only test without the check and **fails** with it, naming the path.

**One correction.** My first version also called the check from `insert_with`, and it broke `rebuilding_after_snapshot_relinks_children_the_install_did_not`. That was the assertion being wrong, not the code: the rebuild buffers its writes and merges them once per child, so mid-walk the level below still holds its pre-write state and every check reports drift that isn't there. `insert_with` is a pure function of the rows it *reads* — deferred writes are the contract. The check is now on the `StorageAdaptor` path only, with that reasoning in the doc comment; `insert_with`'s counts stay pinned by the oracle test from #1, by equivalence.

### Also since the last pass

- **Rebased onto master.** The branch was 8 commits behind, which is why `E2E - Rust Apps` was red: master added the delegated-authorship job and its workflow file, and CI ran master's `.github` against a tree that didn't have the file yet. Green now.
- **`chat_wall` was broken** (`b96217929`) — it passed `sender_username`, which mero-chat has since dropped from `send_message`, so every call failed argument deserialization and the probe walled at 0. It reported that as a broken harness rather than a storage result, which is the only reason it was caught. Worth knowing it will rot again: it drives a contract in another repo and is `#[ignore]`d, so nothing here runs it.
- **Re-measured with it, ceiling raised to 60,000:** `send_message` gas 16.0M → 18.9M from n=100 to n=60,000, reads flat at 305, ~4.7ms. No write wall below 60,000. `get_messages` is now the binding limit — last OK at 30,000, exhausted at 32,000, and 30,000 spends 99.83% of the budget so the real ceiling is nearer 28,000. Against the 394-message write wall this started from, that's the append wall gone and roughly 76x usable capacity.

The remaining read cost is ours, not the app's: paging by index goes through `Collection::nth`, which materialises the whole child-id list from the trie on first use — one O(n) id walk per call, no payloads. That's the ordered-index gap; it wants its own PR.

### Still not done, unchanged

Grouping rebuild links by parent; `collections.rs` reaching `ChildTrie` directly (justification is in the PR body instead); trie rows excluded from the sync-path double-read cleanup; the rebuild scanning the whole State column and filtering in Rust. `deleted_children` remains #3635.

Verification: `cargo test --workspace` clean, `cargo clippy --workspace --all-targets` zero warnings, `cargo fmt --check` clean under 1.88.

crates/node/src/sync/level_sync.rs
```diff
@@ -984,8 +984,12 @@ fn get_local_hashes_at_level(
     match parent_ids {
         None => {
             // Level 0: get direct children of root
-            if let Some(children) = root_index.children() {
-                for child in children {
+            {
```
chefsale, line 990: 💡 **get_local_hashes_at_level re-reads the index row it already has, via get_children_of**

Both branches of `get_local_hashes_at_level` already fetch and decode `root_index`/`parent_index` (lines 975 and 1007) and only use the fetched value for `let _ = &root_index;` / `let _ = &parent_index;` before calling `Index::<MainStorage>::get_children_of(root_id)` / `get_children_of(parent_storage_id)`. `Index::get_children_of` (crates/storage/src/index.rs) is implemented as `let _index = Self::get_index(parent_id)?.ok_or(...)?; Ok(<ChildTrie<S>>::new(parent_id).children())` — i.e. it re-reads and re-borsh-decodes the same index row purely to repeat an existence check the caller already performed, then walks the trie. This is the exact 'doubled read on the sync hot path' anti-pattern this same PR explicitly removes in `hash_comparison_protocol.rs` (`collect_leaves_recursive`, `get_local_tree_node`) and in `delta_store.rs::load_rotation_log_direct`, each with a comment explaining why `get_children_of` must be avoided when the index is already in hand. `level_sync.rs` was not updated to match, so `get_local_hashes_at_level` — which runs per parent, per level, in the LevelWise sync protocol this module says is 'optimized for wide, shallow trees' with 'avg_children_per_level > 10' — pays one avoidable extra index-row read+decode per parent.

**Suggested fix:**
```
Call `calimero_storage::child_trie::ChildTrie::<MainStorage>::new(id).children()` directly (add `use calimero_storage::child_trie::ChildTrie;`) instead of `Index::<MainStorage>::get_children_of(id)` at both call sites (line 992 and lines 1010-1012), since `root_index`/`parent_index` already prove the entity exists.
```

chefsale (commented):
## 🤖 AI Code Reviewer

**Reviewed by 3 agents** | Review time: 0.0s

💡 1 suggestions, 📝 1 nitpicks. See inline comments.

---

<sub>🤖 Generated by [AI Code Reviewer](https://github.com/calimero-network/ai-code-reviewer) | Review ID: `review-60a57d57`</sub>

crates/storage/src/collections/vector.rs
```diff
@@ -274,6 +283,56 @@ where
...
+
+        let Some(mut entry) = self.inner.get_mut(id)? else {
+            return Ok(None);
+        };
+
+        let old = mem::replace(&mut *entry, value);
+        Ok(Some(old))
+    }
+
+    /// Read the value stored under `id`, whatever position it occupies.
```
chefsale: 📝 **Nit: doc comment for get_by_id got merged onto collection_id, leaving get_by_id undocumented**

The doc block `/// Read the value stored under \`id\`, whatever position it occupies. /// # Errors /// If an error occurs when interacting with the storage system.` is attached to `collection_id()`, immediately followed (in the same doc block, no blank-line-then-`///` separation issue but clearly a different topic) by `/// The id of the inner collection — the parent every entry of this vector is a child of. ... Exposed so the authored layer can check...` which is `collection_id`'s real doc. The function that the first paragraph actually describes, `get_by_id()`, has no doc comment of its own. Rustdoc output for `collection_id` will now read as a description of a getter method, which is confusing and wrong.

**Suggested fix:**
```
Split the doc comment: move the "Read the value stored under `id`..." paragraph (plus its `# Errors` section) onto `get_by_id`, and keep only the "The id of the inner collection..." paragraph on `collection_id`.
```

crates/node/src/sync/level_sync.rs
xilosada: Fixed in 61c19f71a — and you were right that this is the same anti-pattern, not a nit. It was a call site this PR should have converted and didn't.

Two things I'd add to your read:

**The `unwrap_or_default()` mattered as much as the extra read.** As you noted for `collect_leaves_recursive`, the only failure `get_children_of` can return here *is* that re-read — so a `DeserializationError` on a row this function had already proved decodable became "parent with no children", and got fed to hash comparison as fact. Going to the trie removes the swallow along with the doubled read.

**I dropped the `let _ = &root_index;` / `let _ = &parent_index;` bindings** rather than keeping them alongside the fix. `get_index` is itself the existence check, so `Ok(Some(_)) => {}` states that directly. That orphaned binding is what tipped off both the `delta_store::load_rotation_log_direct` bug and the snapshot one earlier in this branch, so I'd rather it stop reading as an idiom.

Verified: `cargo test --workspace` clean, clippy zero.

crates/storage/src/collections/vector.rs
xilosada: Fixed in 61c19f71a. Correct diagnosis — I inserted `collection_id` anchored on the `get_by_id` signature, so it landed between that function's doc comment and the function, and rustdoc would have shipped the getter's docs on `collection_id` with `get_by_id` bare.

Split as you suggested: the "Read the value stored under `id`" paragraph and its `# Errors` section back onto `get_by_id`, only the inner-collection paragraph on `collection_id`.

PR #3638, opened 2026-08-24
test(e2e): wait for the migration instead of sleeping past it

rtb-12:
Thirty-three scenarios slept a fixed 15s after `upgrade_group` and then read, with nothing establishing that the migration had finished. A cohort stalled part-way through reads as successfully as a converged one, so the sleep supplied the only ordering and no assertion at all.

## Changes

**Migration suite — 30 scenarios.** `assert_migration_complete` already polls `get_migration_status` until every member reports a converged schema with zero residue, and aborts early if a member enters `failed`. Its own docstring says it exists to save authors from hand-rolling a `wait` loop, which is what these were.

- 5 scenarios already called it, but *after* the reads it should have ordered. Those now call it before, and the sleep goes.
- 25 had no barrier at all. They gain one in place of the sleep.

**3 scenarios outside the migration suite** slept before a cross-node read of a specific value. Those get `expect:` on the barrier that already precedes them, so the wait ends when the value arrives rather than when the hashes agree, which can happen a few milliseconds early:

```yaml
    expect:
      method: get
      args:
        key: "alpha"
      equals:
        output: "from_node_1_first"
```

**36 sleeps removed, 491 seconds.** Every barrier exits as soon as its condition holds, so the happy path gets faster, not slower.

## Deliberately left alone

- `22-scenario-identity-downgrade.yml` and `39-missing-abi-refused.yml` carry `expected_failure` on the upgrade. A completion barrier there would assert the opposite of the test.
- `35-two-namespaces-coexist.yml` captures no namespace id to poll on.
- The `sync-resilience` sleeps, which are measured against `with_idle_connection_timeout` rather than guessed.

## Test plan

The suite is the test, and this is a correctness change as much as a timing one: where no barrier existed, a stalled migration previously passed. If any scenario now fails, that is the barrier reporting something the sleep was hiding, and the failure names the member and its residue rather than surfacing as a confusing read.

- All 133 workflows validate against merobox master: 0 failures
- 36 `type: wait` steps removed, confirmed by diffing on `seconds:` at step indent; no other step type touched
- 2 incidental blank-line removals, no other whitespace churn

### DIFF
--- a/apps/scaffolding-e2e/workflows/account-device-revoke-lockout.yml
+++ b/apps/scaffolding-e2e/workflows/account-device-revoke-lockout.yml
@@ -480,9 +480,22 @@ steps:
   # device-addressed, everything above would pass and then the node would
   # quietly re-fetch the rotated key on its next sync round.
 
-  - name: Give the recovery loop several rounds to try
-    type: wait
-    seconds: 20
+  - name: Wait for the recovered write to reach the revoked device's replacement
+    type: wait_for_sync
+    context_id: '{{context_id}}'
+    nodes:
+      - account-revoke-node-1
+      - account-revoke-node-3
+    timeout: 60
+    trigger_sync: true
+    # Waiting on the read itself, not just a converged hash: agreeing
+    # hashes can precede the delta actually applying.
+    expect:
+      method: get
+      args:
+        key: "from-laptop"
+      equals:
+        output: "laptop-wrote-this"
 
   # Asserted as the PROPERTY, not as a log signal, because this shape cannot
   # trigger the pull path at all. A node asks for a key when it is a member of a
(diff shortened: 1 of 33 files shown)


rtb-12:
Closing: the premise is wrong.

These migrations are lazy - a node migrates when it serves a read. This PR puts `assert_migration_complete` *before* the reads, so it waits for something only the steps it blocks can produce. Every migration-suite job shows the same thing:

```
attempt 1: 0/2 migrated, 2 in-progress, 0 failed
attempt 7: 0/2 migrated, 2 in-progress, 0 failed
```

`0 failed` throughout - nothing in core errored, the migration simply was never triggered.

Three further problems:

- Five scenarios already called `assert_migration_complete` **after** the reads, which was correct. Moving it earlier broke them.
- The deleted sleeps were not waiting for the migration. They were waiting for the upgrade metadata to gossip to the second node before its lazy read, which the comment I removed as stale had said explicitly.
- `account-device-revoke-lockout` failed separately: the replacement carries `trigger_sync: true`, forcing the sync that scenario verifies does not happen on its own.

Replacing with an additive change: add `assert_migration_complete` **after** the reads in the scenarios that never check completion, leaving every sleep in place.

PR #3644, opened 2026-08-24
ci(e2e): fail the run when a node panicked

rtb-12:
Exactly one scenario in the suite greps its node logs for a panic (`apps/kv-store/workflows/ephemeral-presence-e2e.yml`). The other 89 matrix jobs go green over a node that panicked and restarted, or that panicked on a path the scenario never asserts against, because nothing reads the logs for it.

## What it matches

`merod` installs a panic hook (`crates/merod/src/main.rs:181`) that logs the message, thread and location as structured fields under `Application panic occurred`, then chains to the previous hook at `:217` — so a panic also leaves Rust's default `thread '...' panicked at`.

The gate matches either. The structured line is absent for a panic raised before the hook is installed, and the default line is what a non-merod process leaves. Matching the two known formats rather than the bare word `panicked` keeps prose out of it.

It sits with the existing hard gates over `docker-logs/`, so it covers every matrix job rather than the single scenario that thought to ask.

## Test plan

The gate was run against synthetic log directories and observed failing before being believed:

| input | expected | result |
| --- | --- | --- |
| clean logs | pass | rc=0 |
| `Application panic occurred` with fields | **fail** | rc=1 |
| `thread 'tokio-runtime-worker' panicked at crates/node/src/x.rs:42:9` | **fail** | rc=1 |
| the same line carrying ANSI colour escapes | **fail** | rc=1 |
| prose containing the word `panicked` | pass, no false positive | rc=0 |

The ANSI case matters because escapes sit between a field name and its `=` in node logs; the existing op-store gate strips them for the same reason and this follows it.

## Expectation

If a scenario is currently panicking, this turns that from invisible into a red run naming the workflow, the file and the message. That is the intended outcome, not a regression — the suite has had no way to see it.

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -1030,7 +1030,8 @@ jobs:
               # in node logs, so the raw pattern never matches.
               grep -rhE "Application panic occurred|thread '[^']*' panicked at" docker-logs/ \
                 | sed -E 's/\x1b\[[0-9;]*m//g' \
-                | grep -oE "panic\.message=[^ ]*|thread '[^']*' panicked at [^ ]*" \
+                | sed -E "s/.*(panic\.message=.*) panic\.thread=.*/\\1/; \
+                         s/.*(thread '[^']*' panicked at [^,]*).*/\\1/" \
                 | sort | uniq -c | head -10
               exit 1
           else
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -1007,6 +1007,36 @@ jobs:
               echo "no gov_divergence_pull_triggered markers — no scope_root governance divergence pulled this run"
           fi
 
+      - name: Assert no node panicked (HARD gate)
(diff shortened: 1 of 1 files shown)


rtb-12:
Superseded by calimero-network/merobox#377, which puts the check in merobox instead.

Five workflows in this repo drive merobox (`app-migration-e2e`, `e2e-rust-apps`, `e2e-rust-apps-release`, `fuzzy-load-test`, `sync-regression`). This PR gated one of them, so the same 30 lines of shell would have had to be pasted into four more and then kept in step. merobox starts the nodes and already persists their logs, so the check covers every scenario in every workflow — and every other repo using merobox — from one place.

The prototype did its job before being closed: on its first run it caught a real panic in `namespace-owner-writes-in-a-foreign-open-subgroup` — `Mailbox has closed`, 1 occurrence — in a scenario that is green on master today. That finding stands and deserves its own issue.

It also surfaced a bug in itself that the merobox version now has a regression test for: the message extraction stopped at the first space, reporting `panic.message=Mailbox` for `Mailbox has closed`.

PR #3649, opened 2026-08-25
fix(node): deliver the namespace's application when a member joins it

rtb-12:
The application belongs to the namespace — the root owns `target_application_id` and subgroups inherit it — but only context sync ever delivered it.

**Before:** joining a namespace installed the app only if that namespace already had a context. Auto-follow joined that context, and the context sync fetched the bundle. A namespace with no context yet delivered nothing.

**After:** joining a namespace installs the app, context or not.

That gap is why a member who joins and then creates the namespace’s **first** context failed on `application not found` — for an application the invitation named and the inviter held.

## Changes

- `BlobRequest` carries `namespace_id`, and a bundle is authorized against that namespace’s `target_application_id` instead of a context’s. Optional and `#[serde(default)]`, so older peers keep working in both directions.
- On join, fetch and install the bundle when the application is absent. Best-effort — a joiner that reaches no provider is where it was before.
- Reverts the two explicit `install_application` steps from #3645; they were standing in for this.

## Test plan

Five namespace-scoped tests, each observed failing first:

| break | result |
| --- | --- |
| ignore the namespace, serve any app blob | 1 failed |
| match bytecode but not the compiled artifact | 1 failed |
| restored | 14 passed |

All 9 pre-existing tests still pass. `cargo fmt` and clippy clean. The two reverted workflows are the end-to-end proof.


chefsale:
We had this behaviour before and we deleted it, I think we want to force the user to install the app either from registry or if for local development manually. If you go to any app it gets installed before joining. cc: @xilosada also to chip in on this.

Issue #3648, opened 2026-08-25
Pairing should be scoped by application (default: all), not by namespace


rtb-12:
## In one sentence

Pairing a second device should be scoped by **application**, defaulting to all,
instead of by namespace.

## Before and after

**Today**

```
pair a device  ->  you must name ONE namespace
                   the link fans out to every namespace anyway
                   the new device only listens on the one it was told
                   so one workspace works, the rest go quiet
```

The caller has to name a namespace, cannot find out which namespaces are valid,
and gets `500 Internal server error` when it guesses wrong.

**Agreed**

```
pair a device  ->  you choose APPS (default: all)
                   core resolves apps -> namespaces
                   the link fans out to exactly those
                   the new device installs those apps and joins their contexts
```

The user answers one question they can actually answer. Namespace-level pairing
goes away; it was the least useful granularity and nobody asked for it.

## Why application is the right unit

A namespace is an implementation unit, so scoping by it leaves the rest of the
outcome undecided. An application list settles four things at once:

1. which namespaces receive the device binding
2. which scope keys are delivered
3. which applications the new device installs
4. which contexts it joins

The resolver already exists:
`GET /admin-api/namespaces/for-application/:application_id`
(`crates/server/src/admin/service.rs:347`).

## Decisions

| # | Question | Decision |
| --- | --- | --- |
| 1 | One device identity, or one per namespace? | **One.** Matches today's behaviour. |
| 2 | Namespaces and apps added after pairing? | **Automatic.** New namespaces bind; a new app installed on one device is visible to the others. |
| 3 | How does a client list your devices? | **New account-level endpoint**, showing per-device scope. |
| 4 | Adding apps to an already-paired device? | **Extend scope**, not re-pair. |
| 5 | Does revoke stay all-or-nothing? | **Yes.** Security action; scoping it reintroduces a fixed bug. |
| 6 | Where does a device name live? | **Alias**, like contexts and applications. |
| 7 | Where does the new device get bytecode? | **Same mechanism as a namespace invite.** |
| 8 | What happens to a revoked device? | **Falls back to its own account.** |

Decision 1 also settles the tension noted in the first comment thread.
`crates/governance-store/src/node_device.rs` justifies the per-namespace device
row as preventing a peer in one namespace from testing whether a device in
another is the same machine. `docs/src/content/docs/protocol/accounts.mdx:100-108`
already says that correlation is trivially available anyway - the genesis names
[...]

PR #3657, opened 2026-08-26
feat(account)!: account-level device pairing, scoped by application

### LINKED ISSUE #3650: Device pairing refusals are all flattened to 500 Internal server error, so no client can explain a failure
## Summary

`parse_api_error` (`crates/server/src/admin/service.rs:633-712`) maps a known
set of `ContextError`s to real status codes and falls through to `500` with the
literal body `{"error":"Internal server error"}`. Both device pairing handlers
raise every one of their refusals with `eyre::bail!`, so all of them take that
fallthrough.

Five distinct refusals are currently indistinguishable on the wire:

| Refusal | Raised at |
| --- | --- |
| The confirmation code does not match the key material | `crates/context/src/handlers/pair_device_complete.rs`, gate 4 |
| The statement fails verification | same file, gate 3 |
| This node holds no current scope key for the namespace | `pair_device_complete.rs:190` |
| This node has no namespace identity for the namespace | `pair_device_complete.rs:76` |
| The device is already linked to a different account here | `crates/governance-store/src/node_device.rs:709` |

Only shape errors reach a caller usefully: hex width and empty-code checks in
`ValidatedJson` return `400` with a real message. Everything semantic is a
`500`.

## Impact

No client can tell a user what went wrong, and the two most likely real-world
failures have opposite remedies:

- *"you mistyped the confirmation code"* - try again, read it off the other
  screen
- *"this machine already started as its own account here"* - pair a fresh
  install, or revoke the existing device first

Both arrive as `Internal server error`.

The last row is the worst case, because it is an expected outcome rather than
an edge case: it is what happens whenever someone installs the app, uses it,
and only then tries to pair the machine to an existing account. Its own message
text explains the fix precisely -

> this node already holds device {} for account {} in {namespace:?}, and it is
> linked - its replica state is held under that id. Moving a machine between
> accounts means revoking the existing device first

\- and that text goes only to the node log.

There is a second-order effect on #3648. `pair-complete` requires the caller to
name a namespace where this node holds the current scope key, and nothing
reports which namespaces qualify, so callers guess. Because a wrong guess is
indistinguishable from a wrong confirmation code, a client cannot even retry
intelligently against another namespace - it has no way to know which of the
two happened.

## Steps to reproduce

1. Pair a device through `POST /admin-api/namespaces/:namespace_id/account/pair-init`.
2. Call `POST /admin-api/namespaces/:namespace_id/account/pair-complete` with
   the returned payload but a wrong `confirmationCode` of the correct shape,
   e.g. `DEAD-BEEF-DEAD-BEEF`.

**Actual:** `500 {"error":"Internal server error"}`.

**Expected:** a 4xx naming the confirmation code as the mismatch.

`apps/scaffolding-e2e/scripts/account-pair-refusals.sh` already exercises this
case and the substituted-key case; both currently assert only that the call
fails.

## Notes toward a fix

Typed `ApiError` variants for the refusals in the two context handlers, mapped
in `parse_api_error` rather than falling through. A minimum useful split:

- `403` - authority refusals: not the account holder, device already linked to
  another account
- `400` - verification failures: confirmation code mismatch, invalid statement
- `409` or `503` - this node is not ready in that namespace: no namespace
  identity, no current scope key

Same class as #3400 (`get_group_info` returns 500 for a non-member instead of
the intended 403), and the same root cause: a domain refusal raised as an
untyped error and flattened at the boundary.

## Related

- #3648 - pairing should be scoped by application, where this is the reason a
  client cannot recover from a bad namespace guess
- #3400 - same flattening on `get_group_info`

rtb-12:
## Summary

Device pairing becomes **account-level and scoped by application**, instead of being keyed to a single namespace.

A device is a replica of an *account*, and an account speaks in many namespaces. Pairing did not reflect that: it took one namespace in the path, so the device it minted listened on one topic while the link the holder published fanned out across every namespace the account spoke in. Every other namespace stayed silent, and a device that joined no context could not run the application at all.

Three changes make pairing follow the account rather than freeze at one namespace:

1. **The device is scoped by application, not by namespace.** "Which applications may this device use" is a question the person doing the pairing can answer; "which namespaces" is an implementation unit they never named. The holder resolves applications to namespaces itself and publishes the link to exactly that set.
2. **One device across the whole namespace set.** The certificate is root-signed and names no namespace, so minting per namespace would hand the holder several ids to certify and several confirmation codes to read aloud for one machine.
3. **Bindings are re-established as the account grows.** The holder persists every `DeviceCert` it signs, so a namespace created or joined *later* binds the cached devices whose scope covers it. Pairing stops being a snapshot taken at `pair-complete` time.

---

### The routes

| | Before | After |
| --- | --- | --- |
| Pair a device | `POST /namespaces/:namespace_id/account/pair-init` and `pair-complete` | `POST /admin-api/account/pair-init` and `pair-complete` |
| What the path names | one namespace | nothing; the account is the subject |
| Repair a device's bindings | no route | `POST /admin-api/account/devices/:device_id/relink` |
| List this account's devices | no route | `GET /admin-api/account/devices` |
| List this account's applications | no route | `GET /admin-api/account/applications` |
| Revoke | `POST /namespaces/:namespace_id/account/revoke` | unchanged; a revocation names the namespace it publishes into |

The namespace-scoped pairing routes are **removed**, not deprecated. Keeping them would have meant two ways to pair with different reach, and the narrower one silently produces the half-bound device this change exists to prevent.

### What a device is scoped by

| | Before | After |
| --- | --- | --- |
| Scope named at | `pair-init`, as one namespace in the path | `pair-complete`, as an application list in the body |
| Naming nothing | not expressible | every namespace this node takes part in |
| Device identity | one per namespace paired | one per device, whatever the set |
| Confirmation codes to read aloud | one per namespace | one |
| What the namespace set on `pair-init` decides | everything | only what the new device *subscribes* to |

The two lists are deliberately independent. The namespace set decides what the device listens on; the application set decides where the holder publishes. A binding published where the device is not yet listening is picked up when it subscribes.

### When a binding exists

| | Before | After |
| --- | --- | --- |
| Namespaces bound at pair time | the one in the path | every one the application scope resolves to |
| A namespace joined afterwards | never bound; the device silently never sees it | bound on gain, from the cached certificate |
| Repairing drift | re-pair the device | `relink`, with no handshake, no code, and the device offline |
| Certificates the holder keeps | none | every `AccountProof<DeviceCert>` it signs, plus the pairing's scope, node-locally |
| Certificates signed by a sibling device | not observable | cached when the link op applies |

Auto-bind on namespace gain is best-effort by construction and cannot fail the join or create that triggered it.

### What a refusal returns

| Refusal | Before | After |
| --- | --- | --- |
| Mismatched confirmation code | 500 | **400** |
| Unverifiable device statement | 500 | **400** |
| Node not ready in the named scope | 500 | **409** |
| Caller is not the account holder | 500 | **403** |
| Device linked to another account | 500 | **403** |
| Relink of a revoked device | 500 | **403** |
| Unknown device | 500 | **404** |

A blanket 500 told a client nothing it could act on: retrying a mistyped code is pointless, retrying a scope this node does not yet take part in is exactly right, and neither is a broken node. Closes #3650.

---

## Why this is required for application pairing

A paired device is a member of nothing and follows no context. That is the point of separating the two ids, and it is also why it never reaches the context bootstrap that installs an application. Before this change such a device held the stub row the `ContextRegistered` apply wrote and had nothing to run.

Acquisition drives the joiner's existing install chain off that row instead, triggered where the application metadata first materialises. The stub is distinguished from a real install by `size == 0` rather than by presence, because the stub names the creator's real blob and so satisfies a bare `has_application` check while carrying no package, version or size.

`GET /admin-api/account/applications` is the other half: it is the only route by which a device that is a member of nothing can be told which applications its account speaks in, because a namespace summary is withheld from non-members.

---

## Breaking changes

- `POST /namespaces/:namespace_id/account/pair-init` and `/pair-complete` are removed. Callers move to `POST /admin-api/account/pair-init` and `/pair-complete`, which take the namespace set in the body and an optional application scope instead of a namespace in the path.
- `meroctl` pairs over the account-level routes.

## Fixes folded in

- A degenerate KEM key no longer aborts a pairing half-done: the scope key is wrapped before the link publishes, and a failing namespace is skipped rather than failing the rest.
- `bytecode_id` is healed from the registration op on replicas that learned a context without an invitation. It previously stayed zeroed, which disabled upgrade cascades on paired devices.
- `GET /admin-api/identity` no longer reports a revoked device as this node's identity; it falls back to the node's own account root.
- Sync's blob share passed its rogue-sender ceiling to `add_blob`, which reads that argument as an exact length. A receiver holding a size-0 stub advertises 0, so the cap became the 500 MiB maximum and every correctly received bundle failed `fatal: blob size mismatch` after being stored. The cap is applied to the reader instead; integrity still rests on the received blob id.

Implements #3648, except device aliases, which are #3658.

## Test plan

Per-crate suites, all green:

| Crate | Tests |
| --- | --- |
| `calimero-governance-store` | 618 |
| `calimero-context` | 271 |
| `calimero-server` | 131 |
| `calimero-store` | 81 |
| `calimero-context-client` | 37 |
| `calimero-server-primitives` | 32 |

Full workspace: 5181 passed, 0 failed. `cargo fmt --check` and `cargo clippy --all-targets` clean.

Every new branch and refusal is covered by a test observed failing under a deliberate mutation before being trusted.

End-to-end, `account-pairing-application-scope.yml` is registered in the e2e matrix and green. It covers scope-narrowed pairing, relink widening, auto-bind on a namespace gained afterwards, bytecode arrival on a device that is a member of nothing, revoke fan-out, the identity fallback, and each typed refusal status over the wire. Node-2 appears in no `wait_for_sync`: a namespace barrier reads `groupStateHash`, which a node answers only for a member, so its state is polled instead.


chefsale:
Heads-up on a collision, with the specific thing that will bite — #3664 touches
`governance-store/src/node_device.rs`, `governance-store/src/test_fixtures.rs`, and
`context/src/handlers/pair_device_complete.rs`, which this PR also edits heavily.

**The mechanical half.** `ensure_account_root` is gone, split in two:

- `require_account_root()` — reads, errors when there is none. What production code
  wants.
- `provision_account_root()` — mints. Called only from `merod init`.

Every call site becomes a compile error naming itself, so this part is a rename.

**The half the compiler will not find, and the reason I'm commenting.** `test_store()`
in `governance-store/src/test_fixtures.rs` now *provisions a root*, because every node
that ran `merod init` has one. Use `test_store_without_account_root()` to model a
root-free node deliberately.

That single behaviour change broke **28 tests across three crates**, and every one of
them compiled cleanly — `cargo check --workspace --all-targets` was green the whole
time. They fail at runtime, in test *setup*, because they had been silently relying on
a read minting a root as a side effect. Three shapes:

1. A test asserting the store starts root-free (now it doesn't).
2. A restore test whose precondition proved recovery wasn't reading a leftover root.
3. A shared helper — `create_test_store()` in `context/src/handlers/update_application`,
   and `SimStorage::new` in `node/tests/sync_sim/storage.rs` — where only a couple of
   many callers actually reach a path needing a root.

The `sync_sim` one is worth calling out because it was genuinely hard to diagnose: 27
of its 28 failures reported only `"responder failed"`, since the sim doesn't surface
the responder's inner error. Exactly one test asserted on error *text*, and that's the
only reason the cause was legible rather than looking like a sync bug.

**So:** after rebasing, if you see failures in tests that construct their own `Store`,
check whether they need `provision_account_root()` in setup before digging into logic.
The rule that now holds: any test building its own `Store` needs a root, and nothing
will warn you.

Happy to do the rebase onto #3664 myself if that's easier — say the word.

chefsale (approved):
ship it , get rid of the scripts next -> integrate them into merobox.

PR #3663, opened 2026-08-26
feat(merod): provision the account root at init, not on first use

chefsale:
Groundwork for #3430. First of two: this names the moment a node gets an account; the next lets a node have none.

## The problem

An account root is created **lazily** by `ensure_account_root`, reachable from four production call sites — including `account_for_group`, which is a **resolver**. So a read can bring an identity into being as a side effect, and the moment a node acquires an account depends on which request happened to arrive first.

Nothing is wrong with the key that results. What is wrong is that there is no point in a node's life you can name as the one where it got an account.

`init` now names it.

## Why this is a prerequisite, not tidiness

A node meant to hold **no** root — root in cold storage, this device enabled by an imported certificate — cannot exist while any read will mint one. `--no-account-root` would be a lie today: the flag would set nothing, and the first resolver call would hand the node an account anyway.

Those call sites can only start refusing once the root legitimately comes from somewhere else. This is that somewhere.

**The flag is deliberately not in this PR.** It arrives with the refusals that make it true, rather than as an option that quietly does nothing.

## Behaviour is unchanged

Every existing deployment gets the same key it would have got on first use — this only moves *when*. The lazy path stays and simply finds the root already present.

## How the ordering was found

By getting it wrong. Making the resolver refuse first looked like the small, self-contained step. Two tests said otherwise, and both describe real states rather than fixtures:

- `the_creation_time_resolver_trap_is_gone` covers *"exactly the state `init` runs in"* — no context→group row, no root. The resolver's mint is what provisions a fresh node, so refusing there breaks creating your first context.
- `a_revoked_device_stops_naming_the_account_it_spoke_for` covers a node **paired into someone else's account** — it never had a root — whose device is then revoked. It previously minted one at that moment to have something to name.

That second case still needs a deliberate answer before the refusals can land: does such a node error, or is it required to hold its own root? Worth settling in review of the next PR rather than assumed.

## Verified

- New test asserts the root exists after `init`, before anything is served. Mutation-tested: removing the provisioning call fails it and nothing else.
- Manually: a fresh `merod init` logs the account, and `merod account export` returns a mnemonic immediately afterwards.
- `merod` + `calimero-governance-store`: 661 tests pass. `cargo fmt` and `clippy -D warnings` clean.

### DIFF
--- a/crates/governance-store/src/node_device.rs
+++ b/crates/governance-store/src/node_device.rs
@@ -419,11 +419,32 @@ impl<'a> NodeDeviceRepository<'a> {
 
         let existing = self.account_root()?;
         if let Some(previous) = &existing {
-            if !force {
+            // Gate on whether the root being replaced has CERTIFIED something, not
+            // merely on whether one exists.
+            //
+            // The cost this guard protects against is losing a root that devices
+            // depend on: replacing it orphans every certificate it signed, and
+            // there is no second copy. A root that has certified nothing carries
+            // none of that cost.
+            //
+            // That distinction became load-bearing when `init` started
+            // provisioning a root. The primary recovery flow is total loss — wipe
+            // the disk, `merod init`, restore the phrase — and the fresh node now
+            // always has a root, so gating on existence turned every restore into a
+            // `--force` ceremony. That trains an operator to reach for the flag
+            // precisely where it is dangerous: on a node where the root being
+            // replaced HAS certified devices and the loss is unrecoverable.
+            //
+            // One device row, so this is a read rather than a scan.
+            let certified_something = self
+                .get()?
+                .is_some_and(|row| row.account == previous.account());
+            if certified_something && !force {
                 eyre::bail!(
-                    "this node already has an account root ({}), and replacing it \
-                     cannot be undone: a root that has already certified devices \
-                     has no second copy",
+                    "this node already has an account root ({}) that has certified \
+                     a device, and replacing it cannot be undone: the certificates \
+                     it signed have no second copy. Export it first, then pass \
+                     --force to replace it",
                     previous.public_key()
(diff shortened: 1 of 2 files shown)


chefsale:
Fixed. The `account-root-backup-restore` failure was a real consequence of this change, not a flake.

**What broke.** The scenario's last phase is total-loss recovery: wipe the disk, `merod init`, import the backed-up phrase. Provisioning a root at init means the fresh node always has one, and `try_import_account_root` refused on mere existence — so every restore now needed `--force`.

**Why I did not just add `--force` to the runbook.** That flag exists to stop someone destroying a root whose certificates cannot be recovered. Making the ordinary recovery path require it trains an operator to reach for it exactly where it is dangerous — on a node where the root being replaced *has* certified devices.

**The fix** makes the guard ask what its own message always claimed: has this root **certified** anything? A root that has signed no device carries none of the loss, so overwriting it is safe. One that has still needs `--force`, and the error now names which case you are in and what to do about it. One device row, so it is a read rather than a scan.

Verified against the binary, which is the flow the e2e drives:

```
init → export → wipe → init → import (no --force)
  Replaced the account root 4TeUciDv…
  Imported account root BM35F989…
```

Two unit tests: the refusal (with a device enrolled to make replacement costly) and the restore-over-an-untouched-root case — which the e2e exercised but nothing unit-tested, which is why this reached CI.

Also worth flagging for review: `clippy -D warnings` caught a `#[test]` I had stranded above a doc comment, which silently detached a test from its attribute. The suite still reported the same count, because a new test took the old one's place in the ordering — so the number could not reveal it. I added a scan for that shape while fixing it; there are none left.

593 tests pass, `fmt` and `clippy -D warnings` clean.

PR #3690, opened 2026-08-28
feat(blob): bound the DHT lookup and let a caller opt out of waiting

xilosada:
An image shared in a chat never loaded on the other nodes — it spun forever. Not slow: **forever**.

`NetworkClient::query_blob` awaits a oneshot resolved only when kad reports the query progressed, with no timeout of its own. A lookup that doesn't terminate parks the task for the life of the process, so the HTTP request never answered, so the browser's fetch never settled, so the client's own error path never ran either. A caller can recover from "not found". It cannot recover from silence.

### Three changes, smallest first

**The query is bounded.** `BLOB_QUERY_TIMEOUT` (5s) per attempt, treated exactly like an empty result: back off, retry, and if the attempts run out report the blob as undiscoverable. Generous next to a warm-cluster lookup of a few hundred ms, and far below kad's own query timeout — which is what previously decided how long a caller waited.

**A failed lookup is 404, not 500.** The final attempt returned `Err`, so a blob with no reachable provider answered `500 Internal server error` — indistinguishable, to a client, from the node falling over.

**Waiting is now optional.** `?lazy=true` reads locally, starts discovery in the background and answers `202`; the outcome arrives as `NodeEvent::Blob` (`BlobReady` / `BlobUnavailable`), routed by context so it reaches exactly the subscribers already watching that conversation.

### Why lazy is opt-in

This endpoint's answer to "fetch a blob from a peer" is the bytes, and merobox's `download_blob` step, mero-js and meroctl all read it that way — `blob-cross-node-transfer` asserts the sha256 of a cross-node fetch. Answering `202` by default would break all of them. What changed for existing callers is only that their wait is now **finite**.

Concurrent asks are deduped through an `Arc<DashSet<BlobId>>`: a conversation full of images asks for the same blob once per rendered message, and a virtualised list asks again on every scroll. One fetch serves them all.

### Verified on a live 2-node pair, not just in tests

| | result |
|---|---|
| `?lazy=true`, blob absent | **202 in 1.1ms** (`fetch_state=Started`) |
| background fetch | completed **570ms** later, `size=200046` |
| follow-up read | **200, 200046 bytes, 3ms** |
| 8 concurrent lazy requests | **1 `Started` + 7 `AlreadyFetching`** |
| synchronous, no provider | **404 in 3.6s** — previously an indefinite hang, then 500 |

### One thing deliberately cut

I had every fetcher re-announce the blob so a shared image wouldn't depend forever on one node. Removed in 55ba163: `announce_blob` is a kad `put_record` — one record per key, last writer wins — so fetchers contend for the key, and if the winner later evicts the blob, discovery fails *even though the origin still has it*. I watched a near-miss of that shape while verifying. Real multi-provider discovery wants `start_providing`/`get_providers`, which is a separate change needing a 3-node test.

### Checks

87 + 140 + 327 tests pass, clippy and `cargo check --workspace` clean, fmt via the pinned 1.88 toolchain. Round-trip tests pin the new event's shape — `NodeEvent` is `untagged`, so a variant that isn't distinguishable by shape silently deserialises as the WRONG one rather than erroring.

### DIFF
--- a/crates/node/primitives/src/client.rs
+++ b/crates/node/primitives/src/client.rs
@@ -9,12 +9,13 @@ use calimero_crypto::SharedKey;
 use calimero_governance_types::SignedNamespaceOp;
 use calimero_network_primitives::client::{is_no_peers_subscribed_error, NetworkClient};
 use calimero_network_primitives::config::GOSSIPSUB_MESH_N_LOW;
+use calimero_primitives::blobs::BlobId;
 use calimero_primitives::context::{Context, ContextId};
 use calimero_primitives::events::NodeEvent;
 use calimero_primitives::identity::{PrivateKey, PublicKey};
 use calimero_store::Store;
 use calimero_utils_actix::LazyRecipient;
-use dashmap::DashMap;
+use dashmap::{DashMap, DashSet};
 use eyre::{OptionExt, WrapErr};
 use futures_util::Stream;
 use libp2p::gossipsub::TopicHash;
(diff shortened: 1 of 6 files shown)


chefsale (approved):

xilosada:
yagni

Issue #3700, opened 2026-08-29
Encrypt every op payload; direct admission is the prerequisite


chefsale:
Every op payload should be encrypted. Root ops are not, and the reason is a
bootstrap circularity that **direct admission removes** — so these are one piece
of work, not two.

## What is in the clear today

`RootOp` is cleartext by design — *"visible to all namespace members"*. Group ops
carry a cleartext `group_id` for routing and encrypt the mutation. The problem is
that the transport cannot enforce the "members" half: `ns/<hex>` is a gossipsub
topic, subscription is ungated, and mesh scoring is soft-only. So *member-visible*
is in practice *anyone-visible*.

All eleven root ops, and what a subscriber learns:

| op | leaks |
|---|---|
| `GroupCreated` / `GroupReparented` / `GroupDeleted` | the group topology and every change to it, incl. cascade context ids |
| `AdminChanged` | who administers what, and every handover |
| `PolicyUpdated` | `policy_bytes` |
| `MemberJoined` / `MemberJoinedAt` | member account, join time, **the full signed invitation**, and the joiner's credential |
| `MemberJoinedOpen` / `NamespaceCreated` | member/founder account and credential |
| `MemberJoinedViaTeeAttestation` | member key, quote hash, **`mrtd`**, `tcb_status`, role |
| `KeyDelivery` | who received a key (the envelope itself is wrapped, so the key is safe) |

Every `JoinAccountCredential` is a root-signed device cert, so the
account-to-device-to-key graph is public too.

"Operators never see user data" is true of **application state** — that is
group-scoped and encrypted. It is not true of **membership metadata**.

## Why root ops are cleartext, and how that ends

Circularity: to join you must publish an op; to encrypt it you need the key; to
get the key you must be a member. Cleartext was the way out.

**Direct admission dissolves it.** If the joiner never publishes its own join op —
it presents a claim to an admitter named by the inviter, who publishes on its
behalf — then every publisher of a root op is already a member holding the
namespace key. Nothing left needs to be readable by a non-member.

The namespace key already exists and is already used for encryption elsewhere
(`group_governance_publisher.rs`), so this is not new machinery.

## Honest scope

Achievable: **all op payloads encrypted.** That is the content.

Not achievable without a much larger project, and worth stating so
"fully encrypted" is not read as more than it is:

- **The namespace id.** `ns/<hex>` is the routing topic. An observer always
  learns which namespace a message belongs to and which peers subscribe. Only
[...]

PR #3714, opened 2026-08-30
feat(invite): default admitters to the group's admins and TEE nodes

chefsale:
Stacked on #3703 — **merges into `feat/direct-admission`, not master.** Implements the defaulting decision from calimero-network/core#3700.

An invitation nobody restricted is claimable by broadcast, which staples it to a readiness beacon and publishes it as plain borsh on the namespace topic. Any subscriber can read and redeem it, silently — the consumed-invitation row is account-keyed, so a thief and the intended joiner both succeed. Restricting admission only helps if the unrestricted case stops being what you get by saying nothing.

A caller naming no admitters now gets the group's **current admins + every TEE-admitted account**. TEE nodes are included whatever their role: they are admitters because of what they run, not what they may do in the group.

### The snapshot, and why it's safe

The list is signed into the invitation, so it is a snapshot — an admin removed afterwards can still admit until expiry. That is acceptable *only* because #3703 capped expiry at `MAX_INVITATION_VALIDITY_SECS` (24h): the stale window is bounded by the invitation's life, not the group's.

Resolving live at admit time would close that window, at the cost of the property that makes this usable — a joiner that has synced nothing can read who to approach off the invitation alone.

### Behaviour change, and why it isn't in #3703

This changes every invitation. `beacon_admission_provable` refuses a restricted one, so scenarios that join via the readiness beacon take the admitter path instead. That is the intended end state and it is what makes the broadcast claim path deletable — but it needed to not ride in on a PR that was already green and verified.

**This wants the full scenario matrix before merge**, not just the standard checks.

### Verification

`default_admitters` has two tests: admins are named and a read-only member is not (a reader appearing there would hand admission authority to the whole group), and a group with no admin returns `Ok(vec![])` rather than an error. Both pass. Workspace call sites compile clean.

### DIFF
--- a/.github/actions/setup-merobox/action.yml
+++ b/.github/actions/setup-merobox/action.yml
@@ -53,7 +53,7 @@ runs:
         # Adding a field to an admin-api response therefore does not reach a
         # scenario until this pin moves. It is the same chain a `crates/client` fix
         # travels, for the same reason.
-        MEROBOX_VERSION="0.6.69"
+        MEROBOX_VERSION="0.6.70"
         TAG="v${MEROBOX_VERSION}"
 
         case "$(uname -m)" in
--- a/crates/context/src/handlers/create_group_invitation.rs
+++ b/crates/context/src/handlers/create_group_invitation.rs
@@ -43,6 +43,19 @@ impl Handler<CreateGroupInvitationRequest> for ContextManager {
                 "create group invitation",
             )?;
 
+            // An invitation nobody restricted is claimable by broadcast, which
+            // publishes it to every subscriber of the namespace topic. Callers
+            // that say nothing get the group's admins and TEE nodes rather than
+            // that, so the exposed case stops being the one you reach by
+            // omission.
+            let admitters = if admitters.is_empty() {
+                calimero_governance_store::NamespaceMembershipService::default_admitters(
+                    &datastore, &group_id,
+                )?
+            } else {
+                admitters
+            };
+
             let private_key = PrivateKey::from(node_sk);
 
             let mut rng = rand::thread_rng();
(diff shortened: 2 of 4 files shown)


chefsale:
## The 93 failures were one bug, in the release chain

`invalid invitation signature: Verification equation was not satisfied` (`membership.rs:350`).

merobox's bundled client-py pinned a core rev predating `admitters`. It deserialises an invitation and re-serialises it on the way to the node, so the field was dropped and the invitation stopped matching the inviter's signature over it. Not one of the 93 logs pointed at the client.

**Why #3703 was green with the same incompatibility already present:** `admitters` carries `skip_serializing_if = "Vec::is_empty"`. While the list was always empty the field never reached the wire, nothing could drop it, and borsh over an empty `Vec` matched at both ends. This PR is the first thing that populates it. A conditionally-serialised field proves nothing about compatibility until something fills it in.

## Merge order — three clicks and a re-run

1. **calimero-client-py#102** — repins core to `ca3d1f977`, adds the `admit_join` binding and `admitters` on `create_namespace_invitation`. Publishes **0.6.37**.
2. **merobox#390** — floor to `>=0.6.37` in *both* `pyproject.toml` and `requirements.txt` (they have drifted before). Publishes **0.6.70**.
3. Re-run this PR — the pin to 0.6.70 is already committed here.
4. Merge this.

The pin sits on this branch deliberately, not master: master's invitations still carry an empty list, so nothing there needs 0.6.70.

## Expect one red that is not a failure

If merobox#390's CI reports `No matching distribution found for calimero-client-py>=0.6.37`, that is PyPI's `/simple/` index lagging its JSON API. It is CDN-served per edge, so a successful install from a laptop says nothing about what a runner resolves. Re-run the failed jobs; do not weaken the floor.

Also: re-run this PR **in full**, not `--failed`. The job that downloads the merobox release may have passed against the old pin, and `--failed` would skip it and re-test the stale binary.

PR #1940, opened 2026-02-11
feat(sync): 013 - WASM Merge Callback for Custom Types

xilosada:
## Summary

Implements issue #1780: Add WASM callback interface for custom type merging during sync operations.

- Add `__calimero_merge_root_state` WASM export for merging root application state
- Add `__calimero_merge` WASM export for merging custom types by type name
- Implement `WasmMergeCallback` trait in runtime to invoke WASM merge exports
- Add `try_merge_by_type_name()` for string-based merge function lookup

## Test plan

- [x] Unit tests pass (`cargo test -p calimero-storage -p calimero-runtime`)
- [x] E2E test passes: `team-metrics-custom` workflow with concurrent CRDT operations
- [x] Verified custom `Mergeable` implementation is invoked via WASM callback
- [ ] CI passes

Implements #1780

---

> [!NOTE]
> **Medium Risk**
> Touches core sync/merge plumbing across runtime, macros, and storage, and introduces unsafe WASM memory/ABI handling (ptr/len packing, memory growth) that could cause merge failures or traps if contracts drift.
> 
> **Overview**
> Enables sync to CRDT-merge *root state* and *custom types* by calling new WASM exports (`__calimero_merge_root_state`, `__calimero_merge`) via a runtime-provided `RuntimeMergeCallback` implementing `WasmMergeCallback`.
> 
> Extends the storage merge registry to support **type-name-based lookup** (`try_merge_by_type_name`) and adds injectable/registry-backed callbacks (plus a no-op LWW fallback) to decouple merge behavior from global state; the SDK `#[app::state]` macro now emits the required WASM exports and a compact Borsh boundary result type for passing success/error payloads across the WASM boundary.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9a38615750eb7f48884e5c551977a92baad25f10. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

Issue #1780, opened 2026-02-01
[Sync Protocol] 013: WASM Merge Callback for Custom Types


xilosada:
# Issue 013: WASM Merge Callback for Custom Types

**Priority**: P1  
**CIP Section**: Appendix A - WASM Merge Callback Interface  
**Depends On**: 012-builtin-crdt-merge

## Summary

Implement WASM callback interface for merging custom `Mergeable` types defined by applications.

## When to Use

- `crdt_type == Custom { type_name }`
- Application defines `impl Mergeable for MyType`
- Cannot merge in storage layer alone

## Callback Interface

```rust
/// Trait for WASM merge callback
pub trait WasmMergeCallback: Send + Sync {
    /// Merge custom type via WASM
    fn merge(
        &self,
        local: &[u8],
        remote: &[u8],
        type_name: &str,
    ) -> Result<Vec<u8>, MergeError>;
    
    /// Merge root state (always custom)
    fn merge_root_state(
        &self,
        local: &[u8],
        remote: &[u8],
    ) -> Result<Vec<u8>, MergeError>;
}
```

## WASM Module Export

Applications must export merge functions:

```rust
// In application WASM
#[no_mangle]
pub extern "C" fn __calimero_merge(
    local_ptr: *const u8,
    local_len: usize,
    remote_ptr: *const u8,
    remote_len: usize,
    type_name_ptr: *const u8,
    type_name_len: usize,
) -> *mut MergeResult { ... }

#[no_mangle]
pub extern "C" fn __calimero_merge_root_state(
    local_ptr: *const u8,
    local_len: usize,
    remote_ptr: *const u8,
    remote_len: usize,
) -> *mut MergeResult { ... }
```

## Runtime Implementation

```rust
pub struct RuntimeMergeCallback {
    module: WasmModule,
}

impl WasmMergeCallback for RuntimeMergeCallback {
    fn merge(&self, local: &[u8], remote: &[u8], type_name: &str) -> Result<Vec<u8>> {
        // Call WASM export __calimero_merge
        self.module.call("__calimero_merge", local, remote, type_name)
    }
    
    fn merge_root_state(&self, local: &[u8], remote: &[u8]) -> Result<Vec<u8>> {
        // Call WASM export __calimero_merge_root_state
        self.module.call("__calimero_merge_root_state", local, remote)
    }
}

impl RuntimeMergeCallback {
    /// Create callback from loaded module (if exports exist)
[...]

PR #3743, opened 2026-08-31
fix(storage): stop persisting rustc's type_name output in CrdtType

### LINKED ISSUE #3670: CrdtType persists rustc's unstable type_name output into storage and the wire
## Summary

`std::any::type_name::<T>()` output is written into `CrdtType`, which is Borsh-serialized into the entity index and sent on the wire during sync. `type_name` makes **no stability guarantee** about its output, and it just changed under us: on #3666 (rustc 1.88 → 1.98) `Contained<u8>` became `Contained<'_, u8>` because 1.98 began rendering elided lifetimes.

Found while fixing two `calimero-primitives` tests that broke for the same reason.

## Where it happens

Not just one collection — every generic one, via `CrdtMeta::crdt_type()` and the re-key paths:

- `collections/crdt_impls.rs:140,288,357,416,483,532` — `lww_register`, `unordered_map`, `sorted_map`, `unordered_set`, `sorted_set`, `vector`
- `collections/sorted_map.rs:127,185,292`
- `collections/vector.rs:56,120,142`
- `collections/sorted_set.rs:95,141,166`
- `collections/unordered_set.rs:76`
- `collections/counter.rs:238,239`

These land in `CrdtType::{LwwRegister{inner_type}, UnorderedMap/SortedMap{key_type,value_type}, UnorderedSet/SortedSet/Vector{element_type}}` (`primitives/src/crdt.rs`), which derives `BorshSerialize`/`BorshDeserialize` and `Serialize`/`Deserialize`.

## Blast radius — narrower than it first looks

I traced every consumer. **There is no divergence path today**, for four independent reasons:

1. **Not in the Merkle hash.** `own_hash = Sha256(borsh(child))`, and `Element` marks `metadata` `#[borsh(skip)]` (`entities.rs:199-204`) — only `id` serializes. `full_hash = fold(own_hash, child_trie_root)`. So `crdt_type` never reaches a hash, and two nodes on different toolchains cannot fork on it.
2. **Not used for dispatch.** `merge_by_crdt_type` (`storage/src/merge.rs:372`) matches on the *variant* and discards the strings — `CrdtType::SortedMap { .. }`.
3. **Never compared across nodes.** The only string comparisons are against local sentinel constants: `OPAQUE_LEAF_CRDT_TYPE_NAME` (`sync/helpers.rs:361`) and `JS_ROOT_CRDT_TYPE_NAME` (`crdt.rs:215`), both on `LwwRegister.inner_type`.
4. **Not in id derivation.** `compute_collection_id(parent_id, field_name)` ignores `crdt_type`.

And `key_type` / `value_type` / `element_type` are **write-only** — constructed, persisted, transmitted, and never read anywhere in the tree.

So this is a **latent trap, not a live bug**: unstable compiler output is committed to a persisted and transmitted format, and happens to be harmless only because nothing consumes it. That is a property no one is currently enforcing, and it would stop being true the moment someone reasonably decides to compare or hash `crdt_type`.

## How to fix it

**A. Drop the write-only fields (recommended).** Remove `key_type`/`value_type`/`element_type` from the collection variants; nothing reads them. This deletes the whole problem class for those variants rather than papering over it. Cost: a breaking Borsh change to `CrdtType`, so it needs a format bump or migration — there is precedent (`entities.rs:888` records an earlier such break).

**B. Make `inner_type` a declared enum, not a string.** Its two real uses are sentinels, so the string is doing an enum's job. Something like `LwwRegister { kind: LwwKind }` with `LwwKind = { Opaque, JsRoot, App }` removes string comparison entirely and is self-documenting. Combined with A, no `type_name` output is persisted at all.

**C. If the strings must stay for debuggability, canonicalize them.** Route every `type_name` call through one normalizer that strips lifetimes and normalizes whitespace, so rustc rendering changes cannot leak into stored bytes. Cheapest option, but it only narrows the window — a future rustc could change path rendering too, and the field is still unstable-by-construction.

**D. Add a recorded-bytes regression test regardless.** Assert the exact serialized `CrdtType` for a representative set of `K`/`V` (including a type with a lifetime parameter). That turns the next silent drift into a red test instead of a surprise, and matches how refactors are proven elsewhere in this repo.

Recommended: **A + B + D**. A and B remove the unstable input; D guards against it coming back.

## Note

If A or B is adopted, check the JS/ABI surfaces at the same time. `wasm-abi` has its own `inner_type` (`schema.rs:247`), but it is a structured `TypeRef` rather than a rustc type name, so it looks independent — worth confirming rather than assuming.

rtb-12:
## The problem

Every CRDT collection carried a small label describing what Rust type it holds, filled from a debugging helper:

```rust
CrdtType::unordered_map(
    std::any::type_name::<K>(),   // "alloc::string::String"
    std::any::type_name::<V>(),   // "u64"
)
```

rustc gives no stability guarantee for how `type_name` renders a type. That would be harmless for a debug string, but these labels do not stay local. They are written into:

- **persisted rows** - `EntityIndex.metadata` and the child trie's buckets;
- **the sync wire** - offset 0 of `LeafMetadata`, plus every `Action`'s `Metadata` and each `ancestors[].metadata`;
- **a hash** - `CausalDelta::compute_id`'s preimage includes `borsh::to_vec(ancestors)`, and `content_address_matches` re-derives that id as a gossip receive gate.

So two nodes built with different compilers can produce different bytes for identical data, disagree on a delta id, and silently drop each other's updates. A compiler upgrade alone is enough, with no source change anywhere.

merobox cannot catch this: every node in an e2e run is the same build, so both sides always agree.

## What the labels were for

The original design (#1978) planned to hand types the storage layer could not merge to a WASM callback, and the label said which type that was. Only one case ever needed it - `Custom`, for app-defined types - and it got its own variant. Every other variant is merged by the storage layer matching on the variant and discarding the label:

```rust
CrdtType::UnorderedMap { .. } => merge_unordered_map(...)   // label thrown away
```

The labels have been write-only ever since.

## What this PR does

The four vestigial payloads are deleted, and `LwwRegister`'s free-form string becomes a declared enum:

```rust
// before
LwwRegister  { inner_type: String }   // "alloc::string::String" | "Opaque" | "JsRoot"
UnorderedMap { key_type: String, value_type: String }
SortedMap    { key_type: String, value_type: String }
UnorderedSet { element_type: String }
SortedSet    { element_type: String }
Vector       { element_type: String }

// after
LwwRegister  { kind: LwwKind }        // Value | Opaque | JsRoot
UnorderedMap, SortedMap, UnorderedSet, SortedSet, Vector
```

`"Opaque"` and `"JsRoot"` were the only strings anything read back, and they were compared by hand in three files. They are now `LwwKind` variants, so `hash_comparison_protocol.rs`, `level_sync.rs` and `sync/helpers.rs` call `is_opaque_leaf()` and the local `OPAQUE_LEAF_CRDT_TYPE_NAME` const is gone. `Custom(String)` is untouched - it is the one payload that is genuinely read.

## Why removing them is safe

Deleting fields changes the borsh layout of a struct that crosses the sync wire, which has no version negotiation. This is a **declared lockstep upgrade**, the same way `sync/wire.rs:82` and `sync/snapshot.rs:825` already handle theirs. Two properties make it safe rather than reckless:

**Old data still decodes.** `BorshDeserialize` is hand-written and accepts the legacy `0x00-0x0D` tags, reading and dropping their payloads. A row holding `"alloc::string::String"` maps to `LwwKind::Value`, `"Opaque"` to `Opaque`. No migration.

**A stale peer fails loudly.** Current tags are the legacy tag plus `0x80`:

```rust
let (variant, legacy) = match tag.checked_sub(CRDT_TYPE_TAG_V2) {
    Some(variant) => (variant, false),   // current format
    None          => (tag, true),        // legacy row
};
```

Without that shift a pre-upgrade binary would read `LwwRegister`, then consume the next four bytes as a `u32` string length; if those bytes are small it decodes "successfully" with every field behind it shifted. The tag change turns that silent misparse into an unknown-discriminant error.

A node can always read its own data: `content_address_matches` runs only on the gossip receive path, never when replaying stored deltas. The cross-version effect is confined to new nodes rejecting old nodes' deltas until the upgrade completes.

## Tests

A recorded byte string is the only thing that catches this class of drift, so:

- `borsh_encodings_are_byte_frozen` - every variant's encoding, both directions, including a populated `Custom` and all three `LwwKind`s.
- `legacy_type_name_payloads_still_decode` - hardcoded pre-change bytes for all fourteen legacy tags plus both sentinels.
- `current_tags_are_unknown_to_a_legacy_reader` - the two tag ranges cannot collide, and a tag from neither range is rejected.
- `leaf_metadata_encoding_is_byte_frozen` - one fully-populated `LeafMetadata`, since `CrdtType` sits at its offset 0.
- `test_borsh_discriminant_tags_are_stable` - repinned to the new base; it caught the tag move rather than being deleted for it.

Closes #3670.

## Test plan

```
$ cargo fmt --check
clean

$ cargo clippy -p calimero-primitives -p calimero-node-primitives -p calimero-storage \
    -p calimero-node -p calimero-context-client -p merodb \
    --all-targets --features calimero-storage/testing -- -D warnings
exit 0

$ cargo test -p calimero-primitives --features borsh crdt
test result: ok. 15 passed; 0 failed

$ cargo test -p calimero-node-primitives --lib
test result: ok. 328 passed; 0 failed; 1 ignored
```

The full workspace suite and merobox e2e run on CI for this branch.

### Observed failing first

Each frozen test was watched failing before being believed, by swapping the first two `CrdtType` variants so every discriminant behind them shifts; reverted afterwards.

Two pins then failed for real on the deliberate tag move and were repinned rather than removed: `test_borsh_discriminant_tags_are_stable` (`left: 128, right: 0`) and `leaf_metadata_encoding_is_byte_frozen`, whose first byte went from `04` plus two empty strings to a bare `84`.


xilosada:
@rtb-12 please share a report about this. 
Why the fields are not deleted: dropping them changes the borsh layout of a struct that crosses the sync wire with no version negotiation, so a new node and a not-yet-restarted old one would disagree on every frame, and every persisted row would need a migration. That is an ungated flag-day change, deferred until sync carries a wire version. Until then old data stays decodable, which legacy_type_name_payloads_still_decode pins by decoding hardcoded pre-change bytes.`

rtb-12:
Report below - and the short version first: **the paragraph you quoted is superseded.** The PR now deletes the fields, and I have rewritten the description to match. Two of the three reasons I gave for deferring did not survive scrutiny.

**"No version negotiation on the sync wire"** - accurate, but I treated it as a blocker when this repo already has an accepted answer for it. `sync/wire.rs:82` and `sync/snapshot.rs:825` both declare lockstep upgrades in prose. At `0.11.0-rc` that is a normal move here, not an exotic one, so "wait until sync carries a wire version" was the wrong bar.

**"Every persisted row would need a migration"** - wrong. `BorshDeserialize` is now hand-written and accepts the legacy `0x00-0x0D` tags, reading and discarding their payloads: a row holding `"alloc::string::String"` maps to `LwwKind::Value`, `"Opaque"` to `Opaque`. Existing rows decode unchanged. Only the writer changes.

**"A new node and an old one would disagree on every frame"** - true, but the framing hid the part that actually mattered. The danger was never disagreement, it was *how*. With the tags left alone, a pre-upgrade binary reads `LwwRegister`, then consumes the next four bytes as a `u32` string length; if those bytes are small it decodes "successfully" with every field behind it shifted. Silent corruption, no error. Current tags are therefore the legacy tag plus `0x80`, which turns that into an unknown-discriminant failure.

What made it safe to do now rather than defer: `content_address_matches` is called only on the gossip receive path (`handlers/state_delta/mod.rs`, two sites), never when a node replays its own stored deltas. So no node can lose its own data. The cross-version effect is confined to new nodes rejecting old nodes' deltas until the upgrade completes, which is the declared flag day.

Also folded in, since the string was the reason they existed: `"Opaque"` and `"JsRoot"` are no longer magic strings. `LwwRegister` carries a declared `LwwKind`, the sync layer calls `is_opaque_leaf()`, and `hash_comparison_protocol.rs`'s local `OPAQUE_LEAF_CRDT_TYPE_NAME` is deleted. `Custom(String)` is untouched - it is the one payload anything reads.

New coverage: `legacy_type_name_payloads_still_decode` now pins all fourteen legacy tags plus both sentinels, and `current_tags_are_unknown_to_a_legacy_reader` pins that the two tag ranges cannot collide. `test_borsh_discriminant_tags_are_stable` caught the deliberate tag move and was repinned to the new base rather than deleted.

Happy to split the removal back out into its own PR if you would rather review the byte-stability fix and the layout break separately.

chefsale:
Solid PR. The `0x80` tag shift, the legacy-tag decode path, and the frozen-byte tests are the right calls, and I like that each frozen test was watched failing before being believed rather than asserted into existence.

I checked the blast radius and it's confined to core: `wasm-abi`'s `crdt_type` is its own `CrdtCollectionType` (`crates/wasm-abi/src/schema.rs:297`), structurally declared and unrelated to `primitives::CrdtType`, and no out-of-repo consumer (`calimero-sdk-js`, `calimero-client-py`) touches it. `merodb` does, and it's updated here. So this is node-to-node only — no SDK → client-py → merobox release chain, which makes the lockstep break considerably cheaper than it first reads.

One thing I'd like settled before approving, plus two cleanups.

## 1. Should `Custom(String)` go too?

The description keeps it as "the one payload that is genuinely read". I don't think that survives a look at the tree:

- `CrdtType::Custom` is **never constructed in production**. Every hit is a test or a match arm.
- `merge.rs:421` binds `type_name` into `MergeError::WasmRequired` — whose only caller, `interface.rs:3882`, logs a warning and falls back to LWW.
- `tools/merodb/src/abi.rs:249` matches `Custom(_)` and discards the string.
- The SDK offers no way for an app to declare one.

The consumer it was the lookup key for never landed. #1780 / #1940 shipped the *root-state* half — `__calimero_merge_root_state` and `__calimero_register_merge` are live today — but the per-type by-name half is absent: zero hits for `try_merge_by_type_name` or a bare `"__calimero_merge"` export. And custom `Mergeable` structs ended up solved a different way entirely, by deterministic re-key (#2579, #2650), which converges them structurally so there's nothing left to hand to WASM.

So by this PR's own standard — delete payloads nothing consumes — `Custom`'s string qualifies alongside the five being removed.

The timing is the real argument rather than the purity: you're already paying a fleet-wide lockstep tag break here. Dropping `Custom` in the same change is nearly free; dropping it later costs a **second** coordinated upgrade. If someone does intend to revive by-name dispatch, keeping it is fine — but that intent should be written down somewhere, because right now nothing in the tree points at it.

## 2. Stale comment at `interface.rs:3882`

```rust
// Types that need WASM callback (LwwRegister, collections, Custom)
// For now, fall back to LWW. PR #1940 will add WASM callback support.
```

#1940 is closed. Since this PR is already editing that neighbourhood, worth repointing the comment at what actually happens rather than at a PR that isn't coming.

## 3. Not yours, but adjacent

"a struct that crosses the sync wire, which has no version negotiation" is more true than the description implies. `SYNC_PROTOCOL_VERSION` (`crates/node/primitives/src/sync/handshake.rs:16`) and `is_version_compatible()` exist, carrying the doc comment *"Increment on breaking changes to ensure nodes can detect incompatibility"* — and `SyncHandshake` has **zero production call sites**. The mechanism built for exactly this situation was never wired up, so bumping it would accomplish nothing. Noting it here only so the next person doesn't go looking for it; filing separately, no action needed in this PR.

chefsale (approved):
Good change, and well argued. The `0x80` tag shift is the part that matters most — turning a pre-upgrade binary's silent field-shift into an unknown-discriminant error is the difference between a lockstep break and a data-corruption incident — and the hand-written `BorshDeserialize` accepting the legacy `0x00-0x0D` tags means nothing on disk needs migrating. Good call watching each frozen test fail before believing it, and repinning the two that caught the deliberate tag move rather than deleting them.

I traced the blast radius and it's confined to core: `wasm-abi`'s `crdt_type` is its own `CrdtCollectionType` (`crates/wasm-abi/src/schema.rs:297`), structurally declared and unrelated to `primitives::CrdtType`, and no out-of-repo consumer (`calimero-sdk-js`, `calimero-client-py`) touches it. `merodb` does and is updated here. So there's no SDK → client-py → merobox release chain to sequence, which makes this considerably cheaper than a wire break usually is.

Two follow-ups, both filed separately so they don't hold this up:

- `CrdtType::Custom(String)` is kept here as the one payload that's genuinely read. It's read only into a `MergeError::WasmRequired` whose caller falls back to LWW, and the variant is never constructed in production — the per-type by-name dispatch it keys never landed (#1780 / #1940 shipped only the root-state half). Whether it should go depends on which direction #3785 takes, so it stays for now; the tag base is explicit after this PR, so removing it later is mechanical.
- `interface.rs:3882` still points at #1940 for WASM callback support. #1940 is closed; tracked in the same follow-up.

Nothing blocking. Approving.

PR #3745, opened 2026-08-31
chore(deps): move the wasmer family to 7.4

rtb-12:
## Summary

Moves the wasmer family from 6.1.0 to `=7.4.0` (`wasmer`, `wasmer-types`, `wasmer-middlewares`), superseding dependabot #3720.

**Why #3720 looked inconsistent.** It asked for `wasmer = "7.1.0"` beside `wasmer-types`/`wasmer-middlewares = "7.3.0"`. That requirement string never took effect: the family exact-pins itself in every direction (`wasmer 7.3.0` requires `wasmer-types "=7.3.0"`, `wasmer-middlewares 7.3.0` requires `wasmer "=7.3.0"`), so #3720's own lockfile already resolved 7.3.0. None of the suspected causes were real - no yank, no MSRV block, no dependency conflict. It was inert text advertising a floor the family forbids.

**Why exact pins.** A wasmer *minor* changed memory strategy, which is exactly what this PR has to absorb. `=` makes crossing one a deliberate review rather than something a routine `cargo update` does.

## The 7.4 change, and the knob that went with it

7.4 reduces `BaseTunables` to a unit struct and makes `MemoryStyle::Static` fieldless, reserving 8 GiB of virtual address space per static memory (4 GiB wasm32 space plus a 4 GiB guard, per upstream's own doc comment). Virtual address space, not committed memory - pages commit as the guest touches them.

The only field this workspace set was `static_memory_bound`, fed from `VMLimits::max_memory_pages`:

```rust
let base = BaseTunables {
    static_memory_bound: Pages(limits.max_memory_pages),
    ...
};
```

That was never a cap. Reading wasmer 6/7.3's own `memory_style`, it picks a strategy:

```rust
let maximum = memory.maximum.unwrap_or_else(Pages::max_value);
if maximum <= self.static_memory_bound {
    MemoryStyle::Static { bound: self.static_memory_bound, .. }   // pre-reserve, elide bounds checks
} else {
    MemoryStyle::Dynamic { .. }                                    // grow as needed
}
```

Nothing is rejected either way, and `WasmerTunables`' `create_host_memory` / `create_vm_memory` are bare passthroughs to `self.base`. So the default of 1024 pages was pushing any memory above 64 MiB onto the slower dynamic path, not limiting guests. Real bounds come from the module's declared `MemoryType.maximum` and from gas.

With the field gone there is nothing for `max_memory_pages` to feed, so it goes too, along with its rows in `HOST_FUNCTIONS.md`, `MEMORY_MODEL.md` and the protocol limits page. That page now says what actually bounds guest memory instead of implying a 64 MiB cap. `VMLimits` has no serde, so this is a Rust API change only - no config file or on-disk format is affected.

`WasmerTunables` keeps its `vmconfig`, which still carries `wasm_stack_size` from `VMLimits::max_stack_size`.

## Other break

`TrapCode` gained four variants (`UninitializedExnRef`, `YieldOutsideAsyncContext`, `HostInterrupt`, `ReadonlyTableModified`). All map to `Indeterminate` and the match stays wildcard-free, so a future variant is a compile error rather than a silent default.

Verified unchanged: the metering API **and** the `wasmer_metering_remaining_points` global name (a rename there would have disabled metering while still compiling), headless engines, `Module::serialize`/`deserialize`, and the `profiling` feature.

## Precompiled module caches

No upgrade hazard. `from_precompiled` and `Module::to_bytes` have no production callers - every reference outside their definitions is a test. The only module cache (`crates/context/src/lib.rs`) is an in-memory `BoundedCache` filled by `Engine::compile`, so an upgrading node has no persisted artifact to invalidate.

Covered anyway: a test feeds valid `wasmer-universal` magic with a foreign payload and asserts a typed `Deserialize` error plus that the engine still recompiles from source. The pre-existing all-zeros test never reached payload deserialization, so it could not have caught this.

## Test plan

```
$ cargo fmt --check
clean

$ cargo clippy -p calimero-runtime --all-targets -- -D warnings
exit 0

$ cargo check -p calimero-runtime --features profiling
exit 0

$ cargo test -p calimero-runtime
test result: ok. 153 passed; 0 failed
test result: ok. 26 passed; 0 failed      # builds and executes a real app wasm
test result: ok. 2 passed; 0 failed
test result: ok. 2 passed; 0 failed
```

Lockfile census: one version each of `wasmer`, `wasmer-types`, `wasmer-middlewares` and `wasmer-compiler-cranelift`, all 7.4.0.

The full workspace suite, the ABI contract and WASM size gates run on CI for this branch.

### Observed failing first

The precompiled-cache fallback test was watched failing before being believed, by breaking the fallback: `panicked at lib.rs: engine must still recompile from source after a rejected cache entry`. Reverted afterwards.


chefsale (approved):

Cargo.toml
```diff
@@ -298,9 +298,11 @@ trybuild = "1.0"
 ureq = "2.9.7"
 url = "2.5.2"
 velcro = "0.5.4"
-wasmer = "6.1.0"
-wasmer-types = "6.1.0"
-wasmer-middlewares = "6.1.0"
+# Pinned exact: the family exact-pins itself internally, and 7.4 drops the
+# BaseTunables memory knobs `WasmerTunables` needs (every memory becomes 8 GiB static).
+wasmer = "=7.3.0"
```
chefsale: why this one?

rtb-12: Fair question, and it changed the PR: it is now bumped to `=7.4.0`.

I had stopped at 7.3.0 because 7.4 empties `BaseTunables`, and the comment I left claimed that dropped a memory knob we needed. That was wrong, so I checked what the field actually did. In 6.x/7.3 `static_memory_bound` only picks a strategy:

```rust
let maximum = memory.maximum.unwrap_or_else(Pages::max_value);
if maximum <= self.static_memory_bound {
    MemoryStyle::Static { bound: self.static_memory_bound, .. }   // pre-reserve, elide bounds checks
} else {
    MemoryStyle::Dynamic { .. }                                    // grow as needed
}
```

Nothing is rejected either way, and our `create_host_memory`/`create_vm_memory` are bare passthroughs to `self.base`. So `max_memory_pages` was never a cap - at 1024 it was pushing any memory above 64 MiB onto the slower dynamic path. Real bounds come from the module's declared `MemoryType.maximum` and from gas. 7.4 making every static memory reserve 8 GiB of *virtual* address space is a simplification, not a lost control.

With `static_memory_bound` gone there is nothing for `max_memory_pages` to feed, so it goes too, along with its rows in `HOST_FUNCTIONS.md`, `MEMORY_MODEL.md` and the protocol limits page - that page implied a 64 MiB guest cap that never existed. `VMLimits` has no serde, so it is a Rust API change only.

The `=` pins stay, for the reason this exchange demonstrates: a wasmer minor changed memory strategy, so crossing one should be a deliberate review rather than a `cargo update`.

`clippy -D warnings`, `--features profiling` and `cargo test -p calimero-runtime` (153 + 26 + 2 + 2, zero failures, including the suite that builds and runs a real app wasm) all pass on 7.4.

Issue #3313, opened 2026-07-25
Secrets written to disk are owner-only on unix but inherit the directory ACL on Windows (signing key, node config, meroctl config)


rtb-12:
## Summary

Every place core writes a secret to disk restricts it to the owner on unix only. On Windows the file inherits the containing directory's ACL, so on a shared or loosely-permissioned host the secret can be readable by other local users. Windows is a supported target, so this is a real gap rather than a theoretical one.

Three sites, all with the same `#[cfg(unix)]` / `#[cfg(not(unix))]` shape and no ACL handling on the Windows branch:

- `crates/config/src/lib.rs:547` - node config (`config.toml`), which holds node identity material.
- `crates/meroctl/src/config.rs:67` - meroctl config, which holds session credentials.
- `tools/mero-sign/src/lib.rs` (`write_private_key`) - the unencrypted 32-byte Ed25519 signing seed, whose signer is part of `ApplicationId = SHA-256(borsh((package, signerId)))`.

There is no `windows-sys` / `windows` / `windows-acl` dependency in the workspace, so nothing sets a DACL anywhere today.

## Impact

- The signing key is the highest-value of the three: whoever reads it can publish bundles under the victim's publisher identity, and therefore under the same `ApplicationId`.
- The failure is silent: on unix the operator gets 0600, so the reasonable assumption is that the tooling protects the file everywhere. Only `mero-sign` currently warns; the two config writers do not.
- CI and developer laptops are the realistic exposure: keys and credentials generated in a shared build directory or a roaming profile inherit whatever that directory allows.

## Steps to reproduce

1. On Windows, create a directory that grants read access to `Users` (or use any inherited-permissive location).
2. Run `cargo mero key generate --output <that-dir>\key.json` (or `merod init`, or a `meroctl` login that persists config).
3. Inspect the resulting file with `icacls`.
   - **Actual:** the file inherits the directory ACL; other local users can read it.
   - **Expected:** the ACL grants only the creating user (plus SYSTEM/Administrators as appropriate).

## Criteria for resolving

- [ ] One shared helper writes owner-only secrets, used by all three call sites, rather than each re-implementing the `cfg(unix)` dance (they already differ: `meroctl` uses `create(true).truncate(true)`, which keeps an existing file's permissions; `mero-sign` uses remove-then-`create_new`, which does not).
[...]

PR #3747, opened 2026-09-01
ci: build and publish Windows binaries

chefsale:
## Why

`merod` has never been built for Windows in this repository. Releases carry Linux and macOS archives only, and two consumers depend on a Windows archive that does not exist:

- **The desktop app's in-app merod version switcher.** `merod_versions.rs` resolves release assets by target triple and asks for `x86_64-pc-windows-msvc`, so installing or switching merod versions cannot work on Windows at all today.
- **The desktop app's Windows CI job.** It worked around the gap by cloning this repository at the pinned tag and building merod from source on the runner. That step is what failed, and it is why Windows builds were disabled in `calimero-network/tauri-app#96`.

Core issue #3313 already treats Windows as a supported target and asks for a Windows CI job; this is the runner half of that.

## What was actually missing

Not portability. The source has been kept Windows-clean throughout:

- every `std::os::unix` use is `#[cfg(unix)]`-guarded (`config`, `meroctl`, `auth`, `merod/cli`),
- `mod watchdog` is `#[cfg(unix)]` and `--exit-on-eof` bails on non-unix by construction,
- both `shutdown_signal()` implementations (`node/run.rs`, `auth/server.rs`) have a `cfg(not(unix))` arm.

What was missing was a runner.

## What this does

Adds `x86_64-pc-windows-msvc` to the release build matrix and the artifact download in `release-binaries`.

Windows publishes **`merod` and `meroctl`**, not the full binary set. Those are the two with a consumer; `mero-abi`, `cargo-mero` and `mero-auth` have never been compiled for the platform, and a break in any of them would red the whole release for a binary nobody has asked for. The narrower list lives in a new `WINDOWS_BINARIES`, resolved once into `TARGET_BINARIES` so the build and compress steps cannot drift apart.

Three things the platform needs that unix does not:

| Need | Why |
| --- | --- |
| `LIBCLANG_PATH` | `rocksdb` is pulled with `bindgen-runtime`, so librocksdb-sys needs a libclang. Linux installs one; the Windows image ships LLVM but nothing points bindgen at it. |
| NASM | aws-lc-sys (rustls' provider, via libp2p-tls) assembles its x86_64 primitives with NASM on windows-msvc and fails without it. |
| `defaults.run.shell: bash` | The existing steps are bash (`readarray`, `tar`, `$BINARIES`). Windows runners default to pwsh. |

Only the file *inside* the archive gains the `.exe` Windows requires. The archive keeps the `<binary>_<target>.tar.gz` name every consumer already resolves by target triple, so adding a platform changes nothing about how any of them find an asset.

## Before merging

The matrix's `pull_request` arm carries a **`TEMPORARY`-marked** Windows entry, so this PR proves a Windows build green before it reaches master — no Windows build has ever run here. **That line must be removed before merge**, leaving Windows master-only, exactly as aarch64 and macOS already are.

## Follow-up

`calimero-network/tauri-app` re-enables its Windows job once a release from this change exists and `merod-config.json` is bumped to it. That branch is ready and holds on this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -208,13 +208,7 @@ jobs:
     strategy:
       matrix:
         # Keep legacy check names for required status checks.
-        #
-        # TEMPORARY (remove before merge): the pull_request arm normally stays
-        # Linux-only for cost, exactly as aarch64 and macOS are excluded. The
-        # Windows entry is duplicated into it here only so THIS pull request
-        # proves a Windows build green before it reaches master, since no
-        # Windows build has ever run in this repository.
-        include: ${{ fromJSON(github.event_name == 'pull_request' && '[{"name":"ubuntu-24.04-8cpu, x86_64-unknown-linux-gnu","runner":"big-machine","target":"x86_64-unknown-linux-gnu"},{"name":"windows-latest, x86_64-pc-windows-msvc","runner":"windows-latest","target":"x86_64-pc-windows-msvc"}]' || '[{"name":"ubuntu-24.04-8cpu, x86_64-unknown-linux-gnu","runner":"big-machine","target":"x86_64-unknown-linux-gnu"},{"name":"ubuntu-24.04-arm, aarch64-unknown-linux-gnu","runner":"ubuntu-24.04-arm","target":"aarch64-unknown-linux-gnu"},{"name":"macos-latest, aarch64-apple-darwin","runner":"macos-latest","target":"aarch64-apple-darwin"},{"name":"windows-latest, x86_64-pc-windows-msvc","runner":"windows-latest","target":"x86_64-pc-windows-msvc"}]') }}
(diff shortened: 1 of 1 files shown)


chefsale:
## Windows build verified green ✅

The temporary `pull_request` matrix entry did its job and has been removed in fc940c37f — Windows is now master-only, exactly as aarch64 and macOS already are.

**Run [33473825711](https://github.com/calimero-network/core/actions/runs/33473825711)** — first Windows build ever in this repository:

| Job | Result |
| --- | --- |
| `Build Binaries (windows-latest, x86_64-pc-windows-msvc)` | ✅ success |
| `Build Binaries (ubuntu-24.04-8cpu, x86_64-unknown-linux-gnu)` | ✅ success |

Linux passing matters as much as Windows here: it is what confirms the `TARGET_BINARIES` refactor of the shared build/compress steps left the unix path alone.

Artifact `artifacts-x86_64-pc-windows-msvc` (20,663,722 bytes), downloaded and inspected:

```
meroctl_x86_64-pc-windows-msvc.tar.gz  ->  meroctl.exe
merod_x86_64-pc-windows-msvc.tar.gz    ->  merod.exe
$ file merod.exe
merod.exe: PE32+ executable (console) x86-64, for MS Windows
```

That is exactly the shape both downstream consumers already look for — `prepare-merod.mjs` scores triple-match + `.tar.gz` highest, and `merod_versions.rs::score_merod_asset` resolves the same triple — so neither needs a change to start finding it.

Once this merges and a release carries the asset, `calimero-network/tauri-app` can bump `merod-config.json` and re-enable its Windows job; that branch is ready and holds on this.

PR #3752, opened 2026-09-01
test(storage-cost): gate every shipped collection, and make the tolerance check two-sided

xilosada:
**Stack 3 of 5.** Base: #3751. Independent of #3751 in content — it touches only `tools/storage-cost/` and could be rebased onto #3750 directly if you would rather land it first.

Widens the blocking cost snapshot from 24 rows to 48: adds `rga`, `lww_register` and `nested_map` alongside the three collections #3660 landed with.

## Two findings came out of doing it

**1. `RGA::insert(pos, char)` is O(n) per call**, so building a document one character at a time is quadratic — `n + ~47` storage rows per character, measured 63.5 / 147.7 / 547.1 / 2,047.0 at n = 10 / 100 / 500 / 2,000. No existing `CostShape` could describe that honestly: `FlatPerEntry` fails for the wrong reason and `KnownLinearInN` describes a point read, not a build. Adds `QuadraticBuild`, with a ratchet that fails if the curve worsens **and**, with a different message, if it silently stops being quadratic.

**2. The tolerance check only failed one way.** `reproducible.rs` failed a tolerance that was too *tight* — a flake generator — while nothing failed one that was gratuitously wide. Three separate comments claimed it caught both. Live consequence: `vector_get_nth` declared 25% against a measured spread of 10.5% at n=10 and under 3% above, so its 13,328 committed rows admitted anything to ~16,660 and a real 20% regression in `Vector::get` passed silently. The check is now two-sided and the band is 18%, verified by breaking a row and watching the gate catch what the old one let through.

## Deliberately not registered

`sorted_map`. Without `RuntimeEnv::with_index` wired, `index_bridge()` returns `None` and every `storage_index_*` call falls through to a thread-local mock that never touches the counting closures — so measuring it would attribute **zero** cost to all index maintenance while printing plausible numbers. The reasoning lives in the registry, not only in a review comment.

## Verification

`flat_curve` and `reproducible` pass; snapshot regenerated to 48 rows; `./scripts/check-storage-cost.sh` green across repeated runs; break-and-revert proof performed against a **new** row (`nested_map_insert`), confirming the widened gate bites on the rows this PR adds and names the workload at all four sizes.

## Note for the reviewer

`reproducible.rs` and `check-storage-cost.sh` arrive byte-identical from #3660; the two-sided rule changes someone else's work. The rule itself (3× worst observed spread + 8 points) is empirically tuned across 11 live measurements rather than derived from first principles — if you want something stricter, this is the line to argue about.

### DIFF
--- a/tools/storage-cost/src/lib.rs
+++ b/tools/storage-cost/src/lib.rs
@@ -190,10 +190,23 @@ mod tests {
     /// the property the entire snapshot gate rests on: if per-process
     /// thread-local state leaks between measurements, the snapshot is unstable
     /// and the gate flakes.
+    ///
+    /// The collection is built with `new_with_field_name`, NOT `new`. `new`
+    /// mints a RANDOM entity id, and a parent's children live in a hash trie
+    /// (core#3633) whose descent depth follows the hash of that id — so two
+    /// runs of an identical workload read a different number of trie nodes
+    /// while writing exactly the same set. That is why `rows_written` was
+    /// stable at 313 and only `rows_read` moved, and why this failed under
+    /// `calimero-storage/testing` (which CI enables) roughly always rather than
+    /// occasionally. The sibling `byte_counts_are_not_reproducible` already
+    /// records that ids are not deterministic; this test simply must not
+    /// depend on them being so.
     #[test]
     fn row_counts_are_deterministic_across_calls() {
         let workload = || {
-            let mut map = Root::new(UnorderedMap::<String, String, MainStorage>::new);
+            let mut map = Root::new(|| {
+                UnorderedMap::<String, String, MainStorage>::new_with_field_name("determinism")
+            });
             for i in 0..16 {
                 map.insert(format!("k{i}"), "v".to_owned())
                     .expect("insert should succeed");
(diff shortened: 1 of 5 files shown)


xilosada:
Pushed `56a7ead2c` — fixes the `Rust` failure on this PR, and the same failure on #3750 and #3754 once they pick it up.

**The bug.** `row_counts_are_deterministic_across_calls` built its workload with `UnorderedMap::new()`, which mints a **random entity id**. A parent's children live in a hash trie (core#3633) whose descent depth follows that id's hash, so two runs of an identical workload read a different number of trie nodes while writing exactly the same set.

That accounts for both clues in the CI output: `rows_written` pinned at 313 while only `rows_read` moved (1088 vs 1098), and the failure being ~deterministic under `calimero-storage/testing` — which CI's `cargo test` enables — rather than an occasional flake. The sibling `byte_counts_are_not_reproducible` already records that ids are not deterministic; this test just should not have depended on them.

**The fix** is `new_with_field_name`, giving the collection a deterministic id, which is what the property under test actually needs.

**Measured on this branch**, same command, with and without the change:

```
without fix:  0 pass /  6
with fix:    10 pass / 10
```

plus 10/10 under default features and `./scripts/check-storage-cost.sh` green on all rows — the change touches only the test's own workload, never a measured row.

Two things I got wrong on the way, in case they are in anyone's notes: this is **not** stale-base (the tool is byte-identical between this branch and a master-merged one, and it reproduces on both), and it is **not** flaky — it fails essentially always under CI's feature set.

chefsale (approved):

PR #3754, opened 2026-09-01
test(runtime): probe where storage operations stop working entirely

xilosada:
**Stack 5 of 5.** Base: #3753. This is the one with a product finding in it.

Two gas-wall probes in the shape of the existing `chat_wall`: expensive sweeps `#[ignore]`d, each with a cheap non-ignored discriminator so an ignored probe cannot rot unnoticed.

## Finding 1 — a per-call write ceiling of ~450 entries, on every collection

`bulk_write_wall` answers what the cost gate cannot: how much can **one call** write before it traps? Binary-searched to a real `GasExhausted`:

| One call | Lands | Exhausts | Gas at largest success |
|---|---:|---:|---:|
| `UnorderedMap::insert` | 447 | 448 | 999,642,805 |
| `UnorderedSet::insert` | 458 | 459 | 999,052,999 |
| `Vector::push` | 459 | 460 | 997,234,828 |
| `RGA::insert_str` | 491 | 492 | 998,651,983 |

Within 9% of each other. **This is a platform-wide ceiling, not any collection's defect** — we initially attributed the 492 to RGA and were wrong.

The constant behind it: **~33,600 gas per storage row**, against `DEFAULT_MAX_GAS` of 1,000,000,000, giving **~29,700 rows per call**. Divide by a collection's rows-per-entry (46–98, per the snapshot in #3752) and you have its batch limit. That number is what an app developer needs to size a contact import, a document seed, a state restore or a bulk migration — and it appears nowhere in the docs, the SDK, or any error message.

## Finding 2 — RGA additionally has a total ceiling

`rga_wall` measures what *is* specific to RGA: its cost grows with its own length, because both the read path and the write path run through `get_ordered_chars()`, which linearises the whole document. So it has a total ceiling as well as a per-call one, near **12,000 characters** — and the typing wall and the read wall are coincident, so a document does not degrade into read-only first. It becomes unopenable, permanently, while still on disk and still replicating.

Note this is *not* what the child trie (#3633) fixed. The trie removed the sibling-fold cost on writes — and it works, visibly: `unordered_map_insert` is flat at 66 rows/entry from n=10 to n=10,000. A trie makes looking up **one** child cheap; it cannot make "give me all n children, in order" cheaper than n. That is semantics, not layout.

## Suggested follow-ups (not in this PR)

1. Document the per-call row budget.
2. Make the ceiling fail politely — a batch over budget should be refused with a message naming the limit, not trapped with `GasExhausted`, which to an app is indistinguishable from the node falling over.
3. Give RGA an ordered index; it fixes the read cost and the write growth together.

## Status of the numbers

The four per-call walls are **executed**. The two RGA document walls are currently **extrapolated** from an executed 8,400 characters at 70% of the ceiling; a run to real exhaustion is in progress and I will update this PR with the executed figures.


xilosada:
## Update: the document walls are now executed

The two RGA figures in the description were extrapolated from 8,400 characters. The probe has since been run to real gas exhaustion:

| Wall | Result |
|---|---|
| Typing, one char per call | **12,544 characters land; 12,545 exhausts** at `limit=1,000,000,000` |
| Reading, `get_text()` | bracketed to **(12,500, 12,545]** |

The extrapolation said ~12,190, so it was low by 2.9% — direction and magnitude both hold, and the read and write walls are confirmed coincident rather than merely predicted to be.

Every number in this PR is now executed rather than modelled.

xilosada:
Pushed the `storage-cost` determinism fix — this is the `Rust` failure on this PR.

`row_counts_are_deterministic_across_calls` built its workload with `UnorderedMap::new()`, which mints a **random entity id**. A parent's children live in a hash trie (core#3633) whose descent depth follows that id's hash, so two runs of an identical workload read a different number of trie nodes while writing exactly the same set. That is the `rows_written: 313` / `rows_read: 1088 vs 1098` signature in the CI log. `new_with_field_name` gives the collection a deterministic id, which is what the property under test actually needs.

Measured on **this branch**, same command:

```
without fix:  0 pass / 6
with fix:     8 pass / 8
```

Applied as a cherry-pick rather than a rebase: the test exists on every branch in the stack including `feat/storage-cost-gates`, so the fix's natural home is the bottom — but cherry-picking onto each affected branch keeps your stack ordering untouched. When the stack is next rebased, git will drop the duplicates by patch-id. Same commit is on #3752.

Not stale-base and not flaky, in case either is in your notes: the tool is byte-identical across the stack, and it fails essentially always under CI's feature set.

chefsale (approved):

PR #3757, opened 2026-09-01
feat(merod): a graceful stop that works on Windows

chefsale:
## The problem

Windows has no `SIGTERM`, and `taskkill` without `/F` posts `WM_CLOSE` to windows a console process does not have. So a supervisor's only way to stop merod there is `taskkill /F` — `TerminateProcess`, which runs no code.

The consequence is concrete: `StopCause::flushes` never runs, and the desktop app hard-kills a node mid-write on **every quit**. `kill_pids` in tauri-app already notes merod needs ~10s to flush and that hard-killing it was a bug on unix; on Windows it is the only behaviour available today.

`--exit-on-eof` cannot cover this. It takes a `RawFd` and watches it with `tokio::net::unix::pipe`, neither of which exists off unix — `run.rs` already bails with *"--exit-on-eof is only supported on unix"*.

## Why stdin rather than a console control event

`GenerateConsoleCtrlEvent` is the usual answer, and it addresses a process group **sharing the caller's console**. A GUI supervisor has no console, so it would have to allocate or attach one purely to deliver the event — and `CREATE_NEW_PROCESS_GROUP` disables `CTRL_C` for the child, so the payload has to become `CTRL_BREAK` too. That is a lot of platform machinery for one signal, and it fights `CREATE_NO_WINDOW`, which the desktop app needs so merod does not sit behind a black console window.

Closing a pipe needs no console, no process group, and no platform code. It also carries both meanings on one channel:

- the supervisor **closes stdin** to ask for a graceful stop, and
- the OS **closes it anyway** if the supervisor dies

which is the same guarantee `--exit-on-eof` gives on unix, spelled portably.

## Shape

`--exit-on-stdin-close`, off by default — merod inherits a terminal's stdin when run by hand, and a node must not exit because someone piped it from a finished command.

Two details worth keeping:

- **Read on a plain OS thread, not `spawn_blocking`.** A thread parked in `read` cannot be cancelled and runtime shutdown waits for the blocking pool, so a node whose supervisor outlives it would hang on the way out. `parent_closed` documents the same trap. A detached thread does not delay process exit.
- **`parent_closed` and `data_dir_replaced` stay unix-only** — one needs an inherited fd, the other identifies a directory by `(dev, ino)`. Off unix the stdin watch is the whole watchdog, so `mod watchdog` is no longer `cfg(unix)`.

## Verification

`the_node_stops_when_its_stdin_closes` drives a real node and is deliberately **not** `cfg(unix)`: restricting it would exercise everything except the case it exists for.

```
node exited 100.397125ms after its stdin closed
test the_node_stops_when_its_stdin_closes ... ok
```

The unit test covers the dangerous direction for a default-off watch — that it never resolves when disabled, since a watch that fires spontaneously stops a node nobody asked to stop.

`cargo test -p merod --bin merod`: 79 passed. CI's `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`: clean.

## Before merging

The matrix's `pull_request` arm carries a **`TEMPORARY`-marked** Windows entry. This PR's entire subject is behaviour under `cfg(not(unix))`, which is never type-checked when building for a unix host — without it, the Windows arm of the new stop watch would first compile on master. **Remove that line before merge.**

## Downstream

`calimero-network/tauri-app` then spawns merod with piped stdin and this flag, and closes the pipe to stop it — replacing a `taskkill` graceful phase that cannot do anything to a console process.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -208,13 +208,7 @@ jobs:
     strategy:
       matrix:
         # Keep legacy check names for required status checks.
-        #
-        # TEMPORARY (remove before merge): this PR's whole subject is behaviour
-        # under `cfg(not(unix))`, which is never type-checked when building for
-        # a unix host. The pull_request arm is Linux-only for cost, so without
-        # this the Windows arm of the new stop watch would first compile on
-        # master. Remove once this PR is green.
-        include: ${{ fromJSON(github.event_name == 'pull_request' && '[{"name":"ubuntu-24.04-8cpu, x86_64-unknown-linux-gnu","runner":"big-machine","target":"x86_64-unknown-linux-gnu"},{"name":"windows-latest, x86_64-pc-windows-msvc","runner":"windows-latest","target":"x86_64-pc-windows-msvc"}]' || '[{"name":"ubuntu-24.04-8cpu, x86_64-unknown-linux-gnu","runner":"big-machine","target":"x86_64-unknown-linux-gnu"},{"name":"ubuntu-24.04-arm, aarch64-unknown-linux-gnu","runner":"ubuntu-24.04-arm","target":"aarch64-unknown-linux-gnu"},{"name":"macos-latest, aarch64-apple-darwin","runner":"macos-latest","target":"aarch64-apple-darwin"},{"name":"windows-latest, x86_64-pc-windows-msvc","runner":"windows-latest","target":"x86_64-pc-windows-msvc"}]') }}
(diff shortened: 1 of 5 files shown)


chefsale:
## Windows verified, temporary matrix entry removed

`Build Binaries (windows-latest, x86_64-pc-windows-msvc)` passed, which is the only place the new `cfg(not(unix))` stop watch is ever type-checked — building for a unix host never compiles that arm. Its purpose served, the `TEMPORARY` line is gone and Windows is back to master-only, as aarch64 and macOS already are.

Every check on this PR is now green.

### Verification recap

| What | Result |
| --- | --- |
| `the_node_stops_when_its_stdin_closes` (real node) | `node exited 100.397125ms after its stdin closed` |
| `cargo test -p merod --bin merod` | 79 passed |
| `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` | clean |
| Windows compile of the `cfg(not(unix))` arm | ✅ |

The live test is deliberately not `cfg(unix)`: gating it would exercise everything except the case it exists for.

### Downstream

`calimero-network/tauri-app` then spawns merod with piped stdin and this flag and closes the pipe to stop it, replacing a `taskkill` graceful phase that cannot reach a console process with no window.

Worth noting for that side: the desktop can run **any** merod through `VersionId`, not just the bundled one, so it will probe `merod run --help` for the flag rather than gate on a version number — an older pinned merod would otherwise fail to start on an unknown argument.

PR #3773, opened 2026-09-01
fix(deps): pin wasmer to 6.1.0, the newest release that supports Windows

chefsale:
Reverts #3745 (`b7a0e783c`) and records why the pin exists. **Master's Windows build is broken right now.**

## The bisect

| run | commit | Windows |
| --- | --- | --- |
| 33534721084 | `b7a0e783c` — #3745, wasmer 6.1 → 7.4 | **failure** |
| 33526003495 | `c0c3ac89b` | success |

## wasmer 7 removed *every* compiler backend on Windows

Not just Cranelift. All three carry the identical guard, and it is present from **7.0.0** onward — the whole 7.x line, not one bad release:

```rust
// wasmer-compiler-cranelift / -singlepass / -llvm, src/lib.rs
#[cfg(target_os = "windows")]
compile_error!("The <X> compiler backend is not supported on Windows. Use the V8 backend instead.");
```

I checked 7.0.0, 7.1.0, 7.2.0, 7.3.0 and 7.4.0 — all blocked. And there is no 6.1.1 or 6.2.0, so **6.1.0 is the newest wasmer that supports Windows at all.** The pin is the exact boundary, not a conservative guess.

## Why we cannot take wasmer's advice

V8 is a *runtime* backend, not a compiler backend. There is no `impl CompilerConfig` for it anywhere in the crate, so there is no `push_middleware` — and `wasmer_middlewares::Metering` is how a guest is stopped from running forever.

An unmetered Windows node would accept executions every other node rejects. That is a divergence in what nodes agree on, not a performance trade-off. On 7.x the choice is Windows **or** gas metering, and we need both.

LLVM was worth checking specifically, since it is where wasmer is steering and it *is* a sys compiler that keeps metering. It carries the same guard — and it would have been expensive to find out the hard way: `llvm-sys 221.1.0` means **LLVM 22.1**, required via `LLVM_SYS_221_PREFIX` on every machine and runner that builds merod.

## Impact if left

The next release ships **no `merod.exe`**, and the desktop's Windows build fails on its next merod bump — undoing what v0.0.88 just delivered.

## #3745 was not careless

It absorbed 7.4's `BaseTunables` change and removed a `static_memory_bound` knob that was never a cap. Nothing in its reasoning was wrong. It went unnoticed because **the release matrix is Linux-only on pull requests**, so no Windows build ran before merge. #3771 adds the lane that catches this class of breakage — it landed one PR too late.

## Two notes, so this does not recur

**In the workspace `Cargo.toml`**, above the pin — where whoever bumps wasmer next will actually be looking. It states what 7.x removed, why V8 is not an option, and what would have to change for 7.x to become viable.

**In `.github/dependabot.yml`**, an ignore for `version-update:semver-major` on the three wasmer crates. Deliberately majors only — minor and patch updates inside 6.x still arrive normally. Without it dependabot re-proposes 7.x on its next weekly run and the comment is advisory at best.

## Verification

- `wasmer-compiler-cranelift 6.1.0` carries no Windows guard, where 7.0.0 through 7.4.0 all do.
- `cargo test -p calimero-runtime`: 182 pass.
- `cargo fmt --check --all`: clean.

## Revisit when

wasmer restores a compiler backend on Windows, or metering moves off the middleware. Merging #3771 first means whichever way that goes, CI proves it on Windows before it merges.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,22 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 10
+    ignore:
+      # wasmer 7 removed every compiler backend on Windows (cranelift,
+      # singlepass and llvm all `compile_error!` there from 7.0.0), and its
+      # suggested V8 backend has no `CompilerConfig`, so gas metering cannot be
+      # installed. A 7.x bump therefore drops `merod.exe` from the release --
+      # see the pin note in the workspace Cargo.toml.
+      #
+      # Ignoring the majors, not the family: minor and patch updates within 6.x
+      # still arrive normally. Drop this once wasmer has a Windows compiler
+      # backend again.
+      - dependency-name: "wasmer"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "wasmer-types"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "wasmer-middlewares"
+        update-types: ["version-update:semver-major"]
     groups:
       # Families that version in lockstep, matched first so no member is ever bumped alone:
       # each appears in the others' public API, so a lone bump lands two incompatible types.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -208,13 +208,7 @@ jobs:
     strategy:
       matrix:
         # Keep legacy check names for required status checks.
-        #
-        # TEMPORARY (remove before merge): this PR exists to unbreak the Windows
-        # build, and the pull_request arm is Linux-only for cost -- so without
-        # this it cannot demonstrate the one thing it claims. #3771 adds a
-        # permanent Windows lane; until that merges, this is how a Windows fix
-        # proves itself.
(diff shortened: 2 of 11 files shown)


chefsale:
## Windows build verified on wasmer 6.1.0

`Build Binaries (windows-latest, x86_64-pc-windows-msvc)` **passed**, so the pin does restore `merod.exe`. The temporary matrix entry has served its purpose and is removed in 5b57730a7; Windows is back to master-only.

That entry existed because a PR whose entire subject is the Windows build could not otherwise demonstrate it — the `pull_request` arm is Linux-only for cost. Which is exactly how #3745 landed unseen. #3771 makes that lane permanent so the workaround stops being necessary.

### The version boundary is exact

I probed each release rather than assuming 6.1.0 was simply "the previous one":

| version | Windows |
| --- | --- |
| 7.4.0, 7.3.0, 7.2.0, 7.1.0, 7.0.0 | blocked — `compile_error!` in cranelift, singlepass **and** llvm |
| 6.1.1, 6.2.0 | do not exist |
| **6.1.0** | **supported** |

So 6.1.0 is not a conservative fallback; it is the newest wasmer that supports Windows at all.

### Merge order

**This first, then #3771.** #3771's branch predates #3745, so its green Windows lane ran against a master that still had wasmer 6.1. Merging it onto master as it stands today would install a permanently-red Windows job, because merod cannot build there until this lands.

PR #3774, opened 2026-09-01
feat(account): say whether this node holds the root of the account it speaks for

rtb-12:
## Summary

A caller cannot tell the account holder from a device paired into it. Both report the same `accountId` and the same `accountRootPublicKey`, because a paired node adopts the account it was certified into. The only way to learn that a node cannot certify another device was to attempt `pair-complete` and be refused with `PairingNotTheAccountHolder`.

`GET /admin-api/identity` now reports `holdsAccountRoot`.

```rust
/// Whether this node holds the root key of the account it speaks for. False
/// means no root key available: the node runs on a delegate device key, so it
/// cannot certify another device into the account.
#[serde(default)]
pub holds_account_root: bool,
```

It is read where the identity is already assembled, so it adds no failure path of its own. The two branches of that read are the two answers: the device row names the account this node adopted, and the fallback reports the node's own root, which is the account it holds the root of by construction.

```rust
if let Some(held) = devices.unrevoked_device()? {
    let holds_root = devices
        .account_root()?
        .is_some_and(|root| root.account() == held.account);
    ...
}
Ok(devices.account_root()?.map(|root| (root.account(), ..., true)))
```

**The account has to match rather than a root merely existing.** A node that ran `merod init` holds a root and can still be paired into another account, and certifying into that one is not its to do. That combination is not hypothetical: `a_node_whose_own_device_was_revoked_falls_back_to_its_own_account_root` already builds it, and a bare `is_some` reports such a node as the holder.

Defaulted on deserialization so a response from a node predating the field still parses, which is what `crates/client`'s identity test pins.

The caller this is for is the desktop's account panel, which offered "Add a device" on every node and so offered a button a paired device's node would always refuse.

## Test plan

- `cargo test -p calimero-server --lib get_node_identity`: 7 passed.
- Three new cases: the holder reports `true`; a node holding its own root but paired elsewhere reports `false` for the adopted account; a node with no root reports `false`.
- Replacing the condition with a bare `is_some()` fails `a_paired_node_does_not_hold_the_root_of_the_account_it_adopted` and nothing else, confirming the account match is what the test pins.
- `cargo test -p calimero-client --lib a_caller_finds_itself_by_matching_its_account`: 1 passed, covering an identity response that omits the field.
- `cargo fmt --all --check` clean.


crates/server/src/admin/handlers/identity/get_node_identity.rs
```diff
@@ -89,6 +99,18 @@ pub async fn handler(Extension(state): Extension<Arc<AdminState>>) -> impl IntoR
         }
         .into_response();
     };
+    let holds_account_root = match holds_account_root(store, account) {
+        Ok(holds) => holds,
+        Err(err) => {
+            error!(error = ?err, "Failed to read this node's account root");
```
chefsale: be more specific, failed to read can mean anything.

rtb-12: Fixed in 9d13609. It now names the record and what its absence leaves unknown:

```
failed to read this node's account root record, so whether it may certify
another device into its account is unknown
```

The log line says `Failed to read this node's account root record` rather than just "account root", so the two read as the same failure.

chefsale: Maybe something along the lines: this node holds only device key so the account root is not available? as i still dont get the reason of failure :D

rtb-12: Yes updating only , I also dont feel this right . Claude just pushed it and replied

chefsale (approved):

PR #3828, opened 2026-09-03
feat(server,meroctl,client): make delegated execution reachable, grantable and discoverable

chefsale:
## Description

Delegated authorship shipped complete on the wire — a warrant, two signature domains, a nonce ledger, a capability enforced at the API and again at the cut. It was unusable for the callers it exists for, and none of the reasons were protocol reasons.

**Nothing could grant the capability.** `CAN_AUTHOR_ON_BEHALF` is implied by neither membership nor admin and is not propagated by the subgroup cascade, so every group is authorship-closed by default. But `meroctl group members set-capabilities` had no flag for it — the feature could be enforced and never enabled. `check-access` also omitted the row, so an operator could not even see it. (The mask is *replaced*, not merged; re-pass any flag the member already holds.)

**The client could not learn what to sign.** Every input to a warrant is something the author already holds except `executor`, which names the *relay's* account — a content address no client can derive — and whether that relay may act in this context, a capability row no client can read. New `GET /admin-api/contexts/:context_id/intents` answers both, on the path the intent will be presented to:

```json
{ "data": { "executorAccount": "…", "canAuthorOnBehalf": true, "groupId": "…" } }
```

`canAuthorOnBehalf: false` comes back as an *answer*, not an error, because it is the default state of every context: a client reads it to say "ask an admin of `groupId`" instead of presenting a warrant that will be refused. Asking before signing is the point — minting a warrant spends a nonce from a monotonic per-device sequence, and one naming the wrong executor is unspendable.

**And the endpoint was closed to exactly the callers it exists for.** `POST .../intents` sits on the guarded admin API. For a member using their own node that is right and free. For a browser tab or an agent holding one signing key it is fatal: they have no relationship with the relay and so no credential on it — the [known gap](docs/src/content/docs/protocol/direct-admission.mdx) the direct-admission docs record. So `server.admin.public_intents` (`merod init --public-intents`) serves those two routes, and nothing else, without a node credential — because they are the two that carry their own:

- the warrant is signed by the author's device key and commits to this context, this method and these exact arguments;
- it is refused unless unexpired, its nonce unspent, and the relay holds `CAN_AUTHOR_ON_BEHALF` — all before anything is executed or published;
- the delta is attributed to the author, and every peer re-checks both signature layers and re-authorizes at the cut.

A node token proves none of that, and none of that needs one.

**Off by default**, and the docs are explicit that it is the gate only under embedded auth: under the default proxy mode the node installs no guard on either router (`with_optional_auth` returns the router unchanged when there is no auth service), so the setting alone neither opens nor closes the path, and a deployment must drive it and its ingress exemption from one source of truth. The drift is silent in the unsafe direction — which is why mero-tee's node image derives both from a single Ansible variable.

### `meroctl context intent` was broken for the same case (`5b208294`)

Found while wiring the fleet relay. The command took the warrant's `executor` from `GET admin-api/identity`, which is wrong twice over:

- **It is guarded**, so it needs a credential on the relay — and on the credential-free relay this feature exists for, the one call the command made to prepare its warrant was the one call it could not make.
- **An identity read does not report the grant.** So the command signed, spent `--nonce`, and learned otherwise from the refusal. Peers refuse a repeat, so the next attempt must pick a higher number and the gap never closes.

Both facts now come from the descriptor via a new `Client::get_intent_relay`, and the command refuses *before* signing when the grant is missing — printing the `set-capabilities` line to hand an admin, with the group and account filled in. `get_node_identity` keeps its other caller (`meroctl account show`).

### Granting at fleet scale (`def24841`)

The per-node grant is the obvious reading of this PR and it does not scale: a namespace admitting TEE nodes continuously would need one admin-signed op per admission. The group's **default-capability mask** already answers that, and two tests in `crates/governance-store/src/warrant_gate.rs` pin it — set the mask once on the namespace and every non-admin member admitted afterwards inherits it, with no per-node op.

Both assert through `account_may_author` (the function `POST .../intents` and the descriptor both call) rather than reading the capability row, since a row the gate does not read would prove nothing. Admission runs through `MembershipPolicy::admit_member_if_absent` — the call both TEE-attestation apply handlers make — rather than the `add_member` it currently delegates to, so a refactor that stops routing admission through `add_member` fails the test instead of passing while production regresses. The second test is the control: without it the first would pass just as happily if admission granted authorship to every TEE node regardless of the default.

This also pins a topology nothing exercised before: a **read-only** TEE node relaying for someone else. That is coherent by design — every `is_read_only_for_context` call site passes a delta's *author*, never its executor — but it was previously true only by code reading.

### Breaking

`AdminConfig::new` now takes `(enabled, public_intents)` — **breaking for `calimero-server`**. One constructor rather than a defaulting one, because this flag decides whether a node exposes a write path to callers holding no credential on it, and a convenience default is how a relay ships with the posture nobody intended, in either direction. One call site in-tree (`merod init`).

## Test plan

```bash
cargo fmt --check
cargo clippy -p calimero-server -p calimero-server-primitives -p meroctl -p merod \
             -p calimero-client -p calimero-governance-store --all-targets -- -D warnings
cargo test -p calimero-server -p calimero-server-primitives -p meroctl \
           -p calimero-client -p calimero-governance-store
```

All green on this branch. Specifically:

- `route_manifest` — passes with the new `GET` row in `endpoints.json`.
- `wire_fixtures` — 13 fixtures including the new `contexts/intent_relay.res.json`, and the regenerate-clean step is a no-op (`94865968` fixed a hand-written fixture whose key order did not match the canonical writer's).
- `meroctl` capability-encoding tests — every flag maps to its own bit (now 8, each distinct), including `--can-author-on-behalf`.
- `calimero-governance-store` — 644 tests, including the two new default-capability tests. Verified by mutation: removing the `set_default_capabilities` call makes the positive one fail.
- `perform_intent`'s unit tests are unchanged and still cover the clock, wrong-context, wrong-args and wrong-method refusals.

**End-to-end coverage** is the paired mero-js PR's `delegated-intent` e2e, which asserts the new descriptor **either side of the grant** — `canAuthorOnBehalf: false` before `setMemberCapabilities`, `true` after — so the field is pinned to the capability rather than to a constant.

### ⚠️ What is NOT covered, stated plainly

**`server.admin.public_intents` has no test.** Grepping the repo for it returns only its implementation and doc comments. The flag whose whole purpose is admitting a caller with no credential is exercised by nothing, and that is a harness gap rather than something closable in this diff:

- the mero-js e2e authenticates in `beforeAll`, so every request carries a session;
- `apps/scaffolding-e2e/workflows/delegated-authorship.yml` opens with `type: login`, reads the relay account via `type: node_identity` (an authenticated admin read), and grants with `capabilities: 512` — a raw integer through the API, never `meroctl`;
- **no scenario anywhere runs `auth_mode: proxy` behind a reverse proxy** — 0 of 137 — which is the default mode, what every hosted node runs, and the only topology where the ingress exemption this flag pairs with exists.

Both suites also relay through the node that is *also* the namespace admin, whereas a fleet node is admitted `ReadOnlyTee` — so the non-admin-relay axis is untested end to end too.

Tracked in #3829 with a reproduction and resolution criteria. The cheapest path there is extending `scripts/e2e-auth-seam.sh`, which already drives raw HTTP and asserts exact status codes from a scenario that deliberately has no `login` step — in both directions, since a test asserting only "the intents routes answer" would pass against a node with no guard at all.

## Proof the blockers were real

- **Before:** `set-capabilities --help` listed seven flags, none setting bit 9 — no CLI invocation could produce a mask containing `CAN_AUTHOR_ON_BEHALF`, and `all_flags_or_together` asserted `count_ones() == 7`. The unit test encoded the gap as correct.
- **After:** `--can-author-on-behalf` is present, `each_flag_maps_to_its_own_bit` covers it, and the same test asserts `count_ones() == 8`.
- **Before:** `GET /admin-api/contexts/<ctx>/intents` → 404. A client had to compose the executor account from `/admin-api/identity` plus a group read it has no group id for and no credential to make.
- **After:** the route returns the descriptor, and the paired SDK e2e shows it flipping with the grant.
- **Before:** `meroctl context intent` read the executor from a guarded route and never checked the grant, so it spent a nonce to discover a refusal.
- **After:** it reads the descriptor and refuses before signing, naming the group and account in the grant command.

## Wire contract (SDK gate)

- [x] Regenerated wire fixtures — added `crates/server/primitives/fixtures/wire/contexts/intent_relay.res.json` plus its row. `IntentRelayApiResponseData` is a new DTO that three SDKs mirror by hand, and `executorAccount` is the one field whose value a client copies verbatim into a *signed* object: rename it and every relay client mints warrants naming nobody, surfacing as a 403 nowhere near the cause rather than as a decode error. The account and the group id are both 64 hex, so the fixture gives them deliberately different values.
- [x] Updated `crates/server/endpoints.json` — one added row, `GET /admin-api/contexts/{context_id}/intents`. Covered by the SDK e2e rather than baselined.
- [x] Linked the matching mero-js PR — below. Purely additive to `calimero-server-primitives` (two new response types); no existing DTO changed.

```
sdk-ref: claude/delegated-execution-rollout-3hlfmf
```

## Documentation update

- `docs/protocol/delegated-authorship.mdx` — how an admin grants the capability, the new discovery read and why it must be answerable *before* the grant exists, how to reach a relay you have no account on (including the embedded-vs-proxy distinction, which is where the mistakes live), and that `meroctl context intent` now makes the read itself.
- `docs/protocol/direct-admission.mdx` — the known gap now records that the *write* half is answered while `/admit` is not, so the remaining work is named rather than implied.
- `CHANGELOG.md` — every addition plus the breaking constructor note.

## Rollout — please read before releasing

This is the core half of a four-repo change. The fleet side is blocked on a release carrying it:

- **mero-tee** pins `merodVersion` and now **fails the image build** when delegated execution is enabled without a merod that has `--public-intents`. It is on `0.11.0-rc.17`, so the relay image cannot be built until a release carries this. That gate is deliberate: without it the image builds, boots, joins namespaces and reports itself to the cloud, then a pre-feature merod ignores the unknown config key and 404s the discovery read — so the cloud advertises a relay that never answers, and it surfaces in someone's browser days later.
- Bumping `merodVersion` also crosses six breaking wire changes between rc.17 and rc.32, one of which (#3796, map entries value-first) does **not** announce itself. So the fleet takes one coordinated jump straight to the release carrying this, rather than paying that flag-day twice.

Paired PRs (branch `claude/delegated-execution-rollout-3hlfmf` in each): **mero-js** calimero-network/mero-js#135 (SDK: cloud client, relay client, `connectCloud`, the e2e above), **mdma** calimero-network/mdma#210 (relay discovery + the machine view), **mero-tee** calimero-network/mero-tee#244 (the fleet node as a relay).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@
   everywhere, but no CLI could grant it, so delegated execution could not be
   turned on at all. Note the mask is replaced, not merged.
 
+- **`Client::get_intent_relay`** in `calimero-client`, and `meroctl context
+  intent` now uses it. The command previously took `executor` from
+  `GET admin-api/identity`, which needs a credential on the relay — so on the
+  relay the feature exists for, the credential-free one, it could not read it at
+  all — and which does not report the grant, so it signed and spent `--nonce`
+  before learning the write would be refused. It now reads both from the
+  descriptor and refuses beforehand when the grant is missing, naming the group
+  and the account in the `set-capabilities` command to ask an admin for.
+
 - **`server.admin.public_intents`** (and `merod init --public-intents`): serve
   the two delegated-execution routes above without a node credential. **Off by
   default.** It opens exactly those two and nothing else, because they carry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,42 @@
 
 ## [Unreleased]
 
+### Added
+
+- **`GET admin-api/contexts/:context_id/intents`** — the relay descriptor a
+  keyholder needs *before* minting a warrant: `executorAccount` (whom the
+  warrant's `executor` must name), `canAuthorOnBehalf` (whether this node holds
+  the grant on the owning group) and `groupId` (whose admin grants it).
+  Both facts belong to the node, so a client could not compose them, and asking
+  after the fact is too late: minting a warrant spends a nonce from a monotonic
+  per-device sequence, and one naming the wrong executor is unspendable.
+  `canAuthorOnBehalf: false` is an answer rather than an error — it is the
+  default state of every context, since the capability is implied by neither
+  membership nor admin.
+
+- **`--can-author-on-behalf`** on `meroctl group members set-capabilities`, and
+  the same row in `check-access`. The capability existed and was enforced
+  everywhere, but no CLI could grant it, so delegated execution could not be
+  turned on at all. Note the mask is replaced, not merged.
+
+- **`server.admin.public_intents`** (and `merod init --public-intents`): serve
+  the two delegated-execution routes above without a node credential. **Off by
+  default.** It opens exactly those two and nothing else, because they carry
+  their own credential — the warrant commits to this context, method and
+  arguments, is single-use, and is refused before execution unless the node
+  holds `CAN_AUTHOR_ON_BEHALF`. A node token proves none of that, and requiring
+  one makes the feature unreachable for the callers it exists for: a browser
+  tab or an agent holding one signing key and no relationship with the relay.
+  This is the write half of the known gap recorded in
+  [direct admission](docs/src/content/docs/protocol/direct-admission.mdx);
+  `/admit` is still behind the guard.
+
+  **Breaking** for `calimero-server`: `AdminConfig::new` now takes
+  `(enabled, public_intents)`. One constructor rather than a defaulting one,
+  because this flag decides whether a node exposes a write path to callers
+  holding no credential on it, and a convenience default is how a relay ships
+  with the posture nobody intended — in either direction.
+
 ### Removed
 
 - **`upgradePolicy`** from the namespace and group-info responses (`GET
--- a/crates/client/src/client/contexts.rs
+++ b/crates/client/src/client/contexts.rs
@@ -4,7 +4,7 @@ use calimero_primitives::context::ContextId;
 use calimero_server_primitives::admin::{
     CreateContextRequest, CreateContextResponse, DeleteContextApiRequest, DeleteContextResponse,
     GetContextIdentitiesResponse, GetContextResponse, GetContextStorageResponse,
-    GetContextsResponse, PerformIntentApiRequest, PerformIntentApiResponse,
+    GetContextsResponse, IntentRelayApiResponse, PerformIntentApiRequest, PerformIntentApiResponse,
     ResyncContextApiRequest, ResyncContextApiResponse, SyncContextResponse,
     UpdateContextApplicationRequest, UpdateContextApplicationResponse,
 };
@@ -29,6 +29,23 @@ where
                 &format!("admin-api/contexts/{context_id}/application"),
                 request,
             )
+            .await?;
+        Ok(response)
+    }
+
+    /// Ask what this node would need to be named as, and whether it may act, before
+    /// minting a warrant for it.
+    ///
+    /// This is the read half of `perform_intent`, and it is a separate call from
+    /// `get_node_identity` for two reasons. It reports `canAuthorOnBehalf`, which
+    /// no identity read carries — and authorship is the default-off capability, so
+    /// the common answer is "no" and finding that out from a refusal costs the
+    /// author a nonce it cannot reuse. And it is readable without a credential on
+    /// this node, which is the whole point of a relay a keyholder has no account on.
+    pub async fn get_intent_relay(&self, context_id: &str) -> Result<IntentRelayApiResponse> {
+        let response = self
+            .connection
+            .get(&format!("admin-api/contexts/{context_id}/intents"))
             .await?;
         Ok(response)
     }
--- a/crates/governance-store/src/warrant_gate.rs
+++ b/crates/governance-store/src/warrant_gate.rs
@@ -365,6 +365,71 @@ mod tests {
         );
     }
 
+    /// A group's default capabilities reach an attested TEE node on admission,
+    /// so a fleet relay does not need a per-node grant.
+    ///
+    /// This is the ergonomic answer to "every fleet node lands
+    /// authorship-closed": set the mask once on the namespace and every
+    /// non-admin member admitted afterwards inherits it. It is asserted through
+    /// `account_may_author` rather than by reading the capability row, because
+    /// that function is what `POST .../intents` and the relay descriptor both
+    /// call — a row that the gate does not read would prove nothing.
+    ///
+    /// The role matters: `ReadOnlyTee` is read-only for its OWN writes, and the
+    /// read-only gate is only ever applied to a delta's author, never to its
+    /// executor. So a read-only TEE node relaying for someone else is coherent,
+    /// and this pins that it is also reachable.
+    #[test]
+    fn a_groups_default_capabilities_reach_an_admitted_tee_node() {
+        let w = seed(7);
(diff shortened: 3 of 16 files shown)


chefsale:
## Coverage note — correcting the test plan above

The test plan says end-to-end coverage is the paired mero-js `delegated-intent` e2e. That is true but incomplete in a way a reviewer should not have to discover, so stating it plainly:

**`server.admin.public_intents` has no test.** Grepping the repo for it returns only `crates/server/src/admin/service.rs` and `crates/merod/src/cli/init.rs` — implementation and doc comments. The flag whose whole purpose is admitting a caller with no credential is exercised by nothing.

That is not an oversight in the diff so much as a gap in the harness, and it is not currently closable here:

- The mero-js e2e authenticates in `beforeAll` (`await mero.authenticate({username, password})`), so every request in it carries a session.
- `apps/scaffolding-e2e/workflows/delegated-authorship.yml` opens with `type: login`, reads the relay's account via `type: node_identity` (an authenticated admin read), and grants the capability with `capabilities: 512` — a raw integer through the API, never `meroctl`. Each is a shortcut only an operator has.
- No scenario anywhere runs merod in `auth_mode: proxy` behind a reverse proxy — 0 of 137 — which is the default mode, what every hosted node runs, and the only topology where the ingress exemption this flag pairs with actually exists.

Filed as #3829 with the reproduction and resolution criteria. The cheapest path there is extending `scripts/e2e-auth-seam.sh`, which already drives raw HTTP and asserts exact status codes (its helper treats 401/403 as "the auth layer speaking") from a scenario that deliberately has no `login` step — so the pattern exists and just needs pointing at these two routes, in both directions: the intents routes answer without a credential **and** a neighbouring `/admin-api/contexts/...` route still refuses. A test asserting only the first half would pass against a node with no guard at all.

## What `def24841` adds

Two tests in `crates/governance-store/src/warrant_gate.rs`, on a different axis — how the capability gets granted at fleet scale:

- `a_groups_default_capabilities_reach_an_admitted_tee_node` — setting the group's default-capability mask to `CAN_AUTHOR_ON_BEHALF` means a node admitted afterwards as `ReadOnlyTee` can relay with **no per-node governance op**. This matters operationally: the per-node grant is the obvious reading of this PR and it does not scale, since a namespace admitting nodes continuously would need one admin-signed op per admission.
- `an_admitted_tee_node_is_closed_when_no_default_is_set` — the control. Without it the first test would pass just as happily if admission granted authorship to every TEE node regardless of the default, which is the failure it exists to rule out.

Both assert through `account_may_author`, the function `POST .../intents` and the descriptor both call, rather than reading the capability row — a row the gate does not read would prove nothing. Admission goes through `MembershipPolicy::admit_member_if_absent`, the call both TEE-attestation apply handlers make, rather than the `add_member` it currently delegates to, so a refactor that stops routing admission through `add_member` fails the test instead of passing while production regresses. Verified by mutation: removing the `set_default_capabilities` call makes the first test fail.

This also pins a topology nothing exercised before — a **read-only** TEE node relaying for someone else. That is coherent by design (every `is_read_only_for_context` call site passes a delta's *author*, never its executor) but it was previously only true by code reading. Note both suites above relay through the node that is also the namespace **admin**, whereas a fleet node is admitted `ReadOnlyTee`; the non-admin-relay axis is still untested end to end, and is noted in #3829.

---

PR #3834, opened 2026-09-04
fix(governance-store): resolve an authorship grant through the membership anchor

chefsale:
# [core] an authorship grant reaches wherever membership reaches

> **No longer stacked.** #3833 merged as `48e68538`, so this now targets `master` directly
> and stands alone. #3833 added the *observation* (`grantedOnGroupId`, which reports where
> a grant lives); this PR changes the *decision*. That split was deliberate — it keeps this
> to a one-function change with a focused security argument.

## Description

`CAN_AUTHOR_ON_BEHALF` was read on the group owning the context and nowhere else,
so a relay fleet admitted once at the namespace root was refused for every context
living in a subgroup — the shape it actually runs in. There was no row to write for
an inherited member either, short of admitting the relay directly to every subgroup.
The descriptor #3833 added reported the root grant while the gate refused it: the node
telling a caller where the grant was, and then turning it away.

`holds_authorship` now defers to `authorship_grant_source`, which resolves the grant
on the owning group and, for an inherited member, on the anchor its membership comes
through. **One function, two directions.**

### Widening — an ancestor grant counts

For an **inherited** member only. Bounded by membership, in three ways that are each
pinned by a test:

- a `Restricted` subgroup is still a wall — it required its own admission, so it
  requires its own grant;
- a node deny-listed off an `Open` subgroup is still refused there (the deny entry
  *is* the removal when there is no direct row to delete);
- a **direct** member of the target does not reach the ancestor at all. A direct row
  is a per-group decision someone made, so the capability row beside it is that
  group's own statement and must not be overridden from above.

The fleet path is the inherited one, which is what makes the third bullet safe rather
than a gap: TEE admission runs through `admit_member_if_absent`, which gates on the
inheritance-aware `is_member` and so writes **no** row for a node that already
inherits, and `MemberJoinedOpen` rejects `Direct` outright. A node that *did* need its
own admission is exactly the node that needs its own grant.

The cost, stated rather than hidden: a root grant now also reaches `Open` subgroups
created *after* it. This reverses a decision `CAN_AUTHOR_ON_BEHALF`'s own doc comment
recorded as deliberate ("every new context is authorship-closed until an admin says
otherwise"), and that comment is rewritten here rather than left contradicting the
code. The scoping did not disappear — it moved to where the admin writes the grant:
on the subgroup for one group, on the root for a fleet.

### Narrowing — a bare capability row no longer authorizes

The old read was the row and nothing else, so the bit alone was permission. No writer
produces a row without membership deliberately — `MemberCapabilitySet` bails unless
the account is already a direct member, and `remove_member` deletes the capability row
alongside the member row — so this closes a class rather than a known live hole: the
non-atomic window in those three deletes, and the next writer that forgets. What it
buys structurally is that the gate's answer is one deny-list-aware predicate instead
of two that can drift apart.

### Peers must agree

Both directions are authorization evaluated **at the cut**. A node running this and a
node running the old read disagree about whether the same delegated delta is
authorized, and would then hold different state. **One coordinated upgrade, not a
rolling one.** There is no wire-format change and no migration; the divergence is in
the verdict.

Determinism is unaffected, and for a stronger reason than the owning-group read gave:
`check_path` already reads `member_capability(anchor, identity)` to check
`CAN_JOIN_OPEN_SUBGROUPS` and decide the inheritance. This reads the same key and looks
at a different bit. A peer that cannot read the anchor cannot resolve the membership
either, and refuses the delta a step earlier.

## Test plan

`crates/governance-store/src/warrant_gate.rs` — 17 tests in the module, 652 in the
crate, all green. Also `calimero-server-primitives` (31 + 13 wire fixtures + 2 of
#3830's `deny_unknown_fields`), `calimero-context-config`, `calimero-context` and
`meroctl`; `cargo fmt --check`; `cargo clippy --workspace --all-targets --features
calimero-storage/testing -- -D warnings` (exit 0, only a pre-existing
`proc-macro-error2` future-incompat note on a dependency); docs `npm run check`.

Three tests are new here and two gained gate assertions:

| Test | Pins |
| --- | --- |
| `a_namespace_grant_reaches_a_subgroup_context` | the widening (was asserting the old refusal) |
| `a_bare_capability_row_with_no_membership_may_not_author` | the narrowing |
| `a_direct_member_of_the_subgroup_does_not_reach_the_ancestor_grant` | the fallback's boundary |
| `a_namespace_grant_is_not_reported_across_a_private_subgroup_boundary` | `Restricted` wall, now at the gate too |
| `a_deny_listed_node_reports_no_grant_for_that_subgroup` | deny-list revocation, now at the gate too |

**Not covered, and worth saying plainly:** there is no merobox scenario for this. The
product-layer proof would be a context in an `Open` subgroup with the grant at the
namespace root and an inherited relay — and `apps/scaffolding-e2e/workflows/delegated-authorship.yml`
creates its context *in the namespace* and grants there, so it cannot fail for this
reason. I could not validate a new scenario in this environment (no Docker daemon), and
an unrunnable scenario touching subgroup key delivery and auto-follow is worse than a
stated gap. It should be a follow-up.

## Proof the fix works

**Reproduction** — restore the old one-row body of `holds_authorship`:

```rust
let Some(bits) = CapabilitiesRepository::new(store).member_capability(group_id, &account)? else {
    return Ok(false);
};
Ok(MemberCapabilities::from_bits_truncate(bits).contains(MemberCapabilities::CAN_AUTHOR_ON_BEHALF))
```

**Before** — `cargo test -p calimero-governance-store warrant_gate`:

```
test warrant_gate::tests::a_bare_capability_row_with_no_membership_may_not_author ... FAILED
test warrant_gate::tests::a_deny_listed_node_reports_no_grant_for_that_subgroup ... FAILED
test warrant_gate::tests::a_namespace_grant_reaches_a_subgroup_context ... FAILED
test result: FAILED. 13 passed; 3 failed
```

**After** — with the fix in place:

```
test result: ok. 17 passed; 0 failed
```

Exactly the three behaviour-change tests fail without it, and every other test in the
file stays green. `a_namespace_grant_is_not_reported_across_a_private_subgroup_boundary`
passes **either way** — which is the point: it is the invariant the widening must not
break, so a mutation that made it fail would mean the widening had become a hole.

## Wire contract (SDK gate)

No DTO or route change — this is an authorization verdict, not a wire shape. The
descriptor's field and its fixture landed in #3833.

- [x] Regenerated wire fixtures — n/a, no DTO changed. Verified anyway after merging
      master: `UPDATE_FIXTURES=1` leaves them byte-identical, so #3830's
      `deny_unknown_fields` does not disturb the field #3833 added.
- [x] Updated `crates/server/endpoints.json` — n/a, no routes changed
- [x] Linked the matching mero-js PR — n/a for this PR. The `IntentRelayInfo` mirror is
      mero-js#135 and pairs with #3828/#3833. Its *semantics* change here (the
      "granted on an ancestor" case now means allowed, not refused), so the mero-js
      guide note needs the same edit — tracked as a follow-up, not required to land this.

## Documentation update

Done in this PR, because three places documented the old rule as deliberate and would
otherwise contradict the code:

- `MemberCapabilities::CAN_AUTHOR_ON_BEHALF` — the reversal is written as a reversal,
  with the cost named.
- `docs/.../capability-inheritance.mdx` — a `CAN_AUTHOR_ON_BEHALF` row in the bit table,
  and an aside marking this as the **one documented exception** to capabilities being
  read from the target group alone, including the `Inherited`-only boundary.
- `docs/.../delegated-authorship.mdx` — the `grantedOnGroupId` table (the third row now
  means "honoured here"), and a note that `<GROUP_ID>` may be the namespace root so a
  fleet needs one grant, not one per channel.
- `CHANGELOG.md` — a `Changed` entry with the coordinated-upgrade warning. It also
  amends #3833's entry, whose third-row semantics and "should the gate later widen"
  paragraph this PR makes wrong.

### DIFF
--- a/.github/actions/setup-merobox/action.yml
+++ b/.github/actions/setup-merobox/action.yml
@@ -9,7 +9,7 @@ runs:
       run: |
         set -euo pipefail
 
-        MEROBOX_VERSION="0.6.72"
+        MEROBOX_VERSION="0.6.75"
         TAG="v${MEROBOX_VERSION}"
 
         case "$(uname -m)" in
--- a/.github/workflows/e2e-rust-apps-release.yml
+++ b/.github/workflows/e2e-rust-apps-release.yml
@@ -64,8 +64,11 @@ jobs:
       - name: Build scaffolding-e2e app
         run: cargo mero build --manifest-path apps/scaffolding-e2e/Cargo.toml
 
-      - name: Build scaffolding-e2e multi-service bundle
+      - name: Build scaffolding-e2e bundle
         run: (cd apps/scaffolding-e2e && cargo mero bundle --dev --no-icon)
+
+      - name: Build scaffolding-e2e-multi bundle
+        run: (cd apps/scaffolding-e2e-multi && cargo mero bundle --dev --no-icon)
 
       - name: Build xcall-example app
         run: cargo mero build --manifest-path apps/xcall-example/Cargo.toml
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -1030,6 +1030,7 @@ jobs:
       # Nodes fetch bundles from here rather than from each other. The URL is
       # discovered on this runner, so it is stamped in rather than prebuilt.
       - name: Serve app bundles from a fixture registry
+        if: matrix.image != 'merod:local-dht'
         run: |
           url=$(./scripts/e2e-fixture-registry.sh)
           ./scripts/set-registry-mode.sh merod:local http "$url"
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -30,6 +30,8 @@ on:
       - "apps/team-metrics-custom/**"
       - "apps/team-metrics-macro/**"
       - ".github/workflows/e2e-rust-apps.yml"
+      - ".github/workflows/e2e-rust-apps-release.yml"
+      - "scripts/lint-e2e-bundle-coverage.py"
       - ".github/workflows/build-merod-image.yml"
       - ".github/actions/setup-merobox/**"
       - ".github/actions/build-local-merod/**"
@@ -55,6 +57,8 @@ on:
       - "apps/team-metrics-custom/**"
       - "apps/team-metrics-macro/**"
       - ".github/workflows/e2e-rust-apps.yml"
+      - ".github/workflows/e2e-rust-apps-release.yml"
+      - "scripts/lint-e2e-bundle-coverage.py"
       - ".github/workflows/build-merod-image.yml"
       - ".github/actions/setup-merobox/**"
       - ".github/actions/build-local-merod/**"
@@ -104,6 +108,8 @@ jobs:
         run: python3 scripts/lint-e2e-convergence.py
       - name: Lint merobox scenarios for unnamed multi-service bundles
         run: python3 scripts/lint-e2e-bundle-shape.py
+      - name: Lint e2e workflows for unbuilt scenario bundles
+        run: python3 scripts/lint-e2e-bundle-coverage.py
 
   # ── Auth seam: client-token e2e over real HTTP ──
   # Proves a client token minted the way the auth frontend mints it can call
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@
   everywhere, but no CLI could grant it, so delegated execution could not be
   turned on at all. Note the mask is replaced, not merged.
 
+- **`Client::get_intent_relay`** in `calimero-client`, and `meroctl context
+  intent` now uses it. The command previously took `executor` from
+  `GET admin-api/identity`, which needs a credential on the relay — so on the
+  relay the feature exists for, the credential-free one, it could not read it at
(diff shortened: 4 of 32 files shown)


chefsale:
One caveat on this PR's check list, since it is easy to read as more coverage than it is.

`E2E - Rust Apps`, `App migration e2e` and `ABI Guard` are **not running here**, and their absence is not a skip you can see in the checks — they simply never triggered. All three declare `pull_request: branches: [master]`, and this PR's base is `claude/relay-descriptor-grant-scope`, so the filter does not match. #3833 does run all three, on the same `crates/**` paths.

That is worth stating on a PR that changes an authorization verdict. The stack still gets that coverage before anything reaches master, in either merge order:

- if this merges into #3833's branch first, that push re-triggers #3833's runs, which do include all three, against the combined change;
- if #3833 merges to master first and this is retargeted to master, the filter starts matching and they run here.

What it does mean is that **the heavy suites have not yet run against this diff**, so please do not treat the green here as end-to-end validation. The evidence for this change is the unit level: 17 tests in `warrant_gate`, mutation-verified, with the three that fail on the old read spelled out in the description. The one thing genuinely missing at the product layer — a merobox scenario for a context in an `Open` subgroup with the grant at the namespace root — is called out in the Test plan as a follow-up; I could not validate a new scenario in this environment.

---

PR #3843, opened 2026-09-04
feat(merod): reset a revoked device identity so the node can re-pair

rtb-12:
Stacked on #3837 (`fix/pair-complete-revoked-device`), which is this PR's base. Rebase onto master once that merges.

## Summary

A revoked device id is spent for good. A node revoked while it was offline keeps presenting the same id, every certificate minted for it is refused by every peer, and nothing re-mints until an enrolment happens to consult the namespace holding the tombstone. There was no way to force that.

`merod --node <name> account device --reset` drops the device row on demand.

**Why merod and not meroctl.** The operation is `NodeDeviceRepository::delete()`, which opens the node's datastore directly, and RocksDB's lock is exclusive while `merod run` is up. That rules out an admin route: there is no running node to serve it. It is the same offline shape as `account export` and `account import`, so it lands as a flag on the existing `merod account device` subcommand rather than as a new command family.

**The rule is reused, not re-derived.** `NodeDeviceRepository::reset_device()` gates on the existing private `stored_identity_still_serves`, evaluated across every namespace this node takes part in. A live device's id *is* its replica lineage (counter slots and an HLC seed), so dropping a row the enrolment path would have kept strands that state with nothing able to author under it again. A device that still serves is refused with a message naming the device, the namespace, and what to do instead.

## Test plan

**Unit** (`crates/governance-store/src/node_device.rs`), both written before the implementation and observed failing:

- `resetting_drops_a_revoked_device` - enrol, tombstone, reset returns the spent id, the row is gone, and the next enrolment mints a different one.
- `resetting_a_device_that_still_serves_is_refused` - enrol, reset is refused, the row is untouched.

Before:

```
error[E0599]: no method named `reset_device` found for struct `NodeDeviceRepository<'a>`
error: could not compile `calimero-governance-store` (lib test) due to 2 previous errors
```

After:

```
test node_device::tests::resetting_a_device_that_still_serves_is_refused ... ok
test node_device::tests::resetting_drops_a_revoked_device ... ok
```

**E2E.** `apps/scaffolding-e2e/workflows/account-device-revoke-lockout.yml` gains a phase 8 on top of the fixture it already builds. `node_exec` runs an offline `merod` subcommand against a stopped node's data directory and is a native merobox step (0.6.75), already used by four scenarios in this repo, so no Rust integration test was needed.

The phase stops node-3, runs `account device --reset`, asserts the id it printed is exactly the device that was revoked, then runs plain `account device` under `expected_failure` to prove the row is actually gone. That step is what makes the reset load-bearing: a `pair-init` re-mints on its own once the tombstone has landed, so the re-pair would succeed either way, and only a direct observation of the empty slot proves the command did the dropping. Node-3 then restarts, re-pairs onto the same account, and the new device id is asserted different from the revoked one under the same account id. Finally the guard is exercised from the other side: the fresh device is live, so a second `--reset` is asserted to be refused.

**Manual smoke.**

```
$ merod --home … --node n1 account device --reset
Error:
   0: this node holds no device identity, so there is nothing to reset
```

**Gates.** `cargo fmt --check`, `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`, `cargo clippy -p merod --all-targets --features mock-attestation -- -D warnings`, `cargo test -p calimero-governance-store` (646 passed), `cargo test -p merod` - all pass.

**Docs.** `docs/src/content/docs/operate/merod.mdx` gains a `### device` section under the account family and the stopped-node caution now names `device` alongside `export` and `import`.

### DIFF
--- a/apps/scaffolding-e2e/workflows/account-device-revoke-lockout.yml
+++ b/apps/scaffolding-e2e/workflows/account-device-revoke-lockout.yml
@@ -83,6 +83,12 @@ description: >
        line no code path emits in this shape. The refusal is still logged at
        `info` on the responder for the day a scenario can drive it.
 
+    8. The lockout has an exit, and it is an operator command. A revoked id is
+       spent for good, so `merod account device --reset` drops the row offline and
+       the re-pair mints a fresh replica under the same account. The same rule
+       guards it in reverse: the fresh device still serves, so resetting it again
+       is refused rather than stranding the state held under it.
+
   What this does NOT cover (intentionally):
     * Self-service revocation (an account revoking its own lost device without
       an admin). Needs the root-signed revocation proof, which is a wire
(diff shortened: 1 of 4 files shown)


rtb-12:
Folded into #3837 as one PR.

PR #3851, opened 2026-09-07
feat(storage): FugueText - a non-interleaving replacement for RGA

xilosada:
## Summary

Adds `FugueText`, a new opt-in text collection that orders characters with Tree-Fugue (Weidner, Gentle, Kleppmann, arXiv 2305.00583).
When two people type passages at the same position while offline, each passage stays contiguous after merge.
`ReplicatedGrowableArray` interleaves them in that case (`SBAbbbaaa` where Fugue gives `SAaaaBbbb`), and is otherwise unchanged by this PR.

`FugueText` is a new type, not a replacement.
It is not storage-compatible with RGA, there is no conversion helper, and nodes that predate this PR cannot decode its two new `CrdtType` tags, so a context should only adopt it once every member runs a build that has it.

### Design

- **Ordering.** `fugue.rs` is the pure algorithm: node `(id, value, parent, side)`, in-order traversal, siblings ascending by id, insert as right child of the left origin unless it already has one, else left child of the right origin.
  This is plain Fugue, not FugueMax.
  The two differ only where some interleaving is unavoidable for any algorithm (paper Figure 7: `AYXBC` vs `AXYBC`); FugueMax needs document order at compare time, which a stateless rebuild cannot provide cheaply.
  A test pins the residual case.
- **Storage.** Text is stored as run-length blocks `TextBlock { start_id, text, parent, side, tombstones }` in an `UnorderedMap`, one row per run rather than per character.
  A run stops growing at `MAX_RUN_LEN = 256` characters, so a keystroke rewrites and ships a bounded block, and a full block's text never changes again.
  A mid-run insert needs no split because every node is addressable as `(replica, counter + i)`.
- **No derived state.** Every read and write rebuilds the tree from the stored blocks.
  There is no cached index, so gas is identical on every replica regardless of cache warmth.
- **Merge.** Blocks are mutable under one key (the owner appends, anyone sets tombstone bits), so last-writer-wins would drop data.
  Rows are tagged `CrdtType::FugueTextBlock` and merged by a join on every apply path: tombstones are OR-ed, and the rest of the record is the maximum under `(node count, text, parent, side)`.
  For honest replicas that is "longer run wins"; for any other input it still converges.
- **Multi-character insert.** `insert_str` resolves the insert rule once for the first character and writes each touched block once.
  Stored bytes are identical to inserting one character at a time.
- **ABI.** `FugueText` has its own `CrdtCollectionType::FugueText` (category `Replayable`), so an RGA to FugueText field swap is reported as `Breaking` by `calimero-abi diff`.

### Known limits

- A document is read by rebuilding it, so cost is linear in length.
  With the default gas budget one keystroke stops fitting at about 21,500 characters mid-document; apps that need more should hold one `FugueText` per paragraph.
- Deleted characters are kept as tombstones; there is no garbage collection.
- A node that loses its state and writes before resyncing can mint an id twice.
  Replicas still converge, and one side's characters lose.
- `insert` and `insert_str` panic inside a migration; `insert_str_with_replica` with a replica constant shared by all nodes is the supported path there.

### Cost

One `insert_str` into an empty document, delta size and apply time (debug build):

| chars | delta | apply |
|---|---|---|
| 64 | 742 B | 0.38 ms |
| 256 | 934 B | 0.41 ms |
| 512 | 1,541 B | 0.76 ms |

The storage-cost workloads, the bytes-per-keystroke gate and the gas probes behind the document-size limit are in the pull request stacked on this one.

### Also in this PR

- `apps/fugue-collab`, the e2e guest app, with five three-node scenarios.
- `ABI Contract` resolves a paired `mero-devtools-js` ref the way `SDK e2e` resolves `mero-js`: a `devtools-ref:` line in the PR body, else a same-named branch, else `main`.
- `merodb` no longer decodes a `FugueText` field as an RGA character map.
- The Fugue suites share one storage harness (`crates/storage/tests/fugue_harness`).

## Test plan

Assertions are on exact text, stored bytes and Merkle roots, not on convergence alone: a merge that loses data does so identically on every replica.

- [ ] `cargo test -p calimero-storage --lib fugue`: algorithm, blocks, run cap, join laws over seeded random blocks including non-prefix pairs, batched insert byte-equivalence, apply-path regressions for each historical data-loss shape
- [ ] `cargo test -p calimero-storage --test fugue_laws --test fugue_conformance` (11 + 7): join commutative, associative, idempotent; paper Figure 2, Figure 7 and the three Table I columns; RGA control asserting the interleaved result
- [ ] `random_scripts_agree_with_the_blockless_control`: seeded differential test against `FugueTextSimple`
- [ ] `cargo test -p calimero-node --test sync_sim text_` (5, under 1 s): lossy, reordered and duplicated delta delivery; partition and heal; late joiner over HashComparison and over level-wise sync; stale and forged blocks; each asserts oracle text, contiguous passages, unique characters, equal Merkle roots and surviving `FugueTextBlock` tags
- [ ] `cargo test -p calimero-storage --test abi_crdt_shapes`, `cargo test -p mero-abi`, `./scripts/verify-abi.sh`, state-schema conformance
- [ ] `ABI Contract`, validated against the paired devtools branch below (calimero-network/mero-devtools-js#55)
- [ ] e2e on three merod nodes: `fugue-collab-figure2`, `fugue-collab-partition-heal`, `fugue-collab-restart-mid-edit`, `fugue-collab-late-joiner`, `fugue-collab-large-paste`
- [ ] `cargo fmt --check` and `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`

devtools-ref: feat/abi-fugue-text

---

> [!NOTE]
> **High Risk**
> New persisted CRDT types and merge paths affect sync/reconciliation across the network; mixed-version nodes cannot decode the new tags, and incorrect block merge would silently lose text.
> 
> **Overview**
> Introduces **`FugueText`**, an opt-in collaborative text CRDT (Tree-Fugue) stored as run-length blocks with per-block `FugueTextBlock` merge semantics, plus new **`CrdtType`** wire tags that require an all-members upgrade. **`ReplicatedGrowableArray` is unchanged**; ABI conformance gains a `fugue_text` field so codegen sees the new collection type.
> 
> Adds **`apps/fugue-collab`** (insert/delete/read RPC) and wires **five merobox E2E scenarios** (Figure 2, partition heal, restart mid-edit, late joiner, large paste) into the Rust-apps workflow, along with a **collaborative-editor concurrent** RGA scenario. **Sync-sim `text_*` tests** exercise real deltas under loss, partitions, and join paths against a pure-tree oracle.
> 
> **CI:** ABI Contract now resolves a paired **`mero-devtools-js`** ref via `resolve-paired-ref.sh` (PR `edited` + `devtools-ref:`), and SDK E2E runs tests for that script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75eba50f022994b6819a781e910eba3251701d1f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xilosada:
Caught up with `master` (71 commits, through d388e7507) and **green**.

The four apply-path reds were one bug: `UnorderedMap::insert_with_storage_type` now stores an entry as `(value, key)`, and `merge_block_entry_bytes` still decoded `Entry<(BlockKey, TextBlock)>` and joined `.1` — so the join ran over the *key*. It never failed loudly: a `TextBlock` opens with a 12-byte `start_id` and a `BlockKey` is exactly 12 bytes, so borsh decoded the reversed tuple happily and the defect showed up only as a wrong merged value. 757 of 761 tests passed with it in place.

That is the same failure shape as the three critical bugs this collection already had, caught by the same apply-path suite — which is the argument for keeping that suite pointed at merged *values* rather than at convergence.

Also folded in three rules master tightened: `MergeStrategy` for the bench fixture, `rand::rng()` for rand 0.10, and two clippy lints in the wall probes.

Now green: 761 storage lib tests + every integration binary, clippy `-D warnings` clean, fmt clean on 1.98.0. #3852 is closed — this is the only PR.

xilosada:
**Demo re-run against the merged branch — reproduces.**

merod 0.11.0-rc.32 (`5d5fe62a5`), master through `d388e7507`, three native nodes, JSON-RPC:

```
FugueText  existing text. HEAD2PASSAGE2HEAD3PASSAGE3HEAD1PASSAGE1   green
RGA        existing text. HEAD3HEAD2HEAD1PASSAGE3PASSAGE2PASSAGE1   red
```

Both converge across all three replicas, both keep all 54 characters, and isolation is asserted before all six writes on both sides (`view_p*` = the bare seed, `view_h*` = the seed plus that writer's own passage). Only the ordering rule differs. Fugue's block order came out different from the pre-merge run — allowed, since what is guaranteed is that no contribution is broken up, not where it lands.

Getting there needed two deployment changes that are worth flagging on their own, because they are not specific to this branch and any pre-existing scenario will hit them:

1. ** no longer accepts a bare `.wasm`** — it answers `not a signed application bundle`. Apps now need `cargo mero bundle` and a `[package.metadata.calimero]` reverse-DNS id.
2. **Bytecode is acquired lazily** (`calimero_app_downloader`). A node that never held the blob cannot execute and falls back to a configured source — so any scenario that isolates or partitions nodes must install on every node *before* isolation, or it fails with `bytecode blob ... not found in blobstore`, which reads as a runtime error rather than a deployment one.

Also, for anyone running merobox locally: `pipx install merobox` currently gives 0.6.56, which cannot drive current core (it sends a `metadata` field the server now rejects). 0.6.75 works. The binary-mode mdns drop is still present in 0.6.75.

PR #3852, opened 2026-09-07
wip: carry FugueText across master's CrdtType and merge-path refactors

xilosada:
Draft, and deliberately red. This is the integration branch for #3851 against current `master`; it exists so the resolution is reviewable rather than sitting in a working tree.

`cargo check --workspace` is clean and 757 of 761 storage tests pass. The four reds are the apply-path suite that pins the `FugueTextBlock` join, so **this must not be merged into #3851 until they are green.**

## What master changed under us, in four days and 71 commits

| refactor | resolution |
|---|---|
| `CrdtType` tags are no longer positional — hand-written borsh based at `0x80`, and `Custom` moved to offset **14**, where `FugueText` sat | assigned 15/16 explicitly, with serializer and deserializer arms; corrected the enum's now-false "discriminants are positional" note |
| `Mergeable` demands a declared strategy | `MergeStrategy { DISPATCHED = false }` for `FugueText`, as a built-in with a fixed rule, exactly like `ReplicatedGrowableArray` |
| `UnorderedMap::insert_with_storage_type` replaced the crdt-typed variant and flipped the entry tuple to `(value, key)` | kept our explicit tag, layered over master's `#[app::mergeable]` derivation with `or_else` |
| the timestamp-blind intercept gained `Custom(_)` and an `is_app_root_entry` guard | our APPLIED-only `FugueTextBlock` arm OR'd alongside it |

The tag collision is the one worth a second reader: these bytes are persisted, sent on the wire and hashed into `delta_id`, so 15/16 is a decision, not a detail.

## A merge artifact worth knowing about

Git kept master's `Collection::insert_with_storage_type` signature but wired its body to our `_and_crdt_type` with a literal `None` — silently dropping every entry's `crdt_type`. It compiled, and 757 tests passed. Only the apply-path suite noticed.

## Where the remaining four stand

`coalesce_meeting_a_delete_on_the_apply_path_keeps_every_node`, plus the three exhaustive/fuzz apply-path sweeps.

Traced as far as: the join is reached correctly — `origin=Applied`, `is_builtin_crdt` true, `merge_fugue_text_block` called with 69 bytes on **both** sides. The equal sizes are the clue. Replica A's coalesced `"abc"` should serialise longer than B's tombstoned `"ab"`, so the suspicion is that **coalescing no longer happens through master's new insert path** — the early `get_mut` return that replaces a value in place — rather than the join being wrong. Next step is to assert block contents directly after a coalescing append, before looking at the merge again.


xilosada:
Folding this into #3851 — splitting the master catch-up into its own PR was needless ceremony. The resolution work moves onto `feat/fugue-on-master` directly.

PR #3857, opened 2026-09-08
feat(governance): seal the joiner's own join when it holds the key

chefsale:
# [governance] seal the joiner's own join when it holds the key

## Description

Follow-on to #3856. That PR sealed a join an **admitter relays**; this one seals the join a **node-side joiner publishes for itself**, which is the common path and the last big source of cleartext membership metadata.

`root_op_is_sealable` says the two invitation joins cannot be sealed, and that is the right answer *for the variant*: it has to answer identically on every node, and some of their publishers hold no namespace key — a browser client signing offline never does.

But the ordinary node-side joiner **does** hold one by the time it publishes. `join_group` unwraps and stores the key the join bundle carried at `join_group.rs:264`, well before the publish at line 517; the `needs_key_wait` branch just below exists only for the case where the bundle carried none. So the publisher can answer a narrower question the classifier cannot: *do I hold the key right now*.

`seal_root_op_if_keyed` asks exactly that, and `join_group` seals its own join when the answer is yes. What that removes from the namespace topic is the one thing a cleartext join tells every non-member: which account joined which group, and when.

Cleartext stays the fallback rather than becoming a failure, deliberately. An unkeyed joiner has nowhere else to go — its key arrives in a `KeyDelivery` an admin publishes *on seeing this op* — so refusing to publish unsealed would deadlock the join rather than protect anything. That is also why this is a separate function from `seal_root_op_for_publish`, which refuses when unkeyed: there an unkeyed publisher is a broken node, here it is a legitimate state.

## What this does NOT do, and why

The classification is **not** flipped to a hard refuse, and one cleartext case remains: an invitation that targets a **subgroup**. `NamespaceRepository::resolve` walks parents to the root, so `group_id` may be a subgroup — and then the bundle delivers that subgroup's key, not the namespace key, so the joiner cannot seal.

That case cannot be closed by relaying through an admitter either, and it is worth recording why, because the relay is the obvious next idea:

- The admitter would publish envelope `Y` while the joiner holds inner op `X`.
- The joiner cannot apply `Y` instead — it lacks the namespace key, which is exactly what defines this case.
- So it applies `X`, a DAG node no peer has. Peers resolving its cited `governance_dag_heads` get `CutMembership::Incomplete` → `DeltaAuthOutcome::Buffer { needed }` (`state_delta/verify.rs:116`), and re-resolve on drain — so the joiner's state deltas buffer **forever**, waiting for an op that was never published. Silently. That is worse than the leak.

Closing it needs an answer to a trust-model question rather than more code: **may an outsider invited into a subgroup hold the namespace key?** Give it to them and they can read all namespace-level governance, which is a privacy widening and the opposite of this series' purpose. Withhold it and they cannot seal, cannot open their own membership op, and cannot be relayed. Filed separately rather than guessed at here.

For completeness, the other publishers are already covered: offline clients (mero-js `signMemberJoinOp`) go through `POST /admin-api/namespaces/:id/admit`, which is the sealed relay from #3856.

## Test plan

`a_joiner_seals_its_own_join_only_when_it_holds_the_namespace_key` pins both branches, and each is load-bearing in the opposite direction — sealing when keyed is the privacy win, and the cleartext fallback when unkeyed is what keeps an unkeyed joiner able to join at all. It also asserts the seal is openable back into the same op, since sealing under the key *id* instead of the key encrypts fine and never opens.

Mutation-checked: making `seal_root_op_if_keyed` seal unconditionally (defaulting a missing key to zeros) fails the unkeyed assertion —

```
an unkeyed joiner must be told it cannot seal, not handed a seal under a key it does not have
```

Full gate, run locally:

```
cargo fmt --check
cargo test -p calimero-governance-store   # 668 passed
cargo test -p calimero-context            # 292 passed + suites
cargo test -p calimero-governance-types
cargo test -p calimero-server
cargo clippy -p merod --all-targets --features mock-attestation -- -D warnings
cargo test -p merod --features mock-attestation   # 105 + 2 passed
```

`mero-auth`'s `storage::providers::rocksdb::tests::test_rocksdb_specific_errors` fails in my sandbox and does so identically with these changes stashed — that crate's only calimero dependency is `calimero-build-utils`, so nothing here reaches it.

The networked join paths are covered by CI rather than locally (merobox needs Docker, unavailable in my sandbox): `38-namespace-invite-join-regression`, `group-join-mesh-not-ready`, `group-join-via-inheritance`, `group-join-then-list-from-joiner` and `offline-account-root` all exercise joins, and the mero-js `SDK E2E (paired)` job covers the admit path with a real offline-signed op.

## Proof the fix works

This is a hardening change rather than a bug fix, so there is no failing-before symptom to show. The evidence that it *takes effect* is the mutation above plus the assertion that the op now travels as `NamespaceOp::RootSealed` and reopens to the same `MemberJoinedAt`:

- **Before**: `join_group` handed `NamespaceOp::Root(RootOp::MemberJoinedAt { .. })` straight to the publisher — readable by every subscriber on the namespace topic, member or not.
- **After**: the same op arrives as `RootSealed { key_id, encrypted }` whenever the joiner holds the key, and the test proves `decrypt_root_op` recovers the identical join.

## Wire contract (SDK gate)

No HTTP wire DTO or route changed, and neither `crates/server/primitives/**` nor `crates/primitives/**` is in the diff, so this gate does not trigger.

The gossip wire form of an existing op changes shape on the sealing path — no new variant, so no new discriminant; a `MemberJoinedAt` published by a keyed joiner now arrives as `RootSealed`. mero-js's own signing path is untouched (it produces the inner op, which is unchanged), so **no paired SDK PR is needed**. Backwards compatibility across a staggered upgrade is explicitly not a goal, so there is no version negotiation and no migration path; this is a coordinated-upgrade change like the rest of the sealing series.

- [ ] Linked the matching mero-js PR — n/a, no SDK change required

## Documentation update

`docs/src/content/docs/protocol/encryption.mdx` — the relayed-join row now states that a joiner with a node of its own seals its own join whenever it already holds the namespace key, and that only a still-unkeyed joiner falls back to cleartext, because its key arrives in answer to that very op.

---


chefsale:
The `E2E Rust Apps Failed` comment above is stale — **CI is green on `d6d7ea84`**. All 167 check runs are `success` (bar the release jobs, which are `skipped` on a PR as usual).

Eight merobox jobs did fail at 15:52Z, but not on anything in this diff. Each died at its *first* step, `actions/download-artifact@v8` for `e2e-app-bundles`:

```
Unable to download artifact(s): Failed to ListArtifacts: Received non-retryable error:
Failed request: (403) Forbidden: Error from intermediary with HTTP status code 403 "Forbidden"
```

No node ever booted, so `docker-logs/` was absent and every post-run gate step printed "(none)" rather than a real assertion failure. Two things confirm it was not the change: every join scenario passed (`38-namespace-invite-join-regression`, `group-join-mesh-not-ready`, `group-join-via-inheritance`, `group-join-then-list-from-joiner`), and two of the eight — `blob-announce-membership` and `shared-storage` — never exercise a join at all.

I re-ran the failed jobs once at 16:26Z and all eight passed: `blob-announce-membership`, `shared-storage`, `group-context-isolation`, `group-open-self-join-key-delivery`, `group-leave-namespace`, `group-subgroup-visibility-inheritance`, `group-subgroup-cold-sync`, `group-list-subgroups-restricted-hidden`. `Report Failed Workflow` also flipped to `success`.

The stale text is just how the reporter works — it comments off the run's original conclusion and does not revisit it after a partial re-run. Judge each job by its newest run in the checks tab. No re-run budget remains on this head: if CI goes red again it gets root-caused, not re-run.

---

Issue #3806, opened 2026-09-02
A paired device never learns namespaces its account gains after pairing


rtb-12:
## Summary

A device paired with `pair-init` / `pair-complete` takes part only in the namespaces named at `pair-init`.
When the holder later creates or joins a namespace, `bind_known_devices` (`crates/governance-store/src/device_link.rs`) binds the device into it on the holder's side, but nothing tells the device the new namespace id.
The device never subscribes to that namespace's topic, never hears its readiness beacons, and never pulls its governance DAG, so the namespace stays invisible to it.
Core has no account-level channel that announces a namespace to the account's other devices; the id only ever travels out of band, inside the pairing invite.

After #3797 a device that missed a publish catches up from beacons, but only on namespaces it already listens on, which is exactly the set this issue is about.

## Impact

Correctness and usability for every deployment that pairs devices, which is the desktop's device-linking flow.
Concrete scenario: a user pairs their phone, then starts a new project on their laptop; the phone is bound into the new namespace on the laptop's side yet never shows it, and no restart of the phone helps, because the startup sweep only walks the namespaces the node already takes part in.
The existing scenario `account-pairing-application-scope` covers the holder half of this (step 4 asserts the binding on node-1) and leaves the device's view unasserted.

## Steps to reproduce

1. On node-1 install an application and create namespace A targeting it. On node-2 run `pair-init` naming A, then `pair-complete` on node-1. Wait until node-2 lists A on `GET /admin-api/namespaces`.
2. On node-1 create namespace B targeting the same application. `GET /admin-api/account/devices` on node-1 shows node-2's device bound in both A and B.
3. Poll `GET /admin-api/namespaces` on node-2 for several minutes, then restart node-2 and poll again.

Actual: node-2 lists only A throughout.
Expected: node-2 lists B with its target application and follows B's contexts, the same as it does for A.

Workaround today: run `pair-init` on node-2 again naming A and B (it is idempotent); the beacon rescue then pulls B.
This needs B's id handed to the device out of band, which is the gap.

## Root cause

The account is a first-class principal, one person with many devices, but it has no replicated state of its own.
Its device list and the application scope of each device live only in the holder node's local store.
Its namespace set exists only as binding rows scattered across other namespaces' DAGs.
So every account-level fact has to travel some other way: the namespace list rides the pairing invite, widening is a fan-out from one node, and revocation reaches only the namespaces the holder takes part in.
[...]

PR #3869, opened 2026-09-09
feat(governance-store): cache sibling certificates by the account a device speaks for

rtb-12:
## Summary

- `NodeDeviceRepository::is_account_holder` tells a node holding its own account's root from a paired one.
- The gain-time bind on a non-holder binds no cached device: only the holder stores application scopes.
- The holder binds devices of its own account only: a cached certificate naming any other account is out of scope in every namespace, which covers the certificate rows a forced `account import --force` leaves behind after it discards the account and deletes the device row.
- The sibling certificate cache keys on the enrolled device's account instead of the node's root, and is fed from founder and join credentials as well as link ops, so a paired device holds its siblings' proofs.

No wire or on-disk format changed: the cache row is the same `NodeAccountDeviceCertValue` under the same key, and only which node writes it changed.

Stacked on #3868. Part of #3806.

## Test plan

- [x] `cargo test -p calimero-governance-store` - 681 pass, including three new apply-level tests (a sibling's join credential and a sibling founder's genesis credential are cached; a stranger's is not) and the holder-account check in `device_link`.
- [x] `cargo test -p calimero-context` - 294 pass; it is the only other crate that drives the bind fan-out.
- [x] `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings`, `cargo fmt --check`.

merobox, `merod:local` built from this branch, each scenario from a clean node state:

- [x] `account-paired-device-founds-namespace` - 15/15 steps; "Founding C did not change what the phone speaks as" green.
- [x] `account-pairing-application-scope` - 41/41 steps; the pair-complete refusals (400/400/409), the app-A-only reach, the relink widening, and revocation reaching A1, B1 and A2.
- [x] `account-facts-cross-namespace` - 26/26 steps; `revoked_in` contains both namespaces.
- [x] `shared-storage-account-writers-two-devices` - all steps.
- [x] `account-device-revoke-lockout` - all steps.
- [x] `account-root-backup-restore` - all steps.
- [ ] `account-pairing-missed-publish` - not runnable on the macOS host used here: its `partition_peers` step shells out to `sudo iptables` on the host, which needs a Linux host with passwordless sudo. It failed at that step before reaching any assertion; CI runs it on Linux.

### DIFF
--- a/crates/governance-store/src/device_link.rs
+++ b/crates/governance-store/src/device_link.rs
@@ -27,6 +27,27 @@ enum BindPlan {
     Skip(BindOutcome),
 }
 
+/// Does this node know `cert`'s device belongs in `namespace`?
+///
+/// Only the holder stores scopes, so a paired device cannot answer for any
+/// sibling and binds none.
+fn in_scope_here(
+    store: &Store,
+    devices: &NodeDeviceRepository<'_>,
+    namespace: &ContextGroupId,
+    cert: &KnownDeviceCert,
+) -> EyreResult<bool> {
+    if !devices.is_account_holder()? {
+        return Ok(false);
+    }
+    // A namespace whose metadata has not synced names no application, and is
+    // reachable only by a scope that names none either.
+    let application = MetaRepository::new(store)
+        .load(namespace)?
+        .map(|meta| meta.target.application_id);
+    Ok(cert.covers(application))
+}
+
 /// Everything that decides whether the pair of publishes is worth making,
 /// cheapest question first.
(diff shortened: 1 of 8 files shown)


rtb-12:
Converted to draft and held. The holder check and the non-holder-binds-nobody rule are interim guards until device scopes are shared through the account namespace, and the sibling cache feed is superseded once the account namespace is the single source of certificates. The account namespace stack lands first; this branch stays as the reference for the account-keyed cache and its tests.

rtb-12:
Closed: superseded by the account-namespace stack. In the final design the account namespace itself is the device registry (certificates arrive in its link rows, scopes as root-signed ops), so the node-local sibling cache and the holder check are not needed. Staging is on #3806.

PR #3870, opened 2026-09-09
feat(store,governance-store): a paired device carries its certifier into namespaces it gains

rtb-12:
## Summary

- A device records the member key that endorsed its own link, in a new node-local singleton row beside its imported certificate.
- The gain-time bind on a non-holder binds the cached sibling certificates whose signing key is that endorser, treated as unscoped, and nobody else.
- That bind is also guarded on the account: a cert cached under an account this device no longer speaks for is out of scope even when its signing key matches, so a forced root import followed by a re-pairing cannot leak a discarded account's device into the new namespace.
- The scenario `account-paired-device-founds-namespace` now also asserts the laptop is bound in the namespace the phone founded.

Stacked on #3869, which stacks on #3868. Part of #3806.

## Test plan

- [x] `cargo test -p calimero-governance-store -p calimero-store`
- [x] "The phone carried the laptop into C" fails on the previous PR's head (output below) and passes here (output below).
- [x] `account-pairing-application-scope` and `account-facts-cross-namespace` still pass.
- [x] `account-paired-device-founds-namespace` re-run against a freshly rebuilt `merod:local` after the account guard: 16/16 steps, output below.

### Unit test - RED then GREEN

Before (this task's own new test, run against the pre-change branch):

```
$ cargo test -p calimero-governance-store certified_it
test device_link::tests::a_paired_device_binds_the_device_that_certified_it_and_no_other_sibling ... FAILED
assertion `left == right` failed
  left: OutOfScope
 right: Linked { key_delivered: true }
```

After:

```
$ cargo test -p calimero-governance-store -p calimero-store
test result: ok. 686 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 83 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

### Scenario - RED then GREEN

Before, on the previous PR's head (`82aa6e2ac`):

```
Executing The phone carried the laptop into C (assert_api_response)...
✗ assert_api_response on paired-founder-node-2: GET /admin-api/account/devices missed 1 of 3 assertion(s)
    namespaces: expected to contain
    '424d2009a6fa44d1fa9360e9d83a949bc155a081498b88f4ecb9bd6462cc203d', got
    ['42427cc925d5f988ceeb5eec5ab0290309baa7f4589febbac51c42d7888e4949']
❌ Step 'The phone carried the laptop into C' failed
```

with the node log showing the bind decided `OutOfScope` for the only device it considered:

```
INFO calimero_governance_store::device_link: carried this account's devices into a namespace namespace_id=ContextGroupId(...) outcomes=[(DeviceId([172, 69, 85, ...]), OutOfScope)]
```

After, on this branch (re-run after the account guard, on an image rebuilt from this head):

```
Executing Founding C did not change what the phone speaks as (json_assert)...
✓ JSON equality passed
✓ JSON equality passed
✓ Step 'Founding C did not change what the phone speaks as' completed

Executing The phone carried the laptop into C (assert_api_response)...
✓ assert_api_response: GET /admin-api/account/devices on paired-founder-node-2 satisfied 3 assertion(s)
✓ Step 'The phone carried the laptop into C' completed

🎉 Workflow 'E2E Account - A Paired Device Founds A Namespace And Carries Its Certifier' completed successfully!
```

All 16 steps passed; the phone still reports the same account and device after founding C.

`account-pairing-application-scope` (41 steps) and `account-facts-cross-namespace` (26 steps) both still pass in full, including the pair-complete refusals, relink scope widening, and revocation-reaches-every-namespace assertions.


rtb-12:
Converted to draft and held. The certifier row and the certifier-only bind rule exist only until device scopes are shared through the account namespace, at which point they would be deleted. The account namespace stack lands first; this branch stays as the reference for the guest-list write path and its scenario.

rtb-12:
Closed: superseded by the account-namespace stack. In the final design every device binds its in-scope siblings from the account namespace's registry, so the certifier row and the certifier-only bind rule are not needed. Staging is on #3806.

Issue #3871, opened 2026-09-09
RootOp::KeyDelivery applies with no authorization and stores an unbound key


chefsale:
## Summary

`RootOp::KeyDelivery` reaches its effect with **no authorization on the deliverer** and stores the delivered key **without binding its content to any signed op**. Any holder of the namespace key can have a key of their own choosing written into another node's keyring for any group in that namespace, and become the current key that node encrypts under.

The apply handler is a deliberate no-op — `crates/governance-store/src/ops/namespace.rs:154`:

```rust
RootOp::KeyDelivery { .. } => Ok(()),
```

Authorization in this crate is opt-in per handler (compare `ops/namespace/context.rs:83` `require_namespace_admin`, or `admin_changed.rs` / `group_reparented.rs` / `policy_updated.rs`, which all gate on `require_namespace_admin`). `dispatch_root_op` (`ops/namespace.rs:35`) and `apply_root_op` add no generic gate, and the receive path (`crates/node/src/handlers/network_event/namespace.rs:143`) checks only namespace-vs-topic and the signature.

The whole effect runs in `root_op_side_effects` → `apply_received_group_key_envelope_at_depth` (`crates/governance-store/src/namespace/governance.rs`), which is called with `expected_key_id: None`. The gates that *are* present all pass for an arbitrary namespace member:

- the cross-namespace pin passes whenever the named group is in this namespace;
- `addressed_to_us` passes for a `Member`-addressed envelope naming the recipient's identity;
- `unwrap_any(..., Some(&responder_identity))` → `check_sender` (`group_keys.rs`) only requires `envelope.sender == op.signer`, which a self-signed op satisfies by construction.

There is no `is_member`, admin, capability or key-holder check on the deliverer, and no check that it has any relationship to the group being delivered for.

Because `expected_key_id` is `None`, the content-binding block is skipped. Its own comment states the threat:

> any key-holding member of the group can mint a valid wrap around a key of its own choosing… a node that stores a chosen key seals its own subsequent writes under it — readable by whoever chose it.

Minting the envelope requires no secret: `GroupKeyring::wrap_for_member(sender_sk, recipient_pk, group_id, group_key)` takes the attacker's own key, the victim's **public** identity key, an arbitrary group id and an arbitrary key.

[...]

PR #3886, opened 2026-09-09
fix(governance): bind a delivered key to the id an awaiting op names

chefsale:
# [governance] bind a delivered key to the id an awaiting op names

Closes the content half of #3871. The authorization half landed in #3875.

## Description

The anchor gate from #3875 answers **who may deliver** a key. It cannot answer **what was delivered**, and those are separate powers: an admin of a group may hand out its key without being entitled to choose which key the group uses. Only the first was checked.

### The correction this PR rests on

`RootOp::KeyDelivery` names no `key_id`, and #3875 read that as "no hash can check this" and passed the expected-id set empty. That was the wrong conclusion from a correct observation, and it is stated as fact in #3875's own body and in #3871's checklist:

> "This closes the authorization half; the content half cannot be closed without a `key_id` on the op, which is a wire change."

**A `key_id` on the op would be security theatre.** It would be signed by the deliverer alongside the key it chose, so the hash would match by construction and check nothing. Binding is only worth something when the id comes from a **different** signed op.

And one is usually to hand. A buffered `NamespaceOp::Group` envelope is authored by a group *member* and names the `key_id` its ciphertext was encrypted under — a third party relative to whoever is delivering. `awaited_key_ids_for` collects those for the group, which is the same source `apply_received_group_key` is already handed on the pull path (`namespace_group_keys_awaiting` → `awaited_group_keys` → `collect_buffered_group_op_keys`).

So the content half closes with **no wire change, no new discriminant, and no coordinated upgrade** — the opposite of what was recorded on the issue.

### A set, not one id

`expected_key_id: Option<[u8; 32]>` becomes `expected_key_ids: &[[u8; 32]]` through the three apply signatures and the public entry point.

A group stranded across a rotation has buffered ops naming two different keys, and **both are legitimately deliverable**. Picking one and refusing the other would reject a key the node is genuinely waiting for, so the check accepts any of them. The pull path passes `Option::as_slice()`, which is exactly a 0-or-1 set.

Empty means "nothing signed names a key for this group yet". That is the ordinary first-delivery state, and it leaves the anchor gate as the only check — precisely as `key_servers_allowed` decides for an unverifiable key on the pull path. The two paths now agree on both halves.

### A failed lookup refuses

An I/O failure is not evidence that no op names a key, and reading it as such is the exact downgrade — checked to unchecked — that this binding exists to prevent. It joins the "could not tell" outcome #3875 added, logged as itself and left to the pull rather than propagated (this arm's errors never propagate: a failed apply orphans every op causally after it).

## Test plan

Run through `./scripts/check-like-ci.py` (#3883), which reads the steps out of the workflow:

```
  Cargo format                     ok              5.9s
  Naming gate                      ok              0.2s
  Cargo clippy                     ok            758.0s
  Cargo clippy (mock-attestation)  ok            553.8s

4/4 passed, 10 not selected
```

Plus, directly:

```bash
cargo test -p calimero-governance-store --no-fail-fast   # 694 passed, 0 failed (all targets)
cargo test -p calimero-node --lib                        # 586 passed, 0 failed
cargo test -p calimero-node --features mock-attestation --lib   # 608 passed, 0 failed
```

The `mock-attestation` pair is run explicitly because it is default-off: that is the step whose absence broke the previous round of this work, and no default-feature run compiles it.

Two new tests:

| test | what it pins |
| --- | --- |
| `an_anchors_key_delivery_is_refused_when_it_is_not_the_awaited_key` | the property, at its sharpest: the signer **is** a trusted anchor of the group and is still refused, because the substituted key hashes to nothing a signed op names. Authorization alone is not enough |
| `a_key_delivery_matching_the_awaited_key_is_applied` | the positive control — same store, same anchor, same envelope shape, the awaited key. Without it, a fixture that refused everything would look like a working binding |

**Verified by neutering the binding** (passing the set empty): the refusal test fails, the positive control passes either way. That split is what makes the pair worth having.

`clippy` also caught a real defect in this PR's own code: `Option::as_slice()` exists, and the first version hand-rolled `map_or(&[][..], std::slice::from_ref)`. Fixed and folded in.

**Not run: merobox.** No Docker daemon here. A scenario for this needs the same `script` step #3871 already wants — a hand-minted delivery, which no supported client emits — and the substitution case additionally needs a buffered group op to exist first, so it is a strictly harder scenario than the authorization one.

## Proof the fix works

**Reproduction** (`crates/governance-store/src/namespace/tests.rs`, `an_anchors_key_delivery_is_refused_when_it_is_not_the_awaited_key`): namespace `N` keyed, Restricted subgroup `S` whose meta names `A` as admin, so `A` is a genuine anchor of `S` and passes #3875's gate. A buffered `NamespaceOp::Group` for `S` names the real key's id. `A` then delivers `[0xEE; 32]` instead.

**Before** (binding neutered, i.e. the state master is in after #3875):

```
thread '...an_anchors_key_delivery_is_refused_when_it_is_not_the_awaited_key' panicked:
being allowed to deliver this group's key is not being allowed to choose it: the
substituted key does not hash to the id a signed op names, so it must not be stored
```

The substituted key was stored and — per the `key_rank` reasoning in #3875 — became `S`'s current key. The deliverer was fully authorized; only the content was wrong.

**After:** refused, `S` holds no key, the DAG apply still succeeds (a refusal is a skipped effect), and the same anchor's delivery of the *awaited* key still applies.

## Wire contract (SDK gate)

No HTTP wire DTO, route, on-disk format or op encoding changed. `RootOp` and `NamespaceOp` are untouched — that is the point of the design correction above.

- [ ] Regenerated wire fixtures — not applicable, no DTO changed
- [ ] Updated `crates/server/endpoints.json` — not applicable, no routes changed
- [ ] Linked the matching mero-js PR — not applicable, no wire change

`apply_received_group_key`'s signature changes, but it is an internal Rust API with one caller outside the crate (`crates/node/src/sync/manager/namespace_sync.rs`), updated here.

**No coordinated upgrade.** A node running this refuses to adopt a key that contradicts an op it already holds; nothing a compliant publisher emits is affected.

## Documentation update

- `docs/src/content/docs/protocol/encryption.mdx` — the *trusted anchor* section now opens on the two halves ("who delivered it, and what was delivered, are separate powers"), states why a `key_id` on the op would check nothing, names the third-party source, and explains why any awaited id is accepted rather than one.
- `crates/governance-store/AGENTS.md` — the same as an invariant.
- Both replace text from #3875 that described every delivery as unverifiable. That was true of the op in isolation and false of the store's knowledge, which is the whole of this change.

## Related

- #3875 — the authorization half; its body's claim that this needs a wire change is corrected here.
- #3871 — the issue. After this, only the `key_rank` hardening (item 4) remains open on it, and it is independent of both halves.
- #3883 — the CI-parity runner used for the verification above.

---


chefsale:
## Correction: this PR's description overstates what the content binding proves

Flagging it here because the body is merged and cannot be meaningfully edited, and a reader would otherwise take it at face value.

This PR's description presents the binding as resting on an attestation from a *different* signed op — effectively a third-party check on which key a group uses. **It is not.** The `key_id` it binds against is read from the **cleartext** `key_id` field of a buffered `NamespaceOp::Group` envelope, and nothing checks that field:

- the receive path verifies only the topic and the signature — not membership, not an account binding, not revocation (`crates/node/src/handlers/network_event/namespace.rs:147-155`);
- an unresolvable `key_id` decrypts nothing, runs no authority gate and raises no error (`namespace/governance.rs:643-673`) — the ciphertext is never inspected, because not holding the key is *why* the op is buffered;
- the op is head-advanced and written to the op log regardless.

So the same principal can mint the hint and then satisfy it. The binding is a **decryption hint and a content check**, never authority over what a group's key is.

This mattered concretely. #3887's first form exempted a *bound* delivery from the seed-not-replace rule on the strength of this claim, which made "bound" strictly more powerful than "unbound" — the precise escalation that PR existed to prevent. The exemption was removed before merge, and #3888 was filed for the same field's effect on the pull path (closed by #3892 as `a76005de`).

The accurate account is in #3887's and #3892's descriptions, and the invariant is recorded in `crates/governance-store/AGENTS.md`. Nothing here needs reverting — the code this PR merged is sound, and it is the surrounding claim that was wrong. I wrote that claim, and I had not checked the field before asserting it.

---

Issue #3888, opened 2026-09-09
A forged awaited key_id makes a group-key pull acceptable from any peer


chefsale:
## Summary

The direct group-key pull decides how far to trust a responder from whether the requested key is "verifiable" — meaning some op names its `key_id`. That id comes from a **cleartext, unauthenticated field of a buffered op**, so a caller can mint it. Doing so flips `key_servers_allowed` from *anchor-only* to *every candidate*, and the pull then adopts a key of the attacker's choosing.

Found during a composed security review of the key-delivery path (#3871, #3875, #3886, #3887). It is **not introduced by any of those** — it pre-dates them — but they are what prompted looking.

`key_servers_allowed` (`crates/node/src/sync/peers.rs:275-287`) returns every candidate when the request is verifiable, and anchors-only otherwise:

```rust
if verifiable {
    return (total_candidates > 0).then_some(total_candidates);
}
if !anchors_known || anchor_count == 0 {
    return None;
}
Some(anchor_count)
```

`verifiable` is `key_id.is_some()` (`crates/node/src/sync/manager/namespace_sync.rs`, the op-driven request set), and that id comes from `namespace_group_keys_awaiting` → `awaited_group_keys` (`namespace/retry.rs:191-206`) → `collect_buffered_group_op_keys` (`namespace/op_log.rs:611-618`), which reads the `group_id` and `key_id` **cleartext fields** of buffered `NamespaceOp::Group` envelopes.

Nothing on that route checks the field:

- the receive path verifies only the topic match and the Ed25519 signature over `op.signer` — not membership, not an account binding, not revocation (`crates/node/src/handlers/network_event/namespace.rs:147-155`);
- `NamespaceOp::validate` only bounds ciphertext length;
- in `apply_signed_op`, an unresolvable `key_id` means no decrypt is attempted, no authority gate runs and no error is raised (`namespace/governance.rs:643-673`) — the ciphertext is never inspected, because not holding the key is precisely why the op is buffered;
- the op is then head-advanced and written to the op log regardless;
- `awaited_group_keys` keeps any pair whose `key_id` the node does not hold — and an attacker-minted id is by construction not held.

So the "an op is awaiting this key" signal, which the pull treats as authority to widen who may serve it, is authored by whoever wants it to exist.

## Impact

**Confidentiality**, for any group whose current key is at epoch `0` — i.e. any group that has never rotated, including a namespace root still on its genesis key.

[...]

PR #3896, opened 2026-09-10
feat(node,ci): report which ground authorized each group-key recovery

chefsale:
# [node, ci] report which ground authorized each group-key recovery

Closes the coverage gap behind #3892's two red E2E rounds. Observe-only: one log field, one informational workflow step, no assertion.

## Description

`key_server_accepted` allows exactly two grounds for serving a group key (#3888, #3892):

- the responder is a **trusted anchor** of the group;
- or it proved, with a certificate chaining to this node's account root, that it is a device of **this node's own account**.

The accept path logged `"recovered group key via direct delivery"` and **not which one**. So the suite could tell us a key arrived, never that the second ground was the reason it arrived.

**That distinction is not academic.** Every scenario exercising this path passed before the own-account ground existed — and it then shipped **inert**: a node that provisioned its own root holds no certificate for its own device, so the proof was always empty and acceptance quietly collapsed back to anchor-only. Two red E2E rounds and one security hole (a replayable public certificate) came out of that. A green suite could not have distinguished the working mechanism from the broken one, because "a key was recovered" was true in both cases.

`key_recovery_authorized_by` now names the ground, and a new workflow step reports the distribution across each scenario's node logs.

### Why this asserts nothing

Deliberate. Nobody yet knows which ground each scenario legitimately uses — in a three-node mesh the anchor and the own-account sibling may both be reachable, and which one answers first is not fixed. Pinning a guess is how a gate starts failing for the wrong reason, and I would rather add a step that reports than one I have to keep re-tuning.

Observe first, assert once the distribution is known. That is how the `unified_projection_*` and `op_store_incomplete` gates in this same workflow were built, and their comments say so: *"Observe-only per scenario, and deliberately so."*

## Test plan

```
  Cargo format                     ok              4.0s
  Naming gate                      ok              0.1s
  Cargo clippy                     ok             45.7s
  Cargo clippy (mock-attestation)  ok             41.9s
4/4 passed
```

via `./scripts/check-like-ci.py` (#3883), plus `cargo test -p calimero-node --lib` → **586 passed, 0 failed**.

**The extraction pipeline was validated locally**, against synthetic ANSI-styled log lines in both field forms:

```
      1 key_recovery_authorized_by=anchor
      1 key_recovery_authorized_by="own_account_device"
```

That check is not ceremony. `tracing` writes escapes *between* a field name and its `=`, so the raw pattern silently matches nothing — the exact failure the neighbouring steps in this workflow carry warnings about (*"which is how 'planes seen:' used to print an empty list under a failing gate"*). The YAML was also parsed to confirm the step is well-formed.

**Not run: merobox.** Confirmed unavailable here — the `docker` binary is present but its daemon is unreachable, and `merobox` is not installed. This PR's own E2E run is what produces the first distribution, which is the point of it.

## What this does and does not change

- **Behaviour: none.** One `info!` field added on a path that already logged, and one `if: always()` reporting step that never fails.
- No HTTP wire DTO, route, on-disk format or op encoding changed.

- [ ] Regenerated wire fixtures — not applicable
- [ ] Updated `crates/server/endpoints.json` — not applicable

## A correction worth recording

I repeatedly described the key-delivery paths as having "no merobox coverage". Auditing the suite before writing anything, that was **too strong**: 20+ of the 68 scenarios touch key delivery, pull, or subgroup keys, and that coverage is exactly why four regressions were caught today rather than shipped.

The accurate gap is narrower and is what this PR addresses: the *paths* are exercised, but the *security properties* the recent fixes established are not distinguished from the behaviour that preceded them. Attack coverage remains out of reach for a different reason — the suite has no step type that emits a hand-crafted governance op, so "a stranger mints a `key_id`" is not expressible in YAML at all.

I also considered asserting the `"deferring a KeyDelivery"` marker from #3895 and did not: it only appears when the delivery wins a race against `GroupCreated`, so asserting it would be flaky by construction.

## Related

- #3892 — added the own-account ground this reports on (merged as `a76005de`).
- #3888 — closed; the forgeable-`key_id` widening.
- #3894 — records the traps that made the inert shipment possible.
- #3895 — the deferral fix, in flight; its marker is deliberately not asserted here.

---

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -1283,6 +1283,41 @@ jobs:
           fi
           echo "no blob chunk hash mismatches — every chunk served passed its content-hash check"
 
+      - name: Report which ground authorized each group-key recovery (observe-only)
+        if: always()
+        run: |
+          # `key_server_accepted` allows exactly two grounds: the responder is a
+          # trusted anchor of the group, or it proved with a certificate chaining
+          # to this node's account root that it is a device of this node's OWN
+          # account (#3888, #3892). Every scenario that exercises the key pull
+          # passed before the second ground existed, so a green suite is not
+          # evidence that it works -- and it shipped INERT once already, because
+          # a node that provisioned its own root holds no certificate for its own
+          # device, so the proof was always empty and acceptance quietly
+          # collapsed back to anchor-only. Two red rounds and a security hole
+          # came out of that, and the suite could not have told us.
+          #
+          # So this reports the distribution rather than asserting one: nobody
+          # yet knows which ground each scenario legitimately uses, and pinning a
+          # guess is how a gate starts failing for the wrong reason. Observe
+          # first, assert once the distribution is known -- the same way the
+          # projection and op-store gates were built. INFORMATIONAL, never fails.
+          #
+          # ANSI escapes sit between a field name and its `=` in node logs, so
+          # the sed is not optional; matching the raw pattern finds nothing.
+          if grep -rl "key_recovery_authorized_by" docker-logs/ 2>/dev/null | grep -q .; then
+              echo "group-key recoveries by authorizing ground:"
+              grep -rho 'key_recovery_authorized_by.*' docker-logs/ \
+                | sed -E $'s/\x1B\\[[0-9;]*[mK]//g' \
+                | grep -oE 'key_recovery_authorized_by="?[a-z_]+"?' | sort | uniq -c | sort -rn
+          else
+              # Not a failure: most scenarios never need the direct pull at all,
+              # because the delivery reaches them first. Silence here and silence
+              # from a broken marker look identical, which is why the suite-wide
+              # view is what matters rather than any one scenario.
+              echo "no key_recovery_authorized_by markers - this scenario recovered no group key by direct pull"
+          fi
+
       - name: Report scope_root governance-divergence pulls (P6.S2, observe-only)
         if: always()
         run: |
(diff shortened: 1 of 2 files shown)


chefsale:
First distribution is in, from the `e2e-rust-apps` run on the merge commit (run 34473093881). **Both grounds are live — the own-account path is not dead code:**

| scenario | `key_recovery_authorized_by` |
| --- | --- |
| `account-device-revoke-lockout` | 2 × `"own_account_device"` |
| `group-subgroup-cold-sync` | 1 × `"anchor"` |
| `group-late-joiner`, `group-kick-and-rejoin-keyshare` | no markers (delivery arrived before any direct pull was needed) |

That answers the question this step was built for. When #3892 shipped inert, acceptance silently collapsed to anchor-only and the suite still went green — so a green suite was never evidence the second ground worked. It now is: `account-device-revoke-lockout` recovers its key **only** via `own_account_device`, so an inert proof would show up here as a changed distribution rather than as silence.

The split also matches the design rather than accidentally covering it: account/device scenarios recover from a second device of the *same* account (`own_account_device`), cross-account group scenarios recover from a group anchor (`anchor`).

Two follow-ups this makes possible, both out of scope for the observe-only step:

1. A suite-wide roll-up (the way `projection-coverage` aggregates the projection gate) that asserts the whole suite sees **at least one recovery on each ground**. Per-scenario assertion is still the wrong shape — most scenarios legitimately emit nothing — but "neither ground was exercised anywhere" is the fail-open case worth gating.
2. Once (1) exists, the per-scenario step can stay informational permanently.

---

PR #3902, opened 2026-09-10
ci: assert both key-serve grounds are exercised across the e2e suite

chefsale:
# [ci] assert both key-serve grounds are exercised across the e2e suite

## Description

The follow-up #3896 explicitly deferred. `key_server_accepted` allows exactly two grounds — the responder is a **trusted anchor** of the group, or it proved by certificate that it is a device of this node's **own account** (#3888, #3892) — and #3896 added a per-scenario step that *prints* which ground authorized each recovery, observing rather than asserting because the legitimate distribution was unknown.

**Printing is not gating.** If a ground stopped working entirely, every scenario would print zero of it and the suite would stay green. That is not hypothetical: #3892 shipped the own-account ground **inert** — a node that provisioned its own account root holds no certificate for its own device, so the proof was always empty and acceptance silently collapsed back to anchor-only — and the whole suite passed anyway, twice. A green suite was never evidence the second ground worked.

New job `key-recovery-coverage` pools every scenario's logs and fails if either ground went unexercised across the whole run. The per-scenario step stays informational, and its comment is updated to point at the floor instead of promising one.

## The distribution the floor is set from

Measured on **this PR's own run**, which is the whole suite rather than a sample:

```
25 anchor
 6 own_account_device
```

| ground | suppliers |
| --- | --- |
| `own_account_device` (6) | `shared-storage-account-writers-two-devices` ×2, `account-device-revoke-lockout` ×2, `account-pairing-missed-publish`, `account-facts-cross-namespace` |
| `anchor` (25) | `auto-follow-*` (4 scenarios, 6 recoveries), `group-leave-member`/`-namespace`/`-context`, `group-private-add-then-write`, `group-kick-and-readd-deny-list` ×3, `group-membership`, `group-metadata`, `group-subgroup`, `group-subgroup-queued-deltas`, `group-subgroup-cold-sync`, `group-list-subgroups-restricted-hidden`, `dm-subgroup-privacy`, `account-pairing-application-scope` |

The split matches the design rather than accidentally covering it: account/device scenarios recover from a second device of the *same* account, cross-account group scenarios recover from a group anchor.

**So losing a ground takes ~20 scenarios going quiet at once, not one.** That is not a plausible timing coincidence — it is what a real regression looks like, which is what the gate should fire on. `own_account_device` is the narrower floor at 4 suppliers, still well above one, and its scenarios are structural rather than racy: a newly paired device has no delivery addressed to it, so pulling from its own account's other device is the only route.

> **Superseded caveat, kept for the record.** An earlier revision of this description warned that the floor rested on a single supplier per ground and asked a reviewer to weigh that flake risk. That came from sampling 6 of 88 scenarios by guessing which ones pull a key — and the guess was wrong twice over: `group-late-joiner`, `group-kick-and-rejoin-keyshare` and `kv-store-joiner-cold` supply nothing, the `auto-follow-*` family I never checked supplies six, and two probes cut off before reaching the step. The measured numbers above replace it. Lowering the floor to match a future regression is still the one wrong answer, and the job's failure text says so.

### Why suite-wide and not per scenario

The same split as the projection gate and `projection-coverage`, for the reason that job states in its own comment: asserting a *negative* is satisfied just as well by a suite that checked nothing. Most scenarios never perform a direct pull at all — the delivery reaches them first — so demanding a marker from each one would fail the quiet ones. The systemic failure worth gating is a ground going unexercised **everywhere**.

To make a red gate cheap to diagnose, the job prints **per-scenario attribution on every run** (artifacts are merged into one directory and every log file is named after its scenario, so the filename is the attribution). When it fails, the first question is "which supplier stopped supplying?", and the diff in that list is the answer — without downloading an artifact.

## Test plan

No Rust changed, so `fmt`/`clippy`/`test` cannot be affected by this diff. What *can* be, and was run:

```
python3 scripts/check-scenario-coverage.py   # OK: 137/138 scenarios registered; 1 baselined
./scripts/check-naming.sh                    # exit 0
python3 -c "import yaml; yaml.safe_load(...)" # parses; 10 jobs; comment.needs updated
```

The assertion logic was validated by **extracting the step's own `run:` script out of the YAML** — so the tested text is the shipped text, not a retyped copy — and running it under `bash -e` against fixtures whose log lines carry the real ANSI escapes (`key_recovery_authorized_by<ESC>[0m<ESC>[2m=<ESC>[0m"anchor"`):

| case | expected | result |
| --- | --- | --- |
| both grounds present | pass | pass, both counted |
| only `anchor` | fail naming `own_account_device` | ✓ |
| only `own_account_device` | fail naming `anchor` | ✓ |
| logs present, marker absent | fail naming both | ✓ |
| an unrecognized third ground | pass **+ warning** | ✓ |
| matrix ran, no logs downloaded | fail (not a silent pass) | ✓ |
| matrix skipped / cancelled | pass with notice | ✓ |

That exercise earned its keep: it caught two `[ test ] && var=value` statements which, under the runner's `bash -e`, **exit the step when the test is false** — i.e. on the happy path, where a ground *is* present. They are now `if` blocks, with a comment saying why.

Also checked: `e2e-rust-apps-release.yml` carries none of this instrumentation family (no projection, op-store or key-recovery steps), so it needs no matching change and the PR is not widened to it.

CI on this head: **0 failed jobs of 30**, including all ~90 scenarios and the new gate.

## Proof the fix works

Not a bug fix — this closes an instrumentation gap — but the before/after is the point, so:

**Before.** A run where the own-account ground is inert prints `anchor` counts in the scenarios that use it, `no key_recovery_authorized_by markers` everywhere else, and the suite is green. That is the exact shape of #3892's two red rounds.

**After**, on that same input (fixture case "only `anchor`"):

```
::error::no group-key recovery anywhere in the suite was authorized by: own_account_device
(anchor=1, own_account_device=0) — unexercised, so a regression breaking it would leave this
suite green, which is exactly how #3892 shipped inert
```

## Wire contract (SDK gate)

No HTTP wire DTO, route, on-disk format or op encoding changed — this is a CI workflow only.

- [ ] Regenerated wire fixtures — not applicable
- [ ] Updated `crates/server/endpoints.json` — not applicable

## Documentation update

None required: the reasoning lives in the workflow comments beside the job, which is where the next person to see it go red will be reading. One thing an admin may need to do, which this PR cannot: `key-recovery-coverage` is a new job, so it reports through the `comment` job but will not *block* merges until it is added to the branch-protection required-checks list — the same caveat that applies to `projection-coverage`.

## Related

- #3896 — added the observe-only per-scenario report and deferred exactly this assertion.
- #3892 / #3888 — the two grounds, and the inert shipment this gate exists to catch.

### DIFF
--- a/.github/workflows/e2e-rust-apps.yml
+++ b/.github/workflows/e2e-rust-apps.yml
@@ -1297,11 +1297,12 @@ jobs:
           # collapsed back to anchor-only. Two red rounds and a security hole
           # came out of that, and the suite could not have told us.
           #
-          # So this reports the distribution rather than asserting one: nobody
-          # yet knows which ground each scenario legitimately uses, and pinning a
-          # guess is how a gate starts failing for the wrong reason. Observe
-          # first, assert once the distribution is known -- the same way the
-          # projection and op-store gates were built. INFORMATIONAL, never fails.
+          # This step stays INFORMATIONAL and never fails, deliberately: most
+          # scenarios never perform a direct pull at all, because the delivery
+          # reaches them first, so asserting a marker HERE would fail the quiet
+          # ones. The floor now sits in `key-recovery-coverage`, which pools the
+          # whole suite and fails if either ground went unexercised everywhere --
+          # the same split as the projection gate and `projection-coverage`.
           #
           # ANSI escapes sit between a field name and its `=` in node logs, so
           # the sed is not optional; matching the raw pattern finds nothing.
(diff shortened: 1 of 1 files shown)


chefsale:
**Correcting the caveat in the PR body: the floor is not thin.** The gate ran on this PR and reported the real suite-wide numbers:

```
25 anchor
 6 own_account_device
both key-serve grounds exercised across the suite (anchor=25, own_account_device=6)
```

spread across **~20 distinct scenarios for `anchor`** and **4 for `own_account_device`**:

| ground | suppliers |
| --- | --- |
| `own_account_device` (6) | `shared-storage-account-writers-two-devices` ×2, `account-device-revoke-lockout` ×2, `account-pairing-missed-publish`, `account-facts-cross-namespace` |
| `anchor` (25) | `auto-follow-*` (4 scenarios, 6 recoveries), `group-leave-member`/`-namespace`/`-context`, `group-private-add-then-write`, `group-kick-and-readd-deny-list` ×3, `group-membership`, `group-metadata`, `group-subgroup`, `group-subgroup-queued-deltas`, `group-subgroup-cold-sync`, `group-list-subgroups-restricted-hidden`, `dm-subgroup-privacy`, `account-pairing-application-scope` |

So please disregard "the floor rests on few suppliers" and the flake risk built on it. My PR body said I sampled 6 of 88 scenarios and found one supplier per ground — that sample was simply unrepresentative: I picked scenarios by guessing which ones would pull a key, and guessed wrong (`group-late-joiner`, `group-kick-and-rejoin-keyshare`, `kv-store-joiner-cold` supply nothing, while the `auto-follow-*` family, which I never checked, supplies six). Two of my probes also cut off before reaching the step, so I under-counted twice over.

What this means for review:

- **Losing a ground now takes ~20 scenarios going quiet at once, not one.** That is not a plausible timing coincidence — it is what a real regression looks like, which is exactly what the gate should fire on.
- The `own_account_device` floor is the narrower of the two at 4 suppliers, still comfortably above one, and its scenarios are structural rather than racy (a newly paired device has no delivery addressed to it, so pulling from its own account's other device is the only route).
- The per-scenario attribution above is printed by the job on **every** run, so if this ever does go red, the diff in that supplier list is the diagnosis.

The rest of the PR body stands. CI is green on this head, including all ~90 scenarios and the new job.

---

Issue #3904, opened 2026-09-10
A joiner that holds no covering key publishes its join in the clear


chefsale:
> [!IMPORTANT]
> **This body is corrected below and was substantially wrong when filed. Implemented by #3910.**
>
> I wrote it, then found the premise stale while starting the work. The corrections are inline as `⚠️ CORRECTION` notes and in [this comment](https://github.com/calimero-network/core/issues/3904#issuecomment-5621951249); the original text is kept as the record rather than quietly rewritten. Two claims below are **false** and the design question the issue builds toward is **already answered**:
>
> 1. *"`MemberJoined` needs **no** endorsement at apply"* — **#3804 changed this.** `join_group` now hard-fails without an admitter endorsement. Reaching an admitter is a requirement, not a hint.
> 2. *"holds the namespace key by definition"* — **false.** `handle_namespace_join_request` endorses a join and serves an **empty** key envelope when it holds no key itself, which is one of the ways a joiner ends up unkeyed in the first place.
>
> Consequently the **fail-closed vs fall-back decision this issue asks the team to make does not exist**: routing through the admitter adds no availability dependency, because a join that reached no admitter has already failed. And most of what is described as missing was closed by #3857 / #3858 / #3859 *before* this was filed. The remaining scope is one relay exchange, which is what #3910 does.
>
> **Retitled** from *"Invitation-based joins publish membership in the clear, though the sealed relay path already exists"* — the old title implied all invitation joins, when only the unkeyed case was affected.

---

## Summary

`RootOp::MemberJoined` and `RootOp::MemberJoinedAt` are published as cleartext `NamespaceOp::Root` on the namespace gossip topic (`root_op_is_sealable` returns `false` for both, `crates/governance-types/src/lib.rs:888`). Every other root op is now sealed, so these two are the remaining membership metadata in the clear.

[...]

Issue #3859, opened 2026-09-08
An invitation into an Open-chain subgroup admits a joiner that can be served no usable key, silently


chefsale:
## Summary

An invitation may target any group that resolves into the namespace, including a subgroup: `MembershipRepository::validate_open_invitation` (`crates/governance-store/src/namespace/membership.rs:194`) checks the inviter signature, inviter permission, expiry, and that `NamespaceRepository::resolve(&group_id)` equals this namespace — there is no root-only requirement. When the targeted subgroup sits on an **Open chain to the namespace**, the joiner is admitted as a direct member and then holds no key that decrypts anything in that subgroup.

**The subgroup's own key exists but is never the one in use.** This is the subtlety that makes the bug hard to see:

- A key row **is** minted for every subgroup at creation, unconditionally — visibility is not consulted. Both creation paths do it: `crates/server/src/admin/handlers/namespaces/create_group_in_namespace.rs:112` and `crates/context/src/handlers/create_group.rs:275`.
- But **nothing encrypts to it** when the chain to the namespace is Open. Every publisher picks the encrypting group with the same predicate, `CapabilitiesRepository::is_open_chain_to_namespace` (`crates/governance-store/src/capabilities.rs:147`) — `true` ⇒ the **namespace** key:
  - state deltas: `crates/context/src/handlers/execute/mod.rs:414`
  - governance ops: `crates/governance-store/src/group_governance_publisher.rs:359`
  - rotation eligibility: `crates/governance-store/src/pending_rotation.rs:75`
- And it is **never distributed**: the apply handler `ops/namespace/group_created.rs` mints no keyring, so any peer that merely *received* `GroupCreated` holds no row for that subgroup at all.

So the minted subgroup key is dead weight on the creating node: real for the keyring, unused by every writer.

Note the predicate is the **whole chain**, not one hop. An Open subgroup beneath a *Restricted* parent gets `false` and therefore uses its own key — that case is fine and out of scope. `group_id == namespace_id` also returns `false`, so the root always uses its own key. This issue is only about a subgroup whose every ancestor up to the namespace is Open.

[...]

Issue #3858, opened 2026-09-08
Seal a Restricted-subgroup invitation join under the subgroup key — the last cleartext governance op


chefsale:
## Summary

After #3850, #3856 and #3857 every governance root op travels sealed except one: a `MemberJoined` / `MemberJoinedAt` published by a joiner whose invitation targets a **subgroup** rather than the namespace root. `NamespaceRepository::resolve` (`crates/governance-store/src/namespace/core.rs:453`) walks parents to the root, so an invitation's `group_id` may be a subgroup; `join_group` then stores the key the join bundle carried for *that* group (`crates/context/src/handlers/join_group.rs:264`) — the subgroup key, not the namespace key. Sealing a root op needs the namespace key, so this joiner publishes its membership in the clear, and `root_op_is_sealable` (`crates/governance-types/src/lib.rs:774`) still cannot be flipped to a hard refuse for these two variants.

This issue is scoped to a subgroup on a **Restricted chain**, whose own key is genuinely the one in use. An invitation targeting a subgroup on an **Open chain to the namespace** is a separate, functional (not privacy) defect: everything in such a subgroup is encrypted under the *namespace* key (`CapabilitiesRepository::is_open_chain_to_namespace`, consulted by every publisher), so the joiner is admitted and can decrypt nothing — silently, because the responder may still serve it the unused key row the subgroup was created with. Filed as #3859.

## Decision

**Option 3: give a subgroup-scoped invitation join its own wire variant, sealed under the subgroup key.** The two alternatives are recorded below as dead, so nobody re-derives them.

The design principle this follows, stated plainly: *a member of only a subgroup must not be able to read namespace-level data.* That principle is already enforced everywhere the namespace key can be distributed, and this issue must not become the exception that breaks it:

- **Pull path** — the key-delivery pull in `NamespaceGovernance` gates on `MembershipRepository::new(self.store).is_member(&group_gid, &account)` for the *requested* group and returns an empty key list when the requester is not a member of it (`crates/governance-store/src/namespace/governance.rs`, the pull arm). A subgroup-only member asking for the namespace key gets nothing. This is easy to misread: `member_account_in_namespace` (`crates/governance-store/src/account_bindings.rs:209`) resolves only a *device binding* and does not itself check group membership — the membership gate lives in the caller.
[...]

PR #3910, opened 2026-09-10
feat(governance): relay a join its author cannot seal instead of publishing it in the clear

chefsale:
## Description

Closes the last cleartext membership op on the namespace topic. Refs #3904 —
whose body was substantially stale and is now corrected inline there.

`RootOp::MemberJoined` / `MemberJoinedAt` name an account, the group it joined
and when. A joiner that holds the key covering its invitation's group already
seals its own join (#3857 for the namespace root, #3858/#3859 for the covering
group). A joiner that holds **no** such key could not, and the answer was to
publish unsealed — putting exactly the membership graph the sealing series
(#3847, #3850, #3856–#3859) exists to hide in front of every non-member
subscribed to the topic. Topic subscription is not authorization-gated, so
knowing a namespace id was enough to watch a private group's membership
assemble in real time.

Refusing outright is not available either, and this is the part that shaped the
design: the key such a joiner lacks arrives in a `KeyDelivery` an admin
publishes *on seeing that very op*. There is no order in which it could seal
for itself. That is the cold-start trap recorded in
`crates/governance-store/AGENTS.md` — a cold-start path cannot be gated on
folded state, because the state is behind the key it is trying to obtain — and
four attempts across #3888/#3891 were reverted for that same shape.

**So route the publish through someone who can seal.** `join_group` now signs
and applies the join locally *without publishing*, then hands the signed op to
a namespace keyholder over a new exchange on the existing namespace-join
stream. The pre-existing `relay_signed_join` handler seals it as
`NamespaceOp::RootRelaySealed` and publishes under its own envelope signature.

Three properties make that safe, and none of them is new — `RootRelaySealed`
and its apply path already shipped for the no-node keyholder case:

- **The relayer wraps; it never re-authors.** The envelope carries the joiner's
  whole `SignedNamespaceOp`, and `open_relayed_join` verifies that inner
  signature after decrypting. The outer signature proves only who carried it.
  `join_op_proves_ownership` then requires `signer == account.statement.sign_pk`,
  so a relayer that substituted a member produces an op every peer rejects.
- **Keyless verification survives.** The outer envelope is an ordinary
  signature over the ciphertext, so a non-member still authenticates the
  skeleton it stores without holding a key — it simply learns nothing about who
  joined.
- **No new availability requirement.** Since #3804 a join with no admitter
  endorsement fails outright, so the first peer asked is one the joiner has
  *already* completed an exchange with.

### One thing worth reviewing closely

An admitter is **not** guaranteed to hold the key. `handle_namespace_join_request`
will endorse a join and serve an *empty* key envelope when it holds none itself
— which is one of the ways a joiner ends up unkeyed in the first place. So the
initiator tries the endorsing admitter first (it is the one peer known
reachable: the endorsement arrived over a stream to it) and then falls through
to the rest of the namespace topic. Widening past the admitter is sound because
a relayer supplies a seal and an envelope signature, never authority — the
endorsement lives inside the op and is self-authenticating. For the same
reason the responder deliberately runs **no** membership or authority check on
the requester; a gate there would only reject legitimate relays.

**A relay that finds no keyholder fails the join.** There is no cleartext
fallback, because a fallback is defeatable by exactly the adversary it defends
against — anyone who can keep the joiner from reaching a keyholder forces the
leak — and per `relay_signed_join`'s own reasoning a receiver cannot tell that
fallback from a publisher that skipped sealing deliberately.

### Scope, and why it is smaller than #3904 originally said

#3904's body was stale in two load-bearing ways. It claimed (1)
*"`MemberJoined` needs no endorsement at apply"* — #3804 changed that, which is
precisely why this adds no availability dependency; and (2) that a responding
node *"holds the namespace key by definition"* — it does not, per the
empty-envelope path above. Most of what the issue described as missing was
closed by #3857/#3858/#3859 before it was filed. The issue body now carries
those corrections inline.

## Test plan

```
cargo fmt --check                                                        ok
cargo clippy --workspace --all-targets \
  --features calimero-storage/testing -- -D warnings                     ok
cargo clippy -p merod -p calimero-node -p calimero-server \
  -p calimero-tee-attestation --all-targets --features mock-attestation  ok
./scripts/check-naming.sh                                                ok
python3 scripts/check-scenario-coverage.py                               ok
cargo test -p calimero-node                          296 passed; 0 failed
cargo test -p calimero-context                       591 passed; 0 failed
cargo test -p calimero-server                        162 passed; 0 failed
cargo test -p calimero-governance-store              702 passed; 0 failed
cargo test -p calimero-node-primitives               322 passed; 0 failed
cargo test -p calimero-context-client                 37 passed; 0 failed
```

Six new unit tests. Five pin the responder gate in
`sync/manager/relay_sealed_join.rs`:

```
relay_sealed_join::tests::refuses_an_op_signed_for_another_namespace       ok
relay_sealed_join::tests::refuses_a_join_with_no_endorsement               ok
relay_sealed_join::tests::refuses_a_join_whose_signature_does_not_verify   ok
relay_sealed_join::tests::an_admitter_cannot_substitute_the_member         ok
relay_sealed_join::tests::refuses_a_root_op_that_is_not_a_join             ok
```

`an_admitter_cannot_substitute_the_member` is the one #3904 asks for: it
re-points the join at a different account and asserts the joiner's signature no
longer verifies, so wrapping cannot change who joined. The ops are built
through `SignedNamespaceOp::sign` rather than by filling the struct, so a
hand-written signature cannot make them pass for the wrong reason.

One in governance-store —
`signing_without_publishing_still_applies_and_feeds_the_local_path` — pins that
the new sign-and-apply-only helper still fires the readiness notification and
the local-apply feed. The local apply is not optional: a later
`MemberJoinedOpen` causally parents onto this op.

The `init_pop_gate_tests` pair pins the new payload's proof-of-possession
classification, so a future edit that moves it in or out fails a unit test.

The "non-member authenticates the outer envelope keyless" half of #3904's
criterion was already covered by the existing
`a_relayed_join_is_authenticated_by_the_joiner_not_the_admitter` and the
park/retry and re-seal-replay tests in `governance-store/src/namespace/tests.rs`.

### New merobox scenario

`apps/scaffolding-e2e/workflows/group-join-never-published-in-the-clear.yml`,
registered in the e2e matrix with `registry_fetch: "false"` (governance only).

It asserts on the `op_kind` of governance-store's "namespace governance op
published" line, because that label comes from `NamespaceOp::op_kind_label` —
the wire variant actually handed to the publisher. `member_joined` /
`member_joined_at` are emitted only for a cleartext
`NamespaceOp::Root(RootOp::MemberJoined*)`, so their absence is a direct test of
the wire form. It is deliberately **not** asserted on the log line this PR
deletes: that string is gone, so an absence check on it would pass while
testing nothing.

Three things I got wrong writing it, each of which would have shipped a gate
that passes for the wrong reason — worth knowing if you write one of these:

- **The patterns match the quoted value, not `op_kind="..."`.** A raw node log
  carries ANSI escapes *between* a field name and its value, so the fuller
  pattern matches nothing — and for `assert_log_absent` that is a silently
  vacuous pass, not a loud failure. The projection-coverage gate in
  `e2e-rust-apps.yml` hit this and now strips escapes before grepping; merobox
  greps container output as-is.
- **The positive assertion belongs on the joiner, not the admitter.** These
  labels exist on the publish path only — a *receiving* node logs `add` /
  `join` / `remove` from `MembershipOpKind::label` — so asserting on node-1
  would have matched its own `GroupCreated` / `KeyDelivery` publishes and never
  a join at all.
- **The convergence check compares an account**, since the member list is
  `Vec<AccountId>`; `memberIdentity` is a signing key and would never match.

### What is still NOT covered, stated plainly

- **The relay path itself has no scenario.** Reaching it needs an admitter that
  endorses a join while holding no covering key, and merobox has no declarative
  way to produce a keyless-but-endorsing responder. `root_relay_sealed` is
  therefore not asserted present; what the scenario asserts is the property
  that must hold on every path including the relay — no join is ever in the
  clear. The relay branch's own gate is the five unit tests.
- **The subgroup-targeted invitation join is not exercised.** The invite
  endpoint refuses a non-root group (*"namespace_id must reference a root group
  (namespace)"*), so it is reachable only via the `recursive` flag, which the
  merobox step does not expose. An admin-push `add_group_members` is no
  substitute — that publishes a `MemberAdded` GroupOp, not a join.
- **No test of the decision itself.** The seal-locally-vs-relay branch lives in
  `join_group`, and the repo has no test harness for that handler — the
  in-process governance e2e never calls it. Building one is real work and is not
  attempted here.

## Proof the fix works

This is a confidentiality fix rather than a crash, so the "before" is a code
path and a log line rather than a failing test.

**Reproduction.** Run any two-node join where the responder serves no key
(e.g. an admitter still awaiting its own `KeyDelivery`) and read a non-member's
view of the `ns/<hex>` topic.

**Before** — `join_group.rs:612` fell through to a cleartext publish,
announcing it:

```rust
Ok(None) => {
    info!(?group_id, "publishing this join in the clear: no namespace key held
           yet, so it cannot be sealed and the key it needs arrives in answer to it");
    NamespaceOp::Root(join_root)
}
```

A non-member on the topic reads `Root(MemberJoinedAt)` — joiner account, group
id, join time, causal position.

**After** — that arm is gone (`git grep -c "publishing this join in the clear"`
→ 0). The same case now logs

```
no key to seal this join under; relaying it to the admitter to be sealed and published
```

and a non-member reads `RootRelaySealed { key_id, encrypted }`. When no peer
will relay, the join fails with a message naming why rather than publishing:

```
join could not be published: this node holds no key to seal it under and no
admitter would relay it, and publishing it in the clear would disclose the membership
```

so "sealed" and "leaked" are never the same silence — #3904's observability
criterion, and the fail-open shape #3902 closed for group-key recovery.

**Regression locked in by:** the five responder-gate unit tests, of which
`an_admitter_cannot_substitute_the_member` is the security-relevant one, plus
the new scenario's `"member_joined"` / `"member_joined_at"` absence assertions
on both nodes.

## Wire contract (SDK gate)

Not applicable — no HTTP DTO and no route changed, so no fixture or manifest
regeneration and no paired SDK PR. The change is to the **P2P sync wire**
(borsh), not the server wire.

- [ ] Linked the matching mero-js PR — n/a, no server-wire change

**Mixed-version note, since merobox cannot catch it.** Both new variants are
appended at their enum tails, so every existing discriminant is unchanged and
no currently-encodable message changes bytes. Borsh rejects an unknown
discriminant outright, so an older responder cannot misread the request — it
drops the stream, and the joiner then *fails* rather than falling back. During a
rolling upgrade a new joiner that reaches only pre-relay admitters will fail its
join instead of leaking it. That is the intended direction, but operators
upgrading a fleet should know the failure mode exists.

## Documentation update

Done in this PR:

- `docs/.../protocol/encryption.mdx` — the cleartext-set table row, the
  subgroup-seal section and the metadata-exposure list no longer describe a
  cleartext fallback.
- `docs/.../protocol/direct-admission.mdx` — the node-joiner flow now says the
  join is published sealed either way, and why the unkeyed case can route
  through the admitter without a new dependency.
- `crates/governance-store/AGENTS.md` — the relayed-join bullet records that the
  relay is no longer only for a keyholder with no node.
- `crates/node/AGENTS.md` — a gotcha bullet for the new module, including the
  admitter-may-be-keyless trap.
- `crates/governance-types/src/lib.rs` — `root_op_is_sealable`'s comment for
  these two variants, which #3904 correctly called actively misleading. It
  ended on "hand it to the admitter does not work"; `RootRelaySealed`
  post-dates that reasoning and answers it. The arm still returns `false`, and
  the rewrite explains why that is now load-bearing: the joiner's inner op is a
  cleartext `Root(..)` by construction, so flipping it would make
  `refuse_unsealed_sealable_root` refuse the joiner's own local apply and break
  the route at its first step.
- `crates/context/src/test_support.rs` — `published_join`'s doc, which claimed
  to return "the wire form production publishes for a JOIN".

Not needed: no README or public API surface changed.


chefsale:
## `Windows node smoke` is red, and it is not this PR's — with the root cause and a patch

**Failing check:** `Windows node smoke` → `merod::watchdog_live::the_node_keeps_running_while_its_data_directory_is_intact` ([job](https://github.com/calimero-network/core/actions/runs/34543386628/job/103090710322)). The other two tests in that binary passed.

```
failed to listen on '/ip4/0.0.0.0/udp/58487/quic-v1'
An attempt was made to access a socket in a way forbidden by its access permissions. (os error 10013)
  at crates\network\src\behaviour.rs:385
```

### Why it is not this PR's

The same check **passed on this PR's previous head** `84865cd4b` (check run `103075237340`, ~70 minutes earlier). The only commit added since is the merge of `master`, and `8cca1a2da` (#3908) touches `.github/fleet.json` and `.github/scripts/bump-fleet.sh` — **no Rust at all**. So byte-identical Rust passed and then failed.

This PR's diff is namespace-join relay logic and governance sealing; it does not touch UDP binding, the swarm transport, or `merod`'s startup. Nothing in it can reach a QUIC listen call.

### Root cause, which is a real pre-existing bug in the test helper

`crates/merod/tests/watchdog_live.rs:64`:

```rust
/// Asked of the OS rather than hardcoded, so parallel jobs on one runner cannot
/// collide on a port.
fn free_port() -> u16 {
    TcpListener::bind("127.0.0.1:0").expect("bind").local_addr().expect("addr").port()
}
```

The port is discovered by binding **TCP**, then handed to `--swarm-port` (line 113), which merod also listens on for **UDP/QUIC**. A port being free for TCP says nothing about UDP. On Windows, a UDP bind inside a Hyper-V / WinNAT *excluded port range* fails with exactly `WSAEACCES` (10013) — those ranges are reserved dynamically, so which ports are poisoned varies per boot. That is why this is intermittent and Windows-only, and why the comment's "cannot collide on a port" reasoning does not hold for the QUIC listener.

It is also latent for every other test using `free_port()` for a swarm port, not just this one.

### Proposed patch — deliberately not applied here

A Windows test-harness fix is unrelated to this diff, so pushing it would widen the PR. Happy to open it separately, or for whoever owns the Windows lane to take it:

```rust
/// Asked of the OS rather than hardcoded, so parallel jobs on one runner cannot
/// collide on a port — and probed for UDP as well as TCP, because a swarm port
/// carries QUIC too. On Windows a port inside a Hyper-V excluded range binds
/// for TCP and then fails the UDP bind with WSAEACCES (10013), which surfaces
/// as a node that never logs readiness.
fn free_port() -> u16 {
    for _ in 0..32 {
        let port = TcpListener::bind("127.0.0.1:0")
            .expect("bind tcp")
            .local_addr()
            .expect("addr")
            .port();
        if UdpSocket::bind(("127.0.0.1", port)).is_ok() {
            return port;
        }
    }
    panic!("no port was bindable for both TCP and UDP after 32 attempts");
}
```

(There is an inherent race between probing and merod binding, which the retry does not remove — but it converts a hard failure into a vanishingly unlikely one, and the excluded-range case it targets is deterministic per port rather than racy.)

### What I did

I attempted the one permitted re-run of the failed job and got `403 This workflow is already running` — the run still has jobs in flight. I will spend that re-run once it completes, and I am keeping this PR watched until it is green rather than leaving it here. If the re-run reproduces the failure on a *different* random port, that would contradict the excluded-range explanation and I will re-diagnose rather than call it a flake again.

Everything else on this head is green, including the new merobox scenario (`Test (group-join-never-published-in-the-clear)`), the full `Rust` job, `E2E - Rust Apps` (106 jobs), `SDK E2E (paired)`, `Sync regression` and `App migration e2e`.


---

chefsale:
Re-run spent, and it passed — `Windows node smoke` is green on attempt 2 ([job](https://github.com/calimero-network/core/actions/runs/34543386628/job/103098497942)), same commit, different port draw. That is the outcome the excluded-range diagnosis predicted, so I am not re-diagnosing.

**All 12 workflows are now green on `254133d3e`**: CI (incl. the full `Rust` job — test, clippy, both `mock-attestation` passes, `cargo deny`, `cargo machete`, scenario coverage), E2E - Rust Apps, SDK E2E (paired), Sync regression, App migration e2e, ABI Contract, ABI Guard, Docs CI, Release, PR Title.

The `free_port()` TCP-probe-for-a-UDP-port bug above is still real and still unfixed — it will bite another PR on another random port. Left out of this diff on purpose; worth its own change.

This is ready for review.


---

PR #3912, opened 2026-09-11
refactor(runtime): delete the fetch host function

### LINKED ISSUE #1694: Add whitelist to `fetch` host function
Add whitelist of the allowed host to be called by the `fetch()` host function. Potentially, allow this whitelist to be specified at the constructor of the app and/or in the app manifesto.

chefsale:
## Description

Closes #1694 by removing `fetch` rather than adding the allowlist that issue proposed.

`fetch` has been hard-disabled since it was introduced. The impl opened with `const BLOCKED: bool = true` and short-circuited every call to `"Host function 'fetch' is temporarily disabled."`, which left the ~100 lines of `ureq` request code below it unreachable. Nothing in the tree called it, the SDK wrapper was `#[doc(hidden)]`, and the docs site never listed it among the host functions.

**Why removal instead of the allowlist.** As the [design comment on #1694](https://github.com/calimero-network/core/issues/1694#issuecomment-4582139606) works out in detail, an allowlist does not address the actual blocker: `fetch` is non-deterministic network I/O inside a replicated, synced state machine, so a result feeding committed state diverges nodes. That is almost certainly *why* it was hard-disabled in the first place. Plumbing a `FetchPolicy` end-to-end would add config surface, SSRF hardening and redirect handling to a function that still could not be enabled at the end of it. Reintroducing HTTP access needs the determinism question settled first, and a fresh design with it — so the dead code goes now and the design starts clean when there is a decision to build on.

What comes out:

| Area | Removed |
| --- | --- |
| `crates/runtime/src/logic/host_functions/utility.rs` | the `fetch` impl and its `test_fetch_is_blocked` test, plus the `borsh::from_slice` / `std::io::Read` imports that only it used |
| `crates/runtime/src/logic/imports.rs` | the host-import registration |
| `crates/sys/src/lib.rs` | the `wasm_imports!` declaration |
| `crates/sdk/src/env/ext.rs` | the whole file — it was only the `fetch` wrapper — and the `pub mod ext;` that exposed it |
| `crates/runtime/examples/fetch.rs` | existed only to demo it |
| `crates/runtime/Cargo.toml` + `Cargo.lock` | `ureq`, whose only use in the crate was this function |

Net 333 deletions, 6 insertions. No dependency changes beyond dropping `ureq` from `calimero-runtime` (it stays in the workspace; `cargo-mero` still uses it).

### One compatibility note for reviewers

A blob that *imports* `env.fetch` will now fail to link at `Instance::new` instead of erroring at call time. Such a blob could never have gotten a successful fetch — it only ever received the "temporarily disabled" failure — so nothing working regresses. I did **not** leave a compat shim, though `imports.rs` keeps one for `executor_id` in a similar pre-split case. Happy to add the equivalent alias if reviewers would rather have belt-and-braces here; it is a one-line addition.

## Test plan

Not a bug fix and not a behavior change — the function returned failure before and no longer exists, so there is nothing new to assert and no regression test to add. The removed `test_fetch_is_blocked` only pinned the disabled stub's error message.

Verification is that the tree still builds and lints with the surface gone. Run locally, all passing:

```bash
cargo fmt --check
cargo check --workspace
```

| Check | Result |
| --- | --- |
| `cargo fmt --check` | pass |
| `cargo check --workspace` | pass, 0 errors |
| `cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings` | pass, 0 errors |
| `cargo check -p calimero-runtime -p calimero-sys -p calimero-sdk` | pass |

Clippy under `-D warnings` is the check that matters most here: it is what would catch the `borsh::from_slice` / `std::io::Read` imports left stranded by deleting the function body.

Two things I could **not** run in my sandbox, deferred to CI:

- **`cargo test` / the full `./scripts/check-like-ci.py`.** The `mero-auth` and `calimero-server` build scripts fetch frontend archives from GitHub, which fails TLS behind my proxy (`invalid peer certificate: UnknownIssuer`) — the exact limitation `check-like-ci.py` warns about. The checks above were run with `CALIMERO_AUTH_FRONTEND_SRC` / `CALIMERO_WEBUI_SRC` pointed at a local stub, per that script's own advice.
- **`cargo machete`** (not installed locally). This is the step that would flag a stale `ureq` had I missed it, so it is worth a glance on the CI run.

CI is authoritative for both; I will push a fixup if either reddens.

## Proof the fix works

N/A — this is a dead-code removal, not a bug fix.

## Wire contract (SDK gate)

No HTTP wire DTO or route changes. This touches the WASM host-import ABI, not the HTTP surface, so the wire-contract gate does not apply.

- [ ] Regenerated wire fixtures — not applicable, no DTO changed
- [ ] Updated `crates/server/endpoints.json` — not applicable, no routes changed

Worth a second opinion on: the mero-js SDK does not bind host imports (those are guest-side, reached through `calimero-sdk`), so I do not believe a paired SDK change is needed. Flag it if the JS app runtime surfaces `fetch` somewhere I did not find.

## Documentation update

Done in this PR, no follow-up needed:

- `crates/runtime/HOST_FUNCTIONS.md` — dropped the `fetch` row from the Utility catalog
- `crates/runtime/AGENTS.md` — `utility.rs` description and the `examples/` tree
- `crates/sys/AGENTS.md` — the host-service grouping ("Network/misc" → "Misc", since `fetch` was the only network entry)
- `crates/sdk/AGENTS.md` — the `src/env/` file tree, which listed only the deleted `ext.rs`; corrected to the files actually there (`host.rs`, `subscriber.rs`)

The `docs/` site needed no change — it never listed `fetch` in `protocol/execution.mdx` or `build/host-functions.mdx`.

https://claude.ai/code/session_012qEJBiVwyjzGYSCbKWe97F


chefsale:
### `Windows e2e (3-node sync, binary mode)` failed — not this PR's change

The failing step is `calimero-server`'s build script downloading a release asset, and it got a server-side HTTP 500:

```
failed to download https://github.com/calimero-network/admin-dashboard/releases/latest/download/admin-dashboard-build.zip
  HTTP status server error (500 Internal Server Error) for url
  (https://github.com/calimero-network/admin-dashboard/releases/download/v1.13.1/admin-dashboard-build.zip)

Location: crates\build-utils\src\fetch.rs:125:10
  11: calimero_build_utils::fetch::fetch_and_extract
  12: build_script_build::try_main::{closure#1}
```

Why this isn't attributable to the diff:

- The failure is in `crates/server/build.rs` → `crates/build-utils/src/fetch.rs`. **This PR touches neither** `crates/server/**` nor `crates/build-utils/**` — the diff is confined to `crates/runtime`, `crates/sys`, `crates/sdk` and their docs.
- **`Windows e2e (auth seam, binary mode)`** — the sibling Windows job, same commit, same workflow, same build — **passed**. So the Windows build of this change is fine; the two jobs differ only in which scenario they run afterwards.
- **`Windows node smoke`** passed.
- Every Linux job passed, including **`Rust`** (clippy `-D warnings`, `cargo test`, `mock-attestation`, `cargo machete`), **`Build Apps`**, **`Build WASM fixtures`**, **`WASM size gate`**, **`cargo-mero e2e`**, **`ABI Validation`**, and all ~40 migration scenarios plus the full merobox e2e suite.

An HTTP 500 from GitHub's release CDN is upstream infrastructure, and `fetch_and_extract` already exhausted its retries against it. There is no fix to port — nothing in this repo's code causes or can avoid a 500 on that host beyond the retry logic already there.

I have re-run the failed job once to confirm that reading. If it fails again with the same 500 I will say so here rather than treating it as flake a second time; if it fails for any *other* reason, that is a real signal and I will root-cause it.

---

PR #3939, opened 2026-09-15
fix(abi): regenerate the resolved conformance golden for derived intents

chefsale:
**Fixes ABI Guard on master.** Regression from my own #3936 — apologies for the red.

## What broke

#3936 made `MethodIntent` derivable from a method's receiver, so every method taking `&self` or `&mut self` now emits an `intent` field in the ABI.

Two golden files pin that output. I regenerated `apps/abi_conformance/abi.expected.json` and **missed its sibling** `apps/abi_conformance_resolved/abi.expected.json`. `scripts/verify-abi.sh` diffs the emitted ABI against both, so ABI Guard has been red on master since 09c77e9.

Reproduced before fixing:

```diff
--- apps/abi_conformance_resolved/abi.expected.json
+++ extracted from the built wasm
       "returns": { "kind": "string" }
-      }
+      },
+      "intent": "read_only"      # describe
...
+      "intent": "mutating"       # join
+      "intent": "mutating"       # record
+      "intent": "read_only"      # tally
```

## The fix

Regenerated through the same path the check uses — `cargo mero build`, then `mero-abi extract` off the embedded `calimero_abi_v1` section — rather than hand-edited. Four `intent` fields added; nothing else in the file changes.

## Why this one was missable, and why it should not have been

The second golden is deliberately not a copy of the first. The comment above its `verify_golden` call says so:

> Aliases, macro-generated and re-exported types: only the compiler resolves these, so this golden is what proves they are described correctly.

Which is exactly why it has to move in lockstep with any change to what the macro emits — the case for its existence is the case for regenerating it. I checked the diff's own files rather than searching for every golden the emitted ABI feeds; `git ls-files | grep abi.*expected` would have found it in one step.

## Validation

Ran the three steps `abi-guard.yml` runs, in its order:

```
./scripts/verify-abi.sh                                    ABI verify: OK
./apps/state-schema-conformance/verify-state-schema.sh     === All tests passed! ===
cargo test -p calimero-wasm-abi --test invariants          9 passed
```

`verify-abi.sh` covers both goldens plus the jq spot checks (`view_constant` → `read_only`, `xcall_noop` → `xcall_callable`, nullable params) and the identity-downgrade self-diff.

---


chefsale:
Still required after #3938 merged — updating in case that looks like it superseded this.

I had merged master into #3938 and ported this same golden fix onto it, so that its merge with master wouldn't go red from the same cause. The squash took only #3938's original commit (20 files / 1854 insertions), not the merged head that carried the port. So master at 424c3a4 has the derived-intent macro and the auth provider, but **not** the regenerated golden:

```
$ git show origin/master:apps/abi_conformance_resolved/abi.expected.json | grep -c '"intent"'
0
```

ABI Guard is therefore still red on master, and this is the fix.

Rebased onto current master and re-validated there, rather than relying on the earlier run against 09c77e9:

```
./scripts/verify-abi.sh     ABI verify: OK
```

---

Issue #3930, opened 2026-09-15
Add an account-proof auth provider so a device key can obtain a session


chefsale:
## Summary

`mero-auth` implements exactly one authentication provider — `crates/auth/src/providers/impls/user_password.rs` — so the only way to obtain a session on a node is a username and password. An account holder who runs no node has no credential there, and per-relay provisioning is not an option on a machine that deliberately serves several tenants.

Everything in the delegated-execution client past "post a warrant" needs a session. This is the largest item in #3928 and the one the other work depends on.

## Impact

**Blocks the entire client beyond writes** — reads, app grants and request proofs all need a session first.

It is also becoming urgent rather than merely wanted: `user_password` is already switched off on fleet relay images (calimero-network/mero-tee, pending release), because a Calimero-operated multi-tenant relay should not carry interactive logins. Once that ships there is **no** authentication path into a relay at all. This issue is what replaces it.

## Scope

**A challenge endpoint.** Issues a challenge the client signs. Stateless is preferable — a MAC over node id, nonce and expiry — with a short lifetime and a spent-set so a challenge is single-use. Exact shape comes from #3929.

**A provider in `crates/auth/src/providers/impls/`.** Given an `AccountProof<DeviceCert>` and a signature over the challenge, it verifies the device belongs to the account and the signature is that device's, then mints a token whose subject is the **account**.

`providers` is a `HashMap<String, bool>` in `AuthConfig`, so registering a new provider is additive and nothing that exists changes behaviour. Note that naming an unregistered provider in config parses and silently matches nothing — the new provider must be registered in the provider registry, not merely named in a config file.

**Propagation.** The account must reach the handlers. Traefik's `auth-node` forwardAuth already propagates `X-Auth-User` and `X-Auth-Permissions` from `/auth/validate`, so the account id should travel in `X-Auth-User`. #3930 consumes it.

## Out of scope

- App grants and scoped sessions. A first cut may issue a session for the account with membership-derived scope; grants are a later refinement.
- Per-request proof signatures, if #3929 defers them.

## Criteria for resolving

- [ ] A client holding an account root and a device key obtains a session against a node it does not own, with no password involved.
- [ ] A signature from a device that is not certified by the named account is refused.
- [ ] A challenge is single-use and expires; a replayed challenge is refused.
- [ ] The account id reaches request handlers via `X-Auth-User`.
[...]

Issue #3928, opened 2026-09-15
Epic: unblock the delegated-execution client — key login, reads, warrant v2


chefsale:
## Summary

A client for delegated execution is being built against the node, and the node is missing the pieces it needs. Writes already work end to end; **everything past "post a warrant" is blocked**. This epic is the node-side work that unblocks that client, in the order that unblocks the most per unit of our effort.

Scope is deliberately narrow: what a client cannot proceed without. Related work that matters but does **not** block them is listed at the bottom so we do not gold-plate while they wait.

## What already works — build against it now

Verified against `e0210ab`. None of this needs doing:

| Capability | Where |
|---|---|
| Delegated write, end to end | `POST /admin-api/contexts/:id/intents` accepts `Warrant` + `AccountProof<DeviceCert>` |
| Warrant-minting metadata | `GET /admin-api/contexts/:id/intents` → `executorAccount`, `canAuthorOnBehalf`, `grantedOnGroupId` |
| Relay discovery | mdma `GET /api/cloud/me/namespaces/{ns}/relays` → `relay_url`, `executor_account`, `authorship_ready` |
| Author-attributed execution | `Principal` (`crates/context/src/handlers/execute/principal.rs`) already runs a delegated write as the author, not the node |

## What blocks them

1. **#3929 — Freeze the interface contracts.** Cheapest unblock by a wide margin. Fix the challenge/`LoginStatement` exchange, the warrant field set and the read request/response shape, and hand over a stub that serves them. A day of our time against weeks of theirs, because the whole client can then be built against stubs and swapped onto the real node later. **Do this before writing any of 2–5.**
2. **#3930 — Account-proof auth provider (key-based login).** The largest item. Every client flow past posting a warrant needs a session, and today the only provider is `user_password` — which the target architecture deletes, and which is already switched off on fleet relay images (calimero-network/mero-tee#283). Once that ships there is no way in at all.
3. **#3931 — Delegated reads.** Without these a client can write but cannot display anything, which blocks every screen. Depends on #3930.
4. **#3932 — Derive `MethodIntent` from the receiver.** Reads gate on `MethodIntent::ReadOnly` and fail closed, and only **2** methods across 573 `pub fn`s declare it today. But the annotation is redundant: the macro already parses `SelfType::{Immutable, Mutable}` and already branches on it to decide whether to emit `app.commit()`. Deriving the intent turns this from an ecosystem campaign into a macro change. Must ship *with* #3931, or reads return nothing and will look broken.
[...]

PR #3940, opened 2026-09-15
feat(context,server): let an authenticated account read a context it is a member of

### LINKED ISSUE #3931: Let an authenticated account read a context it is a member of
## Summary

Delegated execution has a write path and no read path. An account that runs no node can author a write through a relay but cannot read anything back, so a client can change state and then cannot display it.

The only way such an account observes state today is as a side effect of a write: `PerformIntentApiResponseData` carries the method's return value, so reading means spending a warrant nonce, holding `CAN_AUTHOR_ON_BEHALF`, and publishing a delta — for a question that changes nothing.

Part of #3928. Depends on #3930.

## Impact

**Blocks every screen the client builds.** A client that cannot read is a client that cannot render.

## Why this is small

A read publishes nothing — no delta, no envelope, no DAG entry, no peer ever sees it. So there is no wire format change, no nonce, no peer-side verification and no signing. It also sidesteps the precondition that made delegated authorship wait: the sync path refuses a `User` leaf whose owner is not its author's account, and a read produces no leaf.

Most of the machinery is already in core and enforced:

| Piece | Where |
|---|---|
| `MethodIntent::ReadOnly`, declared with `#[app::view]`, emitted into the ABI | `crates/wasm-abi/src/schema.rs`; `extract_read_only_set` at `crates/context/src/handlers/execute/mod.rs:2708` |
| Read-only calls take a **shared** lock instead of the exclusive write lock | `execute/mod.rs:224` |
| A read-only call that produces a root hash is rejected **post-execution** | `execute/mod.rs:2198` |
| Read-only *identities* (`GroupMemberRole::{ReadOnly, ReadOnlyTee}`) with the same rejection | `governance-store/src/namespace/core.rs:91`; `execute/mod.rs:2181` |
| Separating the identity the app observes from the key that signs | `crates/context/src/handlers/execute/principal.rs` |

## Scope

1. Build `Principal` from the authenticated caller (`X-Auth-User`) rather than defaulting to the node's own identity. `principal.rs` already names this as a known cost: *"an application enforcing per-member permissions sees the node rather than the caller, so a member whose in-application permissions are lower than the node owner's executes at the higher level."*
2. Gate the delegated read to `MethodIntent::ReadOnly`. **Fail closed on `Unspecified`** — most methods are unannotated and must be refused, not guessed at.
3. Re-check membership of the target context's owning group **per call**, not once per session. A relay serves several tenants, so a session must not carry a cached right to read.
4. Keep the existing post-execution `root_hash.is_some()` rejection as defence in depth.

## Note on why writes still need a warrant

A session authenticates the caller **to this relay**. A warrant proves to peers who never saw the request that the author consented to that operation. Reads have no peers to convince; writes do. This issue does not change the write path.

## Criteria for resolving

- [ ] An authenticated account that is a member of a context can call an `#[app::view]` method on it and receive the return value.
- [ ] A method that is `Unspecified` or `Mutating` is refused with a distinguishable error.
- [ ] A non-member is refused, and membership is evaluated per call.
- [ ] The app observes the **caller's** identity, not the node's.
- [ ] A member holding the `ReadOnly` role can read and still cannot write.
- [ ] Ships together with #3932, or reads return nothing and will look broken.
- [ ] `cargo fmt`, `clippy --workspace --all-targets`, tests pass.

chefsale:
Closes #3931. Part of #3928. Completes the delegated-execution client's read path — with #3930 merged, an account holding only a root and a device key can now log in, read, and write through a node it does not own.

> **Validation note up front:** workspace clippy was still running locally when this opened, deliberately, so CI starts in parallel rather than after it. Everything else below has been run. If clippy comes back with anything I'll push the fix rather than leave it.

## The shape

`POST /admin-api/contexts/{id}/query` with `{ method, argsJson }`, returning the method's return value.

A read needs no warrant: a warrant proves to peers who never saw the request that the author consented, and a read has no peer to convince because it publishes nothing. What it does need is proof this caller may see this context, which is a question about membership *now* — a session answers that, a warrant does not.

Before this, the only way an account running no node could observe state was as a side effect of a write: `PerformIntentApiResponseData` carries the return value, so reading meant spending a nonce, holding `CAN_AUTHOR_ON_BEHALF`, and publishing a delta — to ask a question that changes nothing.

## Four things worth a reviewer's attention

**1. The read-only gate is not the `is_read_only_call` used for lock selection.** That one looks like exactly the right check and isn't: lock selection may answer conservatively, because a cold module cache yields `false`, it takes the write lock, and the only cost is contention. Reused as an authorization gate, that same `false` refuses a genuinely read-only method whenever its module hasn't loaded — intermittent 409s that depend on cache warmth. The gate sits beside the xcall gate instead, after module load, where the set is the declared one. A module with no ABI refuses the read.

**2. `read_as` does not weaken `principal.rs`'s "never deserialized from a request" invariant.** It is set by the server from the authenticated session, never from a body; refused unless the ABI declares the method read-only, failing closed on `Unspecified`; and membership is re-checked per call. The account/device halves may differ here where they may not for a write, because the rule that would break — `user_leaf_author_is_its_owner` on the receive path — is about a leaf, and a read writes none.

**3. Membership is re-checked every call**, through `MembershipRepository::is_member` so inheritance resolves and a subgroup member reads a context held through a parent. A context owned by no group refuses rather than falling through to "no rule, so allowed".

**4. A latent auth bug, fixed here because this is where it surfaced.** `get_public_key_for_key_id` ends in `key.and_then(|k| k.public_key)`, collapsing *"no row"* and *"row whose public_key is None"* into one `Ok(None)` — and the guard granted `AuthenticatedNodeOwner` to both, on a comment asserting no other key with an absent public key could exist. #3930 made that false: an `account_proof` session stores no row by design.

Not exploitable as it stands — the permission check runs first and such a token carries only `context:intent`/`context:query`, `AuthenticatedNodeOwner` is read by three `leave_*` handlers needing permissions it lacks, and the default `AuthMode::Proxy` installs no guard. But it is a false claim about who the caller is, which a new handler trusting the marker or a widened `session_permissions` would make live. `key_row_exists` asks the question the lookup couldn't; a row keeps the existing grant, no row yields `AuthenticatedAccount`, neither grants no identity, and a storage error rejects.

## The trap I walked into anyway

The route needed a `context:query` mapping in `validator.rs` or it falls to the `/admin-api/*` admin default-deny — a delegated client with exactly the right session would get a 403 reading as a bug in its own code. `/intents` carries a comment saying precisely this about its own mapping, and I still missed it on the first pass, because the routes and the permission table live in different crates. `the_query_route_requires_its_own_scoped_permission` pins it.

`context:query` is its own verb, not `Execute` (which covers join/leave/resync) and not `PerformIntent` (which submits writes). A token minted so a client can render should carry neither.

## Statuses

| | | |
|---|---|---|
| `401` | session is not account-anchored | names no account; a node owner reads via the execute path |
| `403` | not a member | a person must add you |
| `409` | method is not read-only | mint a warrant — your code called the wrong thing |

403 and 409 are separate because they are different instructions. `ExecuteError` derives `Copy`, so `NotReadOnly` can't carry the method name; the handler supplies it rather than stripping `Copy` from a type used everywhere.

## Validation

```
cargo check -p calimero-context -p mero-auth -p calimero-server   0 errors
cargo test -p mero-auth  (4 new permission tests)                 pass
cargo test -p calimero-server  (4 new refusal tests)              pass
cargo test -p calimero-server --test route_manifest               pass
cargo fmt --check                                                 clean
cargo clippy --workspace --all-targets -D warnings                RUNNING — see note above
```

Not run locally: the docs site build. I cleared `docs/node_modules` to escape a disk ceiling; the change is prose using an already-imported `<Aside>`, so CI is verifying it rather than me.

Pre-existing, verified on master and not introduced here: `test_rocksdb_specific_errors` fails when the suite runs as root, because it asserts that opening `/nonexistent/...` errors.

---


chefsale:
Closing the loop on the validation caveat in the description: local workspace clippy has since finished clean on this head.

```
EXIT=0, 0 errors
```

Everything the description listed as run has now been run. The one remaining gap is unchanged and stated there: the docs site build was not run locally (I had cleared `docs/node_modules` for disk), so CI is verifying that rather than me. The change is prose using an already-imported `<Aside>`.

---

Issue #3929, opened 2026-09-15
Freeze the delegated-execution client contracts and ship a stub node


chefsale:
## Summary

The client work for delegated execution cannot start on sessions, reads or a warrant builder until the shapes are fixed, and each of those shapes is currently either undecided (warrant v2) or non-existent (login, reads). Every day this is open, the client team is either idle or building against a guess.

This issue is not implementation. It fixes three wire contracts and provides something that serves them, so the client can be built end to end against stubs and swapped onto the real node later.

Part of #3928. **Do this before the other four.**

## Impact

Cheapest unblock in the epic by a wide margin — roughly a day of our time against weeks of theirs. It also protects against the expensive failure: they build a warrant builder against today's `Warrant`, we add `method` and `app_version`, the signed bytes change, and the work is thrown away.

## Scope

Three contracts, written down and version-tagged.

**1. Challenge and `LoginStatement`.** The exchange that turns an account's device key into a session:

- challenge endpoint: path, method, response shape, lifetime, whether the challenge is stateless (MAC over node id, nonce and expiry) and how spend is recorded
- `LoginStatement` fields, its signing domain, and its borsh encoding
- the session response: token shape, what identifies the account to downstream handlers, expiry
- whether request proofs are required per request in v1 or deferred

**2. Warrant v2 field set.** Today (`crates/account/src/warrant.rs`):

```rust
pub struct Warrant {
    pub context: ContextId,
    pub author_account: AccountId,
    pub author_device_key: PublicKey,
    pub executor: AccountId,
    pub intent_hash: [u8; 32],   // H(method ‖ args) — method NOT in the clear
    pub nonce: u64,
    pub not_after: u64,          // wall clock, seconds
    pub signature: [u8; 64],
}
```

Decide and freeze which of these land, each with a reason: `method` in the clear, `app_version`, `executor_node`, `account_heads`, `governance_floor`, and whether `not_after` becomes a governance height. Every addition changes the signed bytes, so this is one decision, not five.

**3. Read request and response.** Path, how the caller names the context and method, argument encoding, response shape, and the error a non-read-only method returns.

## Deliverable

- A contract document in `docs/` covering the three above, with the borsh layouts and one worked example per exchange.
- A stub that serves them — a small binary or a test fixture is fine — so the client can run its full flow without a node.

## Criteria for resolving

- [ ] The three contracts are written down with field-level types and signing domains.
[...]

PR #3946, opened 2026-09-15
feat(account)!: warrant v2 — plaintext method, app version, cited heads

### LINKED ISSUE #3933: Warrant: add the fields delegated execution needs, breaking the wire
## Summary

Today's `Warrant` (`crates/account/src/warrant.rs`) carries the method only as a hash, with no app version and no cited governance state:

```rust
pub struct Warrant {
    pub context: ContextId,
    pub author_account: AccountId,
    pub author_device_key: PublicKey,
    pub executor: AccountId,
    pub intent_hash: [u8; 32],   // H(method ‖ args) — method NOT in the clear
    pub nonce: u64,
    pub not_after: u64,          // wall clock, seconds
    pub signature: [u8; 64],
}
```

The target architecture needs a plaintext `method` (to select a per-method write-set), a pinned `app_version` (so a relay cannot wait for an upgrade that widens the method), and cited `account_heads` / `governance_floor`. Possibly `executor_node`, and `not_after` may become a governance height.

Part of #3928. The field set is decided in #3929; this issue lands it.

## Approach: change it in place

**No backwards compatibility.** The project is in alpha and fresh networks are acceptable, so this adds the fields to `Warrant` directly and bumps the signing domain, rather than carrying a v2 struct alongside v1 with a dual-accept intake path.

```rust
// crates/account/src/domain.rs:45
pub(crate) const WARRANT_SIGN_DOMAIN: &[u8] = b"calimero.warrant.v1";   // → v2
```

Bumping the domain rather than reusing it is still worth doing, and is cheap: it means an old signature cannot verify against the new layout under any parse, instead of relying on borsh to fail. The crate already has a test asserting every domain is pairwise distinct — keep it passing.

What this costs, stated so it is a choice rather than a surprise: **every warrant in flight becomes invalid**, and a node on an older build refuses the new ones and vice versa. In alpha that is acceptable; it would not be later.

## Impact

The client is about to write a warrant builder. Every one of these fields changes the signed bytes, so shipping them later throws that work away. Landing them early — even before verification uses the new fields — is what lets the client build once.

## Scope

- Add the fields decided in #3929 to `Warrant`, bump `WARRANT_SIGN_DOMAIN`, update `signing_payload` and every construction and verification site.
- Verification of the *new* fields can land incrementally: pinning `app_version` at the cut and selecting write-sets by `method` are follow-on work. This issue is the struct, the domain, the encoding and the intake.

## Two things to get right

**`not_after` must not gate an apply.** `crates/governance-store/src/warrant_gate.rs` states the rule and the reason:

> "Wall-clock expiry must not gate an apply. Peers apply at different times, so a warrant that expired between two receivers would be accepted by one and refused by the other, and authorization would stop converging."

The bound belongs at the relay's API boundary, where one clock decides and nothing has converged. If this replaces `not_after` with a governance height, that height is a **shared** check evaluated at the cut — a different rule, not the same one relocated.

**`governance_floor` is only as good as its source.** A floor the relay supplied is a floor the relay chose. It must come from the account's own devices or a co-signing peer outside the relay's operator; otherwise it is an accepted risk rather than a guarantee, and should be documented as one where the field is defined.

## Criteria for resolving

- [ ] `Warrant` carries the field set frozen in #3929, under a bumped signing domain.
- [ ] A warrant signed under the old domain does not verify — with a test asserting it.
- [ ] The pairwise-distinct domain test still passes.
- [ ] No wall-clock field gates an apply; any new height-based bound is evaluated at the cut.
- [ ] The `governance_floor` provenance limitation is documented where the field is defined.
- [ ] `cargo fmt`, `clippy --workspace --all-targets`, tests pass.

chefsale:
Closes #3933. Part of #3928 — its last issue. Based on `master` (#3940 merged).

```
sdk-ref: claude/delegated-execution-rollout-3hlfmf
```

**Paired SDK PR:** [calimero-network/mero-js#148](https://github.com/calimero-network/mero-js/pull/148) — `signWarrant` there was producing v1 bytes, and the `sdk-ref:` above points `SDK E2E (paired)` at the branch that fixes it.

**Merge order — correcting what this said earlier.** It said "mero-js first, or this PR's SDK e2e goes red against `master` mero-js". That was wrong twice over: the `sdk-ref:` above already insulates *this* PR whichever lands first, and mero-js#148 cannot go green on its own at all — its `E2E (vs released merod)` suite runs against the newest **released** merod by design, and no released merod verifies v2. So the order is **this PR first, then a merod release, then mero-js#148**. Merging mero-js first would red other core PRs' `SDK E2E (paired)` (which defaults to `master` mero-js) *and* leave that repo red anyway.

## Description

The v1 warrant carried the method as a hash, no app version, and no cited state. #3933 lands the field set frozen in #3929, in place and under a bumped domain — no dual-accept intake, because the project is in alpha and fresh networks are acceptable.

```rust
pub struct Warrant {
    pub context: ContextId,
    pub author_account: AccountId,
    pub author_device_key: PublicKey,
    pub executor: AccountId,
    pub app_version: ApplicationId,        // NEW — the build, not a version string
    pub method: String,                    // NEW — in the clear
    pub intent_hash: [u8; 32],
    pub account_heads: Vec<[u8; 32]>,      // NEW — capped at 64
    pub governance_floor: Vec<[u8; 32]>,   // NEW — capped at 64
    pub nonce: u64,
    pub not_after: u64,
    pub signature: [u8; 64],
}
```

`WARRANT_SIGN_DOMAIN` → `calimero.warrant.v2`. Every warrant in flight becomes invalid; that is the stated cost.

## Five things worth a reviewer's attention

**1. The counts in `signing_payload` are load-bearing.** `domain_hash` length-prefixes every part, so each head is individually unambiguous — but the two lists are *adjacent*, so without a count in front of each, `account_heads = [a, b], governance_floor = []` hashes identically to `[a], [b]`. A relay could relabel which plane a head was cited from. Three tests cover it: mutating a list, emptying one, and (on the SDK side) moving a head between them.

**2. `sign` takes `WarrantTerms`, not eleven positional arguments.** Four of the eleven are `[u8; 32]`. The v1 signature had seven and was already a swap waiting to happen; eleven is not a list anyone checks by eye.

**3. Nothing verifies the new fields yet, by #3933's own scope** ("pinning `app_version` at the cut and selecting write-sets by `method` are follow-on work; this issue is the struct, the domain, the encoding and the intake"). They are in the signed bytes now so a client builds against the final encoding once instead of rewriting its signer when enforcement lands. Said in the field docs, in the client contract's status table, and again beside the vectors — a signed-but-unchecked field that nobody flags reads as a guarantee.

**4. The minting clients differ because their situations do.** `meroctl context intent` reads `app_version` from the context it is about to write to, the way it already reads the executor account from the relay descriptor — a client *guessing* the build would pin the wrong one, which is worse than not pinning. `merod account warrant` is deliberately offline and has nothing to read, so `--app-version` is a flag defaulting to all-zeros; mero-js's `appVersion` is optional with the same default, so a caller with nothing to read it from produces the bytes merod would. None cites heads: none tracks a log, and an empty list says so rather than fabricating a view.

**5. A v1 warrant now gets a 400, where it used to get a 500.** The two `warrant` decode failures in `perform_intent` raised bare `eyre!` strings while the `authorProof` pair right beside them raised `IntentRefusal::Malformed` — so a malformed warrant fell through to a 500 and a malformed credential got its 400. Invisible while nothing sent an undecodable warrant; this PR makes it the single most common refusal there is, and "Internal server error" points a client at the node rather than at its own encoder. `decode_warrant` is extracted so the refusal is testable without a `ContextClient`, and both tests assert the *status*, since the old behavior failed too.

## There was a fourth minting client, and it was inside the test harness

`Test (delegated-authorship)` went red on this branch with the node's own new refusal:

```
✗ expected the failure to mention 'CAN_AUTHOR_ON_BEHALF', but it was:
  HTTP 400: warrant is not a valid statement (240 bytes)
```

240 is the v1 fixed width. The scenario's `sign_warrant` step is a **merobox** step type, and merobox reaches the encoding through `from calimero_client_py import sign_warrant` — a crate that takes `calimero-account` as a git dependency on core's `master`. So the bytes that scenario signed came from whatever was bundled into the pinned merobox release binary, not from the code under test. That is also circular: getting v2 into that binary needs calimero-client-py published, merobox rebuilt and `MEROBOX_VERSION` bumped here, and the first of those wants this change on `master` already.

The scenario now mints through `merod account warrant` via the `node_exec` step the `sign-cert` step above it already used. `allow_running: true` is safe where it is not for `sign-cert`: `WarrantCommand::run(self)` takes no root args and opens no datastore, so it never contends for RocksDB's exclusive lock.

**calimero-client-py needs its own follow-up regardless of this PR**: `src/warrant.rs` calls `Warrant::sign` positionally, so it stops compiling against `master` the moment this lands.

## The encoding is no longer fixed-width

v1 was pure concatenation — no length prefixes, no tags — which is what made a non-borsh implementation viable. `method` is a `String` and each head list a `Vec`, so three `u32` LE counts now sit inside the encoding, **while the signing preimage hashes its counts as `u64`**. Same numbers, different serialization, and exactly the kind of thing a second implementation gets wrong silently. Both widths are pinned on both sides.

| | v1 | v2 |
|---|---|---|
| wire length | 240, fixed | 351 for the fixture's inputs, variable |
| preimage | `a5d5f7aa…` | `cd03cfaa…` |

That also broke three mero-js call sites reading the warrant by constant offset — they read a neighbouring field rather than failing, and one compared the method's length prefix to a nonce and still passed as "different". mero-js#148 adds `parseWarrant`, beside the encoder so the two cannot drift.

It also moved a frozen vector in this repo: `the_delegated_preimage_is_byte_frozen` in `calimero-node-primitives`. The delegated delta preimage embeds the warrant **whole**, so `app_version`, the plaintext method and the two head counts move those bytes by construction. Re-frozen, with the domain-bump question the surrounding comment block asks answered in the test: the bump happened on the *warrant*, and `DOMAIN_SEPARATOR_DELEGATED` needs none of its own, because a v1 warrant no longer verifies at all — no old delegated preimage survives to be confused with a new one. The self-authored preimage is byte-identical and its test is untouched, which is the claim that second domain exists to make.

## Criteria for resolving (#3933)

- [x] `Warrant` carries the #3929 field set under a bumped signing domain
- [x] A warrant signed under the old domain does not verify — `a_warrant_signed_under_the_old_domain_does_not_verify` signs the v2 field set under `calimero.warrant.v1` and *nothing else*, so it proves the domain rather than the layout: same bytes, same key, same order, refused
- [x] The pairwise-distinct domain test still passes
- [x] No wall-clock field gates an apply — `not_after` is unchanged and still checked only at the relay's API boundary; no height-based bound added
- [x] The `governance_floor` provenance limitation is documented where the field is defined ("a floor the relay supplied is a floor the relay chose"), in the client contract, and on the SDK's own field
- [x] `cargo fmt`, `clippy --workspace --all-targets`, tests

## Test plan

```
cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings   0 errors
cargo fmt --check                                                                           clean
cargo test -- --nocapture                                                       138 suites ok
```

The one failure in that workspace run is `storage::providers::rocksdb::tests::test_rocksdb_specific_errors`, which asserts that opening `/nonexistent/…` errors and so fails whenever the suite runs as root. Pre-existing on `master`, green in CI.

The `node_exec` rewrite was validated against the **pinned merobox's own schema**, with its `docker` and `aiohttp` imports stubbed so the config models load without Docker — `NodeExecStepConfig` 3/3, `validate_workflow_config` valid with no warnings — and the exact argv replayed against `target/debug/merod`, which mints 287 bytes on one line matching the step's capture regex. The scenario itself needs Docker and was not run locally; CI is its check.

mero-js#148:
```
vitest run (unit)                                              383 passed, 1 skipped
eslint .                                                       0 errors
tsc --noEmit -p tsconfig.contract.json / tsconfig.e2e.json     clean
CALIMERO_CORE_DIR=<this branch> npm run test:contract          11 passed
```

## Proof the fix works

Not a bug fix — a deliberate wire break. Two pieces of evidence.

**The fixture**, which is the artifact a second implementation is diffed against:

```
                  v1                                                                 v2
signing preimage  a5d5f7aa368f8d79edd0667f60c41239a7dedaf8765e4e72ef3031bbacd15f35   cd03cfaad8fb5f5aa143c76f35f39363b6608b820ab318a53ed01be901c7c6c7
wire              240 bytes, pure concatenation                                      351 bytes, field by field, with the three u32 counts spelled out
intent_hash       dc066cc8…                                                          dc066cc8…  (unchanged — its own domain, untouched by this)
```

`the_intent_hash_is_stable` still passes on the old digest, which is the control: the commitment domain did not move, only the signing one.

**The break, reproduced and then closed.** Before the paired SDK branch, this PR's own `SDK E2E (paired)` run named it exactly — `delegated-intent.test.ts`, 4 failed:

```
× produces the same bytes merod does, for the same inputs
  Expected: …ddab7343 [0000…0000] [03000000 736574] fbc11918… [00000000] [00000000] 6300…  ← merod (v2)
  Received: …ddab7343                              fbc11918…                       6300…  ← mero-js (v1)

× refuses the intent before the relay is granted authorship
  expected /CAN_AUTHOR_ON_BEHALF/ but got 'HTTP 500 Internal Server Error'        ← the 500 fixed above
```

The v1 bytes are missing `app_version`, the method's length prefix and its text, and both head counts — the node's borsh decode then failed, which is what surfaced as the three 500s. After the SDK branch, mero-js reproduces core's 351-byte fixture independently (its conformance test asserts the same signature `4007d416…` from the same inputs), and the `sdk-ref:` above is what makes this PR's run use it.

## Wire contract (SDK gate)

This changes the borsh encoding of a type mero-js hand-mirrors. No HTTP DTO or route changed — the warrant travels as `"warrant": "<hex>"` inside the unchanged `PerformIntentApiRequest`, so the JSON shape is identical and only the hex bytes differ.

- [x] Regenerated wire fixtures — `crates/account/src/tests/warrant_wire_fixture.rs` (the account crate's own fixture; no `calimero-server-primitives` DTO changed, so `wire_fixtures` is untouched)
- [x] Updated `crates/server/endpoints.json` — n/a, no route changed
- [x] Linked the matching mero-js PR — mero-js#148, via `sdk-ref:`

## Documentation update

Done in this PR. `docs/src/content/docs/build/delegated-execution-client.mdx` carries the new struct, the new preimage recipe with the count prefixes, the 351-byte field-by-field layout, the u32-vs-u64 warning, a caution that v1 no longer verifies, and a note on where `governance_floor` must come from. The status table now reads "Live; `app_version` and cited heads signed but not enforced" rather than "field set changes".


chefsale:
`Rust` is red on `e64a018495` and I have not been able to localize it. Recording where I got to rather than leaving it silent.

**What is failing:** step 8, `Cargo test`, and only that step — Build, fmt, clippy, clippy/test under `mock-attestation`, the all-features store test, deny and machete all pass ([job](https://github.com/calimero-network/core/actions/runs/35005147612/job/104503294685)). It fails in **1m47s**, against 6m45s for the same step on #3940, so it is an early failure rather than a slow convergence test.

**It is this PR's**, not inherited: #3940 at `9bf836720` passes the same step, and the only delta is warrant v2.

**What I ruled out locally on this exact head**, all green:

| crate | result |
|---|---|
| `calimero-account` | 89 passed |
| `calimero-governance-store` | 708 passed |
| `calimero-node-primitives` | 300 passed |
| `calimero-context` | passed, incl. `delegated_authorship` 6/6 |
| `calimero-server` + `-primitives` | 172 / 31 / 13 / 2 / 1 passed |
| `meroctl` | 51 passed |
| `mero-auth` | 204 passed, 1 failed — `test_rocksdb_specific_errors`, the known as-root failure, pre-existing on master and green in CI |

That covers every crate whose fixtures this PR rewrites for `WarrantTerms`, plus both CLI minters. I also checked that no new doc-test code block was introduced, and that `crates/node/src`, `crates/node/tests`, `context/primitives`, `authz`, `dag` and `runtime` contain no `Warrant` reference at all.

**What blocks me:** `merod` and `calimero-node` are the only crates left, and this container cannot link them — the linker dies with `No space left on device` / `ld terminated with signal 7` partway through `merod`'s binary. I have reclaimed disk four times in this session (deleting `target/debug/deps` and ~250 duplicate `target/debug/build/<crate>-<hash>` dirs each time) and it is not enough for that link. Neither crate references `Warrant` outside `merod/src/cli/account.rs`, whose test module I read line by line: the `--app-version` flag I added carries a default and participates in no `conflicts_with`, so the two argument-parsing tests there are unaffected as far as reading can establish.

I also could not read the failure from CI: the logs API caps the tail at ~5000 lines and the whole tail of that job is `cargo deny` / `cargo machete` output, both of which passed, and the check run's `output.summary`/`output.text` come back empty.

**What would settle it in one step:** anyone with log access opening step 8 on that job and pasting the failing test name. I will fix it immediately. Failing that I will keep bisecting as disk allows, and this PR stays watched until it is green.

---

chefsale:
Two updates: the `Cargo test` failure is solved, and the E2E failure that replaced it has a diagnosis with a fourth v1 signer in it.

## `Cargo test` (step 8) — fixed

It was one test out of 322, in a crate early in the run order, which is why the step died in 1m47s while #3940's took 6m45s:

```
sync::delta_auth::tests::the_delegated_preimage_is_byte_frozen
test result: FAILED. 321 passed; 1 failed
```

The delegated delta preimage embeds the warrant **whole** — `the_warrant_is_embedded_verbatim_in_the_preimage` is the pin that says so — so `app_version`, the plaintext `method` and the two head counts move those bytes by construction. The diff is exactly those fields and nothing else; `intent_hash`, `nonce` and `not_after` are unchanged, and the self-authored preimage is byte-identical with its test untouched, which is the claim `DOMAIN_SEPARATOR_DELEGATED` exists to make.

The comment block above those tests says a failure there is never answered with a new constant unless already-signed deltas still verify. They don't, and that's the intended effect, so the answer is written into the test rather than left to the reader: the domain bump it asks for happened on the *warrant* (`calimero.warrant.v1` → `v2`), and `DOMAIN_SEPARATOR_DELEGATED` needs none of its own, because a v1 warrant no longer verifies at all — no old delegated preimage survives for a second domain to separate from a new one.

Re-frozen in `0430cc394`. `master` is also merged in (`0f586eb39`); its one conflict was the docs status aside, where `master` still said the warrant's field set was the thing still moving.

```
cargo clippy --workspace --all-targets --features calimero-storage/testing -- -D warnings   0 errors
cargo test -- --nocapture --skip bundle_produces_signed_mpk                                 138 suites ok
```

The single remaining failure is `storage::providers::rocksdb::tests::test_rocksdb_specific_errors`, which asserts that opening `/nonexistent/...` errors and so fails whenever the suite runs as root. Pre-existing on `master`, green here.

## `Test (delegated-authorship)` — this PR's, and blocked on a release chain

```
Executing Refused before the relay is granted authorship (perform_intent)...
✗ expected the failure to mention 'CAN_AUTHOR_ON_BEHALF', but it was:
  HTTP 400: warrant is not a valid statement (240 bytes): Unexpected length of input.
  A warrant signed under the v1 layout no longer decodes …
```

240 bytes is the v1 fixed width, so the scenario is minting v1. The PR description says there are three minting clients. There are **four**: the scenario's `sign_warrant` step is a merobox step type, and merobox reaches it via `from calimero_client_py import sign_warrant`.

That signer is not a reimplementation — `calimero-client-py`'s `src/warrant.rs` calls `calimero_account::Warrant` directly, with `calimero-account` taken as a git dependency on **`core` branch `master`**. Two consequences:

1. It will produce v2 bytes automatically once this lands, so there is no second encoding to keep in sync — but it will not *compile* against post-#3946 `master` until it is adapted to `WarrantTerms`.
2. The failure is circular as things stand. CI downloads merobox pinned at `v0.6.75` with a checksum, a PyInstaller binary bundling a `calimero-client-py` built against pre-#3946 `master`. Getting a v2 signer into that binary needs calimero-client-py published, merobox rebuilt, and `MEROBOX_VERSION` bumped here — and the first of those wants this PR on `master` first.

The way out that needs no release chain: merobox has a **`node_exec`** step that runs a merod subcommand against a node with `capture` regexes for its stdout, and `merod account sign-warrant` — updated to v2 in this PR — is offline (`WarrantCommand::run(self)` takes no `root_args` and opens no store, so it is safe alongside a running node). Minting through it makes the scenario exercise the product's own path instead of a signer inside the test harness, which is where this coupling should not have been.

I'm implementing that. Flagging it rather than pushing quietly because it is a scenario redesign rather than a constant, and because the calimero-client-py adaptation is needed either way before the next merobox build.

---

PR #3963, opened 2026-09-16
fix(deps): remove duplicate wasmparser entries from Cargo.lock

rtb-12:
## Summary

- A dependency bump left two identical `wasmparser 0.259.0` `[[package]]` entries and a stale `0.258.0` one in `Cargo.lock`, so cargo refuses to parse the lockfile on a clean checkout of master (`package 'wasmparser' is specified twice`).
- Drops the duplicate and the stale entry; cargo regenerates the section to one `0.259.0` entry.

## Test plan

- [x] `cargo metadata --locked` parses on this branch and fails on master (`failed to parse lock file`).
- [x] `cargo test -p calimero-server-primitives` builds and passes.


rtb-12:
Superseded: master already carries a clean Cargo.lock as of #3831 (7b918e1ee). #3964 is retargeted to master.

Issue #3947, opened 2026-09-15
Blobs have no owner, so /admin-api/blobs cannot be scoped to a caller


chefsale:
## Summary

Split out of #3941, which scoped `GET /admin-api/contexts` and `GET /admin-api/namespaces` to the caller and could **not** scope `GET /admin-api/blobs`. The reason is not a missing index — it is that the data model has no owner to index.

`BlobMeta` (`crates/store/src/types/blobs.rs:10`) is the whole of a blob's metadata:

```rust
pub struct BlobMeta {
    pub size: u64,
    pub content_hash: ContentHash,
    pub links: Box<[key::BlobMeta]>,
    /// Number of live references to this content-addressed blob.
    pub refs: u32,
}
```

No context, no group, no account. And `refs` is not incidental — its own doc comment says why it exists:

> Blobs are deduplicated by content hash: a given byte sequence is stored once but may be referenced by many owners — **the same bytes added by several contexts**, or a chunk shared by several root blobs.

So ownership is genuinely **many-to-many**, by design. `list_handler` (`crates/server/src/admin/handlers/blob.rs:211`) iterates the flat `key::BlobMeta` keyspace, and `GET /admin-api/blobs/{id}` performs no ownership check at all: a `blob:get` scope reaches any blob on the node.

## Impact

**Cross-tenant disclosure on a multi-tenant relay, and the more serious half of #3941.** Context and namespace ids leak identifiers; blobs leak *content*. `GET /admin-api/blobs/{id}` will serve one tenant's file to another holding `blob:get`.

Bounded today by the permission layer exactly as #3941 was — an `account_proof` session carries only `context:intent` and `context:query`, so it reaches neither route. The exposure is to any caller granted `blob:list`/`blob:get` on a shared node, which is what those scopes are for.

## Why this needs a decision, not an index

#3941 added `GroupMemberByAccount` and the endpoints followed. The same move does not work here, because there is no single answer to "whose blob is this":

1. **Two tenants can legitimately own the same bytes.** Dedup means one row, `refs: 2`. An owner *set* is the honest model, and it has to be maintained as refs rise and fall — the reference counting already exists, so this is a matter of recording who each reference is for.
2. **Chunks are not user-facing.** `list_handler` already excludes blobs referenced as chunks. An ownership model has to decide whether a chunk inherits its root's owners or is simply never listed.
[...]

Issue #3943, opened 2026-09-15
A context identity a client holds, rather than one the node owns


chefsale:
## Summary

Across the app fleet, "who am I in this context" is answered by asking the node what **it** owns. `GET /admin-api/contexts/:id/identities-owned` is called by 7 of 16 apps; `POST /admin-api/identity/context`, which mints a fresh context keypair, is called by **none**.

That is the deepest mismatch with a device that runs no node, and it is in the app layer rather than the protocol. A delegated device has no identity the node owns, so every app that routes through `getContextIdentitiesOwned` has no answer to give.

It surfaces most concretely on blobs. `GET /admin-api/blobs/:id?context_id=` is load-bearing, and the reason is recorded in an app:

> Pass `context_id` so the node does P2P network discovery for blobs it doesn't have locally **AND signs the blob-access auth with our owned identity in that context**.
> — `apps/mero-design/app/src/api/rpc.ts:225`

So blob reads are not a transport problem. Without an owned identity there is nothing to sign the access with.

## Impact

**Blocks blobs for delegated devices entirely** (6 of 16 apps use them: mero-sheets, mero-issue-tracker, mero-design, mero-pixart, mero-sign, scaffolding-e2e), and leaves the other 7 apps' identity resolution with no answer.

Worth recording because it inverts an assumption: mero-drive and mero-stream, the two apps that look blob-heavy, use no blobs at all — mero-drive moves Yjs updates through `execute`, mero-stream ships video frames through ephemeral presence. Both are *harder* to delegate than blob storage, not easier.

## Steps to reproduce

1. Obtain an account-anchored session on a node the caller does not own (#3930).
2. `GET /admin-api/contexts/{id}/identities-owned`.

Expected, for a delegated caller: an identity derived from the account that the app can use as the executor. Actual: the node's own owned identities, which are not the caller's.

3. `GET /admin-api/blobs/{id}?context_id={id}` as that caller — the node signs blob access with an identity that is not the caller's.

## Scope

The question this issue must answer first, before any code: **what is a delegated device's context identity?**

Two shapes, and they are not equivalent:

- **The account is the identity.** `Principal` already separates the observed account from the signing device, and #3931 uses exactly that for reads. Simplest, and consistent with governance rows being account-keyed. But blob access is signed, and an account is a content address, not a key.
- **A device-derived per-context key**, certified under the account root, that the client holds and signs with. Answers the blob case, at the cost of a new credential and a delivery path.

[...]

Issue #3942, opened 2026-09-15
A delegated device receives no events: SSE has no account-authenticated path


chefsale:
## Summary

With #3930 and #3931, an account that runs no node can log in, read and write. It still receives no events, so the UI never updates by itself.

SSE is the event transport (`crates/server/src/sse/`): `GET /sse` to connect, `POST /sse/subscription` with `{contextIds?, groupIds?}` to subscribe, both Bearer-authenticated. The subscribe path resolves the caller through `caller_account::for_context` / `for_group` (`ws/subscribe.rs:172`, `:311`) — which resolves a **key** to an account. An `account_proof` session has no key row; its identity is the account itself, so that resolution has nothing to work with.

## Impact

**Blocks the client's usability even once reads and writes work.** A delegated device that can read and write but gets no events shows a UI that looks frozen: state changes on other devices never arrive, and the only remedy a client has is to poll, which most do not implement.

Measured against the app fleet (`calimero-network/apps`, 16 frontends): **13 of 16 subscribe to events**, most with no polling fallback. `useSubscription` from mero-react accounts for 6; 5 more use `SseClient` from mero-js directly. Two apps depend on it for their core function rather than for freshness — mero-stream ships encoded video frames through ephemeral presence over SSE, and mero-drive carries Yjs document updates.

The WebSocket alternative exists but is `@experimental` and used by zero apps.

## Steps to reproduce

1. Obtain a session via the `account_proof` provider (#3930).
2. `GET /sse` with that Bearer token.
3. `POST /sse/subscription` with a context the account is a member of.

Expected: the subscription is accepted and events for that context arrive. Actual: the caller cannot be resolved to an account by the key-based path, so the subscription does not establish for that account.

## Scope

- Resolve an account-anchored session on the SSE subscribe path, alongside the existing key-based resolution — the `AuthenticatedAccount` extension added in #3931 is the same input.
- Evaluate membership **per subscription and on re-subscribe**, not once at connect, for the same reason reads do: a relay serves several tenants and a long-lived stream outlives a removal.
- Decide what happens to an established stream when the account is removed from a group mid-stream. Closing it is the safe answer; silently continuing to deliver is not.
- A permission scope for subscribing, mirroring `context:query` — reusing `context:execute` would hand a subscriber join/leave/resync.

## Criteria for resolving

- [ ] An account-authenticated session can subscribe to a context it is a member of and receives its events.
[...]

PR #3967, opened 2026-09-16
test(e2e): prove a keyholder can log in and read, with no password

chefsale:
## Description

Epic #3928 states its own criterion:

> A client holding only an account root and a device key can **obtain a session against a node it is not the owner of**, read a context it is a member of, and submit a delegated write — with no password anywhere in the flow.

Nothing exercised the session half. `delegated-authorship.yml` proves such a client can **write** through a relay, but it logs in with a username and password — the provider this architecture exists to remove.

Three walls had to come down before this could run at all, and each was a wall rather than a gap:

| | |
| --- | --- |
| Nothing could **produce** a login statement | #3962 |
| No node would **accept** one — `node_key` was set by nothing at all | #3962, corrected in #3966 |
| Nothing **enables** the provider | deliberate; this scenario opts in per node |

## ✅ Run against live nodes — green

The scenario has now executed against real `merod` containers and passed. The keyholder path works end to end:

```
challenge: db22ab6a…
statement signed, session key 1038836c…
session minted with no password
delegated read served: {"data":{"returns":"read-by-a-keyholder"}}
```

`read-by-a-keyholder` is the value the scenario writes two steps earlier, so the read returned *this context's state* rather than an empty or defaulted answer.

Getting there took two fixes, both for defects this run exposed and no unit test could have:

- **`fix(auth): give an account-proof login the key record the token path needs`** — a verified proof authenticated and then died on `{"error":"Failed to generate tokens"}`. `generate_token_pair` resolves the subject through the key manager and fails closed when absent; this provider minted no record, by design. The first successful proof now registers the account, which also hands an operator the only revocation handle they have over a keyholder.
- **`fix(auth,server): identify an account-proof session by what minted it`** — that record then broke the auth layer's classifier, which inferred "this is an account" from the *absence* of a row. Recording one made an account-proof session resolve as `AuthenticatedKey` instead, and the delegated read refused it. The layer now classifies on the record's own `auth_method`, consulted before either inference and failing closed, so an account is never read as the node owner — which on a relay is another tenant.

## Why the scenario enables the provider rather than core

`account_proof` is off by default and stays that way. `allowed_audiences` defaults to **empty, which accepts any audience**, so enabling the provider globally would ship cross-surface session reuse as the default posture for every node anyone runs.

So the scenario does what an operator would, through `init_args` — merobox re-initialises a node on start, so a `merod config` write between `stop_node` and `start_node` is silently discarded:

```
--device-key-login --device-key-login-audience cli
```

The audience is pinned rather than left open for the same reason — an empty list would wave the audience check through instead of exercising it.

## What makes each assertion load-bearing

**`node_key` is deliberately not configured.** The node fills it from its own signing key at startup, and that is the behaviour under test. The scenario reads the value back from `/admin-api/identity`, which reports it through the *same accessor* the auth service reads (`NamespaceRepository::node_identity`), so the two cannot disagree. A constant pinned in the YAML would verify against itself and nowhere else.

**The signing device joins nothing.** It is minted offline for node-2's account via `account sign-cert --generate`, then logs in to **node-1** — a node it does not own, which is precisely the case the statement's `node` field exists to make safe. A device-keyed membership check would refuse the read, because a thin client's key sits in no group's binding rows; the *account* is the member.

**The read asserts the value, not the shape.** A value is written before the session is minted, and the check matches that exact value. An earlier version grepped for a field name this route does not return (`"output"`; the DTO is `QueryContextApiResponseData { returns }`) — it failed while the read was succeeding, and in the other direction a bare field-name grep would pass on any body containing the word, including an error mentioning it.

## A defect this fixes in the script from #3962

The script subscribed to SSE, slept 8 seconds and required an event. **That could never have passed**: merobox runs steps sequentially, so nothing writes during that sleep and the stream could only time out.

The honest trigger is a delegated *write* from the same session — a warrant minted by the device and spent by the relay, which is #3942's real shape and needs the authorship grant `delegated-authorship` sets up. Worth doing, and not this scenario's first job: what has never run is the session. The assertion is removed rather than left to fail.

## What this does not cover

- **Delegated writes** — already covered by `delegated-authorship.yml`.
- **Events for a delegated session** (#3942) — see the defect note above.
- **Blobs** — #3943 and #3947 are open, and a delegated device cannot fetch blobs at all until #3943 decides what its context identity is.


chefsale:
## `delegated-session` is red, root-caused, and blocked on a decision rather than a fix

Two real defects found by the first-ever runs of this scenario, one fixed and one not fixable from inside this PR.

### Fixed (`7c43fa66f`)

`install_application` pointed at `./apps/kv-store/res/kv_store.wasm` — relative to the repo root rather than the workflow file, and a raw `.wasm` where the step takes an `.mpk`. Now installs the same bundle `delegated-authorship.yml` does.

### Not fixed: the scenario cannot enable the provider

On `7c43fa66f` every step passes — namespace, invite, join, context, write, sync, `sign-cert`, both `merod config` writes, restart, `node_identity` (a real account and device), the assert — and then the script gets **404 on `GET /auth/challenge`**:

```rust
if !state.0.config.providers.get("account_proof").copied().unwrap_or(false) {
    return error_response(StatusCode::NOT_FOUND, "The account_proof provider is not enabled on this node", None);
}
```

The provider is off at runtime. The restart step says why:

```
Executing Start node-1 with device-key login enabled (start_node)...
Initializing node delegated-session-node-1...
✓ Applied e2e-style defaults to delegated-session-node-1
```

**merobox's `start_node` re-initialises the node and rewrites `config.toml`**, discarding the two `merod config` writes. `delegated-authorship.yml` survives its own stop/start only because `sign-cert` reads and never writes.

### Every other route is closed

| route | why not |
| --- | --- |
| env vars | `mero-auth` reads only `MERO_AUTH_ADMIN_*` and the KEK; nothing for `providers` |
| `nodes.init_args` | real and per-node, but passes flags to `merod init`, which has **no auth-provider flag** (all 25 checked) |
| `docker restart` from a script | no precedent anywhere in the suite |

### The actual gap

**There is no supported way to stand up a node with device-key login.** Not awkward — none. Hand-editing `config.toml` is the only route, and merobox overwrites it. That is the same shape as the two gaps this branch already closed: the feature exists and nothing can switch it on.

The smallest honest fix is a `merod init` flag (`--account-proof`, and probably an audience argument), used here via `init_args`. It survives re-initialisation, and it is the explicit per-node opt-in that the "off by default" posture implies — you can decide device-key login is opt-in, but there has to be something to opt in *with*.

That is CLI surface on a test PR, so it is **not** being added unilaterally. Flagging the decision here rather than quietly widening the diff or leaving the job red without explanation.

Nothing else on this PR is failing; the rest of CI is green on `7c43fa66f`.

---

chefsale:
`Release Mero Auth container` is red on `cc73ed4b6` and it is not this PR's.

It failed at **"Booting builder"**, pulling `moby/buildkit:buildx-stable-1` from Docker Hub:

```
ERROR: Error response from daemon: Head "https://registry-1.docker.io/v2/moby/buildkit/manifests/buildx-stable-1":
read tcp 10.0.0.43:34728->104.18.43.178:443: read: connection reset by peer
```

Three things place it outside this diff:

- **Checkout never ran.** `Log in to GitHub Container Registry`, `Extract metadata`, `Resolve buildx annotations` and `Build and push image` all report `outcome=skipped`. Nothing from this branch was fetched, let alone compiled.
- **`Release Merod container` succeeded in the same run, at the same second**, through the same buildx setup against the same registry.
- This PR's diff is confined to `crates/auth`, `crates/merod`, `apps/scaffolding-e2e` and one workflow file — nothing Docker-related.

Every other workflow on this head is clean: CI (`Rust`, both Windows jobs, `cargo-mero e2e`, `WASM size gate`), App migration e2e (468 steps), SDK E2E, Coverage gate, ABI Contract, ABI Validation and Lint all report zero failed jobs.

**On the re-run:** the single re-run this warrants has been refused three times with `403 This workflow is already running` — `Release App Bundle` has been queued in that run since 20:28. I'll keep retrying until the run goes idle. If it then fails a second time on the same registry pull, it stops being treatable as transient and I'll say so here rather than re-running again.

Still outstanding on this PR: `Test (delegated-session)` has not executed yet — its workflow run (`35146687489`) has been `pending` with no jobs created since 20:27. That scenario is the point of the PR, so this is not green until it runs.

---

PR #3969, opened 2026-09-17
feat(context): create a namespace open to delegated authorship

chefsale:
# [context] create a namespace open to delegated authorship

> ⚠️ **Known defect, found after opening — do not merge as-is.** The claim that this mask cannot diverge between peers is **wrong**, and the fix is not yet in this branch. See *The value is replicated — no, it is not* below and [comment 5712442071](https://github.com/calimero-network/core/pull/3969#issuecomment-5712442071). Everything else in this description stands.

## Description

`CAN_AUTHOR_ON_BEHALF` is implied by nothing — not membership, not admin, not the subgroup-admit cascade — so a namespace was created authorship-closed and **every attested relay admitted to it afterwards needed its own admin-signed op**. For a namespace that keeps admitting fleet nodes that is one governance op per admission, published at the moment a node is assigned and the admin is not watching. The grant then arrives late or not at all, and the symptom is a write refused at `POST .../intents` *after* the author has already spent a warrant nonce on it.

Core already had the mechanism: `MembershipRepository::add_member` seeds a **non-admin** member's capability row from the group's default mask at admission. So putting the bit in that mask is what makes an attested node land able to relay. Creation is the one point that needs no backfill — the mask reaches every member admitted after it and no member admitted before, and at creation there are none.

The hardcoded mask at the local-row write becomes `initial_default_capabilities(parent_group_id.is_none())` — namespace creation is `create_group` with no parent:

| | mask |
| --- | --- |
| namespace (root) | `CAN_JOIN_OPEN_SUBGROUPS \| CAN_AUTHOR_ON_BEHALF` |
| subgroup | `CAN_JOIN_OPEN_SUBGROUPS` (unchanged) |

**Namespaces only, deliberately.** A grant resolves through the membership anchor, so one at the root already reaches every Open subgroup beneath it — which is where contexts live — and a `Restricted` subgroup is a deliberate membership boundary whose admin should decide its own posture rather than inherit one from a parent it was created to be separate from.

### The value is replicated — no, it is not

**This section previously claimed the opposite, and the claim was false.** It read:

> a joiner adopts a namespace's `default_capabilities` verbatim from the join bundle, so two peers built at different versions cannot disagree about the mask

True for a **joiner**, false in general — and I wrote it without checking the paths that are not joins. Four sites seed a namespace's `default_capabilities` with a hardcoded `CAN_JOIN_OPEN_SUBGROUPS` when none is set:

| site | when it runs |
| --- | --- |
| `crates/governance-store/src/ops/namespace/namespace_created.rs:482` | genesis establish |
| `crates/governance-store/src/ops/namespace/namespace_created.rs:288` | genesis re-arrival |
| `crates/governance-store/src/namespace/governance.rs:1368` | replica bootstrap |
| `crates/context/src/handlers/store_group_meta.rs:86` | gossiped group meta |

Before this change the creator seeded the same value, so all five agreed. This change makes the creator seed `516` and leaves the other four at `4`.

Joiners are unaffected, deliberately: `namespace_sync.rs:882` puts the responder's real value in the join bundle and `join_group.rs:359` writes it *before* catch-up ops, so genesis's `is_none()` gate leaves it alone. That ordering is why `group-default-capabilities-propagation` passes and why CI did not catch this.

The gap is a peer that materialises a namespace **without** a join bundle — genesis apply, replica bootstrap, gossiped meta. It seeds `4` and every `is_none()` gate keeps it there. `add_member_with_keys` copies the mask at admission, so that peer materialises member rows without bit 9 while the creator materialises them with it; `account_may_author` then disagrees across peers about the same relay, and a delegated write accepted at one is dropped at the cut by the other.

**This must be fixed before merge.** Three options, discussed in [comment 5712442071](https://github.com/calimero-network/core/pull/3969#issuecomment-5712442071): share the rule with the genesis sites (smallest, but splits old/new peers on a rolling upgrade); publish a `DefaultCapabilitiesSet` op at creation (preferred — travels the DAG as an op old peers already understand); or carry the mask in `NamespaceCreated` (a wire change).

**What this widens, stated rather than buried.** It reaches every non-admin member of the namespace, not only attested TEE nodes. What it does *not* confer is the ability to **forge**: a delegated write is authorized by the *author's* warrant — signed by their device key, committed to one context, one method and the exact arguments, nonce checked unspent, both signature layers re-verified by every peer at the cut. A holder can only spend a warrant it was deliberately handed; what it gains is the arguments in cleartext and the choice of when, or whether, to publish. Admins never receive it at all (`add_member` seeds the default for non-admin roles only), so a creator's own node still needs an explicit grant.

It is an **initial value, not a rule**: `GroupOp::DefaultCapabilitiesSet` changes it afterwards and nothing re-asserts it.

### Also: meroctl could not express this bit at all

`group settings set-default-capabilities` exposed **three** of the eight flags its sibling `group members set-capabilities` has, and built the mask from zero — so the one command that could open a namespace to a fleet had no flag for the capability, and running it to set *any* flag silently cleared every bit it had no flag for. All eight now, through a by-name encoder like its sibling's (a positional `bool` list is what makes a transposition compile, pass, and grant the wrong capability), plus a `long_about` stating that it **replaces** the mask and is **not retroactive**.

### Also: the delegated-session scenario now covers the WRITE leg

`delegated-session.yml` proved two thirds of epic #3928's criterion — obtain a session against a node you do not own, read a context you are a member of. The third leg, *submit a delegated write*, had **never run from a password-free session**. `delegated-authorship.yml` exercises the warrant mechanics exhaustively, but it logs in with `dev`/`dev-password`, so what it demonstrates is a delegated write by an already-privileged caller. The criterion is the two halves joined.

Added to the same scenario and the same script, so the token spent by the write is the one minted from a signed statement in step 3:

- **5a** — the device mints a warrant offline via `merod account warrant`, naming node-1's own account as executor. One `INTENT_ARGS` variable feeds both the warrant and the request body, because the warrant commits to `H(method ‖ args)`: re-typing that JSON with different spacing mints a warrant that verifies as a signature and is refused as authorization.
- **5b** — the keyholder spends it at `POST .../intents` on its own `Bearer` token. Asserts on `rootHash`, which is present only when the intent executed and advanced state.
- **5c** — reads the value back through the delegated read path, not an admin surface the keyholder does not have.
- **5d** — replays the identical bytes and asserts refusal, so the nonce is proven consumed.
- Scenario level: a `wait_for_sync` and a `json_assert` on node-2, so the write is proven to be a real delta a peer re-verifying both signature layers accepted.

`node_identity` now yields both `publicKey` and `accountId` from one call rather than two, so the key the statement is addressed to and the account the warrant names cannot describe different nodes. The `CAN_AUTHOR_ON_BEHALF` grant in the scenario is **explicit rather than leaning on this PR's new default**: node-1 created the namespace so it is the root admin, and the default mask is seeded for non-admin roles only, so the default would not reach it; and a scenario depending on the default would stop testing the intents path the day someone changed the default.

The script states what it does **not** assert: that the delta is attributed to the author rather than the relay. That is `Principal`'s job and it is real, but it is not observable from the two HTTP responses — the author lives on the delta, not in the intent reply. It is covered by the `Principal` unit tests and by `delegated-authorship.yml`'s DAG assertions.

## Test plan

```bash
cargo test -p calimero-context --lib create_group     # 15 passed
cargo test -p meroctl --bins settings                 #  3 passed
cargo fmt --check                                     # clean
```

All green locally. Clippy was run as the full CI invocation, not a per-crate narrowing.

New tests:

- `a_namespace_is_created_open_to_delegated_authorship` — pins the asymmetry **from both sides**. A test checking only the namespace would pass just as happily if every group were opened.
- `open_subgroup_inheritance_survives_in_both_arms` — losing `CAN_JOIN_OPEN_SUBGROUPS` is silent and late.
- `no_other_capability_is_seeded` — `count_ones() == 2` plus an explicit sweep.
- meroctl: `each_flag_sets_exactly_its_own_bit`, `the_fleet_mask_carries_both_bits`, `no_flags_is_an_empty_mask`.

**These tests all pass and did not catch the defect above**, because each exercises `create_group` in isolation — which is exactly the creator path that is correct. A regression test for the divergence has to involve a second peer that never joins, and it ships with the fix.

**Not run locally, and why:** the `merod --features mock-attestation` clippy pass and the merobox E2E suite. This sandbox has a fixed writable-disk allowance that a second full-feature build does not fit in, and two build scripts (`crates/server`, `crates/auth`) fetch release archives over TLS with bundled webpki roots, so they cannot verify the environment's proxy CA — worked around for the runs above with their supported `CALIMERO_WEBUI_SRC` / `CALIMERO_AUTH_FRONTEND_SRC` escape hatches.

## Proof the fix works

Not a bug fix — a posture change — so there is no before/after log. The behavioral claim is pinned by the tests above, asserted through the same path production uses: `governance-store`'s existing `a_groups_default_capabilities_reach_an_admitted_tee_node` proves a default mask reaches an admitted TEE node **via `account_may_author`**; this PR is what puts the bit in that mask at creation. What is **not** yet proven is that every peer agrees about that mask — see above.

## A note on the merobox pin, so the history reads straight

`b46f838df` pinned merobox 0.6.76 here, ported from #3968 to remove a variable from an unrelated `ephemeral-presence-e2e` failure. That was **half of a coherent change**: 0.6.76 goes with that PR's scenario rewrites, and the pin alone broke three account/revocation scenarios. Reverted in `19b1aa123`. #3968 has since merged, so `f135c237e` merges master and this branch inherits 0.6.76 **with** its rewrites.

That made this the first single-variable run for `ephemeral-presence-e2e`, and **it passed** — so that failure tracked the harness version, not this change, correcting what I said in comment 5711718157. It is one green run.

## Wire contract (SDK gate)

No HTTP wire DTO or route changed. No fixture or manifest regeneration needed.

- [ ] ~~Regenerated wire fixtures~~ — n/a, no DTO changed
- [ ] ~~Updated `crates/server/endpoints.json`~~ — n/a, no route changed
- [x] Linked the matching mero-js PR — calimero-network/mero-js#154 adds `AdminClient.openToDelegatedExecution()` / `grantAuthorship()` over routes that already exist.

## Documentation update

Done in this PR: `docs/src/content/docs/protocol/delegated-authorship.mdx` — namespaces are created carrying the bit, the three limits (namespaces only; not retroactive and never for admins; reaches every non-admin member), and the "initial value, not a rule" framing with the `meroctl` command to close a namespace again. **That page needs a pass with the fix**, since it currently inherits the replication claim corrected above.

---


chefsale:
**CI: `Test (ephemeral-presence-e2e)` failed on `b1fca8419` — I do not believe it is this PR's.** Recording the reasoning rather than re-running silently.

The scenario failed one assertion out of 21:

```
=== 20 passed, 1 failed ===
FAIL node 2 SSE subscriber received node 1 presence over gossip:
     SSE: no live delta carrying the published slice within 15000ms — gossip not delivered over SSE
```

What passed alongside it is what makes this look like a transport-timing flake rather than a regression:

- `ok  node 2 subscriber received node 1 presence over gossip` — the **WS** subscriber received the *same* presence event.
- `ok  node 2 accepted the SSE subscription (production transport)` — the SSE stream was established.
- `ok  SSE subscriber was seeded with the existing presence` / `ok  SSE seed carries ageMs (got: 2425)` — Phase 3b, SSE delivery working.
- Phase 4's load-bearing no-DAG-growth guard passed.

So SSE is up and delivering; only the *live* delta missed its 15s window, on one transport, for one event the other transport received.

**Why it is not plausibly this diff.** This PR changes three files: the default-capability mask a namespace is *created* with (`create_group.rs`), `meroctl`'s `set-default-capabilities` flags, and one docs page. It touches nothing in the server, the SSE/WS subscription layer, or gossip. `ephemeral-presence-e2e` never reads or sets capabilities — `grep -n "capabilit" apps/kv-store/workflows/ephemeral-presence-e2e.yml apps/kv-store/scripts/ephemeral-presence-e2e.*` returns nothing — so the one behaviour this PR changes is not on the path the failing assertion exercises. Every governance gate in this same job passed, including the projection-divergence hard gate (`2 concluded comparison(s), result="agree"`) and the op-store completeness gate.

**Re-run:** attempted, and refused with `403 This workflow is already running` — run 35194491892 still has jobs in flight. I will re-run the failed job once the run completes, and if it fails a second time I will treat it as real and root-cause it rather than re-running again.

I am not proposing a fix here: the assertion does look timing-sensitive (a 15s deadline on live SSE delivery), but making it robust means changing a scenario this PR has no other reason to touch, and widening the PR to do that would be wrong unless the second run says otherwise.

---

chefsale:
**Follow-up: the re-run failed the same way. This is deterministic, not a flake — correcting my previous comment.**

I said above that if it failed a second time I would treat it as real. It did (job `105124924268`, same head `b1fca8419`, same single assertion), so here is the root-cause work rather than another re-run. I have **not** spent a second re-run and do not intend to.

### Three candidate mechanisms linking it to this PR, all refuted

This PR changes the capability **value** seeded into a non-admin member's row at admission (`CAN_JOIN_OPEN_SUBGROUPS | CAN_AUTHOR_ON_BEHALF` = 516, was 4). I looked for a path from that to a dropped SSE delta:

1. **The scenario asserts on a capability bitmask.** No — `grep -n "capabilit" apps/kv-store/workflows/ephemeral-presence-e2e.yml apps/kv-store/scripts/ephemeral-presence-e2e.*` returns nothing. It never reads or sets capabilities.
2. **The extra bit triggers an SSE re-authorization prune.** No — `prune_stale_grants` is driven by `NodeEvent::GroupMembership` (`crates/server/src/sse/events.rs:115-135`). This PR changes no membership event, only the u32 written alongside one; a *wider* mask cannot revoke a grant, and `CAN_AUTHOR_ON_BEHALF` gates nothing about subscriptions. Decisively, the **WS** subscriber goes through the same prune path and received the delta.
3. **SSE serialises the live delta differently** (the predicate requires `ageMs == null`, and SSE *seeds* demonstrably carry `ageMs: 2425`). No — `age_ms: None` is set on the live path in the **node** layer, `crates/node/src/handlers/ephemeral/inbound.rs:300,306`, before either transport. The payload is identical for WS and SSE.

I also ruled out a script-ordering race: the WS `waitFor` is awaited before the SSE one, but the SSE watcher replays already-buffered events (`events.find(pred)` in `ephemeral-transport.js:254`), so a delta arriving during the WS wait is still matched.

So the failure sits in **SSE live delivery for ephemeral presence** — the payload is built, the WS subscriber gets it, the SSE stream is open (its seed assertions pass in Phase 3b), and the live delta does not arrive within 15s.

### What I could not establish, stated plainly

**Whether this reproduces on `master`.** That is the check that would settle ownership, and I could not run it: the tooling available to me cannot surface a recent `master` run of `E2E - Rust Apps` (workflow/branch filters on run listings return unrelated runs), and the raw job-log storage is unreachable from my environment. So I am **not** claiming this is definitively not this PR's — I am reporting that three concrete causal hypotheses are refuted by the code and I found no fourth.

Supporting, not conclusive: 114 of 116 jobs on this exact commit passed, including `App migration e2e` and `SDK E2E (paired)`; the only other failure is `Report Failed Workflow`, the downstream reporter. (The `E2E Rust Apps Failed` bot comment naming *scaffolding-e2e, xcall-example* is boilerplate — every job in both groups passed.)

### What would settle it, for someone with the access I lack

1. Run `ephemeral-presence-e2e` on `master` at `21bd859`. If it fails there, this is pre-existing and not this PR's.
2. If it passes on master, the next probe is the SSE event task on node 2: `docker-logs/ephemeral-presence-e2e-ephemeral-node-2.log` around the `set_ephemeral` timestamp, checking whether the context event reached the SSE session's subscription set at all (`crates/server/src/sse/events.rs:120-124` drops events for contexts the session does not watch).

**I am not pushing a fix.** Loosening or re-ordering the assertion would be editing a scenario this PR has no other reason to touch, and doing that on a failure I cannot attribute would bury a possibly-real SSE delivery bug rather than fix it. I would rather this blocks the PR and gets a correct answer.

---

chefsale:
**Correction: I was wrong. The evidence now points at this PR.**

My two comments above argued this failure was not plausibly caused by this diff. A controlled comparison has since become available and it contradicts that:

| branch | contains this change | `Test (ephemeral-presence-e2e)` |
| --- | --- | --- |
| `test/account-scenarios-native-steps` @ `b4f0637bc` | **no** | **passed** (run 35197607387, 0 failures across 116 jobs) |
| this PR @ `b1fca8419` | **yes** | **failed twice** |

Both branch from the same base — master `21bd8594` — and ran on the same infrastructure minutes apart. `git merge-base --is-ancestor b1fca8419 b4f0637bc` confirms the passing branch does not carry this commit. That is as close to a controlled experiment as this repo offers, and the only variable is this one commit.

So: **this is a regression in this PR**, not a pre-existing failure and not a flake. My earlier reasoning — "three candidate mechanisms refuted, therefore not this diff" — was the wrong shape of argument. Refuting the mechanisms I happened to think of is not evidence of absence, and I presented it with more confidence than it earned. The comparison above is the check I should have found before commenting at all; I had claimed it was unavailable, when in fact any concurrent PR's E2E run serves the same purpose.

**I have not yet found the mechanism.** What the change does is narrow: a namespace is now created with `default_capabilities = CAN_JOIN_OPEN_SUBGROUPS | CAN_AUTHOR_ON_BEHALF` (516) instead of `CAN_JOIN_OPEN_SUBGROUPS` (4), and `add_member` copies that into every non-admin member's capability row at admission. In this scenario that means node 2's row in the namespace is 516 rather than 4. Why that stops a *live SSE* presence delta arriving within 15s while the WS subscriber receives the same event, I cannot yet explain — and I am not going to invent an explanation.

The lead I would pursue next, stated so it is not lost: `store_group_meta.rs:81-90` seeds a group's `default_capabilities` with a **hardcoded** `CAN_JOIN_OPEN_SUBGROUPS` when the gossiped meta arrives before the real value does. With this change the authoritative value is no longer that constant, so a peer can briefly seed member rows from 4 while the creator uses 516. My PR description asserts this cannot happen because the value is "replicated, not recomputed" via the join bundle — that claim is now suspect, and this hardcoded fallback is the specific place it may not hold.

I am leaving the PR red rather than pushing a speculative fix or touching the scenario. Anyone picking this up should treat the PR description's replication argument as unverified.

---

chefsale:
**Correcting my correction: the comparison above is not controlled, and I should not have called it one.**

My previous comment presented `test/account-scenarios-native-steps` passing vs this branch failing as a single-variable experiment and concluded "this is a regression in this PR". That was wrong. Both branches do sit on master `21bd8594`, but that branch **also bumps merobox 0.6.75 → 0.6.76** (`.github/actions/setup-merobox/action.yml`, `.github/workflows/e2e-rust-apps.yml`). Two variables, not one.

The harness difference is not cosmetic, and it is visible in the artifacts:

| run | merobox | container logs uploaded |
| --- | --- | --- |
| this PR @ `b1fca8419` (failed) | 0.6.75 | **4** — `node-1`, `node-1-init`, `node-2`, `node-2-init` |
| `b4f0637bc` (passed) | 0.6.76 | **2** — `node-1`, `node-2` |

0.6.75 starts `-init` containers for each node; 0.6.76 does not. The failing assertion is a 15-second deadline for a *live* delta on an SSE stream whose seed assertions pass — node startup and readiness differences are exactly the kind of thing that moves it. That is now a live candidate alongside this PR's own change, and **neither is established**.

So I have made the opposite error twice: first asserting the failure was not this PR's on the strength of mechanisms I had personally failed to think of, then asserting it *was* on the strength of a comparison I had not checked for confounds. The honest position is that I do not know, and I should have said so both times.

**What I pushed** (`b46f838df`): the merobox 0.6.76 bump, ported from the branch it comes from. This is not a claim that the harness is at fault — it removes the one variable I can remove, so the next run has a single cause to attribute rather than two. It no-ops once master carries the same bump.

**How to read the next run:**

- still fails on 0.6.76 → the remaining variable is this PR's commit, the regression is mine, and the `store_group_meta.rs:81-90` hardcoded-fallback lead from my previous comment is where I would start;
- passes on 0.6.76 → the failure tracked the harness version, and this PR's change is exonerated — but only once, so I would want it green twice before trusting it.

I have still not touched the scenario or its assertions, and I will not until the cause is known.

---

chefsale:
**Reverted the merobox bump — it was half a change, and it broke three green scenarios.**

`b46f838df` pinned merobox 0.6.76, ported from `test/account-scenarios-native-steps` to remove the second variable from the `ephemeral-presence-e2e` comparison. On that head, three scenarios that were passing went red:

- `Test (account-device-revoke-lockout)`
- `Test (account-lost-holder-revoked-everywhere)`
- `Test (group-remove-from-root-revokes-inherited)`

All three are files that the branch I took the bump from **also rewrites** (`account-device-revoke-lockout.yml`, `account-lost-holder-revoked-everywhere.yml`, `group-remove-from-root-revokes-inherited.yml` are in its diff). 0.6.76 changes step semantics those scenarios depend on — the version bump and the scenario updates are one coherent change, and I took only the version half.

Porting the other half is not the answer: 17 files of someone else's in-flight work, well past this PR's subject, and guaranteed to conflict when their PR merges. 0.6.76's proper home is that branch; this one inherits it from master once it lands.

The experiment never returned a verdict anyway — the run was cancelled by a later push before `ephemeral-presence-e2e` reported on 0.6.76 — so reverting costs nothing. That scenario returns to failing on 0.6.75, which is the honest state: **unresolved**, rather than obscured by three new failures this PR introduced.

Reverted in `19b1aa123`. The open question from my previous comment stands unchanged: whether `ephemeral-presence-e2e`'s live-SSE timeout is this PR's change or the harness. It needs a run of that scenario on master at `21bd8594`, which I cannot trigger from here.

Also on this branch now: `a723e382e`, which closes the epic's third acceptance leg — the delegated WRITE on the password-free session. That is unrelated to the failure above and stands on its own.

---

chefsale:
Merged master (`55c24dcc0`) in — `f135c237e`.

#3968 landed, which is exactly where I said merobox 0.6.76 belonged: *"The right home for 0.6.76 is that branch; this one inherits it from master once it lands."* This branch now carries 0.6.76 **together with** the scenario rewrites that go with it, rather than the pin alone, which is what broke `account-device-revoke-lockout`, `account-lost-holder-revoked-everywhere` and `group-remove-from-root-revokes-inherited` when I ported half of it in `b46f838df` and reverted in `19b1aa123`.

The merge was clean and touched none of this PR's files. `account-api.sh` lost 119 lines to #3968 but kept `fail`, `node_url` and `offline_merod` with unchanged signatures, which is all `delegated-session.sh` sources from it.

**On `ephemeral-presence-e2e`, stated plainly because I have got this wrong twice.** First I claimed it was "not plausibly this diff" by refuting three mechanisms I happened to think of — evidence of absence from a search I designed. Then I claimed a controlled comparison proved it *was* this diff — but that comparison branch also bumped merobox, so it was two variables, not one. This run is the first single-variable one: same merobox and same scenario definitions as master, with only this PR's commits on top. If `ephemeral-presence-e2e` still fails here, the remaining variable is my own commit and I will treat it as mine.

This run is also the first execution of the new delegated-write leg (`a723e382e`).

---

chefsale:
Two things from the run on `f135c237e`, and they point opposite ways.

## `ephemeral-presence-e2e` passed

This was the first single-variable run — same merobox and same scenario definitions as master, only this PR's commits on top — and it is green. So that failure tracked the harness version, not this change. That **closes the open question in comment 5711718157**, where I said the opposite.

It is one green run, so the confidence is limited accordingly. The three scenarios I broke by porting half of #3968 (`account-device-revoke-lockout`, `account-lost-holder-revoked-everywhere`, `group-remove-from-root-revokes-inherited`) are green too.

## The two red checks are an install failure, not a test failure

`Test (group-subgroup-cold-sync)` and `Test (group-member-channel-lifecycle)`. The first one's log says it plainly, in the step that fetches the pinned merobox asset:

```
curl: (22) The requested URL returned error: 500
curl: (22) The requested URL returned error: 500
curl: (22) The requested URL returned error: 500
curl: (22) The requested URL returned error: 500
##[error]Process completed with exit code 22.
```

Four retries, four 500s, exit 22 — before any node started. Both jobs then show `docker-logs/` missing, "this scenario folded no membership op", and 0 concluded comparisons. No `merod` code ran at all, so no diff could have produced this. I read the error line directly for `group-subgroup-cold-sync`; for `group-member-channel-lifecycle` I have the identical shape (~25s, no `docker-logs/`, every post-scenario gate reporting on nothing) but did not see its own `curl` line, and I am flagging that rather than asserting it.

Over 100 scenarios in the same run passed, including the ones that would actually break if this change were wrong: `group-default-capabilities-propagation`, `group-capabilities`, `group-subgroup`, `group-subgroup-visibility-inheritance`, `group-membership`, `delegated-authorship`, `namespace-owner-writes-in-a-foreign-open-subgroup`.

`rerun_failed_jobs` is refused with *"This workflow is already running"* while the rest of the run finishes. I will spend the one re-run once it completes.

## A real bug in this PR, which CI did not catch

Separately — and this one **is** mine. This PR's description claims:

> a joiner adopts a namespace's `default_capabilities` verbatim from the join bundle, so two peers built at different versions cannot disagree about the mask

That is true for a **joiner** and false in general, and I wrote it without checking the other paths. Four sites seed a namespace's `default_capabilities` with a hardcoded `CAN_JOIN_OPEN_SUBGROUPS` when none is set:

| site | when it runs |
| --- | --- |
| `crates/governance-store/src/ops/namespace/namespace_created.rs:482` | genesis establish |
| `crates/governance-store/src/ops/namespace/namespace_created.rs:288` | genesis re-arrival |
| `crates/governance-store/src/namespace/governance.rs:1368` | replica bootstrap |
| `crates/context/src/handlers/store_group_meta.rs:86` | gossiped group meta |

Before this PR the creator also seeded `CAN_JOIN_OPEN_SUBGROUPS`, so all five agreed. **This PR makes the creator seed `CAN_JOIN_OPEN_SUBGROUPS | CAN_AUTHOR_ON_BEHALF` (516) and leaves the other four at 4.**

Joiners are fine, and deliberately so: `namespace_sync.rs:882` puts the responder's real value in the join bundle, and `join_group.rs:359` writes it *before* applying catch-up ops, so genesis's `is_none()` gate then leaves 516 alone. That ordering is why `group-default-capabilities-propagation` passes and why this did not show up in CI.

The gap is a peer that materialises a namespace **without** a join bundle — genesis apply, replica bootstrap, or gossiped meta. It seeds 4, and every `is_none()` gate keeps it there permanently. `add_member_with_keys` copies the mask at admission, so that peer materialises member rows *without* bit 9 while the creator materialises them *with* it. `account_may_author` then answers differently on the two peers about the same relay, and a delegated write accepted at one is dropped at the cut by the other.

So the "replicated, not recomputed" argument in the description does not hold as written. I am correcting the description now, and not merging this until the seed comes from one place. The options differ enough to be worth a decision rather than a quiet pick:

1. **Share the rule** — have the genesis/bootstrap sites call the same `initial_default_capabilities(true)`. Smallest diff, but it is *recomputation*: an old peer computes 4 and a new one 516, so a rolling upgrade gets exactly the split above. `CLAUDE.md` names this class as needing its own migration path, and merobox cannot see it.
2. **Publish a `DefaultCapabilitiesSet` op at namespace creation.** The value then travels the DAG as an explicit op that old peers already understand, and every `is_none()` gate is respected because that op writes unconditionally. One extra governance op at creation.
3. **Carry the mask in `NamespaceCreated` itself.** Genuinely replicated, but a wire change with its own migration.

My preference is (2): it is the only one that is correct across mixed versions without a new wire format, and it uses replication the system already has rather than adding a second way for the value to travel.

---

chefsale:
The run on `f135c237e` finished. Reading it in full rather than by the red badge:

## `Test (delegated-session)` passed

First execution ever of the write leg — steps 5a–5d, plus the peer read-back. So epic #3928's third criterion now runs end to end in CI: a device holding only an account root and a device key obtains a session against a node it does not own, reads a context, and submits a delegated write that a *second* node accepts after re-verifying both signature layers. No password anywhere in the flow.

## The red is still only the two install failures

`Test (group-subgroup-cold-sync)` and `Test (group-member-channel-lifecycle)` — `curl: (22) error: 500` four times fetching the pinned merobox asset, exit 22, before any node started. Diagnosed in [5712442071](https://github.com/calimero-network/core/pull/3969#issuecomment-5712442071); nothing has changed about that reading. `Report Failed Workflow` is the reporter for those two, not an independent failure.

Everything else on that run is green, including every check that would break if this change were wrong: `group-default-capabilities-propagation`, `group-capabilities`, `group-subgroup`, `group-subgroup-visibility-inheritance`, `group-membership`, `group-governance-*`, `delegated-authorship`, `namespace-owner-writes-in-a-foreign-open-subgroup`, plus `Rust`, `Lint`, `ABI Contract`, `SDK E2E (paired)`, the migration matrix and both Windows e2e jobs.

## Pushed a base merge instead of spending the re-run

The PR had gone `behind` — master moved to `50139073b` (#3884, blob presence on HEAD). `5fe53f4e7` merges it. That was needed on its own, and it starts a fresh run, which re-attempts the asset fetch without burning the one re-run this PR is allowed. `rerun_failed_jobs` was refused anyway while `Windows e2e (3-node sync)` was still going in that run.

The merge was clean and touched none of this PR's five files; #3884 is confined to blob handlers, node primitives and admin-api docs.

## Still not mergeable, and not because of CI

The propagation defect in *The value is replicated — no, it is not* is unfixed. Since writing that I found it is worse than stated: `governance.rs:1253 seed_bootstrap_admin_if_absent` says in its own comment that a **TEE replica** bootstrapping the namespace root admits its own `MemberJoinedViaTeeAttestation` row snapshotting the group's default caps at apply time. That is the attested relay this whole change exists to serve, and it is seeded from the hardcoded `4` — so for a relay admitted down that path the capability never arrives at all.

The fix is agreed in shape (publish a `DefaultCapabilitiesSet` op at namespace creation, so the value travels the DAG and causally precedes every admission) but needs a second part, because that bootstrap path deliberately runs ahead of the gossip it is named for. It ships with a regression test involving a peer that never joins.

---

PR #3972, opened 2026-09-17
feat(governance,server): discover an account's applications outside a device's scope

rtb-12:
## Summary

Stacked on #3977, which takes schema version 14 and store prefix `0x52`.

A device paired with an application scope only follows namespaces whose target application its scope covers, so an application its account speaks in elsewhere is invisible to it. `GET /admin-api/account/applications` made that worse by deriving its answer from this node's LOCAL participation set, which by construction can never contain such a namespace - and the account's namespace set carried no registry coordinates, so even knowing about the application there was nowhere to install it from.

- **New appended op `GroupOp::AccountNamespaceTargetNamed { namespace, application, package, version }`**, published into the ACCOUNT NAMESPACE under the same authority and gate as `AccountNamespaceGained`. `SIGNED_GROUP_OP_SCHEMA_VERSION` goes 14 -> 15: the variant is appended so no prior ordinal moves and a v14 peer fails at the version gate rather than partway through a DAG. Coordinates are bounded by the existing `check_coords` / `bounds::MAX_COORD_BYTES` before apply, and an empty pair is rejected there.
- **Applied onto a new key rather than a wider value.** `GroupAccountNamespace` rows are bare borsh `Option<ApplicationId>`; widening that value type would fail to decode every row already on disk. The coordinates go under `GroupAccountNamespaceTarget` (prefix `0x53`) beside it. The apply only writes for a namespace the set already names - one it left or never gained is a no-op, not an error - and `AccountNamespaceLeft` drops both rows.
- **Two publish sites, because a gain is never restated.** `account_namespace::publish` names the coordinates straight after a Gained the set took; and `account_follow` re-announces them when it folds a `TargetApplicationSet` (a new `OpEvent`) that left the recorded pair stale. That staleness check *is* the single-publisher rule: every device of the account folds the same op, whichever lands first silences the rest as they fold the result, and a concurrent second publish writes the same pair. A node with no identity in the account namespace skips silently.
- **`GET /admin-api/account/applications` now reads the account's set**, unioned with local participation for anything the set has not learned. Entries keep `applicationId` and `namespaces` unchanged and gain optional `package` / `version` plus a `followed` boolean. A node holding no account still answers `None`.

The route manifest records routes, not response shapes (see the header of `crates/server/tests/route_manifest.rs`), so no regeneration is needed.

## Test plan

- `governance-types`: the appended op round-trips at ordinal 32 with the right `op_kind_label`, and its coordinates are rejected when empty or over `MAX_COORD_BYTES`; the golden-ordinal test's version assertion moves to 15.
- `governance-store`: the set records, replaces and drops coordinates with the row it belongs to; through the real apply pipeline, coordinates attach only to a namespace the account holds, an admin-at-the-cut bound to no account here renames nothing, and a leave drops them.
- `context`: coordinates need both a target and a real registry pair (the raw-wasm placeholder is refused); a gain names where the namespace is published; a gain made before the coordinates folded names none and the refresh supplies them, after which nothing re-announces; a namespace the account never gained is never refreshed.
- `server`: an application outside this device's scope is listed unfollowed with its coordinates beside one the device does follow; a namespace the account set never named keeps the previous behaviour.
- e2e `account-application-outside-scope`: node-2 pairs scoped to app A while the account also holds a namespace for app B, and node-2's route lists app B with the package and version node-1's installed row carries and `followed: false`, and app A with `followed: true`. Registered in `e2e-rust-apps.yml` beside the other account scenarios.


rtb-12:
Closing: the intent was a consent step before an in-scope application is installed on a second device, not a listing of applications outside a device's scope. A narrowed device not learning about the rest of the account is the cleaner behaviour. The consent step is being built as a node setting that stops automatic first installs.

PR #4007, opened 2026-09-21
fix(server): answer 404 when the caller names something this node does not have

### LINKED ISSUE #3996: 'group not found' returns HTTP 500 instead of 404, so callers read a real failure as 'already left'
## Summary

"group not found" is raised as a bare `eyre` error and surfaces as **HTTP 500**, not 404:

```rust
// crates/context/src/handlers/get_member_capabilities.rs:19-23
if MetaRepository::new(&self.datastore).load(&group_id)?.is_none() {
    bail!("group '{group_id:?}' not found");
}
```

`bail!` produces an untyped error, and `parse_api_error` has nothing to map, so a perfectly ordinary "you asked about something that is not here" is reported as a server fault.

## Impact

Two concrete harms, both observed during the fleet-HA re-admission incident:

**1. A control-plane script read the failure as success.** The fleet sidecar's wrapper around namespace leave saw the 500 and logged:

```
namespace leave for f72d… returned rc=101 (likely already left / not a member)
```

It treated a server error as "already gone" and continued. The real condition — the group genuinely is not present on that node — was never reported. A 404 is what that code was written to interpret; a 500 is not.

**2. The dashboard hot-loops on it.** Something polls member capabilities roughly every 1.4 s, and each call logs an `ERROR` with a stack location:

```
ERROR … group 'ContextGroupId(…f72d…)' not found at crates/context/src/handlers/get_member_capabilities.rs:22
```

On a fleet node that journal is shipped to a central store, so a routine UI poll against an absent group becomes a permanent error stream — noise in exactly the place someone looks when diagnosing a real fault.

Neither is a correctness bug in the handler. Both are consequences of the status code lying about the category of the failure.

## Steps to reproduce

1. Call `GET /admin-api/groups/<any-unknown-id>/members/<id>/capabilities` on any node.
2. Observe HTTP 500 and an `ERROR`-level log line.
3. Expected: 404, and not at `ERROR` — a client asking about an absent resource is not a server fault.

## Criteria for resolving

- [ ] "group not found" returns **404**.
- [ ] "identity is not a member of group" — the next `bail!` in the same handler — gets the same treatment; it is the same category.
- [ ] Neither logs at `ERROR`. A 4xx is a client-shaped outcome; on a fleet node an `ERROR` per poll is shipped to the operator's log store.
- [ ] The other `bail!("… not found")` sites reachable from the admin API are audited for the same mapping, rather than this one being fixed in isolation.
- [ ] A test asserts the status code. It must fail against the current code.

Found while diagnosing the fleet-HA disable→re-enable incident, where harm (1) hid the real failure.

chefsale:
Closes #3996.

## The defect

`"group not found"` was a bare `bail!`, so `parse_api_error` had nothing to match and it reached the caller as `500 {"error":"Internal server error"}`. A perfectly ordinary "what you asked about is not here" was reported as a server fault.

Two things went wrong with that during the fleet-HA incident:

**A control-plane script read the failure as success.** The fleet sidecar's wrapper around namespace leave saw the 500 and logged `rc=101 (likely already left / not a member)` — treating a server error as "already gone" and continuing past a real failure. That code was written to interpret a 404.

**The dashboard hot-loops on it.** Something polls member capabilities about every 1.4 s, each call logging an `ERROR` with a stack location. On a fleet node that journal ships to a central store, so a routine UI poll against an absent group became a permanent error stream — noise in exactly the place someone looks when diagnosing a real fault.

Neither is a correctness bug in the handler. Both follow from the status code lying about the *category* of the failure.

## The fix

Four typed variants — `GroupNotFound`, `NamespaceNotFound`, `ApplicationNotFound`, `ContextNotFound` — map to **404**, keeping their message (which carries only the id the caller supplied, so echoing it leaks nothing).

The identity-not-a-member case reuses `MembershipError::MemberNotFound`, already mapped to 404.

`NotANamespaceMember` joins `NotAGroupMember` at **403**: the group exists, the caller just has no standing in it. Answering 404 there would send them looking for something that is right in front of them.

## The audit

The issue asked for the other `bail!("… not found")` sites reachable from the admin API to be audited rather than this one fixed in isolation. All of them are converted:

`get_member_capabilities` (both), `get_group_info` (both), `delete_group`, `delete_namespace`, `create_context`, `join_context`, `create_group_invitation`, `retry_group_upgrade`, `create_group`, `update_application` (both), `upgrade_group` (7 sites), `get_group_upgrade_status`, `get_cascade_status`, `lib.rs`.

**Two deliberately left alone**, both because 404 would be the same kind of lie in the other direction:

- `sync_group`'s *"group not found locally; wait for P2P replication"* is a retry-later condition, not an absent resource. A 404 tells a client to stop when it should retry — the reasoning `AuthorityNotYetResolvable` already uses for its 503.
- `execute`'s *"bytecode blob not found in blobstore"* is genuinely this node's state being broken. 500 is the honest answer.

## The log level

`ApiError` gains `is_client_fault()`, and the handlers that can now legitimately answer 4xx pick `debug` over `error` with it. Scoped to those handlers — 82 files use this `error!`-on-any-failure pattern and sweeping them all is a different change.

The *other* `ERROR` in the reported line came from `parse_api_error`'s `"unhandled admin-api error"` fallback (that's where the `at crates/context/…:22` suffix came from — eyre's location in the `Debug` render). It no longer fires for these, now that they're typed.

## Tests

Each status code, plus that "not a member" stays 403. And `an_untyped_not_found_still_falls_through_to_500` pins **why typing was needed at all** — it asserts the untyped form these sites used still returns a scrubbed 500, which is what every one of them did before this change.

Locally: 379 + 215 pass across `calimero-server` and `calimero-context`, clippy `-D warnings` clean.

---

> [!NOTE]
> **Medium Risk**
> Changes HTTP status semantics for many admin API paths; callers that treated 500 as success or retry may behave differently, though that is the intended fix.
> 
> **Overview**
> Replaces untyped `"not found"` / membership `bail!` strings across context handlers with **`ContextError`** variants so the admin API can answer with the right HTTP status instead of a generic **500**.
> 
> **404** for resources this node does not have: `GroupNotFound`, `NamespaceNotFound`, `ApplicationNotFound`, `ContextNotFound`, plus **`MemberNotFound`** when querying capabilities for a non-member. **403** for **`NotAGroupMember`** and the new **`NotANamespaceMember`** (resource may exist; this node has no standing).
> 
> **`parse_api_error`** maps those types explicitly; **`ApiError::is_client_fault()`** lets affected group admin handlers log client failures at **`debug`** rather than **`error`**, cutting noise from polling absent groups in central logs.
> 
> Tests pin 404 vs 403 behavior and that untyped not-found messages still fall through to scrubbed 500s.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39cb2b88bf005ab880318e1c277c43c37076bfde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chefsale:
Both Bugbot findings were correct and are fixed in 0d996f9.

## 1. Wrong principal named — confirmed

`create_context` takes `identity_secret` **straight from the request body** and only falls back to the node's own key when the caller omits it, so the identity being checked is routinely somebody else's. `NotAGroupMember`'s "node is not a member of group X" named the wrong principal.

New `IdentityNotAGroupMember` carries the identity and says so.

I kept it at **403 rather than the suggested 404**, and the difference from `get_member_capabilities` is deliberate: there, the identity is the *subject* of a lookup, so "not found" is the honest answer. Here the caller holds the key and is acting **as** that identity, so the refusal is about standing, not about something being absent. The capability check on the very next lines refuses the same call for the same shape of reason; splitting them across 404 and 403 would be incoherent.

## 2. Audit incomplete — confirmed, and worse than reported

The audit had **two** blind spots, not one:

- `ok_or_eyre` was never searched — only `bail!` and `eyre!`.
- `eyre::eyre!(` with the message on the *following* line is invisible to a line-based grep, because no single line holds both the macro and the text. `create_group`'s parent-group miss is exactly that shape.

Re-ran it as a multi-line sweep over every not-found-shaped string literal in `crates/context/src`, classified by constructor. Beyond the three locations reported, that also turned up:

- `create_context` — application misses at two further call sites
- `update_application` — `"new application with id '…' not found in database"`
- **three member-mutation handlers** still answering 500 for a member the caller named: `set_member_capabilities`, `set_member_auto_follow`, `update_member_role`

Those three now take `MembershipError::MemberNotFound` → **404**, matching `get_member_capabilities` — the caller asked about somebody who is not in this group. That is the mirror image of finding 1, where the identity is the actor rather than the subject.

## Unchanged

The two documented exclusions stand, for the reasons already in the description: `sync_group`'s *"wait for P2P replication"* (retry-later, not absent — a 404 would tell a client to stop when it should retry) and `execute`'s missing bytecode blob (this node's state is genuinely broken, so 500 is honest).

## Verification

A test asserts the refusal names the identity **and** does not claim the node is the one without standing. 379 + 215 tests pass across `calimero-server` and `calimero-context`; clippy `-D warnings` clean.

